PR #15454, opened 2023-03-05
chore(misc): update lerna to 6.5.1

fahslaj:
Updates Lerna to 6.5.1, allowing Nx to make use of the new idempotency of `lerna publish`.


fahslaj:
Closing this for now, as the latest version of lerna don't export commands as independently importable code in js. I'll need to refactor nx-release.ts to call lerna with `exec` instead of directly through an import statement.

PR #15459, opened 2023-03-06
cleanup(misc): add known preset dependencies before generating it

leosvelperez:
## Current Behavior

The `ensurePackage` utility adds some perf overhead to plugins initialization.

## Expected Behavior

During workspace creation, before generating the preset, install known preset dependencies that are installed using `ensurePackage`, so the `ensurePackage` doesn't perform an individual installation and perf is improved.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/workspace/src/generators/new/generate-preset.ts
+++ b/packages/workspace/src/generators/new/generate-preset.ts
@@ -4,7 +4,11 @@ import {
   Tree,
 } from '@nrwl/devkit';
 import { Preset } from '../utils/presets';
-import { nxVersion } from '../../utils/versions';
+import {
+  angularCliVersion,
+  nxVersion,
+  typescriptVersion,
+} from '../../utils/versions';
 import { getNpmPackageVersion } from '../utils/get-npm-package-version';
 import { NormalizedSchema } from './new';
 import { join } from 'path';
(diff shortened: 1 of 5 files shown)


Coly010 (approved):

packages/workspace/src/generators/new/generate-preset.ts
```diff
@@ -93,20 +97,29 @@ function getPresetDependencies(preset: string, version?: string) {
...
+      };
 
     case Preset.NextJs:
       return { dependencies: { '@nrwl/next': nxVersion }, dev: {} };
 
     case Preset.ReactMonorepo:
-      return { dependencies: {}, dev: { '@nrwl/react': nxVersion } };
-
     case Preset.ReactStandalone:
       return { dependencies: {}, dev: { '@nrwl/react': nxVersion } };
```
FrozenPandaz, line 124: Doesn't this need typescript?

Is it because we have a `js` option?

1. I'm not sure we have the ability for them to pass the `--js` option
2. If there is the ability.. we could maybe check for the option here?

FrozenPandaz (commented):
Does this completely remove the installs during preset portion of `create-nx-workspace --preset react-standalone`?

Could you also post the timing difference here so we can see the difference it makes?

packages/workspace/src/generators/new/generate-preset.ts
leosvelperez: Yeah, it's because of the `js` option.

I'm hesitant of doing so, I think it's very fragile to assume:

- Whether the option is passed
- The default value in the preset generator if not passed

Any of those things are controlled somewhere else and if they change, it's hard to know that the logic here would need to be changed. If we still decide to do such assumptions, we could do the same and assume whether `@nrwl/jest`, `nrwl/cypress`, `@nrwl/js` need to be installed.

AgentEnder (approved):

PR #15461, opened 2023-03-06
feat(repo): use caching for pnpm workspaces in e2e tests

meeroslav:
Use workspace caching for pnpm as well.

This will significantly cut down on CI times.

We had it switched off, since there were some issues in the past for pnpm but that has been resolved a long time ago.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/utils/create-project-utils.ts
+++ b/e2e/utils/create-project-utils.ts
@@ -39,7 +39,7 @@ export function newProject({
   packageManager = getSelectedPackageManager(),
 } = {}): string {
   try {
-    const useBackupProject = packageManager !== 'pnpm';
+    const useBackupProject = /*packageManager !== 'pnpm'*/ true;
     const projScope = useBackupProject ? 'proj' : name;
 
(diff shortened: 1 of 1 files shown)


e2e/utils/create-project-utils.ts
```diff
@@ -39,7 +39,7 @@ export function newProject({
   packageManager = getSelectedPackageManager(),
 } = {}): string {
   try {
-    const useBackupProject = packageManager !== 'pnpm';
+    const useBackupProject = /*packageManager !== 'pnpm'*/ true;
```
barbados-clemens, line 42: should just get rid of the flag in this case.

barbados-clemens (changes requested):

e2e/utils/create-project-utils.ts
meeroslav: True, I left it for testing. Let me remove it quickly

barbados-clemens (approved):

PR #15479, opened 2023-03-06
feat(core): move string utils to devkit

Coly010:
## Current Behavior

There are string utils being used from `@nrwl/workspace`. 

## Expected Behavior

Move the string utils to devkit
Deprecate the existing utils in favour of the devkit package to allow plugin authors time to adjust their code.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/angular-extensions/src/misc.test.ts
+++ b/e2e/angular-extensions/src/misc.test.ts
@@ -7,7 +7,7 @@ import {
   uniq,
   updateFile,
 } from '@nrwl/e2e/utils';
-import { classify } from '@nrwl/workspace/src/utils/strings';
+import { classify } from '@nrwl/devkit/src/utils/string-utils';
 
 describe('Move Angular Project', () => {
   let proj: string;
--- a/e2e/next/src/next.test.ts
+++ b/e2e/next/src/next.test.ts
@@ -1,4 +1,5 @@
 import { detectPackageManager, joinPathFragments } from '@nrwl/devkit';
+import { capitalize } from '@nrwl/devkit/src/utils/string-utils';
 import {
   checkFilesExist,
   cleanupProject,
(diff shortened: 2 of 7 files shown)


packages/js/src/utils/string-utils.ts
```diff
@@ -0,0 +1,153 @@
+/**
```
meeroslav, line 1: Why is this not moved over to `devkit`?

Doesn't feel like it belongs to `js` package

meeroslav (changes requested):
This feature doesn't seem to belong to `js` only. Perhaps move it to devkit?

packages/js/src/utils/string-utils.ts
Coly010: It can be discussed. I was under the impression that Devkit should be for plugin-specific code, whereas JS is a better place for JS-specific logic

FrozenPandaz: When you move this to `devkit` make sure to consider there is a util there called `names(name: string): Record<string, string>` we don't want to export 2 ways of creating names.

mandarini (approved):

jaysoo (approved):

leosvelperez (approved):

xiongemi (approved):

packages/js/src/utils/string-utils.ts
meeroslav: I think it should be combined with https://github.com/nrwl/nx/blob/master/packages/devkit/src/utils/names.ts.

For example, we could use those string functions to replace the arbitrary functions `toClassName`, `toPropertyName`, `toConstantName etc.

meeroslav (approved):
Looks good to me

FrozenPandaz (approved):

PR #15482, opened 2023-03-06
chore(misc): update lerna to 6.5.1

fahslaj:
Updates lerna to 6.5.1 so that the nx repo can make use of the idempotency on lerna publish.

### DIFF
--- a/scripts/nx-release.ts
+++ b/scripts/nx-release.ts
@@ -1,14 +1,14 @@
 #!/usr/bin/env node
-import * as yargs from 'yargs';
 import { execSync } from 'child_process';
 import { existsSync, readFileSync, writeFileSync } from 'fs';
-import { URL } from 'url';
 import { join } from 'path';
+import { URL } from 'url';
+import * as yargs from 'yargs';
 
 import { parse } from 'semver';
 
-import * as version from '@lerna/version/index';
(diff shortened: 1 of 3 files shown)


fahslaj:
@FrozenPandaz This Lerna version bump includes the idempotency update for `lerna publish`.

AgentEnder (approved):

scripts/nx-release.ts
```diff
@@ -1,14 +1,14 @@
...
 import { existsSync, readFileSync, writeFileSync } from 'fs';
-import { URL } from 'url';
 import { join } from 'path';
+import { URL } from 'url';
+import * as yargs from 'yargs';
 
 import { parse } from 'semver';
 
-import * as version from '@lerna/version/index';
-import * as publish from '@lerna/publish/index';
```
FrozenPandaz, line 11: Any idea why this had to change?

FrozenPandaz (approved):

scripts/nx-release.ts
fahslaj: Yes, the commands are no longer exported from separate packages and are instead exported by `lerna`. We don't really encourage direct import of the functions like this anyway so the typescript types aren't exported either.

PR #15484, opened 2023-03-06
feat(node): add nestjs support to node app plugin

ndcunningham:
Esbuild's compatibility is not as smooth as other frameworks we have currently supported (express, fastify, koa).
So, currently, this PR is compatible with `webpack`.

A follow-up PR will be to support `esbuild`.

### DIFF
--- a/docs/generated/packages/node/generators/application.json
+++ b/docs/generated/packages/node/generators/application.json
@@ -91,7 +91,7 @@
       "framework": {
         "description": "Generate the node application using a framework",
         "type": "string",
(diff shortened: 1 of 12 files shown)


nartc (approved):

jaysoo:
Approved, but can you also add an e2e test to `node-server.test.ts`? The other three frameworks are covered there so let's add a forth one.

jaysoo (approved):

AgentEnder (approved):

PR #15493, opened 2023-03-07
feat(angular): move replaceAppNameWithPath to Angular

Coly010:
## Current Behavior

replaceAppNameWithPath is used in multiple plugins via a deep import to @nrwl/workspace

## Expected Behavior

Move the function to Angular
Deprecate the existing fuction in @nrwl/workspace to allow plugin authors time to migrate.

## Related Issue(s)

Fixes #


leosvelperez (commented):
I have my doubts about making this function part of the DevKit public API. It's probably not something generic enough to make it available for everyone. The logic is Angular-specific and it's only used on the Angular plugin. I would inline it in the Angular plugin utils.

packages/angular/src/generators/application/lib/update-config-files.ts
```diff
@@ -8,11 +8,12 @@ import {
   removeProjectConfiguration,
   updateJson,
 } from '@nrwl/devkit';
-import { replaceAppNameWithPath } from '@nrwl/workspace/src/utils/cli-config-utils';
+
```
leosvelperez, line 11: Remove blank line

packages/angular/src/generators/application/angular-v14/lib/update-config-files.ts
```diff
@@ -8,14 +8,15 @@ import {
   removeProjectConfiguration,
   updateJson,
 } from '@nrwl/devkit';
-import { replaceAppNameWithPath } from '@nrwl/workspace/src/utils/cli-config-utils';
+
```
leosvelperez, line 11: Let's remove this blank line, it's not a needed change.

packages/angular/src/utils/cli-config-utils.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { replaceAppNameWithPath } from './cli-config-utils';
+
+describe('replaceAppNameWithPath', () => {
+  describe('when node is `application`', () => {
+    describe('and appName is `app`', () => {
```
leosvelperez, line 5: Too many `describe`s, we can simplify this.

packages/angular/src/utils/cli-config-utils.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+        const node = 'application';
+        const appName = 'app';
+        const root = 'apps/app';
+        expect(replaceAppNameWithPath(node, appName, root)).toEqual(node);
+      });
+    });
+  });
+
+  describe('when node is `library`', () => {
+    describe('and appName is `lib`', () => {
```
leosvelperez, line 16: Same as previous.

packages/angular/src/utils/cli-config-utils.ts
```diff
@@ -0,0 +1,50 @@
...
+      return node;
+    }
+  } else if (Array.isArray(node)) {
+    return node.map((j) => replaceAppNameWithPath(j, appName, root));
+  } else if (typeof node === 'object' && node) {
+    const forbiddenPropertyList: string[] = [
+      'prefix',
+      'builder',
+      'executor',
+      'browserTarget',
```
leosvelperez, line 30: I know this is a copy and paste from existing functionality and not an issue introduced by this PR, but we also have a `buildTarget` property in some of our builders that is similar to `browserTarget`. We should probably include it in this list as well.

leosvelperez (changes requested):

mandarini (approved):

leosvelperez (approved):

packages/workspace/src/utils/cli-config-utils.ts
```diff
@@ -34,6 +34,10 @@ export function serializeTarget({ project, target, config }) {
   return [project, target, config].filter((part) => !!part).join(':');
 }
 
+/**
+ *
+ * @deprecated This function will be removed from `@nrwl/workspace` in version 17. Prefer importing from `@nrwl/devkit`
+ */
 export function replaceAppNameWithPath(
```
FrozenPandaz, line 41: I don't think anyone is using this:

https://grep.app/search?q=replaceAppNameWithPath&case=true&filter[lang][0]=TypeScript

Let's remove it.

FrozenPandaz (changes requested):

AgentEnder (approved):

FrozenPandaz (approved):

PR #15523, opened 2023-03-08
fix(core): issue with registering file dependencies

forivall:
## Current Behavior
Files wont have all the dependencies in the project graph

## Expected Behavior
Files are once again reported with the correct dependencies in the project graph

## Related Issue(s)
When i encountered the issue was having inconsistent (race-condition) behaviour based on the execution of the dependency analysis workers, so I didnt want to create a repro right out of the gate, so no issue, unfortunately. Now that I know the root cause, i can create a reproduction if desired, but I think the unit test is sufficient. This may fix some linting bugs.

Fixes #

### DIFF
--- a/packages/nx/src/project-graph/project-graph-builder.spec.ts
+++ b/packages/nx/src/project-graph/project-graph-builder.spec.ts
@@ -148,6 +148,43 @@ describe('ProjectGraphBuilder', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


forivall:
Added a unit test which fails without the fix, and is fixed by the fix.

FrozenPandaz (approved):
Thank you for your contribution! LGTM! :tada:

PR #15544, opened 2023-03-08
chore(core): small fixes around e2e logging

mandarini:
* Added some optional chaining to some places around the logs, because it would throw a wrong error.
* Added some more `isVerbose` to `stdio`

### DIFF
--- a/e2e/utils/command-utils.ts
+++ b/e2e/utils/command-utils.ts
@@ -72,16 +72,16 @@ export function runCommand(
   try {
     const r = execSync(command, {
       cwd: tmpProjPath(),
-      stdio: ['pipe', 'pipe', 'pipe'],
+      stdio: isVerbose() ? 'inherit' : 'pipe',
       env: {
         ...getStrippedEnvironmentVariables(),
         ...childProcessOptions?.env,
         FORCE_COLOR: 'false',
-      },
+      } as any,
       encoding: 'utf-8',
       ...childProcessOptions,
-    }).toString();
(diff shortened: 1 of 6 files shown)


e2e/utils/command-utils.ts
```diff
@@ -72,19 +72,19 @@ export function runCommand(
...
       ...childProcessOptions,
-    }).toString();
-    if (process.env.NX_VERBOSE_LOGGING) {
-      console.log(r);
+    });
+    if (isVerbose()) {
+      console.log(r?.toString());
     }
-    return r;
+    return r?.toString();
```
meeroslav, line 87: This is already string thanks  to encoding so no need to call `toString()`

meeroslav (approved):
Left some minor improvement points, but otherwise looks good

e2e/utils/command-utils.ts
```diff
@@ -72,19 +72,19 @@ export function runCommand(
...
+      stdio: isVerbose() ? 'inherit' : ['pipe', 'pipe', 'pipe'],
       env: {
         ...getStrippedEnvironmentVariables(),
         ...childProcessOptions?.env,
         FORCE_COLOR: 'false',
-      },
+      } as any,
       encoding: 'utf-8',
       ...childProcessOptions,
-    }).toString();
```
mandarini, line 83: @meeroslav it was actually `toString()` before, I just moved it. I'll put it back. But if `encoding` makes sure what's returned it's already a string, maybe we should change this throughout the utils.

leosvelperez (approved):

PR #15548, opened 2023-03-08
fix(core): refactor the logging logic in e2e tests

mandarini:
* change `['pipe', 'pipe', 'pipe']` to `pipe`
* enhance the way `isVerbose()` is used
* change look of output if `isVerbose()`
* make sure error is always caught and logged
* remove unnecessary `.toString()` since we're using `encoding` to all `execSync` commands
* add optional chaining where needed

## Comments

Output, if error, is always displayed. 

The thing is, however, that in the test summary, if there was an error, still the stack trace of `error on execSync` is displayed. We can take a look at whether this is configurable or not at a later PR.

### DIFF
--- a/e2e/utils/command-utils.ts
+++ b/e2e/utils/command-utils.ts
@@ -1,4 +1,4 @@
-import { PackageManager } from '@nrwl/devkit';
+import { output, PackageManager } from '@nrwl/devkit';
 import { packageInstall, tmpProjPath } from './create-project-utils';
 import {
   detectPackageManager,
@@ -70,9 +70,9 @@ export function runCommand(
 ): string {
   const { failOnError, ...childProcessOptions } = options ?? {};
   try {
-    const r = execSync(command, {
+    const r = execSync(`${command}${isVerbose() ? ' --verbose' : ''}`, {
       cwd: tmpProjPath(),
-      stdio: ['pipe', 'pipe', 'pipe'],
+      stdio: 'pipe',
       env: {
         ...getStrippedEnvironmentVariables(),
         ...childProcessOptions?.env,
@@ -80,16 +80,23 @@ export function runCommand(
       },
       encoding: 'utf-8',
       ...childProcessOptions,
-    }).toString();
-    if (process.env.NX_VERBOSE_LOGGING) {
-      console.log(r);
+    });
+
+    if (isVerbose()) {
+      output.log({
+        title: `Command: ${command}`,
+        bodyLines: [r as string],
+        color: 'green',
+      });
     }
-    return r;
+
+    return r as string;
   } catch (e) {
     // this is intentional
     // npm ls fails if package is not found
+    logError(`Original command: ${command}`, `${e.stdout}\n\n${e.stderr}`);
     if (!failOnError && (e.stdout || e.stderr)) {
-      return e.stdout?.toString() + e.stderr?.toString();
+      return e.stdout + e.stderr;
     }
     throw e;
   }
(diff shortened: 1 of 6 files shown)


leosvelperez (approved):

Coly010 (approved):

e2e/utils/command-utils.ts
```diff
@@ -70,26 +71,33 @@ export function runCommand(
 ): string {
   const { failOnError, ...childProcessOptions } = options ?? {};
   try {
-    const r = execSync(command, {
+    const r = execSync(`${command}${isVerbose() ? ' --verbose' : ''}`, {
```
meeroslav, line 74: We need to be careful with this since it changes the output from the command.

e2e/utils/command-utils.ts
```diff
@@ -70,26 +71,33 @@ export function runCommand(
...
       cwd: tmpProjPath(),
-      stdio: ['pipe', 'pipe', 'pipe'],
+      stdio: 'pipe',
       env: {
         ...getStrippedEnvironmentVariables(),
         ...childProcessOptions?.env,
         FORCE_COLOR: 'false',
-      },
+      } as any,
```
meeroslav, line 81: Do we need type here? It should be automatically inferred based on the `encoding`.

If yes, please use `NodeJS.ProcessEnv` instead `any

e2e/utils/command-utils.ts
```diff
@@ -189,7 +197,7 @@ export function runCommandAsync(
           CI: 'true',
           ...(opts.env || getStrippedEnvironmentVariables()),
           FORCE_COLOR: 'false',
-        },
+        } as any,
```
meeroslav, line 200: Same as above

e2e/utils/command-utils.ts
```diff
@@ -218,14 +226,14 @@ export function runCommandUntil(
       CI: 'true',
       ...getStrippedEnvironmentVariables(),
       FORCE_COLOR: 'false',
-    },
+    } as any,
```
meeroslav, line 229: Same as above

e2e/utils/command-utils.ts
```diff
@@ -218,14 +226,14 @@ export function runCommandUntil(
...
   });
   return new Promise((res, rej) => {
     let output = '';
     let complete = false;
 
     function checkCriteria(c) {
-      output += c.toString();
+      output += c?.toString();
```
meeroslav, line 236: No need for this. The `checkCriteria` is called on `data`, so there will always be an input.

e2e/utils/command-utils.ts
```diff
@@ -272,25 +280,35 @@ export function runNgAdd(
...
-    })
-      .toString()
-      .replace(
-        /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g,
-        ''
-      );
+    const result = execSync(
+      pmc.run(
+        `ng g ${packageName}:ng-add`,
+        `${command}${isVerbose() ? ' --verbose' : ''}` ?? ''
```
meeroslav, line 286: Same as above, we need to make sure to run all the tests to ensure this doesn't break anything.

meeroslav: Also can, we move this interpolation before the `execSync` call so that it's easier to read?

e2e/utils/create-project-utils.ts
```diff
@@ -169,13 +170,27 @@ export function runCreateWorkspace(
...
+
+    if (isVerbose()) {
+      output.log({
+        title: `Command: ${command}`,
+        bodyLines: [create as string],
+        color: 'green',
+      });
+    }
+
+    return create ?? '';
```
meeroslav, line 189: The `create` will always be string thanks to `utf-8` encoding so this is not needed. We can just `return create`

e2e/utils/create-project-utils.ts
```diff
@@ -212,13 +227,27 @@ export function runCreatePlugin(
...
```
meeroslav, line 246: Same comment as above

e2e/utils/create-project-utils.ts
```diff
@@ -233,16 +262,37 @@ export function packageInstall(
...
+      output.log({
+        title: `Command: ${command}`,
+        bodyLines: [install as string],
+        color: 'green',
+      });
     }
-  );
-  return install ? install.toString() : '';
+
+    return install ?? '';
```
meeroslav, line 291: Same as above

e2e/utils/log-utils.ts
```diff
@@ -40,7 +40,7 @@ export function logSuccess(title: string, body?: string) {
  * @returns
  */
 export function stripConsoleColors(log: string): string {
-  return log.replace(
+  return log?.replace(
```
meeroslav, line 43: log is always string. Calling `replace` on an empty string does nothing.

We don't need this change

meeroslav (changes requested):
I have some minor changes.

e2e/utils/command-utils.ts
mandarini: Yes, but I think it's fair to change the output. We do say "if you want to get a full log run with verbose" and I think that it makes sense to add the `--verbose` to these commands directly.

e2e/utils/command-utils.ts
mandarini: It was throwing an error, which is not any more 🤔

e2e/utils/command-utils.ts
```diff
@@ -272,25 +280,33 @@ export function runNgAdd(
...
       env: { ...(opts.env || getStrippedEnvironmentVariables()) },
       encoding: 'utf-8',
-    })
-      .toString()
-      .replace(
-        /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g,
-        ''
-      );
+    })?.replace(
+      /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g,
```
barbados-clemens, line 293: we can use the stripConsoleColors here to keep it a a bit dry.

barbados-clemens (commented):
Thanks for this!
just a place to reuse the stripConsoleColor is the only thing I see in the PR.

e2e/utils/command-utils.ts
mandarini: Thanks!! :D Done! Can you check again please?

barbados-clemens (approved):

meeroslav (approved):

AgentEnder (approved):

PR #15560, opened 2023-03-09
cleanup(angular): fix unit tests setup so peer deps are not installed with ensurePackage

leosvelperez:
## Current Behavior

When running the Angular plugin unit tests, several package installations occurs. This happens because the `init` generator (invoked from app, lib, host, remote generators) install the plugin peer deps if they are not found in the `package.json` that's on the `Tree`. This also causes a degradation in the performance of the tests.

## Expected Behavior

When running the Angular plugin unit tests, no package installations should occur. The `Tree` used in the tests should contain the required peer deps in the `package.json`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/migrations/update-15-2-0/update-typescript-target.spec.ts
+++ b/packages/angular/src/migrations/update-15-2-0/update-typescript-target.spec.ts
@@ -1,3 +1,4 @@
+import { UnitTestRunner } from '@nrwl/angular/src/utils/test-runners';
 import {
   readJson,
   readProjectConfiguration,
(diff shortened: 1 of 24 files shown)


packages/angular/src/migrations/update-15-2-0/update-typescript-target.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { UnitTestRunner } from '@nrwl/angular/src/utils/test-runners';
```
Coly010, line 1: Should this import be relative?

Coly010 (approved):

packages/angular/src/migrations/update-15-2-0/update-typescript-target.spec.ts
leosvelperez: Absolutely. It was already there so I didn't notice (the change was from VSCode organizing imports feature). Fixed now.

PR #15569, opened 2023-03-09
feat(nextjs): add --appDir support to next.js app generator

jaysoo:
This PR enables `app` directory for Next.js, which is currently experimental.

## TODO
- [ ] Fix issue where exported `metadata` in RSC is being erroneously marked as a client component (even though `use client` is not present.
- [ ] Add e2e test to cover `app` directory.


packages/next/src/generators/application/lib/create-application-files.ts
```diff
@@ -28,11 +28,27 @@ export function createApplicationFiles(host: Tree, options: NormalizedSchema) {
 
   generateFiles(
     host,
-    join(__dirname, '../files'),
+    join(__dirname, '../files/common'),
     options.appProjectRoot,
     templateVariables
   );
 
+  if (options.appDir) {
```
ndcunningham, line 36: You should also update `next.config.js` to include `experimental: { appDir: true }`

PR #15541, opened 2023-03-08
feat(repo): add new flag for verbose e2e logging

meeroslav:
Existing NX_VERBOSE_LOGGING is too broad for E2E tests and creates a lot of noise in the output.
The new NX_E2E_VERBOSE_LOGGING reports just relevant logs.

## Related Issue(s)

Fixes #

PR #15587, opened 2023-03-10
fix(angular): fix e2e tests to support pnpm v7

meeroslav:
## Related Issue(s)

Cherry picks https://github.com/nrwl/nx/pull/15695, which should be merged in first


e2e/utils/command-utils.ts
```diff
@@ -70,7 +70,7 @@ export function runCommand(
 ): string {
   const { failOnError, ...childProcessOptions } = options ?? {};
   try {
-    const r = execSync(`${command}${isVerbose() ? ' --verbose' : ''}`, {
+    const r = execSync(command, {
```
meeroslav, line 73: This is fixed in https://github.com/nrwl/nx/pull/15541

e2e/utils/command-utils.ts
```diff
@@ -279,10 +279,7 @@ export function runNgAdd(
   try {
     const pmc = getPackageManagerCommand();
     packageInstall(packageName, undefined, version);
-    const fullCommand = pmc.run(
-      `ng g ${packageName}:ng-add`,
-      `${command}${isVerbose() ? ' --verbose' : ''}` ?? ''
-    );
+    const fullCommand = pmc.run(`ng g ${packageName}:ng-add`, command ?? '');
```
meeroslav, line 282: This is fixed in https://github.com/nrwl/nx/pull/15541

e2e/utils/command-utils.ts
```diff
@@ -369,9 +366,7 @@ export function runLernaCLI(
 ): string {
   try {
     const pm = getPackageManagerCommand();
-    const fullCommand = `${pm.runLerna} ${command}${
-      isVerbose() ? ' --verbose' : ''
-    }`;
+    const fullCommand = `${pm.runLerna} ${command}`;
```
meeroslav, line 369: This is fixed in https://github.com/nrwl/nx/pull/15541

e2e/angular-core/src/ng-add.test.ts
```diff
@@ -72,7 +72,7 @@ describe('convert Angular CLI workspace to an Nx workspace', () => {
   });
 
   afterEach(() => {
-    cleanupProject();
+    cleanupProject({ skipReset: true });
```
meeroslav, line 75: These tests do not start with standard create-nx-workspace so running `nx reset` fails after the test.

packages/angular/src/generators/host/files/src/main.server.ts__tmpl__
```diff
@@ -39,7 +39,7 @@ export function app(): express.Express {
   );
 
   // All regular routes use the Universal engine
-  server.get('*', (req, res) => {
+  server.get('*', (req: any, res: any) => {
```
meeroslav, line 42: Build fails due to Typescript errors

packages/angular/src/generators/host/lib/add-ssr.ts
```diff
@@ -65,7 +67,10 @@ export async function addSsr(tree: Tree, options: Schema, appName: string) {
       express: expressVersion,
       '@module-federation/node': moduleFederationNodeVersion,
     },
-    {}
+    {
+      '@types/cors': typesCorsVersion,
```
meeroslav, line 71: Build fails due to unknown types

package.json
```diff
@@ -46,7 +46,7 @@
     "@babel/preset-typescript": "^7.15.0",
     "@cypress/react": "^6.0.0",
     "@floating-ui/react": "0.19.2",
-    "@monodon/rust": "1.0.0",
+    "@monodon/rust": "1.1.1",
```
meeroslav, line 49: This is cherry picked from https://github.com/nrwl/nx/pull/15695

yarn.lock
```diff
@@ -4281,10 +4281,10 @@
     npmlog "^6.0.2"
     write-file-atomic "^4.0.1"
 
-"@ltd/j-toml@1.24.0":
-  version "1.24.0"
-  resolved "https://registry.yarnpkg.com/@ltd/j-toml/-/j-toml-1.24.0.tgz#4c993610586964fb21fd38a84bb41913f6e4f5ff"
-  integrity sha512-XWEvSNLJ2YeCvlNDd/DQfTMldJqBZWX8+RKgPMY2i3MShqsB9XvtY2UfHqcuPb+eRrNO1WR7w4xlNfzCzop+Ig==
+"@ltd/j-toml@1.38.0":
```
meeroslav, line 4284: This is cherry picked from https://github.com/nrwl/nx/pull/15695

mandarini (approved):

e2e/utils/create-project-utils.ts
```diff
@@ -81,7 +81,7 @@ export function newProject({
       packageInstall(packages.join(` `), projScope);
 
       // stop the daemon
-      execSync('nx reset', {
+      execSync('npx nx reset', {
```
barbados-clemens, line 84: should'd we use the `packageManaer` instead?

meeroslav: We could but it's not necessary in this case. This should happen only once per test project run and ensures daemon is stopped before we copy workspace to another folder.

Coly010 (approved):

packages/angular/src/generators/host/files/src/main.server.ts__tmpl__
leosvelperez: These TS type errors are very interesting. They only arise with pnpm v7. I tried with pnpm 6 and it's all good. Also, the IDE doesn't report such an issue, even when the packages are installed with pnpm v7.

That said, by adding `@types/express` the types are inferred correctly and therefore, we don't need to add `any`.

e2e/angular-core/src/ng-add.test.ts
leosvelperez: They don't start with `create-nx-workspace` but still end up being Nx workspaces at the end of the tests. Each test converts the initial Angular CLI workspace into an Nx workspace. Surely that starts the daemon and needs to be reset, isn't it?

I can see a couple of test cases don't run any command, so maybe those don't have the daemon running, but the rest run commands. We might need to inline the `cleanupProject` calls to the end of each test case and pass the `skipReset: true` to those than don't run commands.

leosvelperez (changes requested):
Left a couple of comments in some of the comments you initially added.

packages/angular/src/generators/host/files/src/main.server.ts__tmpl__
meeroslav: You are correct, adding types solved it

e2e/angular-core/src/ng-add.test.ts
meeroslav: Actually, by ensuring proper cleanup the `nx reset` issue is now gone so I reverted in back to clean `cleanupProject`

leosvelperez (approved):

PR #15608, opened 2023-03-10
fix(js): Always create tmp tsconfig

### LINKED ISSUE #15043: Nx build: Derived tsconfig is not working when project has 0 dependencies.
### Current Behavior

When building a lib with 0 dependencies, which uses a derived tsconfig.base the nx build commands gives an error:
`Cannot read properties of undefined (reading 'paths')`.

This is because the derived tsconfig is not taken into account when building a lib that doesn't have any dependencies.

### Expected Behavior

When running nx build on a project without any dependencies I would expect that the `extends` property is still taken into account.

Also Nx build should succeed without any error in such a scenario.

### GitHub Repo

https://github.com/janwiebedroid/nx-tsconfig-bug/tree/main/nx-ts-config-bugs

### Steps to Reproduce

1. Ensure that there are no compiler options in packages/testlib1/tsconfig.lib.json
2. Run `nx build testlib1`
3. Error occurs

When running nx build on a project with dependencies it works:
1. Ensure that there are no compiler options in packages/testlib-with-deps/tsconfig.lib.json
2. Run `nx build testlib-with-deps`
3. No error occursn

### Nx Report

```shell

   Node : 16.17.0
   OS   : win32 x64
   npm  : 8.15.0

   nx                      : 15.7.1
   @nrwl/linter            : 15.7.1
   @nrwl/workspace         : 15.7.1
   @nrwl/cli               : 15.7.1
   @nrwl/devkit            : 15.7.1
   @nrwl/eslint-plugin-nx  : 15.7.1
   @nrwl/js                : 15.7.1
   @nrwl/tao               : 15.7.1
   typescript              : 4.8.4
```

### Failure Logs

_No response_

### Additional Information

The issue seems to be in [check-dependencies.ts](https://github.com/nrwl/nx/blob/master/packages/js/src/utils/check-dependencies.ts).

If there are 0 dependencies a temp tsconfig is never created.

janwiebedroid:
CheckDependencies didnt generate tmpTsConfig when no project dependencies were found even though a base tsconfig might still exist

Closes #15043

## Current Behavior
When building a js lib with 0 dependencies, which uses a derived tsconfig.base the nx build commands gives an error:
Cannot read properties of undefined (reading 'paths').

## Expected Behavior
No error occurs when building a js lib with 0 dependencies

## Related Issue(s)
Fixes #15043

### DIFF
--- a/packages/js/src/utils/check-dependencies.ts
+++ b/packages/js/src/utils/check-dependencies.ts
@@ -22,23 +22,13 @@ export function checkDependencies(
     context.configurationName
   );
   const projectRoot = target.data.root;
-
-  if (dependencies.length > 0) {
-    return {
-      tmpTsConfig: createTmpTsConfig(
-        tsConfigPath,
-        context.root,
-        projectRoot,
-        dependencies
-      ),
-      projectRoot,
-      target,
-      dependencies,
-    };
-  }
-
   return {
-    tmpTsConfig: null,
+    tmpTsConfig: createTmpTsConfig(
+      tsConfigPath,
+      context.root,
+      projectRoot,
+      dependencies
+    ),
     projectRoot,
     target,
     dependencies,


janwiebedroid:
Locally the e2e tests are succeeding. Could someone help me figuring this one out? Maybe the build just needs a restart?

jaysoo:
Thanks for the PR! We actually don't need to create the temporary tsconfig file. This has been patched here for Nx 16 https://github.com/nrwl/nx/commit/afb37c8691159081ff919a9d593dd11848859847#diff-7977b9747825cac3082c1aaa4912365558ea4c0fa09d78749eeef4221144dfcf

PR #15627, opened 2023-03-13
feat(angular): remove protractor

Coly010:
## Current Behavior

We currently still offer ways to use protractor

## Expected Behavior

We should not offer methods to use protractor unless user has migrated a workspace with an existing setup

## Related Issue(s)

Fixes #


packages/angular/src/utils/versions.ts
```diff
@@ -24,7 +24,6 @@ export const autoprefixerVersion = '^10.4.0';
 export const tsNodeVersion = '10.9.1';
 
 export const jestPresetAngularVersion = '~13.0.0';
-export const protractorVersion = '~7.0.0';
```
leosvelperez, line 27: Please check other package versions here that were only needed for Protractor and remove them as well.

packages/angular/src/utils/backward-compatible-versions.ts
```diff
@@ -33,7 +33,6 @@ export const backwardCompatibleVersions: Record<
     autoprefixerVersion: '^10.4.0',
     tsNodeVersion: '10.9.1',
     jestPresetAngularVersion: '~12.2.3',
-    protractorVersion: '~7.0.0',
```
leosvelperez, line 36: Please check other package versions here that were only needed for Protractor and remove them as well.

leosvelperez (changes requested):

bcabanes (approved):

leosvelperez (approved):

FrozenPandaz (approved):

PR #15630, opened 2023-03-13
feat(angular): add directive generator

Coly010:
## Current Behavior

We rely on the @schematics/angular package to be able to generate directives

## Expected Behavior

We should not have a reliance on @schematics/angular and be able to control better the generation of directives

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/directive/directive.spec.ts
+++ b/packages/angular/src/generators/directive/directive.spec.ts
@@ -1,0 +1,163 @@
+import { addProjectConfiguration, Tree } from '@nrwl/devkit';
+import { createTreeWithEmptyWorkspace } from '@nrwl/devkit/testing';
+import { directiveGenerator } from './directive';
+import type { Schema } from './schema';
+
+describe('directive generator', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+
+    addProjectConfiguration(tree, 'test', {
+      root: 'test',
+      sourceRoot: 'test/src',
+    });
+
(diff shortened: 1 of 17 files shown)


bcabanes (approved):

packages/angular/src/generators/directive/directive.ts
```diff
@@ -0,0 +1,160 @@
...
+  getProjects,
+  joinPathFragments,
+  names,
+  readNxJson,
+  readProjectConfiguration,
+} from '@nrwl/devkit';
+import type { Schema } from './schema';
+import { checkPathUnderProjectRoot } from '../utils/path';
+import { dirname } from 'path';
+import { insertNgModuleProperty } from '@nrwl/angular/src/generators/utils';
```
leosvelperez, line 14: Make the import path relative.

packages/angular/src/generators/directive/directive.spec.ts
```diff
@@ -0,0 +1,163 @@
+import { addProjectConfiguration, Tree } from '@nrwl/devkit';
+import { createTreeWithEmptyWorkspace } from '@nrwl/devkit/testing-pre16';
```
leosvelperez, line 2: Is there a reason to use the `pre16` one?
```suggestion
import { createTreeWithEmptyWorkspace } from '@nrwl/devkit/testing';
```

leosvelperez (changes requested):

AgentEnder (approved):
e2e/workspace-create looks fine

leosvelperez (approved):

PR #15659, opened 2023-03-14
feat(angular): add pipe generator

Coly010:
## Current Behavior

We currently rely on @schematics/angular to generate pipes

## Expected Behavior

We should not rely on @schematics/angular to generate pipe

## Related Issue(s)

E2Es will be fixed when https://github.com/nrwl/nx/pull/15630 merges

### DIFF
--- a/packages/angular/src/generators/pipe/schema.json
+++ b/packages/angular/src/generators/pipe/schema.json
@@ -1,0 +1,73 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "SchematicsAngularPipe",
+  "title": "Angular Pipe Options Schema",
+  "type": "object",
+  "cli": "nx",
+  "additionalProperties": false,
+  "description": "Creates a new, generic pipe definition in the given project.",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "The name of the pipe.",
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      },
+      "x-prompt": "What name would you like to use for the pipe?"
+    },
+    "path": {
+      "type": "string",
+      "format": "path",
+      "$default": {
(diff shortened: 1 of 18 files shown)


bcabanes (approved):

packages/angular/src/generators/directive/directive.ts
```diff
@@ -10,10 +10,7 @@ import {
 } from '@nrwl/devkit';
 import type { Schema } from './schema';
 import { checkPathUnderProjectRoot } from '../utils/path';
-import { dirname } from 'path';
-import { insertNgModuleProperty } from '../utils';
-import { insertImport } from '@nrwl/js';
-import { ensureTypescript } from '@nrwl/js/src/utils/typescript/ensure-typescript';
+import { addToNgModule, findModule } from '@nrwl/angular/src/generators/utils';
```
leosvelperez, line 13: Change the import path to a relative path.

packages/angular/src/generators/pipe/schema.json
```diff
@@ -0,0 +1,70 @@
...
+        "$source": "workingDirectory"
+      },
+      "description": "The path at which to create the pipe, relative to the workspace root.",
+      "visible": false
+    },
+    "project": {
+      "type": "string",
+      "description": "The name of the project.",
+      "x-dropdown": "projects"
+    },
```
leosvelperez, line 32: Let's add the default source here so the default project is auto-populated:
```suggestion
    "project": {
      "type": "string",
      "description": "The name of the project.",
      "$default": {
        "$source": "projectName",
        "index": 0
      },
      "x-dropdown": "projects"
    },
```

leosvelperez (changes requested):

leosvelperez (approved):

PR #15681, opened 2023-03-14
fix(node): fix standalone linter setup

meeroslav:
The current implementation of `node-server` preset (node standalone) will produce `.eslintrc.json` that is broken.

This PR fixes that.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/linter/src/linter.test.ts
+++ b/e2e/linter/src/linter.test.ts
@@ -473,7 +473,7 @@ describe('Linter', () => {
         'plugin:@nrwl/nx/javascript',
       ]);
 
-      runCLI(`generate @nrwl/react:lib ${mylib} --unitTestRunner=jest`);
+      runCLI(`generate @nrwl/workspace:lib ${mylib} --unitTestRunner=jest`);
       // should add new tslint
       expect(() => checkFilesExist(`.eslintrc.base.json`)).not.toThrow();
       const appEslint = readJson(`.eslintrc.json`);
@@ -494,10 +494,7 @@ describe('Linter', () => {
         'plugin:cypress/recommended',
         '../.eslintrc.base.json',
       ]);
-      expect(libEslint.extends).toEqual([
-        'plugin:@nrwl/nx/react',
-        '../../.eslintrc.base.json',
-      ]);
+      expect(libEslint.extends).toEqual(['../../.eslintrc.base.json']);
       // should have no plugin extends
       expect(appEslint.overrides[0].extends).toBeUndefined();
       expect(appEslint.overrides[1].extends).toBeUndefined();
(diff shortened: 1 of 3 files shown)


e2e/linter/src/linter.test.ts
```diff
@@ -572,10 +569,61 @@ describe('Linter', () => {
...
+      ]);
+      expect(e2eEslint.overrides[0].extends).toEqual([
+        'plugin:@nrwl/nx/typescript',
+      ]);
+      expect(e2eEslint.overrides[1].extends).toEqual([
+        'plugin:@nrwl/nx/javascript',
       ]);
+
+      runCLI(`generate @nrwl/workspace:lib ${mylib} --no-interactive`);
+      // should add new tslint
```
barbados-clemens, line 607: nitpick: comment isn't right? was it supposed to be eslint?

e2e/linter/src/linter.test.ts
```diff
@@ -572,10 +569,61 @@ describe('Linter', () => {
...
+        'plugin:@nrwl/nx/typescript',
+      ]);
+      expect(rootEslint.overrides[1].extends).toEqual([
+        'plugin:@nrwl/nx/javascript',
+      ]);
+      expect(e2eEslint.overrides[0].extends).toEqual([
+        'plugin:@nrwl/nx/typescript',
+      ]);
+      expect(e2eEslint.overrides[1].extends).toEqual([
+        'plugin:@nrwl/nx/javascript',
```
barbados-clemens, line 603: nitpick: personally this doesn't read as an e2e test and instead is more unit test like. 
instead we should just run the lint target and assert that it works instead of checking properties of generated files (where a unit test is more likely to already cover this). since this test is now dependent on the order of the file contents instead of how the user interacts with Nx like a true e2e test should be.

I'd also recommend using expect.objectContaining and expect.arrayContaining instead of asserting each index of the override otherwise this test will have to be updated every time the array order changes which is a bit too much 'context' that an e2e test should have IMO.

Just a though.

e2e/linter/src/linter.test.ts
meeroslav: It's actually not adding anything. I will remove that line

e2e/linter/src/linter.test.ts
meeroslav: The reason we are checking the expected content, if that there is a scenario where `nx lint ...` would pass but running `eslint ...` would fail.

But you are correct about running `lint` command to verify things are working and using potentially smarter checks.

meeroslav: I improved the tests, unfortunately, this doubled the duration of the test suite.

barbados-clemens (approved):

ndcunningham (approved):

jaysoo (approved):

FrozenPandaz (approved):

PR #15709, opened 2023-03-16
fix(storybook): enable e2e tests

mandarini:
Storybook e2e tests still had some tests and assertions that corresponded to outdated code!

* Remove two tests that do not make sense any more
* Remove react tests from Angular tests
* Re-enable tests
* Remove assertions that don't make sense any more
* Skip tests that are running or building storybook, since there are issues with Storybook v6 and `pnpm`, and with Node 18.

### DIFF
--- a/e2e/storybook-angular/src/storybook-angular.test.ts
+++ b/e2e/storybook-angular/src/storybook-angular.test.ts
@@ -2,10 +2,8 @@ import {
   checkFilesExist,
   cleanupProject,
   getPackageManagerCommand,
-  isNotWindows,
   killPorts,
   newProject,
-  readFile,
   runCLI,
   runCommand,
   runCypressTests,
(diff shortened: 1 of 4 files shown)


Coly010 (approved):

mandarini:
@meeroslav it's not the `buildable=true`. I removed it and the issue is still there.

e2e/storybook-angular/src/storybook-angular.test.ts
```diff
@@ -2,321 +2,120 @@ import {
...
-      } else {
-        json['pnpm'] = {
-          overrides: {
-            'enhanced-resolve': '5.10.0',
-          },
-        };
-      }
-      return json;
+    process.env.SELECTED_PM = 'yarn';
+    runCreateWorkspace(proj, {
```
meeroslav, line 25: This function is not cached. Please revert back to using `newProject`

e2e/storybook-angular/src/storybook-angular.test.ts
```diff
@@ -2,321 +2,120 @@ import {
...
-      `
-        module.exports = {
-          stories: [],
-          addons: ['@storybook/addon-essentials'],
-        };
-
-        console.log('hi there');
-      `
-    );
+    runCommand(getPackageManagerCommand().install);
```
meeroslav, line 36: Install should be part of the `@nrwl/angular:storybook-configuration`. If it's not we need to fix it there

e2e/storybook-angular/src/storybook-angular.test.ts
```diff
@@ -2,321 +2,120 @@ import {
...
-        };
-      } else {
-        json['pnpm'] = {
-          overrides: {
-            'enhanced-resolve': '5.10.0',
-          },
-        };
-      }
-      return json;
+    process.env.SELECTED_PM = 'yarn';
```
meeroslav, line 24: We need to test if this functionality works across all the package managers. This way we are constantly checking just `yarn`.

e2e/storybook-angular/src/storybook-angular.test.ts
```diff
@@ -2,321 +2,120 @@ import {
...
-
-        runCLI(
-          `generate @nrwl/angular:storybook-configuration ${myAngularLib} --configureCypress --generateStories --generateCypressSpecs --no-interactive`
-        );
-        runCLI(
-          `generate @nrwl/angular:stories ${myAngularLib} --generateCypressSpecs --no-interactive`
-        );
-
+    it('should execute e2e tests using Cypress running against Storybook', async () => {
+      if (isNotWindows() && runCypressTests()) {
```
meeroslav, line 63: Perhaps the `isNotWindows` check is not outdated and should be removed

meeroslav (changes requested):
I left some notes that should be addressed.

I also think this PR diverged a bit. Let's have one PR that fixes existing tests and then have a followup that checks for storybook 7.

Mixing those two just leads to more moving parts that potentially break.

e2e/storybook-angular/src/storybook-angular.test.ts
meeroslav: Reverting this will likely fix one of the E2E tests that are currently timing out because we set the limit to 1000s and it took 1002s.

mandarini: Hmm I made that change because the `newProject` installs ALL packages, and not all packages are needed. I can add a parameter there maybe?

e2e/storybook-angular/src/storybook-angular.test.ts
mandarini: Oops yes that was left there from the `enhance` package addition!

mandarini:
@meeroslav most of the tests (building/serving) were disabled, because Storybook cannot build/serve with v6.5+Node 18 for webpack/vite. This is why I thought that it's not worth it. But yes I can disable these, and then enable them in another PR!

e2e/storybook-angular/src/storybook-angular.test.ts
mandarini: `pnpm` is not supported by 6.5, it needs Storybook 7

mandarini: https://github.com/storybookjs/storybook/issues/13428#issuecomment-1368062647

e2e/utils/create-project-utils.ts
```diff
@@ -57,27 +58,29 @@ export function newProject({
...
-        `@nrwl/react`,
-        `@nrwl/storybook`,
-        `@nrwl/vite`,
-        `@nrwl/web`,
-        `@nrwl/webpack`,
-        `@nrwl/react-native`,
-        `@nrwl/expo`,
-      ];
+      const packages = packagesToInstall?.length
+        ? packagesToInstall
```
meeroslav, line 62: This is not so simple. We use a single cache per machine which would lead to all other E2E projects running on that same machine after storybook ending up with only `angular` and `storybook`.

Can you revert this change, please?

meeroslav (changes requested):
I think the addition to newProject should be done in a separate PR and carefully planned out. I'm afraid the current implementation would create randomly flaky tests.

e2e/utils/create-project-utils.ts
mandarini: But let's find a way to not have to install ALL packages, or maybe improve the `runCreateWorkspace` function!

barbados-clemens (approved):

meeroslav (approved):

PR #15714, opened 2023-03-16
cleanup(angular): replace wrapAngularDevkitSchematics in library generator

Coly010:
## Current Behavior

We use wrapAngularDevkitSchematics when we create Angular libraries.

## Expected Behavior

Generate the files ourselves

## Related Issue(s)

Fixes #


packages/angular/src/generators/library/files/base/tsconfig.json__tpl__
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "files": [],
+  "include": [],
+  "references": [
+    {
+      "path": "./tsconfig.lib.json"
+    },<% if(unitTesting) { %>
+    {
+      "path": "./tsconfig.spec.json"
+    }<% } %>
```
leosvelperez, line 14: I just realized we shouldn't handle any test-related config & files (except spec files) in the `library` generator. We invoke the `jestProject` generator and that one sets up any file and config needed for a project. It handles this change and others we're doing in this PR.

The same applies to what we already did in the `application` generator. We need to amend that.

packages/angular/src/generators/library/files/base/tsconfig.spec.json__tpl__
leosvelperez, line 1: Remove this file

packages/angular/src/generators/library/files/base/package.json__tpl__
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "<%= npmScope %><%= libFileName %>",
```
leosvelperez, line 2: `npmScope` might not be defined. It's not a required property of the `nx.json` and we're missing the `/` here.
```suggestion
  "name": "<% if (npmScope) { %><%= npmScope %>/<% } %><%= libFileName %>",
```

packages/angular/src/generators/library/lib/create-files.ts
```diff
@@ -0,0 +1,92 @@
...
+          project.sourceRoot,
+          `lib/${options.libraryOptions.fileName}.module.ts`
+        )
+      );
+    }
+  }
+
+  if (options.libraryOptions.unitTestRunner === UnitTestRunner.None) {
+    tree.delete(joinPathFragments(project.root, `tsconfig.spec.json`));
+  }
```
leosvelperez, line 78: Remove this

packages/angular/src/generators/library/files/lib/README.md__tpl__
leosvelperez, line 1: Why are we removing the `README`?

packages/angular/src/generators/library/lib/update-project.ts
```diff
@@ -1,188 +1,13 @@
...
-import { join } from 'path';
 
 export async function updateProject(
   host: Tree,
   options: NormalizedSchema['libraryOptions']
 ) {
-  createFiles(host, options);
   updateProjectTsConfig(host, options);
-  fixProjectWorkspaceConfig(host, options);
   updateNgPackage(host, options);
```
leosvelperez, line 10: Do we still need to update `ng-package.json` here? The template should have all that's needed, isn't it?

packages/angular/src/generators/library/lib/update-project.ts
```diff
@@ -1,188 +1,13 @@
...
 import { NormalizedSchema } from './normalized-schema';
 import { updateNgPackage } from './update-ng-package';
-import { join } from 'path';
 
 export async function updateProject(
   host: Tree,
   options: NormalizedSchema['libraryOptions']
 ) {
-  createFiles(host, options);
   updateProjectTsConfig(host, options);
```
leosvelperez, line 9: Do we still need to update the tsconfig here? The template should have all that's needed, isn't it?

leosvelperez (changes requested):

packages/angular/src/generators/library/files/base/tsconfig.json__tpl__
leosvelperez: I created a PR for the application generator to address this https://github.com/nrwl/nx/pull/15731.

packages/angular/src/generators/library/files/lib/README.md__tpl__
Coly010: bad copy i think

Coly010: actually, it's just been moved

leosvelperez (approved):

PR #15723, opened 2023-03-16
feat(nextjs): update nextjs version to 13.2.3

ndcunningham:


packages/next/migrations.json
```diff
@@ -295,6 +295,23 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "15.8.6": {
+      "version": "15.8.6-beta.0",
+      "packages": {
+        "next": {
+          "version": "13.2.3",
```
FrozenPandaz, line 303: Latest is `13.2.4`. Let's use that.

packages/next/migrations.json
```diff
@@ -295,6 +295,23 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "15.8.6": {
+      "version": "15.8.6-beta.0",
```
FrozenPandaz, line 300: Can you change this to `15.9.0-beta.4` please?

packages/next/src/utils/versions.ts
```diff
@@ -1,8 +1,8 @@
 export const nxVersion = require('../../package.json').version;
 
-export const nextVersion = '13.1.1';
-export const eslintConfigNextVersion = '13.1.1';
-export const sassVersion = '1.55.0';
+export const nextVersion = '13.2.3';
```
FrozenPandaz, line 3: Use `13.2.4`

FrozenPandaz (changes requested):

terrymun:
I noticed that the [storybook build seems to have stalled at 64%](https://app.circleci.com/pipelines/github/nrwl/nx/29535/workflows/d8f79c38-7292-429a-87a7-9ff0c3505ab2/jobs/299152/parallel-runs/0/steps/0-114) due to an error thrown by webpack:

```
ERR! => Failed to build the preview
ERR! Module not found: Error: Can't resolve 'crypto' in '/tmp/nx-e2e--1922-u6fp8WcLr0u2/nx/proj/node_modules/next/dist/shared/lib/bloom-filter'
ERR! 
ERR! BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
ERR! This is no longer the case. Verify if you need this module and configure a polyfill for it.
ERR! 
ERR! If you want to include a polyfill, you need to:
ERR! 	- add a fallback 'resolve.fallback: { "crypto": require.resolve("crypto-browserify") }'
ERR! 	- install 'crypto-browserify'
ERR! If you don't want to include a polyfill, you can use an empty module like this:
ERR! 	resolve.fallback: { "crypto": false }
```

I've experienced the same too, and had to remove `crypto` as a fallback in the webpack config on storybook's `main.js` or `main.ts` file:

```ts
webpackFinal: async (config) => {
  if (config.resolve) {
    config.resolve = {
      ...config.resolve,
      fallback: {
        ...config.resolve.fallback,
        crypto: false,
      },
    };
  }
 
  return config;
},
```

PR #15731, opened 2023-03-17
cleanup(angular): remove tsconfig.spec.json handling from application generator

leosvelperez:
## Current Behavior

The application generator is handling the `tsconfig.spec.json` even though that's already handled by the Jest project generator.

## Expected Behavior

The application generator should not handle the `tsconfig.spec.json` and leave it to the Jest project generator.

## Related Issue(s)

Fixes #


packages/angular/src/generators/application/files/base/tsconfig.json__tpl__
```diff
@@ -8,12 +8,6 @@
   "references": [
     {
       "path": "./tsconfig.app.json"
-    },
-    <% if(unitTesting) { %>{
-      "path": "./tsconfig.spec.json"
-    },<% } %>
-    {
-      "path": "./tsconfig.editor.json"
```
leosvelperez, line 16: Adding the `tsconfig.editor.json` is handled by the generator's code to ensure it is the last reference.

Coly010 (approved):

FrozenPandaz (approved):

PR #15732, opened 2023-03-17
chore(graph): remove e2e-graph-client from e2e-matrix

philipjfulcher:
## Current Behavior
`e2e-graph-client` is run during the nightly e2e-matrix

## Expected Behavior
`e2e-graph-client` is not run during the nightly e2e-matrix

There some issues with flakiness and Cypress caching with the `e2e-graph-client` when run during the e2e matrix. Because the graph client is covered by e2e tests on PRs, I don't think there's a need to run it during the e2e matrix. The tests doesn't run against a real Nx install, it's just mocking data and testing the frontend. The frontend is isolated from changes to operating system changes, external changes, Nx itself, so the e2e matrix likely won't catch any regressions.

### DIFF
--- a/.github/workflows/e2e-matrix.yml
+++ b/.github/workflows/e2e-matrix.yml
@@ -86,7 +86,6 @@ jobs:
           - e2e-webpack
           - e2e-workspace-create
           - e2e-e2e-workspace-create-npm
-          - e2e-graph-client
         include:
           # os short names
           - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


FrozenPandaz (approved):

meeroslav:
I'm afraid I have to disagree with this PR being merged in this form.

If a test is (supposed to be) run on CI, that doesn't mean we shouldn't run it nightly as well. Our CI uses caching and affected, all of which is non-existent on nightly run. Additionally, we had cases where things were accidentally skipped on CI. Nightly provides a second layer of defense.

I strongly believe we should have a test for graph without mocked data, but even with mocked data we should run it at least for one combination - e.g. Node 16, Npm, Ubuntu.

@philipjfulcher please consider bringing it back or implementing another test that would not be flaky.

PR #15769, opened 2023-03-20
cleanup(angular): replace wrapAngularDevkitSchematics for Web Worker

Coly010:
## Current Behavior

We use `wrapAngularDevkitSchematics` in Web Worker Generator

## Expected Behavior

Replace `wrapAngularDevkitSchematics` in the generator

## Related Issue(s)

Fixes #


packages/angular/src/generators/web-worker/lib/normalize-options.ts
```diff
@@ -1,10 +1,37 @@
...
+import { getProjects, joinPathFragments } from '@nrwl/devkit';
 
 export function normalizeOptions(
+  tree: Tree,
   options: WebWorkerGeneratorOptions
 ): WebWorkerGeneratorOptions {
+  if (!options.project) {
+    throw new Error('You must provide a project!');
+  }
+
```
leosvelperez, line 12: This shouldn't be needed. The `project` option is required in the schema and it will be automatically validated. Same as the `name` which we are not validating here.
```suggestion
```

packages/angular/src/generators/web-worker/lib/normalize-options.ts
```diff
@@ -1,10 +1,37 @@
...
+  const selectedProject = projects.get(options.project);
+  if (selectedProject.projectType !== 'application') {
+    throw new Error(
+      `Web Worker can only be added to an application. Project '${options.project}' is a library.`
+    );
+  }
+
+  const path =
+    options.path ?? joinPathFragments(selectedProject.sourceRoot, 'app');
```
leosvelperez, line 26: `sourceRoot` is optional, let's handle it:
```suggestion
  const path =
    options.path ?? selectedProject.sourceRoot
      ? joinPathFragments(selectedProject.sourceRoot, 'app')
      : joinPathFragments(selectedProject.root, 'src', 'app');
```

leosvelperez (changes requested):

packages/angular/src/generators/web-worker/lib/normalize-options.ts
```diff
@@ -1,10 +1,41 @@
...
+  joinPathFragments,
+  readProjectConfiguration,
+} from '@nrwl/devkit';
 
 export function normalizeOptions(
+  tree: Tree,
   options: WebWorkerGeneratorOptions
 ): WebWorkerGeneratorOptions {
+  if (!getProjects(tree).has(options.project)) {
+    throw new Error(`Project '${options.project} does not exist!`);
```
leosvelperez, line 14: Missing ending quote:
```suggestion
    throw new Error(`Project '${options.project}' does not exist!`);
```

leosvelperez (changes requested):

leosvelperez (approved):

PR #15776, opened 2023-03-20
feat(core): do not affect all if lock changed but not root package.json

### LINKED ISSUE #10842: Yarn+NX rebuilds all workspaces with any package.json dependency change
## Current Behavior

NX rebuilds all yarn workspaces when the package.json dependencies of any workspace change. Since we are using internally-published packages and iterating them frequently, this causes the project to rebuild way too often.

## Expected Behavior

NX only rebuilds workspaces with actual dependency changes.

## Steps to Reproduce

1. Clone https://github.com/gbisaga/nx-rebuild-all
2. `yarn install --frozen-lockfile`
3. `yarn build:all`
4. `yarn workspace @ps-refarch/nxpackage3 add jsonwebtoken`
5. `yarn build:all`

Expected behavior: only @foo-monorepo/nxpackage3 rebuilds
Actual behavior: all 3 workspaces rebuild

### Environment

   Node : 16.13.1
   OS   : win32 x64
   yarn : 1.22.18

   nx : 14.3.1
   @nrwl/angular : Not Found
   @nrwl/cypress : Not Found
   @nrwl/detox : Not Found
   @nrwl/devkit : 14.3.1
   @nrwl/eslint-plugin-nx : Not Found
   @nrwl/express : Not Found
   @nrwl/jest : 14.3.1
   @nrwl/js : Not Found
   @nrwl/linter : 14.3.1
   @nrwl/nest : Not Found
   @nrwl/next : Not Found
   @nrwl/node : Not Found
   @nrwl/nx-cloud : Not Found
   @nrwl/nx-plugin : Not Found
   @nrwl/react : Not Found
   @nrwl/schematics : Not Found
   @nrwl/storybook : Not Found
   @nrwl/web : Not Found
   @nrwl/workspace : 14.3.1
   typescript : 4.7.3
   Community plugins:

ianldgs:
## Current Behavior

Given an NX workspace using yarn, npm or pnpm workspaces, with projects `a`, `b`, and `c`, where each of them has a package.json;

Installing or updating a dependency to a specific package:
- `yarn workspace a add {package}`
- `pnpm i --filter a {package}`
- or `npm install {package} -w a`

Causes all projects (`a`, `b`, and `c`) to be affected (implicit).

## Expected Behavior

Only project `a` should be affected (explicit).

## Related Issue(s)
[](https://github.com/nrwl/nx/issues/10842)

Fixes #10842

## Notes

The logic is simple: only consider the lock files as global/implicit inputs if the root package.json changed.
Locator `getImplicitlyTouchedProjects` will not return any project otherwise.
Projects that were really affected will be return by `getTouchedProjects`.

This is mostly safe because the lock file should only change when there are changes in any package.json.
The root package json is where NX and shared dependencies are declared, so updating it will likely trigger a full build anyway.
Changing a project's package.json is expected to only affect the touched project.

### DIFF
--- a/packages/nx/src/project-graph/affected/locators/workspace-projects.spec.ts
+++ b/packages/nx/src/project-graph/affected/locators/workspace-projects.spec.ts
@@ -76,6 +76,21 @@ describe('getTouchedProjects', () => {
       getTouchedProjects(fileChanges, buildProjectGraphNodes(projects))
     ).toEqual(['b']);
   });
+
+  it('should return projects with the package.json changed', () => {
+    const fileChanges = getFileChanges([
+      'libs/a/package.json',
+      'libs/c/package.json',
+    ]);
+    const projects = {
+      a: { root: 'libs/a' },
+      b: { root: 'libs/b' },
+      c: { root: 'libs/c' },
+    };
+    expect(
+      getTouchedProjects(fileChanges, buildProjectGraphNodes(projects))
+    ).toEqual(['a', 'c']);
+  });
 });
 
 describe('getImplicitlyTouchedProjects', () => {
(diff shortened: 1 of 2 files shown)


meeroslav:
Thank you for this PR @ianldgs, but unfortunately, such functionality would not be accurate and would lead to false affected graph.

If you just run the `install` for example (e.g. `yarn install`) your packages and lock file could be changed, without ever touching `package.json`, and some of those changes might influence all the projects (imagine you have `nx` defined as `^15`).

Not sure how the original issue could be tackled, but even if we check whether the project's nested `package.json` was changed, that still doesn't give us enough information on what scope of changes happened on lock file.

ianldgs:
@meeroslav although that is true, what I've seen lately is that `npm install` with later versions of NPM almost never changes the lock. Same with pnpm. Not sure about yarn.
And people are anyway not supposed to run those commands. They should be running `npm ci` or using the `--frozen-lockfile` flag.
But I understand the point.

How about changing `getTouchedNpmPackages` to track that? It could look at the lock file instead and check which versions really changed.

meeroslav:
We are currently working on refactoring that part of the code, so `package.json` will no longer be checked, only the lock files.

Unfortunately, this doesn't really help to resolve your problem, but we'll see if there might be some easy flag to resolve this.

PR #15777, opened 2023-03-20
feat(bundling): add for esbuild to enable/disable package.json generation

jaysoo:
This PR adds a `generatePackageJson` option to `@nrwl/esbuild:esbuild`, same as webpack, so the user can opt out of `package.json` generation.

**Notes:**
- Migration is provided for existing projects
- New Node apps and JS libs are created with `generatePackageJson: true` (so behavior is intact)
- User who choose to disable it will have `package.json` copied to dist as-is from source

## Current Behavior
Cannot disable `package.json` generation.

## Expected Behavior
Can disable `package.json`, and have it copied as an asset.

## Related Issue(s)

Fixes #


ndcunningham (approved):
LGTM

ndcunningham:
```
Object literal may only specify known properties, and 'project' does not exist in type 'NormalizedEsBuildExecutorOptions'.
```

Maybe you are still passing `project` inside your unit test `build-esbuild-options.spec.ts`

PR #15778, opened 2023-03-20
fix(bundling): respect --sourcemap option for esbuild

jaysoo:
This PR fixes an issue where `--sourcemap` from CLI or `project.json` is not respected when running the build. Note, this option should supercede anything in `esbuildOptions` since the former can be from the CLI, which takes precedence when running in different contexts (e.g. CI, dev, prod, etc.).

## Current Behavior
`--sourcemap` does nothing

## Expected Behavior
`--sourcemap` should be passed down to esbuild as an option

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/esbuild/executors/esbuild.json
+++ b/docs/generated/packages/esbuild/executors/esbuild.json
@@ -114,7 +114,13 @@
         "default": false
       },
       "sourcemap": {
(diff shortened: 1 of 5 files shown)


ndcunningham (approved):
LGTM

PR #15797, opened 2023-03-21
feat(web): remove hard dependencies on cypress, rollup, jest, and linter

jaysoo:
This PR cleans up the dependencies of `@nrwl/web` so that it can be used by other plugins without pulling in unnecessary packages. Use `ensurePackage` to install the packages as needed from generators.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/web/src/generators/application/schema.d.ts
+++ b/packages/web/src/generators/application/schema.d.ts
@@ -1,4 +1,4 @@
-import { Linter } from '@nrwl/linter';
+import type { Linter } from '@nrwl/linter';
 
 export interface Schema {
   name: string;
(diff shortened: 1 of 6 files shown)


packages/web/src/generators/application/schema.d.ts
```diff
@@ -1,4 +1,4 @@
-import { Linter } from '@nrwl/linter';
+import type { Linter } from '@nrwl/linter';
```
ndcunningham, line 2: Should the `linter` here be marked with `none` as well?

`linter?: Linter | 'none'`

jaysoo: `'none'` is already in `Linter`, we just never updated web :)

mandarini (approved):

ndcunningham (approved):

PR #15803, opened 2023-03-21
feat(angular): generate serve-static for e2e tests rather than run the dev-server

jaysoo:
This PR updates the existing Angular setup such that a `serve-static` target is added to the app, and e2e project will use `serve-static` rather than the dev-server. If the application is already built, the start-up is much faster e.g. in CI.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/application/lib/add-e2e.ts
+++ b/packages/angular/src/generators/application/lib/add-e2e.ts
@@ -1,12 +1,20 @@
 import type { Tree } from '@nrwl/devkit';
 import type { NormalizedSchema } from './normalized-schema';
 
+import {
+  readProjectConfiguration,
+  updateProjectConfiguration,
+} from '@nrwl/devkit';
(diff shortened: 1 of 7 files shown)


packages/angular/src/generators/application/lib/add-e2e.ts
```diff
@@ -16,6 +24,32 @@ export async function addE2e(tree: Tree, options: NormalizedSchema) {
       standaloneConfig: options.standaloneConfig,
       skipPackageJson: options.skipPackageJson,
       rootProject: options.rootProject,
+      devServerTarget: 'serve-static',
```
barbados-clemens, line 27: I'm not sure if we should change the targets to always use the static file server or only in the CI configuration. 

Personally, I think we should only update the CI configuration for the e2e tests to use the static server as locally using the static server to iterate on work provides a poor experience, since you have to wait for a full build.

PR #15816, opened 2023-03-22
cleanup(angular): remove v14 folder for init

Coly010:
## Current Behavior

We have duplicated code to support v14 for Angular in the init generator

## Expected Behavior

Instead of the duplicated code, use branches to determine the correct versions to install

## Related Issue(s)

Fixes #


packages/angular/src/generators/init/init.ts
```diff
@@ -60,7 +50,9 @@ export async function angularInitGenerator(
     if (!packageVersion) {
       devkitVersion ??=
         getInstalledPackageVersion(tree, '@angular-devkit/build-angular') ??
-        angularDevkitVersion;
+        (installedAngularVersionInfo.major === 14
```
leosvelperez, line 53: Instead of repeating this conditional code everywhere, what about doing something like:

```ts
import * as latestVersions from '../../utils/versions';
import { backwardCompatibleVersions } from '../../utils/backward-compatible-versions';

...

//  this could be turned into a utility with a switch/case to handle multiple older versions (e.g. when we support v16, we'll have v14 and v15)
const versions = installedAngularVersionInfo.major === 14
  ? backwardCompatibleVersions.angularV14
  : latestVersions;

...

devkitVersion ??=
  getInstalledPackageVersion(tree, '@angular-devkit/build-angular') ?? 
  versions.angularDevkitVersion
```

Coly010: Can we address this in a further PR? Right now i just want to get rid of the duplicated code, but i can do another pass of cleanup where i create a util that i can use to replace all occurrences of this syntax in all our generators

leosvelperez: Up to you. It's a very simple util though with just a couple of lines (most of the code is in the snippet above) that could be used in the next PRs you're cleaning up. You're already cleaning up. It's the best chance to do these things. Otherwise, you'll cleanup the cleanup :)

leosvelperez (approved):

PR #15833, opened 2023-03-22
fix(core): fix broken path elevation in npm lockfile pruning

meeroslav:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/lock-file/npm-parser.ts
+++ b/packages/nx/src/lock-file/npm-parser.ts
@@ -580,10 +580,20 @@ function elevateNestedPaths(
       `${segs.join('/node_modules/')}/node_modules/${packageName}`;
 
     // check if grandparent has the same package
-    while (
-      segments.length > 1 &&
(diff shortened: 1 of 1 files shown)


packages/nx/src/lock-file/npm-parser.ts
```diff
@@ -580,10 +580,20 @@ function elevateNestedPaths(
...
-      segments.length > 1 &&
-      !result.has(getNewPath(segments.slice(0, -1)))
-    ) {
+    const shouldElevate = (segs: string[]) => {
+      const newPath = getNewPath(segs.slice(0, -1));
+      if (result.has(newPath)) {
+        const match = result.get(newPath);
+        const source = remappedPackages.get(path);
+        return (
+          match.valueV1?.version === source.valueV1?.version &&
```
meeroslav, line 589: If a parent already exists, check if it's the same version. In which case, it's ok to pop the segment and shorten the path.

FrozenPandaz (approved):

PR #15834, opened 2023-03-22
cleanup(misc): remove internal rootProject flag from generators that can derive it

leosvelperez:
## Current Behavior

An internal `rootProject` flag is used by generators to know whether they are running for a standalone project. Some generators shouldn't need this flag and instead, they shoul derive it from the project configuration.

## Expected Behavior

The project generators from `cypress` and `jest` shouldn't have the `rootProject` flag. They should derive the information from the project configuration.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/cypress/src/generators/cypress-project/cypress-project.ts
+++ b/packages/cypress/src/generators/cypress-project/cypress-project.ts
@@ -43,6 +43,7 @@ import { Schema } from './schema';
 export interface CypressProjectSchema extends Schema {
   projectName: string;
   projectRoot: string;
+  rootProject: boolean;
 }
 
 function createFiles(tree: Tree, options: CypressProjectSchema) {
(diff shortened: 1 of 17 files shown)


packages/cypress/src/generators/cypress-project/cypress-project.ts
```diff
@@ -43,6 +43,7 @@ import { Schema } from './schema';
 export interface CypressProjectSchema extends Schema {
   projectName: string;
   projectRoot: string;
+  rootProject: boolean;
```
FrozenPandaz, line 46: This is getting added? Shouldn't it be getting removed?

FrozenPandaz (approved):

barbados-clemens (approved):
Thank you! I've been meaning to clean this up so much appricated!

Coly010 (approved):

PR #15840, opened 2023-03-22
fix(core): resolve some strip-source-code bugs

forivall:
I found some bugs, added some test cases, and used my knowledge of js parsing quirks around template and regex literals to resolve them.

The regex handling was adapted from https://github.com/Microsoft/typescript/blob/2c952fe8505fc427cdfe30ed4b40343f0f5f39bc/src/services/classifier.ts#L260

## Current Behavior
Imports after a template literal with 2 variables are not analyzed

## Expected Behavior
These imports should be included in the dependency graph

## Related Issue(s)

Fixes #


FrozenPandaz (approved):
This looks great! Thank you so much for looking into this. :tada:

PR #15842, opened 2023-03-23
fix: workaround `undefined` value crashing build

n370:
## Related Issue(s)

Fixes #


n370:
I'm sorry, this won't work. Build complete but results in unusable bundle. I'll reopen this and add more details if I figure out whats is really going on.

PR #15849, opened 2023-03-23
fix(angular): use @nrwl/web:file-server

Coly010:
## Current Behavior

We have a duplicated file-server executor in nrwl/angular

## Expected Behavior

Use the @nrwl/web as the file-server executor

## Related Issue(s)

Fixes #


leosvelperez (changes requested):
There's the following TODO that was left for when this was done https://github.com/nrwl/nx/blob/master/packages/angular/src/generators/application/lib/add-e2e.ts#L15-L16.

mandarini (approved):

packages/angular/src/migrations/update-15-9-0/update-file-server-executor.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    if (projectChanged) {
+      updateProjectConfiguration(tree, projectName, project);
+    }
+  }
+
+  if (needsNrwlWeb) {
+    addDependenciesToPackageJson(tree, {}, { '@nrwl/web': nxVersion });
+  }
```
leosvelperez, line 52: Sorry, I missed it before, let's format files at the end of the migration.

leosvelperez (changes requested):

leosvelperez (approved):

jaysoo (approved):

bcabanes (approved):

PR #15866, opened 2023-03-23
Update one-storybook-with-composition.md

rudrakumarg:
## Related Issue(s)

Fixes #


isaacplmann:
The link works without this change.

PR #15867, opened 2023-03-23
fix(core): remove dependency to linter from workspace plugin

jaysoo:
## Related Issue(s)

Fixes #


FrozenPandaz:
Changes are in https://github.com/nrwl/nx/pull/15865

PR #15888, opened 2023-03-24
fix(js): run prettier with default options when theres no prettierrc

FrozenPandaz:
## Current Behavior

When creating a new repo, files will be formatted with single quotes.

## Expected Behavior

When creating a new repo, files will not be formatted with single quotes.

## Related Issue(s)

Fixes #


AgentEnder (approved):

packages/angular/src/generators/stories/__snapshots__/stories-app.spec.ts.snap
```diff
@@ -175,6 +175,7 @@ Object {
     101,
     110,
     116,
+    44,
```
leosvelperez, line 178: Could you update the test cases for this snapshot so we match the actual `string` instead of a `Buffer`? This is an existing issue, but let's take the chance to correct it.

packages/vite/src/generators/vitest/vitest-generator.ts
```diff
@@ -41,7 +41,7 @@ export async function vitestGenerator(
 
   addOrChangeTestTarget(tree, schema, testTarget);
 
-  const initTask = initGenerator(tree, {
+  const initTask = await initGenerator(tree, {
```
mandarini, line 44: is this `await` here needed?

mandarini (approved):

PR #15898, opened 2023-03-27
docs(nx-dev): update typedoc version

leosvelperez:
## Current Behavior

The TypeDoc version installed doesn't support TypeScript 5.0 which will be needed soon for Angular 16.

## Expected Behavior

The TypeDoc version installed supports TypeScript 5.0.

Note: The new version **0.23.28** also supports older TS versions, so we can make this change ahead of the Angular 16 support. This reduces the size of the Angular 16 support PR and allows us to have a more focused PR.

## Related Issue(s)

Fixes #


bcabanes:
That should be okay if we go without the `@category`, I see it is still sorted but by type now. @isaacplmann what do you think?

isaacplmann:
I think this looks fine

bcabanes (approved):

isaacplmann (approved):

FrozenPandaz (approved):

PR #15910, opened 2023-03-27
feat(nx-dev): add table of content for documents

bcabanes:
It adds a table of content on the right side of documents for large viewports.


juristr (changes requested):
Looks a bit weird in dark mode

image

Optional improvement
- can we somehow specify on a page to **not have a TOC**? Thinking like on the main docs entry page where it might not be necessary

isaacplmann (approved):

juristr (approved):

PR #15940, opened 2023-03-28
fix(angular): handle projects with no targets in angular migration

FrozenPandaz:
## Current Behavior

Angular file-server migration errors when there is a project with no targets.

## Expected Behavior

Angular file-server migration skips over projects with no targets and does not error.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/migrations/update-15-9-0/update-file-server-executor.spec.ts
+++ b/packages/angular/src/migrations/update-15-9-0/update-file-server-executor.spec.ts
@@ -64,4 +64,22 @@ describe('updateFileServerExecutor', () => {
       '@nrwl/web:file-server'
     );
   });
+
+  it('should handle projects with no targets', async () => {
+    // ARRANGE
+    const tree = createTreeWithEmptyWorkspace();
+
+    const project = {
+      name: 'test',
+      root: '/',
+      sourceRoot: '/src',
+    };
+
+    addProjectConfiguration(tree, 'test', project);
+
+    // ACT
+    expect(async () => {
+      await updateFileServerExecutor(tree);
+    }).not.toThrow();
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/angular/src/migrations/update-15-9-0/update-file-server-executor.ts
```diff
@@ -32,21 +34,17 @@ export default async function updateFileServerExecutor(tree: Tree) {
...
-    for (const [targetName, target] of Object.entries(project.targets)) {
-      if (target.executor === oldExecutor) {
-        project.targets[targetName].executor = newExecutor;
-        projectChanged = true;
-        needsNrwlWeb = true;
-      }
+  forEachExecutorOptions(
+    tree,
+    '@nrwl/angular:file-server',
+    (currentValue, project, target, configuration) => {
```
leosvelperez, line 40: We are not updating options and configurations. Using `forEachExecutorOptions` would needlessly execute the callback for every configuration and once for the options, but we only need it once per target.

We could have a different utility like `forEachTargetWithExecutor` for these use cases, but while we don't have it, it's better here to just change the existing code like:

```diff
- for (const [targetName, target] of Object.entries(project.targets)) {
+ for (const [targetName, target] of Object.entries(project.targets ?? {})) {
```

packages/angular/src/migrations/update-15-9-0/update-file-server-executor.spec.ts
```diff
@@ -64,4 +64,22 @@ describe('updateFileServerExecutor', () => {
       '@nrwl/web:file-server'
     );
   });
+
+  it('it should change the executor correctly in nx.json', async () => {
```
leosvelperez, line 68:
```suggestion
  it('it should handle projects with no targets', async () => {
```

leosvelperez (changes requested):

packages/angular/src/migrations/update-15-9-0/update-file-server-executor.ts
Coly010: I concur. I had originally wrote this with forEachExecutorOptions, but as Leo has mentioned, it runs needlessly.

FrozenPandaz: Sure, we can do that.

leosvelperez (approved):

PR #15959, opened 2023-03-29
chore(repo): add Cypress cache to top-level install job

jaysoo:
This PR updates `CYPRESS_CACHE_FOLDER` to be an absolute path so it still works when running Cypress e2e tests within our Nx e2e tests. Previously, it was set as `CYPRESS_CACHE_FOLDER=.cypress` which would resolve to something like `/tmp/nx-e2e/nx/<proj>/.cypress`, and since we do not cache this folder it would fail with `Cypress binary not found` error.

To make it more resilient, there is a `Install Cypress` task in the case that the cache isn't found. This step is skipped if the cache is found, and if cache misses _but_ package manager install (e.g. `npm install`) already downloaded the binary it would be a no-op.

Here's an example of a successful React nightly test run that was previously failing due to missing Cypress binary: https://github.com/nrwl/nx/actions/runs/4554668922/jobs/8032914216.

## Current Behavior
Nightly tests do not use correct Cypress cache location.

## Expected Behavior
Nightly tests work with cached Cypress install.

### DIFF
--- a/.github/workflows/e2e-matrix.yml
+++ b/.github/workflows/e2e-matrix.yml
@@ -12,7 +12,7 @@ on:
         default: false
 
 env:
-  CYPRESS_CACHE_FOLDER: '.cypress'
+  CYPRESS_CACHE_FOLDER: ${{ github.workspace }}/.cypress
 
 permissions: {}
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-matrix.yml
```diff
@@ -56,6 +56,17 @@ jobs:
         if: steps.cache-modules.outputs.cache-hit != 'true'
         run: yarn install --prefer-offline --frozen-lockfile --non-interactive
 
+      - name: Cache Cypress
+        id: cache-cypress
+        uses: actions/cache@v3
+        with:
+          path: '${{ github.workspace }}/.cypress'
```
meeroslav, line 63: Please add "lookup-only" to skip fetching cache on the pre install job

.github/workflows/e2e-windows.yml
```diff
@@ -46,6 +46,17 @@ jobs:
         if: steps.cache-modules.outputs.cache-hit != 'true'
         run: yarn install --prefer-offline --frozen-lockfile --non-interactive
 
+      - name: Cache Cypress
+        id: cache-cypress
+        uses: actions/cache@v3
+        with:
+          path: '${{ github.workspace }}/.cypress'
```
meeroslav, line 53: Same as the other commend, please add lookup only

meeroslav (changes requested):
Left two minor change requests. Otherwise looks good

meeroslav (approved):

PR #15961, opened 2023-03-29
fix(misc): skip formatting files in nested generator calls

leosvelperez:
## Current Behavior

- Generators invoking other generators don't pass `skipFormat: true` to the nested generators which results in multiple `formatFiles` invocations during a generator run.
- Some generators invoked as part of other generator runs don't support the `skipFormat` flag.

The above have performance implications in the generator runs.

## Expected Behavior

Generators invoking other generators explicitly pass the `skipFormat: true` to prevent multiple `formatFiles` runs during a generator run.

## Related Issue(s)

Fixes #


mandarini (approved):

AgentEnder (approved):
nx-plugin changes look good

meeroslav (approved):

FrozenPandaz (approved):

PR #15987, opened 2023-03-30
feat(core): allow referencing other packages to specify implementations for executors + generators

FrozenPandaz:
## Current Behavior

Generators and executors cannot specify that their schema and implementation are in a different package.

## Expected Behavior

Generators and executors can specify that their schema and implementation are in a different package.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/config/workspaces.ts
+++ b/packages/nx/src/config/workspaces.ts
@@ -40,7 +40,6 @@ import {
 import { getNxRequirePaths } from '../utils/installation-directory';
 import { getIgnoredGlobs } from '../utils/ignore';
 import {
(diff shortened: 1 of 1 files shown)


AgentEnder (changes requested):
I don't think this will work -- This makes the migration story for plugin's really hard. If a community plugin extends one of our plugin's executors collections its easy for updating Nx to break that plugin's targets since their options would not get migrated.

At minimum we need to only allow this for @nrwl -> @nx, and remove it after v17. Either way this is quite ugly imho, I'm not sure its the path we should go down 🤔

AgentEnder (approved):

PR #16017, opened 2023-03-31
fix(angular): fix path and selector handling in directive generator

leosvelperez:
## Current Behavior

The directive generator:

- doesn't handle that `npmScope` could be `undefined`
- doesn't add the `app` or `lib` segment to the path based on the project type

## Expected Behavior

The directive generator builds correctly the selector and path.

Note: Follow-up PRs will make this logic more consistent across `component`, `directive` and `pipe` generators.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/directive/lib/normalize-options.ts
+++ b/packages/angular/src/generators/directive/lib/normalize-options.ts
@@ -1,0 +1,54 @@
+import type { ProjectConfiguration, Tree } from '@nrwl/devkit';
+import {
+  joinPathFragments,
+  names,
+  readNxJson,
+  readProjectConfiguration,
+} from '@nrwl/devkit';
+import { parseName } from '../../utils/names';
+import type { Schema } from '../schema';
+
+export function normalizeOptions(tree: Tree, options: Schema) {
+  const { prefix, projectType, root, sourceRoot } = readProjectConfiguration(
+    tree,
+    options.project
+  ) as ProjectConfiguration & { prefix?: string };
+
+  const projectSourceRoot = sourceRoot ?? joinPathFragments(root, 'src');
+  const { name, path: namePath } = parseName(options.name);
+
+  const path =
+    options.path ??
+    joinPathFragments(
+      projectSourceRoot,
+      projectType === 'application' ? 'app' : 'lib',
+      namePath
+    );
+
+  const selector =
+    options.selector ?? buildSelector(tree, name, options.prefix, prefix);
+
+  return {
+    ...options,
+    name,
+    path,
+    projectRoot: root,
+    projectSourceRoot,
+    selector,
+  };
+}
+
+function buildSelector(
+  tree: Tree,
+  name: string,
+  prefix: string,
+  projectPrefix: string
+): string {
+  let selector = name;
+  prefix ??= projectPrefix ?? readNxJson(tree).npmScope;
+  if (prefix) {
+    selector = `${prefix}-${selector}`;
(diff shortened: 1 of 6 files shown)


packages/angular/src/generators/utils/names.ts
```diff
@@ -0,0 +1,11 @@
+import { normalizePath } from '@nrwl/devkit';
+
+export type NameInfo = { name: string; path: string };
+
+export function parseName(rawName: string): NameInfo {
```
Coly010, line 5: Is this function in devkit? How do we handle this for creating projects?

packages/angular/src/generators/directive/lib/validate-options.ts
```diff
@@ -0,0 +1,13 @@
...
+import type { Schema } from '../schema';
+
+export function validateOptions(tree: Tree, options: Schema): void {
+  const projects = getProjects(tree);
+  if (!projects.has(options.project)) {
+    throw new Error(`Project "${options.project}" does not exist!`);
+  }
+
+  checkPathUnderProjectRoot(tree, options.project, options.path);
+}
```
Coly010, line 13: I wonder if we could make this shared between multiple generators. What are your thoughts?

packages/angular/src/generators/directive/lib/normalize-options.ts
```diff
@@ -0,0 +1,54 @@
...
+  const path =
+    options.path ??
+    joinPathFragments(
+      projectSourceRoot,
+      projectType === 'application' ? 'app' : 'lib',
+      namePath
+    );
+
+  const selector =
+    options.selector ?? buildSelector(tree, name, options.prefix, prefix);
```
Coly010, line 29: Could this be extracted to a shared util for other generators? 

This same logic would be used for pipe + component too right?

We'd need to change buildSelector to potentially accept a final argument, something like

```ts
export function buildSelector(Tree, string, string, string, keyof ReturnType<names>)

// then it could be used like

buildSelector(tree, name, options.prefix, projectPrefix, 'propertyName')
```

Coly010 (approved):
Approved! But added some comments for consideration

packages/angular/src/generators/directive/lib/validate-options.ts
leosvelperez: Possibly. As mentioned in the PR description: `Follow-up PRs will make this logic more consistent across component, directive and pipe generators`.

When making this consistent across the other similar generators, I'll extract whatever makes sense.

packages/angular/src/generators/directive/lib/normalize-options.ts
leosvelperez: Yeah. As mentioned in the description, I'll make the logic consistent in follow-up PRs. That includes extracting and making reusable whatever makes sense.

packages/angular/src/generators/utils/names.ts
leosvelperez: This is how the Angular DevKit directive schematic does and what we have in the component generator (https://github.com/leosvelperez/nx/blob/master/packages/angular/src/generators/component/lib/normalize-options.ts#L15-L17). I just made sure to make it consistent and have the directive generator behave the same. It's not meant to be used for creating projects. It's a parsing that would handle the scenario in which someone specifies a path in the name. I think it makes sense to handle it. It doesn't break existing behavior and it's aligned with the Angular DevKit directive schematic.

PR #16024, opened 2023-03-31
feat(nest): adding simpleName option to library generator

### LINKED ISSUE #14654: Add ability to specify library name not based on directory path using @nrwl/nest:library

## Description

Using the following command, the following output is generated:

```ts
yarn nx generate @nrwl/nest:library item --directory=api/feature --no-interactive --dry-run

CREATE libs/api/feature/item/README.md
CREATE libs/api/feature/item/src/index.ts
CREATE libs/api/feature/item/tsconfig.json
CREATE libs/api/feature/item/tsconfig.lib.json
CREATE libs/api/feature/item/project.json
UPDATE tsconfig.base.json
CREATE libs/api/feature/item/.eslintrc.json
CREATE libs/api/feature/item/jest.config.ts
CREATE libs/api/feature/item/tsconfig.spec.json
CREATE libs/api/feature/item/src/lib/api-feature-item.module.ts
```

However, the `module` file that is created contains the directory path name. But what if I just wanted it to be called `item.module.ts` rather than `api-feature-item.module.ts`?

It doesn't look like this is possible with the current implementation of the CLI.

## Motivation

Following this NX implementation [here](https://gist.github.com/trungk18/7ef8766cafc05bc8fd87be22de6c5b12), where every lib has a `feature`, `data-access`, `utils` sub-libs, it would be convenient to have shorter file names within these sub-libs. It makes so sense to have such long names IMO, since the directory paths are different, there would be no collisions.

It's not only the file names that are extra long but the class names within these files follow the same structure, only using camel case. (i.e., `class ApiFeatureItemController` for example)

## Suggested Implementation

Add an additional flag to the CLI, something like `--filename`:

```ts
yarn nx generate @nrwl/nest:library item --directory=api/feature --filename=item --no-interactive --dry-run
```

ashley-hunter:
## Current Behavior

Directory is always included in the module name.

## Expected Behavior

Allow a simple name, the same as we do in Angular.

## Related Issue(s)

Fixes #14654


jaysoo (approved):

jaysoo:
Thanks. I will rebase and merge.

ashley-hunter:
> Thanks. I will rebase and merge.

Thank you very much!

PR #16025, opened 2023-03-31
feat(js): adding simpleName option to library generator

### LINKED ISSUE #15564: Add simpleName option to @nrwl/js:lib generator
## Description
The @nrwl/angular:library generator has a [simpleName](https://nx.dev/packages/angular/generators/library#simplename) option that omits the directory from the generated files. It would be very nice to include this option in the @nrwl/js:lib generator as well. Using grouping folders as recommended [here](https://nx.dev/more-concepts/grouping-libraries) results in unwieldy file names otherwise.

## Motivation
Using grouping folders to organize libs becomes impractical without this option.

## Suggested Implementation
Analogous to [simpleName](https://nx.dev/packages/angular/generators/library#simplename)

## Alternate Implementations
-

ashley-hunter:
## Current Behavior

Directory name is always included in the generate file name

## Expected Behavior

Add option to have a simple file name

## Related Issue(s)

Fixes #15564

### DIFF
--- a/packages/js/src/generators/library/library.spec.ts
+++ b/packages/js/src/generators/library/library.spec.ts
@@ -183,7 +183,6 @@ describe('lib', () => {
           name: 'myLib2',
           directory: 'myDir',
           tags: 'one,two',
-          simpleModuleName: true,
         });
         projects = Object.fromEntries(getProjects(tree));
         expect(projects).toMatchObject({
@@ -1145,6 +1144,22 @@ describe('lib', () => {
 
       const project = readProjectConfiguration(tree, 'my-lib');
       expect(project.targets.build.options.assets).toEqual([]);
+    });
+  });
+
+  describe('--simpleName', () => {
+    it('should generate a simple name', async () => {
+      await libraryGenerator(tree, {
+        ...defaultOptions,
+        name: 'myLib',
+        simpleName: true,
+        directory: 'web',
+      });
+
+      expect(tree.read('libs/web/my-lib/src/index.ts', 'utf-8')).toContain(
+        `export * from './lib/my-lib';`
+      );
+      expect(tree.exists('libs/web/my-lib/src/lib/my-lib.ts')).toBeTruthy();
     });
   });
 });
--- a/packages/js/src/generators/library/library.ts
+++ b/packages/js/src/generators/library/library.ts
@@ -487,7 +487,7 @@ function normalizeOptions(
     ? name
     : projectDirectory.replace(new RegExp('/', 'g'), '-');
   const fileName = getCaseAwareFileName({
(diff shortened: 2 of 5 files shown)


AgentEnder:
Hey @ashley-hunter! I'm updating your branch to fix the issues with the docs. Thanks for the contribution 🎉

ashley-hunter:
> Hey @ashley-hunter! I'm updating your branch to fix the issues with the docs. Thanks for the contribution 🎉

Thank you very much!

packages/js/src/generators/library/library.spec.ts
```diff
@@ -1152,4 +1152,20 @@ describe('lib', () => {
       expect(project.targets.build.options.assets).toEqual([]);
     });
   });
+
+  describe('--simpleName', () => {
+    it('should generate a simple name', async () => {
```
AgentEnder, line 1157: This test isn't really checking that the `simpleName` flag did anything. It should pass a `directory` and then use `readProjectConfiguration('my-lib')` to ensure that the generated name doesn't contain the directory.

packages/js/src/generators/library/library.ts
```diff
@@ -451,7 +451,8 @@ function normalizeOptions(
 
   const projectName = projectDirectory.replace(new RegExp('/', 'g'), '-');
   const fileName = getCaseAwareFileName({
-    fileName: options.simpleModuleName ? name : projectName,
+    fileName:
+      options.simpleName || options.simpleModuleName ? name : projectName,
```
AgentEnder, line 455: We don't support simpleModuleName here (its not in the schema.json), so we shouldn't need to check it.

```suggestion
      options.simpleName ? name : projectName,
```

AgentEnder (changes requested):

packages/js/src/generators/library/library.ts
ashley-hunter: There is a `simpleModuleName` in the schema type, would you like it removed? I wasn't sure why it existed previously as its not in the schema.json

image

packages/js/src/generators/library/library.spec.ts
ashley-hunter: Thanks, oversight on my part I will correct that. 

I don't think I can use `readProjectConfiguration` to verify, as its the name of the file that is created that is simplified rather than the index file and it is not available in any of the options in the project configuration. I have tests covering that the file (with the simple name) exists and is exported correctly in the index file.

Let me know if there is any thing else you would like tested, thanks

jaysoo:
Will rebase and merge this as well.

jaysoo (approved):

AgentEnder (approved):

PR #16037, opened 2023-04-01
feat(core): allow local execution transpiler overriding with env var

yharaskrik:
## Current Behavior

Currently the transpiler being picked for local execution is entirely depending on if `@swc-node/register'` is installed or not. While this works for the overwhelmingly large majority of cases there are times when we may want to force the transpiling to use ts-node even though swc-node is installed. For example in our CI, we use CircleCI arm64 resources to run all of our build steps (as they end up cheaper than the x86/amd64 resources) except for when we dockerize our servers, we dockerize our servers on x86 Docker containers (since they are being deployed to x86 servers so we can't* build them on arm resources). This means that if we have `@swc-node/register` installed then when our dependencies are installed on arm resources we have arm swc native code compiled which doesn't run when we call local executions in our step that dockerizes (since its x86 and we do not have x86 swc compiled binaries).

* we COULD but I was not able to get building for a different architecture working with docker regardless of what I tried.

## Expected Behavior

This PR would allow us to set an env var to tell Nx to use `ts-node` even though swc-register is installed. This allows us to have local executions happening for all cases it can work, like dev'ing locally, or on most of our CI steps, but then set it to ts-node when we know we need it to fallback.

There may be another way to get around this I am missing so if there is I am more than happy to close this and do that!

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/reference/environment-variables.md
+++ b/docs/shared/reference/environment-variables.md
@@ -3,7 +3,7 @@
 The following environment variables are ones that you can set to change the behavior of Nx in different environments.
 
 | Property                         | Type    | Description                                                                                                                                                                                                                  |
-|----------------------------------| ------- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| -------------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


AgentEnder (changes requested):
This makes sense, I'd like to move forward with it.

Can you rename it and document it inside of `environment-variables.md`?

yharaskrik:
> This makes sense, I'd like to move forward with it.
> 
> Can you rename it and document it inside of `environment-variables.md`?

Done!

PR #16039, opened 2023-04-02
Update storybook-7-setup.md

JulianTurner:
Fixed wrong command

## Current Behavior
The Command is not working

## Expected Behavior
The Command works

## Related Issue(s)

Fixes # wrong example command

### DIFF
--- a/docs/generated/packages/storybook/documents/storybook-7-setup.md
+++ b/docs/generated/packages/storybook/documents/storybook-7-setup.md
@@ -53,7 +53,7 @@ npm install -D @nrwl/storybook
 You can generate Storybook configuration for an individual project with this command:
 
 ```shell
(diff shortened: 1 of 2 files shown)


mandarini (dismissed):

mandarini:
Hi @JulianTurner ! Thanks for your PR. It had 2 issues and I fixed them. For your reference, and if you need to make a contribution in the future, these are the issues that I fixed:

1. You forgot to run `yarn documentation`, so the docs check failed. `yarn documentation` makes changes to the generated files for our docs.
2. Your commit did not match our commit guidelines, which are `type(scope): subject`. You can run `yarn commit` to get help when writing the commit message (that's what I do at least)

These two issues probably mean that you pushed with `--no-verify`, because our pre-commit hooks catch both of these issues, and they also run `yarn documentation` for you, and ask you to add and commit the changes.

So, please keep in mind these two things in the future! In any case, THANK YOU for contributing to Nx! :D

PR #16045, opened 2023-04-02
feat(vite): add generatePackageJson option to vite executor (#16042)

### LINKED ISSUE #16042: @nrwl/vite:build add generatePackageJson option

## Description

Other executors (esbuild/webpack/swc) have the ability to generate a package JSON file for the dependencies used.  Since we can also use vite to build libraries, or even use something like [vavite](https://github.com/cyco130/vavite) to build non-frontend code, it would be useful to have this available as an option.

## Motivation

Its primary use case would be for using the vite executor to build libraries. Currently you need to manually build the package.json and copy it over as an asset in order to provide them.  

## Suggested Implementation

Reuse most of existing implementation of esbuild `generatePackageJson` feature to add similar functionality to vite executor.

## Alternate Implementations

TBD

jensbodal:
* moves getExtraDependencies helper from esbuild package to js package

## Current Behavior

No ability to generate package.json when building with vite executor 

## Expected Behavior

Allow ability to generate package.json using vite executor

## Related Issue(s)

* https://github.com/nrwl/nx/issues/16042

## Example

```
    "build": {
      "executor": "@nrwl/vite:build",
      "outputs": ["{options.outputPath}"],
      "defaultConfiguration": "production",
      "options": {
        "outputPath": "dist/apps/api",
        "ssr": true
      },
      "configurations": {
        "development": {
          "mode": "development"
        },
        "production": {
          "generatePackageJson": true,
          "main": "apps/api/src/main.ts",
          "tsConfig": "apps/api/tsconfig.app.json",
          "mode": "production"
        }
      }
    },
```


jensbodal:
I'm not sure why CircleCI Pipeline failed to run

jensbodal:
> but the approach is wrong because we're pulling in other options that don't affect the build (besides package.json generation).

I just want to call out that that's what the esbuild executor does for two of the options.  I had wanted to just put these options in a nested object that were specific to `generatePackageJson` but I didn't because I didn't want to diverge from existing implementations, and they _might_ have some other purpose for the general build later and it would be messy to later require moving `main` and `tsconfig` to the root build options.

## esbuild executor

* [`main`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/schema.json#L9)
* [`tsConfig`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/schema.json#L27)
* [`format`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/schema.json#L47)
* [`external`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/schema.json#L76)

The above are used for both `generatePackageJson` and `esbuild.build` ([esbuild.build options](https://github.com/evanw/esbuild/blob/main/lib/shared/types.ts))

* [`thirdParty`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/schema.json#L129)
  - [if not set](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/esbuild.impl.ts#L62) determines if [`updateBuildableProjectDepsInPackageJson`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/esbuild.impl.ts#L81) flag is set when calling [`updatePackageJson`](https://github.com/nrwl/nx/blob/master/packages/js/src/utils/package-json/update-package-json.ts#L58)
  - sets [`excludeLibsInPackageJson`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/esbuild.impl.ts#L80) which is used for generating the package.json
  - determines additional logic for [generating the package.json](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/esbuild.impl.ts#L85)
* [`dependenciesFieldType `](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/schema.json#L134)
  - [not currently in use](https://github.com/search?q=org%3Anrwl+dependenciesFieldType&type=code) but I would imagine this is only related to the generate of the packageJson

---

If preferred I could just remove the allowed overrides that I provided and just set them statically instead of being defaults 

```
      buildableProjectDepsInPackageJsonType = 'dependencies',
      excludeLibsInPackageJson = true,
      format = 'esm',
      generateLockfile = true,
```

This would just leave setting `main`, `tsconfig`, and `generatePackageJson`.  Possibly there's a way to pull `main` and `tsconfig` from the vite configuration.

I'd like to work with you to make whatever changes in order to get this in since it's something I'd very much like to use.

jensbodal:
I can remove the `tsconfig` option and just generate that from `${buildConfig.root}/tsconfig.json`

`main` is a little trickier since it's really only used to make the generated `package.json` "valid".  I could either 

1. default this to `./src/index.js` and someone could override the setting by having an existing `package.json` in their directory with the `main` field populated
2. make `generatePackageJson` a `string` option and provide it there
3. make `generatePackageJson` an `object` and require setting the value (and other overrides there)
4. require having a `package.json` with the value set to use `generatePackageJson`

jensbodal:
I think I have a better idea now than continuously adding this feature to different executors: https://github.com/nrwl/nx/issues/16042#button-description-0-0

PR #16069, opened 2023-04-03
Switch packages to use the @nx scope

FrozenPandaz:
# Switch packages to use the @nx scope

Nx's first party plugins and other packages will be published under `@nx` starting in v16. `@nrwl/*` packages will no longer be published in Nx 17.

This PR does the following:
* The `@nrwl/` packages are now published under the `@nx` scope.
  * `@nrwl/js` => `@nx/js`
  * `@nrwl/react` => `@nx/react`
  * `@nrwl/angular` => `@nx/angular`
  * etc.
  * The one exception is `@nrwl/tao` which does not need to be changed to `@nx/tao` but needs to collect consistent statistics.
* `@nrwl/*` packages are still published but they are shells of their `@nx/` counterparts
  * Any public apis re-export from their `@nx` counterparts
  * Any generators will still be available (`@nrwl/react:library`)
  * Any executors will still be available (`@nrwl/webpack:webpack`)
  * These packages have a new `README.md` which says that the package has been deprecated and to use the `@nx/*` counterpart
  * Everything that can be done today with the `@nrwl/*` packages should be able to be done the same way in Nx 16
  * In Nx 17, `@nrwl/*` packages will cease to be published 
* The `@nx` packages have been added to the package group
* The `@nx` packages have replaced the `@nrwl` packages in the First party plugins for `nx list`
* This PR changes any **imports** have been modified to `@nx/*` imports
  * For now in this PR, other usages of other packages remain `@nrwl/`
    * For example, `create-nx-workspace` will still install `@nrwl/`
    * I used this to ensure that the `@nrwl/*` packages are working properly.
    * Before Nx 16, is released, these usages will change to `@nx`
    * This will be the next phase of this effort before Nx 16 is released.
      * Those PRs will be more granular :sweat_smile:
* Docs have not changed yet in this PR
  * When Nx 16 has been released, we will update the docs
  * There is a temporary change to the docs generation to generate the docs for `@nx/*` packages as `@nrwl/*` docs

Misc changes that were necessary to make this possible:
* `@nx/angular` is now built with `@nx/angular:package` instead of `@angular-devkit/build-angular:ng-packagr`
* You can reference other packages for the implementation or schema of an executor
  * `@nrwl/jest:jest` references `@nx/jest:jest` for its implementation
  * Note: We did not want to allow extending a collection of `executors` like you can `generators` because it makes less sense for `executors`. 

## Current Behavior

Packages are published under the `@nrwl` scope.

## Expected Behavior

Packages are published under the `@nx` scope.

We will still publish `@nrwl/` packages for v17 but they will be wrappers of their `@nx` counterparts.

We plan to stop publishing `@nrwl/` packages starting in v18.

There is a migration to help people automatically move to the `@nx` packages.

## Related Issue(s)

Fixes #


packages/cypress/src/generators/cypress-e2e-configuration/files/v9/__dir__/plugins/index.js
```diff
@@ -11,7 +11,7 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
-const { preprocessTypescript } = require('@nrwl/cypress/plugins/preprocessor');
+const { preprocessTypescript } = require('@nx/cypress/plugins/preprocessor');
```
barbados-clemens, line 14: this file should just be removed since we dropped the preprocessor. I can do this in a separate PR to following this one up.

packages/cypress/src/generators/cypress-project/files/v9-and-under/src/plugins/index.js
```diff
@@ -11,7 +11,7 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
-const { preprocessTypescript } = require('@nrwl/cypress/plugins/preprocessor');
+const { preprocessTypescript } = require('@nx/cypress/plugins/preprocessor');
```
barbados-clemens, line 14: as mentioned above, not needed since we removed the preprocessor, but I can follow up a PR to remove it to not clutter up this PR.

scripts/patched-jest-resolver.js
```diff
@@ -86,9 +86,9 @@ module.exports = function (path, options) {
 
     // TODO(v17): Remove this workaround
     // We have some weird d.ts + .js business going on for these 2 imports so this is a workaround
-    if (path === '@nrwl/devkit') {
+    if (path === '@nx/devkit') {
```
barbados-clemens, line 89: on line 65 you'll probably want to switch from '@nrwl/' to '@nx/' to match.

barbados-clemens (approved):
just a couple comments to consider.

leosvelperez (approved):

packages/angular/migrations.spec.ts
```diff
@@ -3,9 +3,11 @@ import json = require('./migrations.json');
 
 describe('Angular migrations', () => {
   it('should have valid paths', () => {
-    Object.values(json.schematics).forEach((m) => {
+    Object.values(json.schematics).forEach((m: any) => {
```
Coly010, line 6: A heads up more than anything. There's a PR that changes this json to `generators` instead of `schematics`.

https://github.com/nrwl/nx/pull/16291

packages/angular/src/migrations/update-16-0-0-add-nx-packages/update-16-0-0-add-nx-packages.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
+import replacePackage from './update-16-0-0-add-nx-packages';
+
+describe('update-16-0-0-add-nx-packages', () => {
+  let tree: Tree;
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+
+  it('should remove the dependency on @nrwl/angular', async () => {
```
Coly010, line 11: Should this also check that the nx package has been added correctly?

Coly010 (approved):

packages/devkit/src/utils/replace-package.ts
```diff
@@ -0,0 +1,166 @@
...
+        packageJson.peerDependencies ?? {},
+        packageJson.optionalDependencies ?? {},
+      ]) {
+        if (oldPackageName in deps) {
+          deps[newPackageName] = NX_VERSION;
+          delete deps[oldPackageName];
+        }
+      }
+      return packageJson;
+    });
```
meeroslav, line 53: Shouldn't this be similarly to others be marked with `needsUpdate` and if `true` run `installPackagesTask` at the end?

Or is this only meant to be run within the migration (which prompts for `install`)?

meeroslav (approved):
Looks good to me. I left one question.

packages/devkit/src/utils/replace-package.ts
FrozenPandaz: It's only used for migrations for now... I could make it return the task though.

nartc (approved):

xiongemi (approved):

philipjfulcher (approved):

PR #16078, opened 2023-04-04
cleanup(angular): update component generator to be more consistent with similar generators

leosvelperez:
## Current Behavior

The `component` generator implementation is not fully aligned with similar generators' implementation (e.g. `directive`).

## Expected Behavior

The `component` generator implementation is aligned as much as possible with similar generators' implementation. Some common functionality is reused for better consistency.

## Related Issue(s)

Fixes #


packages/angular/src/generators/component/lib/normalize-options.ts
```diff
@@ -1,29 +1,33 @@
...
     options.project
-  );
-  const projectSourceRoot = sourceRoot ?? joinPathFragments(root, 'src');
+  ) as ProjectConfiguration & { prefix?: string };
 
-  const parsedName = options.name.split('/');
-  const name = parsedName.pop();
-  const namedPath = parsedName.join('/');
+  const projectSourceRoot = sourceRoot ?? joinPathFragments(root, 'src');
+  const { name, path: namePath } = parseName(options.name);
```
Coly010, line 17: Sorry, I know this function was added in a previous PR, but looking back it here, what do you think of renaming it to this?

It can be done in a follow-up PR.

```suggestion
  const { name, path: namePath } = parseNameWithPath(options.name);
```

Coly010 (approved):
This looks so good now after the changes, awesome! 🎉

Left one minor comment

PR #16084, opened 2023-04-04
feat(core): deprecate implicitDependency in nx.json

meeroslav:
## Related Issue(s)

Fixes #


AgentEnder (dismissed):

docs/shared/deprecated/global-implicit-dependencies.md
```diff
@@ -1,6 +1,6 @@
 # Global Implicit Dependencies
 
-As of Nx 14.4, it is better to use [`inputs` and `namedInputs`](/more-concepts/customizing-inputs) instead of the `implicitDependencies` previously defined in `nx.json`. [`implicitDependencies` in the project configuration](/reference/project-configuration#implicitdependencies) are still the best way to manually set up a dependency between two projects that Nx is not able to detect automatically.
+Since v14.4 Nx supports [`inputs` and `namedInputs`](/more-concepts/customizing-inputs) for setting up implicit dependencies. As of Nx v16, the `implicitDependencies` defined in `nx.json` are deprecated and do not influence the affected graph. This field will be removed in v17. The [`implicitDependencies` in the project configuration](/reference/project-configuration#implicitdependencies) are still the best way to manually set up a dependency between two projects that Nx is not able to detect automatically.
```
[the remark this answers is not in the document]
meeroslav: I get what you are saying. 

It was de facto deprecated since v15, and there is an automated migration from the `implicitDependencies` to `namedInputs`. Unfortunately, we did not mark it as `deprecated` so we cannot fully remove it just yet.

Right now if you try using it it will give you a warning:
```
Using "implicitDependencies" is deprecated. Use "namedInputs" instead.
For more information about the usage of "namedInputs" see https://nx.dev/deprecated/global-implicit-dependencies#global-implicit-dependencies'
```

Perhaps changing this text would be better? Something along:
```
Using `implicitDependencies` for global implicit dependencies configuration is no longer supported. 
Use `namedInputs` instead.
For more information about the usage of "namedInputs" see https://nx.dev/deprecated/global-implicit-dependencies#global-implicit-dependencies'
```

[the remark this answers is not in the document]
AgentEnder: Yeah, I like that better if we are removing it. I just don't want to tell folks its deprecated and then it not work, since people are used to deprecated things still currently working.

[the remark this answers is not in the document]
meeroslav: Ok, the wording has been changed so please have a look when you have time.

packages/nx/src/utils/assert-workspace-validity.ts
```diff
@@ -19,44 +16,16 @@ export function assertWorkspaceValidity(
 
   const invalidImplicitDependencies = new Map<string, string[]>();
```
FrozenPandaz, line 17: We don't need this map right?

packages/nx/src/utils/assert-workspace-validity.ts
```diff
@@ -19,44 +16,16 @@ export function assertWorkspaceValidity(
...
-      );
-      return map;
-    }, invalidImplicitDependencies);
+  if (nxJson.implicitDependencies) {
+    output.warn({
+      title:
+        'Using `implicitDependencies` for global implicit dependencies configuration is no longer supported.',
+      bodyLines: [
+        'Use "namedInputs" instead.',
+        'For more information about the usage of "namedInputs" see https://nx.dev/deprecated/global-implicit-dependencies#global-implicit-dependencies',
```
FrozenPandaz, line 25: Can you include a line to use `nx repair`? But I noticed that `nx repair` will mess up the `production` fileset. Can you fix that as well please?

FrozenPandaz (changes requested):

packages/nx/src/utils/assert-workspace-validity.ts
meeroslav: We do need it for the project level `implicitDependencies` below (line 59).

packages/nx/src/utils/assert-workspace-validity.ts
meeroslav: Sure, fix is in a separate PR: https://github.com/nrwl/nx/pull/16145

FrozenPandaz (approved):

PR #16098, opened 2023-04-04
feat(nx-dev): add markdoc title card component

bcabanes:


nx-dev/ui-markdoc/src/lib/tags/cards.component.tsx
```diff
@@ -79,3 +79,17 @@ export function Card({
     </div>
   );
 }
+
+export function TitleCard({ title, url }: { title: string; url: string }) {
+  return (
+    <a
+      key={title}
+      href={'blahblah'}
```
juristr, line 87: Not the "blahblah" was intentional

juristr (changes requested):
I think we should make it s.t. the cards expand on the entire width.

Right now they look as follows:
image

I'd rather prefer something like this

image

bcabanes:
@juristr They take all the space by default, it respects the number of columns you put in `{% cards cols="x" %}`.

juristr (approved):

isaacplmann (approved):

PR #16112, opened 2023-04-05
cleanup(core): deprecate and remove Rules from workspace

Coly010:
## Current Behavior

Some functions are marked as to be removed in v16.
Some functions have no usages and are not exported from public api.
Some functions are based on Angular Schematic usage which will no longer be supported.

## Expected Behavior

Remove what can be removed, deprecate the others for removal in v17.

## Related Issue(s)

Fixes #


packages/workspace/src/utils/testing.ts
FrozenPandaz, line 1: followup: I think these are purely internal. We can remove them if they aren't being used by us.

FrozenPandaz (approved):

Issue #15830, opened 2023-03-22
Next 13 apps directory 'use client' directive is used on all exports of a library


merijnhofsteenge:
### Current Behavior

Since nx 15.8.7 it is possible to use the 'use client' directive on nx libraries. It seems however that when an import is made from a library, all the other components exported from the same library get explicitly checked for whether they are client or server components. Some examples of this behaviour:

1. A library contains two files, each containing one client component. They are both exported. One component is used by the next app, and the other component is unused. The app only builds if the unused component is marked with 'use client'.
2. A library contains two files, one containing a server component and one containing a client component. The client component gets used in a client component in a different library. This is not possible, since next now sees the server component as being marked with 'use client', even though this is not explicitly done.

### Expected Behavior

1. It should not matter if the component has the 'use client' directive or not, since it is not used.
2. Client and server components should be usable in the same library, even when some client components are used by different client libraries.

### GitHub Repo

https://github.com/merijnhofsteenge/nx-next-13-use-client

### Steps to Reproduce

Follow the readme of the repo

### Nx Report

```shell
Node : 18.12.0
   OS   : darwin arm64
   npm  : 8.19.2
   
   nx                      : 15.8.7
   @nrwl/js                : 15.8.7
   @nrwl/jest              : 15.8.7
   @nrwl/linter            : 15.8.7
   @nrwl/workspace         : 15.8.7
   @nrwl/cli               : 15.8.7
   @nrwl/cypress           : 15.8.7
   @nrwl/devkit            : 15.8.7
   @nrwl/eslint-plugin-nx  : 15.8.7
   @nrwl/next              : 15.8.7
   @nrwl/react             : 15.8.7
   @nrwl/tao               : 15.8.7
   @nrwl/webpack           : 15.8.7
   @nrwl/nx-cloud          : 15.2.4
   typescript              : 4.9.5
```

ndcunningham:
This should be fixed with `15.9.0` which includes https://github.com/nrwl/nx/pull/15650
Try upgrading your Nx version and see if the issue is fixed.

jaysoo:
@merijnhofsteenge there is an issue with how Next.js is detecting server vs client components. One workaround for now is to do deep imports from a client component to another client component.

For example, if you update `tsconfig.base.json` with the following entry:

```
{
  ...
  "compilerOptions": {
    ...
    "paths": {
      "@nx-next-13-use-client/client-lib": [
        "libs/client-lib/src/index.ts"
      ],
[...]

PR #16130, opened 2023-04-05
feat(nextjs): update to Next.js 13.3.0

jaysoo:
This PR updates the installed version of Next.js to 13.3.0 ~~and migrates existing workspaces~~.

We will hold off on migrating existing workspaces to 13.3 until we can put together a better migration experience for those using Storybook (they need to go to v7).

Also includes a few fixes. See below.

## ~~Use Storybook v7 for React/Next.js~~

~~Storybook v7 is stable now, and v6 is no longer maintained. Next.js 13.2 requires v7 to work, so it will be the default moving forward unless the user passes `--storybook7Configuration=false` when generating Storybook config (i.e. `nx g @nrwl/react:storybook-configuration --storybook7Configuration=false`).~~

> Note: Another PR just landed to update Storybook to v7 for new projects. No changes needed here anymore.

## Disable type checks by default (app/layout types are incorrect)

This only applies when appDir is enabled in `next.config.js`.

The Next.js type generation has a bug where the paths are wrong inside a monorepo. The PR is merged but waiting for a release (https://github.com/vercel/next.js/pull/47534).

Once released we can re-enable the type-check by default. In the meantime the user can update `next.config.js` config to enable it if the patch is released before we update.

## Generate libraries with separate RSC entry

Add a server-only entry for libraries generated with `@nrwl/next:lib` to help with libraries that export both client and server components. So users can import client-only components using deep import path. See: https://github.com/nrwl/nx/issues/15830

e.g. 

Say our library `my-lib` exported both RSC and client components in `index.ts`.

```ts
// my-lib/src/index.ts
export { MyClientOnlyComponent } from './my-client-only-component';
export { MyServerOnlyComponent } from './my-server-only-component';
```

Then importing this in a client component will error out.

**Fails:**

```ts
'use client';
import { MyClientOnlyComponent } from '@acme/my-lib';

export function ParentComponent() {
  return <MyClientOnlyComponent />;
}
```

But if we separate the two components into separate entry files.

```ts
// my-lib/src/index.ts
export { MyClientOnlyComponent } from './my-client-only-component';

// my-lib/src/server.ts
export { MyServerOnlyComponent } from './my-server-only-component';
```

Then importing from client entry file will work in both server _and_ client components.

**Works:**
```ts
'use client';
import { MyClientOnlyComponent } from '@acme/my-lib';

export function ParentComponent() {
  // Works because the main entry no longer has server-only components
  return <MyClientOnlyComponent />;
}
```

**Also works:**
```ts
import { MyClientOnlyComponent } from '@acme/my-lib';
import { MyServerOnlyComponent } from '@acme/my-lib/server';

export async function SomePage() {
  // Can render client component from server component
  return (
    <MyClientOnlyComponent/>
    <MyServerOnlyComponent/>
  );
}
```

This isn't an Nx issue because it's coming from `@next/swc`:
- How imports are collected:  https://github.com/vercel/next.js/blob/2820f0787512b6514e36c02cbddd4d3ab69f0251/packages/next-swc/crates/core/src/react_server_components.rs#L64
- The check that fails due to forbidden imports: https://github.com/vercel/next.js/blob/2820f0787512b6514e36c02cbddd4d3ab69f0251/packages/next-swc/crates/core/src/react_server_components.rs#L384

### DIFF
--- a/e2e/workspace-create-npm/src/create-nx-workspace-npm.test.ts
+++ b/e2e/workspace-create-npm/src/create-nx-workspace-npm.test.ts
@@ -136,6 +136,7 @@ describe('create-nx-workspace --preset=npm', () => {
     const tsconfig = readJson(`tsconfig.base.json`);
     expect(tsconfig.compilerOptions.paths).toEqual({
       [libName]: [`packages/${libName}/src/index.ts`],
+      [`${libName}/server`]: [`packages/${libName}/src/server.ts`],
     });
   });
 
--- a/packages/js/src/utils/typescript/ts-config.ts
+++ b/packages/js/src/utils/typescript/ts-config.ts
@@ -52,38 +52,22 @@ export function getRootTsConfigFileName(tree: Tree): string | null {
   return null;
 }
 
-export function updateRootTsConfig(
-  host: Tree,
-  options: {
-    name: string;
-    importPath?: string;
-    projectRoot: string;
-    js?: boolean;
-  }
+export function addTsConfigPath(
+  tree: Tree,
+  importPath: string,
+  lookupPaths: string[]
 ) {
-  if (!options.importPath) {
-    throw new Error(
-      `Unable to update ${options.name} using the import path "${options.importPath}". Make sure to specify a valid import path one.`
-    );
-  }
-  updateJson(host, getRootTsConfigPathInTree(host), (json) => {
+  updateJson(tree, getRootTsConfigPathInTree(tree), (json) => {
     const c = json.compilerOptions;
-    c.paths = c.paths || {};
-    delete c.paths[options.name];
+    c.paths ??= {};
 
-    if (c.paths[options.importPath]) {
+    if (c.paths[importPath]) {
       throw new Error(
-        `You already have a library using the import path "${options.importPath}". Make sure to specify a unique one.`
+        `You already have a library using the import path "${importPath}". Make sure to specify a unique one.`
       );
     }
 
-    c.paths[options.importPath] = [
-      joinPathFragments(
-        options.projectRoot,
-        './src',
-        'index.' + (options.js ? 'js' : 'ts')
-      ),
-    ];
+    c.paths[importPath] = lookupPaths;
 
     return json;
   });
--- a/packages/next/src/generators/library/library.ts
+++ b/packages/next/src/generators/library/library.ts
@@ -2,6 +2,7 @@ import {
   convertNxGenerator,
   formatFiles,
   GeneratorCallback,
+  getImportPath,
   getWorkspaceLayout,
   joinPathFragments,
   names,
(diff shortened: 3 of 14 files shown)


ndcunningham (approved):
LGTM

barbados-clemens (approved):

mandarini (approved):

meeroslav (approved):

xiongemi (approved):

e2e/workspace-create-npm/src/create-nx-workspace-npm.test.ts
```diff
@@ -152,6 +152,7 @@ describe('create-nx-workspace --preset=npm', () => {
     const tsconfig = readJson(`tsconfig.base.json`);
     expect(tsconfig.compilerOptions.paths).toEqual({
       [libName]: [`packages/${libName}/src/index.ts`],
+      [`${libName}/*`]: [`packages/${libName}/*`],
```
jaysoo, line 155: export `server.ts` for server-only components.

packages/react/src/generators/library/lib/create-files.ts
```diff
@@ -47,6 +47,19 @@ export function createFiles(host: Tree, options: NormalizedSchema) {
     host.delete(`${options.projectRoot}/package.json`);
   }
 
+  if (options.additionalImportPaths) {
+    for (const additionalImportPath of options.additionalImportPaths) {
```
jaysoo, line 51: This is just creating an empty file for additional import paths (e.g. `src/server.ts` for RSC). Other generators using this generator should fill them in with more useful content.

packages/js/src/utils/typescript/ts-config.ts
```diff
@@ -85,6 +88,17 @@ export function updateRootTsConfig(
...
+      for (const additionalImportPath of options.additionalImportPaths) {
+        c.paths[`${options.importPath}/${additionalImportPath}`] = [
+          joinPathFragments(
+            options.projectRoot,
+            `./src/${additionalImportPath}.${options.js ? 'js' : 'ts'}`
+          ),
+        ];
+      }
+    }
+
```
FrozenPandaz, line 101: I don't know if I'm a huge fan of having this here..

Not a big deal probably... but this currently doesn't check if those paths already exist and would just overwrite them :shrug:. 

I feel like we should rework this function's API to be more like `addTsConfigImportPaths(tree, importPath: string, importModules: string[])`

```ts
function jsLibraryGenerator(tree) {
  addTsConfigPath(tree, importPath, [joinPathFragments(projectRoot, 'src/index.ts')])
  addTsConfigPath(tree, joinPathFragments(importPath, 'server'), [joinPathFragments(projectRoot, 'src/server.ts')])
}
```

This would open us up to adding import paths which are not under `src`.

We could add the new function and keep the current usages the same.. then transition them over. What do you think?

packages/next/src/generators/library/library.ts
```diff
@@ -33,9 +33,28 @@ export async function libraryGenerator(host: Tree, options: Schema) {
     ...options,
     compiler: 'swc',
     skipFormat: true,
+    // Always include deep import paths so users can mix server and client react components in the same library.
+    // See: https://github.com/nrwl/nx/issues/15830
+    additionalImportPaths: ['server'],
```
FrozenPandaz, line 38: I think entry points work better? It's like multiple entry points into this project.

```suggestion
    additionalEntryPoints: ['server'],
```

I kind of see why it's in the `react:library` since I guess it enables RSC... why not just do everything in here for now though?

I don't like how a blank file is created in the react library.. and then this has to write the same exact file.

FrozenPandaz (changes requested):

packages/next/src/generators/library/library.ts
jaysoo: I think it's too hard to know what should be in the file since it's very Next.js specific. The React generator can only generate blank files as to not assume what's in there. When RSC is more stable there might be some consensus or standard on how libraries should be built.

packages/js/src/utils/typescript/ts-config.ts
jaysoo: Yeah we can do something like that.

packages/next/src/generators/library/library.ts
jaysoo: Updated it so this option is not added anymore. Instead, the Next.js library generator will add the file + import path that is specific to Next.js apps.

packages/js/src/utils/typescript/ts-config.ts
jaysoo: Removed the original function and added `addTsConfigPath` so each library generator can add the import paths that are needed. For now, the new `server.ts` entry is only for Next.js apps, and we can look at adding something similar to `@nrwl/react:lib` later when it's more clear the direction of RSCs.

FrozenPandaz (approved):
Sweet!

leosvelperez (approved):

PR #16132, opened 2023-04-05
feat(nextjs): add support for experimental appDir

ndcunningham:
This adds support for `--appDir` when creating a new nextjs project using nx

`npx nx generate @nrwl/next:application my-app --appDir` 

Should populate your workspace with something similar to this:
```tree
apps/my-app
├── app
├── index.d.ts
├── jest.config.ts
├── next-env.d.ts
├── next.config.js
├── project.json
├── public
├── specs
├── tsconfig.json
└── tsconfig.spec.json
```


packages/next/src/generators/application/files/app/page.tsx__tmpl__
```diff
@@ -0,0 +1,28 @@
...
+%>
+  <%- style !== 'styled-jsx' ? `import styles from './page.module.${style}';` : '' %>
+<% }
+%>
+
+<% if (styledModule && styledModule !== 'styled-jsx') { %>
+const StyledPage = styled.div`<%- pageStyleContent %>`;
+<% }%>
+
+export function Index() {
```
jaysoo, line 14: Let's make it an async function since it is a server component. Users can do any async data fetch needed within the function.

jaysoo (approved):

PR #16198, opened 2023-04-10
chore(vite): update vite version

mandarini:
Make sure to set the latest version of `vite` on `package.json`.


barbados-clemens (approved):
Does this also need bump vitest versions?

mandarini:
> Does this also need bump vitest versions?

I don't think we need to, but I could add it sure!

barbados-clemens:
> > Does this also need bump vitest versions?
> 
> I don't think we need to, but I could add it sure!

No worries, just wanted to make sure they weren't gonna get out of sync but looks like vitest dep is set to any minor so should be fine. 👍

packages/vite/src/executors/test/vitest.impl.ts
```diff
@@ -93,14 +93,15 @@ async function getSettings(
 ) {
   const projectRoot = context.projectGraph.nodes[context.projectName].data.root;
   const offset = relative(workspaceRoot, context.cwd);
-  // if reportsDirectory is not provides vitest will remove all files in the project root
+  // if reportsDirectory is not provided vitest will remove all files in the project root
```
lukascivil, line 96: The comma is missing, IMO.

`
// if reportsDirectory is not provided, vitest will remove all files in the project root
`

pumano:
looks like PR is too old, 4.3.3 released also @angular-devkit/build-angular 16.0.0-rc2 depends indirectly from vite 4.3.1

barbados-clemens (approved):

FrozenPandaz (approved):

jaysoo (approved):

PR #16245, opened 2023-04-12
docs(core): improve nx init description

leosvelperez:
## Current Behavior

The `nx init` command docs page doesn't have much info.

## Expected Behavior

The `nx init` command docs page is more descriptive.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/nx-commands.ts
+++ b/packages/nx/src/command-line/nx-commands.ts
@@ -303,12 +303,8 @@ export const commandsObject = yargs
   })
   .command({
     command: 'init',
-    describe: `Adds Nx to any type of workspace. It performs the following changes:
-
-- Installs nx
-- Creates an nx.json configuration file
-- Sets up distributed caching (optional)
-- If migrating an Angular CLI, CRA or NestJS workspace, it installs the relevant Nx plugins for better developer experience`,
+    describe:
+      'Adds Nx to any type of workspace. It installs nx, creates an nx.json configuration file and optionally sets up distributed caching. For more info, check https://nx.dev/recipes/adopting-nx.',
     builder: (yargs) => withIntegratedOption(yargs),
     handler: async (args: any) => {
       await (await import('./init')).initHandler(args);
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/nx-commands.ts
```diff
@@ -303,7 +303,12 @@ export const commandsObject = yargs
   })
   .command({
     command: 'init',
-    describe: 'Adds nx.json file and installs nx if not installed already',
+    describe: `Adds Nx to any type of workspace. It performs the following changes:
+
+- Installs nx
+- Creates an nx.json configuration file
+- Sets up distributed caching (optional)
+- If migrating an Angular CLI, CRA or NestJS workspace, it installs the relevant Nx plugins for better developer experience`,
```
leosvelperez, line 311: @juristr @isaacplmann I'd welcome any suggestion to make this better.

I thought about including links to the individual pages migrating different types of workspaces, but I wasn't sure if we wanted to do that given this is also shown in the terminal help. I didn't see any other command with links, so I wasn't sure if that was something we were avoiding or if it was that other commands don't need any links.

bcabanes (changes requested):

packages/nx/src/command-line/nx-commands.ts
isaacplmann: There are a few other terminal commands that use links, but I would only use one link and shorten the description.

leosvelperez: @isaacplmann I don't think links are rendered in the description. I made the change and added a link but it's not rendered as a link. It's just plain text. Which other commands use links in their description? I couldn't find one.

isaacplmann: Maybe I'm thinking of create-nx-workspace: packages/create-nx-workspace/src/utils/preset/point-to-tutorial-and-course.ts

leosvelperez: Yeah, that only appears in the terminal. @bcabanes is there a way to get a link rendered in the command docs page description?

bcabanes (approved):

packages/nx/src/command-line/nx-commands.ts
leosvelperez: After talking to Ben about it, we'll keep the link as plain text until the doc generation is reworked to support such scenarios.

AgentEnder (approved):

PR #16281, opened 2023-04-13
fix(documentation): generated package.json does not currently match the documentation. 

MatthewJamesBoyle:
## Current Behavior
Whilst running through the example, I experienced the error:
```
 >  NX   Cannot find project 'blog'
```

This was because the generated project only had a single * in packages.workspace. This small amendment got me working again.

## Expected Behavior
The documentation matches the CLI output.

### DIFF
--- a/docs/shared/core-tutorial/01-create-blog.md
+++ b/docs/shared/core-tutorial/01-create-blog.md
@@ -48,6 +48,13 @@ The `package.json` file contains this property:
 
 ```json
   "workspaces": [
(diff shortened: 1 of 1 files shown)


isaacplmann:
I just tried this and you don't need to have two asterisks for Nx to pick up the blog project.  It could be that you were running into a different bug.

We also are no longer linking to this tutorial.  So people will only get to it if they actively search for it.

I'll close this for now.

PR #16292, opened 2023-04-13
feat(next): support nextjs cypress component testing

### LINKED ISSUE #14363: Support Cypress Component Testing with Next.js

## Description
Cypress Component Testing seems like it would be a very useful workflow, but unfortunately Nx doesn't seem to support it for Next.js projects.

Running a command like: 
```
nx g @nrwl/react:cypress-component-configuration --project=my-project --build-target=web:build
```
_(by the way, the [official documentation](https://nx.dev/packages/react/generators/cypress-component-configuration) states to run `react:cypress-component-project` instead of `-configuration`, which is incorrect)_

results in this errors:

```
 ERROR  The provided build target, web:build, uses the '@nrwl/next:build' executor.
But only the follow executors are allowed
- @nrwl/webpack:webpack
```

## Motivation

I was pretty excited to use Cypress Component Testing to have a visual debugging environment for UI component tests, and Nx seemed to support it, but I was frustrated to discover the support was only for vanilla React projects, not Next.js.

## Suggested Implementation

Since Nx already supports Cypress Component Testing and Cypress Component already [supports Next.js](https://docs.cypress.io/guides/component-testing/overview#Supported-Frameworks), it seems like there's only a small missing piece required to allow it to use a Next build executor instead of Webpack.

## Alternate Implementations

If it's already possible to do this and I just haven't figured it out, extending the documentation to explain what to do for this use case would be extremely helpful.

barbados-clemens:
## Current Behavior

No preset for running CT with NextJS

## Expected Behavior

Nx + NextJS CT support

## Related Issue(s)

Fixes #14363


mandarini (approved):

packages/react/src/generators/cypress-component-configuration/cypress-component-configuration.ts
```diff
@@ -25,8 +25,19 @@ export async function cypressComponentConfigGenerator(
...
+      '@nx/webpack:webpack',
+      '@nx/vite:build',
+      '@nrwl/webpack:webpack',
+      '@nrwl/vite:build',
+    ]),
+  });
+
   await addFiles(tree, projectConfig, options, found);
+
   if (options.skipFormat) {
```
FrozenPandaz, line 41:
```suggestion
  if (!options.skipFormat) {
```

FrozenPandaz (approved):

jaysoo (approved):

PR #16295, opened 2023-04-13
chore(core): nx plugin submission nx-size-limit

LironHazan:
Community Plugin Submission
nx-size-limit

Features
>Size Limit is a performance budget tool for JavaScript. It checks every commit on CI, calculates the real cost of your JS for end-users and throws an error if the cost exceeds the limit.

A .size-limit.json config generator (nx-size-limit:add) which allows setting any size-limit configuration.
An executor (nx-size-limit:size-limit)for running the limitation check
An example of how to use in an NX CI (TBD a nicer gha)


LironHazan:
will open a new PR using yarn commit.

PR #16299, opened 2023-04-13
feat(angular): custom mf config path and ts support

### LINKED ISSUE #15739: Implement custom module federation config path in angular/src/builders/utilities/module-federation.ts

## Description

Currently, the Nx Angular Module Federation Dev Server when calling the _getStaticRemotes_ method use the default path for the module-federation.config(in project root folder 'module-federation.config.js'), which can be inconvenient in certain scenarios and make unable to use a typescript config file(like my module-federation.config.ts). It would be helpful to have the ability to specify a custom path for the module-federation.config file when starting the dev server. This would allow users to load the config from a specific location(ex. a libs containing all the manifest) and to use a ts config.

## Motivation

Implementing a custom module-federation.config path in the Nx Angular Module Federation Dev Server would provide more flexibility, capabilities to use typed config file and convenience for users who need to work with it in non-standard scenarios.
## Suggested Implementation

- Adding a customModuleFederationConfig?: object; options in builders/module-federation-dev-server/schema.d.ts and schema.json;
```json
"customModuleFederationConfig": {
        "type": "object",
        "description": "use a custom module federation config file",
        "properties": {
              "path": { "type": "string" }
         }
}
```

- Change the _getStaticRemotes_ method to use the new path when declaring _const mfConfigPath_.

peppoasap:
## Current Behavior
The module-federation.config.js file is placed in the app/src folder and it's not possible to move it or use a typescript config file due to module federation dev server builder error caused by the static path to this file.

## Expected Behavior
Allow users to specify an option in the project.json with the path to the config file. Fallback to origin location.
The config file can also be a typescript file.

## Related Issue(s)

Fixes #15739 

> PS: This is my first PR in a open source project, I followed the contributions guidelines and some other fixes and pr made on the project to try to fit better into the codebase. I hope that's fine, if there is any problem let me know and I will do my best to fix it asap, thanks :)

### DIFF
--- a/packages/devkit/package.json
+++ b/packages/devkit/package.json
@@ -35,7 +35,7 @@
     "semver": "7.3.4"
   },
   "peerDependencies": {
-    "nx": ">= 14.1 <= 16"
+    "nx": ">= 15 <= 17"
   },
   "publishConfig": {
     "access": "public"


packages/angular/src/builders/module-federation-dev-server/schema.json
```diff
@@ -122,6 +122,13 @@
     "pathToManifestFile": {
       "type": "string",
       "description": "Path to a Module Federation manifest file (e.g. `my/path/to/module-federation.manifest.json`) containing the dynamic remote applications relative to the workspace root."
+    },
+    "moduleFederationConfigOptions": {
+      "type": "object",
+      "description": "Custom Module Federation configuration to be used instead of the default configuration generated by the builder. This can be useful for multi-repository module federation setups where the host application uses a remote application from an external repository.",
+      "properties": {
+        "path": { "type": "string" }
+      }
```
Coly010, line 131: We should simplify this to just be of type string

```json
"moduleFederationConfig": {
  "type": "string",
  "description": "...",
}
```

Coly010: It might also be worth noting that if this supports paths outside of the project root, that the inputs for caching for this project will also need to be updated to include the path to this file. 

This warning should potentially be added to the description.

If the project's inputs are not updated to include this file (if it lives outside the project root), then you can encounter cache hits even when you change this config file.

packages/angular/src/builders/module-federation-dev-server/module-federation-dev-server.impl.ts
```diff
@@ -49,14 +49,45 @@ export function executeModuleFederationDevServerBuilder(
     pathToManifestFile = userPathToManifestFile;
   }
 
+  let pathToModuleFederationConfigFile = join(
+    context.workspaceRoot,
+    project.sourceRoot,
```
Coly010, line 54:
```suggestion
    project.root,
```

packages/angular/src/builders/utilities/module-federation.ts
```diff
@@ -127,3 +128,43 @@ export function validateDevRemotes(
     );
   }
 }
+
+export function resolveModuleFederationConfigFile(path: string): {
+  remotes: Remotes;
+} {
+  tsNodeRegister(path);
+  const mfeConfigFile = require(path);
+  console.log('mfeConfigFile', mfeConfigFile);
```
Coly010, line 137: Remove

packages/angular/src/utils/mf/utils.ts
```diff
@@ -40,6 +40,9 @@ export async function getModuleFederationConfig(
   determineRemoteUrl: (remote: string) => string,
   options: { isServer: boolean } = { isServer: false }
 ) {
+  //check if mfConfig is from a js or ts configuration and take default values accordingly
+  mfConfig = (mfConfig as any).default ?? mfConfig;
```
Coly010, line 44: Is this necessary? Doesn't this PR already return the config from either default export or as is when resolving the config?

packages/angular/src/builders/module-federation-dev-ssr/schema.json
```diff
@@ -83,6 +83,13 @@
     "pathToManifestFile": {
       "type": "string",
       "description": "Path to a Module Federation manifest file (e.g. `my/path/to/module-federation.manifest.json`) containing the dynamic remote applications relative to the workspace root."
+    },
+    "moduleFederationConfigOptions": {
+      "type": "object",
+      "description": "Custom Module Federation configuration to be used instead of the default configuration generated by the builder. This can be useful for multi-repository module federation setups where the host application uses a remote application from an external repository.",
+      "properties": {
+        "path": { "type": "string" }
+      }
```
Coly010, line 92: We should simplify this to just be of type string

```json
"moduleFederationConfig": {
  "type": "string",
  "description": "...",
}
```

It might also be worth noting that if this supports paths outside of the project root, that the inputs for caching for this project will also need to be updated to include the path to this file.

This warning should potentially be added to the description.

If the project's inputs are not updated to include this file (if it lives outside the project root), then you can encounter cache hits even when you change this config file.

packages/angular/src/builders/module-federation-dev-ssr/schema.d.ts
```diff
@@ -14,4 +14,5 @@ export interface Schema {
   skipRemotes?: string[];
   verbose: boolean;
   pathToManifestFile?: string;
+  moduleFederationConfigOptions?: { path: string };
```
Coly010, line 17: This should be changed in line with the schema.json changes

```
moduleFederationConfig?: string
```

packages/angular/src/builders/module-federation-dev-server/schema.d.ts
```diff
@@ -20,4 +20,5 @@ export interface Schema {
   devRemotes?: string[];
   skipRemotes?: string[];
   pathToManifestFile?: string;
+  moduleFederationConfigOptions?: { path: string };
```
Coly010, line 23: This should be changed in line with the schema.json changes

```
moduleFederationConfig?: string
```

Coly010 (changes requested):
Hey! Thanks for contributing to Nx with this PR 🎉

I've added some comments to address from what is in this PR already.

Also, it would be worthwhile to write an e2e test that uses the TS Config that you're trying to add support for.
This is useful for two reasons. 

1. It will verify that it works
2. It will provide an example of envisioned usage of this.

peppoasap:
I'm so sorry, I did a huge mess with a badly rebase and accidentally, automatically asked for the review of a lot of people, sorry about that :S (I forced push a fix for that, but all the reviewers are still here...sorry)

**Really sorry again, I need to study more how to git correctly in those cases.**

_________
_Related to the commit:_
I made some fixes based on the suggestions you gave me, I had some problems for a while to correctly handle TS config file both in project.json and webpack custom config in withModuleFederation, because the ts-node lazy compiler, double registered the files and actually did the compilation two times, but after a bit of debugging and tries I found a working way, I'm not sure this is the best solution, due to my lack of knowledge about ts-node, that's the best I found actually.
I made some manual test and it work well both with js and ts file, I will write some e2e test as soon as possible to complete the work.
Please let me know if there are any problems or if something need to be fixed or improved, thanks so much.

packages/angular/src/builders/utilities/webpack.ts
```diff
@@ -27,7 +28,7 @@ export async function mergeCustomWebpackConfig(
 export function resolveCustomWebpackConfig(path: string, tsConfig: string) {
   tsNodeRegister(path, tsConfig);
 
-  const customWebpackConfig = require(path);
+  const customWebpackConfig = __importDefault(require(path));
```
Coly010, line 31: Hey :) Can you explain why you're using this from `tslib`? 

We don't want to directly use tslib functions in this code, so if this can be done in a different manner, I would suggest to do that.

Coly010 (changes requested):

peppoasap:
Hey Coly, it's me again, messed up with Git...I've never worked with a Git like that and I'm struggling doing some basic stuffs. I wanted to align my branch with the master because I've seen a lot of commits and changes and also some conflict with some of my changes, so to test on a real scenario I did a rebase and everything worked fine.
When I was trying to push my new changes I got an error from git, because my local branch wasn't on the same ref. of the origin, so after a while of trying I just deleted my branch on origin and pushed the new local one...and here we are ;(

I think this PR is not more useable, what should I do? Can you give me some suggestions on how to work correctly with this Git? Do I need to align my working branch with the Master one everytime I can or I'm going in the wrong way?

____
_Related to the changes:_
- Actually the `__importDefault` was not useful, it was a leftover from some trials to solve the problem I had with ts-node double registration, cleaned that.
- I wrote a small e2e test starting from module-federation.test.ts to verify that the basic operation of an mf configuration with ts was working, but I can't in any way start the e2e tests, it always fails me in workspace creation phase, I tried clearing npx cache as recommended, but nothing, can you suggest me a possible solution?

**PS: Thank you for your patience and sorry for the various wasted time I am creating.**

PR #16306, opened 2023-04-14
feat(testing): add "video" option to cypress executor

### LINKED ISSUE #16305: Add video option to cypress executor

## Description

I would like to be able to set the [video option](https://docs.cypress.io/guides/references/configuration#Videos) from the cypress executor options
## Motivation

Currently we have a "ci" configuration for running our e2e tests in Jenkins. We don't want it to create video's on this run, however the only way to set that currently (afaik) is by copying cypress.config.ts and overriding the "cypressConfig" option. It would be cleaner if I was able to set it as a configuration option in the project.json
## Suggested Implementation

See [PR](https://github.com/nrwl/nx/pull/16306)

JesseZomer:
adds the ability to set the video option from the cypress executor options

https://docs.cypress.io/guides/references/configuration#Videos

closed #16305

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/cypress/executors/cypress.json
+++ b/docs/generated/packages/cypress/executors/cypress.json
@@ -132,6 +132,11 @@
         "description": "A comma delimited list to identify a run with.",
         "aliases": ["t"]
       },
+      "video": {
+        "type": "boolean",
+        "description": "Whether Cypress will capture a video of the tests run with cypress run.",
+        "default": true
+      },
       "port": {
         "oneOf": [
           { "type": "string", "enum": ["cypress-auto"] },
--- a/packages/cypress/src/executors/cypress/cypress.impl.spec.ts
+++ b/packages/cypress/src/executors/cypress/cypress.impl.spec.ts
@@ -440,6 +440,49 @@ A generator to migrate from v8 to v10 is provided. See https://nx.dev/cypress/v1
     );
   });
 
+  it('should forward video option', async () => {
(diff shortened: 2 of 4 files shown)


barbados-clemens:
Hi @JesseZomer thanks for the PR!

Typically the way executor options work is they mimic the CLI tool in which they are supporting.

Since Cypress doesn't have a --video or like option on the CLI I'm not too sure if this is something we want to support as now this opens the door for all options configured in via the cypress config file to be executor options, which is something we don't want to do right now. 

Reading the GH Issue you opened I do see the reasoning behind it, but instead of providing the CLI option, you can look at setting configurations options dynamically via the `setupNodeEvents`
```ts
export default defineConfig({
  e2e: {
    ...nxE2EPreset(__filename),
    setupNodeEvents(on, config) {
      return {
        ...config,
        video: !process.env.CI
      }
    }
  }
})
```

or via environment variables which might be the way to go for your use case. 

https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_

```shell
CYPRESS_VIDEO=false nx affected e2e
```

or just define it globally which will depend on your CI provider. 

Optionally you can write a .env file with that value and Nx will load that file for you automatically.
https://nx.dev/recipes/environment-variables/define-environment-variables

Hopefully that gives you some direction on how to achieve the result you're looking for an explains why we aren't accepting the PR for the --video option.

PR #16324, opened 2023-04-16
feat(vite): add `--testFile` argument to @nrwl/vite:test

### LINKED ISSUE #16280: running @nrwl/vite:test with single file or list of files

## Description

Ability to run `@nrwl/vite:test` with a `--testFile` or `--testFiles` argument. Like what is available in `@nrwl/jest:jest` (jest has [--testFile](packages/vite/src/executors/test/vitest.impl.ts))
## Motivation

It would make the developer experience better when writing and debugging single test files.
## Suggested Implementation

modify `vitestExecutor` in `packages/vite/src/executors/test/vitest.impl.ts` to pass value from `--testFiles` argument to `startVitest`
```diff
+ const testFiles = options.testFiles?.split(' ');
+ const ctx = await startVitest(options.mode, testFiles ?? [], settings);
- const ctx = await startVitest(options.mode, [], settings);
``` 
## Alternate Implementations

For single file (`--testFile`):
```diff
+ const cliFilters = options.testFiles ? [options.testFile] : [];
+ const ctx = await startVitest(options.mode, cliFilters, settings);
- const ctx = await startVitest(options.mode, [], settings);
```

liammcateer:
closed #16280

## Current Behavior

When running the `@nrwl/vite:test` executor, we don't currently have the option to run a specific test file.
## Expected Behavior

After this change, `@nrwl/vite:test` will have a new option `--testFile` available, which can be used to run a single test file.
## Related Issue(s)

#16280 
Fixes #

### DIFF
--- a/docs/generated/packages/vite/executors/test.json
+++ b/docs/generated/packages/vite/executors/test.json
@@ -57,6 +57,10 @@
       "reportsDirectory": {
         "type": "string",
         "description": "Directory to write coverage report to."
+      },
+      "testFile": {
+        "description": "The name of the file to test.",
+        "type": "string"
       }
     },
     "required": [],
--- a/packages/vite/src/executors/test/schema.d.ts
+++ b/packages/vite/src/executors/test/schema.d.ts
@@ -8,4 +8,5 @@ export interface VitestExecutorOptions {
   update?: boolean;
   reportsDirectory?: string;
   coverage?: boolean;
+  testFile?: string;
 }
--- a/packages/vite/src/executors/test/schema.json
+++ b/packages/vite/src/executors/test/schema.json
@@ -56,6 +56,10 @@
     "reportsDirectory": {
       "type": "string",
       "description": "Directory to write coverage report to."
+    },
+    "testFile": {
+      "description": "The name of the file to test.",
+      "type": "string"
     }
   },
   "required": [],
--- a/packages/vite/src/executors/test/vitest.impl.ts
+++ b/packages/vite/src/executors/test/vitest.impl.ts
@@ -56,8 +56,9 @@ export async function* vitestExecutor(
   const nxReporter = new NxReporter(options.watch);
   const settings = await getSettings(options, context);
   settings.reporters.push(nxReporter);
+  const cliFilters = options.testFile ? [options.testFile] : [];
 
-  const ctx = await startVitest(options.mode, [], settings);
+  const ctx = await startVitest(options.mode, cliFilters, settings);
 
   let hasErrors = false;
 


docs/generated/packages/vite/executors/test.json
```diff
@@ -57,6 +57,10 @@
       "reportsDirectory": {
         "type": "string",
         "description": "Directory to write coverage report to."
+      },
+      "testFile": {
+        "description": "The name of the file to test.",
+        "type": "string"
```
barbados-clemens, line 63: the option is called `testNamePattern` (aliased to `-t`) in the vitest cli docs
we should use that name instead as the executor options should closely mimic the CLI of the tool they are wrapping.
https://vitest.dev/guide/cli.html#options

```suggestion
      "testFilePattern": {
        "aliases": ["t"],
        "description": "Run tests with full names matching the pattern",
        "type": "string"
```

packages/vite/src/executors/test/schema.d.ts
```diff
@@ -8,4 +8,5 @@ export interface VitestExecutorOptions {
   update?: boolean;
   reportsDirectory?: string;
   coverage?: boolean;
+  testFile?: string;
```
barbados-clemens, line 11: should be named `testNamePattern` see comment in schema.json

packages/vite/src/executors/test/schema.json
```diff
@@ -56,6 +56,10 @@
     "reportsDirectory": {
       "type": "string",
       "description": "Directory to write coverage report to."
+    },
+    "testFile": {
+      "description": "The name of the file to test.",
+      "type": "string"
```
barbados-clemens, line 62: the option is called `testNamePattern` (aliased to `-t`) in the vitest cli docs
we should use that name instead as the executor options should closely mimic the CLI of the tool they are wrapping.

https://vitest.dev/guide/cli.html#options

```suggestion
      "testFilePattern": {
        "aliases": ["t"],
        "description": "Run tests with full names matching the pattern",
        "type": "string"
```

packages/vite/src/executors/test/vitest.impl.ts
```diff
@@ -56,8 +56,9 @@ export async function* vitestExecutor(
   const nxReporter = new NxReporter(options.watch);
   const settings = await getSettings(options, context);
   settings.reporters.push(nxReporter);
+  const cliFilters = options.testFile ? [options.testFile] : [];
 
-  const ctx = await startVitest(options.mode, [], settings);
+  const ctx = await startVitest(options.mode, cliFilters, settings);
```
barbados-clemens, line 61: unsure if the option should be a `cliFilter` or an option in the settings?

barbados-clemens (changes requested):
Thanks for the contribution, I left some comments on the arg naming. 

and unsure where this arg should be passed. I'd assume it's in the settings passed to vitest as i'm unsure how you'd actually pass args to vitest cli to the cliFilter parameter.

docs/generated/packages/vite/executors/test.json
barbados-clemens: note: this file will auto update when the schema.json file is updated and `yarn documentation` is updated.

liammcateer: `@nrwl/vite:test` already has an option called [testNamePattern](https://nx.dev/packages/vite/executors/test#testnamepattern), which does map to the vitest `testNamePattern`. `testNamePattern` doesn't fully achieve the goal of running a single (or multiple) test files, as it looks at the test names (e.g. inside the `describe`, or `it` block), rather than the file names.

I wasn't able to find `testFilePattern` in the vitest docs.

liammcateer: One potential solution is to pass `options._` to `startVitest` like this `const ctx = await startVitest(options.mode, options._ ?? [], settings);`.

if I run `nx test my-lib myTest.spec.ts` the `options` object looks like this
```js
{
  passWithNoTests: true,
  reportsDirectory: '../../coverage/packages/my-lib',
  _: [ 'myTest.spec.ts' ],
  mode: 'test',
  watch: false,
  update: false,
  coverage: false
}
```

I'm just not sure how this could be documented.

barbados-clemens: Ah I didn't see that. 
yeah okay that makes sense. Looks like Vitest just passes in the 'rest of the args' to the `startVitest` function. 

https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/cli.ts#L50-L52

having a --testFile arg I think probably makes the most sense to me in this case as passing _ is probably not the best move as it'll most likely be subject to changing over time and breaking people without notice. 

any objections @mandarini?

liammcateer: Hey there @barbados-clemens @mandarini, just checking in to see if there are any updates on this PR?

barbados-clemens: @jaysoo could you take look at this PR since @mandarini is out of the office?

mandarini:
Hi there @liammcateer ! Thank you for your PR! I think it looks good! Can you please rebase, and then also run `pnpm documentation` to update the documentation, too?

mandarini (approved):
@liammcateer thanks! :)

barbados-clemens (approved):

PR #16330, opened 2023-04-16
feat(js): standalone executor for generating package.json

jensbodal:
* generates a package.json based on the project's dependencies in node_modules
* moves getExtraDependencies helper from esbuild package to js package

## Current Behavior

Currently each executor implements their own (generally similar) logic for generating a package.json.  Adding this executor and function should allow for

1. Decomposing existing implementations to use a common method
2. Other executors to more easily add the ability to include this option in their schema
3. Any executor that lacks this functionality to use this executor directly to "wrap" an existing executor using `dependsOn`

## Expected Behavior

If an executor is missing the ability to generate a `package.json` file this executor can be called in `dependsOn` to provide the functionality. 

## Related Issue(s)

Related to #16042 which doesn't necessarily fix that issue but can allow generating a `package.json` by calling this executor in the depends on block.

## Example

`project.json`:

```json
{
  "name": "api",
  //...
  "targets": {
    "build": {
      "dependsOn": [
        { "projects": "self", "target": "build-api", "params": "forward" }
      ],
      "executor": "@nx/js:generate-package-json",
      "options": {
        "outputPath": "dist/apps/api",
        "main": "apps/api/src/main.ts",
        "tsConfig": "apps/api/tsconfig.app.json"
      }
    },
    "build-api": {
      "executor": "@nx/vite:build",
      //...
      "configurations": {
        "production": {
          "mode": "production"
        }
      }
    },
    "serve": {
      "executor": "@nx/vite:dev-server",
      //...
      "configurations": {
        "production": {
          "buildTarget": "api:build:production"
        }
      }
    }
  }
}
```


jensbodal:
Is anyone able to rerun the circle ci build? It mentions it can’t find an nx dependency in the failing build

SmoshySmosh:
Looking forward to using this executor.

jensbodal:
Haven't had time to figure it out yet but having issues rebuilding documentation after rebasing with latest changes.  Will see if I have some time to figure this out later this week.

jensbodal:
rebased with master

jensbodal:
> @jensbodal Is there any way I can temporarily use this PR for testing?

You're welcome to if you want. I just rebased with master again. 

As for how the best way to do that I'm not sure since I think part of the release process creates the proper packages from the monorepo.  I'm guessing you could install my fork locally then point your local install to it via a file alias.

meeroslav:
Hi @jensbodal, thank you for this PR. Unfortunately, this is not something we want to implement in nx.

The `@nx/js` executors already have flag `--generatePackageJson` for generating package json and lock file based on the dependencies and we are currently cleaning this up

SmoshySmosh:
> Hi @jensbodal, thank you for this PR. Unfortunately, this is not something we want to implement in nx.
> 
> The `@nx/js` executors already have flag `--generatePackageJson` for generating package json and lock file based on the dependencies and we are currently cleaning this up

@meeroslav I must be misunderstanding here. How does the `@nx/js` executer help create a package.json for the other build executers? They represent two completely different build systems with two different purposes. 

Is it possible that you are saying that the `@nx/js` executer can be used to only create the `package.json` file? If so, can you please provide an example so I can start using it?

meeroslav:
Sorry @SmoshySmosh, your message got somehow lost.

The package json creation logic from `@nx/js` is used under the hood in other executors.

You can see how to use it here: https://nx.dev/recipes/ci/ci-deployment

For libraries, we are now deprecating this functionality and suggest that you use the new linter rule - https://nx.dev/packages/eslint-plugin/documents/dependency-checks

PR #16351, opened 2023-04-17
feat(repo): replace usages of @nrwl/ with @nx/ in tooling packages

FrozenPandaz:
## Current Behavior

Tooling packages generate `@nrwl/` packages.

## Expected Behavior

Tooling packages generate `@nx/` packages.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/migrations/update-15-9-0/update-testing-tsconfig.spec.ts
+++ b/packages/angular/src/migrations/update-15-9-0/update-testing-tsconfig.spec.ts
@@ -4,6 +4,7 @@ import {
   readProjectConfiguration,
   Tree,
   updateJson,
+  updateProjectConfiguration,
 } from '@nx/devkit';
 import { createTreeWithEmptyWorkspace } from 'nx/src/devkit-testing-exports';
 import {
@@ -75,6 +76,18 @@ async function setup(tree: Tree, name: string) {
     skipPackageJson: true,
   });
 
+  const projectConfig = readProjectConfiguration(tree, name);
+  updateProjectConfiguration(tree, name, {
+    ...projectConfig,
+    targets: {
+      ...projectConfig.targets,
+      test: {
+        ...projectConfig.targets.test,
+        executor: '@nrwl/jest:jest',
+      },
+    },
+  });
+
   updateJson(tree, `${name}/tsconfig.spec.json`, (json) => {
     // revert to before jest-preset-angular v13
     delete json.compilerOptions.target;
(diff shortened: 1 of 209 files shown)


packages/angular/src/migrations/update-15-9-0/update-testing-tsconfig.spec.ts
```diff
@@ -75,6 +76,18 @@ async function setup(tree: Tree, name: string) {
...
   });
 
+  const projectConfig = readProjectConfiguration(tree, name);
+  updateProjectConfiguration(tree, name, {
+    ...projectConfig,
+    targets: {
+      ...projectConfig.targets,
+      test: {
+        ...projectConfig.targets.test,
+        executor: '@nrwl/jest:jest',
```
Coly010, line 86: Should this not be `@nx/jest:jest`?

If someone is _only_ migrating to NX 15.9.0 - the Nx 15.9.0 release will have a version of this migration that uses the @nrwl/ scoped packages.

This change will only be applied to people migrating from Nx < 15.9.0 to Nx latest

packages/angular/src/migrations/update-15-9-0/update-testing-tsconfig.spec.ts
```diff
@@ -85,6 +98,18 @@ async function setup(tree: Tree, name: string) {
...
   });
 
+  const libConfig = readProjectConfiguration(tree, `${name}-lib`);
+  updateProjectConfiguration(tree, `${name}-lib`, {
+    ...libConfig,
+    targets: {
+      ...libConfig.targets,
+      test: {
+        ...libConfig.targets.test,
+        executor: '@nrwl/jest:jest',
```
Coly010, line 108: Same as previous

Coly010 (approved):
One comment, but looks good

mandarini (approved):

meeroslav (changes requested):
There are still a lot of files that contain references to `@nrwl/*` e.g.

```
jest-preset.ts
srcipts/depcheck.ts
...
```

nartc (approved):

packages/angular/src/migrations/update-15-9-0/update-testing-tsconfig.spec.ts
FrozenPandaz: This should be `@nrwl/jest:jest`. 

The migrations will run in order, so this will work on anything that's `@nrwl/jest:jest` and then the `16.0.0` migration will change them to `@nx/jest:jest`.

FrozenPandaz:
> There are still a lot of files that contain references to `@nrwl/*` e.g.
> 
> ```
> jest-preset.ts
> srcipts/depcheck.ts
> ...
> ```

Yes, those changes will be elsewhere. We can't change `jest-preset.js` until we update the actual packages that we use. I changed them both in the PR that updates us to Nx `16.0.0-beta.1` (but I'll have to update the PR for `16.0.0-beta.2`

meeroslav (approved):

AgentEnder (approved):

jaysoo (approved):

barbados-clemens (approved):

PR #16373, opened 2023-04-18
fix(react): install rollup package when generating lib

mandarini:
## Current Behavior

The `@nrwl/rollup` package is not getting installed when generating a buildable react library that uses the `@nrwl/rollup:rollup` executor, which results in the following error:

```
npx nx build rollib

 >  NX   Ran target build for project rollib (53ms)

    ✖    0/0 failed
    ✔    0/0 succeeded [0 read from cache]

   View logs and run details at https://nx.app/runs/785BGG9zna

 >  NX   Unable to resolve @nrwl/rollup:rollup.

   Cannot find module '@nrwl/rollup/package.json'
   Require stack:
   - /Users/katerina/Projects/nrwl/test_nx_workspaces/rolls/node_modules/nx/src/utils/package-json.js
   - /Users/katerina/Projects/nrwl/test_nx_workspaces/rolls/node_modules/nx/src/utils/package-manager.js
   - /Users/katerina/Projects/nrwl/test_nx_workspaces/rolls/node_modules/nx/bin/init-local.js
   - /Users/katerina/Projects/nrwl/test_nx_workspaces/rolls/node_modules/nx/bin/nx.js
```

## Expected Behavior

`@nrwl/rollup` package should get installed.

### DIFF
--- a/packages/rollup/src/generators/init/init.spec.ts
+++ b/packages/rollup/src/generators/init/init.spec.ts
@@ -52,6 +52,7 @@ describe('rollupInitGenerator', () => {
       name: expect.any(String),
       dependencies: {},
       devDependencies: {
+        '@nx/rollup': nxVersion,
         tslib: expect.any(String),
       },
     });
--- a/packages/rollup/src/generators/init/init.ts
+++ b/packages/rollup/src/generators/init/init.ts
@@ -33,7 +33,14 @@ export async function rollupInitGenerator(tree: Tree, schema: Schema) {
       }
     );
   } else {
-    task = addDependenciesToPackageJson(tree, {}, { tslib: tsLibVersion });
+    task = addDependenciesToPackageJson(
+      tree,
+      {},
+      {
+        '@nx/rollup': nxVersion,
+        tslib: tsLibVersion,
+      }
+    );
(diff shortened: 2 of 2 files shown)


packages/react/src/generators/library/lib/add-rollup-build-target.ts
```diff
@@ -30,6 +30,7 @@ export async function addRollupBuildTarget(
     host,
     {},
     {
+      '@nx/rollup': nxVersion,
```
mandarini, line 33: The weird thing is that on line 23 of the same file we're doing `ensurePackage` for the `@nx/rollup` package, and it still doesn't get installed :/

jaysoo: it doesn't get added to the workspace, i think it needs to be in the init generator

jaysoo: We need to fix this in `@nx/rollup:init` generator. The bug is that `@nx/rollup` package is only added when compiler is swc. It should be added for all cases.

jaysoo (approved):

PR #16377, opened 2023-04-18
feat(angular): replace usages of @nrwl with @nx

FrozenPandaz:
## Current Behavior

The Angular package generates `@nrwl/*`

## Expected Behavior

The Angular package generates `@nx/*`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/convert-to-with-mf/lib/get-webpack-config-path.ts
+++ b/packages/angular/src/generators/convert-to-with-mf/lib/get-webpack-config-path.ts
@@ -7,9 +7,9 @@ export function getWebpackConfigPath(
   let pathToWebpackConfig = '';
   for (const target of Object.values(project.targets ?? {})) {
     if (
-      target.executor === '@nx/angular:webpack-browser' ||
-      (target.executor === '@nrwl/angular:webpack-browser' &&
-        target.options.customWebpackConfig?.path)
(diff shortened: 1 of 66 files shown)


barbados-clemens (approved):

packages/angular/src/generators/convert-to-with-mf/lib/get-webpack-config-path.ts
```diff
@@ -7,8 +7,9 @@ export function getWebpackConfigPath(
   let pathToWebpackConfig = '';
   for (const target of Object.values(project.targets ?? {})) {
     if (
-      target.executor === '@nrwl/angular:webpack-browser' &&
-      target.options.customWebpackConfig?.path
+      target.executor === '@nx/angular:webpack-browser' ||
+      (target.executor === '@nrwl/angular:webpack-browser' &&
+        target.options.customWebpackConfig?.path)
```
Coly010, line 12: Should this be
```suggestion
      (target.executor === '@nx/angular:webpack-browser' ||
      target.executor === '@nrwl/angular:webpack-browser') &&
        target.options.customWebpackConfig?.path
```

Coly010 (changes requested):
Everything looks good, but there's one conditional that I believe is incorrect

Coly010 (approved):

PR #16379, opened 2023-04-18
feat(node): replace usages of @nrwl with @nx

FrozenPandaz:
## Current Behavior

Node packages generate `@nrwl/`

## Expected Behavior

Node packages generate `@nx/`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/express/src/generators/application/application.ts
+++ b/packages/express/src/generators/application/application.ts
@@ -60,13 +60,8 @@ server.on('error', console.error);
     toJS(tree);
   }
 }
-// TODO (nicholas): Remove After Nx 16
-// @deprecated Use `nx g @nrwl/node:app --framework=express instead.
-export async function applicationGenerator(tree: Tree, schema: Schema) {
-  console.warn(
-    'As of Nx 16 using `nx g @nrwl/express:app` has been deprecated! Use `nx g @nrwl/node:app --framework=express instead.'
-  );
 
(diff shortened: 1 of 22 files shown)


packages/express/src/generators/application/application.ts
```diff
@@ -60,13 +60,8 @@ server.on('error', console.error);
...
   }
 }
-// TODO (nicholas): Remove After Nx 16
-// @deprecated Use `nx g @nrwl/node:app --framework=express instead.
-export async function applicationGenerator(tree: Tree, schema: Schema) {
-  console.warn(
-    'As of Nx 16 using `nx g @nrwl/express:app` has been deprecated! Use `nx g @nrwl/node:app --framework=express instead.'
-  );
 
+export async function applicationGenerator(tree: Tree, schema: Schema) {
```
jaysoo, line 64: I think we either remove this or reuse it in `@nrwl/node:app` when choosing express. Right now it is duplicated.

Out of scope for this PR, but should do a follow-up.

jaysoo (approved):

packages/express/src/generators/application/application.ts
FrozenPandaz: I think we should reuse it.

I think `@nx/node` can do the framework.. but it would proxy to `@nx/express` (which is the real one).

I believe it's similar for `@nx/node:app --framework nest` right?

PR #16284, opened 2023-04-13
feat(repo): migrate to pnpm

meeroslav:
This PR migrates our repo from `yarn classic` to `pnpm`:

- Migrate lock file
- Migrate CI and Nightly configs
- Add missing explicitly used packages
- Fix invalid imports
- Fix local scripts
- Fix broken tests

Current status:
- [X] Lint, Build and Test are passing
- [X]  CI and nightly are running properly
- [x] One `e2e-js` `swc` test is failing and several `e2e-react-native` tests are failing

## Related Issue(s)

Fixes #

PR #16390, opened 2023-04-19
fix(js): update swc options so path mappings can work in all environments

jaysoo:
This PR updates `@nx/js:swc` executor so it properly compiles with `jsc.paths` set in `.swcrc` for all environments. Another PR is failing because the compiled code has wrong path mapping. https://github.com/nrwl/nx/pull/16284

**Details:**
1. It looks like we're setting `jsc.baseUrl` with the wrong default. It should be `.` since `cwd` is set to the project root. That way, when mappings are used in output, it will be relative to the project root. Removing that whole logic and relying on users to set `baseUrl` seems more reliable. Also with local testing (macos, debian) it doesn't look like `baseUrl` is needed by swc at all (at least on latest version).
2. Passing both `--no-swc` and `--config-file` together is causing wrong paths in output as well for some reason (able to replicate this in CI with just `npx swc` and not Nx, but not locally). Since we're passing the config file anyway, we don't need the `--no-swc` option to find the file.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/js/src/js-swc.test.ts
+++ b/e2e/js/src/js-swc.test.ts
@@ -134,8 +134,9 @@ export function x() {
 
     // update .swcrc to use path mappings
     updateJson(`libs/${lib}/.swcrc`, (json) => {
+      json.jsc.baseUrl = '.';
       json.jsc.paths = {
-        'src/*': ['src/*'],
+        '~/*': ['./src/*'],
       };
       return json;
     });
@@ -144,7 +145,7 @@ export function x() {
     updateFile(`libs/${lib}/src/lib/${lib}.ts`, () => {
       return `
 // @ts-ignore
-import { x } from 'src/x';
+import { x } from '~/x';
 
 export function myLib() {
   console.log(x());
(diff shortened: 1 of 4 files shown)


meeroslav (approved):

nartc (approved):

e2e/js/src/js-swc.test.ts
```diff
@@ -134,8 +134,9 @@ export function x() {
 
     // update .swcrc to use path mappings
     updateJson(`libs/${lib}/.swcrc`, (json) => {
+      json.jsc.baseUrl = '.';
       json.jsc.paths = {
-        'src/*': ['src/*'],
+        '~/*': ['./src/*'],
```
nartc, line 139: comment: so the key is to use relative path only for path mappings with swc config file?

jaysoo: yes, and the `baseUrl` is only added to pass CI. It's not needed locally (MacOS/Debian), but seems to be needed in CI even with the same swc versions.

PR #16407, opened 2023-04-19
feat(react): add @nx/web as a dependency since it is needed by cypress/etc.

jaysoo:
We previously removed `@nrwl/web` as a dependency since it added webpack and other packages that may not be needed. Now that webpack and jest are no longer installed by default, we can add it back in.

## Why?

Other optional packages such as `@nx/cypress` pulls in the web package, so having it installed as a dependency initially saves time. Also, `@nx/web:file-server` is used in certain setups in React, such as module federation.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -38,7 +38,8 @@
     "minimatch": "3.0.5",
     "@nx/devkit": "file:../devkit",
     "@nx/js": "file:../js",
-    "@nx/linter": "file:../linter"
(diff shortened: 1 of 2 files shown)


xiongemi (approved):

mandarini (approved):

FrozenPandaz (changes requested):
Can you update any usages of `ensurePackage` please?

FrozenPandaz (approved):

PR #16428, opened 2023-04-20
feat(vite): add option to generate a package.json to build executor

jwalton9:
## Current Behavior

There's currently no way of generating a minimal `package.json` when using the `vite:build` executor

## Expected Behavior

The ability to generate a `package.json` including only the dependencies of the project should be enabled by setting `generatePackageJson: true` in the `vite:build` executor options
## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/vite/src/vite.test.ts
+++ b/e2e/vite/src/vite.test.ts
@@ -9,6 +9,7 @@ import {
   newProject,
   promisifiedTreeKill,
   readFile,
+  readJson,
   rmDist,
   runCLI,
   runCLIAsync,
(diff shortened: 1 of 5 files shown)


packages/vite/src/executors/build/build.impl.ts
```diff
@@ -54,6 +59,25 @@ export async function* viteBuildExecutor(
       },
       context
     );
+  } else if (options.generatePackageJson) {
```
jaysoo, line 62: We should merge the detected dependencies into the original `package.json` file. There could be other things like the name or keyword fields that users want to maintain manually.

e2e/vite/src/vite.test.ts
```diff
@@ -144,6 +146,24 @@ describe('Vite Plugin', () => {
         expect(fileExists(`dist/apps/${myApp}/package.json`)).toBeFalsy();
         rmDist();
       }, 200_000);
+
+      it('qqqqq should build application with package json generation', async () => {
```
jaysoo, line 150: Remove the `qqqqq` part, and we should test with and without existing `package.json` file.

jaysoo (changes requested):
Hey @jwalton9 thanks for the PR! I think it's a good idea. Let me know if you need more clarification and I can elaborate more.

Since Nx 16 is right around the corner, I think we'll leave this feature out until after the release. It is already a huge release as is.

packages/vite/src/executors/build/build.impl.ts
jwalton9: Seems like the `createPackageJson` helper already does this so I've added the test. Also made `generatePackageJson` take precedence as existing `package.json` files were just being copied over

jensbodal:
Fwiw I tried adding this a month ago here https://github.com/nrwl/nx/pull/16045 and then three weeks ago with a more generic solution here: https://github.com/nrwl/nx/pull/16330.  Either way I just want this feature available but seems odd that my first one was just closed without feedback or offer of improving it.

mandarini:
Hello @jwalton9 ! I'm taking a look at this right now, and if all's well I'll approve and merge, thank you for your contribution!

@jensbodal thank you for your comment! We are trying our best to meet our users needs! However, we did reply to your PR with feedback! :) https://github.com/nrwl/nx/pull/16045#issuecomment-1494310689 In any case, I am sorry if you felt that your PR was dismissed without a reason, we are trying our best! :)

mandarini (approved):
Thank you for your PR!

mandarini (approved):
Thank you for your PR!

jaysoo (approved):

PR #16435, opened 2023-04-20
feat(bundling): remove esbuild 0.16.0 support

jaysoo:
esbuild 0.17 changed some of the programmatic interfaces (e.g. watch), and we had to keep two separate implementations to allow users to keep using 0.16 (if they skipped the migration). With Nx 16 we want to remove the old interface support.

## Notes

- If a project uses esbuild v0.16's programmatic interface (not the esbuild CLI) then they will need to update their own code. If they just use it through Nx then it's fine.
- Remove `esbuildVersion` from node package and use the one from js instead.
- Remove e2e test that tests against esbuild 0.16, which is currently failing nightly due to bad peer deps.

### DIFF
--- a/packages/esbuild/src/executors/esbuild/esbuild.impl.ts
+++ b/packages/esbuild/src/executors/esbuild/esbuild.impl.ts
@@ -91,236 +91,103 @@ export async function* esbuildExecutor(
     packageJsonResult = await copyPackageJson(cpjOptions, context);
   }
 
-  if ('context' in esbuild) {
-    // 0.17.0+ adds esbuild.context and context.watch()
-    if (options.watch) {
-      return yield* createAsyncIterable<{ success: boolean; outfile?: string }>(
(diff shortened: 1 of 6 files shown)


packages/esbuild/src/executors/esbuild/esbuild.impl.ts
```diff
@@ -91,236 +91,103 @@ export async function* esbuildExecutor(
...
-          let hasTypeErrors = false;
-          const results = await Promise.all(
-            options.format.map(async (format, idx) => {
-              const esbuildOptions = buildEsbuildOptions(
-                format,
-                options,
-                context
-              );
-              const watch =
+  if (options.watch) {
```
jaysoo, line 94: This code block is the same as what's in `if ('context' in esbuild) {...}` previously.

ndcunningham (approved):

PR #16467, opened 2023-04-21
feat(testing): add --strict flag for jest-preset-angular test env opts

### LINKED ISSUE #16352: enable jest runner to fail when errorOnUnknownElements or errorOnUnknownProperties
## Description
When I run my Jest tests in my Angular Nx monorepo I see that whenever a component or property is not properly imported in the .spec.ts file the test does not fail, although in the console an error is clearly visible. 
Such as
image

After looking for solutions I found that Angular has a way to fix it as described here: https://github.com/angular/angular/issues/36430#issuecomment-1117801535

I would like to configure at a root level my projects. I do not intend to configure every single test files for catching these errors. I would like to alter e.g. the root `jest.preset.js` file.

At another library (https://thymikee.github.io/jest-preset-angular/docs/getting-started/test-environment/) I saw that it should be possible with proper setup, adding `errorOnUnknownElements: true` and `errorOnUnknownProperties: true` to the jest config (at the `testEnvironmentOptions` property which in theory exists in `jest.preset.js` as well), but adding it to the file did not affect at all.

## Motivation
I think if an error occures on the console a test should fail. The current behaviour is quite misleading, allowing tests to pass, although they did not.

## Suggested Implementation
With adding some configurations to the root `jest.preset.js` file one should be able to toggle this behaviour. In my opinion this should be the default behaviour.

## Alternate Implementations
This could also be set at an app/lib level, but it is code duplication and can be opted out which is not good in my opinion.

barbados-clemens:
## Current Behavior

jest-preset-angular has options to configure ng TestBed to error on unknown properties and elements but must be manually configured in each test-setup file

## Expected Behavior

when creating an angular project. providing the --strict flag should auto setup jest-preset-angular to fail when unknown properties/elements are used

## Related Issue(s)

Fixes #16352

### DIFF
--- a/packages/angular/src/generators/application/lib/add-unit-test-runner.ts
+++ b/packages/angular/src/generators/application/lib/add-unit-test-runner.ts
@@ -1,4 +1,4 @@
-import type { Tree } from '@nx/devkit';
+import { Tree, joinPathFragments } from '@nx/devkit';
 import type { NormalizedSchema } from './normalized-schema';
 
 import { jestProjectGenerator } from '@nx/jest';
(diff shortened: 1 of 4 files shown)


Coly010 (approved):

docs/generated/packages/jest/generators/jest-project.json
```diff
@@ -75,6 +75,11 @@
         "type": "boolean",
         "default": false,
         "description": "Use JavaScript instead of TypeScript for config files"
+      },
+      "strict": {
```
FrozenPandaz, line 79: Can you remove this option and keep it scoped to the Angular generator?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #16469, opened 2023-04-21
fix(nextjs): remove the need to install @nx/next for production builds

### LINKED ISSUE #15931: @nrwl/next: Cannot find module '@nrwl/next/plugins/with-nx'
### Documentation issue

- [ ] Reporting a typo
- [X] Reporting a documentation bug
- [ ] Documentation improvement
- [ ] Documentation feedback

### Is there a specific documentation page you are reporting?

[@nrwl/next -> Setting up Next.js](https://nx.dev/packages/next#setting-up-next.js)

### Additional context or description

With changes from https://github.com/nrwl/nx/pull/15824 package `@nrwl/next` needs to be included in `dependencies`, not in the `devDependencies`, otherwise generated `package.json` doesn't have `@nrwl/next` and it's not possible to run the Next's application because of error:

```
error - Failed to load next.config.js, see more info here https://nextjs.org/docs/messages/next-config-error
Error: Cannot find module '@nrwl/next/plugins/with-nx'
```

Workspace generated by command `npx create-nx-workspace@latest --preset=next` have `@nrwl/next` in the `dependencies`.

jaysoo:
This PR brings back the compiled version of the `withNx` function. This allows `@nx/next` to not be installed as a production dependency, which avoids pulling in `@nx/devkit` and other dev packages.

## Current Behavior
`@nx/next` or `@nrwl/next` must be installed as production dependency.

## Expected Behavior
`@nx/next` or `@nrwl/next` don't need to be installed for the compiled app to run.

## Related Issue(s)

Fixes #15931

### DIFF
--- a/packages/next/index.ts
+++ b/packages/next/index.ts
@@ -6,4 +6,4 @@ export { componentGenerator } from './src/generators/component/component';
 export { libraryGenerator } from './src/generators/library/library';
 export { pageGenerator } from './src/generators/page/page';
 export { withNx } from './plugins/with-nx';
-export { composePlugins } from './src/utils/config';
+export { composePlugins } from './src/utils/compose-plugins';
--- a/packages/next/plugins/with-nx.ts
+++ b/packages/next/plugins/with-nx.ts
@@ -1,23 +1,14 @@
-import {
-  createProjectGraphAsync,
-  joinPathFragments,
-  offsetFromRoot,
-  parseTargetString,
-  ProjectGraph,
-  ProjectGraphProjectNode,
-  Target,
-  workspaceRoot,
-} from '@nx/devkit';
-import {
-  calculateProjectDependencies,
-  DependentBuildableProjectNode,
-} from '@nx/js/src/utils/buildable-libs-utils';
+/**
+ * WARNING: Do not add development dependencies to top-level imports.
+ * Instead, `require` them inline during the build phase.
+ */
+import * as path from 'path';
 import type { NextConfig } from 'next';
 import { PHASE_PRODUCTION_SERVER } from 'next/constants';
-
-import * as path from 'path';
-import { createWebpackConfig, NextConfigFn } from '../src/utils/config';
-import { NextBuildBuilderOptions } from '../src/utils/types';
+import type { NextConfigFn } from '../src/utils/config';
+import type { NextBuildBuilderOptions } from '../src/utils/types';
+import type { DependentBuildableProjectNode } from '@nx/js/src/utils/buildable-libs-utils';
+import type { ProjectGraph, ProjectGraphProjectNode, Target } from '@nx/devkit';
 
 export interface WithNxOptions extends NextConfig {
   nx?: {
@@ -78,6 +69,7 @@ function getNxContext(
   targetName: string;
   configurationName?: string;
 } {
+  const { parseTargetString } = require('@nx/devkit');
   const targetConfig = getTargetConfig(graph, target);
 
   if (
@@ -119,7 +111,6 @@ function getNxContext(
     );
   }
 }
-
 /**
  * Try to read output dir from project, and default to '.next' if executing outside of Nx (e.g. dist is added to a docker image).
  */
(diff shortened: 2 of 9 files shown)


packages/next/index.ts
```diff
@@ -6,4 +6,4 @@ export { componentGenerator } from './src/generators/component/component';
 export { libraryGenerator } from './src/generators/library/library';
 export { pageGenerator } from './src/generators/page/page';
 export { withNx } from './plugins/with-nx';
-export { composePlugins } from './src/utils/config';
```
jaysoo, line 9: Moved `composePlugins` to its own file so we can copy the compiled version to dist during build.

packages/next/plugins/with-nx.ts
```diff
@@ -130,21 +121,35 @@ async function determineDistDirForProdServer(
...
-    const { options, node: projectNode } = getNxContext(graph, originalTarget);
-    const outputDir = `${offsetFromRoot(projectNode.data.root)}${
-      options.outputPath
-    }`;
-    return nextConfig.distDir && nextConfig.distDir !== '.next'
-      ? joinPathFragments(outputDir, nextConfig.distDir)
-      : joinPathFragments(outputDir, '.next');
-  } else {
-    return '.next';
+  try {
```
jaysoo, line 124: This logic will only be needed when running through Nx. i.e. `nx serve <app> --prod`

packages/next/plugins/with-nx.ts
```diff
@@ -130,21 +121,35 @@ async function determineDistDirForProdServer(
...
+      }`;
+      return nextConfig.distDir && nextConfig.distDir !== '.next'
+        ? joinPathFragments(outputDir, nextConfig.distDir)
+        : joinPathFragments(outputDir, '.next');
+    }
+  } catch {
+    // ignored -- fallback to Next.js default of '.next'
   }
+
+  return '.next';
```
jaysoo: For deployment platforms like Vercel, or a self-contained Docker image, the folder should just be `.next` (or whatever the user sets it to).

packages/next/plugins/with-nx.ts
```diff
@@ -130,21 +121,35 @@ async function determineDistDirForProdServer(
...
+      }`;
+      return nextConfig.distDir && nextConfig.distDir !== '.next'
+        ? joinPathFragments(outputDir, nextConfig.distDir)
+        : joinPathFragments(outputDir, '.next');
+    }
+  } catch {
+    // ignored -- fallback to Next.js default of '.next'
   }
+
+  return nextConfig.distDir || '.next';
```
jaysoo, line 150: For Vercel and other platforms (including self-contained Docker images), use `.next` or whatever the user sets in their `next.config.js`.

packages/next/plugins/with-nx.ts
```diff
@@ -155,9 +160,16 @@ export function withNx(
       const { nx, ...validNextConfig } = _nextConfig;
       return {
         ...validNextConfig,
-        distDir: await determineDistDirForProdServer(validNextConfig),
+        distDir: await determineDistDirForProdServer(_nextConfig),
       };
     } else {
+      const {
```
jaysoo, line 166: Inline all of the `require` to run only during build phase. We don't need these to production.

packages/next/src/executors/build/lib/create-next-config-file.ts
```diff
@@ -13,7 +27,77 @@ export function createNextConfigFile(
...
-    copyFileSync(nextConfigPath, join(options.outputPath, 'next.config.js'));
+    const helpersPath = join(options.outputPath, '.nx-helpers');
+    mkdirSync(helpersPath, { recursive: true });
+    copyFileSync(
+      join(__dirname, '../../../utils/compose-plugins.js'),
+      join(helpersPath, 'compose-plugins.js')
+    );
+    writeFileSync(join(helpersPath, 'with-nx.js'), getWithNxContent());
+    writeFileSync(
+      join(helpersPath, 'compiled.js'),
```
jaysoo, line 40: A smaller version of `with-nx.js` that does not depend on Nx packages.

ndcunningham (approved):

PR #16471, opened 2023-04-21
fix(nextjs): support workspace libs with standalone Next.js app

jaysoo:
This PR fixes the setup when generating and using workspace libs in a Next.js standalone app.

1. When `@nx/next:lib` is used, babel files should not be generated since we are using SWC.
2. `compilerOptions.paths: {}` should not exist for standalone `tsconfig.json`, because migrating after generating lib will override the base paths in `tsconfig.base.json` file.

### DIFF
--- a/packages/next/src/generators/application/lib/create-application-files.ts
+++ b/packages/next/src/generators/application/lib/create-application-files.ts
@@ -63,6 +63,10 @@ export function createApplicationFiles(host: Tree, options: NormalizedSchema) {
 
       let { extends: _, ...updatedJson } = json;
 
+      // Don't generate the `paths` object or else workspace libs will not work later.
+      // It'll be generated as needed when a lib is first added.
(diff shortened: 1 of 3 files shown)


packages/next/src/generators/library/library.spec.ts
```diff
@@ -152,14 +152,14 @@ describe('next library', () => {
     expect(appTree.read('my-lib/src/index.ts', 'utf-8')).toContain(
       'React client components'
     );
-    expect(appTree.read('my-lib/src/server.ts', 'utf-8')).toContain(
+    expect(appTree.read('my-lib/src/server.tsx', 'utf-8')).toContain(
```
jaysoo, line 155: Had to generate with a default component, otherwise the Standalone app complains about no exports being found from module.

ndcunningham (approved):

ndcunningham (approved):

PR #16475, opened 2023-04-21
fix(testing): fix react CT w/ vite and dependant projects

barbados-clemens:
## Current Behavior

When using vite with react component testing and trying to import a dependent project, vite will throw an error saying it's outside of the allows list of reading from the file system

## Expected Behavior

vite should be allowed to react dependent projects files in CT

## Related Issue(s)

Note this is only an issue in CT because cypress loads the given vite file which is missing the fs allow entry, but isn't a problem when serving projects via nx due to vite executor providing this value automatically (which cypress does not do)

TODO:
- [x] write tests

### DIFF
--- a/packages/react/plugins/component-testing/index.ts
+++ b/packages/react/plugins/component-testing/index.ts
@@ -5,6 +5,7 @@ import {
 import type { CypressExecutorOptions } from '@nx/cypress/src/executors/cypress/cypress.impl';
 import {
   ExecutorContext,
+  joinPathFragments,
   logger,
   parseTargetString,
   ProjectGraph,
@@ -19,7 +20,8 @@ import {
   getProjectConfigByPath,
 } from '@nx/cypress/src/utils/ct-helpers';
 
-import type { Configuration } from 'webpack';
+import { existsSync, lstatSync } from 'fs';
+import { dirname, join } from 'path';
 type ViteDevServer = {
   framework: 'react';
   bundler: 'vite';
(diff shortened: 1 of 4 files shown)


packages/react/plugins/component-testing/index.ts
```diff
@@ -61,11 +63,38 @@ export function nxComponentTestingPreset(
...
+              command: 'serve',
+            },
+            viteConfigPath
+          );
+          return mergeConfig(resolved.config, {
+            server: {
+              fs: {
+                allow: [
+                  workspaceRoot,
+                  joinPathFragments(workspaceRoot, 'node_modules/vite'),
```
barbados-clemens, line 92: I tried using the same values contained within the vite utils, but they only allowed the library itself and the node_nodules. 

https://github.com/nrwl/nx/blob/master/packages/vite/src/utils/options-utils.ts#L107-L112

Unsure if they were supposed to add all paths being imported by the source library or I misunderstand how that works searchForWorkspaceRoot works.

but went with just the workspace root for now since that works, but is a bit greedy.

jaysoo: I don't think this even matters since that option is to serve up files in whitelisted folders, and `searchForWorkspaceRoot` adds Vite's default behavior back.

Maybe check with @mandarini on why we need that in the first place?

packages/react/plugins/component-testing/index.ts
```diff
@@ -66,6 +68,34 @@ export function nxComponentTestingPreset(
       specPattern: 'src/**/*.cy.{js,jsx,ts,tsx}',
       devServer: {
         ...({ framework: 'react', bundler: 'vite' } as const),
+        viteConfig: async () => {
+          const normalizedPath = lstatSync(pathToConfig).isDirectory()
```
jaysoo, line 72: Do we need to access the fs, or can we check the extension like `.ts'` or `.js`?

packages/react/plugins/component-testing/index.ts
barbados-clemens: all it does is return the path to the project root, which doesn't include the other libraries, so unsure if there is something else extra being done with our vite configs that make it work when using the vite executor or not that I'm missing.

I'll sync with @mandarini when she get's back from PTO. 👍

packages/react/plugins/component-testing/index.ts
barbados-clemens: we can check for the if it's just a .ts/.js file instead of using `lstatSync()`, though unsure why we don't want to use fs?

```ts
const normalizedPath = ['.ts', '.js'].some((ext) => pathToConfig.endsWith(ext))
 ? pathToConfig
 : dirname(pathToConfig);
```

mandarini (approved):

meeroslav (approved):

xiongemi (approved):

FrozenPandaz (approved):

jaysoo (approved):

PR #16488, opened 2023-04-23
feat(react): refactor util `getModuleFederationConfig` to avoid to pass function to determinate the remote url

fyodorovandrei:
## Current Behavior
For the utils `getModuleFederationConfig` you need to pass the parameter  `determineRemoteUrl`.

## Expected Behavior
For the util `getModuleFederationConfig` now no more needs to pass  `determineRemoteUrl`. The util by itself understands which function to call

---

I open a [PR](https://github.com/module-federation/universe/pull/598) in the Module Federation repository to add supports Module Federation remote containers in Storybook. To consume remote MF apps/components in Storybook I created a new [util](https://github.com/module-federation/universe/pull/598/files#diff-e60e4799154ca5a2fbe68da693c5fd0776177ac2b19522cd5776120f039406bc) `withModuleFederation` to replace the existing NX util. In the new util I'm using `getModuleFederationConfig`. With this PR we can use util `getModuleFederationConfig` without passing the parameter `determineRemoteUrl`


fyodorovandrei:
@Coly010 You know more about :) If you have time to take a look will be nice. Thx 👍

jaysoo:
Hey @fyodorovandrei Thanks for the PR. I'll discuss with @Coly010 about this PR, it looks reasonable to me. We do have the Nx 16 release this week, so we will be delayed in reviewing feature PRs.

packages/react/src/module-federation/utils.ts
```diff
@@ -6,17 +6,42 @@ import {
...
+
+    if (!serveTarget) {
+      throw new Error(
+        `Cannot automatically determine URL of remote (${remote}). Looked for property "host" in the project's "serve-server" target.\n
+      You can also use the tuple syntax in your webpack config to configure your remotes. e.g. \`remotes: [['remote1', 'http://localhost:4201']]\``
+      );
+    }
+
+    const host = serveTarget.options?.host ?? 'http://localhost';
+    const port = serveTarget.options?.port ?? 4201;
```
brunos3d, line 32: What about using/creating a similar function to the [findNextAvailablePort](https://github.com/nrwl/nx/blob/37038463ed7a33411ade8aaca6510a3c5be56ac3/packages/angular/src/generators/remote/lib/find-next-available-port.ts#L5) to prevent local port collisions?

Coly010 (approved):

fyodorovandrei:
@Coly010 Why is a lot of duplicated code?

Coly010:
There's no good common place that can be used here and there minor differences between react and angular that need to be accounted for. This is fine as is now

jaysoo (approved):

PR #16491, opened 2023-04-24
fix(node): use consistent default port

philipjfulcher:
## Current Behavior
The Node application generator has two different default ports: 3000 in the `schema.json` and 3333 in the generator itself. This results in a failing e2e test after generation.

## Expected Behavior
The default port is consistent.


nartc:
Some frontend generators use port `3333` to setup proxy conf if consumers pass in `--backendProject`. Can we change the default in `schema.json` to `3333` instead?

jaysoo (approved):

PR #16509, opened 2023-04-24
fix(nextjs): do not generate test for appDir page.tsx

barbados-clemens:
## Current Behavior

when using the new appDir apps are generated with a spec file pointing to the old pages dir

## Expected Behavior

tests are not generated for appDir since page.tsx is an async server side component.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/next/src/generators/application/application.spec.ts
+++ b/packages/next/src/generators/application/application.spec.ts
@@ -452,12 +452,12 @@ describe('app', () => {
 
       expect(tree.exists('apps/testApp/pages/styles.css')).toBeFalsy();
 
-      expect(tree.exists('apps/testApp/app/global.css'));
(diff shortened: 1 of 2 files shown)


packages/next/src/generators/application/files/app/page.tsx__tmpl__
```diff
@@ -14,7 +14,7 @@
 const StyledPage = styled.div`<%- pageStyleContent %>`;
 <% }%>
 
-export default async function Index() {
+export default function Index() {
```
ndcunningham, line 17: Let's add back the `async` since it is a server component. 
Users would be already setup to fetch data.

ndcunningham (changes requested):

ndcunningham (approved):

PR #16547, opened 2023-04-25
fix(core): do not replace legacy package mentions in binary files

FrozenPandaz:
## Current Behavior

Mentions of old packages are replaced in binary files such as images and `.jar` files.

## Expected Behavior

Mentions of old packages are not replaced in binary files such as images and `.jar` files.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/utils/binary-extensions.ts
+++ b/packages/devkit/src/utils/binary-extensions.ts
@@ -1,0 +1,48 @@
+import { extname } from 'path';
+
+const binaryExtensions = new Set([
+  // // Image types originally from https://github.com/sindresorhus/image-type/blob/5541b6a/index.js
+  '.jpg',
+  '.jpeg',
+  '.png',
+  '.gif',
+  '.webp',
+  '.flif',
+  '.cr2',
+  '.tif',
+  '.bmp',
+  '.jxr',
+  '.psd',
+  '.ico',
+  '.bpg',
+  '.jp2',
+  '.jpm',
+  '.jpx',
+  '.heic',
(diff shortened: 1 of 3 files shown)


AgentEnder (approved):

packages/devkit/src/utils/binary-extensions.ts
```diff
@@ -0,0 +1,39 @@
...
+  '.jxr',
+  '.psd',
+  '.ico',
+  '.bpg',
+  '.jp2',
+  '.jpm',
+  '.jpx',
+  '.heic',
+  '.cur',
+  '.tgz',
```
jaysoo, line 23: `.avif` and `.dcm` extensions is missing for images

packages/devkit/src/utils/binary-extensions.ts
```diff
@@ -0,0 +1,39 @@
...
+  // Font files
+  '.ttf',
+  '.otf',
+  '.woff',
+  '.woff2',
+  '.eot',
+]);
+
+export function isBinaryFile(path: string) {
+  return binaryExtensions.has(extname(path));
```
jaysoo, line 38: Do we want to list out more extensions like `.pdf`, `.gz`, `.zip`, etc.? 

There's a long list here: https://github.com/sindresorhus/binary-extensions/blob/main/binary-extensions.json

FrozenPandaz: I added these to the list.. let's see if we want to import this list differently in the future. :thinking:

PR #16576, opened 2023-04-26
fix(core): keep all explicit dependencies in the graph

### LINKED ISSUE #15964: Cache is not cleared when external package is removed or version changed
### Current Behavior

**A build uses the cache and passes** after a dependency package is removed from the project's root `package.json` via npm

### Expected Behavior

**A build does not use the cache and fails** after a dependency package is removed from the project's root `package.json` via npm

### GitHub Repo

https://github.com/jepstein-eagle/package-test

### Steps to Reproduce

1. `npx nx build buildable` (seed cache)
2. `npm rm date-fns` (remove dependent package used in `buildable/src/lib/date/date.component.ts`)
3.  `npx nx build buildable` (uses cache when should not)

### Nx Report

```shell
>  NX  Falling back to ts-node for local typescript execution. This may be a little slower.
  - To fix this, ensure @swc-node/register and @swc/core have been installed

   Node : 16.19.0
   OS   : darwin x64
   npm  : 8.19.3

   nx                      : 15.8.9
   @nrwl/js                : 15.8.9
   @nrwl/jest              : 15.8.9
   @nrwl/linter            : 15.8.9
   @nrwl/workspace         : 15.8.9
   @nrwl/angular           : 15.8.9
   @nrwl/cli               : 15.8.9
   @nrwl/cypress           : 15.8.9
   @nrwl/devkit            : 15.8.9
   @nrwl/eslint-plugin-nx  : 15.8.9
   @nrwl/tao               : 15.8.9
   @nrwl/webpack           : 15.8.9
```

### Failure Logs

_No response_

### Additional Information

This use case is essential to many users because a typical pattern is to change dependent external package versions, and this causes lib builds to pass when they should not if the cache was correctly cleared.

We can use the `implicitDependencies` in the `nx.json` with the following content to resolve:

```
 "implicitDependencies": {
    "package.json": {
      "dependencies": "*"
    }
  }
```

but this should not be required in nx 15 per the [docs](https://nx.dev/deprecated/global-implicit-dependencies#dependencies-on-sections-of-the-root-file) and causes any root `pacakge.json` change, even unrelated to the lib, to require a rebuild.

Thank you in advance for your help 😄

meeroslav:
## Current Behavior
Explicit dependencies not matched to internal or external nodes are ignored during the graph creation.
This leads to:
- inaccurate representation of the actual state
- broken graph when adding missing packages subsequently (see #15964)

## Expected Behavior
All explicit dependencies of a project should be presented in `dependencies` regardless whether we can map it to an actual node or not.

## Methodology

* File dependencies can be created to nodes which do not exist.
* When calculating the updated graph, we omit reporting file dependencies which point to nodes that do not exist so this has no impact on the `ProjectGraph` which is created.
* Once the npm package is installed, recalculating the updated project graph will yield the new dependencies as they can now be resolved.

## Related Issue(s)

Fixes #15964

### DIFF
--- a/e2e/linter/src/linter.test.ts
+++ b/e2e/linter/src/linter.test.ts
@@ -232,7 +232,7 @@ describe('Linter', () => {
         import '@${projScope}/${invalidtaglib}';
         import '@${projScope}/${validtaglib}';
 
-        const s = {loadChildren: '@${projScope}/${lazylib}'};
+        const s = {loadChildren: '@secondScope/${lazylib}'};
       `
         );
 
--- a/packages/nx/src/plugins/js/package-json/create-package-json.ts
+++ b/packages/nx/src/plugins/js/package-json/create-package-json.ts
@@ -234,8 +234,6 @@ function findAllNpmDeps(
           seen,
           dependencyPatterns
         );
-      } else {
-        throw new Error(`Could not find ${dep} in the project graph.`);
       }
     }
   }
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-project-dependencies.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-project-dependencies.ts
@@ -40,19 +40,28 @@ export function buildExplicitTypeScriptDependencies(
             importExpr,
             f.file
           );
+          let targetProjectName;
           if (target) {
             if (!isRoot(source) && isRoot(target)) {
               // TODO: These edges technically should be allowed but we need to figure out how to separate config files out from root
               return;
             }
 
-            res.push({
-              sourceProjectName: source,
-              targetProjectName: target,
-              sourceProjectFile: f.file,
-              type,
-            });
+            targetProjectName = target;
+          } else {
+            // treat all unknowns as npm packages, they can be eiher
+            // - mistyped local import, which has to be fixed manually
+            // - node internals, which should still be tracked as a dependency
+            // - npm packages, which are not yet installed but should be tracked
+            targetProjectName = `npm:${importExpr}`;
           }
+
+          res.push({
+            sourceProjectName: source,
+            targetProjectName,
+            sourceProjectFile: f.file,
+            type,
+          });
         }
       );
     });
--- a/packages/nx/src/project-graph/project-graph-builder.spec.ts
+++ b/packages/nx/src/project-graph/project-graph-builder.spec.ts
@@ -96,13 +96,6 @@ describe('ProjectGraphBuilder', () => {
     expect(() =>
       builder.addExplicitDependency(
         'source',
-        'source/index.ts',
-        'invalid-target'
-      )
-    ).toThrowError();
-    expect(() =>
-      builder.addExplicitDependency(
-        'source',
         'source/invalid-index.ts',
         'target'
       )
--- a/packages/nx/src/project-graph/project-graph-builder.spec.ts
+++ b/packages/nx/src/project-graph/project-graph-builder.spec.ts
@@ -33,6 +33,8 @@ describe('ProjectGraphBuilder', () => {
     expect(() =>
       builder.addImplicitDependency('source', 'invalid-target')
     ).toThrowError();
+    // this should not break, but should not exist in resulting dependencies either
+    builder.addStaticDependency('source', 'invalid-target', 'source/index.ts');
 
     // ignore the self deps
     builder.addDynamicDependency('source', 'source', 'source/index.ts');
--- a/packages/nx/src/project-graph/project-graph-builder.ts
+++ b/packages/nx/src/project-graph/project-graph-builder.ts
@@ -229,20 +229,26 @@ export class ProjectGraphBuilder {
 
       const fileDeps = this.calculateTargetDepsFromFiles(sourceProject);
       for (const [targetProject, types] of fileDeps.entries()) {
-        for (const type of types.values()) {
-          if (
-            !alreadySetTargetProjects.has(targetProject) ||
-            !alreadySetTargetProjects.get(targetProject).has(type)
-          ) {
+        // only add known nodes
+        if (
+          this.graph.nodes[targetProject] ||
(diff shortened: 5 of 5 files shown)


packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-project-dependencies.ts
```diff
@@ -40,19 +40,28 @@ export function buildExplicitTypeScriptDependencies(
...
+            targetProjectName = target;
+          } else {
+            // treat all unknowns as npm packages, they can be eiher
+            // - mistyped local import, which has to be fixed manually
+            // - node internals, which should still be tracked as a dependency
+            // - npm packages, which are not yet installed but should be tracked
+            targetProjectName = `npm:${importExpr}`;
           }
+
+          res.push({
```
meeroslav, line 59: We are attaching all found explicit dependencies to the file.

packages/nx/src/project-graph/project-graph-builder.ts
```diff
@@ -268,7 +274,8 @@ export class ProjectGraphBuilder {
     }
     if (
       !this.graph.nodes[targetProjectName] &&
-      !this.graph.externalNodes[targetProjectName]
+      !this.graph.externalNodes[targetProjectName] &&
+      !sourceProjectFile
```
meeroslav, line 278: Do not break if adding explicit dependency to a file could not be found. We will filter it later in the `getUpdatedProjectGraph` if not resolved by then

e2e/linter/src/linter.test.ts
```diff
@@ -232,7 +232,7 @@ describe('Linter', () => {
         import '@${projScope}/${invalidtaglib}';
         import '@${projScope}/${validtaglib}';
 
-        const s = {loadChildren: '@${projScope}/${lazylib}'};
+        const s = {loadChildren: '@secondScope/${lazylib}'};
```
meeroslav, line 235: This test was broken and it only worked so far because we ignored unknown deps

packages/nx/src/plugins/js/package-json/create-package-json.ts
```diff
@@ -234,8 +234,6 @@ function findAllNpmDeps(
           seen,
           dependencyPatterns
         );
-      } else {
```
meeroslav, line 237: We are collecting deps from files, so some of those might be missing

packages/nx/src/project-graph/project-graph-builder.spec.ts
```diff
@@ -91,13 +93,6 @@ describe('ProjectGraphBuilder', () => {
         'target'
       )
     ).toThrowError();
-    expect(() =>
```
meeroslav, line 94: This no longer breaks, but does not create dependency in `dependencies` either, only in `files`

packages/nx/src/project-graph/project-graph-builder.ts
```diff
@@ -175,41 +175,11 @@ export class ProjectGraphBuilder {
...
-    if (
-      !this.graph.nodes[targetProjectName] &&
-      !this.graph.externalNodes[targetProjectName]
-    ) {
-      throw new Error(`Target project does not exist: ${targetProjectName}`);
-    }
-
-    const fileData = source.data.files.find(
-      (f) => f.file === sourceProjectFile
+    this.addStaticDependency(
```
meeroslav, line 178: The logic was already duplicated in `addDependency` so it's easier to make changes only in one place

packages/nx/src/project-graph/project-graph-builder.ts
FrozenPandaz: This makes sense for this use case. I'm a bit concerned that this would make it somewhat confusing for legitimate mistakes without any feedback.

If dependencies were drawn to projects that did not exist and will never exist because of some legitimate issue, the plugin author would not be told that the dependency was incorrect right? They would just wonder why they don't see it in the graph?

I suppose we could add a flag or a new method which would allow plugin authors to be explicit that they are adding a dependency which they know does not exist... But I'm hesitant to expand the API.

FrozenPandaz (approved):
This looks good to me.. but don't merge it yet.

@vsavkin what do you think about this?

FrozenPandaz (changes requested):
@vsavkin could you also take a look?

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-project-dependencies.ts
```diff
@@ -40,19 +40,28 @@ export function buildExplicitTypeScriptDependencies(
...
-              sourceProjectFile: f.file,
-              type,
-            });
+            targetProjectName = target;
+          } else {
+            // treat all unknowns as npm packages, they can be eiher
+            // - mistyped local import, which has to be fixed manually
+            // - node internals, which should still be tracked as a dependency
+            // - npm packages, which are not yet installed but should be tracked
+            targetProjectName = `npm:${importExpr}`;
```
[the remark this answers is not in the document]
meeroslav: We are simply saying, assume this to be an external (`npm`) package. If it is, this will be resolved once we install it. If it's not, that means it's a local package and will be resolved either by fixing this import or renaming the project (in which case the graph will be fixed).

packages/nx/src/project-graph/project-graph-builder.ts
meeroslav: Please note that `move` is broken - tracking it in a separate issue https://github.com/nrwl/nx/issues/16879

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-project-dependencies.ts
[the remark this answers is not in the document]
meeroslav: This code is only visible to `TS/JS` workspaces so it's safe to assume an `npm` package.

FrozenPandaz (approved):

PR #16586, opened 2023-04-26
fix(vite): build executor should not overwrite package.json in dist

dmitry-stepanenko:
## Current Behavior

Right now if any plugins process/create package.json during the build process, it will be overwritten by `@nrwl/vite:build` executor.

For example, using `rollup-plugin-generate-package-json` plugin, if you put in the app's vite config the following
```
    generatePackageJson({
      inputFolder: './',
      outputFolder: `./dist/apps/myapp`,
    }),
```
it will take no effect if built with nx
## Expected Behavior

Should check for the existing package.json under dist and not overwrite it
## Related Issue(s)

Fixes #


mandarini (approved):
Thank you Dmitry!

Issue #16391, opened 2023-04-19
Error: The module could not be resolved in React native project created with preset npm


adithep:
### Current Behavior

Once app is being compiled there is the following error:

```
error: Error: The module could not be resolved because none of these files exist:
* /Users/---/repo/happynrwl/packages/shared-ui-layout/src/index.ts(.native|.android.js|.native.js|.js|.android.jsx|.native.jsx|.jsx|.android.json|.native.json|.json|.android.ts|.native.ts|.ts|.android.tsx|.native.tsx|.tsx|.android.svg|.native.svg|.svg)
* /Users/---/repo/happynrwl/packages/shared-ui-layout/src/index.ts/index(.native|.android.js|.native.js|.js|.android.jsx|.native.jsx|.jsx|.android.json|.native.json|.json|.android.ts|.native.ts|.ts|.android.tsx|.native.tsx|.tsx|.android.svg|.native.svg|.svg)
  at resolveModulePath (/Users/---/repo/happynrwl/node_modules/metro-resolver/src/resolve.js:134:9)
  at Object.resolve (/Users/---/repo/happynrwl/node_modules/metro-resolver/src/resolve.js:37:12)
  at tsconfigPathsResolver (/Users/---/repo/happynrwl/node_modules/@nrwl/react-native/plugins/metro-resolver.js:84:30)
  at /Users/---/repo/happynrwl/node_modules/@nrwl/react-native/plugins/metro-resolver.js:26:13
  at Object.resolve (/Users/---/repo/happynrwl/node_modules/@react-native-community/cli-plugin-metro/node_modules/metro-resolver/src/resolve.js:27:12)
  at ModuleResolver.resolveDependency (/Users/---/repo/happynrwl/node_modules/@react-native-community/cli-plugin-metro/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:111:31)
  at DependencyGraph.resolveDependency (/Users/---/repo/happynrwl/node_modules/@react-native-community/cli-plugin-metro/node_modules/metro/src/node-haste/DependencyGraph.js:260:43)
  at Object.resolve (/Users/---/repo/happynrwl/node_modules/@react-native-community/cli-plugin-metro/node_modules/metro/src/lib/transformHelpers.js:177:21)
  at Graph._resolveDependencies (/Users/---/repo/happynrwl/node_modules/@react-native-community/cli-plugin-metro/node_modules/metro/src/DeltaBundler/Graph.js:432:35)
[...]

PR #16597, opened 2023-04-27
fix(core): set workspace layout for npm

xiongemi:
## Current Behavior

workspaceLayout is not set for preset.NPM, but set for preset.TS

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/workspace/src/generators/new/generate-workspace-files.ts
+++ b/packages/workspace/src/generators/new/generate-workspace-files.ts
@@ -51,6 +51,10 @@ function setPresetProperty(tree: Tree, options: NormalizedSchema) {
       delete json.targetDefaults;
       delete json.workspaceLayout;
       delete json.npmScope;
+      json.workspaceLayout = {
+        appsDir: 'packages',
+        libsDir: 'packages',
+      };
     }
     return json;
   });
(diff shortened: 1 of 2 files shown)


packages/workspace/src/generators/new/generate-workspace-files.ts
```diff
@@ -51,6 +51,10 @@ function setPresetProperty(tree: Tree, options: NormalizedSchema) {
       delete json.targetDefaults;
       delete json.workspaceLayout;
       delete json.npmScope;
+      json.workspaceLayout = {
```
FrozenPandaz, line 54: This was removed because it should be automatic as long as the directory exists, which it does for the TS preset.

FrozenPandaz (changes requested):

packages/workspace/src/generators/new/generate-workspace-files.ts
xiongemi: i was trying to fix https://github.com/nrwl/nx/issues/16391 about react-native with npm preset. 
`workspaceLayout().libsDir` always returns libs
i realize that for workspaceLayout() function from devkit just returns whatever in nx.json. so i create another pr https://github.com/nrwl/nx/pull/16624/files#diff-414859da125ccd6aa42020f685ec3a2a647b3b20d3333ba73afc71aa02e8700a.

PR #16646, opened 2023-04-28
fix(testing): ensure e2e cypress can use 4200 port in the test

meeroslav:
## Current Behavior
In some cases, port 4200 remains occupied (probably some test does not clean its resources), in which case this test fails as it explicitly expects 4200 to be free.

## Expected Behavior
Ensure port 4200 is freed before we start the test

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/cypress/src/cypress.test.ts
+++ b/e2e/cypress/src/cypress.test.ts
@@ -4,7 +4,7 @@ import {
   cleanupProject,
   createFile,
   ensureCypressInstallation,
-  killPorts,
+  killPort,
   newProject,
   readJson,
   runCLI,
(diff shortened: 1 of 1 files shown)


e2e/cypress/src/cypress.test.ts
```diff
@@ -140,10 +140,13 @@ describe('env vars', () => {
...
 
-    expect(await killPorts(4200)).toBeTruthy();
+    expect(await killPort(4200)).toBeTruthy();
   }, 1000000);
 
-  it('should run e2e in parallel', () => {
+  it('should run e2e in parallel', async () => {
+    // ensure ports are free before running tests
+    await killPort(4200);
+
```
barbados-clemens, line 149: interesting, because the whole point of this feature is that it can handle ports already taken, I guess makes sense to have a known clean slate, but test _ideally_ should not have been failing from that.

barbados-clemens (approved):

PR #16695, opened 2023-05-02
docs(angular): update nx and angular version compatibility matrix with angular 16

leosvelperez:
## Current Behavior

The Nx and Angular Version Compatibility Matrix docs don't have an entry for Angular v16 (it's also missing `~15.2.0`).

## Expected Behavior

The Nx and Angular Version Compatibility Matrix docs have an entry for Angular v16.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/packages/angular/angular-nx-version-matrix.md
+++ b/docs/shared/packages/angular/angular-nx-version-matrix.md
@@ -1,6 +1,6 @@
 # Nx and Angular Versions
 
-The latest version of Nx supports the last two major versions of Angular (current and previous). Workspaces in any of those versions are recommended to use the latest version of Nx to benefit from all the new features and fixes.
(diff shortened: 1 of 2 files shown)


docs/shared/packages/angular/angular-nx-version-matrix.md
```diff
@@ -1,6 +1,6 @@
 # Nx and Angular Versions
 
-The latest version of Nx supports the last two major versions of Angular (current and previous). Workspaces in any of those versions are recommended to use the latest version of Nx to benefit from all the new features and fixes.
+The latest version of Nx supports the [LTS versions of Angular](https://angular.io/guide/releases#actively-supported-versions). Workspaces in any of those versions are recommended to use the latest version of Nx to benefit from all the new features and fixes.
```
dlq84, line 3: Doesn't this imply that Nx does NOT support the current version "a.k.a Active" (16 as of now). 16 is not going to be LTS until 17 is released. Is this the intention, to only support it while in LTS stage?

leosvelperez: Good catch, thanks!

Coly010 (approved):

PR #16703, opened 2023-05-02
fix(expo): expo generator on windows

deroude:
## Current Behavior
Running `npx create-nx-workspace --preset=expo` or `nx g @nx/expo:app my-app` on Windows (Node 16) now breaks with the error 'Unexpected token <some char> in JSON at position <some position>`

The reason is that normalize-options for expo uses node's `path.join`, as opposed to most of the other generators that use `@nx/devkit`.

The following Node script:

```
const devkit = require('@nx/devkit')
const path = require('path')

console.log('DEVKIT', devkit.joinPathFragments('apps','my-app'))
console.log('PATH', path.join('apps','my-app'))
```

will output:

```
DEVKIT apps/my-app
PATH apps\my-app
```

The backslash is breaking the `package.json` and causing `JSON.parse` to throw the error. 

## Expected Behavior

The expo generator should work :)

## Related Issue(s)

Fixes #

The fix is to use `devkit.joinPathFragments` instead of `path.join`

### DIFF
--- a/packages/expo/src/generators/application/lib/normalize-options.ts
+++ b/packages/expo/src/generators/application/lib/normalize-options.ts
@@ -1,5 +1,4 @@
-import { getWorkspaceLayout, names, Tree } from '@nx/devkit';
-import { join } from 'path';
+import { getWorkspaceLayout, joinPathFragments, names, Tree } from '@nx/devkit';
 import { Schema } from '../schema';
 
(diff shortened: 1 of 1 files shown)


xiongemi (approved):

xiongemi:
@deroude the pr looks good. could you squash your commit and fix the commit message?

PR #16710, opened 2023-05-02
feat(core): add skipTypeCheck option to rollup plugin options

marwan38:
## Current Behavior
Can't disable the rollup typescript plugin type checker

## Expected Behavior
Allow users to disable it if needed

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/rollup/src/executors/rollup/lib/normalize.ts
+++ b/packages/rollup/src/executors/rollup/lib/normalize.ts
@@ -44,6 +44,7 @@ export function normalizeRollupExecutorOptions(
     project,
     projectRoot,
     outputPath,
+    skipTypeCheck: options.skipTypeCheck || false,
   };
 }
 
--- a/docs/generated/packages/rollup/executors/rollup.json
+++ b/docs/generated/packages/rollup/executors/rollup.json
@@ -159,6 +159,11 @@
         "type": "boolean",
         "description": "Generate package.json with 'exports' field. This field defines entry points in the package and is used by Node and the TypeScript compiler.",
         "default": false
+      },
+      "skipTypeCheck": {
+        "type": "boolean",
+        "description": "Whether to skip TypeScript type checking.",
+        "default": false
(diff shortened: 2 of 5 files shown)


packages/rollup/src/executors/rollup/lib/normalize.ts
```diff
@@ -44,6 +44,7 @@ export function normalizeRollupExecutorOptions(
     project,
     projectRoot,
     outputPath,
+    skipTypeCheck: options.skipTypeCheck || false,
```
marwan38, line 47: Staying consistent with the swc compiler option

mandarini (approved):

mandarini (changes requested):
Can you add this extra option to the `schema.json` as well, and then run `pnpm documentation`? Thanks!

marwan38:
@mandarini Thanks for getting back to me so quickly. I've added it to the schema and generated the docs. I copied over the exact schema definition from esbuild and swc. No idea what `x-priority` is, but, if both esbuild and swc have it set to `important` I thought that's what I should be doing.

mandarini (changes requested):
Hi there @marwan38 ! Thanks for updating the code! Can you please squash the commits and remove the `x-priority` attribute? We add this as `important` usually for required values or values we want to bring attention to, so that they are pushed at the top of the executor options list. The new option you added will still show in the list, just not at the top!

I tried to make these changes myself to your repo, but I did not have permission to push :)

marwan38:
@mandarini all done. thanks again for walking me through it :)

mandarini (approved):
All is great now! Thanks! :) However, unfortunately, a new PR got merged and there are some conflicts with your branch, so you need to rebase with latest `master`. I would do it for you, but I don't have permission to push to your fork :( You can "allow edits by maintainers" in your fork, if you want!

marwan38:
@mandarini Done!

PR #16738, opened 2023-05-03
Update test-examples.md

mikhin:


mandarini:
@mikhin thank you for the PR! Can you please fix the commit message (run `pnpm commit` to help you write the commit) and then run `pnpm documentation` to update the generated docs as well?

mandarini (changes requested):
@mikhin thank you for the PR! Can you please fix the commit message (run `pnpm commit` to help you write the commit) and then run `pnpm documentation` to update the generated docs as well?

mandarini:
@mikhin hey sorry just realized, this PR is wrong! The example is actually correct. It means that "create a new configuration that is called `watch` and in there add the `watch: true` setting".

PR #16739, opened 2023-05-03
docs(core): add example of custom/ejected github workflow

meeroslav:
## Current Behavior
Our docs suggest using `nrwl/ci` as a default GitHub workflow leading to users believe this is the only way and opening issues with requests we are not willing (or cannot, due to reusable workflow limitations) implement.

## Expected Behavior
We should also have an example of a custom workflow, like we do for other CI providers.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/monorepo-ci-github-actions.md
+++ b/docs/shared/monorepo-ci-github-actions.md
@@ -139,11 +139,17 @@ jobs:
           # list of commands to be run on main has env flag NX_CLOUD_DISTRIBUTED_EXECUTION set to false
           NX_CLOUD_DISTRIBUTED_EXECUTION=false npx nx-cloud record -- npx nx format:check & pids+=($!)
 
-          # add other parallel commands here
-          npx nx affected -t lint --parallel=3 & pids+=($!)
-          npx nx affected -t test --parallel=3 --configuration=ci & pids+=($!)
-          npx nx affected -t build --parallel=3 & pids+=($!)
+          # list of commands to be run on agents
+          npx nx affected -t lint --parallel=3 & 
+          pids+=($!)
 
+          npx nx affected -t test --parallel=3 --configuration=ci & 
+          pids+=($!)
+
+          npx nx affected -t build --parallel=3 & 
+          pids+=($!)
+
+          # run all commands in parallel and bail if one of them fails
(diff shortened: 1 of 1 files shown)


docs/shared/monorepo-ci-github-actions.md
```diff
@@ -71,4 +71,130 @@ jobs:
...
+          pids=()
+          # list of commands to be run on main has env flag NX_CLOUD_DISTRIBUTED_EXECUTION set to false
+          NX_CLOUD_DISTRIBUTED_EXECUTION=false npx nx-cloud record -- npx nx format:check & pids+=($!)
+
+          # add other parallel commands here
+          npx nx affected -t lint --parallel=3 & pids+=($!)
+          npx nx affected -t test --parallel=3 --configuration=ci & pids+=($!)
+          npx nx affected -t build --parallel=3 & pids+=($!)
+
+          for pid in \${pids[*]}; do
```
isaacplmann, line 147: what is this section doing?

docs/shared/monorepo-ci-github-actions.md
```diff
@@ -71,4 +71,130 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha }}
+          # We need to fetch all branches and commits so that Nx affected has a base to compare against.
+          fetch-depth: 0
+
+      - uses: actions/checkout@v3
+        name: Checkout [Default Branch]
+        if: ${{ github.event_name != 'pull_request' }}
+        with:
+          # We need to fetch all branches and commits so that Nx affected has a base to compare against.
+          fetch-depth: 0
```
isaacplmann, line 115: This is great.  I've gotten stuck on this part multiple times trying to set up CI.

docs/shared/monorepo-ci-github-actions.md
isaacplmann: it would be nice to add a comment here

isaacplmann (changes requested):

docs/shared/monorepo-ci-github-actions.md
meeroslav: Added! Please re-review

isaacplmann (approved):

PR #16749, opened 2023-05-03
fix(react-native): fix buildable react native library

xiongemi:
## Current Behavior

currently react native and expo library uses rollup to build libraries; however, it does not install @nx/rollup when adding a buildable library.

## Expected Behavior

when adding a buildable library, it should add @nx/rollup to devDependencies.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/guides/react-native.md
+++ b/docs/shared/guides/react-native.md
@@ -1,7 +1,5 @@
 # React Native with Nx
 
-![React Logo](/shared/react-logo.png)
-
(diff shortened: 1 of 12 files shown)


jaysoo (approved):

jaysoo:
LGTM

Future: Let's look at Vite for the buildable libraries as well since we support it for React/Web.

bcabanes (approved):

PR #16758, opened 2023-05-04
fix(angular): ssr distFolder path should be more correctly defined

Coly010:
## Current Behavior

When adding SSR to an Angular App, we rewrite the output path to match integrated monorepo style directory structure. We then assume this value to follow that structure in the template files.

## Expected Behavior

We should only modify the output path of the build to add what we need, `/browser` and then pull the value stored in the project's config to use as the path in the template files.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/setup-ssr/lib/update-project-config.ts
+++ b/packages/angular/src/generators/setup-ssr/lib/update-project-config.ts
@@ -15,8 +15,8 @@ import type { Schema } from '../schema';
 export function updateProjectConfig(tree: Tree, schema: Schema) {
   let projectConfig = readProjectConfiguration(tree, schema.project);
   const buildTarget = projectConfig.targets.build;
-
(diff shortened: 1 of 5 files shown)


packages/angular/src/generators/setup-ssr/lib/update-project-config.ts
```diff
@@ -15,8 +15,8 @@ import type { Schema } from '../schema';
 export function updateProjectConfig(tree: Tree, schema: Schema) {
   let projectConfig = readProjectConfiguration(tree, schema.project);
   const buildTarget = projectConfig.targets.build;
-
-  buildTarget.options.outputPath = `dist/apps/${schema.project}/browser`;
+  const baseOutputPath = `${buildTarget.options.outputPath}`;
```
leosvelperez, line 18: nit:
```suggestion
  const baseOutputPath = buildTarget.options.outputPath;
```

leosvelperez (approved):
Just left a non-blocking suggestion.

PR #16762, opened 2023-05-04
fix(core): make sure verbose logging is not affecting the cache output in e2e test

meeroslav:
This is one of the three failing suites on nightly

## Current Behavior
Verbose logging flag added on nightly changes the expected output and causes the test to fail

## Expected Behavior
We should be aware of the verbose output when checking for matches

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/nx-run/src/cache.test.ts
+++ b/e2e/nx-run/src/cache.test.ts
@@ -1,5 +1,6 @@
 import {
   cleanupProject,
+  isVerboseE2ERun,
   listFiles,
   newProject,
   readFile,
@@ -61,7 +62,7 @@ describe('cache', () => {
       'read the output from the cache'
     );
 
-    if (process.platform != 'linux') {
+    if (process.platform != 'linux' && !isVerboseE2ERun()) {
       // TODO(vsavkin): This should be always be matched output once you fix output watching on linux
       expectMatchedOutput(outputWithBuildApp2Cached, [myapp2]);
     } else {


e2e/nx-run/src/cache.test.ts
```diff
@@ -61,7 +62,7 @@ describe('cache', () => {
       'read the output from the cache'
     );
 
-    if (process.platform != 'linux') {
+    if (process.platform != 'linux' && !isVerboseE2ERun()) {
```
meeroslav, line 65: Verbose logging is changing the terminal output leading to failed test

AgentEnder (approved):

e2e/nx-run/src/cache.test.ts
FrozenPandaz: So `expectMatchedOutput` doesn't work with the verbose logging?

Shouldn't we fix this test so that it works even with verbose logging?

FrozenPandaz (changes requested):
We shouldn't disable this expectation

e2e/nx-run/src/cache.test.ts
meeroslav: Not sure what you mean by doesn't work. The output works, but because of the `--verbose` we are not printing out compact view (containing expected `read the output from the cache`) but full output which has `local cache` next to tasks that were retrieved from the cache.

Code works as expected, but test does not take into account different environment in which we run it.

meeroslav:
Closing this PR since the issue is elsewhere. This should resolve it https://github.com/nrwl/nx/pull/16808 hopefully

PR #16767, opened 2023-05-04
fix(misc): restore using legacy-peer-deps for npm installs

FrozenPandaz:
## Current Behavior

Conflicting `peerDependencies` cause any `npm` installs to fail. 

## Expected Behavior

 `npm` installs are run with `legacy-peer-deps`. Users will still incur an error if they manually run `install` which they will need to pass the flag manually.. or add a `.npmrc` to their workspace.

## Related Issue(s)

Fixes #


packages/nx/src/utils/package-manager.ts
```diff
@@ -88,6 +88,9 @@ export function getPackageManagerCommand(
       };
     },
     npm: () => {
+      // TODO: Remove this
```
barbados-clemens, line 91: is this just temporary or is this supposed to be removed in future versions of nx?

barbados-clemens (approved):

AgentEnder (approved):

PR #16786, opened 2023-05-04
fix(react): remove outdated packages

### LINKED ISSUE #11882: Potentially redundant dependencies in Next.js template

## Description
When generating a simple Next.js project with `npx create-nx-workspace --preset=next`, I see at least three dependencies where I don't know why they come installed out of the box:

```
"core-js": "^3.6.5",
"regenerator-runtime": "0.13.7",
"react-test-renderer": "18.0.0"
```

`core-js` and `regenerator-runtime` seem to be compatibility layers which shouldn't be necessary in a modern project.

jest tests also work without `react-test-renderer`.

All three of these dependencies are never references in a freshly installed project.

## Motivation
Get rid of redundant dependencies

ndcunningham:
The features should be available by default or they are available via a different library.
Example
`react-test-renderer` -> `@testing-library`

closed #11882

### DIFF
--- a/packages/react/src/generators/init/init.ts
+++ b/packages/react/src/generators/init/init.ts
@@ -16,7 +16,6 @@ import {
   babelPresetReactVersion,
   nxVersion,
   reactDomVersion,
-  reactTestRendererVersion,
   reactVersion,
   testingLibraryReactVersion,
   tsLibVersion,
@@ -63,7 +62,6 @@ function updateDependencies(host: Tree, schema: InitSchema) {
     '@types/react': typesReactVersion,
     '@types/react-dom': typesReactDomVersion,
     '@testing-library/react': testingLibraryReactVersion,
-    'react-test-renderer': reactTestRendererVersion,
   });
 }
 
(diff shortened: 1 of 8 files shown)


packages/react/src/generators/init/init.ts
```diff
@@ -63,7 +63,6 @@ function updateDependencies(host: Tree, schema: InitSchema) {
     '@types/react': typesReactVersion,
     '@types/react-dom': typesReactDomVersion,
     '@testing-library/react': testingLibraryReactVersion,
-    'react-test-renderer': reactTestRendererVersion,
```
xiongemi, line 66: is reactTestRendererVersion still being used?

xiongemi:
do we need a migration script to remove these libraries in package.json?

packages/react/src/generators/init/init.ts
ndcunningham: I don't think so.

jaysoo (approved):

PR #16806, opened 2023-05-05
feat(angular): add ngrx feature store generator

Coly010:
## Current Behavior

We have one NgRx Generator that currently tries to handle two different use cases and as such has come conflicting logic and a lot of branching to determine which use case to run.

## Expected Behavior

We should have two separate generators that focus purely on each use case:
1. Add NgRx Root Setup to an application
2. Add an NgRx Feature Store to an application or library

This PR addresses 2.

There are cases when there should also be a root state, so the root state generator can then invoke this generator for that case.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/manifests/menus.json
+++ b/docs/generated/manifests/menus.json
@@ -4015,6 +4015,14 @@
                 "disableCollapsible": false
               },
               {
+                "id": "ngrx-feature-store",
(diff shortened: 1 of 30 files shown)


AgentEnder (approved):
Signing off on the generated docs changes

leosvelperez (approved):

PR #16811, opened 2023-05-05
feat(angular): add ngrx root store generator

Coly010:
## Current Behavior

Part 2 of the NgRx Generator Refactor

This will focus on generating root state for an application only

### DIFF
--- a/packages/angular/src/generators/ngrx-root-store/lib/validate-options.ts
+++ b/packages/angular/src/generators/ngrx-root-store/lib/validate-options.ts
@@ -1,0 +1,53 @@
+import type { Tree } from '@nx/devkit';
+import { getProjects, readProjectConfiguration } from '@nx/devkit';
+import { Schema } from '../schema';
+import {
+  getInstalledAngularVersionInfo,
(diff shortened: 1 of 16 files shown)


packages/angular/src/generators/ngrx-root-store/lib/validate-options.ts
```diff
@@ -0,0 +1,57 @@
...
+  getInstalledPackageVersionInfo,
+} from '../../utils/version-utils';
+import { getPkgVersionForAngularMajorVersion } from '../../../utils/version-utils';
+import { coerce, lt, major } from 'semver';
+import { isNgStandaloneApp } from '../../../utils/nx-devkit/ast-utils';
+
+export function validateOptions(tree: Tree, options: Schema): void {
+  if (!options.project) {
+    throw new Error('Please provide a value --project');
+  }
```
leosvelperez, line 15: nit: we don't need this check. The schema and the type definition require the `project` option.

leosvelperez (approved):

jaysoo (approved):

PR #16858, opened 2023-05-08
feat(linter): allow ESLint Flat Config to be opted into manually

JamesHenry:
## Current Behavior

ESLint's new (not yet default) so called "Flat Config" cannot be used with Nx workspaces.

## Expected Behavior

Users can manually opt into using ESLint's Flat Config style in their Nx workspaces prior to it being the official primary configuration style in ESLint v9 later in the year.

Appropriate migrations and generators will follow on from this once we get much closer to Flat Config being the primary default in ESLint.

https://eslint.org/docs/latest/use/configure/configuration-files-new

P.S. I also took the opportunity to clean up how ESLint is resolved and instantiated, it had become organically a bit messy over time

### DIFF
--- a/packages/linter/src/executors/eslint/utility/eslint-utils.ts
+++ b/packages/linter/src/executors/eslint/utility/eslint-utils.ts
@@ -1,36 +1,43 @@
-import { ESLint } from 'eslint';
+import type { ESLint } from 'eslint';
 import type { Schema } from '../schema';
 
-export async function loadESLint() {
-  let eslint;
+async function resolveESLintClass(
+  useFlatConfig = false
+): Promise<typeof ESLint> {
   try {
-    eslint = await import('eslint');
-    return eslint;
+    if (!useFlatConfig) {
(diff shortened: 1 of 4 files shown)


packages/linter/src/executors/eslint/utility/eslint-utils.ts
```diff
@@ -1,50 +1,90 @@
...
-  options: Schema
-): Promise<ESLint.LintResult[]> {
-  const projectESLint: { ESLint: typeof ESLint } = await loadESLint();
+  options: Schema,
+  useFlatConfig = false
+) {
+  if (
+    useFlatConfig &&
+    eslintConfigPath &&
+    !eslintConfigPath?.endsWith('eslint.config.js')
```
meeroslav, line 27: Nit: The `optional chaining` on like 27 is not needed, since we check for `eslintConfigPath`'s existence on line before

meeroslav (approved):

PR #16880, opened 2023-05-09
chore(js): set unitTestRunner default to none for @nx/js:library to reflect actual behavior

MaxKless:
## Current Behavior
`@nx/js:library`'s `unitTestRunner` doesn't have a default value even though it actually just defaults to `none`. 
In Nx Console, for example, this results in duplicate `empty` and `none` selection options.

## Expected Behavior
Default value should reflect actual behavior

### DIFF
--- a/docs/generated/packages/js/generators/library.json
+++ b/docs/generated/packages/js/generators/library.json
@@ -31,7 +31,8 @@
         "type": "string",
         "enum": ["jest", "vitest", "none"],
         "description": "Test runner to use for unit tests.",
-        "x-prompt": "Which unit test runner would you like to use?"
+        "x-prompt": "Which unit test runner would you like to use?",
+        "default": "none"
       },
       "tags": {
         "type": "string",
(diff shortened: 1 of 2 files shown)


Coly010 (approved):

AgentEnder (changes requested):
I don't think this is accurate - looking at the code for @nx/js:library there is this series of control flow checks in `normalizeOptions`:

```typescript
  if (options.config === 'npm-scripts') {
    options.unitTestRunner = 'none';
    options.linter = Linter.None;
    options.bundler = 'none';
  }
 if (!options.unitTestRunner && options.bundler === 'vite') {
    options.unitTestRunner = 'vitest';
  } else if (!options.unitTestRunner && options.config !== 'npm-scripts') {
    options.unitTestRunner = 'jest';
  }
```

If we set the default value to "none" here, new libraries would not be initialized with Jest or Vitest depending on the bundler, and I don't think thats a desired functional change.

To be clear `none`, and `{blank}` are functionally different here. When left blank, the test runner is derived from the other settings. When set to `none`, no unit test config is ever generated.

MaxKless:
thanks so much for the review, I missed that part. 
I wonder if there's a way to communicate the `blank` case more clearly - because it's essentially `'unitTestRunner" : "auto"` here, right?

PR #16896, opened 2023-05-09
feat(nextjs): Use next.js cli for build and serve targets

ndcunningham:
This change is to refactor `@nx/next` `executors` from using deeply imported `next.js` build functionality.
Now we will integrate the use of `next.js` CLI which should help bridge the gap between `nx` & `next` commands.

There should be no change to how the commands work for existing users.

### DIFF
--- a/packages/next/src/executors/build/build.impl.ts
+++ b/packages/next/src/executors/build/build.impl.ts
@@ -4,9 +4,9 @@ import {
   readJsonFile,
   workspaceRoot,
   writeJsonFile,
+  logger,
 } from '@nx/devkit';
 import { createLockFile, createPackageJson, getLockFileName } from '@nx/js';
-import build from 'next/dist/build';
 import { join, resolve } from 'path';
 import { copySync, existsSync, mkdir, writeFileSync } from 'fs-extra';
 import { lt, gte } from 'semver';
@@ -17,6 +17,8 @@ import { updatePackageJson } from './lib/update-package-json';
 import { createNextConfigFile } from './lib/create-next-config-file';
 import { checkPublicDirectory } from './lib/check-project';
 import { NextBuildBuilderOptions } from '../../utils/types';
+import { ExecSyncOptions, execSync } from 'child_process';
+import { createCliOptions } from '../../utils/create-cli-options';
 
 export default async function buildExecutor(
   options: NextBuildBuilderOptions,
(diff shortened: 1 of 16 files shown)


jaysoo (approved):

packages/next/src/executors/build/build.impl.ts
```diff
@@ -45,19 +47,21 @@ export default async function buildExecutor(
...
-  // Check the major and minor version numbers
-  if (lt(nextVersion, '13.2.0')) {
-    // If the version is lower than 13.2.0, use the second parameter as the config object
-    await build(root, null, false, debug);
-  } else {
-    // Otherwise, use the third parameter as a boolean flag for verbose logging
-    // @ts-ignore
-    await build(root, false, debug);
+  const args = formatObjectToCli({ experimentalAppOnly, profile, debug });
+  const command = `next build ${args}`;
```
jaysoo, line 53: Needs to be `npx next` to run the locally installed package.

packages/next/src/executors/server/server.impl.ts
```diff
@@ -38,96 +37,24 @@ export default async function* serveExecutor(
...
+  const nextDir = resolve(context.root, buildOptions.outputPath);
 
-  const customServerBuild = await runExecutor(
-    parseTargetString(options.customServerTarget, context.projectGraph),
-    {
-      watch: options.dev ? true : false,
-    },
-    context
-  );
+  const command = `next ${
```
jaysoo, line 45: Needs to be `npx next` to run the locally installed package.

packages/next/src/executors/server/formatObjectToCli.ts
```diff
@@ -0,0 +1,12 @@
+export function formatObjectToCli(obj: { [key: string]: any }): string {
```
jaysoo, line 1: Make sure the file name is using dashes not camelcase.

packages/next/src/executors/export/export.impl.ts
```diff
@@ -25,6 +25,18 @@ import * as chalk from 'chalk';
...
+ * const nextConfig = {
+  nx: {
+    svgr: false,
+  },
+
+  output: 'export'
+};
+ * Read https://nextjs.org/docs/pages/building-your-application/deploying/static-exports
+ **/
 export default async function exportExecutor(
```
jaysoo, line 40: Can we mark this as `x-deprecated: "..."`  in `executors.json` as well?

Issue #15116, opened 2023-02-19
Changes to package-lock.json trigger an Affected on all packages


pedrolourencoribeiro:
### Current Behavior

Update one dependency that affects only one package
run ` npx nx print-affected --select=projects --base=main` 
will print all packages

### Expected Behavior

Update one dependency that affects only one package
run ` npx nx print-affected --select=projects --base=main` 
should print only the affected package

### GitHub Repo

https://github.com/pedrolourencoribeiro/nx-affected

### Steps to Reproduce

1. clone git repo https://github.com/pedrolourencoribeiro/nx-affected
2. checkout zod-update
3. install dependencies (`npm i`)
4. run ` npx nx print-affected --select=projects --base=main`

### Nx Report

```shell
>  NX   Report complete - copy this into the issue template

   Node : 16.18.0
   OS   : darwin arm64
   npm  : 8.19.2
   
   nx                      : 15.7.2
   @nrwl/linter            : 15.7.2
   @nrwl/workspace         : 15.7.2
   @nrwl/cli               : 15.7.2
   @nrwl/devkit            : 15.7.2
   @nrwl/eslint-plugin-nx  : 15.7.2
   @nrwl/js                : 15.7.2
   @nrwl/tao               : 15.7.2
   typescript              : 4.8.4
```

### Additional Information

This was changed in the version 15.4.0 => https://github.com/nrwl/nx/pull/13960#discussion_r1055559215
I don't fully know the reason but this is blocking us from upgrading to a new version since we have quite a lot of packages and we want to know what is really affected when updating a dependency.
Is there a workaround for this? From looking at the code nothing came to my mind but maybe there is.

Thank you for all the work btw the project is amazing and its saving us a lot of time 🙇‍♂️
Ps: I'm available to contribute to this if needed :)

ianldgs:
Hi @AgentEnder / @meeroslav. Sorry to bother again about this, but since we like to release affected projects often, this issue causes our CI to be very busy whenever someone adds a dependency to a local package.json.

Would you consider accepting the changes of https://github.com/nrwl/nx/pull/15776 under a feature flag? (Until a final solution is implemented).
That way, anyone can opt out of this behaviour at their own risk.

brandoncollins7:
I am having the same issue, this is a big problem:(

meeroslav:
Hey all,

The logic behind this is not as trivial as you would have expected.

## Explanation

[...]

PR #16916, opened 2023-05-10
feat(core): add commandExternalDependencies hash input

meeroslav:
Follow up to [#16903](https://github.com/nrwl/nx/pull/16903)

This PR implements new `commandExternalDependencies` input

## Related Issue(s)

Solution for #15116

### DIFF
--- a/docs/shared/reference/project-configuration.md
+++ b/docs/shared/reference/project-configuration.md
@@ -158,6 +158,17 @@ Examples:
 
 Note the result value is hashed, so it is never displayed.
 
+_External Dependencies_
+
+Examples:
+
+- `{externalDependencies: []}`
+  Specify that a target executor is not affected by any changes in installed packages
+- `{externalDependencies: ["lerna"]}`
+  Specify exact packages that affect a target executor
+
+Note the result value is hashed, so it is never displayed. This override is ignored by `@nx/*` executors.
+
 _Named Inputs_
 
 Examples:
--- a/docs/shared/reference/project-configuration.md
+++ b/docs/shared/reference/project-configuration.md
@@ -160,16 +160,47 @@ Note the result value is hashed, so it is never displayed.
 
 _External Dependencies_
 
-This input is useful for ambiguous executors like `nx:run-commands`.
+For official plugins, Nx intelligently finds a set of external dependencies which it hashes for the target. `nx:run-commands` is an exception to this.
+Because you may specify any command to be run, it is not possible to determine which, if any, external dependencies are used by the target.
+To be safe, Nx assumes that updating any external dependency invalidates the cache for the target.
+
+> Note: Community plugins are also treated like `nx:run-commands`
+
+This input type allows for you to override this cautious behavior by specifying a set of external dependencies to hash for the target.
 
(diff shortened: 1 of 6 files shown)


docs/shared/reference/project-configuration.md
```diff
@@ -158,6 +158,19 @@ Examples:
...
+This input is useful for ambiguous executors like `nx:run-commands`.
+
+Examples:
+
+- Specify that a target executor is not affected by any changes in installed packages
+- `{externalDependencies: []}`
+- Specify exact packages that affect a target executor
+- `{externalDependencies: ["lerna"]}`
+
+Note the result value is hashed, so it is never displayed.
```
isaacplmann, line 172: I don't know what this means

docs/shared/reference/project-configuration.md
```diff
@@ -158,6 +158,19 @@ Examples:
 
 Note the result value is hashed, so it is never displayed.
 
+_External Dependencies_
+
+This input is useful for ambiguous executors like `nx:run-commands`.
```
isaacplmann, line 163: This needs more explanation.  Something like

```suggestion
This input is useful for executors that do not have code directly associated with them, like `nx:run-commands`.
```

isaacplmann (changes requested):
Needs clearer documentation

docs/shared/reference/project-configuration.md
isaacplmann: Maybe that's not the right explanation, but whatever the right explanation is needs to go here.

docs/shared/reference/project-configuration.md
```diff
@@ -158,6 +158,19 @@ Examples:
...
+
+Examples:
+
+- Specify that a target executor is not affected by any changes in installed packages
+- `{externalDependencies: []}`
```
FrozenPandaz:
```suggestion
- `{externalDependencies: []}`
  - By default, to err on the side of caution, Specify that a target executor is not affected by any changes in installed packages
```

docs/shared/reference/project-configuration.md
FrozenPandaz:
```suggestion
This input type should only be used with executors which run arbitrary commands/scripts without having a way of determining which, if any, external dependencies are associated with the target. A canonical example is the `nx:run-commands`. Because any command can be provided as an option, it is not possible to tell what the command does or what external dependencies it may use.

> Note: By default, Nx will hash _all_ external dependencies for targets that use `nx:run-commmands` and other unknown executors. This is the safest option because it ensures that any external dependencies are captured. This may lead to unexpected cache misses when updating dependencies. This option is a means to override this behavior and allow for more cache hits by providing exactly which dependencies are used by the target.

Examples:
- Specify that a target executor is not affected by any changes in installed packages
- `{externalDependencies: []}`
- Specify exact packages that affect a target executor
- `{externalDependencies: ["lerna"]}`

Note the result value is hashed, so it is never displayed.
```

isaacplmann (approved):

FrozenPandaz (approved):

Issue #13248, opened 2022-11-18
@nrwl/js:node executor doesn't exit when I using ora package on zsh(mac)


wizardnet972:
### Current Behavior

The process does not exit when I hit command+c to exist from my application.

The problem happens on my mac m1 (zsh).

### Expected Behavior

Stop the process.

### Github Repo

https://github.com/wizardnet972/nx-ora

### Steps to Reproduce

Clone `https://github.com/wizardnet972/nx-ora` repo.

Then run:
```
npm i
npm start (nx serve cli)
```
Press command+c to exit.

Check the processes (run this command): `ps`, and you will see the node process still running. (to exit run `ps -9 #yourprocessnumebr`).

### Nx Report

```shell
yarn run v1.22.19
$ /Users/wizardnet972/repo/nx-ora/node_modules/.bin/nx report

 >  NX   Report complete - copy this into the issue template

   Node : 16.15.1
   OS   : darwin arm64
   yarn : 1.22.19

   nx : 15.2.0
   @nrwl/angular : Not Found
   @nrwl/cypress : Not Found
   @nrwl/detox : Not Found
   @nrwl/devkit : 15.2.0
   @nrwl/esbuild : Not Found
   @nrwl/eslint-plugin-nx : 15.2.0
   @nrwl/expo : Not Found
   @nrwl/express : Not Found
   @nrwl/jest : 15.2.0
   @nrwl/js : 15.2.0
   @nrwl/linter : 15.2.0
   @nrwl/nest : Not Found
   @nrwl/next : Not Found
   @nrwl/node : Not Found
   @nrwl/nx-cloud : Not Found
   @nrwl/nx-plugin : Not Found
   @nrwl/react : Not Found
   @nrwl/react-native : Not Found
   @nrwl/rollup : Not Found
   @nrwl/schematics : Not Found
   @nrwl/storybook : Not Found
   @nrwl/web : Not Found
   @nrwl/webpack : Not Found
   @nrwl/workspace : 15.2.0
   typescript : 4.8.4
   ---------------------------------------
   Local workspace plugins:
   ---------------------------------------
   Community plugins:

✨  Done in 1.10s.
```
```


### Failure Logs

```shell
N/A
```

### Additional Information

1. When I remove `ora` then the process exit. (but I need to use that package).
2. Maybe should check [`tree-kill which nx use`](https://github.com/nrwl/nx/blob/master/packages/js/src/executors/node/node.impl.ts#L11) because it not get the process id.
[...]

Issue #9239, opened 2022-03-08
Serve command doesn't detect nodeJS app process exiting


joshribakoff-sm:
## Current Behavior
Serve a nodeJS program that does `process.exit(1)`. 

It does not output anything and the user has to rely on other observations to deduce that their code exited. In fact editing the code still creates normal webpack output as if the process was reloaded, but the process is dead and is never restarted. The developer has to kill the `serve` command manually and restart it manually to restore the nodeJS app's process.

## Expected Behavior
The `serve` command should output a message saying that my code exited, and what the exit code was. Editing any code should start the process back up [if it was dead]. I expect making an edit to the code and webpack "reloading" to restart a dead nodeJS program automatically.

## Steps to Reproduce
Create a node app that does this
```
process.exit(1)
```

### Environment
Mac OS

This is also a duplicate of https://github.com/nrwl/nx/issues/4054 it appears that issue was closed despite the linked fix being totally tangential to the root issue

hcharley:
When you say:

> Editing any code should start the process back up [if it was dead]

I feel like this should only be true if the `--watch` argument is `true` (which is the default.

In the case where `--watch` is `false`, I think it should notify the user as you say, and then exit.

This will help those of us developing jobs (as opposed to always on servers) to chain subsequent commands/notifications. Right now, it seems like the only choice is to build and run.

joshribakoff-sm:
I would agree 🎉 

Side note: I am in fact writing a job that runs once and exits, but I want to develop with "watch" mode. The use case is to re-run the job on any code change, including re-attempt the job after fixing a crash in a previous job run :)

hcharley:
That's helpful context in your side note @joshribakoff-sm. It made me wonder...

I like that watch auto reloads, but I wish that I could have the option to let the job finish before reloading, or allow for specific Grace periods instead of 'instantly'. 

Maybe this already exists and I've forgotten or don't know about it. 

I think in the past, I've tried to delay on a SIGINT in my code, but what happens (I have to double check) is that the next job starts and then they are both running. 

I probably just could write a better, faster SIGINT cleanup, but it's still my uninformed impression that the grace period is 0 and isn't configurable. 

Anyways, not sure if this wondering is relevant.

hcharley:
BTW, the way I am getting the functionality I need is to use an `expect` script:

```sh
#!/usr/bin/expect -f

set timeout -1

puts "Running with args: $argv\r"

spawn tools/scripts/run-my-app.sh $argv

[...]

PR #16924, opened 2023-05-10
feat(js): improve @nx/js:node executor to be more resilient to many file change events

jaysoo:
This PR provides a new implementation of `@nx/js:node` executor to fix two issues:

1. File changes cause server to not start properly (e.g. port is already in use from old process)
2. Changes to workspace libraries don't cause a rebuild.

**TODO:**

- [x] Support node require overrides that the old executor implements (i.e. buildable lib mapping)
- [x] Cleanup (remove old implementation, etc.)

## What's new
- `@nx/js:node` executor uses Nx Daemon to watch for while changes to app and its dependencies.
- New `--debounce` option to batch file events to avoid unnecessary rebuilds. Defaults to 500 ms.
- Better logging so users know if something errors out or exits unexpectedly.

## Related Issues
https://github.com/nrwl/nx/issues/9239, https://github.com/nrwl/nx/issues/13248

### DIFF
--- a/docs/generated/packages/js/executors/node.json
+++ b/docs/generated/packages/js/executors/node.json
@@ -59,6 +59,11 @@
         "type": "boolean",
         "description": "Enable re-building when files change.",
         "default": true
+      },
+      "debounce": {
+        "type": "number",
+        "description": "Delay in milliseconds to wait before restarting. Useful to batch multiple file changes events together. Set to zero (0) to disable.",
+        "default": 500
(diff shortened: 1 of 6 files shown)


ndcunningham (approved):

mandarini:
Possibly caused:
* https://github.com/nrwl/nx/issues/17451
* https://github.com/nrwl/nx/issues/17429

PR #16928, opened 2023-05-10
feat(js): add verdaccio executor

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/executors/verdaccio/schema.json
+++ b/packages/js/src/executors/verdaccio/schema.json
@@ -1,0 +1,35 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "version": 2,
+  "title": "Verdaccio Local Registry",
+  "description": "Start a local registry with Verdaccio.",
+  "cli": "nx",
+  "type": "object",
+  "properties": {
+    "location": {
+      "type": "string",
+      "description": "Location option for npm config",
+      "default": "user",
+      "enum": ["global", "user", "project"]
+    },
+    "storage": {
+      "type": "string",
+      "description": "Path to the custom storage directory for Verdaccio"
+    },
+    "port": {
+      "type": "number",
+      "description": "Port of local registry that Verdaccio should listen to",
+      "default": 4873
+    },
+    "config": {
+      "type": "string",
+      "description": "Path to the custom Verdaccio config file"
+    },
+    "clear": {
+      "type": "boolean",
+      "description": "Clear local registry storage before starting Verdaccio",
+      "default": true
+    }
+  },
+  "required": ["port"]
+}
--- a/packages/js/src/executors/verdaccio/verdaccio.impl.ts
+++ b/packages/js/src/executors/verdaccio/verdaccio.impl.ts
@@ -1,0 +1,192 @@
+import { ExecutorContext, logger } from '@nx/devkit';
+import { removeSync, existsSync } from 'fs-extra';
+import { ChildProcess, execSync, fork } from 'child_process';
+import { VerdaccioExecutorSchema } from './schema';
+
+let childProcess: ChildProcess;
+
+/**
+ * - set npm and yarn to use local registry
+ * - start verdaccio
+ * - stop local registry when done
+ */
+export async function verdaccioExecutor(
+  options: VerdaccioExecutorSchema,
+  context: ExecutorContext
+) {
+  try {
+    require.resolve('verdaccio');
+  } catch (e) {
+    throw new Error(
+      'Verdaccio is not installed. Please run `npm install verdaccio` or `yarn add verdaccio`'
+    );
+  }
+
+  if (options.clear && options.storage && existsSync(options.storage)) {
+    removeSync(options.storage);
+  }
+  const cleanupFunctions = [setupNpm(options), setupYarn(options)];
+
+  const processExitListener = (signal?: number | NodeJS.Signals) => {
+    if (childProcess) {
+      childProcess.kill(signal);
+    }
+    for (const fn of cleanupFunctions) {
+      fn();
+    }
+  };
+  process.on('exit', processExitListener);
+  process.on('SIGTERM', processExitListener);
+  process.on('SIGINT', processExitListener);
+  process.on('SIGHUP', processExitListener);
+
+  try {
+    await startVerdaccio(options);
+  } catch (e) {
+    logger.error('Failed to start verdaccio: ' + e.toString());
+    return {
+      success: false,
+    };
+  }
+  return {
+    success: true,
+  };
+}
+
+/**
+ * Fork the verdaccio process: https://verdaccio.org/docs/verdaccio-programmatically/#using-fork-from-child_process-module
+ */
+function startVerdaccio(options: VerdaccioExecutorSchema) {
+  return new Promise((resolve, reject) => {
+    childProcess = fork(
+      require.resolve('verdaccio/bin/verdaccio'),
+      createVerdaccioOptions(options),
+      {
+        env: {
+          ...process.env,
+          VERDACCIO_HANDLE_KILL_SIGNALS: 'true',
+        },
+        stdio: ['inherit', 'pipe', 'pipe', 'ipc'],
+      }
+    );
+
+    childProcess.stdout.on('data', (data) => {
+      process.stdout.write(data);
+    });
+    childProcess.stderr.on('data', (data) => {
+      if (
+        data.includes('VerdaccioWarning') ||
+        data.includes('DeprecationWarning')
+      ) {
+        process.stdout.write(data);
+      } else {
+        reject(data);
+      }
+    });
+    childProcess.on('error', (err) => {
+      reject(err);
+    });
+    childProcess.on('disconnect', (err) => {
+      reject(err);
+    });
+    childProcess.on('exit', (code) => {
+      if (code === 0) {
+        resolve(code);
+      } else {
+        reject(code);
+      }
+    });
+  });
+}
+
+function createVerdaccioOptions(options: VerdaccioExecutorSchema) {
+  const verdaccioArgs: string[] = [];
+  if (options.port) {
+    verdaccioArgs.push('--listen', options.port.toString());
+  }
+  if (options.config) {
+    verdaccioArgs.push('--config', options.config);
+  }
+  return verdaccioArgs;
+}
+
+function setupNpm(options: VerdaccioExecutorSchema) {
+  try {
+    execSync('npm --version');
+  } catch (e) {
+    return () => {};
+  }
+
+  let npmRegistryPath;
+  try {
+    npmRegistryPath = execSync(
+      `npm config get registry --location ${options.location}`
+    )
+      ?.toString()
+      ?.trim()
+      ?.replace('\u001b[2K\u001b[1G', ''); // strip out ansi codes
+    execSync(
+      `npm config set registry http://localhost:${options.port}/ --location ${options.location}`
+    );
+    execSync(
+      `npm config set //localhost:${options.port}/:_authToken="secretVerdaccioToken"`
+    );
+    logger.info(`Set npm registry to http://localhost:${options.port}/`);
+  } catch (e) {
+    throw new Error(
+      `Failed to set npm registry to http://localhost:${options.port}/: ${e.message}`
+    );
+  }
+
+  return () => {
+    try {
+      if (npmRegistryPath) {
+        execSync(
+          `npm config set registry ${npmRegistryPath} --location ${options.location}`
+        );
+        logger.info(`Reset npm registry to ${npmRegistryPath}`);
+      } else {
+        execSync(`npm config delete registry --location ${options.location}`);
+      }
+    } catch (e) {
+      throw new Error(`Failed to reset npm registry: ${e.message}`);
+    }
+  };
+}
+
+function setupYarn(options: VerdaccioExecutorSchema) {
+  try {
+    execSync('yarn --version');
+  } catch (e) {
+    return () => {};
+  }
+
+  let yarnRegistryPath;
+  try {
+    yarnRegistryPath = execSync(`yarn config get registry`)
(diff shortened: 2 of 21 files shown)


packages/js/src/generators/init/schema.json
```diff
@@ -25,6 +25,11 @@
       "type": "string",
       "description": "Customize the generated tsconfig file name.",
       "x-priority": "internal"
+    },
+    "publishable": {
+      "type": "boolean",
+      "default": false,
+      "description": "Generate a publishable library."
     }
```
AgentEnder, line 33: Publishable doesn't really make sense on `init` to me. We are creating a publishable library with the library generator, not the init generator.

At minimum, we could change the description to something like `Generate a publish config.`, but I think I'd go further and do something like this:

```suggestion
    "publishConfig": {
      "type": "boolean",
      "default": false,
      "description": "Generate a publish config."
    }
```

packages/js/src/generators/library/library.ts
```diff
@@ -200,6 +200,13 @@ function addProject(
         command: `node ${publishScriptPath} ${options.name} {args.ver} {args.tag}`,
         dependsOn: ['build'],
       };
+
+      projectConfiguration.targets['local-registry'] = {
+        executor: '@nx/js:',
```
AgentEnder, line 205:
```suggestion
        executor: '@nx/js:verdaccio',
```

packages/js/src/generators/library/library.ts
```diff
@@ -200,6 +200,13 @@ function addProject(
         command: `node ${publishScriptPath} ${options.name} {args.ver} {args.tag}`,
         dependsOn: ['build'],
       };
+
+      projectConfiguration.targets['local-registry'] = {
```
AgentEnder, line 204: Should this go on the project, wherever it is, or should it specifically go on the root project? I'd imagine that you wouldn't want multiple of this target to run if you ran e2e in parallel, and e2e would likely depend on this right?

scripts/depcheck/missing.ts
```diff
@@ -67,7 +67,7 @@ const IGNORE_MATCHES_IN_PACKAGE = {
     'identity-obj-proxy',
     '@angular-devkit/schematics',
   ],
-  js: [],
```
AgentEnder, line 70: We should put an optional peer dep in @nx/js/package.json and then this shouldn't be needed.

AgentEnder (changes requested):
Hey Emily 👋, this looks good. Few changes.

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,92 @@
...
+      `npm config set registry http://localhost:${options.port}/ --location ${options.location}`
+    );
+    execSync(`yarn config set registry http://localhost:${options.port}/`);
+
+    const processExitListener = (signal?: number | NodeJS.Signals) => {
+      if (childProcess) {
+        childProcess.kill(signal);
+      }
+      execSync(`npm config delete registry --location ${options.location}`);
+      execSync('yarn config delete registry');
```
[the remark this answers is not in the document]
xiongemi: changed.

packages/js/src/generators/library/library.ts
xiongemi: added a check for root project, so the local-registry target will only be added to the publishable root project.

packages/js/src/generators/init/schema.json
xiongemi: changed.

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,113 @@
...
+      {
+        env: {
+          ...process.env,
+          VERDACCIO_HANDLE_KILL_SIGNALS: 'true',
+          ...(options.storage
+            ? { VERDACCIO_STORAGE_PATH: options.storage }
+            : {}),
+        },
+      }
+    );
```
AgentEnder, line 84: Since we are leaving this executor running to handle verdaccio cleanup, do you think it makes sense to also either let fork inherit at least stdout + stderr or log those out ourselves?

packages/js/src/generators/library/library.ts
```diff
@@ -72,6 +73,12 @@ export async function projectGenerator(
 
   tasks.push(addProjectDependencies(tree, options));
 
+  if (options.rootProject && options.publishable) {
+    tasks.push(
+      addDependenciesToPackageJson(tree, {}, { verdaccio: verdaccioVersion })
+    );
+  }
```
AgentEnder, line 80: I don't think this quite makes sense. We should add the verdaccio setup even if the lib we are generating isn't the root project, I'm just saying that I think the task itself should be added to the root project rather than the plugin project s.t. multiple plugins don't end up starting multiple registries.

xiongemi: changed

packages/js/src/executors/verdaccio/verdaccio.impl.ts
xiongemi: change pipe for stdout + stderr and add on('data') event on these 2

packages/js/src/executors/verdaccio/schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "title": "Verdaccio Local Registry",
+  "description": "Start a local registry with Verdaccio.",
+  "cli": "nx",
+  "type": "object",
+  "properties": {
+    "location": {
+      "type": "string",
+      "description": "Location option for npm config",
+      "default": "user",
+      "examples": ["global", "user", "project"]
```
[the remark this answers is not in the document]
xiongemi: change it to enum

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+export async function verdaccioExecutor(
+  options: VerdaccioExecutorSchema,
+  context: ExecutorContext
+) {
+  try {
+    execSync(
+      `npm config set registry http://localhost:${options.port}/ --location ${options.location}`
+    );
+    execSync(`yarn config set registry http://localhost:${options.port}/`);
```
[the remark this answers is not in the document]
xiongemi: chnaged

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,176 @@
...
+  try {
+    require.resolve('verdaccio');
+  } catch (e) {
+    throw new Error(
+      'Verdaccio is not installed. Please run `npm install verdaccio` or `yarn add verdaccio`'
+    );
+  }
+
+  const cleanNpm = setupNpm(options);
+  const cleanYarn = setupYarn(options);
```
AgentEnder, line 25: This may be cleaner as 
```suggestion
  const cleanupFunctions = [setupNpm(options), setupYarn(options)];
```

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  const processExitListener = (signal?: number | NodeJS.Signals) => {
+    if (childProcess) {
+      childProcess.kill(signal);
+    }
+    cleanNpm();
+    cleanYarn();
```
AgentEnder, line 32: In line with earlier suggestion
```suggestion
    for (const fn of cleanupFunctions) { fn() };
```

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,176 @@
...
+  process.on('SIGINT', processExitListener);
+  process.on('SIGHUP', processExitListener);
+
+  try {
+    await startVerdaccio(options);
+  } catch (e) {
+    return {
+      success: false,
+    };
+  }
```
AgentEnder, line 45: This will hide any errors from startVerdaccio, we should at least log them out.

packages/js/src/generators/library/library.ts
```diff
@@ -72,6 +73,10 @@ export async function projectGenerator(
 
   tasks.push(addProjectDependencies(tree, options));
 
+  if (options.rootProject && options.publishable) {
```
AgentEnder, line 76: We shouldn't need to check root project here. If generating a publishable JS lib, we **_always_** want to make sure that verdaccio has been setup, it just always ends up on the root project
```suggestion
  if (options.publishable) {
```

packages/js/src/generators/setup-verdaccio/files/config.yml
```diff
@@ -0,0 +1,25 @@
+# path to a directory with all packages
+storage: ../tmp/local-registry/storage
+
+# a list of other known repositories we can talk to
+uplinks:
+  npmjs:
+    url: https://registry.npmjs.org/
+    maxage: 60m
+
+packages:
```
AgentEnder, line 10: I wonder if the behaviour we were seeing regarding needing to be auth'd was due to not having the scoped rule from the nx repo's config. If you add it do we still need `npm adduser`?

packages/js/src/generators/setup-verdaccio/generator.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+          options: {
+            port: 4873,
+            config: '.verdaccio/config.yml',
+          },
+        },
+      },
+    });
+  });
+
+  it('should add local-registry target to project.json if it does exist', async () => {
```
AgentEnder, line 34: Can we add a test for not overwriting existing targets?

packages/js/src/generators/setup-verdaccio/generator.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!tree.exists('project.json')) {
+    const { name } = readJson(tree, 'package.json');
+    addProjectConfiguration(tree, name, {
+      root: '.',
+      targets: {
+        ['local-registry']: verdaccioTarget,
+      },
+    });
+  } else {
+    updateJson(tree, 'project.json', (json: ProjectConfiguration) => {
```
AgentEnder, line 39: Can you add a comment explaining why we use `updateJson` here? IIRC its because we know the project.json exists but don't know the project name.

AgentEnder (changes requested):
Looking really close! Thanks Emily.

packages/js/src/executors/verdaccio/verdaccio.impl.ts
xiongemi: changed.

packages/js/src/executors/verdaccio/verdaccio.impl.ts
xiongemi: changed

packages/js/src/generators/setup-verdaccio/generator.spec.ts
xiongemi: added

packages/js/src/executors/verdaccio/verdaccio.impl.ts
xiongemi: changed

packages/js/src/generators/setup-verdaccio/files/config.yml
xiongemi: from https://verdaccio.org/docs/authentication/, it seems like i need to add a line like `//localhost:5555/:_authToken="secretVerdaccioToken"` to `.npmrc`. so i added a command `npm config set //localhost:${options.port}/:_authToken="secretVerdaccioToken"` in the npmSetup and it seems to work without doing `npm adduser`

AgentEnder: That's awesome! Thanks for looking in to it. Thanks @xiongemi

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,188 @@
...
+      `npm config get registry --location ${options.location}`
+    )
+      ?.toString()
+      ?.trim();
+    execSync(
+      `npm config set registry http://localhost:${options.port}/ --location ${options.location}`
+    );
+    execSync(
+      `npm config set //localhost:${options.port}/:_authToken="secretVerdaccioToken"`
+    );
```
AgentEnder, line 132: Can we add this for yarn as well?

AgentEnder (changes requested):
Looks good! One small change and one question.

packages/js/src/executors/verdaccio/verdaccio.impl.ts
```diff
@@ -0,0 +1,188 @@
...
+  try {
+    require.resolve('verdaccio');
+  } catch (e) {
+    throw new Error(
+      'Verdaccio is not installed. Please run `npm install verdaccio` or `yarn add verdaccio`'
+    );
+  }
+
+  if (options.storage && existsSync(options.storage)) {
+    removeSync(options.storage);
```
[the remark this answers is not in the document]
xiongemi: how about just adding a clear flag and defaulting it to true?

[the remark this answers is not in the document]
xiongemi: added.

packages/js/src/executors/verdaccio/verdaccio.impl.ts
xiongemi: i read form https://verdaccio.org/docs/setup-yarn/ that yarn just uses `.npmrc` as the default. i think it gets the auth toekn from .npmrc. i don't think there is way to set it in .yarnrc

AgentEnder (approved):

bcabanes (approved):

ndcunningham (approved):

jaysoo (approved):

PR #16947, opened 2023-05-11
feat(js): refactor usages of npmScope

FrozenPandaz:
## Current Behavior

`create-nx-workspace` and `nx init` set `npmScope` in `nx.json`.

## Expected Behavior

`create-nx-workspace` and `nx init` no longer set `npmScope` in `nx.json`. If `npmScope` is still in `nx.json`, then that is still read but if not then the scope is read from the root `package.json` if available. Otherwise, it will be as if there is no `npmScope`.

## Related Issue(s)

Fixes #


Coly010 (changes requested):
This seems to be missing a file where you would also want to make the change:

https://github.com/nrwl/nx/blob/master/packages/angular/src/generators/utils/project.ts#L63

mandarini (approved):

AgentEnder (approved):

PR #16948, opened 2023-05-11
feat(nextjs): update Next.js to 13.4.1 and default to App Router for new apps

jaysoo:
This PR migrates existing apps to Next.js that are already on `>= 13.0.0` to `13.4.1`. New apps are created with Next.js 13.4.1 and App Router is on by default.

## Related Issue(s)

Fixes #


ndcunningham (changes requested):
You need to also update the `normalizeOptions` inside the application generator
```ts
  const appDir = options.appDir ?? false;
```

Should now be
```ts
  const appDir = options.appDir ?? true;
```

ndcunningham:
Few more places to check `nextAppDir`
- `packages/workspace/src/generators/preset/schema.json`
- `packages/workspace/src/generators/new/schema.json`

ndcunningham (approved):

FrozenPandaz (approved):

packages/next/src/generators/application/schema.json
```diff
@@ -121,9 +121,9 @@
     },
     "appDir": {
       "type": "boolean",
-      "default": false,
-      "description": "Enable experimental app directory for the project",
-      "x-prompt": "Do you want to use experimental app/ in this project?"
+      "default": true,
+      "description": "Enable the App Router for this project.",
+      "x-prompt": "Would you like to use the App Router (recommended)?"
```
ndcunningham, line 126: Should this be removed?

jaysoo: I think we still want it just because next.js also prompts for it. And I think there are still bugs with the App Router, like with emotion, even though they say it's "stable".

PR #16966, opened 2023-05-11
fix(react): do not require webpackConfig for cypress component testing

chrismcleod:
## Current Behavior
If a project has no webpackConfig field set, the component-testing preset will set the final webpackConfig to undefined.

This is because webpackConfig was removed as a default from the build target and so now there is no value to load.

If there is no options.webpackConfig, the buildTargetWebpack returns nothing.

## Expected Behavior
When setting up cypress component testing for a library which targets a build in a different application, the application project.json build target should not require setting "webpackConfig".

### DIFF
--- a/packages/react/plugins/component-testing/index.ts
+++ b/packages/react/plugins/component-testing/index.ts
@@ -236,35 +236,36 @@ function buildTargetWebpack(
     buildableProjectConfig.sourceRoot!
   );
 
-  if (options.webpackConfig) {
-    let customWebpack: any;
+  let customWebpack: any;
 
+  if (options.webpackConfig) {
     customWebpack = resolveCustomWebpackConfig(
       options.webpackConfig,
       options.tsConfig
     );
-
-    return async () => {
-      customWebpack = await customWebpack;
-      // TODO(jack): Once webpackConfig is always set in @nx/webpack:webpack, we no longer need this default.
-      const defaultWebpack = getWebpackConfig(context, {
-        ...options,
-        root: workspaceRoot,
-        projectRoot: ctProjectConfig.root,
-        sourceRoot: ctProjectConfig.sourceRoot,
-      });
-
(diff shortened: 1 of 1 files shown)


packages/react/plugins/component-testing/index.ts
```diff
@@ -236,34 +236,34 @@ function buildTargetWebpack(
...
-      const defaultWebpack = getWebpackConfig(context, {
-        ...options,
-        root: workspaceRoot,
-        projectRoot: ctProjectConfig.root,
-        sourceRoot: ctProjectConfig.sourceRoot,
-      });
+  if(options.webpackConfig) customWebpack = resolveCustomWebpackConfig(
+    options.webpackConfig,
+    options.tsConfig
+  );
```
chrismcleod, line 245:
```suggestion
  if(options.webpackConfig) {
    customWebpack = resolveCustomWebpackConfig(
      options.webpackConfig,
      options.tsConfig
    );
  }
```

barbados-clemens (approved):

barbados-clemens:
Ran tests locally and everything seems fine. CI looks to be failing because of formatting and latest commit message doesn't follow the format. 

Can you squash your commits into the first commit or just reword the second commit to fit the format of `fix(testing): do not require webpackConfig for react component testing` or something similar
and run `nx format`

thanks again for the PR @chrismcleod

chrismcleod:
> fix(testing): do not require webpackConfig for react component testing

when I run `nx format` it ends up changing a ton of files.  If i reset everything but the file I changed and push, it still fails for a format issue.  I tried fixing the format in the one file i changed, but to no avail.  The cicd seems to be hiding the actual format error so I'm not sure what the problem is.

I will squash my commits

chrismcleod:
After squashing, running `nx format` currently produces no changes.

PR #16982, opened 2023-05-12
feat(nx-plugin): Update executor generator

MJez29:
## Current Behavior

The executor generator generates

```ts
import { MyPluginExecutorSchema } from "./schema";

export default async function runExecutor(options: MyPluginExecutorSchema) {
  console.log("Executor ran for MyPlugin", options);
  return {
    success: true,
  };
}
```

## Expected Behavior

The above code does not make it obvious that an executor has a second `context` argument or that it can return an `AsyncIterableIterator`. By typing the function against the `Executor` type the user will have typesafety to know that they are implementing the interface of their executor correctly and an easy way to view the type definition of an executor

```ts
import { Executor } from "@nx/devkit";

import { MyPluginExecutorSchema } from "./schema";

const runExecutor: Executor<MyPluginExecutorSchema> = async (
  options,
  context
) => {
  console.log("Executor ran for MyPlugin", options);
  return {
    success: true,
  };
};

export default runExecutor;
```

### DIFF
--- a/docs/generated/devkit/AsyncIteratorExecutor.md
+++ b/docs/generated/devkit/AsyncIteratorExecutor.md
@@ -1,0 +1,26 @@
+# Type alias: AsyncIteratorExecutor\<T\>
+
+Ƭ **AsyncIteratorExecutor**\<`T`\>: (`options`: `T`, `context`: [`ExecutorContext`](../../devkit/documents/ExecutorContext)) => `AsyncIterableIterator`\<\{ `success`: `boolean` }\>
+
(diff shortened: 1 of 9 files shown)


MJez29:
Looking at the errors does it make sense to create
```ts
export type PromiseExecutor = ...
export type AsyncIterableIteratorExecutor = ...
export type Executor = PromiseExecutor | AsyncIterableIteratorExecutor
```

and by default generate a `PromiseExecutor`?

AgentEnder (approved):

PR #16995, opened 2023-05-12
fix(linter): import from .tsx file lint fix now works correctly

isaacplmann:
When the `enforce-module-boundaries` lint rule detects that you're using an aliased import path to a file within the same project, it correctly shows an error.

But if the file your importing from is a `.tsx` file, the auto fix adds an `x` to the end of the file path. The auto fix would have also failed if the file name was `'something.tsetse.fly.ts'`.

This PR handles both scenarios correctly.


meeroslav (approved):
Lgtm

PR #17005, opened 2023-05-13
docs(core): document all/none for rollup external

toddbaert:
This option now supports `"all"/"none"`. **Moreover [the change that added this](https://github.com/nrwl/nx/pull/16713) dramatically changes default packaging behavior of nx:rollup, so it should be clearly documented.**

If authors don't take action and had `"external": [],` (the default) tons of dependencies will suddenly be bundled into the distributed module. For me, that caused a bunch of issues. The least we can do is add this doc, so it's clear how to achieve the previous behavior.

### DIFF
--- a/docs/generated/packages/rollup/executors/rollup.json
+++ b/docs/generated/packages/rollup/executors/rollup.json
@@ -70,7 +70,7 @@
       },
       "external": {
         "type": "array",
-        "description": "A list of external modules that will not be bundled (`react`, `react-dom`, etc.).",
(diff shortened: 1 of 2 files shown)


mandarini (approved):

toddbaert:
Thanks for merging @mandarini .

IMHO, it's worth trying to highlight [this change](https://github.com/nrwl/nx/pull/16713) as much as possible. I totally see why the change was done, and it certainly was a valid fix. Unfortunately, it also changes the default behavior as I described above, and I'm worried it will give lots of people trouble when they upgrade to 16 and `nx/rollup`.

mandarini:
Thank you for the feedback @toddbaert ! I'll keep this in mind! :)

PR #17069, opened 2023-05-17
feat(nextjs): Add --turbo support

ndcunningham:
Passing `--turbo` should enable turbopack. 
This is only available in dev mode

`--turbo` is still experimental and should **not** be used in **Production**

```shell
❯ npx nx start --turbo

> nx run next-turbo:start --turbo

> @next-turbo/source@0.0.0 start
> nx serve --turbo
> nx run next-turbo:serve:development --turbo
>>> TURBOPACK (beta)
Thank you for trying Next.js v13 with Turbopack! As a reminder,
Turbopack is currently in beta and not yet ready for production.
We appreciate your ongoing support as we work to make it ready
for everyone.
```

### DIFF
--- a/e2e/next/src/next.test.ts
+++ b/e2e/next/src/next.test.ts
@@ -463,6 +463,24 @@ describe('Next.js Applications', () => {
       checkExport: false,
     });
   }, 300_000);
+
+  it('should support --turbo to enable Turbopack', async () => {
+    const appName = uniq('app');
+
+    runCLI(
+      `generate @nx/next:app ${appName} --style=css --appDir --no-interactive`
+    );
+
+    const port = 4000;
+    const selfContainedProcess = await runCommandUntil(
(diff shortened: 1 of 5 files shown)


e2e/next/src/next.test.ts
```diff
@@ -463,6 +463,17 @@ describe('Next.js Applications', () => {
...
+
+  it('should support --turbo to enable Turbopack', async () => {
+    const appName = uniq('app');
+
+    runCLI(
+      `generate @nx/next:app ${appName} --style=css --appDir --no-interactive`
+    );
+
+    const result = runCLI(`serve ${appName}`);
+    expect(result).toContain('>>> TURBOPACK (beta)');
```
jaysoo, line 475: Maybe remove the `(beta)` part so when it is stable it would still work? I guess since we're checking strings it could change at any point anyway.

jaysoo (approved):

PR #17110, opened 2023-05-19
docs(node): add recipes for various databases w/ node

barbados-clemens:
add new 'database' section that for the node examples with dbs

unsure if any extra info should go in the docs since the repo REAME covers what would be said.

### DIFF
--- a/docs/generated/manifests/menus.json
+++ b/docs/generated/manifests/menus.json
@@ -2739,6 +2739,95 @@
         "disableCollapsible": false
       },
       {
+        "name": "Database",
+        "path": "/recipes/database",
+        "id": "database",
+        "isExternal": false,
+        "children": [
+          {
+            "name": "Using Prisma with NestJS",
+            "path": "/recipes/database/nestjs-prisma",
+            "id": "nestjs-prisma",
+            "isExternal": false,
+            "children": [],
+            "disableCollapsible": false
+          },
+          {
+            "name": "Using Mongo with Fastify",
+            "path": "/recipes/database/mongo-fastify",
+            "id": "mongo-fastify",
+            "isExternal": false,
+            "children": [],
+            "disableCollapsible": false
+          },
+          {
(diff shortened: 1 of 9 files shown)


docs/generated/manifests/menus.json
```diff
@@ -2730,6 +2730,31 @@
         "children": [],
         "disableCollapsible": false
       },
+      {
+        "name": "Database",
```
barbados-clemens, line 2734: unsure if this should be the heading or not, also should this be plural "Databases"?

docs/map.json
```diff
@@ -1192,6 +1192,19 @@
             }
           ]
         },
+        {
+          "name": "Database",
```
barbados-clemens, line 1196: unsure if this should be the heading or not, also should this be plural "Databases"?

isaacplmann:
It doesn't make sense to make a new category with only one item in it.  I would put the recipe under other unless you can think of another category that makes sense with more recipes.

barbados-clemens:
> It doesn't make sense to make a new category with only one item in it. I would put the recipe under other unless you can think of another category that makes sense with more recipes.

@isaacplmann there will be more to come. 

so maybe we just hold off on this PR until they are more ready?
currently [fastify+postgres](https://github.com/nrwl/nx-recipes/pull/22) is PR'd with plans for others though the Node Examples effort.

barbados-clemens:
> > It doesn't make sense to make a new category with only one item in it. I would put the recipe under other unless you can think of another category that makes sense with more recipes.
> 
> @isaacplmann there will be more to come.
> 
> so maybe we just hold off on this PR until they are more ready? currently [fastify+postgres](https://github.com/nrwl/nx-recipes/pull/22) is PR'd with plans for others though the Node Examples effort.


updated with all the recipes that were made for databases with node.

isaacplmann (approved):

PR #17112, opened 2023-05-19
feat(react): run remote serve in parallel to speed up server startup

### LINKED ISSUE #12139: React module federation support parallel serve
Currently when we run "nx serve --devRemotes=my-app" in react module federation application, it runs build for all the applications one by one in "remotes" of module-federation.config.js file, serve my-app as dev server and others as static file server. This takes a lot of time to build all the remotes applications, I think the time taken is more than "nx run-many --target=serve --all --parallel". For the benefit of running only one or several remotes, I think it's better to support parallel build/static-serve remotes applications.

jaysoo:
This PR changes the module federation server to start remotes in parallel so startup time is significantly reduced.

TODO:

- [x] Do some local testing around static vs dev remotes and make sure DX is good.

https://github.com/nrwl/nx/assets/53559/2e486cfc-00f2-4ad9-8563-aff74af384c9

## Related Issue(s)

Closes #12139

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -53,6 +53,7 @@
     "@nx/js": "file:../js",
     "@nx/linter": "file:../linter",
     "@nx/react": "file:../react",
+    "@nx/web": "file:../web",
     "@nx/workspace": "file:../workspace"
   },
   "publishConfig": {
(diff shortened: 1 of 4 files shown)


mandarini (approved):

ndcunningham (approved):

mandarini (approved):

minijus:
Does this change apply for Angular MF or only React?

PR #17157, opened 2023-05-23
fix(js): should check for all possible prettierrc formats before creating a default one

dmitry-stepanenko:
## Current Behavior

Prettier can have its config defined in multiple formats https://prettier.io/docs/en/configuration.html. Right now if there's no config defined in default format `.prettierrc`, Nx will create one. This breaks the formatter in the workspace, because "default" one takes priority over other names. 
## Expected Behavior

Should check for other names of prettier config before creating a default one. 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/generators/init/init.spec.ts
+++ b/packages/js/src/generators/init/init.spec.ts
@@ -40,6 +40,20 @@ describe('js init generator', () => {
     expect(prettierignore).toContain('# custom ignore file');
   });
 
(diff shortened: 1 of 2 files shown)


dmitry-stepanenko:
Prettier also supports configuration within `package.json`. Do you guys think we should validate its presence there? It will have higher priority anyway 🤔

AgentEnder (approved):

Issue #17019, opened 2023-05-15
Vite – can't import TypeScript paths in vite.config files


tomekbuszewski:
### Current Behavior

When trying to import `@my/lib` inside `vite.config.ts` and run "test", I am getting an error saying

> NX   Cannot find module '@myorg/mylib'

When I try to run "build", I am getting

> NX   Unexpected token 'export'

When using relative imports, all works fine.

### Expected Behavior

Import paths are allowed in vite.config.ts.

### GitHub Repo

https://github.com/tomekbuszewski/nx-vite-problem

### Steps to Reproduce

1. Create an app or a lib using Vite;
2. Create another one (to be able to link these two);
3. Inside an app's `vite.config`, import something using TS paths (`@my/lib`);
4. Run tests or build.

### Nx Report

```shell
>  NX   Report complete - copy this into the issue template

   Node   : 16.17.0
   OS     : darwin arm64
   npm    : 8.15.0
   Hasher : Native
   
   nx                 : 16.1.4
   @nx/js             : 16.1.4
   @nx/linter         : 16.1.4
   @nx/workspace      : 16.1.4
   @nx/cypress        : 16.1.4
   @nx/devkit         : 16.1.4
   @nx/eslint-plugin  : 16.1.4
   @nx/react          : 16.1.4
   @nrwl/tao          : 16.1.4
   @nx/vite           : 16.1.4
   @nx/web            : 16.1.4
   typescript         : 5.0.4
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

mandarini:
Linking [this issue](https://github.com/nrwl/nx/issues/15277) here.

mandarini:
@danr-za I remember we had solved this, but now it fails due to the `export` issue. Have you noticed this on your end, too?

danr-za:
hmm not that I am aware of yet. I can check

tomekbuszewski:
@mandarini Maybe it'll help you if I add that no files outside of 'src' are being treated with TypeScript. Just tried having Tailwind config import my TS theme and it also fails.

mandarini:
That's not fixed yet, it's partly fixed. Not sure how to fix atm, will revisit.

mandarini:
Updated repro: https://github.com/nrwl/nx/issues/17019

mandarini:
Update on this issue:

@jaysoo looked into it as well, and I present here the findings:

[...]

PR #17187, opened 2023-05-24
fix(vite): call registerTsProject for vite executors

mandarini:
## Current Behavior

Can't import TypeScript paths in vite.config files. Getting error:
```
 >  NX   Unexpected token 'export'
```

## Expected Behavior

Vite should build workspace libraries before importing.

## Related Issue(s)

https://github.com/nrwl/nx/issues/17019


mandarini:
@jaysoo tried this, but it does not seem to do the trick, unfortunately. Will park this here for now.

PR #17248, opened 2023-05-26
feat(nextjs): remove the "root" option from Next.js build executor since it can be inferred

jaysoo:
We never needed the `root` option for `@nx/next:build` executor since we know the project root by reading the graph. This PR:

1. Removes the `root` option from said executor.
2. Update app generator to no longer generate with it.
3. Adds a migration to remove it from existing projects.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/next/src/migrations/update-16-3-0/remove-root-build-option.spec.ts
+++ b/packages/next/src/migrations/update-16-3-0/remove-root-build-option.spec.ts
@@ -1,0 +1,72 @@
+import update from './remove-root-build-option';
+import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing-pre16';
+import {
+  addProjectConfiguration,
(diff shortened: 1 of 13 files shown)


ndcunningham (approved):

packages/next/src/migrations/update-16-3-0/remove-root-build-option.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+            root: 'my-app',
+          },
+        },
+      },
+    });
+
+    await update(tree);
+
+    const updatedConfig = readProjectConfiguration(tree, 'my-app');
+    expect(updatedConfig.targets.build.options.root).toBeUndefined();
```
ndcunningham, line 51: you may need to cast this since its failing at type check

PR #17289, opened 2023-05-29
feat(js): do not generate root babel.config.json for babel projects 

jaysoo:
This PR ensures that `babel.config.json` is not generated in new projects. The default behavior is to only apply `.babelrc` file for individual projects and ignore its dependencies' `.babelrc` files. The implication of not using `rootMode: upward` is that any babel presets/plugins needed by a a dependency also needs to be added to the dependent project's `.babelrc` file.

Additionally, anywhere that we can safely swap to SWC instead of Babel we will do so by default.

## Notes

- Both `@nx/node` and `@nx/express` do not support building projects using Babel, so the `babelJest` option is deprecated and `swcJest` is added to generate Jest project with `@swc/jest` (to speed up test compile).
-  React still needs Babel by default if using styled-components or emotion. Will address it in a separate PR.


packages/react/src/generators/library/lib/create-files.ts
```diff
@@ -37,10 +38,34 @@ export function createFiles(host: Tree, options: NormalizedSchema) {
       options.projectRoot,
       substitutions
     );
+  }
 
-    if (host.exists(joinPathFragments(options.projectRoot, '.babelrc'))) {
-      host.delete(joinPathFragments(options.projectRoot, '.babelrc'));
-    }
+  if (options.bundler === 'rollup' && options.compiler === 'babel') {
```
jaysoo, line 43: Inlined the `.babelrc` creation here rather than in the `files` folder.

packages/next/src/generators/library/library.ts
```diff
@@ -71,36 +71,6 @@ export async function libraryGenerator(host: Tree, rawOptions: Schema) {
   );
   addTsConfigPath(host, `${options.importPath}/server`, [serverEntryPath]);
 
-  updateJson(
```
jaysoo, line 74: This never worked well, and the user should already have emotion and styled-components in the app. As per this PR (https://github.com/nrwl/nx/pull/17286) the app's babel config will now be applied to the libs as well. You never want to be in a situation where different styling solutions are used between apps and libs without a build step in the lib to normalize to CSS or something.

jaysoo: We'll need a follow-up PR to enable emotion and styled-components in SWC for Next.js since they've added some early support for them.

packages/react/src/generators/library/lib/create-files.ts
```diff
@@ -37,10 +38,33 @@ export function createFiles(host: Tree, options: NormalizedSchema) {
       options.projectRoot,
       substitutions
     );
+  }
 
-    if (host.exists(joinPathFragments(options.projectRoot, '.babelrc'))) {
-      host.delete(joinPathFragments(options.projectRoot, '.babelrc'));
-    }
+  if (options.compiler === 'babel') {
```
jaysoo, line 43: Inlined the .babelrc creation here rather than in the files folder.

docs/generated/packages/react/generators/application.json
```diff
@@ -155,7 +155,7 @@
       "compiler": {
         "type": "string",
         "description": "The compiler to use.",
-        "enum": ["babel", "swc"],
+        "enum": ["swc", "babel"],
         "default": "babel"
```
ndcunningham, line 159: should this be set to `swc`?

docs/generated/packages/react/generators/library.json
```diff
@@ -169,7 +169,7 @@
       },
       "compiler": {
         "type": "string",
-        "enum": ["babel", "swc"],
+        "enum": ["swc", "babel"],
         "default": "babel",
```
ndcunningham, line 173: same here

packages/react/src/generators/application/schema.json
```diff
@@ -161,7 +161,7 @@
     "compiler": {
       "type": "string",
       "description": "The compiler to use.",
-      "enum": ["babel", "swc"],
+      "enum": ["swc", "babel"],
       "default": "babel"
```
ndcunningham, line 165: same here

docs/generated/packages/react/generators/application.json
jaysoo: No, I realized it's not quite working with React + SWC, so I need another PR to fix that then change it. If we change it now without the fix then the setup is wrong for CSS-in-JS.

packages/react/src/generators/application/schema.json
jaysoo: The React SWC setup wasn't 100% correct with CSS-in-JS, but that's been fixed here https://github.com/nrwl/nx/pull/17295

We can make it default now.

ndcunningham (approved):

packages/node/src/generators/application/application.spec.ts
```diff
@@ -343,7 +343,33 @@ describe('app', () => {
     });
   });
 
-  describe('--babelJest', () => {
+  describe('--swcJjest', () => {
```
barbados-clemens, line 346: nitpick: 
```suggestion
  describe('--swcJest', () => {
```

barbados-clemens (approved):

packages/react/src/generators/host/schema.json
```diff
@@ -138,7 +138,7 @@
       "type": "string",
       "description": "The compiler to use",
       "enum": ["babel", "swc"],
-      "default": "babel"
+      "default": "swc"
```
ndcunningham, line 141: should this be `babel` as well?

packages/react/src/generators/remote/schema.json
```diff
@@ -143,7 +143,7 @@
       "type": "string",
       "description": "The compiler to use.",
       "enum": ["babel", "swc"],
-      "default": "babel"
+      "default": "swc"
```
ndcunningham, line 146: should this be `babel`

PR #17292, opened 2023-05-29
[Do not merge]: feat(bundling): Add support to rollup for babelUpwardRootMode

ndcunningham:


packages/rollup/src/executors/rollup/schema.d.ts
```diff
@@ -32,4 +32,6 @@ export interface RollupExecutorOptions {
   javascriptEnabled?: boolean;
   generateExportsField?: boolean;
   skipTypeCheck?: boolean;
+  babelConfig?: string;
+  babelUpwardRootMode?: boolean;
```
jaysoo, line 36: Also need to add `babelUpwardRootMode` to the `schema.json` file. And I don't think we need `babelConfig` option here since Babel can find it automatically, we just want to let users enable or disable the root mode.

jaysoo:
There seems to be missing logic to set `rootMode: undefined | 'upward'` in `rollup.impl.ts`. The new `babelUpwardRootMode` for `@nx/rollup:rollup` executor will allow users to set `rootMode: 'upward'` (matching previous behavior), otherwise we should set `rootMode: undefined` to only use the `.babelrc` of the current project being built.

I don't think we need to add the `babelConfig` option to generators. The `.babelrc` file is already added for the user depending on the `--compiler` option when generating a library, we just need to update the executor to not set `rootMode: 'upward'` by default.

jaysoo (approved):

PR #17310, opened 2023-05-30
feat(core): extend option readyWhen to accept multiple strings

### LINKED ISSUE #17308: nx:run-commands: allow option "readyWhen" to accept multiple strings

## Description

The option `readyWhen` for `nx:run-commands` could be either an array of strings (= new feature) _or_ just a single string (= current behavior).

## Motivation

Imagine a complex setup where we use `nx:run-commands` to start multiple dev-servers in parallel. In such a scenario, we want to consider the commands as "ready" as soon as all the dev-servers have started, but we don't always know for sure which one will take the longest to start. And when we have other commands depending on this, like e2e tests for example, it could be problematic to proceed when one of the servers was not ready yet.

## Suggested Implementation

The type for option `readyWhen` could be either a string or an array of strings ( `string | string[]` ).
When the provided value is a single `string`, the behavior remains exactly as it is today. This will still be the most likely use-case and it ensures backwards compatibility.
When the provided value is an array of strings, we should wait until _all_ of the provided strings have been found in `stdout` or `stderr`.

I will open a PR which implements this behavior.

kfrederix:
## Current Behavior

Option `readyWhen` for `nx:run-commands` only accepts a single string.

## Expected Behavior

Option `readyWhen` for `nx:run-commands` accepts either a single string _or_ an array of strings. In case the provided value is an array of strings, we will wait until all strings have appeared in `stdout` or `stderr`.

## Related Issue(s)

Fixes #17308


raqso (approved):

packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -219,24 +232,40 @@ function createProcess(
...
+          readyWhenElement.found = true;
+          if (isReady()) {
+            res(true);
+          }
+          break;
+        }
       }
     });
     childProcess.on('error', (err) => {
+      console.log('error', err);
```
xiongemi, line 263: could you remove this console.log?

xiongemi (approved):
the pr looks good. just 1 comment and could you rebase and resolve the conflict?

PR #17321, opened 2023-05-30
docs(core): deprecate npmScope

isaacplmann:
Adds documentation about npm scope being deprecated


docs/shared/reference/nx-json.md
```diff
@@ -53,10 +52,6 @@ The following is an expanded version showing all options. Your `nx.json` will li
 
 Some presets use the `extends` property to hide some default options in a separate json file. The json file specified in the `extends` property is located in your `node_modules` folder. The Nx preset files are specified in [the `nx` package](https://github.com/nrwl/nx/tree/master/packages/nx/presets).
 
-### NPM Scope
```
FrozenPandaz, line 56: Rather than having a separate doc for `npmScope`, could we keep this section here? Maybe move it to the bottom?

FrozenPandaz (changes requested):

docs/shared/reference/nx-json.md
isaacplmann: npm scope will no longer be part of the nx json file.  We can leave the section there until in v17, but when npm scope is ignored we shouldn't keep it in the nx json reference.

bcabanes: Is this something we want to use an alert or warning callout for?

bcabanes (approved):

FrozenPandaz (approved):

PR #17328, opened 2023-05-31
fix(vite): add ability to specify reportCompressedSize for vite build

friedbizkvit:
vite has build option reportCompressedSize, it enables/disables calculation of gzip size of the output files.
compressing large output files can be slow, so disabling this may increase build performance for large projects
or in cases where it is unnecessary

## Current Behavior
vite executor rewrites the value of specified reportCompressedSize option with 'true'

## Expected Behavior
vite executor uses the value of reportCompressedSize option if specified, otherwise set it to 'true'

## Related Issue(s)

Fixes #


mandarini:
Hi there @friedbizkvit ! Thank you SO much for your PR!!!!

We have made so many changes since you submitted this. Essentially, you can pass any arg to vite, and Nx will pass it down. We are moving away from executors, so now we suggest you either use the Vite CLI or add the options in your `vite.config.ts`.

Unfortunately, this means that I will have to close your PR, but thank you so much for taking the time!!!!

PR #17338, opened 2023-05-31
feat(core): remove node hashing implementation

Cammisuli:
## Current Behavior
Nx has a node and a rust implementation to hash the current workspace

## Expected Behavior
Nx only has the rust implementation to hash the current workspace

## Related Issue(s)

Fixes #


packages/nx/src/command-line/report/report.ts
```diff
@@ -63,14 +59,12 @@ export async function reportHandler() {
...
     outOfSyncPackageGroup,
     projectGraphError,
-    currentHasherImplementation,
   } = await getReportData();
 
   const bodyLines = [
     `Node   : ${process.versions.node}`,
     `OS     : ${process.platform} ${process.arch}`,
     `${pm.padEnd(7)}: ${pmVersion}`,
-    `Hasher : ${currentHasherImplementation}`,
```
FrozenPandaz, line 73: Could we perhaps print if the native code is available?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #17347, opened 2023-06-01
chore(repo): change e2e to use verdaccio executor

xiongemi:
## Related Issue(s)

Fixes #


e2e/angular-core/global-setup.ts
FrozenPandaz, line 1: These files are all the same right? Can we have 1 file which every project uses?

Ultimately, what we will want is for this to logic to live within `@nx/js` so that plugin authors will be able to set it up easily.

I do like the idea of dog fooding this for a while here though.

e2e/workspace-create/global-setup.ts
```diff
@@ -0,0 +1,48 @@
...
+      const childProcess = spawn(
+        `nx`,
+        `local-registry @nx/nx-source --location none --port ${port} --storage ${global.nxLocalRegistryStorage}`.split(
+          ' '
+        )
+      );
+
+      childProcess.stdout.on('data', (data) => {
+        if (data.toString().includes(registry)) {
+          console.log('local registry started');
```
FrozenPandaz, line 28:
```suggestion
          console.log('Local registry started on port ' + port);
```

e2e/workspace-create/global-setup.ts
```diff
@@ -0,0 +1,48 @@
...
+        console.log('local registry error', err);
+        reject(err);
+      });
+      childProcess.on('exit', (code) => {
+        console.log('local registry exit', code);
+        reject(code);
+      });
+    }
+  );
+
```
FrozenPandaz, line 46: Let's add a console.log which says that we are publishing packages.

e2e/workspace-create/global-setup.ts
```diff
@@ -0,0 +1,48 @@
+import { join } from 'path';
+import * as detectPort from 'detect-port';
+import { ChildProcess, execSync, spawn } from 'child_process';
+
+export default async function () {
+  const port = await detectPort(4872);
+  const registry = `http://localhost:${port}`;
+  process.env.npm_config_registry = registry;
```
FrozenPandaz, line 8: Can we add a console.log here which says that npm + yarn have been configured to use the local registry?

nx.json
```diff
@@ -129,18 +129,7 @@
...
-          },
-          {
-            "command": "pnpm e2e-build-package-publish"
-          },
-          {
-            "command": "nx run-e2e-tests {projectName}"
-          }
-        ],
-        "parallel": false
+        "commands": ["nx run-e2e-tests {projectName}"]
```
FrozenPandaz, line 132: Could we make the `targetDefaults` for `e2e` and `e2e-macos` the same as what they are for `run-e2e-tests` with these inputs? We shouldn't need a separate target after this change.

e2e/angular-core/global-setup.ts
```diff
@@ -0,0 +1,48 @@
+import { join } from 'path';
+import * as detectPort from 'detect-port';
+import { ChildProcess, execSync, spawn } from 'child_process';
+
+export default async function () {
+  const port = await detectPort(4872);
```
FrozenPandaz, line 6: Could we build the port detection into the `verdaccio` executor?

FrozenPandaz (changes requested):
Some cleanup

e2e/angular-core/global-setup.ts
xiongemi: moved to setup and teardown to e2e/utils

e2e/workspace-create/global-setup.ts
xiongemi: changed.

e2e/workspace-create/global-setup.ts
xiongemi: chnaged.

nx.json
xiongemi: changed

barbados-clemens (approved):

mandarini (approved):

e2e/js/tsconfig.spec.json
```diff
@@ -15,6 +15,7 @@
     "**/*.spec.jsx",
     "**/*.test.jsx",
     "**/*.d.ts",
-    "jest.config.ts"
+    "jest.config.ts",
+    "local-registry-setup.ts"
```
FrozenPandaz, line 19: I don't think this is necessary?

followup: can you delete it?

e2e/utils/global-setup.ts
```diff
@@ -0,0 +1,53 @@
+import { join } from 'path';
+import { ChildProcess, execSync, spawn } from 'child_process';
+
+export default async function () {
+  let port = 4873;
```
FrozenPandaz, line 5: followup: we don't need port here

e2e/utils/global-setup.ts
```diff
@@ -0,0 +1,53 @@
...
+    process.cwd(),
+    'tmp/local-registry/storage',
+    process.env.NX_TASK_TARGET_PROJECT ?? ''
+  );
+
+  global.nxLocalRegistryProcess = await new Promise<ChildProcess>(
+    (resolve, reject) => {
+      const childProcess = spawn(
+        `nx`,
+        `local-registry @nx/nx-source --location none --port ${port} --storage ${global.nxLocalRegistryStorage}`.split(
```
FrozenPandaz, line 16: followup: we don't need port here

e2e/utils/global-setup.ts
```diff
@@ -0,0 +1,53 @@
+import { join } from 'path';
+import { ChildProcess, execSync, spawn } from 'child_process';
+
+export default async function () {
+  let port = 4873;
+  global.nxLocalRegistryStorage = join(
+    process.cwd(),
+    'tmp/local-registry/storage',
+    process.env.NX_TASK_TARGET_PROJECT ?? ''
+  );
```
FrozenPandaz, line 10: followup: I don't think we actually need the project in the path here right? It can be any tmp directory.a

e2e/utils/global-setup.ts
```diff
@@ -0,0 +1,53 @@
...
+          const registry = `http://localhost:${port}`;
+          process.env.npm_config_registry = registry;
+          process.env.YARN_REGISTRY = registry;
+          console.log('Set npm and yarn config registry to ' + registry);
+
+          resolve(childProcess);
+        }
+      });
+      childProcess.stderr.on('data', (data) => {
+        process.stdout.write(data);
```
FrozenPandaz, line 37: This should go to stderr

e2e/utils/global-teardown.ts
```diff
@@ -0,0 +1,11 @@
+import { existsSync, rmSync } from 'fs-extra';
+
+export default function () {
+  if (global.nxLocalRegistryProcess) {
+    global.nxLocalRegistryProcess.kill();
+  }
+
+  if (existsSync(global.nxLocalRegistryStorage)) {
+    rmSync(global.nxLocalRegistryStorage);
+  }
```
FrozenPandaz, line 10: Let's console.log that stuff has been cleaned up

e2e/utils/global-setup.ts
```diff
@@ -0,0 +1,53 @@
...
+      const childProcess = spawn(
+        `nx`,
+        `local-registry @nx/nx-source --location none --port ${port} --storage ${global.nxLocalRegistryStorage}`.split(
+          ' '
+        )
+      );
+
+      childProcess.stdout.on('data', (data) => {
+        if (data.toString().includes('http://localhost:')) {
+          port ??= parseInt(
```
FrozenPandaz, line 23: This should always define port

FrozenPandaz (changes requested):

e2e/js/tsconfig.spec.json
xiongemi: deleted

e2e/utils/global-setup.ts
xiongemi: removed

e2e/utils/global-setup.ts
xiongemi: changed.

e2e/utils/global-teardown.ts
xiongemi: added

e2e/utils/global-setup.ts
xiongemi: chnaged

FrozenPandaz (approved):

PR #17357, opened 2023-06-01
fix(storybook): transform unsupported options to camelCase

mandarini:
Do not merge yet plz. It's ready for review, but I'm not sure yet if it's the best solution. See this PR too: https://github.com/nrwl/nx/pull/17358.  Either this or the other one should be merged. Not both.

## Current Behavior

Extra args are not converted to camelCase ([ref](https://github.com/nrwl/nx/pull/10347)). So, for example, `--webpack-stats-json` will be passed as `webpack-stats-json` to the Storybook builders. This results in the Storybook builders ignoring it, since they expect `webpackStatsJson` instead.

## Expected Behavior

Not sure, but here I am transforming the kebab-case to camelCase. When I say "not sure" I mean that I am not sure if this is indeed the expected behaviour.

### DIFF
--- a/packages/storybook/src/executors/build-storybook/build-storybook.impl.ts
+++ b/packages/storybook/src/executors/build-storybook/build-storybook.impl.ts
@@ -3,6 +3,7 @@ import * as build from '@storybook/core-server';
 import { CLIOptions } from '@storybook/types';
 import 'dotenv/config';
 import {
+  convertKeysToCamelCase,
   pleaseUpgrade,
   storybookConfigExistsCheck,
   storybookMajorVersion,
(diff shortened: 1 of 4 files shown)


valentinpalkovic:
Hi

I am one of the Storybook Core maintainers! :)

Storybook relies on this fix since CLI flags are converted automatically to camelCase in `storybook build` and `storybook dev` scripts (by using `commander`). Since nx doesn't call `storybook dev/build` directly, CLI arguments have to be adequately prepared before passing them to Storybook.

mandarini:
@valentinpalkovic thanks for jumping in! I created this PR too: https://github.com/nrwl/nx/pull/17358

Let's see which is best for Nx!

mandarini:
I decided to go with the other solution (add all Storybook CLI parameters in `schema.json`) this time. If the issue arises again, we can look into a different solution! Keeping this PR here for future reference.

PR #17361, opened 2023-06-01
cleanup(graph): no differed functionality in async function

alivarastepour:
using the async keyword without the await, is not recommended nor necessary and may lead to confusion for future refactoring.

## Related Issue(s)
may lead to confusion for future changes


scripts/generate-graph.ts
```diff
@@ -4,7 +4,7 @@ import { join } from 'path';
 import * as yargs from 'yargs';
 import { ensureDirSync, statSync } from 'fs-extra';
 
-async function generateGraph(directory: string, name: string) {
```
FrozenPandaz, line 7: This function should probably be async. The IIFE at the bottom of this function is an async function but is not awaited. Could you add an await to that function call?

FrozenPandaz (changes requested):

FrozenPandaz:
I'm going to close this PR as it is stale.

PR #17385, opened 2023-06-02
fix(angular): dynamic host should not generate webpack.prod.config.js

### LINKED ISSUE #17378: Dynamic host (MF) is generating and using a webpack.config.prod.js
### Current Behavior

When generating a host like:

`nx g @nx/angular:host host --dynamic`

It is generating a `webpack.prod.config.js` and using it for production.

### Expected Behavior

Should not generate a `webpack.prod.config.js` since it is dynamic and remotes shouldn't be hardcoded. Also production build should use the other webpack config.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create empty integrated repo
2. npm i -D @nx/angular
3. nx g @nx/angular:host host --dynamic
4. Check that the file exists and it is referenced at project.json

### Nx Report

```shell
There is no report
```

### Failure Logs

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Coly010:
## Current Behavior

We generate a webpack.prod.config.js for dynamic hosts. It's not necessary. The purpose of the webpack.prod.config.js is to overwrite the remote's locations. With dynamic federation, the module-federation.manifest.json is the source of truth for the remote's location and is fetched at runtime. This will be different per env, removing the need for webpack.prod.config.js.
## Expected Behavior

Do not generate webpack.prod.config.js for dynamic hosts and change the production configuration to point to the single webpack.config.js

## Related Issue(s)

Fixes #17378

### DIFF
--- a/packages/angular/src/generators/setup-mf/lib/setup-host-if-dynamic.ts
+++ b/packages/angular/src/generators/setup-mf/lib/setup-host-if-dynamic.ts
@@ -1,5 +1,9 @@
 import type { Tree } from '@nx/devkit';
-import { joinPathFragments, readProjectConfiguration } from '@nx/devkit';
+import {
+  joinPathFragments,
+  readProjectConfiguration,
+  updateProjectConfiguration,
+} from '@nx/devkit';
 import type { Schema } from '../schema';
 
 export function setupHostIfDynamic(tree: Tree, options: Schema) {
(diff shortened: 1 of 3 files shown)


packages/angular/src/generators/setup-mf/lib/setup-host-if-dynamic.ts
```diff
@@ -1,18 +1,40 @@
...
+  if (tree.exists(pathToProdWebpackConfig)) {
+    tree.delete(pathToProdWebpackConfig);
+  }
+
+  project.targets.build.configurations.production = {
+    ...project.targets.build.configurations.production,
+    customWebpackConfig: {
+      path: `${project.root}/webpack.config.js`,
+    },
+  };
```
leosvelperez, line 37: We shouldn't need to repeat the path here, that one is already set in the options, so all configurations will use it unless they overwrite it. We should just delete the overwrite to avoid duplicating the path:
```suggestion
  delete project.targets.build.configurations.production?.customWebpackConfig;
```

leosvelperez: Alternatively, we could handle this in the `changeBuildTarget` function, so we never set it in the first place if it's unnecessary.

Coly010: good point, i wanted to leave all the logic that handles a dynamic host in the one function, for discoverability. I don't mind though. What do you think is best? If you think it's best to keep it in this place but remove the customWebpackConfig option completely, I'll do that, or if you think it's best to change the changeBuildTarget, I'll do that :)

leosvelperez: Yeah, that's ok, I don't mind either way. The important bit is to reduce boilerplate when possible.

leosvelperez (approved):

Issue #12145, opened 2022-09-21
`@nrwl/webpack` breaks jest by exporting `.d.ts` files from `index.js`


hiddenist:
## Current Behavior

Tests utilizing the `@nrwl/webpack` package fail because the `index.js` file in the built javascript is attempting to export two `schema.d.ts` files. I'm getting this in my unit tests for my own generators that extend the react and expo generators.

I was able to address this in my original repo by applying this patch:

```patch
diff --git a/index.js b/index.js
index 222c2849d7855ee4a8ff0f23525ddc7957f10b17..8d019a925e3b1b7ed24634c0be5008b797595455 100644
--- a/index.js
+++ b/index.js
@@ -6,6 +6,4 @@ tslib_1.__exportStar(require("./src/generators/webpack-project/webpack-project")
-tslib_1.__exportStar(require("./src/executors/dev-server/schema"), exports);
 tslib_1.__exportStar(require("./src/executors/dev-server/dev-server.impl"), exports);
 tslib_1.__exportStar(require("./src/executors/webpack/lib/normalize-options"), exports);
-tslib_1.__exportStar(require("./src/executors/webpack/schema"), exports);
 tslib_1.__exportStar(require("./src/executors/webpack/webpack.impl"), exports);
 //# sourceMappingURL=index.js.map
```

## Expected Behavior

Tests including `@nrwl/webpack` or its descendants should not fail (and `.js` files should not be attempting to export typescript)

## Steps to Reproduce

Here's a minimal test that reproduces the issue. Also can be seen in [this repo created with create-nx-plugin](https://github.com/NikkiDelRosso/repro-nx-webpack-schema-bug).
```typescript
import { normalizePluginPath } from '@nrwl/webpack';

describe('@nrwl/webpack normalizePluginPath', () => {
  it('should return a defined value', async () => {
    expect(normalizePluginPath("test", "test")).toBeDefined();
  });
});
```

Running this test using `npx nx test test` results in the following error.

### Failure Logs

```
  ● Test suite failed to run

    Cannot find module './src/executors/dev-server/schema' from '../../node_modules/@nrwl/webpack/index.js'

    Require stack:
      /Users/nikki/repro-nx-webpack-schema-bug/node_modules/@nrwl/webpack/index.js
      src/test.spec.ts


    However, Jest was able to find:
    	'src/executors/dev-server/schema.d.ts'
    	'src/executors/dev-server/schema.json'

[...]

PR #17392, opened 2023-06-02
fix(js): use `export type` for exports from `schema.d.ts` files

### LINKED ISSUE #17391: `@nx/rollup` is exporting a `schema.d.ts` file, which breaks Jest
### Current Behavior

A `schema.d.ts` is being exported. Breaks `jest`

Here's a patch that fixes it:

```diff
# generated by patch-package 6.4.14
#
# declared package:
#   @nx/rollup: 16.3.1
#
diff --git a/node_modules/@nx/rollup/index.js b/node_modules/@nx/rollup/index.js
index d6ac066..be8c315 100644
--- a/node_modules/@nx/rollup/index.js
+++ b/node_modules/@nx/rollup/index.js
@@ -3,6 +3,6 @@ Object.defineProperty(exports, "__esModule", { value: true });
 const tslib_1 = require("tslib");
 tslib_1.__exportStar(require("./src/generators/init/init"), exports);
 tslib_1.__exportStar(require("./src/generators/rollup-project/rollup-project"), exports);
-tslib_1.__exportStar(require("./src/executors/rollup/schema"), exports);
+// tslib_1.__exportStar(require("./src/executors/rollup/schema"), exports);
 tslib_1.__exportStar(require("./src/executors/rollup/rollup.impl"), exports);
 //# sourceMappingURL=index.js.map

```

### Expected Behavior

A `schema.d.ts` would NOT be exported

### GitHub Repo

_No response_

### Steps to Reproduce

1. Call `libraryGenerator` from `@nx/react` in your local generator
2. Run `jest` with the default `jest.config.ts` generated by the `@nx/plugin` generator generator

### Nx Report

```shell

   Node   : 18.16.0
   OS     : darwin arm64
   pnpm   : 6.10.1
   Hasher : Native

   nx (global)        : 16.0.2
   nx                 : 16.3.1
   @nx/js             : 16.3.1
   @nx/jest           : 16.3.1
   @nx/linter         : 16.3.1
   @nx/workspace      : 16.3.1
   @nx/cypress        : 16.3.1
   @nx/devkit         : 16.3.1
   @nx/eslint-plugin  : 16.3.1
   @nx/plugin         : 16.3.1
   @nx/react          : 16.3.1
   @nx/rollup         : 16.3.1
   @nx/storybook      : 16.3.1
   @nrwl/tao          : 16.0.2
   @nx/webpack        : 16.3.1
   typescript         : 4.8.4
   Local workspace plugins:
         [REMOVED as these are enterprise packages]
     - @nrwl/tao@16.0.2

   To fix this, run `nx migrate nx@16.3.1`
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

The exact same problem, but on a different package is described in https://github.com/nrwl/nx/issues/12145. Is there a reason why we're using `schema.d.ts` instead of `schema.ts` files? Looks like the tooling is treating it as a JS file and creating a broken entrypoint for it?

MageJohn:
## Current Behavior

In the rollup and esbuild packages, types held in a `schema.d.ts` are being exported with `export * from`. This export is not removed when TypeScript transforms the file, likely because TypeScript can't be sure that an actual implementation won't be added separately. However, this results in some systems (such as our Jest setup) importing the `schema.d.ts` file and choking on the invalid syntax.

## Expected Behavior

The types should be exported, but the export should not exist at runtime. 

## Related Issue(s)

Fixes #17391

A similar problem was brought up before in #12145, and that specific instance was fixed in #12403. In this PR I have fixed the remaining instances. I did a search for other `.d.ts` files that were being exported and this is all of them.

### DIFF
--- a/packages/esbuild/index.ts
+++ b/packages/esbuild/index.ts
@@ -1,3 +1,6 @@
-export * from './src/executors/esbuild/schema';
+export type {
+  EsBuildExecutorOptions,
+  NormalizedEsBuildExecutorOptions,
+} from './src/executors/esbuild/schema';
 export * from './src/executors/esbuild/esbuild.impl';
 export * from './src/utils/versions';
--- a/packages/rollup/index.ts
+++ b/packages/rollup/index.ts
@@ -1,4 +1,8 @@
 export * from './src/generators/init/init';
 export * from './src/generators/rollup-project/rollup-project';
-export * from './src/executors/rollup/schema';
+export type {
+  AssetGlobPattern,
+  Globals,
+  RollupExecutorOptions,
+} from './src/executors/rollup/schema';
 export * from './src/executors/rollup/rollup.impl';


packages/esbuild/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './src/executors/esbuild/schema';
+export type * from './src/executors/esbuild/schema';
```
ndcunningham, line 1: Can you update this to export only options
`EsBuildExecutorOptions` and `NormalizedEsBuildExecutorOptions`

packages/rollup/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './src/generators/init/init';
 export * from './src/generators/rollup-project/rollup-project';
-export * from './src/executors/rollup/schema';
+export type * from './src/executors/rollup/schema';
```
ndcunningham, line 3: Same here export `RollupExecutorOptions` `Globals` and `AssetGlobPattern`

ndcunningham (changes requested):

packages/esbuild/index.ts
MageJohn: Done!

packages/rollup/index.ts
MageJohn: Also done!

FrozenPandaz (approved):

PR #17407, opened 2023-06-03
fix(core): fix pnpm lockfile parsing in v6.1

jojoa:
Fix the Version check in the pnpm lockfile parsing process from v6.0 to v6.x, as done in the pnpm source, to fix a error in the generatePackageJson Plugin to not find the deps

## Related Issue(s)

Fixes #


meeroslav (approved):
Looks good to me. Thanks!

PR #17415, opened 2023-06-05
fix(nx-cloud): allow regenerating nxcloud access token

rarmatei:
Currently, if someone already has NxCloud enabled, but they'd like to re-generate their access token, or even point nx cloud to a new API URL, using the `NX_CLOUD_API=https://******.nx.app npx nx connect` command - we don't allow them to do this.

This PR re-enables this functionality. We'll be able to use `nx connect` to:
- re-generate an access token
- point NxCloud to a different API URL

### DIFF
--- a/packages/nx/src/command-line/connect/command-object.ts
+++ b/packages/nx/src/command-line/connect/command-object.ts
@@ -7,7 +7,7 @@ export const yargsConnectCommand: CommandModule = {
   describe: `Connect workspace to Nx Cloud`,
   builder: (yargs) => linkToNxDevAndExamples(yargs, 'connect-to-nx-cloud'),
   handler: async () => {
-    await (await import('./connect-to-nx-cloud')).connectToNxCloudCommand();
+    await (await import('./connect-to-nx-cloud')).connectToNxCloudCommand(null, false);
     process.exit(0);
   },
 };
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/connect/command-object.ts
```diff
@@ -7,7 +7,9 @@ export const yargsConnectCommand: CommandModule = {
   describe: `Connect workspace to Nx Cloud`,
   builder: (yargs) => linkToNxDevAndExamples(yargs, 'connect-to-nx-cloud'),
   handler: async () => {
-    await (await import('./connect-to-nx-cloud')).connectToNxCloudCommand();
+    await (
+      await import('./connect-to-nx-cloud')
+    ).connectToNxCloudCommand(null, false);
```
FrozenPandaz, line 12: I think this should still prompt.

packages/nx/src/command-line/connect/connect-to-nx-cloud.ts
```diff
@@ -48,9 +49,10 @@ export async function connectToNxCloudCommand(
     });
     return false;
```
FrozenPandaz, line 50: We could just not `return` here right?

FrozenPandaz (changes requested):

PR #17433, opened 2023-06-06
feat(vite): add incremental builds support to Vite plugin build executor

### LINKED ISSUE #17322: Add incremental builds with buildLibsFromSource option to @nx/vite:build

## Description

An app built with `@nx/vite:build` executor should be able to reference it's dependencies libraries pre-compiled files instead of always rebuilding consuming the source files.
## Motivation

Isn't that the [incremental builds](https://nx.dev/more-concepts/incremental-builds) feature presented in Nx documentation? I work on a very large React standalone application. It continues to grow everyday and I need a solution to make it scalable so that many teams are able to work in the same product with good DX. I'm trying to come up with a Nx PoC to show my manager. Incremental builds is what we're looking for. Unfortunately it's not supported in the Vite plugin executor and moving to Webpack is not an option. This incremental build support would be really decisive for us to adopt Nx.
## Suggested Implementation

Add the `buildLibsFromSource` option just like `@nx/angular:webpack-browser`, `@nx/webpack:webpack` and `@nx/angular:ng-packagr-lite`. It looks like the only thing that's missing is that flag and the path mapping rewrite to the libs build outputs when the flag is falsy.

Maybe something like this? I'm sorry if this doesn't make sense. I'm out of my league here. I'm a user-land dev.
```ts
const metadata = context.projectsConfigurations.projects[context.projectName]
const projectRoot = metadata.root;
    
if (!options.buildLibsFromSource) {
    const { dependencies } = calculateProjectDependencies(
      context.projectGraph,
      context.root,
      context.projectName,
      context.targetName,
      context.configurationName
    );
    const tsConfig =
      createTmpTsConfig(
        resolve(projectRoot, 'tsconfig.json'),
        context.root,
        projectRoot,
        dependencies
      );
    registerTsConfigPaths(resolve(projectRoot, tsConfig));
  } else {
    registerTsConfigPaths(resolve(projectRoot, 'tsconfig.json'));
  }
```

viniciusbsneto:
closed #17322

## Current Behavior

It appears that the Vite plugin build executor is missing the same `buildLibsFromSource` logic from `@nx/angular:webpack-browser` that enables incremental builds. This logic is what makes it possible for the plugin to rewrite the path mappings referencing the build outputs instead of always referencing the source files.
## Expected Behavior

An app built with the "@nx/vite:build" executor should have the ability to reference pre-compiled files of its dependency libraries instead of always rebuilding and consuming the source files. This feature enables incremental builds and should be optional, allowing developers to choose whether or not to utilize it.
## Related Issue(s)

Fixes #17322

### DIFF
--- a/docs/generated/packages/vite/executors/build.json
+++ b/docs/generated/packages/vite/executors/build.json
@@ -16,6 +16,11 @@
         "x-completion-type": "directory",
         "x-priority": "important"
       },
+      "buildLibsFromSource": {
+        "type": "boolean",
+        "description": "Read buildable libraries from source instead of building them separately.",
+        "default": true
+      },
       "base": {
         "type": "string",
         "description": "Base public path when served in development or production.",
--- a/docs/generated/packages/vite/executors/dev-server.json
+++ b/docs/generated/packages/vite/executors/dev-server.json
@@ -18,6 +18,11 @@
         "description": "Target which builds the application. Only used to retrieve the configuration as the dev-server does not build the code.",
         "x-priority": "important"
       },
+      "buildLibsFromSource": {
+        "type": "boolean",
+        "description": "Read buildable libraries from source instead of building them separately.",
+        "default": true
+      },
       "proxyConfig": {
         "type": "string",
         "description": "Path to the proxy configuration file.",
--- a/packages/vite/src/executors/build/build.impl.ts
+++ b/packages/vite/src/executors/build/build.impl.ts
@@ -4,6 +4,7 @@ import { build, InlineConfig, mergeConfig } from 'vite';
 import {
   getViteBuildOptions,
   getViteSharedConfig,
+  registerPaths,
 } from '../../utils/options-utils';
 import { ViteBuildExecutorOptions } from './schema';
 import {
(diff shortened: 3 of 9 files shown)


mandarini:
Updated some things:
* Extracted register function in utils
* Added `buildLibsFromSource` to the dev-server as well

mandarini (approved):

mandarini:
Merged it for the time being, we can revert if we decide that it's a behaviour we don't want to support.

jaysoo:
Yeah I think it's fine. 👍

viniciusbsneto:
Thank you @mandarini @jaysoo! A shoutout to @StalkAltan as well for the support in the Slack channel! So with this behavior Vite (or Rollup) should now be able to resolve the libraries build outputs correctly, right? But I wonder how they're consumed and pulled to the final bundle. I imagine it's not a Nx responsability anymore, it's in the Vite realm. So does anyone know where I can get this information?

viniciusbsneto:
> Thank you @mandarini @jaysoo! A shoutout to @StalkAltan as well for the support in the Slack channel! So with this behavior Vite (or Rollup) should now be able to resolve the libraries build outputs correctly, right? But I wonder how they're consumed and pulled to the final bundle. I imagine it's not a Nx responsability anymore, it's in the Vite realm. So does anyone know where I can get this information?

Hello, @mandarini @jaysoo! I'd like to reiterate on this question. I've run some tests using `buildLibsFromSource: false` and it looks like even though the libs build outputs paths are now correctly registered to the node resolutions, their built files are still not skipped at transformation step by Vite (or Esbuild). I'm not 100% sure, but it appears to be the case. At least I don't see any difference in the transformed modules count (from Vite logger) whether using `buildLibsFromSource: false` or not. Could anyone help me understand that? 🙏🏻

jaysoo:
Hey @viniciusbsneto Thanks for your follow-up. I've added this issue to our list of things to look at next week.

viniciusbsneto:
@jaysoo Thanks! In the meantime, should I open another issue with a reproduction?

barbados-clemens:
@viniciusbsneto were you able to make the reproduction issue? was gonna take a look at the issue this week and wanted to have a repo to go off of to make sure we are seeing the same thing. 
Thanks!

PR #17458, opened 2023-06-07
fix(core): avoid terminal flickering when doing dynamic rendering

leosvelperez:
## Current Behavior

When rendering output with the dynamic rendering terminal output lifecycles, there is some flickering in the terminal output.

## Expected Behavior

The terminal output should be smoothly rendered with the dynamic rendering terminal output lifecycles. There should be no flickering.

The current implementation first clears the lines that are going to be replaced/overwritten and then writes the new lines. This seems to produce flickering due to the rendering of the new lines happening in a different rendering cycle than clearing the lines.

The new implementation directly writes over the existing lines and clears the remaining text, if any. Because the clearing now only happens over text that needs to be cleared and not replaced, no flickering occurs.

### run

- **Before:**

https://github.com/nrwl/nx/assets/12051310/9582be13-0ad4-4504-a0fe-b48ab4efe2cd

- **After:**

https://github.com/nrwl/nx/assets/12051310/a199a544-dcd3-4131-a1ee-f534d8855b77

### run-many

- **Before:**

https://github.com/nrwl/nx/assets/12051310/b266de1b-e386-44b8-ac0a-f04b3081a8b4

- **After:**

https://github.com/nrwl/nx/assets/12051310/e9f6ac66-6a58-4b10-9373-4463cdaabba2

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/life-cycles/dynamic-run-many-terminal-output-life-cycle.ts
+++ b/packages/nx/src/tasks-runner/life-cycles/dynamic-run-many-terminal-output-life-cycle.ts
@@ -83,29 +83,31 @@ export async function createRunManyDynamicOutputRenderer({
   let currentFrame = 0;
   let renderIntervalId: NodeJS.Timeout | undefined;
 
-  const clearPinnedFooter = () => {
-    for (let i = 0; i < pinnedFooterNumLines; i++) {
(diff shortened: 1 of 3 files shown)


JamesHenry (approved):
Tested locally on the nx repo itself and can confirm that any incidents of flickering have gone for me using this change! Great work

PR #17465, opened 2023-06-07
fix(storybook): correctly identify if uses vite-based framework

mandarini:
## Current Behavior

Storybook configuration generator will only add vite-related builder config in `.storybook/main.js` if project has `vite` build target.

## Expected Behavior

Storybook configuration generator should also add vite-related builder config in `.storybook/main.js` if project uses `vite` based Storybook framework.

### DIFF
--- a/packages/storybook/src/generators/configuration/configuration.ts
+++ b/packages/storybook/src/generators/configuration/configuration.ts
@@ -107,7 +107,7 @@ export async function configurationGenerator(
     projectIsRootProjectInStandaloneWorkspace(root),
     !!nextBuildTarget,
     compiler === 'swc',
-    !!viteBuildTarget,
+    !!viteBuildTarget || schema.uiFramework.endsWith('-vite'),
     viteConfigFilePath
   );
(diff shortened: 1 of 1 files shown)


Coly010 (approved):

packages/storybook/src/generators/configuration/configuration.ts
```diff
@@ -107,7 +107,7 @@ export async function configurationGenerator(
     projectIsRootProjectInStandaloneWorkspace(root),
     !!nextBuildTarget,
     compiler === 'swc',
-    !!viteBuildTarget,
+    !!viteBuildTarget || schema.uiFramework.endsWith('-vite'),
```
Coly010, line 110: I've approved the PR, but it might be more valid in the future to have a Set containing the known `-vite` frameworks.  This can allow us to determine stability and support for new frameworks, however, it does prevent people using a third party package that we have not specifically added support for. 

It comes down to whether we want to ensure stability or let users break things

PR #17467, opened 2023-06-07
feat(js): add preset for TS standalone workspace

jaysoo:
This PR adds support for `npx create-nx-workspace --preset=ts-standalone`. It will create a standalone workspace with a single `src` directory (same as React, Angular, and Node standalone projects).

This new preset allows a user to create a repo for a single TS/JS package that has no assumptions about platform (Node, browsers, etc.) nor frameworks (React, Angular, etc.). There is already an existing TS/JS monorepo preset (`--preset=ts`), but we're not currently surfacing it in our prompts.

## Notes

- TS monorepo (`--preset=ts`) does not generate a default package. This is already the case now, but should we generate one?
- Since TS standalone needs to generate default project, we also prompt if they want to use TS or JS.

New stack prompt option:
Screenshot 2023-06-06 at 4 59 02 PM

Prompting for TS/JS for standalone:
Screenshot 2023-06-06 at 4 59 35 PM


barbados-clemens (approved):

e2e/jest/src/jest.test.ts
```diff
@@ -132,16 +134,4 @@ describe('Jest', () => {
...
-    const libName = uniq('babel-test-lib');
-    runCLI(
-      `generate @nx/node:lib ${libName} --buildable --importPath=@some-org/babel-test --publishable --babelJest`
-    );
-
-    const cliResults = await runCLIAsync(`test ${libName}`);
-    expect(cliResults.combinedOutput).toContain(
-      'Test Suites: 1 passed, 1 total'
-    );
-  }, 90000);
```
FrozenPandaz, line 146: This seems unrelated?

e2e/nx-misc/src/workspace.test.ts
```diff
@@ -58,7 +58,7 @@ describe('Workspace Tests', () => {
       const lib1 = uniq('mylib');
       const lib2 = uniq('mylib');
       const lib3 = uniq('mylib');
-      runCLI(`generate @nx/js:lib ${lib1}/data-access`);
+      runCLI(`generate @nx/js:lib ${lib1}/data-access --unitTestRunner=jest`);
```
FrozenPandaz, line 61: Why did this have to change?

packages/create-nx-workspace/src/internal-utils/prompts.ts
```diff
@@ -9,7 +8,9 @@ import {
   PackageManager,
   packageManagerList,
 } from '../utils/package-manager';
+import { Preset } from '../utils/preset/preset';
```
FrozenPandaz, line 11: This doesn't seem to be used?

packages/next/src/generators/library/schema.json
```diff
@@ -72,9 +72,8 @@
     },
     "unitTestRunner": {
       "type": "string",
-      "enum": ["jest", "none"],
-      "description": "Test runner to use for unit tests.",
-      "default": "jest"
+      "enum": ["jest", "vitest", "none"],
```
FrozenPandaz, line 75: This seems unrelated?

FrozenPandaz (approved):

e2e/nx-misc/src/workspace.test.ts
jaysoo: We defaulted to vitest, and some of the file checks are jest specific. Just adding this here since vite setup is covered by other e2e tests already.

packages/next/src/generators/library/schema.json
jaysoo: Making sure we can use Vitest everywhere basically.

isaacplmann (approved):

xiongemi (approved):

PR #17471, opened 2023-06-07
fix(core): ensure createPackageJson does not override existing config

meeroslav:
## Current Behavior
The project's package.json is ignored if `createPackageJson` is invoked in any nested folder.

## Expected Behavior
The project's `package.json` should always be respected when calling `createPackageJson`

## Related Issue(s)

Fixes #


FrozenPandaz (approved):

ndcunningham (approved):
LGTM

PR #17472, opened 2023-06-07
feat(core): build project file map with rust

Cammisuli:
## Current Behavior
The project file map is done on the Node side of Nx. This means that we had to use a slower glob to gather project configurations. This was inefficient because we already traversed the file system to gather the file hashes.  

## Expected Behavior
The project graph creation is now done with Rust. This means that when we traverse the workspace to gather hashes, we can also find which file is a project configuration and build the project file map from it. 

Calling the native code will give us `{ projectFileMap: ProjectFileMap, configFiles: string[], globalFiles: string[] }` and we can use this information to build out the necessary structures for other parts of Nx. 

## Related Issue(s)

Fixes #


packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -152,11 +159,19 @@ async function processCollectedUpdatedAndDeletedFiles() {
...
+
+    let nxJson = new Workspaces(workspaceRoot).readNxJson();
+
+    const { configFiles } = await getWorkspaceFiles(workspaceRoot, nxJson);
+
+    let projectConfigurations = createProjectConfigurations(
+      workspaceRoot,
+      nxJson,
+      configFiles
+    );
```
FrozenPandaz, line 171: Combine these 2 functions and make sure it is invalidated.

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -167,20 +182,47 @@ async function processCollectedUpdatedAndDeletedFiles() {
...
+      if (projectFileMapWithFiles) {
+        projectFileMapWithFiles = updateProjectFileMap(
+          projectConfigurations,
+          projectFileMapWithFiles.projectFileMap,
+          projectFileMapWithFiles.allWorkspaceFiles,
+          updatedFiles,
+          deletedFiles
+        );
+      } else {
+      }
```
FrozenPandaz, line 210: Below should be in here

packages/nx/src/native/tests/hasher.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  it('should hash content', async () => {
+    expect(hashArray).toBeDefined();
+
+    expect(hashArray(['one', 'two'])).toEqual('10960201262927338690');
+  });
+
+  it('should create an instance of NativeHasher', () => {
+    // const nativeHasher = new NativeFileHasher('/root');
+    // expect(nativeHasher instanceof NativeFileHasher).toBe(true);
+  });
```
FrozenPandaz, line 28: rm

FrozenPandaz (changes requested):

FrozenPandaz (changes requested):

leosvelperez (approved):

FrozenPandaz (approved):

PR #17474, opened 2023-06-08
docs(angular): nx and angular cli guide update

isaacplmann:
Update the Nx and the Angular CLI guide


juristr (changes requested):
Just a minor comment, otherwise looks good to me.

juristr (approved):
LGTM

PR #17495, opened 2023-06-08
[do not merge] fix(vite): properly replace env variables

### LINKED ISSUE #17461: Keys defined in .env files that reference another env variable for vite are not working
### Current Behavior

If you have this in your .env
```
KEY=123
API_PATH=/api/v2/$KEY
```
and run `nx build your-app` which runs `"build": "tsc && vite build"` from the `package.json` of `your-app` with its depencies. The passed envs will have these values:
```
KEY=123
API_PATH=/api/v2/$KEY
```

### Expected Behavior

Expectation would be for the values to be:
```
KEY=123
API_PATH=/api/v2/123
```
the problem is that they are passed to runtime so vite does not do anything to those dollar signs anymore and that causes havoc.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create `.env` file
2. Create KEY=123 value
3. Create API_PATH value that references the KEY value
4. Do referenching like this API_PATH=/api/v2/$KEY
5. See that vite is not processing your .env as you would expect

### Nx Report

```shell
Node   : 16.14.2
 OS     : darwin arm64
 pnpm   : 7.27.1
 Hasher : Native
   
 nx (global)  : 16.3.2
 nx           : 16.3.2
 @nrwl/tao    : 16.3.2
 typescript   : 4.9.5
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

mandarini:
## Current Behavior

Current behaviour is explained in #17461

If I have a `.env` file like this:
```
VITE_KEY=123
VITE_KEY2=test$VITE_KEY
```

the result will be `123` and `test$VITE_KEY`.

## Expected Behavior

The result should be `123` and `test123`, as described in the Vite documentation: https://vitejs.dev/guide/env-and-mode.html

This PR introduces a new function `replaceEnvVarsWithinEnv` that enhances the way we handle environment variables in our Vite Build Executor. Previously, references to other environment variables were not being replaced in the Nx executor as they would be when using `vite build`. With this change, we now scan all environment variables for references to other variables (denoted by a $ prefix) and replace them with their corresponding values. The function handles edge cases such as escaped variable references and cases where the referenced variable is undefined, in consistency with Vite.

## Related Issue(s)

Fixes #17461

## Note

I noticed that the same "issue" exists with webpack, which maybe makes sense, since we're calling these builders programmatically, and the substitution maybe happens one step before, when invoking them directly?

### DIFF
--- a/packages/vite/src/executors/build/build.impl.ts
+++ b/packages/vite/src/executors/build/build.impl.ts
@@ -16,7 +16,11 @@ import {
 import { existsSync, writeFileSync } from 'fs';
 import { resolve } from 'path';
 import { createAsyncIterable } from '@nx/devkit/src/utils/async-iterable';
-import { registerPaths, validateTypes } from '../../utils/executor-utils';
+import {
+  registerPaths,
+  validateTypes,
+  replaceEnvVarsWithinEnv,
+} from '../../utils/executor-utils';
 
 export async function* viteBuildExecutor(
   options: ViteBuildExecutorOptions,
@@ -26,6 +30,7 @@ export async function* viteBuildExecutor(
     context.projectsConfigurations.projects[context.projectName].root;
 
   registerPaths(projectRoot, options, context);
+  replaceEnvVarsWithinEnv();
 
   const normalizedOptions = normalizeOptions(options);
 
--- a/packages/vite/src/executors/dev-server/dev-server.impl.ts
+++ b/packages/vite/src/executors/dev-server/dev-server.impl.ts
@@ -11,7 +11,10 @@ import {
 
 import { ViteDevServerExecutorOptions } from './schema';
 import { ViteBuildExecutorOptions } from '../build/schema';
-import { registerPaths } from '../../utils/executor-utils';
+import {
+  replaceEnvVarsWithinEnv,
+  registerPaths,
+} from '../../utils/executor-utils';
 
 export async function* viteDevServerExecutor(
   options: ViteDevServerExecutorOptions,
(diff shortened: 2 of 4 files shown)


tlehtimaki:
Great to see quick action for this, looks good to me. I was thinking is there any possibility this could have a side-effect to anything else?

mandarini:
@tlehtimaki do you have something in mind, that it could have a side-effect on? I will filter the env vars to only pick up those starting with `VITE_`, you are right on that note.

jaysoo:
I would use the same package that Vite uses `dotenv-expand` as we probably don't want to maintain the interpolation logic ourselves. We can scope it to just Vite for now (i.e. use it just for `@nx/vite:build` and `@nx/vite:dev-server`).

And we should pose it as a question on adoption for Nx core because it's weird to have inconsistencies between the different stacks.

tlehtimaki:
> I would use the same package that Vite uses `dotenv-expand` as we probably don't want to maintain the interpolation logic ourselves. We can scope it to just Vite for now (i.e. use it just for `@nx/vite:build` and `@nx/vite:dev-server`).
> 
> And we should pose it as a question on adoption for Nx core because it's weird to have inconsistencies between the different stacks.

That sounds like a good discussion to have. I would see it as a good overall feature and using `dotenv-expand` would be wise from maintenance perspective.

Other question here is that should one be able to have those environment variables to go as a pass through to the commands that vite could do it’s magic. That might pose other challenges though and complicate logic for .env handling. I would vote for having the expansion available.

mandarini:
Added the `[do not merge]` flag to avoid any mistakes!

mandarini:
Closing this for now.

PR #17496, opened 2023-06-08
feat(vite): allow passing multiple test files to vite:test

jwalton9:
## Current Behavior

You can only pass one testfile to run to `vite:test` 

## Expected Behavior

Able to pass multiple test files to run to `vite:test` this is useful for CI for example when using test splitting

### DIFF
--- a/packages/vite/migrations.json
+++ b/packages/vite/migrations.json
@@ -25,7 +25,7 @@
       "implementation": "./src/migrations/update-16-0-0-add-nx-packages/update-16-0-0-add-nx-packages"
     },
     "update-16-4-0-test-file-config": {
-      "version": "16.4.0-beta.2",
+      "version": "16.4.0-beta.8",
       "description": "Changes the testFile config in the vite:test exectutor from a string to an array of strings",
       "cli": "nx",
       "implementation": "./src/migrations/update-16-4-0-test-file-config/update-16-4-0-test-file-config"
--- a/packages/vite/migrations.json
+++ b/packages/vite/migrations.json
@@ -24,11 +24,11 @@
       "description": "Replace @nrwl/vite with @nx/vite",
       "implementation": "./src/migrations/update-16-0-0-add-nx-packages/update-16-0-0-add-nx-packages"
     },
-    "update-16-4-0-test-file-config": {
-      "version": "16.4.0-beta.10",
+    "update-16-4-1-test-file-config": {
+      "version": "16.4.1-beta.0",
       "description": "Changes the testFile config in the vite:test exectutor from a string to an array of strings",
       "cli": "nx",
-      "implementation": "./src/migrations/update-16-4-0-test-file-config/update-16-4-0-test-file-config"
+      "implementation": "./src/migrations/update-16-4-1-update-test-file-config/update-16-4-1-test-file-config"
     }
   },
   "packageJsonUpdates": {
--- a/packages/vite/migrations.json
+++ b/packages/vite/migrations.json
@@ -25,7 +25,7 @@
       "implementation": "./src/migrations/update-16-0-0-add-nx-packages/update-16-0-0-add-nx-packages"
     },
     "update-16-4-0-test-file-config": {
-      "version": "16.4.0-beta.8",
+      "version": "16.4.0-beta.10",
       "description": "Changes the testFile config in the vite:test exectutor from a string to an array of strings",
       "cli": "nx",
       "implementation": "./src/migrations/update-16-4-0-test-file-config/update-16-4-0-test-file-config"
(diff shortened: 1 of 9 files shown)


packages/vite/src/migrations/update-16-4-0-update-test-file-config/update-16-4-0-test-file-config.ts
```diff
@@ -0,0 +1,28 @@
...
+
+          if (target.configurations) {
+            for (const configuration of Object.values(target.configurations)) {
+              if (configuration.testFile) {
+                configuration.testFile = [configuration.testFile];
+              }
+            }
+          }
+        }
+      }
```
barbados-clemens, line 23: You can use `forEachExecutorOption` instead.

```ts
// might just be from '@nx/devkit' now?
import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';

const projects = getProjects(tree);

forEachExecutorOptions<VitestExecutorOptions>(tree, '@nx/vite:test', (options, projectName, targetName, configuration) => {
  const projectConfig = projects.get(projectName);

  // update testFile option if preset

  updateProjectConfiguration(tree, projectName, projectConfig)
})
```

packages/vite/src/migrations/update-16-4-0-update-test-file-config/update-16-4-0-test-file-config.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace({ layout: 'apps-libs' });
+    addProjectConfiguration(tree, 'vitest', {
+      root: 'vitest',
+      targets: {
+        test: {
+          executor: '@nx/vite:test',
+          options: {
+            testFile: 'test-file.ts',
+          },
```
barbados-clemens, line 22: you should add a test that makes sure it doesn't modify a project that doesn't have a `testFile` option.

packages/vite/src/executors/test/schema.d.ts
```diff
@@ -8,5 +8,5 @@ export interface VitestExecutorOptions {
   update?: boolean;
   reportsDirectory?: string;
   coverage?: boolean;
-  testFile?: string;
+  testFile?: string[];
```
barbados-clemens, line 11: now that this is an array does it make sense to rename this to be plural? `testFiles` 🤔

packages/vite/migrations.json
```diff
@@ -23,6 +23,12 @@
       "version": "16.0.0-beta.1",
       "description": "Replace @nrwl/vite with @nx/vite",
       "implementation": "./src/migrations/update-16-0-0-add-nx-packages/update-16-0-0-add-nx-packages"
+    },
+    "update-16-4-0-test-file-config": {
+      "version": "16.4.0-beta.2",
```
barbados-clemens, line 28: you'll want to update this to [16.4.0-beta.5](https://www.npmjs.com/package/nx/v/16.4.0-beta.5)

packages/vite/src/migrations/update-16-4-0-update-test-file-config/update-16-4-0-test-file-config.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
barbados-clemens, line 1: remove eslint comment

barbados-clemens (commented):
Hi thanks for the PR, just a couple things that could be improved!

packages/vite/src/executors/test/schema.d.ts
jwalton9: I'm not sure. I think this makes sense when using the option in `project.json` e.g.
```json
{
   "testFiles": ["tests/foo.test.tsx", "tests/bar.test.tsx"]
}
```
But I prefer `testFile` when using the CLI
```
nx run app:test --testFIle=foo.test.tsx --testFile=bar.test.tsx
```
over 
```
nx run app:test --testFIles=foo.test.tsx --testFiles=bar.test.tsx
```

It just personal preference, so happy to go with `testFiles` but just making the case for keeping as is

jwalton9: Great, thanks for the suggestion, I've updated to use `testFiles` with a `testFile` alias

barbados-clemens (approved):

barbados-clemens:
updated the tests to handle the option rename and updated  the target for the migration

packages/vite/migrations.json
```diff
@@ -23,6 +23,12 @@
       "version": "16.0.0-beta.1",
       "description": "Replace @nrwl/vite with @nx/vite",
       "implementation": "./src/migrations/update-16-0-0-add-nx-packages/update-16-0-0-add-nx-packages"
+    },
+    "update-16-4-1-test-file-config": {
+      "version": "16.4.1-beta.0",
```
ndcunningham, line 28: Nx 16.4 is out so this would an update

ndcunningham (changes requested):

ndcunningham (approved):

PR #17503, opened 2023-06-08
cleanup(js): improve the tsc executor batch implementation

leosvelperez:
## What it does

- [X] Keep the generated tsconfig with the project references in memory instead of writing/reading to/from the filesystem.
- [X] Report each task result as they get executed instead of a single report with the whole batch results when all tasks finish.
- [X] Detach the Typescript compilation from the batch execution logic.
- [X] Handle common target composition setup scenarios.
- [X] Fix an issue with the task scheduler when there are interdependent tasks and batches.


packages/js/src/executors/tsc/lib/get-in-memory-tsconfig.ts
```diff
@@ -0,0 +1,55 @@
...
+  taskName: string,
+  taskOptions: NormalizedExecutorOptions,
+  context: ExecutorContext
+): TypescriptInMemoryTsConfig {
+  const originalTsConfigPath = taskOptions.tsConfig;
+
+  const projectReferences: { path: string }[] = [];
+  for (const depTask of context.taskGraph.dependencies[taskName]) {
+    // if included in the provided map, use it
+    if (taskOptionsMap[depTask]) {
```
jaysoo, line 19: There is an edge case here if the dep doesn't have `tsConfig` option. It shouldn't come up if all deps are using tsc executor, but it is technically possible. Maybe a check on `if (taskOptionsMap[depTask]?.tsConfig)`?

And then at the end of the `for` block, throw and error if no `tsConfig` is found for the dep (either from task or cache).

leosvelperez: Not really. At least not from the intended and current usage. The `taskOptionsMap` object contains only the options of tasks that are part of the batch execution, so all those tasks have the same executor (tsc). The `tsConfig` option is required for that executor, so if not provided, the schema validation should throw before getting here.

Dependencies with a different executor are executed before and are not part of the batch execution.

jaysoo (approved):

mandarini (approved):

packages/nx/src/tasks-runner/forked-process-task-runner.ts
```diff
@@ -63,6 +59,10 @@ export class ForkedProcessTaskRunner {
         });
         this.processes.add(p);
 
+        // ensure the child process is killed when the parent exits
+        process.on('exit', () => p.kill());
+        process.on('SIGTERM', () => p.kill());
```
FrozenPandaz, line 64: Add 1 sweep in the `setupProcessEventListeners` method.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #17511, opened 2023-06-08
docs(angular): tailwind get started section

isaacplmann:
Adds a getting started section to the angular tailwind guide


juristr (approved):
Merging because this already improves this recipe quite a lot. But it needs more work as discussed in Slack

PR #17512, opened 2023-06-08
fix(core): properly hash tasks that depend on outputs of other tasks

FrozenPandaz:
## Current Behavior

1. All files in a output dir are hashed even if they don't match the glob
2. Tasks are hashed upfront even if they depend on the outputs of a dependent task

## Expected Behavior

1. Only files that match the glob are hashed when hashing outputs
2. Tasks are not hashed upfront if they depend on the outputs of dependent tasks

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/native_hasher.rs
+++ b/packages/nx/src/native/native_hasher.rs
@@ -1,7 +1,10 @@
 #![allow(unused)]
 
+use anyhow::anyhow;
 use crossbeam_channel::unbounded;
+use globset::{Glob, GlobSetBuilder};
 use ignore::WalkBuilder;
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::path::Path;
 use std::thread::{self, available_parallelism};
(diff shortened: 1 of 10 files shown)


packages/nx/src/native/native_hasher.rs
```diff
@@ -97,6 +99,75 @@ fn hash_files(workspace_root: String) -> HashMap<String, String> {
...
+
+    let cpus = available_parallelism().map_or(2, |n| n.get()) - 1;
+
+    let mut walker = WalkBuilder::new(&directory);
+    walker.hidden(false);
+
+    let (sender, receiver) = unbounded::<(String, Vec<u8>)>();
+
+    let receiver_thread = thread::spawn(move || {
+        let mut collection: Vec<String> = Vec::new();
```
FrozenPandaz, line 121: This is not correct, the order of these hashes must be deterministic. This will need to return the `Vec<FileData>`

packages/nx/src/native/native_hasher.rs
```diff
@@ -97,6 +120,89 @@ fn hash_files(workspace_root: String) -> HashMap<String, String> {
...
+        return Ok(None);
+    }
+
+    // Sort the file data so that its in deterministically ordered by file path
+    hashes.sort();
+
+    let sorted_file_hashes: Vec<String> = hashes
+        .iter()
+        .map(|file_data| file_data.hash.clone())
+        .collect();
```
Cammisuli, line 202: We can use `into_iter` here so that we can move ownership from hashes to the sorted_file_hashes. This allows us to remove the clone and remove the need for an additional allocation 🙂 
```suggestion
    let sorted_file_hashes: Vec<String> =
        hashes.into_iter().map(|file_data| file_data.hash).collect();
```

Cammisuli (approved):
native changes are good

meeroslav (approved):

PR #17517, opened 2023-06-09
fix(vite): properly find projectRoot

mandarini:
It would throw an error if it could not find the `projectGraph`, but also let's keep the way we get `projectRoot` consistent across executors!


mandarini:
@brandonroberts that should fix your issue!

brandonroberts:
Thanks @mandarini!

ndcunningham (approved):

PR #17525, opened 2023-06-09
fix(core): ignore ignored .gitignore

wSedlacek:
## Current Behavior

If a change happens in a `.gitignore` it will kill the daemon.
For example if a `apps/myapp/ios/.gitignore` is changed like `@nx/expo:prebuild` does then the watch server will be killed.

## Expected Behavior

If `apps/myapp/ios` is already ignored then a change to the `.gitignore` in it will not change the actuall ignored files and thus the server should not restart.

### DIFF
--- a/packages/nx/src/daemon/server/watcher.ts
+++ b/packages/nx/src/daemon/server/watcher.ts
@@ -145,7 +145,9 @@ export async function subscribeToWorkspaceChanges(
           path: normalizePath(relative(workspaceRoot, event.path)),
         };
         if (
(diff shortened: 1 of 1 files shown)


Cammisuli:
Hey @wSedlacek sorry for the delay in responding to this PR. But I don't think it's necessary because we do not use parcel's watcher by default anymore. 

We have a rust watcher that already handles ignored gitignore files. Does this issue still happen for you with the latest versions of Nx?

wSedlacek:
I haven't noticed the issue in a while. Thank you for your response!

PR #17530, opened 2023-06-09
feat(linter): allow globs in onlyDependOnLibsWithTags eslint-plugin configuration option

### LINKED ISSUE #15264: Add globbing to ES Lint `onlyDependOnLibsWithTags`
## Description

Currently, only special-case `*` globbing exists for `onlyDependOnLibsWithTags`.  Add the ability to use `*` to pattern match with partial strings.

Note: I filed this as a request instead of a bug because this feature does not exist, but the current Nx implementation feels like a miss.  See all of these places that support globbing.

- `allow` globs, cf. https://nx.dev/recipes/other/tags-allow-list#tags-allow-list. 
- `sourceTag` globs.  I did not find that in the docs, but it seems to work, when testing locally.
- `onlyDependOnLibsWithTags` currently supports `*` as a special case, cf. https://nx.dev/core-features/enforce-project-boundaries#enforce-project-boundaries
- `bannedExternalImports` globs, cf. https://nx.dev/recipes/other/ban-external-imports#ban-external-imports

If you decide not to implement this, I'd recommend making the documentation more clear that `onlyDependOnLibsWithTags` doesn't support any globbing other than the special case.  My assumption was that it would because everything else globs and you note `*` special case in the docs.

## Motivation

We have a few tag types.  They're perhaps overly picayune, but we're using them to help developers focus in on making more, smaller libraries.  As such we're using a two part tag `main:specifier`, such as `util:types` for a shared, types-only library or `util:data-access`.  Given our wide usage of this pattern we want to be able to set `onlyDependOnLibsWithTags` to `util:*` for things like features or apps.  As we add more `main:new-specifier` libraries, we don't want to have to update the ES Lint tag rules.

Splitting the `main` and `specifier` portions into independent tags creates the same problem.  As we add new things (new specifiers), we still have to go update our ES Lint rules.

ericyd:
## Current Behavior
`onlyDependOnLibsWithTags` does not handle glob patterns in the tags list.  However, several other options for the `eslint-plugin` do support glob patterns.

## Expected Behavior
Update `onlyDependOnLibsWithTags` to handle glob patterns in tag lists. This allows a tagged package to depend on any packages matching the glob, rather than only specific tags.

Note: this behavior was already available, via Regex support. However, the ability to use regular expressions in the tag list was undocumented. This PR adds documentation for the new glob pattern as well as the existing regex pattern.

## Related Issue(s)

Fixes #15264

## Questions to maintainers

1. When I followed the instructions in the CONTRIBUTING.md file, my pnpm-lock file was modified to version 6.1. Should [this command be updated](https://github.com/nrwl/nx/blob/51cadbd4df23a1324cd1f94660f1480c6900640f/CONTRIBUTING.md?plain=1#L60)?
2. When I first cloned the repo, VSCode automatically prompted me to open in the dev container. Great! It also automatically ran `pnpm i` and it added 10k+ files to a `.pnpm-store` directory. Should this directory be added to `.gitignore`? This didn't happen when I ran `pnpm install` outside of the devcontainer, so I'm not clear where the discrepancy comes from

### DIFF
--- a/docs/shared/core-features/enforce-project-boundaries.md
+++ b/docs/shared/core-features/enforce-project-boundaries.md
@@ -94,7 +94,7 @@ First, use your project configuration (in `project.json` or `package.json`) to a
 {% /tab %}
 {% /tabs %}
 
-Next you should update your root lint configuration:
+Next, you should update your root lint configuration:
 
 - If you are using **ESLint** you should look for an existing rule entry in your root `.eslintrc.json` called `"@nx/enforce-module-boundaries"` and you should update the `"depConstraints"`:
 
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/utils/runtime-lint-utils.ts
```diff
@@ -103,6 +103,12 @@ function hasTag(proj: ProjectGraphProjectNode, tag: string): boolean {
     return (proj.data.tags || []).some((t) => regex.test(t));
   }
 
+  // if the tag is a glob, check if the project matches the glob prefix
+  if (tag.endsWith('*')) {
+    const prefix = tag.substring(0, tag.length - 1);
+    return (proj.data.tags || []).some((t) => t.startsWith(prefix));
+  }
```
ericyd, line 110: I know this is a pretty naïve implementation of glob support, but since this option already supports regex I figured it didn't really need to be ultra robust. Happy to revisit if desired

docs/shared/core-features/enforce-project-boundaries.md
```diff
@@ -131,7 +131,31 @@ Next you should update your root lint configuration:
 
 With these constraints in place, `scope:client` projects can only depend on other `scope:client` projects or on `scope:shared` projects. And `scope:admin` projects can only depend on other `scope:admin` projects or on `scope:shared` projects. So `scope:client` and `scope:admin` cannot depend on each other.
 
-Projects without any tags cannot depend on any other projects. If you add the following, projects without any tags will be able to depend on any other project.
+Projects without any tags cannot depend on any other projects, unless you allow all tags (see below).
```
meeroslav, line 134: This should not be changing.

Project with no tag should not be allowed to depend on any other project.

meeroslav (approved):
Added some minor improvements. Thank you!

meeroslav (approved):

isaacplmann (approved):

PR #17532, opened 2023-06-09
feat(linter): add generator for converting to flat config

meeroslav:
A new generator to convert workspace from standard lint configs to new flat eslint config.

Based on: https://github.com/typescript-eslint/typescript-eslint/pull/6836
According to: https://eslint.org/blog/2022/08/new-config-system-part-2/
Following: https://eslint.org/docs/latest/use/configure/migration-guide

- [X] Extend `updateProjectConfiguration` to update `package.json` as well
- [X] Create generator shell with options
- [x] Add unit tests
- [X] Implement nx.json conversion
- [X] Implement project config conversion
- [x] Convert config to new format
  - [X] Covert ignores
  - [X] Convert extends
  - [X] Convert overrides
  - [X] Convert settings
  - [X] Convert rules
  - [x] Convert parsers
  - [x] Convert plugins
  - [x] Linter Options
  - [x] Predefined and Shareable Configs
  - [x] Language Options
- [x] Add E2E test
- [x] Make improvements after sync with ESLint team

### DIFF
--- a/packages/linter/src/generators/convert-to-flat-config/__snapshots__/generator.spec.ts.snap
+++ b/packages/linter/src/generators/convert-to-flat-config/__snapshots__/generator.spec.ts.snap
@@ -1,12 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`convert-to-flat-config generator should add env configuration 1`] = `
-"import nxEslintPlugin from '@nx/eslint-plugin';
-import { FlatCompat } from '@eslint/eslintrc';
+"const nxEslintPlugin = require('@nx/eslint-plugin');
+const { FlatCompat } = require('@eslint/eslintrc');
 const eslintrc = new FlatCompat({
   baseDirectory: __dirname,
 });
-export default [
+module.exports = [
   { plugins: { '@nx': nxEslintPlugin } },
   {
     languageOptions: {
@@ -51,12 +51,12 @@ export default [
 `;
 
 exports[`convert-to-flat-config generator should add global gitignores 1`] = `
-"import nxEslintPlugin from '@nx/eslint-plugin';
-import { FlatCompat } from '@eslint/eslintrc';
+"const nxEslintPlugin = require('@nx/eslint-plugin');
+const { FlatCompat } = require('@eslint/eslintrc');
 const eslintrc = new FlatCompat({
   baseDirectory: __dirname,
 });
-export default [
+module.exports = [
   { plugins: { '@nx': nxEslintPlugin } },
   {
     files: ['*.ts', '*.tsx', '*.js', '*.jsx'],
@@ -93,13 +93,13 @@ export default [
 `;
 
 exports[`convert-to-flat-config generator should add parser 1`] = `
-"import nxEslintPlugin from '@nx/eslint-plugin';
-import typescriptEslintParser from '@typescript-eslint/parser';
-import { FlatCompat } from '@eslint/eslintrc';
+"const nxEslintPlugin = require('@nx/eslint-plugin');
+const typescriptEslintParser = require('@typescript-eslint/parser');
+const { FlatCompat } = require('@eslint/eslintrc');
 const eslintrc = new FlatCompat({
   baseDirectory: __dirname,
 });
-export default [
+module.exports = [
   { plugins: { '@nx': nxEslintPlugin } },
   { languageOptions: { parser: typescriptEslintParser } },
   {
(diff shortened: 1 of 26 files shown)


mandarini (approved):

packages/linter/src/generators/convert-to-flat-config/converters/json-converter.ts
```diff
@@ -0,0 +1,210 @@
...
+      );
+
+      importsList.push(importStatement);
+      configBlocks.push(
+        ts.factory.createSpreadElement(ts.factory.createIdentifier(localName))
+      );
+    });
+  // add plugin extends
+  // TODO(meeroslav): Check if this is the recommended way of doing it
+  const pluginExtends = extendsConfig.filter((e) => e.startsWith('plugin:'));
```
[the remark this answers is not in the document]
JamesHenry: I think maybe anything non-relative is a plugin?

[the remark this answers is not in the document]
meeroslav: Yes, I agree. But I want to check with Nicholas if it should be prefixed with `plugin:` after migration.

packages/linter/src/generators/convert-to-flat-config/__snapshots__/generator.spec.ts.snap
```diff
@@ -0,0 +1,140 @@
...
+const eslintrc = new FlatCompat({
+    baseDirectory: __dirname
+});
+
+export default [
+    {
+        ignores: ["**/*"]
+    },
+    {
+        ignores: ["node_modules"]
```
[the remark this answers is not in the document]
meeroslav: We currently generate it in `.eslintingnore`, so the generator just maps it.

juristr (approved):
LGTM

Coly010 (approved):

packages/linter/src/generators/convert-to-flat-config/generator.ts
```diff
@@ -0,0 +1,115 @@
...
+  // install missing packages
+
+  if (!options.skipFormat) {
+    await formatFiles(tree);
+  }
+}
+
+export default convertToFlatConfigGenerator;
+
+function convertRootToFlatConfig(tree: Tree) {
```
AgentEnder, line 46: I feel like this function should accept the path to the root eslint file as an argument. We've already located it before calling the function.

```suggestion
function convertRootToFlatConfig(tree: Tree, rootEsLintFIlePath: string) {
```

packages/linter/src/generators/convert-to-flat-config/generator.ts
```diff
@@ -0,0 +1,115 @@
...
+      '.eslintrc.base.json',
+      'eslint.base.config.js'
+    );
+  }
+  if (tree.exists('.eslintrc.json')) {
+    convertConfigToFlatConfig(tree, '', '.eslintrc.json', 'eslint.config.js');
+  }
+}
+
+function convertProjectToFlatConfig(
```
AgentEnder, line 60: This function needs to take into account `targetDefaults`. Imagine things like:

`projectConfig.targets.lint === {}`, because all of the config lives in nx.json

packages/linter/src/generators/convert-to-flat-config/generator.ts
```diff
@@ -0,0 +1,115 @@
...
+    );
+  }
+}
+
+// update names of eslint files in nx.json
+// and remove eslintignore
+function updateNxJsonConfig(tree: Tree) {
+  if (tree.exists('nx.json')) {
+    updateJson(tree, 'nx.json', (json: NxJsonConfiguration) => {
+      if (json.targetDefaults?.lint?.inputs) {
```
AgentEnder, line 93:
```suggestion
      if (json.targetDefaults?.lint?.inputs || json.targetDefaults?.['@nx/linter:eslint']?.inputs) {
```

packages/linter/src/generators/init/init.ts
```diff
@@ -25,13 +25,14 @@ export interface LinterInitOptions {
   rootProject?: boolean;
 }
 
-function addTargetDefaults(tree: Tree) {
+function addTargetDefaults(tree: Tree, isRootProject: boolean) {
```
AgentEnder, line 28: I don't see this arg used, do we need it?

AgentEnder (changes requested):
Some small changes

packages/linter/src/generators/convert-to-flat-config/generator.ts
meeroslav: This function converts both `.eslintrc.base.json` and `.eslintrc.json`. The found `rootEsLintFIlePath` could be any one of those two. The second condition on line 56 is not needed, but we still need to check for base one.

But I guess it's cheaper to check the file name's suffix then check for file's existence

AgentEnder (approved):

packages/linter/src/generators/convert-to-flat-config/__snapshots__/generator.spec.ts.snap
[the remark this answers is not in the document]
meeroslav: The new flat config adds that by default, so I removed it

PR #17534, opened 2023-06-09
chore(testing): always log e2e nx-release error

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/utils/global-setup.ts
+++ b/e2e/utils/global-setup.ts
@@ -1,5 +1,5 @@
 import { join } from 'path';
-import { ChildProcess, execSync, fork } from 'child_process';
+import { ChildProcess, exec, fork } from 'child_process';
 import { tmpdir } from 'tmp';
 import { existsSync } from 'fs-extra';
 import { Config } from '@jest/types';
@@ -57,8 +57,25 @@ export default async function (globalConfig: Config.ConfigGlobals) {
     !existsSync('./build')
   ) {
     console.log('Publishing packages to local registry');
-    execSync('pnpm nx-release --local major', {
-      stdio: isVerbose ? 'inherit' : 'ignore',
+    await new Promise<void>((res, rej) => {
+      const publishProcess = exec('pnpm nx-release --local major');
+      let logs = Buffer.from('');
+      if (isVerbose) {
+        publishProcess?.stdout?.pipe(process.stdout);
+        publishProcess?.stderr?.pipe(process.stderr);
+      } else {
+        publishProcess?.stdout?.on('data', (data) => (logs += data));
+        publishProcess?.stderr?.on('data', (data) => (logs += data));
+      }
+      publishProcess.on('exit', (code) => {
+        if (code && code > 0) {
+          if (!isVerbose) {
+            console.log(logs.toString());
+          }
+          rej(code);
+        }
+        res();
+      });
     });
   }
 }


e2e/utils/global-setup.ts
```diff
@@ -57,8 +57,19 @@ export default async function (globalConfig: Config.ConfigGlobals) {
...
+    } catch (err) {
+      if (err?.stdout) {
+        process.stdout.write(err.stdout.toString());
+      }
+      if (err?.stderr) {
+        process.stderr.write(err.stderr.toString());
+      }
+      console.log(err);
+      throw err;
+    }
```
FrozenPandaz, line 73: Your solution would have all of the `stdout` followed by all of the `stderr` instead of interlaced which would be normally how it is printed. Below would be better

```suggestion
    await new Promise<void>((res) => {
      const publishProcess = exec('pnpm nx-release --local major');

      let logs = Buffer.from('');
      if (isVerbose) {
        publishProcess.stdout.pipe(process.stdout);
        publishProcess.stderr.pipe(process.stderr);
      } else {
        publishProcess.stdout.on('data', (data) => (logs += data));
        publishProcess.stderr.on('data', (data) => (logs += data));
      }
      publishProcess.on('exit', (code) => {
        if (!isVerbose && code > 0) {
          console.log(logs.toString());
        }
        res();
      });
```

FrozenPandaz (changes requested):

e2e/utils/global-setup.ts
xiongemi: changed

mandarini (approved):

e2e/utils/global-setup.ts
```diff
@@ -57,8 +57,22 @@ export default async function (globalConfig: Config.ConfigGlobals) {
...
+        publishProcess?.stderr?.pipe(process.stderr);
+      } else {
+        publishProcess?.stdout?.on('data', (data) => (logs += data));
+        publishProcess?.stderr?.on('data', (data) => (logs += data));
+      }
+      publishProcess.on('exit', (code) => {
+        if (!isVerbose && code && code > 0) {
+          console.log(logs.toString());
+        }
+        res();
```
FrozenPandaz, line 74: 1 more thing, we need to reject if the code is > 0

FrozenPandaz (changes requested):

e2e/utils/global-setup.ts
```diff
@@ -57,8 +58,23 @@ export default async function (globalConfig: Config.ConfigGlobals) {
...
+        publishProcess?.stdout?.pipe(process.stdout);
+        publishProcess?.stderr?.pipe(process.stderr);
+      } else {
+        publishProcess?.stdout?.on('data', (data) => (logs += data));
+        publishProcess?.stderr?.on('data', (data) => (logs += data));
+      }
+      publishProcess.on('exit', (code) => {
+        if (!isVerbose && code && code > 0) {
+          console.log(logs.toString());
+          rej(code);
```
FrozenPandaz, line 74: This should move outside of the `if` because even if its `verbose`, we should still exit.

FrozenPandaz (changes requested):

e2e/utils/global-setup.ts
xiongemi: changed

FrozenPandaz (approved):

PR #17551, opened 2023-06-12
fix(misc): mark name as required for ci-workflow generator

MaxKless:
In Nx Console, it looks weird if you have to expand all options when there's only two overall and there's no reason to not see both at the same time.

### DIFF
--- a/packages/workspace/src/generators/ci-workflow/schema.json
+++ b/packages/workspace/src/generators/ci-workflow/schema.json
@@ -34,5 +34,5 @@
       "pattern": "^[a-zA-Z].*$"
     }
   },
-  "required": ["ci"]
+  "required": ["ci", "name"]
 }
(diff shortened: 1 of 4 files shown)


meeroslav (approved):
LGTM

jaysoo (approved):

packages/workspace/src/generators/ci-workflow/schema.json
```diff
@@ -31,7 +31,8 @@
       },
       "default": "CI",
       "x-prompt": "How should we name your workflow?",
-      "pattern": "^[a-zA-Z].*$"
+      "pattern": "^[a-zA-Z].*$",
+      "x-priority": "important"
```
FrozenPandaz, line 35: Is it not `required`?

FrozenPandaz (changes requested):

packages/workspace/src/generators/ci-workflow/schema.json
MaxKless: There's a default value of `CI` that's used when you don't provide a `name`... I guess we could remove that and make `name` required but then again sometimes folks might want to create a workflow and don't care what it's called.

FrozenPandaz (approved):

PR #17560, opened 2023-06-12
feat(testing): add support for additional environment options when generating vite config

jaysoo:
Vite has a bunch of built-in environments (https://vitest.dev/config/#environment), so this PR makes it possible to configure for those. The default is set to `jsdom` match previous behavior, but for `@nx/js:lib` it is set to `node` to make it platform-agnostic.

This is useful once we add the TS/JS preset so that we're not assuming a web package. https://github.com/nrwl/nx/pull/17467

## Related Issue(s)

Fixes #


barbados-clemens (approved):
Looks good. only thing might be if we want to type the environment args a little more strict to be helpful when viewing what values are expected? non blocking.

ndcunningham (approved):

mandarini (approved):

PR #17562, opened 2023-06-12
fix(testing): update default webpack config for react CT

barbados-clemens:
## Current Behavior

Running react CT with webpack generated 3 different webpack errors with buildable libraries
1. `ERROR in Conflict: Multiple assets emit different content to the same filename index.html`
1. `Error: Can't resolve 'core-js/modules/es.object.assign.js'`
1. `Error: ENOENT: no such file or directory, scandir` via the license plugin

Most of these seem to only happen on windows, though sporadically happen on macOS at time.

## Expected Behavior

These errors do not happen when running CT.

### DIFF
--- a/packages/angular/plugins/component-testing.ts
+++ b/packages/angular/plugins/component-testing.ts
@@ -101,7 +101,7 @@ ${e.stack ? e.stack : e}`
   );
 
   return {
-    ...nxBaseCypressPreset(pathToConfig),
+    ...nxBaseCypressPreset(pathToConfig, { testingType: 'component' }),
     // NOTE: cannot use a glob pattern since it will break cypress generated tsconfig.
     specPattern: ['src/**/*.cy.ts', 'src/**/*.cy.js'],
     // cypress defaults to a relative path from the workspaceRoot instead of projectRoot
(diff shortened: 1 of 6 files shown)


jaysoo (approved):

packages/cypress/src/executors/cypress/cypress.impl.ts
```diff
@@ -67,7 +67,10 @@ export default async function cypressExecutor(
   options = normalizeOptions(options, context);
   // this is used by cypress component testing presets to build the executor contexts with the correct configuration options.
   process.env.NX_CYPRESS_TARGET_CONFIGURATION = context.configurationName;
-  let success;
+  // used to set babel settings for react CT.
+  process.env.NX_CYPRESS_COMPONENT_TEST =
```
FrozenPandaz, line 71: Can we try to make this a part of the react `nxComponentTestingPreset`?

FrozenPandaz (changes requested):

Coly010 (approved):

packages/cypress/src/executors/cypress/cypress.impl.ts
barbados-clemens: this does work! 🎉

ndcunningham (approved):

FrozenPandaz (approved):

PR #17604, opened 2023-06-14
fix(core): do not build the project graph in nx report if there's no native support

Cammisuli:
## Current Behavior
If we're on a system that does not have native support, nx report will still try to build the project graph.

Because some dependencies for the project graph uses native code, nx report will output incorrect messages

## Expected Behavior
Nx report does not try to use the project graph when generating the report when there's no native support

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/installation/troubleshoot-installation.md
+++ b/docs/shared/installation/troubleshoot-installation.md
@@ -10,9 +10,10 @@ If you see a message saying that your platform is not supported (or that Nx cann
 are a few reasons why this could potentially happen:
 
 1. Running your install command with `--no-optional` (or the relative flag in yarn, pnpm, etc)
-2. The package-lock.json file was not correctly updated by npm, and missed optional dependencies used by Nx.
+1. Running your install with `--dev` for pnpm.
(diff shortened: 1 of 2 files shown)


docs/shared/installation/troubleshoot-installation.md
```diff
@@ -10,9 +10,10 @@ If you see a message saying that your platform is not supported (or that Nx cann
 are a few reasons why this could potentially happen:
 
 1. Running your install command with `--no-optional` (or the relative flag in yarn, pnpm, etc)
-2. The package-lock.json file was not correctly updated by npm, and missed optional dependencies used by Nx.
+1. Running your install with `--dev` for pnpm.
+1. The package-lock.json file was not correctly updated by npm, and missed optional dependencies used by Nx.
    You can read more about this [issue on the npm repository.](https://github.com/npm/cli/issues/4828)
-3. [Your platform is not supported](#supported-native-module-platforms)
+1. [Your platform is not supported](#supported-native-module-platforms)
```
FrozenPandaz, line 16: These all changed to `1. ` They should be ordered right?

FrozenPandaz (approved):

PR #17615, opened 2023-06-15
feat(node): remove node:node executor

mandarini:
Removing `@nx/node:node` executor, since it's not needed, as [it calls the `@nx/js:node` executor](https://github.com/nrwl/nx/blob/master/packages/node/src/executors/node/node.impl.ts). Also add a migration to move over existing configurations.


jaysoo (approved):
LGTM!

meeroslav (approved):
LGTM

isaacplmann (approved):

PR #17621, opened 2023-06-15
chore(core): update to lerna 7.0.2

fahslaj:
## Current Behavior

Nx packages are published with Lerna 6.6.2.

## Expected Behavior

Nx packages are published with Lerna 7.0.2, and Lerna detects packages in the src tree instead of just the build directory.


packages-legacy/angular/package.json
```diff
@@ -40,5 +40,12 @@
   },
   "nx-migrations": {
     "migrations": "@nx/angular/migrations.json"
+  },
+  "lerna": {
+    "command": {
+      "publish": {
+        "directory": "{workspaceRoot}/build/packages/angular-legacy"
```
fahslaj, line 47: This custom directory config is required for all the legacy packages because {projectRoot} doesn't correlate to its build directory output.

PR #17622, opened 2023-06-15
fix(js): node executor should always log error

### LINKED ISSUE #17083: js: silently fails to run a file
### Current Behavior

I have this:
```
  "targets": {
    "build": {
      "executor": "@nx/js:tsc",
      "outputs": ["{options.outputPath}"],
      "options": {
        "outputPath": "dist/packages/analyze",
        "main": "packages/analyze/src/index.ts",
        "tsConfig": "packages/analyze/tsconfig.lib.json",
        "assets": []
      }
    },
    "run": {
      "executor": "@nx/js:node",
      "options": {
        "buildTarget": "analyze:build",
        "watch": false
      }
    },
```

When I run `nx run analyze:run` - it says ">  NX   Successfully ran target run for project analyze" however the actual file is not being run. I can confirm this by throwing in the file, logging to a file, etc.

I believe it's because in https://github.com/nrwl/nx/pull/16924 in `node.impl.ts` there's no handling of non-zero exit codes in `task.childProcess.once('exit', (code) => {` handler.

After som debugging, I found out that it is trying to run `main.js` file - the default one.

Previously before https://github.com/nrwl/nx/pull/16924 the file to run was determined by `NX_FILE_TO_RUN: event.outfile`.

Now I have to add `"outputFileName": "src/index.js",` to my `build` target options which isn't very nice.

### Expected Behavior

1. When running js file fails - it shows the error
2. I don't have to specify `outputFileName` in `build` target in order to run it (also `main.js` doesn't seem like a good default because even if I had `main.ts` - I'd still have to add `src/` prefix)

### GitHub Repo

_No response_

### Steps to Reproduce

See above

### Nx Report

```shell
Node   : 19.8.1
   OS     : win32 x64
   npm    : 9.5.1
   Hasher : Native

   nx                 : 16.2.1
   @nx/js             : 16.2.1
   @nx/jest           : 16.2.1
   @nx/linter         : 16.2.1
   @nx/workspace      : 16.2.1
   @nx/devkit         : 16.2.1
   @nx/eslint-plugin  : 16.2.1
   @nx/plugin         : 16.2.1
   @nrwl/tao          : 16.2.1
   @nx/vite           : 16.2.1
   nx-cloud           : 16.0.5
   typescript         : 5.0.4
```

### Failure Logs

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

### LINKED ISSUE #17595: Multiple issues running `@nx/js:node` executor (entry file must be named main.ts)
### Current Behavior

On this repository where we're using Nx 14.7.11
https://github.com/DevCycleHQ/js-sdks/tree/8ae38756b1494bf007147480c45693140a141cd6

We have a project named `shared-bucketing-test-data` located in `lib/shared/bucketing-test-data` which is essentially a node script that generates some test data.
It runs using the command `nx build:json shared-bucketing-test-data` and everything works on this version of Nx. The "build" target is configured to use `@nx/js:tsc` and the `build:json` target runs the built output using the `@nx/node:node` executor.

In this commit, I've updated to Nx 16.3.2
https://github.com/DevCycleHQ/js-sdks/tree/e29999ddeddf940dde7ee76feb6097d72ee35822

Now the same `build:json` target no longer executes the Node script defined by this project. The command executes "successfully" with no errors, but also does not actually run the script.

To try and fix it,  I generated a new `@nx/node:application` to see what the latest setup should be for a Node application. The generator now uses `esbuild` and switched from `@nx/node:node` to `@nx/js:node`, so I copied that configuration to my existing project. 

That did not fix the problem, it exhibited the same behaviour of "running successfully" without actually running.

Then I renamed the `index.ts` file to `main.ts` and now the script executes successfully. If I name it anything else, it doesn't work.

This file rename fixed the build for the `esbuild` setup, but applying the same change to my original `tsc` setup did not fix the problem.

It also seems like any errors thrown by the script are being swallowed, so even when the script does run I'll see logs up until an error is thrown, and then nothing that reflects the error (just a clean exist with a "success" from nx)

The final version which fixes the configuration issues is here
https://github.com/DevCycleHQ/js-sdks/tree/889001dc8a435202e5487bca0bc1956660b77463

To summarize, I had to switch from using `@nx/js:tsc` to `@nx/esbuild:esbuild` to build. I also had to make my main file named `main.ts` and not `index.ts` or anything else besides `main.ts`. Finally, the script was also throwing an unrelated error which was never being logged (and still isn't).

### Expected Behavior

- my existing setup should keep working (as far as I can tell) in newer Nx version. I could not get the setup to work at all while using the `@nx/web:tsc` executor to build my files
- the `@nx/web:node` executor should not require the main file to be named `main.ts` (as seems to be the case)
- errors should be logged correctly when thrown by the script being executed

### GitHub Repo

https://github.com/DevCycleHQ/js-sdks/tree/e29999ddeddf940dde7ee76feb6097d72ee35822

### Steps to Reproduce

1. Clone sample repository, run yarn
2. `yarn nx build:json shared-bucketing-test-data --skip-nx-cache`
3. Command completes successfully with no errors, but script does not run (no logs)

### Nx Report

```shell
Node   : 18.16.0
   OS     : darwin arm64
   yarn   : 3.2.0
   Hasher : Native

   nx                 : 16.3.2
   lerna              : 5.6.2
   @nx/js             : 16.3.2
   @nx/jest           : 16.3.2
   @nx/linter         : 16.3.2
   @nx/workspace      : 16.3.2
   @nx/cypress        : 16.3.2
   @nx/detox          : 16.3.2
   @nx/devkit         : 16.3.2
   @nx/esbuild        : 16.3.2
   @nx/eslint-plugin  : 16.3.2
   @nx/next           : 16.3.2
   @nx/node           : 16.3.2
   @nx/react          : 16.3.2
   @nx/react-native   : 16.3.2
   @nx/rollup         : 16.3.2
   @nrwl/tao          : 16.3.2
   @nx/web            : 16.3.2
   @nx/webpack        : 16.3.2
   typescript         : 5.0.4
   Local workspace plugins:
   	 @devcycle/bucketing-assembly-script
   	 @devcycle/openfeature-nodejs-provider
   	 @devcycle/bucketing-test-data
   	 @devcycle/bucketing
   	 @devcycle/devcycle-react-native-sdk
   	 @devcycle/types
   	 @devcycle/nodejs-server-sdk
   	 @devcycle/devcycle-react-sdk
   	 @devcycle/devcycle-js-sdk
```
```


### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

### LINKED ISSUE #17627: js: doesn't show any errors from stderr
### Current Behavior

I run nodejs scripts using `@nx/js:node`
When there's a TS compilation or runtime error - I don't see it, tasks exits without any error output.
This makes debugging a nightmare.

### Expected Behavior

I can see stderr when using @nx/js:node

### GitHub Repo

_No response_

### Steps to Reproduce

1. Add `console.log("test".replaceAll(/a/, "b"));` to your script
2. Run it using `@nx/js:node`

### Nx Report

```shell
Node   : 20.2.0
   OS     : win32 x64
   npm    : 9.6.6
   Hasher : Native

   nx                 : 16.3.2
   @nx/js             : 16.3.2
   @nx/jest           : 16.3.2
   @nx/linter         : 16.3.2
   @nx/workspace      : 16.3.2
   @nx/devkit         : 16.3.2
   @nx/eslint-plugin  : 16.3.2
   @nx/plugin         : 16.3.2
   @nrwl/tao          : 16.3.2
   @nx/vite           : 16.3.2
   nx-cloud           : 16.0.5
   typescript         : 5.1.3
```

### Failure Logs

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

This seems related to https://github.com/nrwl/nx/issues/17083

Temporary fix is to add `console.error(data.toString());` into `node_modules\@nx\js\src\executors\node\node.impl.js`:
```js
// Run the program
task.promise = new Promise((resolve, reject) => {
    var _a;
    task.childProcess = (0, child_process_1.fork)((0, devkit_1.joinPathFragments)(__dirname, 'node-with-require-overrides'), (_a = options.runtimeArgs) !== null && _a !== void 0 ? _a : [], {
        execArgv: getExecArgv(options),
        stdio: [0, 1, 'pipe', 'ipc'],
        env: Object.assign(Object.assign({}, process.env), { NX_FILE_TO_RUN: fileToRun, NX_MAPPINGS: JSON.stringify(mappings) }),
    });
    task.childProcess.stderr.on('data', (data) => {
        console.error(data.toString());
        // Don't log out error if task is killed and new one has started.
        // This could happen if a new build is triggered while new process is starting, since the operation is not atomic.
        if (options.watch && !task.killed) {
            devkit_1.logger.error(data.toString());
        }
    });
    task.childProcess.once('exit', (code) => {
        if (options.watch && !task.killed) {
            devkit_1.logger.info(`NX Process exited with code ${code}, waiting for changes to restart...`);
        }
        if (!options.watch)
            done();
        resolve();
    });
    next({ success: true });
});
```

mandarini:
## Current Behavior

When using the `nx/js:node` executor, error is swallowed unless you're in `watch` mode.

## Expected Behavior

Error should always be logged.

Also, we should infer the `outputFileName` from the `main` entrypoint as set in build options.

## Related Issue(s)

Fixes #17595 #17627 #17083

### DIFF
--- a/e2e/utils/command-utils.ts
+++ b/e2e/utils/command-utils.ts
@@ -23,6 +23,7 @@ export interface RunCmdOpts {
   cwd?: string;
   silent?: boolean;
   verbose?: boolean;
+  redirectStderr?: boolean;
 }
 
 /**
@@ -325,26 +326,25 @@ export function runCLI(
     silenceError: false,
     env: undefined,
     verbose: undefined,
+    redirectStderr: undefined,
   }
 ): string {
   try {
     const pm = getPackageManagerCommand();
-    const logs = execSync(
-      `${pm.runNxSilent} ${command} ${
-        opts.verbose ?? isVerboseE2ERun() ? ' --verbose' : ''
-      }`,
-      {
-        cwd: opts.cwd || tmpProjPath(),
-        env: {
-          CI: 'true',
-          ...getStrippedEnvironmentVariables(),
-          ...opts.env,
-        },
-        encoding: 'utf-8',
-        stdio: 'pipe',
-        maxBuffer: 50 * 1024 * 1024,
-      }
-    );
+    const commandToRun = `${pm.runNxSilent} ${command} ${
+      opts.verbose ?? isVerboseE2ERun() ? ' --verbose' : ''
+    }${opts.redirectStderr ? ' 2>&1' : ''}`;
+    const logs = execSync(commandToRun, {
+      cwd: opts.cwd || tmpProjPath(),
+      env: {
+        CI: 'true',
+        ...getStrippedEnvironmentVariables(),
+        ...opts.env,
+      },
+      encoding: 'utf-8',
+      stdio: 'pipe',
+      maxBuffer: 50 * 1024 * 1024,
+    });
 
     if (opts.verbose ?? isVerboseE2ERun()) {
       output.log({
--- a/packages/js/src/executors/node/node.impl.ts
+++ b/packages/js/src/executors/node/node.impl.ts
@@ -10,11 +10,13 @@ import {
 import { daemonClient } from 'nx/src/daemon/client/client';
 import { randomUUID } from 'crypto';
 import { join } from 'path';
+import * as path from 'path';
 
 import { InspectType, NodeExecutorOptions } from './schema';
 import { createAsyncIterable } from '@nx/devkit/src/utils/async-iterable';
 import { calculateProjectDependencies } from '../../utils/buildable-libs-utils';
 import { killTree } from './lib/kill-tree';
+import { fileExists } from 'nx/src/utils/fileutils';
 
 interface ActiveTask {
   id: string;
@@ -71,12 +73,11 @@ export async function* nodeExecutor(
 
   // Re-map buildable workspace projects to their output directory.
   const mappings = calculateResolveMappings(context, options);
-  const fileToRun = join(
-    context.root,
-    buildOptions.outputPath,
-    buildOptions.outputFileName ?? 'main.js'
-  );
 
+  const outputFileName =
+    buildOptions.outputFileName ?? `${path.parse(buildOptions.main).name}.js`;
+
+  const fileToRun = join(context.root, buildOptions.outputPath, outputFileName);
   const tasks: ActiveTask[] = [];
   let currentTask: ActiveTask = null;
 
@@ -159,7 +160,7 @@ export async function* nodeExecutor(
                   stdio: [0, 1, 'pipe', 'ipc'],
                   env: {
                     ...process.env,
-                    NX_FILE_TO_RUN: fileToRun,
+                    NX_FILE_TO_RUN: fileToRunCorrectPath(fileToRun),
                     NX_MAPPINGS: JSON.stringify(mappings),
                   },
                 }
@@ -168,7 +169,8 @@ export async function* nodeExecutor(
               task.childProcess.stderr.on('data', (data) => {
                 // Don't log out error if task is killed and new one has started.
                 // This could happen if a new build is triggered while new process is starting, since the operation is not atomic.
-                if (options.watch && !task.killed) {
+                // Log the error in normal mode
+                if (!options.watch || !task.killed) {
                   logger.error(data.toString());
                 }
               });
@@ -302,6 +304,25 @@ function runWaitUntilTargets(
       });
     })
   );
+}
+
+function fileToRunCorrectPath(fileToRun: string): string {
+  if (!fileExists(fileToRun)) {
+    const cjsFile = fileToRun.replace(/\.js$/, '.cjs');
+    if (fileExists(cjsFile)) {
+      fileToRun = cjsFile;
+    } else {
+      const mjsFile = fileToRun.replace(/\.js$/, '.mjs');
+      if (fileExists(mjsFile)) {
+        fileToRun = mjsFile;
+      } else {
+        throw new Error(
+          `Could not find ${fileToRun}. Make sure your build succeeded.`
+        );
+      }
+    }
+  }
+  return fileToRun;
 }
 
 export default nodeExecutor;
--- a/e2e/js/src/js-node.test.ts
+++ b/e2e/js/src/js-node.test.ts
@@ -1,0 +1,50 @@
+import {
+  cleanupProject,
+  newProject,
+  runCLI,
+  uniq,
+  updateFile,
+  updateProjectConfig,
+} from '@nx/e2e/utils';
+
+describe('js:node error handling', () => {
+  let scope: string;
+
+  beforeEach(() => {
+    scope = newProject();
+  });
+
+  afterEach(() => cleanupProject());
+
+  it('should log out the error', () => {
+    const esbuildLib = uniq('esbuildlib');
+
+    runCLI(
+      `generate @nx/js:lib ${esbuildLib} --bundler=esbuild --no-interactive`
+    );
+
+    updateFile(`libs/${esbuildLib}/src/index.ts`, () => {
+      return `
+        console.log('Hello from my library!');
+        throw new Error('This is an error');
+        `;
+    });
+
+    updateProjectConfig(esbuildLib, (config) => {
+      config.targets['run-node'] = {
+        executor: '@nx/js:node',
+        options: {
+          buildTarget: `${esbuildLib}:build`,
+          watch: false,
+        },
+      };
+      return config;
+    });
+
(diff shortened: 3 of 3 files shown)


packages/js/src/executors/node/node.impl.ts
```diff
@@ -71,11 +74,14 @@ export async function* nodeExecutor(
 
   // Re-map buildable workspace projects to their output directory.
   const mappings = calculateResolveMappings(context, options);
-  const fileToRun = join(
-    context.root,
-    buildOptions.outputPath,
-    buildOptions.outputFileName ?? 'main.js'
-  );
+
+  let outputFileName = buildOptions.outputFileName;
```
ndcunningham, line 78: I guess you could simplify this by 
```
const outputFileName = buildOptions.outputFileName || getOutputFileName(buildOptions.main);

```

packages/js/src/executors/node/node.impl.ts
```diff
@@ -71,11 +74,14 @@ export async function* nodeExecutor(
...
+
+  if (!outputFileName) {
+    outputFileName = getOutputFileName(buildOptions.main);
+  }
+
+  let fileToRun = join(context.root, buildOptions.outputPath, outputFileName);
```
ndcunningham, line 84: This can be `const` instead of `let` right?

packages/js/src/executors/node/node.impl.ts
```diff
@@ -71,11 +72,13 @@ export async function* nodeExecutor(
...
-    context.root,
-    buildOptions.outputPath,
-    buildOptions.outputFileName ?? 'main.js'
-  );
+
+  const outputFileName =
+    buildOptions.outputFileName ?? `${path.parse(buildOptions.main).name}.js`;
+
+  // since we cannot get the normalized build options here,
+  // should we test fileExists for all possible extensions? eg. '.cjs' | '.mjs' | '.js' ?
```
mandarini, line 80: wdyt @jaysoo ?

e2e/utils/command-utils.ts
```diff
@@ -332,7 +332,7 @@ export function runCLI(
     const logs = execSync(
       `${pm.runNxSilent} ${command} ${
         opts.verbose ?? isVerboseE2ERun() ? ' --verbose' : ''
-      }`,
+      } 2>&1`,
```
meeroslav, line 335: Nice catch

meeroslav: Could we have this as an additional flag to the call? (can be done in a follow up PR)

meeroslav (approved):
LGTM

e2e/utils/command-utils.ts
mandarini: Don't we generally want to keep all logs, though? I mean, when would we not want to show the console.error?

meeroslav: Usually console.error is accompanied by an actual error so this is a special case.

jaysoo: It's probably fine, since additional logs (errors) can just be ignored in the returned output. If it becomes and issue for some reason later we can add the option.

jaysoo (approved):

e2e/utils/command-utils.ts
mandarini: Some e2e tests where failing, so I added the option! :)

e2e/utils/command-utils.ts
```diff
@@ -325,14 +326,15 @@ export function runCLI(
     silenceError: false,
     env: undefined,
     verbose: undefined,
+    redirectStderr: false,
```
mandarini, line 329: @meeroslav I added this option

packages/js/src/executors/node/node.impl.ts
```diff
@@ -304,4 +306,23 @@ function runWaitUntilTargets(
   );
 }
 
+function fileToRunCorrectPath(fileToRun: string): string {
```
mandarini, line 309: @jaysoo I added this extra function here to make sure we get the correct path to the file. We cannot get the extension directly from the build options, because there's a number of things we do to get the extension for esbuild and rollup, so it's just easier to check the three possible extensions.

jaysoo (approved):

ndcunningham (approved):

PR #17629, opened 2023-06-16
fix(bundling): add skipTypeField back for rollup executor

evanjmg:
This was previously removed and needs to be readded back since nx should always support esm modules with cjs dependencies. Forcing modern libraries to be not in esm format isn't good practice.
Previously removed here:
https://github.com/nrwl/nx/commit/575c6a152f4d167343bc30e83730cea9ff311c69

## Current Behavior
No skipTypeField and breaking change was implemented - should be added back

## Expected Behavior
Skip type field allows esm modules with cjs dependencies

## Related Issue(s)
https://github.com/nrwl/nx/issues/17595

I just did this quickly so maintainers please edit the schema and regenerate the docs if you agree with me on this one

### DIFF
--- a/docs/generated/packages/rollup/executors/rollup.json
+++ b/docs/generated/packages/rollup/executors/rollup.json
@@ -169,6 +169,11 @@
         "type": "boolean",
         "description": "Whether to skip TypeScript type checking.",
         "default": false
+      },
+      "skipTypeField": {
+        "type": "boolean",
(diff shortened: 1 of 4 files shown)


jaysoo (approved):

jaysoo:
Thanks for the PR, we'll add it back. This is fine as an escape hatch for users that run into issues.

Issue #17425, opened 2023-06-05
"enforce-module-boundaries" fails only on CI


wojtekolek:
### Current Behavior

For some reason, the linter rule: `enforce-module-boundaries` fails on `Gitlab CI`, which caused tons of unexpected errors. Locally everything works as expected. Everything is configured as should be following the docs (tags etc.).

### Expected Behavior

The `enforce-module-boundaries` check will pass on the CI.

### Steps to Reproduce

Hard to provide some valid steps to reproduce.

### Nx Report

```shell
Node   : 18.16.0
   OS     : darwin arm64
   yarn   : 1.22.19
   Hasher : Native

   nx                 : 16.3.2
   @nx/js             : 16.3.2
   @nx/jest           : 16.3.2
   @nx/linter         : 16.3.2
   @nx/workspace      : 16.3.2
   @nx/cypress        : 16.3.2
   @nx/devkit         : 16.3.2
   @nx/eslint-plugin  : 16.3.2
   @nx/react          : 16.3.2
   @nx/storybook      : 16.3.2
   @nrwl/tao          : 16.3.2
   @nx/web            : 16.3.2
   @nx/webpack        : 16.3.2
   nx-cloud           : 16.0.5
   typescript         : 5.0.4
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

The issue occurs when we decided to update nx from 15.7.* to 16.3.2.
Any idea what may be causing this sort of unexpected thing?

miluoshi:
The same happens for us after updating from 15.4.2 to 16.2.2. 
It fails with the following error only in CI (Github actions cloud runner Ubuntu 22.0.4): `Imports of lazy-loaded libraries are forbidden`

#### Nx Report

```
Node   : 19.9.0
   OS     : darwin arm64
   yarn   : 1.22.19
   Hasher : Native
   
   nx                 : 16.2.2
   @nx/js             : 16.2.2
   @nx/jest           : 16.2.2
   @nx/linter         : 16.2.2
   @nx/workspace      : 16.2.2
   @nx/angular        : 16.2.2
   @nx/cypress        : 16.2.2
   @nx/devkit         : 16.2.2
   @nx/eslint-plugin  : 16.2.2
   @nrwl/nx-plugin    : 16.0.0-beta.1
   @nx/react          : 16.2.2
   @nx/rollup         : 16.2.2
   @nx/storybook      : 16.2.2
[...]

PR #17635, opened 2023-06-16
fix(linter): improve error message for forbidden imports of lazy-loaded libs

miluoshi:
…orts of lazy-loaded libs

* Specify that *static* imports are forbidden
* Include target library name and source file path containing a dynamic import that triggered this rule

## Related Issue(s)

Related to #17425

### DIFF
--- a/packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
+++ b/packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
@@ -8,6 +8,7 @@ import {
 import { isRelativePath } from 'nx/src/utils/fileutils';
 import {
   checkCircularPath,
+  findFileByDependency,
   findFilesInCircularPath,
 } from '../utils/graph-utils';
 import {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
```diff
@@ -125,7 +126,7 @@ export default createESLintRule<Options, MessageIds>({
       noImportsOfE2e: 'Imports of e2e projects are forbidden',
       noImportOfNonBuildableLibraries:
         'Buildable libraries cannot import or export from non-buildable libraries',
-      noImportsOfLazyLoadedLibraries: `Imports of lazy-loaded libraries are forbidden`,
+      noImportsOfLazyLoadedLibraries: `Static imports of lazy-loaded libraries are forbidden.\n\nLibrary "{{targetProjectName}}" is lazy-loaded in "{{file}}"`,
```
meeroslav, line 129: The library could be lazy loaded in several files. This should be a list like we have for circular dependencies.

packages/eslint-plugin/src/utils/graph-utils.ts
```diff
@@ -158,3 +162,17 @@ export function findFilesInCircularPath(
...
+export function findFileByDependency(
+  projectFileMap: ProjectFileMap,
+  sourceProjectName: string,
+  targetProjectName: string
+): FileData {
+  const isDynamicTargetDep = (dep: string | [string, string]) =>
+    fileDataDepType(dep) === DependencyType.dynamic &&
+    fileDataDepTarget(dep) === targetProjectName;
+
+  return projectFileMap[sourceProjectName].find((file) =>
```
meeroslav, line 175: Use `filter` instead of `find` to get all the files.

meeroslav (changes requested):

meeroslav:
Thank you @miluoshi,

I left one small improvement point.

miluoshi:
Thanks for the suggestion. Updated.

meeroslav:
Thank you, I will add one unit test to test this and then we can merge it.

meeroslav (approved):

PR #17637, opened 2023-06-16
feat(core): Allow specifying daemon socket dir

### LINKED ISSUE #17770: Allow specifying daemon socket directory

## Description
Allow overriding the placement of the nx daemon socket.

## Motivation
We use NX to manage our microservice-based web application. Each of our microservices has its own Dockerfile so it can specify its dependencies independent of the other services. For local development, we use docker compose to bring up each service in its own container. This ensures that each dev machine needs only docker to be able to serve our app locally. The added benefit is that we exercise something closer to our production environment during development. 

As our app has grown, we noticed the local development experience degrade. I posited that having N file watchers (one in each container) was the reason for the slowdown. I created a shared container, ran the nx daemon manually inside, and served all microservices from this single container. This resolved the performance issue. The downside is that this shared container is very local specific and needs to contain all dependencies for all our microservices.

Ideally, I'd like to "serve" the nx daemon in its own container and share the daemon socket to the other containers using a shared docker volume. This would give me the separation I'm looking for with the performance benefits I'm getting with the workaround. However, to be able to do this, the nx socket needs to be placed in a known directory.

## Suggested Implementation
Use an environment variable to optionally control the placement of the nx socket.

## Alternate Implementations
Could use configuration (nx.json) or similar.

aovens-quantifi:
Setting NX_DAEMON_SOCKET_DIR will place the daemon socket in the specified directory

## Current Behavior
The NX daemon socket is placed in a random temp folder, uncontrollable by the user

## Expected Behavior
Should allow controlling the placement of the socket

## Related Issue(s)
Fixes #17770

### DIFF
--- a/docs/shared/daemon.md
+++ b/docs/shared/daemon.md
@@ -36,3 +36,8 @@ When using Nx in a CI environment, the Nx Daemon is disabled by default. Whether
 ## Logs
 
 To see information about the running Nx Daemon (such as its background process ID and log output file), run `nx daemon`. Once you have the path to that log file, you could either open it in your IDE or stream updates in a separate terminal window by running `tail -f {REPLACE_WITH_LOG_PATH}`, for example.
+
+## Customizing the socket location
+
+The Nx Daemon uses a unix socket to communicate between the daemon and the Nx processes. By default this socket gets placed in a temp directory. If you are using Nx in a docker-compose environment, however, you may want to run the daemon manually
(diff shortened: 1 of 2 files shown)


FrozenPandaz:
Thank you for taking your time to work on this!

What is the motivation for specifying the output directory? Please open an issue first and we can discuss before implementing it.

PR #17639, opened 2023-06-16
feat(nx-plugin): update e2e tests to use local-registry and be simpler

FrozenPandaz:
## Current Behavior

e2e tests use special utils from `@nx/plugin` which obfuscate what the tests do

## Expected Behavior

e2e tests Node APIs which allow users to easily determine what's going on

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/generators/setup-verdaccio/files/config.yml
+++ b/packages/js/src/generators/setup-verdaccio/files/config.yml
@@ -1,10 +1,6 @@
 # path to a directory with all packages
 storage: ../tmp/local-registry/storage
 
-auth:
-  htpasswd:
-    file: ./htpasswd
-
 # a list of other known repositories we can talk to
 uplinks:
   npmjs:
@@ -26,4 +22,7 @@ packages:
 logs:
   type: stdout
   format: pretty
-  level: http
+  level: warn
+
+publish:
+  allow_offline: true # set offline to true to allow publish offline
--- a/packages/plugin/src/generators/create-package/files/e2e/__cliName__.spec.ts__tmpl__
+++ b/packages/plugin/src/generators/create-package/files/e2e/__cliName__.spec.ts__tmpl__
@@ -1,31 +1,56 @@
-import {
-  checkFilesExist,
-  removeTmpProject,
-  tmpFolder,
-  uniq,
-} from '@nx/plugin/testing';
 import { execSync } from 'child_process';
-import { join } from 'path';
+import { join, dirname } from 'path';
+import { mkdirSync, rmSync } from 'fs';
 
-describe('<%= cliName %> e2e', () => {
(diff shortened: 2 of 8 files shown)


packages/js/src/generators/setup-verdaccio/files/config.yml
```diff
@@ -26,4 +22,7 @@ packages:
 logs:
   type: stdout
   format: pretty
-  level: http
+  level: warn
```
FrozenPandaz, line 25: @xiongemi I could not get this test to work unless I brought this down to `warn`... Do you have any idea why?

AgentEnder (approved):

packages/plugin/src/generators/create-package/files/e2e/__cliName__.spec.ts__tmpl__
```diff
@@ -1,31 +1,56 @@
...
+  mkdirSync(dirname(projectDirectory), {
+    recursive: true,
+  });
+
+  execSync(
+    `<%= packageManagerCommands.exec %> --yes <%= cliName %>@e2e ${projectName} ${args}`,
+    {
+      cwd: dirname(projectDirectory),
+      stdio: 'inherit',
+      env: process.env,
```
jaysoo, line 50: Note: The default is `process.env`, so it's not necessary unless we're adding more values to it.

packages/plugin/src/generators/e2e-project/e2e.ts
```diff
@@ -66,10 +66,21 @@ function validatePlugin(host: Tree, pluginName: string) {
 }
 
 function addFiles(host: Tree, options: NormalizedSchema) {
+  const projectConfiguration = readProjectConfiguration(
+    host,
+    options.pluginName
+  );
+  const { name: pluginPackageName } = readJson(
```
jaysoo, line 73: Do we care if `pluginPackageName` is undefined? It is required in `package.json` so we could it's always there, otherwise it's a user error. We can also validate it and error out if it's missing somehow.

jaysoo (approved):

jaysoo:
Left some comments, LGTM overall

packages/plugin/src/generators/e2e-project/e2e.ts
FrozenPandaz: I'll add this in a follow up.

packages/plugin/src/generators/create-package/files/e2e/__cliName__.spec.ts__tmpl__
FrozenPandaz: Let me try without it in another PR.

I know it shouldn't make any difference... but somehow it does sometimes?

PR #17670, opened 2023-06-19
fix(core): run-commands should not parse strings as numbers

### LINKED ISSUE #17575: Zero gets stripped from run-commands arguments
### Current Behavior

Currently this command implements an `echo {args.buildNumber}`
```
npx nx echo my-project --buildNumber="20230613.10"
```

The result of the echo

20230613.1

The 0 gets chopped off

### Expected Behavior

The result should be

20230613.10

### GitHub Repo

_No response_

### Steps to Reproduce

Currently this command implements an `echo {args.buildNumber}`
```
npx nx echo my-project --buildNumber="20230613.10"
```

The result of the echo

20230613.1

The 0 gets chopped off

### Nx Report

```shell
Node   : 18.16.0
   OS     : darwin x64
   npm    : 9.5.1
   Hasher : Native

   nx                 : 16.3.2
   @nx/js             : 16.3.2
   @nx/jest           : 16.3.2
   @nx/linter         : 16.3.2
   @nx/workspace      : 16.3.2
   @nx/cypress        : 16.3.2
   @nx/devkit         : 16.3.2
   @nrwl/devkit       : 15.9.4
   @nx/eslint-plugin  : 16.3.2
   @nx/next           : 16.3.2
   @nx/react          : 16.3.2
   @nx/rollup         : 16.3.2
   @nrwl/tao          : 16.3.2
   @nx/web            : 16.3.2
   @nx/webpack        : 16.3.2
   nx-cloud           : 16.0.5
   typescript         : 5.0.2
   Community plugins:
   @nx-tools/nx-container : 4.0.4
   Local workspace plugins:
   	 @myproject/components
     - @nrwl/devkit@15.9.4

   To fix this, run `nx migrate nx@16.3.2`
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Coly010:
## Current Behavior

When passing command line args to a target using nx:run-commands whose value resembles a number, it is being parsed as a number. 

This can cause erroneous behaviour when the intention of the value is to be a string.

The main potential of erroneous behaviour here comes from decimal numbers ending in 0, such as `1234.10`. When this is parsed as a number then converted back to a string for interpolation, this value becomes `1234.1`.

## Expected Behavior

Args that are passed to the nx:run-commands via command line should not be parsed as numbers.

We should leave support args that are passed as options in the project.json

## Related Issue(s)

Fixes #17575

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -312,7 +312,15 @@ function parseArgs(options: RunCommandsOptions) {
     const unknownOptionsTreatedAsArgs = Object.keys(options)
       .filter((p) => propKeys.indexOf(p) === -1)
       .reduce((m, c) => ((m[c] = options[c]), m), {});
-    return unknownOptionsTreatedAsArgs;
+
+    const unparsedCommandArgs = yargsParser(options.__unparsed__, {
+      configuration: {
+        'parse-numbers': false,
+        'parse-positional-numbers': false,
+        'dot-notation': false,
+      },
+    });
+    return { ...unknownOptionsTreatedAsArgs, ...unparsedCommandArgs };
   }
   return yargsParser(args.replace(/(^"|"$)/g, ''), {
     configuration: { 'camel-case-expansion': false },
--- a/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
@@ -54,6 +54,26 @@ describe('Run Commands', () => {
     expect(readFile(f)).toEqual('123');
   });
 
+  it.each([
+    [`--key=123`, `args.key`, `123`],
+    [`--key="123.10"`, `args.key`, `123.10`],
+    [`--nested.key="123.10"`, `args.nested.key`, `123.10`],
+  ])(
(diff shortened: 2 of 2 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -312,7 +312,15 @@ function parseArgs(options: RunCommandsOptions) {
     const unknownOptionsTreatedAsArgs = Object.keys(options)
       .filter((p) => propKeys.indexOf(p) === -1)
       .reduce((m, c) => ((m[c] = options[c]), m), {});
-    return unknownOptionsTreatedAsArgs;
+
+    const unparsedCommandArgs = yargsParser(options.__unparsed__, {
```
FrozenPandaz, line 316: Can we please add some unit tests here to make sure that things are working as expected?

FrozenPandaz (changes requested):

packages/nx/src/executors/run-commands/run-commands.impl.ts
Coly010: can do, a unit test actually failed after the change so there is at least one scenario covered

FrozenPandaz (approved):

PR #17684, opened 2023-06-20
docs(angular): simplify tailwind css recipes for angular projects

leosvelperez:
## Current Behavior

The Tailwind CSS usage in Angular projects recipes is pretty verbose and makes it difficult to find the commands to run.

## Expected Behavior

The Tailwind CSS usage in Angular projects recipes is simpler and the commands for the different scenarios can be found more easily.

## Related Issue(s)

Fixes #


isaacplmann (approved):
I think this works well. We could split the different examples into separate recipes if more are added or if they grow any more.

PR #17690, opened 2023-06-20
feat(nest): allow es2021 as target for libs

timonmasberg:
## Current Behavior
Currently, `ES2021` is not a valid target for NestJS libraries. [NestJS itself targets ES2021](https://github.com/nestjs/nest/blob/master/tsconfig.json), since Node v12 support has been dropped, on every package.

## Expected Behavior
`ES2021` should be a valid target for libraries in the generator schema.


gperdomor:
@timonmasberg I was working on the same without knowing about this MR... In my case I setup es2021 as default target in app and lib generators... Maybe can we close this MR in favor of #17726... WDYT?

gperdomor:
@timonmasberg I was working on the same without knowing about this MR... In my case I also setup es2021 as default target in generators... Maybe can we close this MR in favor of #17726

timonmasberg:
> @timonmasberg I was working on the same without knowing about this MR... In my case I also setup es2021 as default target in generators... Maybe can we close this MR in favor of #17726

sure, having the default on es2021 is probably smart, since it aligns with the NestJS schematics

PR #17693, opened 2023-06-20
fix(react): fix nx init with react yarn repo

xiongemi:
- remove yarn.lock before running nx init

## Related Issue(s)

Fixes #


jaysoo:
Why is this needed? What issue is it fixing?

PR #17708, opened 2023-06-21
fix(bundling): check for browserslist when setting terser ecma #17620

### LINKED ISSUE #17620: Terser plugin set to 2020, thus shortening using nullish coalescing operator despite browserslist
### Current Behavior

Including code that Terser can shorten using `??` syntax will be shortened thus due to the choice of `ecma` value `2020` in the TerserPlugin options in the `@nx/webpack` package. This is despite any `.browserslistrc` configuration we choose.

### Expected Behavior

I would expect to be able to build projects without the nullish coalescing operator in my build output. Preferably there would be some way to adjust the plugin options to the TerserPlugin. Or maybe a different/older value chosen based on Nrwl's chosen level of support. I see that you've officially stopped supporting IE11, so something like es2015 or es2016.

### GitHub Repo

https://github.com/kevinbeal/nx-examples

### Steps to Reproduce

1. Write some code or include a package that would be shortened using `??` by terser, like so:
```javascript
function isIndex(value, length) {
  const type = typeof value;
  const MAX_SAFE_INTEGER = 9007199254740991;
  const reIsUint = /^(?:0|[1-9]\d*)$/;
  length = length == null ? MAX_SAFE_INTEGER : length;
  return !!length && (type === "number" || type !== "symbol" && reIsUint.test(value)) && value > -1 && value % 1 === 0 && value < length;
}
```
2. Build with `optimization` flag turned on so that TerserPlugin is used.
3. Observe the build output contains nullish coalesing operator

### Nx Report

```shell
Node   : 16.5.0
   OS     : darwin-arm64
   yarn   : 1.22.19
   
   nx                 : 16.4.0-beta.3
   @nx/js             : 16.4.0-beta.3
   @nx/jest           : 16.4.0-beta.3
   @nx/linter         : 16.4.0-beta.3
   @nx/workspace      : 16.4.0-beta.3
   @nx/angular        : 16.4.0-beta.3
   @nx/cypress        : 16.4.0-beta.3
   @nx/devkit         : 16.4.0-beta.3
   @nx/eslint-plugin  : 16.4.0-beta.3
   @nx/react          : 16.4.0-beta.3
   @nrwl/tao          : 16.4.0-beta.3
   @nx/web            : 16.4.0-beta.3
   @nx/webpack        : 16.4.0-beta.3
   nx-cloud           : 16.0.5
   typescript         : 5.0.4
   Community plugins:
   @ngrx/component-store : 16.0.0
   @ngrx/effects         : 16.0.0
   @ngrx/entity          : 16.0.0
   @ngrx/router-store    : 16.0.0
   @ngrx/store           : 16.0.0
   @ngrx/store-devtools  : 16.0.0
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

A custom webpack config can be used to overwrite the `ecma` value, like so:
```javascript
const { composePlugins, withNx } = require('@nx/webpack');
const { withReact } = require('@nx/react');
module.exports = composePlugins(
  withNx(),
  withReact(),
  (config, { options, context }) => {
    config.optimization.minimizer[0].options.minimizer.options.ecma = "2015"; // <---here
    return config;
  }
);
```
Not that I would recommend this.

Coly010:
## Current Behavior

We do not check browserslist files in the `withNx` helper. 

This can lead to outputs that are not compatible with the targetted browsers

## Expected Behavior

We should let browserslist help define the ecma version for terser to use

## Related Issue(s)

Fixes #17620

## Notes
An improvement could be to find a method to determine the lowest ecma version based on all the browsers matching the browserslist queries

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -32,6 +32,7 @@
     "@babel/core": "^7.15.0",
     "autoprefixer": "^10.4.9",
     "babel-loader": "^9.1.2",
+    "browserslist": "^4.21.4",
     "chalk": "^4.1.0",
     "chokidar": "^3.5.1",
     "copy-webpack-plugin": "^10.2.4",
--- a/packages/webpack/src/utils/with-nx.ts
+++ b/packages/webpack/src/utils/with-nx.ts
@@ -1,12 +1,10 @@
 import * as path from 'path';
-import { Configuration, WebpackPluginInstance, ProgressPlugin } from 'webpack';
+import { join } from 'path';
+import { Configuration, ProgressPlugin, WebpackPluginInstance } from 'webpack';
 import { ExecutorContext } from 'nx/src/config/misc-interfaces';
 import { TsconfigPathsPlugin } from 'tsconfig-paths-webpack-plugin';
 import { readTsConfig } from '@nx/js';
 import { LicenseWebpackPlugin } from 'license-webpack-plugin';
-import TerserPlugin = require('terser-webpack-plugin');
-import nodeExternals = require('webpack-node-externals');
-import ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 
 import { NormalizedWebpackExecutorOptions } from '../executors/webpack/schema';
 import { StatsJsonPlugin } from '../plugins/stats-json-plugin';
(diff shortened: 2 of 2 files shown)


packages/webpack/package.json
```diff
@@ -70,6 +70,9 @@
     "@nx/devkit": "file:../devkit",
     "@nx/js": "file:../js"
   },
+  "peerDependencies": {
+    "browserslist": "^4.21.4"
```
jaysoo, line 74: I would move this to `dependencies` (we used to have it  but removed it when we removed support for legacy browsers), or just load it with a try-catch, so it's completely optional.

jaysoo (approved):

kevinbeal:
Thank you guys for doing this!

This may just be nitpicky stuff I'm bringing up, but the browser I discovered the issue on was a version of Safari, so while this is certainly an improvement, it wouldn't fix the issue for browserlists targeting older Safari but not IE11 (since the fix only checks for IE11 and otherwise picks es2020). Of course, I can just target IE11 and thus also cover old Safari, but I thought I would mention it anyway.

Coly010:
@kevinbeal  Do you know what version of Safari it is? I can add it to an array of browsers to change to ES5.

There's no good way to get the ecmascript version that works across all the browsers found with `browserlist` unfortunately.

Coly010:
> I noticed that too. I wonder why not. Seems like an obvious feature, but maybe that's just hindsight or my ignorance speaking 😅

@kevinbeal It's because some browsers introduced partial support for modern ES features and not total feature parity with each ES spec

PR #17714, opened 2023-06-21
fix(nx-dev): set conf logo to support both themes

bcabanes:
It fixes white logo on white background for nx cloud.


juristr (approved):
LGTM, thank you

PR #17717, opened 2023-06-21
fix(core): fix batch mode not streaming outputs

FrozenPandaz:
## Current Behavior

Batched Nx Runs do not stream output and look terrible.

## Expected Behavior

Batched Nx Runs have their outputs streamed.

## Related Issue(s)

Closes https://github.com/nrwl/nx/pull/17713


barbados-clemens (approved):
confirmed working with jest batch mode 👍

AgentEnder (approved):

PR #17725, opened 2023-06-22
fix(react): ssr should serve correctly and e2e should function

### LINKED ISSUE #17723: nx e2e broken on default run
### Current Behavior

Nx 16.3.2 is broken on create default react application. then run SSR `nx generate setup-ssr`

After serving application first error `node_modules/@nx/webpack/src/executors/ssr-dev-server/ssr-dev-server.impl.js` on line  31 `output.options.target` works after changing to `output.options?.target`

The default e2e application created does not work and running e2e affected runs forever.

Screenshot 2023-06-22 at 11 05 48

Running `npx nx run e2e:e2e --configuration=ci` however runs  

### Expected Behavior

e2e should run 

### GitHub Repo

_No response_

### Steps to Reproduce

1. npx create-nx-workspace@latest
2. Where would you like to create your workspace? · store
✔ Which stack do you want to use? · react
✔ What framework would you like to use? · none
✔ Standalone project or integrated monorepo? · standalone
✔ Which bundler would you like to use? · webpack
✔ Default stylesheet format             · css
✔ Enable distributed caching to make your CI faster · Yes
3. nx generate setup-ssr
4. run all available commands in visual studio Nx Console 

### Nx Report

```shell

   Node   : 18.14.2
   OS     : darwin x64
   npm    : 9.7.2
   Hasher : Native
   
   nx (global)        : 16.3.2
   nx                 : 16.3.2
   @nx/js             : 16.3.2
   @nx/jest           : 16.3.2
   @nx/linter         : 16.3.2
   @nx/workspace      : 16.3.2
   @nx/cypress        : 16.3.2
   @nx/devkit         : 16.3.2
   @nx/eslint-plugin  : 16.3.2
   @nx/react          : 16.3.2
   @nrwl/tao          : 16.3.2
   @nx/web            : 16.3.2
   @nx/webpack        : 16.3.2
   typescript         : 5.0.4
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Coly010:
## Current Behavior

E2E and Serve of React SSR apps not functioning correctly.

The ssr-dev-server is not receiving build options from the js/node executor to be able to determine when the server has been started successfully.
Therefore, it can never set itself to a ready state.

This prevents cypress from being able to run.

The webpack executor also looks for the entry point based on `outputFileName` or else it will default to `main.js`.
The node executor finds the entry point based on `main` in build options. 

Webpack executor is outputting `main.js` as the entrypoint, however, node executor is looking for `server.js` as the main option is `server.ts`.

The setup-ssr generator does not set an `outputFileName`. Make it set an `outputFileName` to ensure consistency and behaviour.
If user's want to change the name of the filename that is output, they can change this option and both executors will still function in unison.

## Expected Behavior

Serving SSR and e2es should work.
setup-ssr generator should set an `outputFileName`

## Related Issue(s)

Fixes #17723 #17706

### DIFF
--- a/packages/js/src/executors/node/node.impl.ts
+++ b/packages/js/src/executors/node/node.impl.ts
@@ -9,8 +9,8 @@ import {
 } from '@nx/devkit';
 import { daemonClient } from 'nx/src/daemon/client/client';
 import { randomUUID } from 'crypto';
-import { join } from 'path';
 import * as path from 'path';
+import { join } from 'path';
 
 import { InspectType, NodeExecutorOptions } from './schema';
 import { createAsyncIterable } from '@nx/devkit/src/utils/async-iterable';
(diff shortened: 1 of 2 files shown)


jaysoo (approved):

ojemuyiwa:
Hi @Coly010 Sadly this doesn't work out of the box, neither does setting `outputFileName:  "server.ts", "dist/parameter-manager/server/main.js"` etc. Used to work out of the box but doesn't work. changing e2e to serve-static works but this doesn’t. 😔
Thanks for looking into it though

Coly010:
There's more than just changing the output file name happening here. 

I have tested it out of the box and can confirm it works. What issue are you seeing?

ojemuyiwa:
> create workspace Standalone react app react framework none CSS Webpack No distributed cache generate SSR
> 
> when you run the auto generated e2e it doesn't work `webpack 5.87.0 compiled successfully (47eed2b4abafa82e) Type-checking in progress... No errors found. Debugger listening on ws://localhost:9229/bae8a966-8180-425f-9609-270786579072 Debugger listening on ws://localhost:9229/bae8a966-8180-425f-9609-270786579072 For help, see: https://nodejs.org/en/docs/inspector`

I.e. cypress does not launch

Coly010:
Yes, this will fix that. The fix hasn't been released yet. It will be released soon with 16.4 of Nx

Issue #17681, opened 2023-06-19
NestJS 10 support


gperdomor:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md))

## Description
Please add support for NestJS 10 
https://trilon.io/blog/nestjs-10-is-now-available

PR #17726, opened 2023-06-22
feat(nest): added es2021 support for nest generators

gperdomor:
## Current Behavior
Nest is using es2015 as ts target

## Expected Behavior
With migration to Nest 10, official templates now uses es2021

## Related Issue(s)
#17681 
#17690


barbados-clemens (dismissed):

packages/nest/src/generators/library/schema.json
```diff
@@ -94,7 +94,7 @@
     "target": {
       "description": "The ES target, Nest suggest using es6 or higher.",
```
ndcunningham, line 95: This should probably be updated as well

barbados-clemens, line 95:
```suggestion
      "description": "The ES target, Nest suggest using es2021 or higher.",
```

barbados-clemens (approved):
Just the one change for the description since we bumping the target.

PR #17761, opened 2023-06-23
docs(core): list deprecations in changelog

isaacplmann:
Lists deprecated features in the changelog page

https://nx-dev-git-fork-isaacplmann-docs-deprecated-timeline-nrwl.vercel.app/changelog

### DIFF
--- a/docs/changelog/16_0_0.md
+++ b/docs/changelog/16_0_0.md
@@ -22,7 +22,7 @@ Here are some of our feature highlights:
 {% cards cols="2" %}
 {% card title="workspace-lint" type="external" url="https://github.com/nrwl/nx/pull/16385" /%}
 {% card title="module federation utils from devkit public api" type="external" url="https://github.com/nrwl/nx/pull/16574" /%}
-{% card title="implicitDependencies" type="external" url="https://github.com/nrwl/nx/pull/16084" /%}
+{% card title="Global implicitDependencies" url="/deprecated/global-implicit-dependencies" /%}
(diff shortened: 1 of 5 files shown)


docs/changelog/16_0_0.md
```diff
@@ -17,3 +16,25 @@ Here are some of our feature highlights:
...
+
+{% cards cols="2" %}
+{% card title="workspace-lint" type="external" url="https://github.com/nrwl/nx/pull/16385" /%}
+{% card title="module federation utils from devkit public api" type="external" url="https://github.com/nrwl/nx/pull/16574" /%}
+{% card title="Global implicitDependencies" url="/deprecated/global-implicit-dependencies" /%}
+{% card title="stylus style option for react and next" type="external" url="https://github.com/nrwl/nx/pull/16135" /%}
+{% card title="@nrwl/tao package" type="external" url="https://github.com/nrwl/nx/pull/16550" /%}
+{% /cards %}
+
+## Breaking Changes
```
juristr, line 30: We could even think about having a reusable component for these that renders them always the same way, as we're going to have again and again. But we can do that at a later point too.

The only thing I'm thinking here is whether we should have here an intro sentence/note that mentions to use the `nx migrate` (with link to migrate docs) to automatically be upgraded avoiding these breaking changes.

juristr (approved):
👍 We should also submit a follow-up PR to add the changelog to the menu s.t. it is visible. 

**Something for another PR:**
Maybe the actual header bar there shouldn't be the current one, but rather pull in the docs one and add a section after the recipes that says "changelog"? 🤔 I think that was something I had discussed with @bcabanes. 
image

PR #17790, opened 2023-06-26
feat(testing): add --force-exit to jest executor

### LINKED ISSUE #17574: Jest - Add "forceExit" option
Currently I cannot find any way to get this param to work. I have migrated a ton of legacy projects to NX that have absolutely no need to clean up many resources when closed.... (docker container is destroyed and kubernetes starts a new one). I am currently having my CI run out of memory due to many different open handles that are opened during tests. I had absolutely no problem running all these tests with mocha using --exit however nx does not support mocha and I was using "babel mocha" but I am unable to get babel to register my library tsconfig paths. 

I was hoping to switch to jest to avoid having to run a babel compile step on my libraries, but not being able to use forceExit like a could with mocha is a deal breaker as there is way too much code and way too many open handles and absolutely no need to refactor the code to close them all as we have no need or repercussions from having them.

barbados-clemens:
## Current Behavior

--force-exit isn't an executor option for jest, but is an arg for the jest cli

## Expected Behavior

--force-exit is an option for jest executor

## Related Issue(s)

Fixes #17574

### DIFF
--- a/docs/generated/packages/jest/executors/jest.json
+++ b/docs/generated/packages/jest/executors/jest.json
@@ -81,6 +81,10 @@
         "description": "Find and run the tests that cover a comma separated list of source files that were passed in as arguments. (https://jestjs.io/docs/cli#--findrelatedtests-spaceseparatedlistofsourcefiles)",
         "type": "string"
       },
+      "forceExit": {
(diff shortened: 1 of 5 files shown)


barbados-clemens (commented):
branched off wrong base so added webpack changes. reverted those. Sorry for the extra review tags.

AgentEnder (approved):

PR #17798, opened 2023-06-26
fix(js): properly identify buildable project

### LINKED ISSUE #17764: js:tsc executor fails when run in a target not named "build"
### Current Behavior

If you try to build something using the `@nrwl/js:tsc` executor that depends on some other packages in the repository, it works as long as the target that is running it is named "build".

It seems that if you name it anything else, it breaks some internal assumptions the package is making and prevents it from rewriting tsconfig path aliases such that the build fails with a bunch of Typescript resolve errors from the dependent packages.

### Expected Behavior

The executor should work in any target regardless of what it's named, unless there's some piece of documentation I'm missing that explains why this is the case.

### GitHub Repo

https://github.com/DevCycleHQ/js-sdks

### Steps to Reproduce

1. clone repo
2. run `nx build js`
3. it builds successfully
4. change the "build" target in `sdk/js/project.json` to some other name
5. run `nx <new build name> js`
6. build now fails with a bunch of Typescript resolve errors

### Nx Report

```shell

   Node   : 18.16.0
   OS     : darwin arm64
   yarn   : 3.6.0
   Hasher : Native

   nx                 : 16.3.2
   lerna              : 5.6.2
   @nx/js             : 16.3.2
   @nx/jest           : 16.3.2
   @nx/linter         : 16.3.2
   @nx/workspace      : 16.3.2
   @nx/cypress        : 16.3.2
   @nx/detox          : 16.3.2
   @nx/devkit         : 16.3.2
   @nx/esbuild        : 16.3.2
   @nx/eslint-plugin  : 16.3.2
   @nx/next           : 16.3.2
   @nx/node           : 16.3.2
   @nx/react          : 16.3.2
   @nx/react-native   : 16.3.2
   @nx/rollup         : 16.3.2
   @nrwl/tao          : 16.3.2
   @nx/web            : 16.3.2
   @nx/webpack        : 16.3.2
   typescript         : 5.0.4
   Local workspace plugins:
   	 @devcycle/bucketing-assembly-script
   	 @devcycle/openfeature-nodejs-provider
   	 @devcycle/bucketing-test-data
   	 @devcycle/bucketing
   	 @devcycle/devcycle-react-native-sdk
   	 @devcycle/types
   	 @devcycle/nodejs-server-sdk
   	 @devcycle/devcycle-react-sdk
   	 @devcycle/devcycle-js-sdk
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

mandarini:
## Current Behavior

Right now, we identify if a project is buildable like this:
```
function isBuildable(target: string, node: ProjectGraphProjectNode): boolean {
  return (
    node.data.targets &&
    node.data.targets[target] &&
    node.data.targets[target].executor !== ''
  );
```

The `target` there is the name of the build target of a parent project. In that function, we try to understand if a project's dependency is buildable. This technique is flawed, since we cannot know for sure the name of the `build` target, and we cannot assume it's the same as the parent project's build target name.

This results in this:  If one project imports another project, both projects' build targets need to have the same name. If projects have different names for the build target, then you get an irrelevant error.

## Expected Behavior

`isBuildable` should not rely on a parent project's build target name, but rather look through the child project's executors, and find if it has an executor which is a build executor.

Retain existing behaviour to avoid breaking changes, and also allow it to work for build executors of other plugins (not `@nx/*`).

## Related Issue(s)

Fixes #17764

### DIFF
--- a/packages/js/src/utils/buildable-libs-utils.ts
+++ b/packages/js/src/utils/buildable-libs-utils.ts
@@ -12,17 +12,48 @@ import type * as ts from 'typescript';
 import { unlinkSync } from 'fs';
 import { output } from 'nx/src/utils/output';
 import { isNpmProject } from 'nx/src/project-graph/operators';
-import { ensureTypescript } from './typescript/ensure-typescript';
 import { readTsConfigPaths } from './typescript/ts-config';
 
-let tsModule: typeof import('typescript');
-
-function isBuildable(target: string, node: ProjectGraphProjectNode): boolean {
-  return (
-    node.data.targets &&
-    node.data.targets[target] &&
-    node.data.targets[target].executor !== ''
-  );
+function isBuildable(node: ProjectGraphProjectNode, target: string): string {
+  let buildTargetName: string | undefined = undefined;
+  // TODO(katerina): Remove @nrwl/* for Nx 17
+  const listOfBuildExecutors = [
+    '@nx/js:swc',
+    '@nx/js:tsc',
+    '@nx/rollup:rollup',
+    '@nx/rspack:rspack',
+    '@nx/vite:build',
+    '@nx/esbuild:esbuild',
+    '@nx/webpack:webpack',
+    '@nx/angular:ng-packagr-lite',
+    '@nx/angular:package',
+    '@nx/next:build',
+    '@nrwl/js:swc',
+    '@nrwl/js:tsc',
+    '@nrwl/rollup:rollup',
+    '@nrwl/rspack:rspack',
+    '@nrwl/vite:build',
+    '@nrwl/esbuild:esbuild',
+    '@nrwl/webpack:webpack',
+    '@nrwl/angular:ng-packagr-lite',
+    '@nrwl/angular:package',
+    '@nrwl/next:build',
+  ];
+  for (const targetName of Object.keys(node.data.targets ?? {})) {
+    const target = node.data.targets[targetName];
+    if (listOfBuildExecutors.includes(target.executor)) {
+      buildTargetName = targetName;
+      break;
+    }
+  }
+  if (!buildTargetName) {
+    return node.data.targets &&
+      node.data.targets[target] &&
+      node.data.targets[target].executor !== ''
+      ? target
+      : undefined;
+  }
+  return buildTargetName;
 }
 
 export type DependentBuildableProjectNode = {
@@ -73,7 +104,8 @@ export function calculateProjectDependencies(
       let project: DependentBuildableProjectNode = null;
       const depNode = projGraph.nodes[dep] || projGraph.externalNodes[dep];
       if (depNode.type === 'lib') {
-        if (isBuildable(targetName, depNode)) {
+        const dependencyTargetName = isBuildable(depNode, targetName);
+        if (dependencyTargetName) {
           const libPackageJsonPath = join(
             root,
             depNode.data.root,
@@ -89,7 +121,7 @@ export function calculateProjectDependencies(
                 overrides: {},
                 target: {
                   project: projectName,
-                  target: targetName,
+                  target: dependencyTargetName,
                   configuration: configurationName,
                 },
               },
--- a/packages/workspace/src/utilities/buildable-libs-utils.ts
+++ b/packages/workspace/src/utilities/buildable-libs-utils.ts
@@ -16,12 +16,46 @@ import { ensureTypescript } from './typescript';
 
 let tsModule: typeof import('typescript');
 
-function isBuildable(target: string, node: ProjectGraphProjectNode): boolean {
-  return (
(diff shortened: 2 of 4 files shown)


packages/workspace/src/utilities/buildable-libs-utils.ts
```diff
@@ -16,12 +16,38 @@ import { ensureTypescript } from './typescript';
...
 
-function isBuildable(target: string, node: ProjectGraphProjectNode): boolean {
-  return (
-    node.data.targets &&
-    node.data.targets[target] &&
-    node.data.targets[target].executor !== ''
-  );
+function isBuildable(node: ProjectGraphProjectNode): string {
+  let buildTargetName: string | undefined = undefined;
+  const listOfBuildExecutors = [
```
jaysoo, line 21: We should add a `TODO(v17)` so we don't forget to clean this up when we remove the `@nrwl/*` scope.

mandarini: Done!

leosvelperez (commented):
One thing to consider is that while this addresses the issue for projects using that list of executors, it leaves out projects that might be using executors from third-party plugins or even `nx:run-commands`.

This would be a breaking change for any workspace using such a setup. Maybe that was never intended to be supported and it was meant to be used only with executors from first-party plugins (I don't know), but the implementation allowed using others and folks might be using it.

If we don't want to add such a constraint and make a breaking change now, what about we add this new solution as a fallback to the previous one? The previous solution is quicker and would correctly catch the buildable dependency for some scenarios. If we don't find the target, then we look for known executors.

That would still address the issue while keeping the existing behavior. I would avoid changing the current behavior because we might change this later by using the `TaskGraph` instead of the `ProjectGraph`. The executor context now contains the task graph, so we can use it now. The project graph knows nothing about target dependencies, but the task graph does.

mandarini:
@leosvelperez sure, I'll keep the old way in there, too, as a fallback to the new way! Sounds like a good practice, indeed!

leosvelperez:
> Maybe this should somehow be in the docs, though, too? @juristr what do you think? For a later addition, of course, not as part of this PR. Like, if you have dependencies, keep the name of the build target consistent, to make sure third-party plugins work?

It's in some docs https://nx.dev/recipes/other/setup-incremental-builds-angular#build-target-name, but it should probably be generalized here https://nx.dev/more-concepts/incremental-builds. That said, the requirement might change soon, if we get to refactor this to use the task graph. When that happens, the target name will no longer be relevant and I think the only relevant thing will be the outputs which what is consumed with buildable libraries. So maybe we don't rush updating the docs because they might change soon.

packages/js/src/utils/buildable-libs-utils.ts
```diff
@@ -12,17 +12,48 @@ import type * as ts from 'typescript';
...
+      buildTargetName = targetName;
+      break;
+    }
+  }
+  if (!buildTargetName) {
+    return node.data.targets &&
+      node.data.targets[target] &&
+      node.data.targets[target].executor !== ''
+      ? target
+      : undefined;
```
leosvelperez, line 54: I would do this first. It's a faster lookup than looping through the targets and a lot of times, folks would have the same target name, so we could determine this pretty quickly and fall back to look if there are targets with known executors.

packages/workspace/src/utilities/buildable-libs-utils.ts
```diff
@@ -16,12 +16,46 @@ import { ensureTypescript } from './typescript';
...
+      buildTargetName = targetName;
+      break;
+    }
+  }
+  if (!buildTargetName) {
+    return node.data.targets &&
+      node.data.targets[target] &&
+      node.data.targets[target].executor !== ''
+      ? target
+      : undefined;
```
leosvelperez, line 56: Same as previous.

packages/js/src/utils/buildable-libs-utils.ts
mandarini: But can we assume that the other target has the same name, and if it does have the same name, it has the same purpose?

leosvelperez: The order in which it is asserted doesn't change that we still make that assumption, but that's the existing behavior. We wouldn't be breaking anything.

leosvelperez: Let's keep it the way you have it. You're right that by checking known executors first, we'd guarantee more correctness if there's a match and we fall back to the old checks for backward compat.

leosvelperez (approved):

AgentEnder (approved):

Issue #17270, opened 2023-05-28
use nextjs custom-server cannot find module .. main.js


evk8:
### Current Behavior

Cannot fine modul main.js not found when using nextjs custom-server.

How can I fix it?

### Expected Behavior

Cannot fine modul main.js not found when using nextjs custom-server.

How can I fix it?

### GitHub Repo

https://github.com/evk8/nx-16.2.2.git

### Steps to Reproduce

1. npx create-nx-workspace@latest --preset=next
2. nx g custom-server kis
3. nx serve kis 
----> failed task
<img width="989" alt="스크린샷 2023-05-28 오전 11 02 44" src="https://github.com/nrwl/nx/assets/90998467/97a1b7e0-f2a1-4229-a149-00bfdfb5369b">
4. fixed main.ts
<img width="929" alt="스크린샷 2023-05-28 오전 11 03 05" src="https://github.com/nrwl/nx/assets/90998467/9efe5c61-62b2-44f1-9c12-54c6b2d292b2">
----->
<img width="916" alt="스크린샷 2023-05-28 오전 11 03 27" src="https://github.com/nrwl/nx/assets/90998467/b36523bb-bb60-4ed5-9db9-7b6fec482924">
4. nx serve kis
<img width="1025" alt="스크린샷 2023-05-28 오전 11 04 57" src="https://github.com/nrwl/nx/assets/90998467/7266a112-db87-4d31-9a15-547bd30c4ee4">

### Nx Report

```shell
Node   : 16.16.0
   OS     : darwin arm64
   pnpm   : 7.18.0
   Hasher : Native
   
   nx                 : 16.2.2
   @nx/js             : 16.2.2
   @nx/jest           : 16.2.2
   @nx/linter         : 16.2.2
   @nx/workspace      : 16.2.2
   @nx/cypress        : 16.2.2
   @nx/devkit         : 16.2.2
   @nx/eslint-plugin  : 16.2.2
   @nx/next           : 16.2.2
   @nx/react          : 16.2.2
   @nrwl/tao          : 16.2.2
   @nx/web            : 16.2.2
   typescript         : 5.0.4
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information
- package manage : pnpm@7.18.0
- node : 16.16.0

_No response_

puku0x:
Hope it helps. https://github.com/nrwl/nx/pull/17266#issuecomment-1565406622

```js
[...]

PR #17806, opened 2023-06-26
fix(nextjs): Custom server should run on fresh apps without errors

ndcunningham:
Before when we ran custom server it would result in a error for fresh NextJs Apps
```
 throw new Error(`Could not find ${fileToRun}. Make sure your build succeeded.`);
 ```

This was because the server code was generated `./server/main.ts` while the `@nx/js:node` executor expects it from the root of the project.

This change should address this issue.

### DIFF
--- a/packages/js/src/executors/node/node.impl.ts
+++ b/packages/js/src/executors/node/node.impl.ts
@@ -73,9 +73,19 @@ export async function* nodeExecutor(
 
   // Re-map buildable workspace projects to their output directory.
   const mappings = calculateResolveMappings(context, options);
+  let outputFileName =
+    buildOptions.outputFileName || `${path.parse(buildOptions.main).name}.js`;
 
-  const outputFileName =
-    buildOptions.outputFileName ?? `${path.parse(buildOptions.main).name}.js`;
+  if (!buildOptions.outputFileName) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by jaysoo — APPROVED** [review]

File: packages/js/src/executors/node/node.impl.ts [hunk 1]
```diff
@@ -73,9 +73,17 @@ export async function* nodeExecutor(
...
   // Re-map buildable workspace projects to their output directory.
   const mappings = calculateResolveMappings(context, options);
-
-  const outputFileName =
-    buildOptions.outputFileName ?? `${path.parse(buildOptions.main).name}.js`;
+  const matches = buildOptions.main.match(/^(.*)\/([^/]*)$/);
+  let outputFileName =
+    buildOptions.outputFileName || `${path.parse(buildOptions.main).name}.js`;
+
+  if (matches && !buildOptions.outputFileName) {
```
jaysoo, line 80: It'd be better to move the match into the  `!buildOptions.outputFileName` if block, so we don't run the match unless required.

mandarini (approved):

puku0x:
This fixes https://github.com/nrwl/nx/issues/17270 . Thank you!

Issue #17767, opened 2023-06-24
  This platform (win32-x64) is currently not supported by Nx. after 16.4.0 update


SteveW94:
### Current Behavior

After updating from 16.3.x to 16.4 today, our CI-Pipeline nx-Tasks stopped working with the following message:

"  This platform (win32-x64) is currently not supported by Nx."

- The exact same tasks worked perfectly fine before
- yarn.lock has been freshly generated
- Latest node version is used
- The Pipeline runs on an AzureDevops Windows 10 64bit Agent

### Expected Behavior

No error messages when using nx-Tasks on Windows 10 CI agent

### Steps to Reproduce

1. Use nx 16.4 Task in CI pipeline

### Nx Report

```shell
-
```

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

canibanoglu:
I'm having the same issue but on GitHub CI with `ubuntu-latest` (`v22.04.2`). Works on my local machine (Mac `arm64`) and this started happening after the 16.4 update for me too.

salihsagdilek:
also same issue digitalocean app platform.

This platform (linux-x64) is currently not supported by Nx.

Cammisuli:
Oh no! That's not right, linux and Windows are definitely supported. This usually happens because the node_modules directory was installed with the `--no-optional` flag. 

You can read more about it here:
https://nx.dev/recipes/ci/troubleshoot-nx-install-issues#troubleshoot-nx-installations

If things still don't work after following the document, let me know.

SteveW94:
@Cammisuli 

I read about this thorough :)

We make sure that in our CI Pipeline we do always create the node_modules from scratch and Also I re-assured multiple times that we nowhere use the no-optionals flag.

Also as I mentioned the same setup works perfectly fine with NX 16.3.2 and stopped working all of a sudden after udpating to 16.4.

canibanoglu:
We're not using the `--no-optional` flag at all in the project, our CI does a clean install every time and this started happening after the update to 16.4. We had to revert to the previous version at which point everything started running as they should

Cammisuli:
Do you all have a reproduction somewhere I can take a look at? 

Particularly the package.json and CI commands that are used?

Cammisuli:
Also, if you take a look inside the node_modules/@nx directory, do you see any packages that are along the lines of `nx-platform-arch` (eg. `nx-win32-x64-msvc`)

[...]

PR #17817, opened 2023-06-27
fix(core): improve error message when native binaries are not found for a supported platform

Cammisuli:
## Current Behavior
When a platform is supported but native binaries are not found, the error message says that the platform is not supported. Even though it is. 

## Expected Behavior
A supported platform with missing dependencies will have a clearer message

## Related Issue(s)

Related https://github.com/nrwl/nx/issues/17767

### DIFF
--- a/packages/nx/bin/nx.ts
+++ b/packages/nx/bin/nx.ts
@@ -20,34 +20,20 @@ function main() {
   if (
     process.argv[2] !== 'report' &&
     process.argv[2] !== '--version' &&
-    process.argv[2] !== '--help' &&
-    !_supportedPlatform()
+    process.argv[2] !== '--help'
   ) {
-    if (
-      process.platform == 'win32' ||
-      process.platform == 'darwin' ||
-      process.platform == 'linux' ||
-      process.platform == 'freebsd'
-    ) {
+    try {
+      assertSupportedPlatform();
+    } catch (e) {
       output.error({
-        title: 'Native binary not found',
+        title: e.name,
         bodyLines: [
-          `The Nx CLI could not find a native binary for your supported platform (${process.platform}-${process.arch}).`,
-          'This likely means that optional dependencies were not installed correctly.',
-          'For troubleshooting steps, please see https://nx.dev/recipes/ci/troubleshoot-nx-install-issues#native-modules',
+          e.message,
+          'For more information please see https://nx.dev/recipes/ci/troubleshoot-nx-install-issues',
         ],
       });
       process.exit(1);
     }
-
-    output.error({
-      title: 'Platform not supported',
-      bodyLines: [
-        `This platform (${process.platform}-${process.arch}) is currently not supported by Nx.`,
-        'For a list of supported platforms, please see https://nx.dev/recipes/ci/troubleshoot-nx-install-issues#supported-native-module-platform',
-      ],
-    });
-    process.exit(1);
   }
 
   const workspace = findWorkspaceRoot(process.cwd());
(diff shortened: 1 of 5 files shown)


AgentEnder (approved):

packages/nx/bin/nx.ts
```diff
@@ -273,12 +259,30 @@ function _getLatestVersionOfNx(): string {
...
+        'This likely means that optional dependencies were not installed correctly, or your system is missing some system dependencies.';
+      if (process.env.NX_VERBOSE_LOGGING == 'true') {
+        err.message += `\n\nAdditional error information: \n${e.message}`;
+      }
+    } else {
+      err.name = 'Platform not supported';
+      err.message = `This platform (${process.platform}-${process.arch}) is currently not supported by Nx.`;
+    }
+
+    throw err;
```
AgentEnder, line 285:
```suggestion
  } catch (e) {
    const isSupportedPlatform = ['win32', 'darwin', 'linux', 'freebsd'].includes(process.platform); 
    const err = new Error(isSupportedPlatform 
    	? `The Nx CLI could not find or load the native binary for your supported platform (${process.platform}-${process.arch}). ` +
        'This likely means that optional dependencies were not installed correctly, or your system is missing some system dependencies.'
      : `This platform (${process.platform}-${process.arch}) is currently not supported by Nx.`, { 
        cause: e
       });
    err.name = isSupportedPlatform ? 'Missing Platform Dependency' : 'Platform not supported'
    if (process.env.NX_VERBOSE_LOGGING == 'true') {
        err.message += `\n\nAdditional error information: \n${e.message}`;
    }    
    throw err;
```

AgentEnder (commented):
Small suggestion for a change to keep the initial error as the cause in case it is helpful. Supported in Node 16.9+, but shouldnt cause issues if its present in older versions.

packages/nx/bin/nx.ts
Cammisuli: TypeScript doesnt seem to like the `{ cause: e }` object 😮 
image

AgentEnder: Interesting, it looks like we'd need to bump `lib` to `es2022` from `es2019`.... I personally think that should be fine, but maybe we should check w/ Jason first?

Cammisuli: Probably not in this pr. This structure changed so that we do the assert in post install too, so throwing the error here isn't applicable anymore.

packages/nx/bin/post-install.ts
```diff
@@ -3,10 +3,13 @@ import { workspaceRoot } from '../src/utils/workspace-root';
 import { fileExists } from '../src/utils/fileutils';
 import { join } from 'path';
 import { daemonClient } from '../src/daemon/client/client';
+import { assertSupportedPlatform } from '../src/native/assert-supported-platform';
 
 (async () => {
   try {
```
FrozenPandaz, line 9: I think we need to move this `try catch` into the `if` block. Otherwise, the error thrown by `assertSupportedPlatform` will not be shown.

FrozenPandaz (changes requested):

packages/nx/bin/post-install.ts
Cammisuli: It is shown, I tested it by publishing a local copy and doing `yarn --no-optional`

FrozenPandaz: Oh because we have `process.exit(1)` in there. Got it.

FrozenPandaz (approved):

Issue #16270, opened 2023-04-12
.env.local file in root of repo not working for cache inputs


lourd:
### Current Behavior

When using a `.env.local` file in the root of a repo, specifying that file as a file in a fileset that's a part of the `inputs` to a target doesn't do anything. Changing that file doesn't not invalidate the cache.

Additionally, specifying the specifying environment variables that are loaded via the `.env.local` file doesn't invalidate the cache. It appears that the caching computation happens before any .env files are loaded.

### Expected Behavior

Specifying `{workspaceRoot}/.env.local` in a fileset and then changing that file should invalidate the cache on the next execution of the project target that uses that fileset as one of its `inputs`.

It should also be made clear in the docs that environment file loading happens after cache computation, if that is the case.

### Steps to Reproduce

1. Create a .env.local file in the root of an Nx repo. Set some environment variable in that file.
1. Create a project.json and target. That target can simply be a command that echos the value of the environment variable.
2. Run the target
3. Change the value in the .env.local file
4. Run the target again — see that it uses a cached output.

### Nx Report

```shell
Node : 16.14.0
   OS   : darwin arm64
   yarn : 1.22.19
   
   nx                      : 15.9.2
   @nrwl/js                : 15.9.2
   @nrwl/jest              : 15.9.2
   @nrwl/linter            : 15.9.2
   @nrwl/workspace         : 15.9.2
   @nrwl/cli               : 15.9.2
   @nrwl/detox             : 15.9.2
   @nrwl/devkit            : 15.9.2
   @nrwl/eslint-plugin-nx  : 15.9.2
   @nrwl/next              : 15.9.2
   @nrwl/node              : 15.9.2
   @nrwl/react             : 15.9.2
   @nrwl/react-native      : 15.9.2
   @nrwl/tao               : 15.9.2
   @nrwl/webpack           : 15.9.2
   typescript              : 5.0.4
```

jensbodal:
Out of curiosity have you tested with another type of file? I was going to file a big for a similar issue but for a file that's in another project.

```
{workspaceRoot}/packages/pkgFoo/my-file
```

Only asking because if that's the case there might be an issue with files as input to caching overall rather than just environment files.

jensbodal:
Is the `.env.local` file in your `.gitignore`? I was trying to figure out a similar issue and discovered that if the file is in there it cannot be used as an input to the cache.  I'm not sure if this is by design and/or if there is a workaround.

FrozenPandaz:
[...]

PR #17831, opened 2023-06-28
Update how-caching-works.md: .gitignore and .nxignore affect how fileset caching works

jensbodal:
Add callout to mention filesets and namedInputs are the default values to a cache input and that any file ignored by .gitignore or .nxignore will not be considered as a cacheable input.

## Current Behavior

filesets and their interaction with the .gitignore and .nxignore files are not mentioned anywhere in the docs that I could find other than here: https://github.com/nrwl/nx/blob/2941ddc6a2da13e8e373c423438c061eea52eda5/docs/shared/using-nx/affected.md?plain=1#L55

## Related Issue(s)

https://github.com/nrwl/nx/issues/16270


jensbodal:
Preview available here: https://nx-fkz7143zd-nrwl.vercel.app/concepts/how-caching-works#filesets

isaacplmann (approved):
Thanks!

PR #17843, opened 2023-06-28
fix(nx-dev): skip docs for private package

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/scripts/documentation/package-schemas/generatePackageSchemas.ts
+++ b/scripts/documentation/package-schemas/generatePackageSchemas.ts
@@ -10,7 +10,7 @@ import {
 } from '@nx/nx-dev/data-access-packages';
 import { NxSchema, PackageMetadata } from '@nx/nx-dev/models-package';
 import { generateJsonFile, generateMarkdownFile } from '../utils';
-import { getPackageMetadataList } from './package-metadata';
+import { findPackageMetadataList } from './package-metadata';
 import { schemaResolver } from './schema.resolver';
 
 function processSchemaData(data: NxSchema, path: string): NxSchema {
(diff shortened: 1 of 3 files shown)


scripts/documentation/package-schemas/generatePackageSchemas.ts
```diff
@@ -37,8 +37,9 @@ export function generatePackageSchemas(): Promise<void[]> {
   console.log(`${chalk.blue('i')} Generating Package Schemas`);
   const absoluteRoot = resolve(join(__dirname, '../../../'));
 
-  const packages = getPackageMetadataList(absoluteRoot, 'packages', 'docs').map(
-    (packageMetadata) => {
+  const packages = getPackageMetadataList(absoluteRoot, 'packages', 'docs')
```
jaysoo, line 40: It's not the responsibility of `generatePackageSchemas` to know that `getPackageMetadataList` might return `null` values in the array. Generally, sparse arrays are hard to work with because you don't expect items to be undefined or null.

I could filter out the empty items in `getPackageMetadataList` instead (might the easier using `reduce` or a for loop).

The other option is to use a visitor pattern, so instead of `getPackageMetadataList` we can use something like:

```
findPackageMetadataList(absoluteRoot, 'packages', 'docs', (packageMetadata) => {
   // ...
});
```

Although in this instance I think just filtering out the values from the list is fine without changing the function signature.

xiongemi: done. update the name to findPackageMetadataList. i move the filter(Boolean) to inside function findPackageMetadataList

jaysoo (approved):
LGTM

isaacplmann (approved):

PR #17844, opened 2023-06-28
feat(vite): add tsconfig paths resolution plugin

### LINKED ISSUE #17922: `buildLibsFromSource: false` doesn't work with Vite executors
### Current Behavior

Suppose an app `my-app` uses the `@nx/vite:build` and/or `@nx/vite:dev-server` executors and depends on a lib `my-lib`. Then when the `buildLibsFromSource` option of the executor is set to `false`, `my-lib` is still built from source within `my-app`.

I created a reproducible example, see below.

This seems to be caused by a combination of two faulty behaviors:
1. When the executor generates a temporary `tsconfig.generated.json` file, it tries to figure out whether the lib is buildable. It does so by checking if it has a target with the same name as the target currently being executed for the app, see https://github.com/nrwl/nx/blob/50d01d15677eb9165e24d759d790f60af22b9fff/packages/vite/src/utils/executor-utils.ts#L43 That means when the target currently being executed is called `serve`, it looks for a target called `serve` in the lib, which doesn't make sense. The `@nx/webpack:dev-server` executor uses a hard-coded `'build'` here, which is probably not optimal either but should work in most cases.
2. When one fixes point 1. by patching `@nx/vite`, the `tsconfig.generated.json` file is generated correctly. However, it is still not respected by the build. I haven't figured out why, but I suspect it's because Nx Vite apps use the `vite-tsconfig-paths` plugin, which needs to be made aware of the generated `tsconfig` file, but isn't.

### Expected Behavior

`my-lib` is not built from source, but its build artifact is used instead.

### GitHub Repo

https://github.com/ms-tng/nx-buildlibsfromsource-repro

### Steps to Reproduce

1. Create a Vite app and a lib, where the app depends on the lib
2. Set `"buildLibsFromSource": false` for the `build` and `serve` targets of the app
3. Build the lib and build or serve the app
4. Open the app in a browser and observe that the lib is built from its sources

### Nx Report

```shell
Node   : 16.14.2
OS     : win32-x64
npm    : 8.5.0

nx                 : 16.4.2
@nx/js             : 16.4.2
@nx/linter         : 16.4.2
@nx/workspace      : 16.4.2
@nx/cypress        : 16.4.2
@nx/devkit         : 16.4.2
@nx/eslint-plugin  : 16.4.2
@nx/react          : 16.4.2
@nrwl/tao          : 16.4.2
@nx/vite           : 16.4.2
@nx/web            : 16.4.2
typescript         : 5.1.6
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [X] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

barbados-clemens:
## Current Behavior

vite-tsconfig-paths plugin isn't able to work with tmp tsconfigs to make incremental building work

## Expected Behavior

nx supports incremental building with vite

## Related Issue(s)

terminal output showing number of modules transformed being less when using incremental building

TODO:
  - [x] validation across different vite setups with/without buildable libs
  - [x] write e2e tests
  - [x] handle TODOs
  - [x] cleanup register work within executor to just generate tsconfigs

Fixes #


packages/vite/plugins/nx-tsconfig-paths.plugin.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+    },
+    resolveId(source: string) {
+      let resolvedFile: string;
+      try {
+        resolvedFile = matchTsPathEsm(source);
+      } catch (e) {
+        logIt('Using fallback path matching.');
+        resolvedFile = matchTsPathFallback?.(source);
+      }
```
barbados-clemens, line 61: This solves the issues I came across where a default @nx/js library will be built with the following package.json
```json
{
  "name": "@acme/js-lib",
  "version": "0.0.1",
  "type": "module",
  "peerDependencies": {
    "tslib": "2.6.0"
  },
  "main": "./src/index.js",
  "types": "./src/index.d.ts"
}
```
and vite will be unable to load it as RollUp claims it cannot find the named exports, or the `tsconfig-paths` will not be able to resolve the exports object since it doesn't exist.

The same issue happens in the original plugin being used as well `vite-tsconfig-paths`. 

This fallbacks to the root level tsconfig where the `.ts` files will be loaded instead so projects that don't include `export` objects or commonjs will still be able to be used in the project with `buildLibsFromSource=false` just will be forced to be build from source. 

This can be fixed in the userland by changing the tsconfig settings to use NodeNext and `type:module` in package.json which allows vite to see the named exports

i.e. the output switches from 

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const tslib_1 = require("tslib");
tslib_1.__exportStar(require("./lib/lib-js"), exports);
//# sourceMappingURL=index.js.map
```
to just

```js
export * from './lib/lib-js';
```

packages/vite/src/migrations/update-16-5-0-change-ts-paths-plugin/change-ts-paths-plugin.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      const withImportChange = tsquery.replace(
+        updatedContent,
+        'ImportDeclaration:has(StringLiteral[value="vite-tsconfig-paths"])',
+        () => {
+          return "import { nxViteTsPaths } from '@nx/vite/plugins/nx-tsconfig-paths.plugin';";
+        }
+      );
+
+      // TODO(caleb): should we try to handle the commented out usage of viteTsConfigPaths()?
```
barbados-clemens, line 51: Unsure if we need to migrate any commented out code?

packages/vite/src/utils/executor-utils.ts
```diff
@@ -43,15 +42,7 @@ export function registerPaths(
...
-    const tmpTsConfig = createTmpTsConfig(
-      tsConfig,
-      context.root,
-      projectRoot,
-      dependencies
-    );
-
-    registerTsConfigPaths(tmpTsConfig);
-  } else {
-    registerTsConfigPaths(tsConfig);
```
barbados-clemens, line 55: As far as I can tell, since these paths never being used by vite downstream, there isn't a reason to register them anymore.

packages/vite/plugins/nx-tsconfig-paths.plugin.ts
barbados-clemens: @jaysoo would like your opinion on this, to see if this makes sense or if we should see what we can do to get cjs built libs working inside vite.

packages/vite/src/migrations/update-16-5-0-change-ts-paths-plugin/change-ts-paths-plugin.ts
mandarini: I think let's not, since it's more complicated and less straightforward, since we will not be able to use tsquery for that. So let's leave commented out code as is, is my opinion. Better left as is, rather than a mistake in migration.

mandarini (commented):
Interesting, and thank you @barbados-clemens !! I'm going to test this out today. I am thinking, should we really drop the use of `vite-tsconfig-paths` plugin, and rely only on the `tsconfig-paths`? I'm thinking that maybe `vite-tsconfig-paths` covers some specific use cases for vite, and using `vite-tsconfig-paths` in the `nx-vite-paths` plugin (instead of `tsconfig-paths`) could potentially save us some maintenance? Again, have not tried out the PR yet, will do today, and not sure what more vite-tsconfig-paths does, so I'm just speculating!

In any case, I'll test it out today! :D thanks! 

I also want to see @jaysoo opinion!

mandarini:
Cool! So, yes, I tested on this repo: https://github.com/mandarini/vite-paths/tree/test-caleb on the branch `test-caleb`, which I added lots of interdependencies, and some to cover Storybook path resolution as well. All works great! So it's a ✅ from me, but with the question whether we really want to drop `vite-tsconfig-paths` completely, and not rely on it at all!

packages/vite/migrations.json
```diff
@@ -29,6 +29,12 @@
       "description": "Changes the testFile config in the vite:test exectutor from a string to an array of strings",
       "cli": "nx",
       "implementation": "./src/migrations/update-16-4-1-update-test-file-config/update-16-4-1-test-file-config"
+    },
+    "16-5-0-change-ts-paths-plugin": {
+      "version": "16.5.0-beta.0",
```
mandarini, line 34: Increment this version, to beta.4 or something? Because if you're already on `next`, which is `beta.2` then it will not get it I think.

packages/vite/src/generators/configuration/__snapshots__/configuration.spec.ts.snap
```diff
@@ -4,7 +4,7 @@ exports[`@nx/vite:configuration library mode should add config for building libr
 "/// <reference types="vitest" />
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
-import viteTsConfigPaths from 'vite-tsconfig-paths';
```
mandarini, line 7: We should remove the installation of `vite-tsconfig-paths` from the init/config generator, and the version from the versions.ts file, I think, if we are to drop it altogether, right?

mandarini (approved):
Approving to move this forward, but still would like one look from @jaysoo

PR #17848, opened 2023-06-29
fix(repo): Make storybook migration more robust if missing fields in tsconfig.json

rodrigc:
## Current Behavior
I ran into some cases in my tree when trying to migrate to nx 16.5.0-beta0, the migration errored out with stack traces, because the tsconfig.json files in my tree were missing certain keys.

## Expected Behavior

Migration should be more robust if keys are missing in tsconfig.json.  Don't error out with stack traces.

### DIFF
--- a/packages/storybook/src/generators/configuration/lib/util-functions.ts
+++ b/packages/storybook/src/generators/configuration/lib/util-functions.ts
@@ -682,6 +682,7 @@ export function renameAndMoveOldTsConfig(
         json.extends = json.extends.replace('../', './');
       }
 
+      json.files ??= [];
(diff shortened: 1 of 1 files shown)


rodrigc:
Before I submitted this patch, these migrations were failing in my tree in a bunch of places
when I did not have certain values set in tsconfig.json.

I submitted this change a while ago.

However, since I submitted this patch, the code has changed, and has more for loops iterating over arrays,
so if certain values are undefined, the migrations will not fail any more.

So I think this can be closed.

mandarini:
Thank you @rodrigc for looking into it! :D

PR #17865, opened 2023-06-29
docs(nx-cloud): ami deplopyment docs

rarmatei:
- docs(nx-cloud): add deployment docs
- docs(nx-cloud): add correct links
- docs(nx-cloud): format files
- docs(nx-cloud): add sourcefiles to map
- docs(nx-cloud): adjust docs map
- docs(nx-cloud): adjust docs map
- docs(nx-cloud): adjust docs map

## Related Issue(s)

Fixes #


docs/nx-cloud/private/ami-setup.md
```diff
@@ -0,0 +1,143 @@
...
+  # samlEntryPoint: 'your_saml_entry_point'
+  # samlCert: 'saml_cert'
+```
+
+### Upgrades
+
+We send out emails with every new NxCloud release to all our Enterprise customers:
+
+1. You can view your current version at the `/version` route: https://your-nx-cloud-url.com/version
+2. [And these are the latest NxCloud releases](https://nx.dev/nx-cloud/reference/release-notes#docker-containers)
```
barbados-clemens, line 131: recomment linking to the top of the page and not part of the way down. 
```suggestion
2. [And these are the latest NxCloud releases](https://nx.dev/nx-cloud/reference/release-notes)
```

docs/nx-cloud/private/get-started.md
```diff
@@ -1,23 +1,62 @@
...
+- Dedicated instances
+- Storage encryption
+- Storage replication / redundancy
+- No external API calls to any services outside of the cluster we set-up for you
+
+This would be a "best of both worlds" option, as it would free you up from managing the instance yourself, but you will get to define specific parameters of how it should it run.
+Your data and the NxCloud will run in complete isolation and will only serve your company. Once you let us know you'd like this option, depending on the agreed requirements,
+it might take a few days to get it set up.
+
+If you'd like to use any of the above options, please mention this to your Enterprise support contact, and they'll be able to set you up with an unlimited use license.
```
barbados-clemens, line 41: Maybe link to the Enterprise contact form?

docs/nx-cloud/private/ami-setup.md
```diff
@@ -0,0 +1,143 @@
...
+# Set the external URL your instance is running on
+nxCloudAppURL: 'https://nx-cloud.on.my-domain.ca'
+
+# Uncomment (along with github secrets below) to enable working with GitHub pull requests or github auth
+#github:
+#  auth:
+#    enabled: false
+
+#gitlab:
+#  apiUrl: '' # this is only needed if you have a self-hosted instance
```
barbados-clemens, line 85: might be just _me_ but I thought this was for is you're doing self hosting nx cloud, not self hosted gitlab.
```suggestion
#  apiUrl: '' # this is only needed if you have a self-hosted gitlab instance
```

docs/nx-cloud/private/ami-setup.md
```diff
@@ -0,0 +1,143 @@
+# Setting up a dedicated NxCloud VM
+
+## AWS EC2
+
+1. Login to your AWS Console and select [the top image published here](https://us-east-1.console.aws.amazon.com/ec2/v2/home?region=us-east-1#Images:visibility=public-images;imageName=nx-cloud;owner=623002322076;sort=desc:imageName)
```
barbados-clemens, line 5: will this force people to create the instances in us-east-1 or can you easily switch to a different region? Ideally link is pointed to just the aws console and region is inferred from aws side?

docs/nx-cloud/private/ami-setup.md
```diff
@@ -0,0 +1,143 @@
...
+4. You will need to SSH into the instance once it's created:
+   - Use an existing SSH key-pair that you already have installed locally.
+   - [Or create a new one and download the keys locally](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html?icmpid=docs_ec2_console#having-ec2-create-your-key-pair)
+     - Then select your new SSH pair from the list
+5. Networking:
+   - Allow the instance to receive HTTP and HTTPS traffic
+   - Allow SSH from your current IP
+6. Leave the storage options as they are (external storage of 50GB)
+7. "Launch instance"
+8. Wait 10 minutes, then navigate to your instance's IP in the browser. You should see the NxCloud dashboard!
```
barbados-clemens, line 17: would it be worth adding in a screenshot of what the dashboard looks like?

barbados-clemens (commented):
Looks good! just a suggestion or two and some of my initial thoughts.

docs/nx-cloud/private/get-started.md
barbados-clemens:
```suggestion
If you'd like to use any of the above options, please mention this to your [Enterprise support contact](https://nx.app/enterprise?utm_source=nx.dev), and they'll be able to set you up with an unlimited use license.
```

docs/nx-cloud/private/ami-setup.md
rarmatei: good shout!

docs/nx-cloud/private/ami-setup.md
rarmatei: ah yes, good spot, didn't realise there was a specific region in the link - I'll update it to be generic

docs/nx-cloud/private/ami-setup.md
rarmatei: yes, you're right! I'll fix the wording!

docs/nx-cloud/private/get-started.md
rarmatei: I removed this line, but you're right, I'll make sure to include a link to the form at the top

docs/nx-cloud/private/ami-setup.md
rarmatei: the top of the page shows releases for the NXCloud runner. In this case, I want to direct to the Docker image versions.

barbados-clemens: Ah okay 👍

docs/nx-cloud/private/ami-setup.md
```diff
@@ -48,7 +52,19 @@ secret:
 scp -r ./myconfiguration.yaml nx-cloud@<your-instance-ip>:~/config/user/update.yaml
 ```
 
-That's it! After a few minutes, you should be able to log-in using your password!
+That's it! After a few minutes, you should be able to log-in with:
+
+- username: `,admin`
```
barbados-clemens, line 57: should it be `admin`?

docs/map.json
```diff
@@ -1578,16 +1583,6 @@
...
-            },
-            {
-              "name": "Kubernetes Setup",
-              "id": "kubernetes-setup",
-              "file": "nx-cloud/private/kubernetes-setup"
-            },
-            {
-              "name": "Standalone",
-              "id": "standalone",
-              "file": "nx-cloud/private/standalone"
```
isaacplmann, line 1590: we need to add redirect rules for these two pages that are deleted.  I'm assuming they should redirect to the new page you made.

isaacplmann (changes requested):

docs/map.json
rarmatei: ah good catch - didn't realise we can add redirects 👍 

should be good now I think!

docs/nx-cloud/private/ami-setup.md
```diff
@@ -0,0 +1,170 @@
...
+# only use this if you'd like to use any of the newer NxCloud version from here: https://nx.dev/nx-cloud/reference/release-notes#docker-containers
+# global.imageTag: ''
+
+# Set the external URL your instance is running on
+nxCloudAppURL: 'https://nx-cloud.on.my-domain.ca'
+
+# Uncomment (along with github secrets below) to enable working with GitHub pull requests or github auth
+#github:
+#  auth:
+#    enabled: false
```
stevepentland, line 106: I think we spoke about this before, but when showing config samples we should encourage the user to set both the secret and the explicit configuration option. While we do have logic that will ensure it is set or unset correctly based on the secrets provided, I would rather they not be relying on that by default in case we change this behaviour for some reason in the future
```suggestion
github:
  auth:
    enabled: true
```

stevepentland (approved):
One small comment about the config example, otherwise it seems solid to me

isaacplmann (approved):

PR #17883, opened 2023-06-30
fix(linter): create-package-json should omit non-npm and ignored packages

meeroslav:
This PR fixes several issues that are blocking adoption of the `dependency-checks` rule:
- Schema typo for `ignoredDependencies`
- Ignore peer dependencies of ignored packages when collecting project dependencies
- Ignore non-npm dependencies (this is a bug also for `createPackageJson` if project contains also non-JS code like `nx` package does)

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -45,7 +45,7 @@ export default createESLintRule<Options, MessageIds>({
         type: 'object',
         properties: {
           buildTargets: [{ type: 'string' }],
-          ignoreDependencies: [{ type: 'string' }],
+          ignoredDependencies: [{ type: 'string' }],
           checkMissingDependencies: { type: 'boolean' },
           checkObsoleteDependencies: { type: 'boolean' },
           checkVersionMismatches: { type: 'boolean' },
--- a/packages/nx/src/plugins/js/package-json/create-package-json.ts
+++ b/packages/nx/src/plugins/js/package-json/create-package-json.ts
@@ -181,6 +181,7 @@ export function findProjectsNpmDependencies(
   target: string,
   options: {
     helperDependencies?: string[];
+    ignoredDependencies?: string[];
   },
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -45,7 +45,7 @@ export default createESLintRule<Options, MessageIds>({
         type: 'object',
         properties: {
           buildTargets: [{ type: 'string' }],
-          ignoreDependencies: [{ type: 'string' }],
+          ignoredDependencies: [{ type: 'string' }],
```
meeroslav, line 48: Optionally, instead of adding another param, we can expose peer dependencies as a nested structure under parent dependency, so that source function (e.g. linter rule) can easily discard all peer dependencies of the ignored packages.

packages/nx/src/plugins/js/package-json/create-package-json.ts
```diff
@@ -181,6 +181,7 @@ export function findProjectsNpmDependencies(
   target: string,
   options: {
     helperDependencies?: string[];
+    ignoredDependencies?: string[];
```
FrozenPandaz, line 184: I think it's better to delete this after the fact...

I also do not like the fact that `helperDependencies` is here.

```ts
const ignoredDependencies = ['react'];
const packageJson = createPackageJson(...);
for (const ignoredDep of ignoredDependencies) {
  delete packageJson.dependencies[ignoredDep];
  delete packageJson.devDependencies[ignoredDep];
  delete packageJson.peerDependencies[ignoredDep];
}
```

FrozenPandaz (changes requested):

FrozenPandaz (approved):

packages/nx/src/plugins/js/package-json/create-package-json.ts
FrozenPandaz: Okay, it wouldn't work for transitive dependencies. I guess this is fine.

PR #17897, opened 2023-06-30
docs(nx-dev): nest 5 minute tutorials under getting started

meeroslav:
This PR introduces following changes:
- Nested menu elements have indentation for better visual orientation
- The 5 minute tutorials is nested under `Getting Started` and collapsible

## Current Behavior
The `5 minute tutorials` is taking too much space, although it's relevant only to beginners in Nx
Screenshot 2023-07-01 at 00 15 30

## Expected Behavior
The `5 minute tutorials` is nested under getting started
Screenshot 2023-07-01 at 00 16 22

The `5 minute tutorials` is collapsible when not active
Screenshot 2023-07-01 at 00 16 42

## Related Issue(s)

Fixes #


juristr:
I see the point...though I wonder whether there's a better way 🤔. I agree that contextually it belongs under the "Getting Started". It started with just two of them but now we're getting more with standalone / package-based / monorepo flavors. There's another discussion about exactly that and how to organize those.

Maybe there needs to be a short section with one or two of them which we think is worth emphasizing & then a "more >>" part that leads to all the other tutorials. Which has my brain circle back to a "Learn" section we had discussed 🤔

bcabanes (approved):

PR #17907, opened 2023-07-01
docs(nx-dev): Fix readability

ShrootBuck:
Just a small tweak, that I think would make it more easy to understand


ShrootBuck:
It looks fine on Github but the docs site renders it differently and that's where I believe it creates confusion.

meeroslav (approved):

juristr (approved):

PR #17910, opened 2023-07-02
docs(misc): update the file path in the react-standalone.md

pjchender:
Update the file path in the react-standalone doc from `modules/products/src/lib/product-list.tsx` to `modules/products/src/lib/product-list/product-list.tsx`.

## Current Behavior

It shows that the file path is `modules/products/src/lib/product-list.tsx`.

image

## Expected Behavior

It should be `modules/products/src/lib/product-list/product-list.tsx`.


juristr (approved):
good catch, thank you!

PR #17919, opened 2023-07-03
feat(angular): add --bundler flag for applications

Coly010:
## Current Behavior

We currently do not have a way for Angular users to opt into using the esbuild builder for Angular.

## Expected Behavior

Add a `--bundler` option that allows users to opt into the esbuild from Angular.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/application/schema.json
+++ b/packages/angular/src/generators/application/schema.json
@@ -164,6 +164,12 @@
       "description": "Generate a Angular app with a minimal setup.",
       "type": "boolean",
       "default": false
+    },
+    "bundler": {
+      "description": "Bundler to use to build the application.",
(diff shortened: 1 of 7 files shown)


packages/angular/src/generators/application/schema.json
```diff
@@ -164,6 +164,12 @@
       "description": "Generate a Angular app with a minimal setup.",
       "type": "boolean",
       "default": false
+    },
+    "bundler": {
+      "description": "Bundler use to build the application.",
```
leosvelperez, line 169:
```suggestion
      "description": "Bundler to use to build the application.",
```

leosvelperez (approved):

PR #17920, opened 2023-07-03
feat(misc): change updateBuildableProjectDepsInPackageJson option default value to false

leosvelperez:
This is the first step to moving from using the current package.json generation in executors to using the new `@nx/dependency-checks` EsLint rule instead. In a follow-up PR, the options will be deprecated.

The PR comes with migrations for targets using the relevant executors and are relying on the default value for the `updateBuildableProjectDepsInPackageJson`. In those cases, the option will be explicitly set to `true` in the target configuration to keep the existing behavior.

## Current Behavior

The `updateBuildableProjectDepsInPackageJson` option of several executors has a default value of `true`.

## Expected Behavior

The `updateBuildableProjectDepsInPackageJson` option of several executors should have a default value of `false`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps.ts
+++ b/packages/js/src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps.ts
@@ -1,0 +1,44 @@
+import {
+  formatFiles,
+  getProjects,
+  Tree,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+
+const executors = new Set([
+  '@nx/js:swc',
+  '@nrwl/js:swc',
+  '@nx/js:tsc',
+  '@nrwl/js:tsc',
+]);
+
+export default async function (tree: Tree) {
+  const projects = getProjects(tree);
+
+  for (const [projectName, project] of projects) {
+    if (project.projectType !== 'library') {
+      continue;
+    }
+
+    let updated = false;
+    for (const [, target] of Object.entries(project.targets || {})) {
+      if (!executors.has(target.executor)) {
+        continue;
+      }
+
+      if (
+        target.options &&
+        target.options.updateBuildableProjectDepsInPackageJson === undefined
+      ) {
+        target.options.updateBuildableProjectDepsInPackageJson = true;
+        updated = true;
+      }
+    }
+
+    if (updated) {
+      updateProjectConfiguration(tree, projectName, project);
+    }
+  }
+
+  await formatFiles(tree);
+}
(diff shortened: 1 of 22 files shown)


isaacplmann (approved):
Looks fine from docs perspective

packages/js/src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps.ts
```diff
@@ -0,0 +1,44 @@
...
+  '@nx/js:swc',
+  '@nrwl/js:swc',
+  '@nx/js:tsc',
+  '@nrwl/js:tsc',
+]);
+
+export default async function (tree: Tree) {
+  const projects = getProjects(tree);
+
+  for (const [projectName, project] of projects) {
```
jaysoo, line 18: `forEachExecutorOptions` might be easier for this case like https://github.com/nrwl/nx/blob/master/packages/node/src/migrations/update-16-3-1/update-webpack-executor.ts

packages/js/src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps.ts
```diff
@@ -0,0 +1,44 @@
+import {
+  formatFiles,
+  getProjects,
+  Tree,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+
+const executors = new Set([
```
jaysoo, line 8: `@nx/angular` also has two executors, are we handling that in another PR?

jaysoo (approved):

packages/js/src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps.ts
leosvelperez: It was a conscious decision not to use it to avoid another loop for the executors. The `forEachExecutorOptions` function works for a single executor, so I would need to loop the executors to use that function.

packages/js/src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps.ts
leosvelperez: The `@nx/angular` executors are handled in a migration of that plugin in this same PR (https://github.com/nrwl/nx/pull/17920/files#diff-9ea34fa47d7f4b3da11f7f800b438573206e052089a46c80b97735bfe6b5d616).

There's also a migration for the `@nx/rollup:rollup` in the `@nx/rollup` plugin in this PR (https://github.com/nrwl/nx/pull/17920/files#diff-0399a3cd84ef9e9977ea32220e0889c4c76ae361a28f84124a64ddccd965c79a).

Issue #17081, opened 2023-05-18
Caching of builds broken in nx > 16.1.4


atsjo:
### Current Behavior

Cached library builds is not reused...

### Expected Behavior

Cached library builds is reused in new builds, as in 16.1.4.

### Steps to Reproduce

1. upgrade to 16.2.1
2. run yarn nx reset
3. build app depending on library
4. build library (library builds again, even if nothing changed)

In 16.1.4, the last build would reuse cached build...

### Nx Report

```shell
Node   : 18.16.0
   OS     : win32 x64
   yarn   : 4.0.0-rc.44
   Hasher : Native
   
   nx                 : 16.2.1
   @nx/js             : 16.2.1
   @nx/jest           : 16.2.1
   @nx/linter         : 16.2.1
   @nx/workspace      : 16.2.1
   @nx/angular        : 16.2.1
   @nx/cypress        : 16.2.1
   @nx/devkit         : 16.2.1
   @nx/esbuild        : 16.2.1
   @nx/eslint-plugin  : 16.2.1
   @nrwl/tao          : 16.2.1
   @nx/webpack        : 16.2.1
   typescript         : 5.0.4
   ---------------------------------------
   Community plugins:
   angular-calendar : 0.31.0
```

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

atsjo:
seems to work fine if not running yarn nx reset after upgrading...

atsjo:
I suspect this https://github.com/nrwl/nx/pull/17139 is a fix for this issue

atsjo:
still the same with 16.3.0... Am I the only one experiencing this? must be something special about my workspace...

If I run yarn nx run-many --target=build and then build one of the individual projects after that, it re-builds, not using the cache....

atsjo:
16.3.2 seems to fix this, even if I can't find anything referring this issue in the changelog... so closing the issue

atsjo:
Unfortunately this was not fixed... Sometimes it works, and sometimes not...

Now I have also migrated my repo to pnpm, and I'm experiencing the same issues as I did with yarn...

atsjo:
I think I managed to track down the problem...

I was bundling firebase functions via esbuild in a dist folder below the project (not dist in repo root), and generated a package.json with external deps there. When this happened many cached project build was invalidated. I avoided this by adding the project dist folder to .nxignore, and then everything worked as expected.

[...]

PR #17926, opened 2023-07-03
fix(core): ensure external dependency hashes are resolved in a deterministic way

### LINKED ISSUE #17917: Cache Invalidation Issue in CI/CD Pipeline (Jenkins) - Inconsistent Hash Value
### Current Behavior

I am encountering an issue with cache invalidation while running tasks in my CI/CD pipeline using Jenkins with the Nx workspace. The problem specifically relates to the inconsistent hash value of the cache, leading to unexpected behavior during the pipeline execution. even after replaying the same pipeline-build without doing any single change, cache is invalidated as hash value is changing

### Expected Behavior

The cache should be utilized effectively during the pipeline run, allowing for faster execution times by reusing previously cached results where possible.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Configure Jenkins to run the Nx tasks in the CI/CD pipeline.
2. Execute the pipeline and observe the cache behavior.

### Nx Report

```shell
npm  : 9.5.1
   
   nx               : 15.7.0
   @nrwl/jest       : 15.9.4
   @nrwl/linter     : 15.9.4
   @nrwl/workspace  : 15.7.0
   @nrwl/angular    : 15.9.4
   @nrwl/cli        : 15.7.0
   @nrwl/cypress    : 15.9.4
   @nrwl/devkit     : 15.9.4
   @nrwl/js         : 15.9.4
   @nrwl/tao        : 15.7.0
   @nrwl/webpack    : 15.9.4
   typescript       : 4.9.5
   Community plugins:
   @ngrx/effects              : 15.3.0
   @ngrx/store                : 15.3.0
   @nx/webpack                : 16.2.1
   rxjs                       : 6.6.7
   @ng-bootstrap/ng-bootstrap : 14.0.1
   @ngrx/store-devtools       : 15.3.0
   @reef/fwk                  : 4.0.7
   ngx-build-plus             : 15.0.0
```

### Failure Logs

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

1. I have observed that the cache behavior in my local development environment is working as expected, with consistent and 
 valid hash values.
2. The cache invalidation issue only arises when running the pipeline in the CI/CD environment using Jenkins.
3. The problem occurs consistently across multiple pipeline runs.

skrtheboss:
## Current Behavior

External dependency hashes are not deterministic and could cause cache misses, since the hash is dependent on which task was hashed when. 

## Expected Behavior

Task hashes should not depend on which task have been hashed earlier.

## Related Issue(s)

Fixes #17917

### DIFF
--- a/packages/nx/src/hasher/task-hasher.ts
+++ b/packages/nx/src/hasher/task-hasher.ts
@@ -332,10 +332,7 @@ class TaskHasherImpl {
                     visited
                   );
                 } else {
-                  const hash = this.hashExternalDependency(
-                    task.target.target,
-                    d.target
-                  );
+                  const hash = this.hashExternalDependency(d.source, d.target);
                   return {
                     value: hash,
                     details: {
@@ -445,20 +442,22 @@ class TaskHasherImpl {
         partialHashes.push(node.data.version);
       }
       // we want to calculate the hash of the entire dependency tree
-      this.projectGraph.dependencies[targetProjectName]?.forEach((d) => {
-        if (
-          !visited.has(
-            this.computeExternalDependencyIdentifier(
-              targetProjectName,
-              d.target
+      if (this.projectGraph.dependencies[targetProjectName]) {
+        this.projectGraph.dependencies[targetProjectName].forEach((d) => {
+          if (
+            !visited.has(
+              this.computeExternalDependencyIdentifier(
+                targetProjectName,
+                d.target
+              )
             )
-          )
-        ) {
-          partialHashes.push(
-            this.hashExternalDependency(targetProjectName, d.target, visited)
-          );
-        }
-      });
+          ) {
+            partialHashes.push(
+              this.hashExternalDependency(targetProjectName, d.target, visited)
+            );
+          }
+        });
+      }
 
       partialHash = hashArray(partialHashes);
     } else {
@@ -494,7 +493,7 @@ class TaskHasherImpl {
       const executorPackage = target.executor.split(':')[0];
       const executorNodeName =
         this.findExternalDependencyNodeName(executorPackage);
-      hash = this.hashExternalDependency(targetName, executorNodeName);
+      hash = this.hashExternalDependency(projectName, executorNodeName);
     } else {
       // use command external dependencies if available to construct the hash
       const partialHashes: string[] = [];
@@ -506,7 +505,7 @@ class TaskHasherImpl {
           const externalDependencies = input['externalDependencies'];
           for (let dep of externalDependencies) {
             dep = this.findExternalDependencyNodeName(dep);
-            partialHashes.push(this.hashExternalDependency(targetName, dep));
+            partialHashes.push(this.hashExternalDependency(projectName, dep));
           }
         }
       }
--- a/packages/nx/src/hasher/task-hasher.ts
+++ b/packages/nx/src/hasher/task-hasher.ts
@@ -332,7 +332,10 @@ class TaskHasherImpl {
                     visited
                   );
                 } else {
-                  const hash = this.hashExternalDependency(d.target);
+                  const hash = this.hashExternalDependency(
+                    task.target.target,
+                    d.target
+                  );
                   return {
                     value: hash,
                     details: {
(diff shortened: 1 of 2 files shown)


skrtheboss:
The regression has been introduced by https://github.com/nrwl/nx/pull/16926.

This issue is preventing us to upgrade to 16.4.x, since our build cache no longer works correctly, and even without changing anything, the build needs to re-execute nearly all targets.

skrtheboss:
Since @meeroslav is the author of the pull request, i am going to mention him.

meeroslav (approved):
Looks good to me

meeroslav:
Have you pinpointed where (and why) exactly is the error happening on CI? The proposed PR looks ok, and I think it improves accuracy, but I don't see any evidence supporting that this would actually fix the linked problem.

skrtheboss:
@meeroslav i have added a test which fails if the patch is not applied.

The hashed value of appB was dependent if appA was hashed prior to it.

packages/nx/src/hasher/task-hasher.ts
```diff
@@ -429,13 +429,14 @@ class TaskHasherImpl {
...
-        this.projectGraph.dependencies[projectName].forEach((d) => {
-          if (!visited.has(d.target)) {
-            partialHashes.push(this.hashExternalDependency(d.target, visited));
-          }
-        });
-      }
+      this.projectGraph.dependencies[projectName]?.forEach((d) => {
+        if (!visited.has(d.target)) {
+          partialHashes.push(
+            this.hashExternalDependency(d.target, new Set(visited))
```
FrozenPandaz, line 435: I don't like that a new `Set` is created each time. A lot more memory will be used.

When we hash an external dependency, there is always a `source` and a `target`.

As you noticed ,when `A -> C` and `B -> C`, these are two different dependencies which should both be visited. I think the `visited` Set should have both `source` and `target` as items. When C is hashed for A, it should record that `A -> C` has been visited. and when C is then visited to be hashed for `B -> C` it won't find it in the `Set` and it will hash.

So I think the signature here should change to:

```ts
  private hashExternalDependency(
    sourceProjectName: string,
    targetProjectName: string,
    visited = new Set<string>()
  ): string {
    visited.add(`${sourceProjectName}->${targetProjectName}`);
    ...
```

FrozenPandaz (changes requested):

packages/nx/src/hasher/task-hasher.ts
skrtheboss: I have changed the code accordingly, let me know if it is correct.

skrtheboss:
Rebased and applied requested changes in the fixup commit.

packages/nx/src/hasher/task-hasher.ts
```diff
@@ -332,7 +332,10 @@ class TaskHasherImpl {
                     visited
                   );
                 } else {
-                  const hash = this.hashExternalDependency(d.target);
+                  const hash = this.hashExternalDependency(
+                    task.target.target,
```
meeroslav, line 336: This should be source from the dependency. Task target would not be accurate, as it's a name of the target not a package.
```suggestion
                    d.source,
```

packages/nx/src/hasher/task-hasher.ts
```diff
@@ -429,22 +445,30 @@ class TaskHasherImpl {
...
       }
       // we want to calculate the hash of the entire dependency tree
-      if (this.projectGraph.dependencies[projectName]) {
-        this.projectGraph.dependencies[projectName].forEach((d) => {
-          if (!visited.has(d.target)) {
-            partialHashes.push(this.hashExternalDependency(d.target, visited));
-          }
-        });
-      }
+      this.projectGraph.dependencies[targetProjectName]?.forEach((d) => {
```
meeroslav, line 448: Please keep the original format as optional chaining adds additional unnecessary checks when compiled to javascript. Also it's a bit more readable in my opinion.

packages/nx/src/hasher/task-hasher.ts
```diff
@@ -470,7 +494,7 @@ class TaskHasherImpl {
       const executorPackage = target.executor.split(':')[0];
       const executorNodeName =
         this.findExternalDependencyNodeName(executorPackage);
-      hash = this.hashExternalDependency(executorNodeName);
+      hash = this.hashExternalDependency(targetName, executorNodeName);
```
meeroslav, line 497: This is not correct. The source is the project name.
```suggestion
      hash = this.hashExternalDependency(projectName, executorNodeName);
```

packages/nx/src/hasher/task-hasher.ts
```diff
@@ -482,7 +506,7 @@ class TaskHasherImpl {
           const externalDependencies = input['externalDependencies'];
           for (let dep of externalDependencies) {
             dep = this.findExternalDependencyNodeName(dep);
-            partialHashes.push(this.hashExternalDependency(dep));
+            partialHashes.push(this.hashExternalDependency(targetName, dep));
```
meeroslav, line 509: Same here, source is the project name
```suggestion
            partialHashes.push(this.hashExternalDependency(projectName, dep));
```

meeroslav (changes requested):

skrtheboss:
Hi @meeroslav i have rebased again and applied requested changes in the 2nd fixup commit.

meeroslav (approved):
Looks good! Thank you!

meeroslav (approved):

FrozenPandaz (approved):

FrozenPandaz:
Thank you for making the changes! LGTM :tada:

atsjo:
Betting this will fix the caching problems I have observed as well since 16.2.x, and reported in https://github.com/nrwl/nx/issues/17081
Managed to make it work most of the time by restructuring one of the projects in my repo, but sometimes it still doesn't use the cache, and it seems non-deterministic.

meeroslav:
@skrtheboss we are planning a patch today

PR #17934, opened 2023-07-04
Update use-environment-variables-in-angular.md

muhamedkarajic:
Just a snippet for how to add enviroment variable + running some nx command for all OS platforms.


muhamedkarajic:
Just added this snippet how to quickly run it without having a .evn file. Useful for docker etc.

isaacplmann:
I appreciate the added snippet, but the formatting is a bit off and I would put it in a different section of the page.  I think I'll create a new PR with your snippet and add your username as a co-author.  I'll link the PR here when its ready

PR #17953, opened 2023-07-04
fix(js): add watchIgnore and runBuildTargetDependencies options to speed up build

### LINKED ISSUE #17070: >  NX  File change detected. Restarting...
### Current Behavior

@nx:js:node dev server is stuck on `>  NX  File change detected. Restarting...` since upgrading to 16.2.1

### Expected Behavior

Dev server to work as should. I did not have this issue with 16.1.4.

### GitHub Repo

_No response_

### Steps to Reproduce

Just run npx nx run <app>:serve with @nx/js:node as the executor.

### Nx Report

```shell
Node   : 18.16.0
   OS     : darwin arm64
   npm    : 9.5.1
   Hasher : Native

   nx                 : 16.2.1
   @nx/js             : 16.2.1
   @nx/jest           : 16.2.1
   @nx/linter         : 16.2.1
   @nx/workspace      : 16.2.1
   @nx/cypress        : 16.2.1
   @nx/devkit         : 16.2.1
   @nx/esbuild        : 16.2.1
   @nx/eslint-plugin  : 16.2.1
   @nx/express        : 16.2.1
   @nx/next           : 16.2.1
   @nx/node           : 16.2.1
   @nx/react          : 16.2.1
   @nrwl/tao          : 16.2.1
   @nx/web            : 16.2.1
   nx-cloud           : 16.0.5
   typescript         : 5.0.4
   Community plugins:
   @nx-tools/nx-container : 4.0.3
   @nx-tools/nx-prisma    : 4.0.3
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

jaysoo:
This PR addresses the slowness issues introduced when we updated the Node executor. Two additional options have been added.

1. ~~`watchIgnore` - A list of glob patterns to skip rebuilds (when the only changed files are ignored).~~
2. `watchRunBuildTargetDependencies` - When set to `true`, the rebuild will rerun dependency tasks. Default is `false`.

~~(1) lets users ignore files they know cannot affect the application. The alternative is to add to `.gitignore` or `.nxignore`, but this provides more granular control that is specific to the application.~~

Instead of (1), users should add files to their `.nxignore`.

(2) is for advanced usage where the application may not be bundled (thus dependencies must be rebuilt).

## Current Behavior
Node executor is slow for everyone.

## Expected Behavior
Node executor is fast for the normal use cases, with and advanced option to opt-into the slower but more correct build behavior (rerunning all dependency targets).

## Related Issue(s)

Fixes #17070

### DIFF
--- a/docs/generated/packages/js/executors/node.json
+++ b/docs/generated/packages/js/executors/node.json
@@ -28,12 +28,14 @@
       "host": {
         "type": "string",
         "default": "localhost",
-        "description": "The host to inspect the process on."
+        "description": "The host to inspect the process on.",
+        "x-priority": "important"
       },
       "port": {
         "type": "number",
         "default": 9229,
-        "description": "The port to inspect the process on. Setting port to 0 will assign random free ports to all forked processes."
+        "description": "The port to inspect the process on. Setting port to 0 will assign random free ports to all forked processes.",
+        "x-priority": "important"
       },
       "inspect": {
         "oneOf": [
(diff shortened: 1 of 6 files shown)


ndcunningham:
What if we have 
-  `watchIgnore: 'libs/my-lib'` 
- `watchRunBuildTargetDependencies: true` 
- `my-lib` is a dependency of the `app` and I make changes to my app.

Is the preference since they both options can affect the same file?

kurt-west:
Personally, I think part of the perceived slowness is that the executor now eats the compilation logs. Previously, the user would see the webpack output so they knew it was working, now it just sits there without any feedback until it boots or fails. It would be convenient to have the output from the build task be written to the logger so the user can see what's actually being build/rebuilt.

With the ignore option, it would be nice to have a flag to write the output of the watcher to the console to see what triggered the rebuild.

thenglong (approved):

urugator:
> eats the compilation log

I also use to put `console.logs` directly to configuration files like `webpack.config.js` for debugging purposes - knowing whether it's even being used, what env vars it sees, what's already configured and how, plugins, etc...

mandarini (approved):

jaysoo:
> What if we have
> 
> * `watchIgnore: 'libs/my-lib'`
> * `watchRunBuildTargetDependencies: true`
> * `my-lib` is a dependency of the `app` and I make changes to my app.
> 
> Is there a preference since both options can affect the same file?

The build for `my-lib` will still happen. The `watchIgnore` option is controlling whether the event triggers in the first place, and the latter controls whether the re-run task dependencies after the event is triggered.

ndcunningham (approved):

PR #17969, opened 2023-07-05
chore(repo): bump versions of ci images

meeroslav:
There are currently some issues with node images on CircleCI.

The latest `cimg/rust-browsers` and `macos.x86.medium.gen2` already come with fresh Node v18, so there's no need to use extra installation via orb.

This also speeds up our pipeline because we skip expensive step of downloading node image and installing it.

## Current Behavior
We use `circleci/node` orb to set up latest node version on top of the existing node installed with executor docker image.

## Expected Behavior
We should use the node bundled within the executor docker image.

## Related Issue(s)

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,6 @@ version: 2.1
 # -------------------------
 orbs:
   nx: nrwl/nx@1.6.1
-  node: circleci/node@5.1.0
   rust: circleci/rust@1.6.0
   browser-tools: circleci/browser-tools@1.4.0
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -20,14 +19,14 @@ executors:
   linux:
     <<: *defaults
     docker:
-      - image: cimg/rust:1.66.1-browsers
+      - image: cimg/rust:1.70.0-browsers
```
meeroslav, line 22: Latest version has latest v18 Node

.circleci/config.yml
```diff
@@ -42,10 +41,22 @@ commands:
...
-          command: |
-            npm install -g @pnpm/exe@8.3.1
+      - when:
+          condition:
+            equal: [<< parameters.os >>, linux]
+          steps:
+            - run:
+                name: Install pnpm package manager (linux)
+                command: |
+                  npm install --prefix=$HOME/.local -g @pnpm/exe@8.3.1
```
meeroslav, line 51: Since we didn't install node in the $HOME folder, we can't install pnpm globally due to access rights on linux, so we need to specify target folder within our $HOME folder

FrozenPandaz (approved):

PR #17972, opened 2023-07-05
feat(nx-dev): changelog anchor links to each version

isaacplmann:
Updates the changelog page to add anchor links for each version.


juristr:
Small nit-pick, but should we maybe have the anchor tag besides the version number? And maybe even use that as the anchor? Because as a user that's what I want to tag and see in the URL
image

isaacplmann:
> Small nit-pick, but should we maybe have the anchor tag besides the version number? And maybe even use that as the anchor? Because as a user that's what I want to tag and see in the URL image

Yep, good call.

juristr (approved):
LGTM now. Thx for fixing

PR #17975, opened 2023-07-05
feat(testing): add @nx/playwright plugin

barbados-clemens:
## Current Behavior

no first party playwright plugin for nx.

## Expected Behavior

@nx/playwright plugin

## Related Issue(s)

- feat(testing): add playwright plugin
- feat(testing): add playwright executor
- chore(testing): add playwright e2e tests
- chore(testing): add legacy-playwright so plugin can be installed

Note: 
assumed generator setup
playwright.config.ts in the project root with testDir set from the project root, and reporter options set to output in the top level 'dist/playwright/<project-root-path>', only tested with html reporter.

I didn't look into what the eslint settings would look like nor tsconfig settings.

Might be able to take inspo from [nxkit playwright plugin](https://github.com/nxkit/nxkit/tree/main/packages/playwright/src/generators/project)

### DIFF
--- a/e2e/playwright/src/playwright.test.ts
+++ b/e2e/playwright/src/playwright.test.ts
@@ -1,0 +1,171 @@
+import {
+  cleanupProject,
+  newProject,
+  uniq,
+  createFile,
+  runCLI,
+  packageInstall,
+  runCommand,
+} from '@nx/e2e/utils';
+
+const TEN_MINS_MS = 600_000;
+describe('Playwright E2E Test runner', () => {
+  beforeAll(() => {
+    newProject({ name: uniq('playwright') });
+    packageInstall('@playwright/test', undefined, '^1.30.0', 'dev');
+    runCommand('npx playwright install --with-deps');
+  });
+
+  afterAll(() => cleanupProject());
+
+  it(
+    'should test example app',
+    () => {
+      //TODO: remove when generators are setup.Need to have basic workspace deps setup
+      runCLI(`g @nx/js:init`);
+      addSampleProject();
+
+      // NOTE: playwright throws errors if it detects running inside jest process. tmp remove and restore the env var for playwright to run
+      const results = runCLI(`e2e demo-e2e`);
+      expect(results).toContain('6 passed');
+      expect(results).toContain('Successfully ran target e2e for project');
+    },
+    TEN_MINS_MS
+  );
+});
+
+// TODO: remove this when there are project generators
+function addSampleProject() {
+  createFile(
+    'apps/demo-e2e/src/example.spec.ts',
+    `
+import { test, expect } from '@playwright/test';
+
+test('has title', async ({ page }) => {
+  await page.goto('https://playwright.dev/');
+
+  // Expect a title "to contain" a substring.
+  await expect(page).toHaveTitle(/Playwright/);
+});
+
+test('get started link', async ({ page }) => {
+  await page.goto('https://playwright.dev/');
+
+  // Click the get started link.
+  await page.getByRole('link', { name: 'Get started' }).click();
+
+  // Expects the URL to contain intro.
+  await expect(page).toHaveURL(/.*intro/);
+});
+`
+  );
+  createFile(
+    'apps/demo-e2e/playwright.config.ts',
+    `
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: './src',
+  outputDir: '../../dist/playwright/apps/demo-e2e/output',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: [
+    [
+      'html',
+      {
+        outputFolder:
+          '../../dist/playwright/apps/demo-e2e/playwright-report',
+      },
+    ],
+  ],
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like await page.goto('/'). */
+    // baseURL: 'http://127.0.0.1:3000',
+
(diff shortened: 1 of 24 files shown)


packages-legacy/playwright/project.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "playwright-legacy",
```
barbados-clemens, line 2: Probably we don't want to make a 'legacy' version of this plugin, but the current publish process adds this as a dep to the actual plugin. Just added to get working, but probably should update the publish process to exclude certain plugins?

packages/playwright/src/executors/playwright/playwright.ts
```diff
@@ -0,0 +1,179 @@
...
+  const projectRoot =
+    context.projectGraph?.nodes?.[context?.projectName]?.data?.root;
+
+  if (!projectRoot) {
+    throw new Error(
+      `Unable to find the Project Root for ${context.projectName}. Is it set in the project.json?`
+    );
+  }
+  const args = createArgs(options);
+  const p = runPlaywright(args, projectRoot);
```
barbados-clemens, line 17: holding off on adding a 'devServerTarget' feature since playwright does that natively with their webserver config options.

packages/playwright/src/executors/playwright/schema.json
```diff
@@ -0,0 +1,156 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "version": 2,
+  "title": "Playwright executor",
+  "description": "",
+  "type": "object",
+  "properties": {
```
barbados-clemens, line 7: unsure if we should add properties for different sub commands like 
show-report, codegen or make them separate executors?

I was thinking probably both

```shell
nx e2e demo-app --show-report # makes sense as an option here
```
```shell
nx codegen demo-app # makes sense as separate executor.  
```

Could be a follow up to this executor before the release.

mandarini (approved):

mandarini:
Nice!!!! :D

packages/playwright/src/executors/playwright/playwright.ts
```diff
@@ -0,0 +1,183 @@
...
+    // TODO(caleb): are there other signals I need to care about?
+    p.on('close', (code) => {
+      resolve({ success: code === 0 });
+    });
+  });
+}
+
+function createArgs(opts: PlaywrightExecutorSchema): string[] {
+  //NOTE: boolean options send at --arg=false might be invalid options.
+  const args: string[] = [];
```
xiongemi, line 29: i am just wondering whether this could be simplified, instead of listing all the options, could we loop through it?

packages/playwright/src/executors/playwright/playwright.ts
```diff
@@ -0,0 +1,183 @@
...
+    args.push(`--ui-port=${opts.uiPort}`);
+  }
+
+  return args;
+}
+
+function runPlaywright(args: string[], cwd: string) {
+  const pm = getPackageManagerCommand();
+  const cmd = [pm.exec, 'playwright', 'test', ...args];
+  return spawn(cmd.join(' '), {
```
xiongemi, line 126: maybe use fork instead of spawn?

packages/playwright/src/executors/playwright/schema.json
jaysoo: I think a separate target makes sense, so it's not confusing when users are used to running both `show-report` and `codegen` as separate commands. It removes any potential confusion from actual `playwright test` options.

I'm not sure if a new executor is needed, or can we just use `command` in the generator? like

```jsonc
{
  "codegen": {
    "command": "playwright codegen"
    "options": { "cwd": "<project-root>" }
  }
}
```

There are also commands that seem like they are unrelated to tests (e.g. `open, `cr`, `ff`, etc.). I thnk we can skip those.

packages-legacy/playwright/project.json
jaysoo: I think it's probably fine for now. We'll remove it in Nx 17.

packages/playwright/src/executors/playwright/playwright.ts
jaysoo: We had issues with `exec` and `spawn` not exiting the child process when the Nx process exits. Something to check here.

Like Emily mentioned, could use `fork` instead.

```
fork(require.resolve('playwright/cli'), [...]);
```

xiongemi (approved):

jaysoo (approved):

e2e/playwright/src/playwright.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    'should test example app',
+    () => {
+      //TODO: remove when generators are setup.Need to have basic workspace deps setup
+      runCLI(`g @nx/js:init`);
+      addSampleProject();
+
+      // NOTE: playwright throws errors if it detects running inside jest process. tmp remove and restore the env var for playwright to run
+      const id = process.env.JEST_WORKER_ID;
+      delete process.env.JEST_WORKER_ID;
+      const results = runCLI(`e2e demo-e2e`);
```
FrozenPandaz, line 31: Move this logic into `getStrippedEnvironmentVariables`

packages/playwright/project.json
```diff
@@ -0,0 +1,53 @@
...
+  "sourceRoot": "packages/playwright/src",
+  "projectType": "library",
+  "targets": {
+    "build": {
+      "executor": "nx:run-commands",
+      "outputs": ["{workspaceRoot}/build/packages/playwright"],
+      "options": {
+        "commands": [
+          "node ./scripts/copy-readme.js playwright",
+          "node ./scripts/add-dependency-to-build.js playwright @nrwl/playwright"
```
FrozenPandaz, line 13: Remove this.

packages-legacy/playwright/.eslintrc.json
```diff
@@ -0,0 +1,25 @@
+{
```
FrozenPandaz, line 1: We don't need this

packages/nx/package.json
```diff
@@ -123,6 +123,8 @@
       "@nrwl/next",
       "@nx/node",
       "@nrwl/node",
+      "@nx/playwright",
+      "@nrwl/playwright",
```
FrozenPandaz, line 127: Remove this legacy one.

packages/playwright/src/executors/playwright/playwright.ts
```diff
@@ -0,0 +1,113 @@
...
+  try {
+    const cli = require.resolve('@playwright/test/cli');
+
+    return fork(cli, ['test', ...args], {
+      stdio: 'inherit',
+      cwd,
+    });
+  } catch (e) {
+    console.error(e);
+    throw new Error('Unable to run playwright. is @playwright/test installed?');
```
FrozenPandaz, line 60:
```suggestion
    throw new Error('Unable to run playwright. Is @playwright/test installed?');
```

FrozenPandaz (changes requested):

AgentEnder (approved):

FrozenPandaz (approved):

PR #17991, opened 2023-07-06
feat(angular): add esbuild incremental build builder

Coly010:
## Current Behavior

We do not have a way to support incremental builds with EsBuild for Angular

## Expected Behavior

Support Incremental Builds with EsBuild for Angular

- Add's a new executor that handles the incremental build support
- Add's a util to allow us to call Angular builder functions from an executor

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/angular-core/src/projects.test.ts
+++ b/e2e/angular-core/src/projects.test.ts
@@ -106,11 +106,21 @@ describe('Angular Projects', () => {
     const appPort = 4207;
     const process = await runCommandUntil(
       `serve ${app1} -- --port=${appPort}`,
-      (output) => output.includes(`listening on localhost:4207`)
+      (output) => output.includes(`listening on localhost:${appPort}`)
     );
 
     // port and process cleanup
     await killProcessAndPorts(process.pid, appPort);
+
+    const esbProcess = await runCommandUntil(
+      `serve my-dir-${esbuildApp} -- --port=${appPort}`,
+      (output) =>
+        output.includes(`Application bundle generation complete`) &&
+        output.includes(`localhost:${appPort}`)
+    );
+
+    // port and process cleanup
+    await killProcessAndPorts(esbProcess.pid, appPort);
   }, 1000000);
 
   it('should lint correctly with eslint and handle external HTML files and inline templates', async () => {
@@ -164,6 +174,11 @@ describe('Angular Projects', () => {
 
   it('should build the dependent buildable lib and its child lib, as well as the app', async () => {
     // ARRANGE
+    const esbuildApp = uniq('esbuild-app');
+    runCLI(
+      `generate @nx/angular:app ${esbuildApp} --bundler=esbuild --no-interactive`
+    );
+
     const buildableLib = uniq('buildlib1');
     const buildableChildLib = uniq('buildlib2');
 
@@ -177,6 +192,27 @@ describe('Angular Projects', () => {
     // update the app module to include a ref to the buildable lib
     updateFile(
       `apps/${app1}/src/app/app.module.ts`,
+      `
+        import { BrowserModule } from '@angular/platform-browser';
+        import { NgModule } from '@angular/core';
+        import {${
+          names(buildableLib).className
+        }Module} from '@${proj}/${buildableLib}';
+
+        import { AppComponent } from './app.component';
+        import { NxWelcomeComponent } from './nx-welcome.component';
+
+        @NgModule({
+          declarations: [AppComponent, NxWelcomeComponent],
+          imports: [BrowserModule, ${names(buildableLib).className}Module],
+          providers: [],
+          bootstrap: [AppComponent],
+        })
+        export class AppModule {}
+    `
+    );
+    updateFile(
+      `apps/${esbuildApp}/src/app/app.module.ts`,
       `
         import { BrowserModule } from '@angular/platform-browser';
         import { NgModule } from '@angular/core';
@@ -224,9 +260,20 @@ describe('Angular Projects', () => {
       };
       return config;
     });
+    updateProjectConfig(esbuildApp, (config) => {
+      config.targets.build.executor = '@nx/angular:browser-esbuild';
+      config.targets.build.options = {
+        ...config.targets.build.options,
+        buildLibsFromSource: false,
+      };
+      return config;
+    });
 
     // ACT
     const libOutput = runCLI(`build ${app1} --configuration=development`);
+    const esbuildLibOutput = runCLI(
+      `build ${esbuildApp} --configuration=development`
+    );
 
     // ASSERT
     expect(libOutput).toContain(
@@ -238,6 +285,9 @@ describe('Angular Projects', () => {
     // the path to dist
     const mainBundle = readFile(`dist/apps/${app1}/main.js`);
     expect(mainBundle).toContain(`dist/libs/${buildableLib}`);
+
+    const mainEsBuildBundle = readFile(`dist/apps/${esbuildApp}/main.js`);
+    expect(mainEsBuildBundle).toContain(`dist/libs/${buildableLib}`);
   });
 
   it('should build publishable libs successfully', () => {
--- a/packages/angular/src/executors/browser-esbuild/browser-esbuild.impl.ts
+++ b/packages/angular/src/executors/browser-esbuild/browser-esbuild.impl.ts
@@ -1,0 +1,57 @@
+import { type EsBuildSchema } from './schema';
+import { validateOptions } from './lib/validate-options';
+import { type DependentBuildableProjectNode } from '@nx/js/src/utils/buildable-libs-utils';
+import { type ExecutorContext, readCachedProjectGraph } from '@nx/devkit';
+import { createTmpTsConfigForBuildableLibs } from './lib/buildable-libs';
+import { createBuilderContext } from 'nx/src/adapter/ngcli-adapter';
+import { type BuilderOutput } from '@angular-devkit/architect';
+import { type OutputFile } from 'esbuild';
+
+export default async function* esbuildExecutor(
+  options: EsBuildSchema,
+  context: ExecutorContext
+) {
+  validateOptions(options);
+  options.buildLibsFromSource ??= true;
+
+  const { buildLibsFromSource, ...delegateExecutorOptions } = options;
+
+  let dependencies: DependentBuildableProjectNode[];
+  let projectGraph = context.projectGraph;
+
+  if (!buildLibsFromSource) {
+    projectGraph = projectGraph ?? readCachedProjectGraph();
+    const { tsConfigPath, dependencies: foundDependencies } =
+      createTmpTsConfigForBuildableLibs(
+        delegateExecutorOptions.tsConfig,
+        context,
+        { projectGraph }
+      );
+    dependencies = foundDependencies;
+    delegateExecutorOptions.tsConfig = tsConfigPath;
+  }
+  const { buildEsbuildBrowser } = await import(
+    '@angular-devkit/build-angular/src/builders/browser-esbuild/index'
+  );
+
+  const builderContext = await createBuilderContext(
+    {
+      builderName: 'browser-esbuild',
+      description: 'Build a browser application',
+      optionSchema: await import(
+        '@angular-devkit/build-angular/src/builders/browser-esbuild/schema.json'
+      ),
+    },
+    context
+  );
+
+  return yield* buildEsbuildBrowser(
+    delegateExecutorOptions,
+    builderContext
+  ) as AsyncIterable<
+    BuilderOutput & {
+      outputFiles?: OutputFile[];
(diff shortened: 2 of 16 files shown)


packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -51,10 +53,73 @@ import {
...
+    },
+    builder: {
+      ...builderInfo,
+    },
+    logger: getLogger(),
+    id: 1,
+    currentDirectory: process.cwd(),
+    scheduleTarget: architect.scheduleTarget,
+    addTeardown(teardown: () => Promise<void> | void) {},
+    validateOptions: architect['_scheduler'].validateOptions,
```
leosvelperez, line 107: I couldn't find `validateOptions` in the `SimpleScheduler` which is the type of the `_scheduler` (https://github.com/angular/angular-cli/blob/333da08ea29e466dd883dfd4742088deb73b25d0/packages/angular_devkit/architect/src/architect.ts#L382). The `validateOptions` in the builder context is implemented here https://github.com/angular/angular-cli/blob/333da08ea29e466dd883dfd4742088deb73b25d0/packages/angular_devkit/architect/src/create-builder.ts#L166-L176. It schedules a job named `..validateOptions`.

Unless I'm missing something, we're effectively setting it to `undefined` which is probably not what we want. The `validateOptions` in the context is used internally by some builders to validate the options of other targets they execute (e.g. builders with an option like `browserTarget`).

From what I saw, this is not an issue with the current `browser-esbuild` builder because it's not using that function, but this function is a generic utility and can be used with other builders.

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -51,10 +53,73 @@ import {
...
+    getBuilderNameForTarget: architectHost.getBuilderNameForTarget,
+    scheduleBuilder: architect.scheduleBuilder,
+    getTargetOptions: architectHost.getOptionsForTarget,
+    reportProgress: architect['_scheduler'].reportProgress,
+    reportRunning: architect['_scheduler'].reportRunning,
```
leosvelperez, line 112: Similar to the previous comment, the scheduler does not define these two functions. Unlike the previous comment, we could provide noop functions for them because we don't use them (similar to `reportStatus` below).

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -15,13 +16,14 @@ import { FileBuffer } from '@angular-devkit/core/src/virtual-fs/host/interface';
 // Importing @angular-devkit/architect here will cause issues importing this file without @angular-devkit/architect installed
 /* eslint-disable no-restricted-imports */
 import type { Architect } from '@angular-devkit/architect';
+import { Target } from '@angular-devkit/architect';
```
leosvelperez, line 19: Non-type imports can bring back the issues that were addressed in https://github.com/nrwl/nx/pull/16595.
```suggestion
import type { Target } from '@angular-devkit/architect';
```

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -51,10 +53,73 @@ import {
   CustomHasher,
   Executor,
   ExecutorConfig,
+  ExecutorContext,
   ExecutorsJson,
   TaskGraphExecutor,
 } from '../config/misc-interfaces';
 import { readPluginPackageJson } from '../utils/nx-plugin';
+import { WorkspaceNodeModulesArchitectHost as AngularWorkspaceNodeModulesArchitectHost } from '@angular-devkit/architect/node';
```
leosvelperez, line 61: This top-level import can reintroduce the issues addressed in https://github.com/nrwl/nx/pull/16595. You can wrap the `WrappedWorkspaceNodeModulesArchitectHost` class declaration in a function, so you make this import inside that function and have the consumers invoke it.

leosvelperez (changes requested):

packages/angular/src/executors/browser-esbuild/browser-esbuild.impl.ts
```diff
@@ -0,0 +1,82 @@
...
+  const { buildEsbuildBrowser } = await import(
+    '@angular-devkit/build-angular/src/builders/browser-esbuild/index'
+  );
+
+  const builderContext = await createBuilderContext(
+    {
+      builderName: 'browser-esbuild',
+      description: 'Build a browser application',
+      optionSchema: await import(
+        '@angular-devkit/build-angular/src/builders/browser-esbuild/schema.json'
```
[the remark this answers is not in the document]
Coly010: The implementation function would be the `createBuilder` wrapper around the `buildEsbuildBrowser` function would it not?

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -15,13 +16,14 @@ import { FileBuffer } from '@angular-devkit/core/src/virtual-fs/host/interface';
...
 import type { Architect } from '@angular-devkit/architect';
+import { Target } from '@angular-devkit/architect';
 import type { NodeModulesBuilderInfo } from '@angular-devkit/architect/node/node-modules-architect-host';
 
 import * as chalk from 'chalk';
 import { Stats } from 'fs';
 import { dirname, extname, join, resolve } from 'path';
 
-import { concat, from, Observable, of, zip } from 'rxjs';
+import { concat, firstValueFrom, from, Observable, of, zip } from 'rxjs';
```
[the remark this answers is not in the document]
Coly010: Ah interesting, I guess Angular CLI does not need to worry about this on their side.

We can branch the logic and use either `firstValueFrom` or `toPromise`

packages/nx/src/adapter/ngcli-adapter.ts
Coly010: Something to look into, but as it is not used by esbuild browser, it could be something that is addressed in a follow up PR.

packages/nx/src/adapter/ngcli-adapter.ts
Coly010: no problem

packages/nx/src/adapter/ngcli-adapter.ts
leosvelperez: Even if not used, providing here something that it just doesn't exist is wrong. If we don't plan to support it now and add the support later, we should add least be explicit about it adding a noop function with a TODO note.

packages/angular/src/executors/browser-esbuild/browser-esbuild.impl.ts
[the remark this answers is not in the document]
leosvelperez: Yeah, good call. We need the actual function, so I guess this is not really feasible.

packages/angular/src/executors/browser-esbuild/browser-esbuild.impl.ts
```diff
@@ -0,0 +1,82 @@
...
+      } else if (!buildRes.value.success) {
+        next({ success: buildRes.value.success });
+        break;
+      } else {
+        next({ success: buildRes.value.success });
+      }
+      buildRes = await iter.next();
+    }
+    done();
+  });
```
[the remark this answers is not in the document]
Coly010: Tried this again and this is needed because of how Angular creates the Async Iterable. We need to loop through until it completes. The same pattern is used in other packages such as the Module Federation Dev Server in the React Package

e2e/angular-core/src/projects.test.ts
```diff
@@ -111,6 +111,14 @@ describe('Angular Projects', () => {
 
     // port and process cleanup
     await killProcessAndPorts(process.pid, appPort);
+
+    const esbProcess = await runCommandUntil(
+      `serve ${esbuildApp} -- --port=${appPort}`,
+      (output) => output.includes(`listening on localhost:4207`)
```
leosvelperez, line 117:
```suggestion
      (output) => output.includes(`listening on localhost:${appPort}`)
```

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -14,14 +15,14 @@ import { FileBuffer } from '@angular-devkit/core/src/virtual-fs/host/interface';
...
-import type { Architect } from '@angular-devkit/architect';
+import type { Architect, Target } from '@angular-devkit/architect';
 import type { NodeModulesBuilderInfo } from '@angular-devkit/architect/node/node-modules-architect-host';
 
 import * as chalk from 'chalk';
 import { Stats } from 'fs';
 import { dirname, extname, join, resolve } from 'path';
 
-import { concat, from, Observable, of, zip } from 'rxjs';
+import { concat, firstValueFrom, from, Observable, of, zip } from 'rxjs';
```
leosvelperez, line 25:
```suggestion
import { concat, from, Observable, of, zip } from 'rxjs';
```

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -85,124 +191,7 @@ export async function scheduleTarget(
   const AngularWorkspaceNodeModulesArchitectHost =
     require('@angular-devkit/architect/node').WorkspaceNodeModulesArchitectHost;
```
leosvelperez, line 192: We can remove this. It's not used anymore.

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -51,10 +52,115 @@ import {
...
+          return firstValueFrom(
+            architect['_scheduler'].schedule('..getProjectMetadata', target)
+              .output
+          );
+        }
+      : (target: Target | string) => {
+          return architect['_scheduler']
+            .schedule('..getProjectMetadata', target)
+            .output.toPromise();
+        };
```
leosvelperez, line 123: We could simplify this and wouldn't need the `getInstalledPackageVersionInfo` function at the bottom of the file or the import from `semver`:
```suggestion
  const { firstValueFrom } = require('rxjs');
  const toPromise = (obs: Observable<any>) =>
    firstValueFrom ? firstValueFrom(obs) : obs.toPromise();

  const validateOptions = (options: json.JsonObject, builderName: string) =>
    toPromise(
      architect['_scheduler'].schedule('..validateOptions', [
        builderName,
        options,
      ]).output
    );

  const getProjectMetadata = (target: Target | string) =>
    toPromise(
      architect['_scheduler'].schedule('..getProjectMetadata', target).output
    );
```

leosvelperez (changes requested):

e2e/angular-core/src/projects.test.ts
Coly010: ah i think there's another line in the file that could benefit from this change, i had just copied that line. this is better

packages/nx/src/adapter/ngcli-adapter.ts
Coly010: ah good catch.

packages/nx/src/adapter/ngcli-adapter.ts
Coly010: good spot

packages/nx/src/adapter/ngcli-adapter.ts
Coly010: I like it

packages/nx/src/adapter/ngcli-adapter.ts
leosvelperez: You missed removing this one.

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -1049,3 +1021,146 @@ function saveProjectsConfigurationsInWrappedSchematic(
...
+        implementation,
+        executorsDir
+      );
+    }
+  }
+
+  return new WrappedWorkspaceNodeModulesArchitectHost(workspace, root);
+}
+
+function getInstalledPackageVersionInfo(
```
leosvelperez, line 1154: This function is not needed/used. You can remove it.

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -51,10 +52,98 @@ import {
   CustomHasher,
   Executor,
   ExecutorConfig,
+  ExecutorContext,
   ExecutorsJson,
   TaskGraphExecutor,
 } from '../config/misc-interfaces';
 import { readPluginPackageJson } from '../utils/nx-plugin';
+import { major } from 'semver';
```
leosvelperez, line 60:
```suggestion
```

leosvelperez (changes requested):

packages/nx/src/adapter/ngcli-adapter.ts
Coly010: 🤦 my bad!

leosvelperez (approved):

jaysoo (approved):

juristr (approved):

PR #18009, opened 2023-07-06
fix(core): parse project configs only in js

FrozenPandaz:
## Current Behavior

Nx parses config files in Rust.. and then reparses it in JS and sometimes assigns the project a different name.

## Expected Behavior

Nx only identifies project config files in rust.. but only parses the file in JS and always assigns the project the same name. The project file map is calculated in JS from the parsed project configs.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/tests/workspace_files.spec.ts
+++ b/packages/nx/src/native/tests/workspace_files.spec.ts
@@ -1,12 +1,24 @@
-import {
-  getConfigFiles,
-  getWorkspaceFilesNative,
-  WorkspaceErrors,
-} from '../index';
+import { getProjectConfigurations, getWorkspaceFilesNative } from '../index';
 import { TempFs } from '../../utils/testing/temp-fs';
 import { NxJsonConfiguration } from '../../config/nx-json';
+import { dirname, join } from 'path';
+import { readJsonFile } from '../../utils/fileutils';
 
 describe('workspace files', () => {
+  function createParseConfigurationsFunction(tempDir: string) {
+    return (filenames: string[]) => {
+      const res = {};
+      for (const filename of filenames) {
+        const json = readJsonFile(join(tempDir, filename));
+        res[json.name] = {
+          ...json,
+          root: dirname(filename),
+        };
+      }
+      return res;
+    };
+  }
+
   it('should gather workspace file information', async () => {
     const fs = new TempFs('workspace-files');
     const nxJson: NxJsonConfiguration = {};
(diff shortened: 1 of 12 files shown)


packages/nx/src/native/workspace/get_nx_workspace_files.rs
```diff
@@ -35,9 +34,10 @@ pub fn get_workspace_files_native(
     let (projects, mut file_data) = get_file_data(&workspace_root, globs)
         .map_err(|err| napi::Error::new(WorkspaceErrors::Generic, err.to_string()))?;
 
-    let root_map = create_root_map(&projects)?;
+    dbg!(&projects);
```
Cammisuli, line 37: Remove or use `trace!`

packages/nx/src/native/workspace/get_nx_workspace_files.rs
```diff
@@ -35,9 +34,10 @@ pub fn get_workspace_files_native(
...
+    let root_set = create_root_set(&projects)?;
 
-    trace!(?root_map);
+    dbg!(&root_set);
```
Cammisuli, line 40: use `trace!`

Cammisuli (changes requested):

packages/nx/src/native/tests/workspace_files.spec.ts
```diff
@@ -152,30 +148,7 @@ describe('workspace files', () => {
...
-          "file": "project.json",
-          "hash": "4357927788053707201",
-        },
-        {
-          "file": "src/index.js",
-          "hash": "3244421341483603138",
-        },
-      ]
-    `);
+    expect(projectFileMap['repo-name']).toMatchInlineSnapshot(`undefined`);
```
Cammisuli, line 151: This should be `projectFileMap['.']`. The undefined in the snapshot isn't right 😅

packages/nx/src/native/tests/workspace_files.spec.ts
```diff
@@ -243,7 +216,7 @@ describe('workspace files', () => {
       expect(getWorkspaceFilesNative(fs.tempDir, globs).projectFileMap)
         .toMatchInlineSnapshot(`
         {
-          "project1": [
+          "libs/project1": [
```
Cammisuli, line 219: We probably dont need this whole `describe('errors')` block anymore. We're not doing any parsing now.

Cammisuli (changes requested):

Cammisuli (approved):

PR #18020, opened 2023-07-07
fix(core): do not cache external dependency hashes if they are not fully resolved

### LINKED ISSUE #17917: Cache Invalidation Issue in CI/CD Pipeline (Jenkins) - Inconsistent Hash Value

skrtheboss:
## Current Behavior

External dependency hashes are not deterministic and could cause cache misses, since the hash is dependent on which task was hashed when.

## Expected Behavior

Task hashes should not depend on which task have been hashed earlier.

## Related Issue(s)

Fixes #17917

### DIFF
--- a/packages/nx/src/hasher/task-hasher.spec.ts
+++ b/packages/nx/src/hasher/task-hasher.spec.ts
@@ -1161,7 +1161,7 @@ describe('TaskHasher', () => {
       expect(hash.value).toContain('|5.0.0|');
     });
 
-    it('should hash entire subtree of dependencies', async () => {
+    it('should hash entire subtree in a deterministic way', async () => {
       const createHasher = () =>
         new InProcessTaskHasher(
           {},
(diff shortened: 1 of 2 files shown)


skrtheboss:
@meeroslav and @FrozenPandaz, the attempt to fix the external dependency hashing in https://github.com/nrwl/nx/pull/17926 did not work. I am still experiencing a lot of cache misses.

So, I have debugged this further and enhanced the test so that it fails. One problem is that the hash should not be stored if the full chain could not be hashed; otherwise, the cache can contain broken hashes. Additionally, the parents/visited set must include the full parent/visited list; otherwise, an endless recursion can occur.

skrtheboss:
Hi @FrozenPandaz and @meeroslav, have had a chance to take a look at this? I think it should really have a higher priority, since this is a tool that focusses on caching performance, and should hash things properly, without performing computations multiple times where not needed..

Thank you.

atsjo:
16.1.4 was the last versjon where caching worked reliably for me... The patch provided in 16.5.1 did not fix my caching problems either...

If I re-run builds without closing vs-code in between, the caching works fine, but if I close vs-code and re-open, it re-builds already built projects sometimes. Seems un-deterministic to me, but maybe related to which project I build...

FrozenPandaz (approved):

atsjo:
thanx @skrtheboss! this finally fixed the caching issues I have experienced since 16.2.0

skrtheboss:
@atsjo nice to hear and thanks @FrozenPandaz for mergin this.

PR #18024, opened 2023-07-07
docs(core): improve visibility of CNW command on intro page

juristr:
Improves visibility of the `create-nx-workspace` and `nx init` commands  on the getting started page

[/getting-started/intro](https://nx-dev-git-fork-juristr-docs-make-cnw-more-prominent-nrwl.vercel.app/getting-started/intro)

image

image

### DIFF
--- a/docs/shared/getting-started/intro.md
+++ b/docs/shared/getting-started/intro.md
@@ -1,6 +1,26 @@
 # Intro to Nx
 
-Nx is a powerful open-source build system that provides tools and techniques for enhancing developer productivity, optimizing CI performance, and maintaining code quality. Learn more about [how Nx works](/getting-started/why-nx).
+Nx is a powerful open-source build system that provides tools and techniques for enhancing developer productivity, optimizing CI performance, and maintaining code quality. Find out more about [why you should use Nx](/getting-started/why-nx).
+
+If instead you want to jump right into it, run the following command. It will guide you through the setup:
+
+{% tabs %}
+{% tab label="npm" %}
+
+{% terminal-command command="npx create-nx-workspace"  /%}
+
+{% /tab %}
+{% tab label="yarn" %}
+
+{% terminal-command command="npx create-nx-workspace --pm yarn"  /%}
+
+{% /tab %}
+{% tab label="pnpm" %}
+
+{% terminal-command command="npx create-nx-workspace --pm pnpm"  /%}
+
+{% /tab %}
+{% /tabs %}
 
 You can use Nx to quickly scaffold a new standalone project or even an entire monorepo. It can be incrementally adopted and will grow with your project as it scales.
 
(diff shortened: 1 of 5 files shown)


docs/shared/getting-started/intro.md
```diff
@@ -1,6 +1,32 @@
...
+Nx is a powerful open-source build system that provides tools and techniques for enhancing developer productivity, optimizing CI performance, and maintaining code quality. If you want to learn more about what Nx is, how it works and why you might want to use it, check out our ["Why Nx"](/getting-started/intro) page.
+
+If instead you want to jump right into it, run the following command. It will guide you through the setup:
+
+{% tabs %}
+{% tab label="npm" %}
+
+```{% command="npm create nx-workspace" path="~" %}
+
+```
```
isaacplmann, line 12:
```suggestion
```shell
npm create nx-workspace
```
```

File: docs/shared/getting-started/intro.md [hunk 2]
```diff
@@ -1,6 +1,32 @@
...
-Nx is a powerful open-source build system that provides tools and techniques for enhancing developer productivity, optimizing CI performance, and maintaining code quality. Learn more about [how Nx works](/getting-started/why-nx).
+Nx is a powerful open-source build system that provides tools and techniques for enhancing developer productivity, optimizing CI performance, and maintaining code quality. If you want to learn more about what Nx is, how it works and why you might want to use it, check out our ["Why Nx"](/getting-started/intro) page.
+
+If instead you want to jump right into it, run the following command. It will guide you through the setup:
+
+{% tabs %}
+{% tab label="npm" %}
+
+```{% command="npm create nx-workspace" path="~" %}
+
```
isaacplmann, line 11:
```suggestion
```shell
npm create nx-workspace
```

isaacplmann: This looks better if you aren't going to show the output of the command

juristr: I actually on purpose used the command because I felt it stands more out than just the normal box. Like one immediately sees this is a CLI command vs a code block etc. Otherwise everything looks the same.

We could adjust the style though by creating a custom `terminal` fence that does the same

juristr: Like I personally find the 2nd better

image

image

But we could check in the component whether there's an output and in that case reduce the padding below the command. + maybe allow writing it like

```
```terminal
npx create-nx-workspace ...
```

isaacplmann: I'm fine with the fake terminal window if you remove the padding on the bottom

juristr: What do you think about just having a single terminal that uses NPM and ignoring the yarn / pnpm? We'd have that in the installation page anyway. It'd make it more concise and less loaded also

meeroslav (approved):
Looks good to me.
I would leave the three tabs even on intro. It doesn't take too much space and it shows that we support all three package managers equally.

PR #18051, opened 2023-07-10
feat(storybook): interaction testing init generator

mandarini:
Initial generator for Storybook Interaction Testing:

* Replace `configureTestRunner` option with `interactionTests` option.
* Add `configureTestRunner` as alias to `interactionTests` to keep existing functionality
* Install all the dependencies needed for interaction tests and the test runner configuration
* add the interaction addon on .storybook/main.ts
* Configure the test-runner target
* Add `deprecated` to Cypress Specs


Coly010 (approved):

packages/storybook/src/generators/configuration/schema.json
```diff
@@ -18,15 +18,22 @@
       "x-dropdown": "projects",
       "x-priority": "important"
     },
+    "interactionTests": {
+      "type": "boolean",
+      "description": "Set up Storybook interaction tests.",
+      "x-prompt": "Do you want to set up Storybook Interaction Tests?",
```
jaysoo, line 24: Are we calling it `Interaction Test` or `Interaction Testing`? If it's used as a noun (implied by the capitalized words) we should be consistent. I think it's better to say `Storybook interaction tests`.

packages/storybook/src/generators/configuration/schema.json
```diff
@@ -18,15 +18,22 @@
...
+      "alias": ["configureTestRunner"],
+      "default": true
+    },
     "configureCypress": {
       "type": "boolean",
       "description": "Run the cypress-configure generator.",
-      "x-prompt": "Configure a cypress e2e app to run against the storybook instance?",
-      "x-priority": "important"
+      "x-deprecated": "Please use Storybook Interaction Testing instead."
```
jaysoo, line 31: Same as above: `Please use  Storybook interaction tests instead`

packages/storybook/src/generators/configuration/lib/util-functions.ts
```diff
@@ -497,6 +497,7 @@ export function createProjectStorybookDir(
   root: string,
   projectType: string,
   projectIsRootProjectInStandaloneWorkspace: boolean,
+  interactionTesting: boolean,
```
jaysoo, line 500: Minor: for consistency I'd rename it to `interactionTests`. Also, maybe for future refactoring, some of these might be better to group in an options object rather than positional args. For example, `{ useVite: true, isNextjs: true }`, which removes the need to know which position is what and makes optional fields easier to handle.

jaysoo (approved):

jaysoo:
Approved. Added some suggestions.

mandarini:
Thanks! I addressed the comments!

PR #18057, opened 2023-07-10
feat(core): throw an error when unknown local cache source

vsavkin:
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/guides/unknown-local-cache.md
+++ b/docs/shared/guides/unknown-local-cache.md
@@ -1,0 +1,86 @@
+# Unknown Local Cache Error
+
+This document will explain why the following error happens and how to address it.
+
+```
+NX Invalid Cache Directory for Task "myapp:build"
+
+The local cache artifact in "node_modules/.cache/nx/786524780459028195" was not been generated on this machine.
+As a result, the cache's content integrity cannot be confirmed, which may make cache restoration potentially unsafe.
+If your machine ID has changed since the artifact was cached, run "nx reset" to fix this issue.
+Read about the error and how to address it here: https://nx.dev/recipes/other/unknown-local-cache
+```
+
+## Nx Tracks Cache Source
+
+Nx can cache tasks, which can drastically speed up your CI and local builds. However, this comes with the potential risk
+of "cache poisoning", where cache artifacts could be intentionally or inadvertently overwritten. If another user
+executes a task that matches the hash of the tainted artifact, they could retrieve the corrupted artifact and use it as
+the outcome of the task. Nx and Nx Cloud contain several safeguards to minimize the likelihood of cache poisoning or, in
+the case of Nx Cloud, completely prevent it.
+
+The error above is one such safeguard.
+
+Nx trusts the local cache. If you executed a task and stored the corresponding cached artifact on your machine, you can
+safely restore it on the same machine without worrying about cache poisoning. After all, in order to tamper with the
+cache artifact, the actor would need access to the machine itself.
+
+However, when artifacts in the local cache are created by a different machine, we cannot make such assumption. By
+default, Nx will refuse to use such artifacts and will throw the "Invalid Cache Directory" error.
+
+## Your MachineId Has Changed
+
+Upgrading your computer's hardware may alter its Machine ID, yielding the error above. To fix it execute `nx reset` to
+remove all the cache directories created under the previous Machine ID. After doing so, you should no longer see the
+error. After doing so, you should no longer see the error.
+
+## You Share Cache with Another Machine Using a Network Drive
+
+You can prefix any Nx command with `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` to ignore the error (
+e.g., `NX REJECT_UNKNOWN_LOCAL_CACHE=0 nx run-many -t build test`). This is similar to
+setting `NODE_TLS_REJECT_UNAUTHORIZED=0` to ignore any errors stemming form self-signed certificates. Even though it
+will make it work, this approach is discouraged.
+
+Storing Nx's local cache on a network drive can present security risks. When a network drive is shared, every CI run has
+access to all the previously created Nx cache artifacts. Hence, it is plausible for every single artifact - for every
+single task hash - to be accessed without leaving any trace. This is feasible due to the network drive's capability to
+allow overwrites.
+
+Instead of sharing the network drive, we highly recommend you to implement the `RemoteCache` interface.
+
+## Implementing Remote Cache Interface
+
+This is the interface:
+
+```typescript
+interface RemoteCache {
+  retrieve(hash: string, cachePath: string);
+
+  store(hash: string, cachePath: string);
+}
+```
+
+> You will need to wrap the default tasks runner to provide the remote cache implementation.
+
+## How Nx Cloud Makes Sure Sharing Cache is Safe
+
+The Nx Cloud runner provides an implementation of `RemoteCache` which does the following things making sharing the cache safe:
+
+1. **Immutable Artifacts:** Nx Cloud allows you to create and store new artifacts without the ability to override the
+   existing ones. This prevents any possibility of poisoning an existing artifact. This is achieved by managing the
+   cache using short-lived signed URLs.
+
+2. **Artifact Accessibility:** Nx Cloud provides access to the cache artifact specifically for the task that is
+   currently being executed. It restricts the ability to list all cache artifacts.
+
+3. **Visibility Control:** Nx Cloud comes with options to manage the visibility of your cache artifacts. For instance,
+   the cache artifacts created in `main` might be accessible by anyone across any branch, whereas the artifacts created
+   in your PR could be shared only within your PR runs.
+
+4. **Access Token Traceability:** Nx Cloud keeps a record of the access token used to create a cache artifact. In case
+   an access token gets compromised it can be easily removed, in turn deleting all the cache artifacts that were created
+   using it.
+
+Nx Cloud is not the only remote cache you can use. If you are using a different remote cache or using your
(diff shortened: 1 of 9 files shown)


docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+Cloud contain several safeguards to minimize the likelihood of cache poisoning or, in the case of Nx Cloud, completely
+prevent it.
+
+The error above is one such safeguard.
+
+Nx trusts the local cache. If you executed a task and stored the corresponding cached artifact on your machine, you can
+safely restore it on the same machine without worrying about cache poisoning. The "Invalid Cache Directory" error is
+thrown when the local cache artifact was created on another machine.
+
+## You MachineId Changed
```
FrozenPandaz, line 28:
```suggestion
## Your Machine ID has Changed
```

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+
+Upgrading your computer's hardware may alter its machineId, yielding the error above. To fix it execute "nx reset" to
```
FrozenPandaz, line 30:
```suggestion
Upgrading your computer's hardware may alter its Machine ID, yielding the error above. To fix it execute `nx reset` to
```

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+remove all "invalid" cache directories. After doing so, you should no longer see the error.
```
FrozenPandaz, line 31:
```suggestion
remove all of the cache directories created under the previous Machine ID. After doing so, you should no longer see the error.
```

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+of "cache poisoning", where cache artifacts could be intentionally or inadvertently overwritten. If another user
+executes a task that matches the hash of the tainted artifact, they could retrieve the corrupted artifact. Nx and Nx
+Cloud contain several safeguards to minimize the likelihood of cache poisoning or, in the case of Nx Cloud, completely
+prevent it.
+
+The error above is one such safeguard.
+
+Nx trusts the local cache. If you executed a task and stored the corresponding cached artifact on your machine, you can
+safely restore it on the same machine without worrying about cache poisoning. The "Invalid Cache Directory" error is
+thrown when the local cache artifact was created on another machine.
```
FrozenPandaz, line 26:
```suggestion
of "cache poisoning", where cache artifacts could be intentionally or inadvertently overwritten. When Nx
executes a task that matches the hash of the tainted artifact, it would retrieve the corrupted artifact and use it as the outcome of the task.

The above error is a safeguard to mitigate the unintentionally usage of a cache artifact from the local cache which may be unsafe.

Nx expects that artifacts in local cache on a machine are created by the machine itself. When this is the case, in order to tamper with the cache artifact, the actor would need access to the machine itself. However, when artifacts in the local cache are created by a different machine, there is reason for concern. By default, Nx will refuse to use such artifacts as they may have been poisoned. The "Invalid Cache Directory" error is
thrown when the local cache artifact was created on another machine.
```

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+remove all "invalid" cache directories. After doing so, you should no longer see the error.
+
+## You Share Cache with Another Machine Using a Network Drive
+
+You can prefix any Nx command with `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` to ignore the error (
+e.g., `NX REJECT_UNKNOWN_LOCAL_CACHE=0 nx run-many -t build test`). This is similar to
+setting `NODE_TLS_REJECT_UNAUTHORIZED=0` to ignore any errors stemming form self-signed certificates. Even though it
+will make it work, this approach is discouraged.
+
+Sharing a network drive can present security risks. When a network drive is shared, every CI run has access to all the
```
FrozenPandaz, line 40:
```suggestion
When Nx's local cache is stored on a network drive, it violates the expectation that the artifacts there created by the same machine. As such, sharing a network drive introduces security vulnerabilities. When a network drive is shared, every CI run has full access to all the
```

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+
+Upgrading your computer's hardware may alter its machineId, yielding the error above. To fix it execute "nx reset" to
+remove all "invalid" cache directories. After doing so, you should no longer see the error.
+
+## You Share Cache with Another Machine Using a Network Drive
+
+You can prefix any Nx command with `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` to ignore the error (
+e.g., `NX REJECT_UNKNOWN_LOCAL_CACHE=0 nx run-many -t build test`). This is similar to
+setting `NODE_TLS_REJECT_UNAUTHORIZED=0` to ignore any errors stemming form self-signed certificates. Even though it
+will make it work, this approach is discouraged.
```
FrozenPandaz, line 38: I think this needs to be moved after the issue is explained.

```suggestion
You can prefix any Nx command with `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` to ignore the error (
e.g., `NX REJECT_UNKNOWN_LOCAL_CACHE=0 nx run-many -t build test`). This is similar to
NodeJS's `NODE_TLS_REJECT_UNAUTHORIZED=0` setting which ignores any errors stemming form self-signed certificates. Even though this setting
will make it work, using it is discouraged because it allows for cache poisoning.
```

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+
+```typescript
+interface RemoteCache {
+  retrieve(hash: string, cachePath: string);
+
+  store(hash: string, cachePath: string);
+}
+```
+
+You will need to wrap the default tasks runner to provide the remote cache implementation.
```
FrozenPandaz, line 58: I would move this to the end. We should first recommend Nx Cloud before recommending this as a solution.

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+
+## How Nx Cloud Makes Sure Sharing Cache is Safe
+
+The Nx Cloud runner provides an implementation of `RemoteCache` which does the following things:
```
FrozenPandaz, line 62: If having multiple machines write to the same cache is unsafe, you may now have some reasonable suspicion that Nx Cloud is unsafe as well. However, Nx Cloud has several safeguards to prevent cache poisoning.

docs/shared/guides/unknown-local-cache.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. **Immutable Artifacts:** Nx Cloud allows you to create and store new artifacts without the ability to override the existing ones. This prevents any possibility of poisoning an existing artifact. This is achieved by managing the cache using short-lived signed URLs.
+
+2. **Artifact Accessibility:** Nx Cloud provides access to the cache artifact specifically for the task that is currently being executed. It restricts the ability to list all cache artifacts.
+
+3. **Visibility Control:** Nx Cloud comes with options to manage the visibility of your cache artifacts. For instance, the cache artifacts created in `main` might be accessible by anyone across any branch, whereas the artifacts created in your PR could be shared only within your PR runs.
+
+4. **Access Token Traceability:** Nx Cloud keeps a record of the access token used to create a cache artifact. In case an access token gets compromised it can be easily removed, in turn deleting all the cache artifacts that were created using it.
+
+If you are implementing the remote cache yourself, consider implementing the same safety mechanisms as Nx Cloud.
```
FrozenPandaz, line 72:
```suggestion
Nx Cloud is by all means, not the only remote cache you can use. If you are using a different remote cache or using your own implementation, we would highly recommend ensuring that the same safety mechanisms as Nx Cloud have been put in place.
```

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -207,6 +225,32 @@ export class Cache {
...
+        sourceMachineId &&
+        sourceMachineId != (await this.currentMachineId())
+      ) {
+        if (
+          process.env.NX_REJECT_UNKNOWN_LOCAL_CACHE != '0' &&
+          process.env.NX_REJECT_UNKNOWN_LOCAL_CACHE != 'false'
+        ) {
+          const error = [
+            `Invalid Cache Directory for Task "${task.id}"`,
+            `The local cache artifact in "${td}" has not been generated on this machine.`,
```
FrozenPandaz, line 244:
```suggestion
            `The local cache artifact in "${td}" was not been generated on this machine.`,
```

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -207,6 +225,32 @@ export class Cache {
...
+        sourceMachineId = await readFile(join(td, 'source'), 'utf-8');
+      } catch {}
+
+      if (
+        sourceMachineId &&
+        sourceMachineId != (await this.currentMachineId())
+      ) {
+        if (
+          process.env.NX_REJECT_UNKNOWN_LOCAL_CACHE != '0' &&
+          process.env.NX_REJECT_UNKNOWN_LOCAL_CACHE != 'false'
```
FrozenPandaz, line 240: I thought we were going to warn even when it is set. Did you change your mind?

FrozenPandaz (changes requested):

packages/nx/src/tasks-runner/cache.ts
vsavkin: Yeah I changed it for now. I can always add it later.

bcabanes (approved):

PR #18058, opened 2023-07-10
fix(linter): skip verifying deps of deps by default in package.json

jaysoo:
This PR fixes an issue with the new `@nx/dependency-checks` where dependencies of dependencies are required to be in the project's `package.json` file.

For example, if `a` depends on `b`, and `b` has a dependency on `npm:lodash`, then the lint rule will require `lodash` to be in `a`'s `package.json` file. This is almost always wrong, unless the user specifically is bundling in external dependencies.

The lint rule after this PR will no longer run in recursive mode, so only dependencies of the project being checked will be used. If the user wants to enable recursive mode they can pass `{ includeTransitiveDependencies: true }` in their eslintrc file.

I tested the new changes on the Nx monorepo, and it correctly found three issues with the Angular plugin.

```
/Users/jack/projects/nx/packages/angular/package.json
  51:5  error  The "chokidar" package is not used by "angular"                    @nx/dependency-checks
  52:5  error  The "http-server" package is not used by "angular"                 @nx/dependency-checks

✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```

## Current Behavior

Lint will fail even though `package.json` is correct for lib.

## Expected Behavior
Lint rule should work as expected.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -1,18 +1,22 @@
+import { join } from 'path';
+import { satisfies } from 'semver';
 import { AST } from 'jsonc-eslint-parser';
-import { normalizePath, workspaceRoot } from '@nx/devkit';
+import { type JSONLiteral } from 'jsonc-eslint-parser/lib/parser/ast';
+import {
+  normalizePath,
+  ProjectGraphProjectNode,
+  FileData,
+  workspaceRoot,
+} from '@nx/devkit';
+import { findNpmDependencies } from '@nx/js/src/utils/find-npm-dependencies';
+
 import { createESLintRule } from '../utils/create-eslint-rule';
 import { readProjectGraph } from '../utils/project-graph-utils';
 import { findProject, getSourceFilePath } from '../utils/runtime-lint-utils';
-import { join } from 'path';
-import { findProjectsNpmDependencies } from '@nx/js/src/internal';
-import { satisfies } from 'semver';
-import { getHelperDependenciesFromProjectGraph } from '@nx/js';
 import {
   getAllDependencies,
   getPackageJson,
-  removePackageJsonFromFileMap,
 } from '../utils/package-json-utils';
-import { JSONLiteral } from 'jsonc-eslint-parser/lib/parser/ast';
 
 export type Options = [
   {
@@ -22,6 +26,7 @@ export type Options = [
     checkVersionMismatches?: boolean;
     checkMissingPackageJson?: boolean;
     ignoredDependencies?: string[];
+    includeTransitiveDependencies?: boolean;
   }
 ];
 
@@ -51,6 +56,7 @@ export default createESLintRule<Options, MessageIds>({
           checkMissingDependencies: { type: 'boolean' },
           checkObsoleteDependencies: { type: 'boolean' },
           checkVersionMismatches: { type: 'boolean' },
+          includeTransitiveDependencies: { type: 'boolean' },
         },
         additionalProperties: false,
       },
@@ -69,6 +75,7 @@ export default createESLintRule<Options, MessageIds>({
       checkObsoleteDependencies: true,
       checkVersionMismatches: true,
       ignoredDependencies: [],
+      includeTransitiveDependencies: false,
     },
   ],
   create(
@@ -80,6 +87,7 @@ export default createESLintRule<Options, MessageIds>({
         checkMissingDependencies,
         checkObsoleteDependencies,
         checkVersionMismatches,
+        includeTransitiveDependencies,
       },
     ]
   ) {
@@ -92,8 +100,7 @@ export default createESLintRule<Options, MessageIds>({
       return {};
     }
 
-    const projectPath = normalizePath(globalThis.projectPath || workspaceRoot);
-    const sourceFilePath = getSourceFilePath(fileName, projectPath);
+    const sourceFilePath = getSourceFilePath(fileName, workspaceRoot);
     const { projectGraph, projectRootMappings, projectFileMap } =
       readProjectGraph(RULE_NAME);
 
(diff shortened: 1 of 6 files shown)


meeroslav (changes requested):
I find this flag too rigid to cover all three possibilities. The code should rather look recursively in non-buildable dependencies (we already have the information about "buildable" targets) while ignoring buildable dependencies's transitive deps.

When importing non-buildable we want to include `lodash` in deps since the entire `util-lib` will be bundled into `buildable`:
```
buildable -> util-lib -> npm:lodash
```

When importing buildable library we want to only include a dependency to that one while ignoring it's transitive deps. In this case `rxjs` is not included:
```
buildable -> another-buildable -> npm:rxjs
```

With `{ recursive: boolean }` problem occurs when we import both types:
```
buildable -> util-lib -> npm:lodash
buildable -> another-buildable -> npm:rxjs
```
While we want to include `lodash`, we do not want to include `rxjs` and the `recursive` flag is all in and does not allow granularity.

jaysoo:
> I find this flag too rigid to cover all three possibilities. 

As discussed offline, the option is now `includeTransitiveDependencies` to be more clear on what it does. It will not be reliable enough for us to pick up all types of situations,  especially since a dependency could be built using `run-commands` on a different target name from the parent's build target.

The option `includeTransitiveDependencies` (either `true` or `false`) should cover most workspace setups i.e. all packages are buildable or none are. In cases where there is a mix of buildable and non-buildable dependencies that are used, users can set `includeTransitiveDependencies: true` to include all transitive dependencies, then exclude the ones they don't want.

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -120,32 +126,16 @@ export default createESLintRule<Options, MessageIds>({
...
-
     const rootPackageJson = getPackageJson(join(workspaceRoot, 'package.json'));
 
-    // find all dependencies for the project
-    const npmDeps = findProjectsNpmDependencies(
+    const npmDependencies = findNpmDependencies(
+      workspaceRoot,
       sourceProject,
       projectGraph,
-      buildTarget,
```
meeroslav, line 136: Why not pass in `buildTarget` so we can get a file set for filtered files (instead of explicit ignoreFilePatterns?

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -350,13 +341,9 @@ export default createESLintRule<Options, MessageIds>({
...
 
         if (expectedDependencyNames.includes(packageName)) {
-          return validateVersionMatchesInstalled(
-            node,
-            packageName,
-            packageRange
-          );
+          validateVersionMatchesInstalled(node, packageName, packageRange);
         } else {
-          return reportObsoleteDependency(node, packageName);
```
meeroslav, line 359: We should either remove `return` everywhere or keep it as it was.

packages/js/src/utils/find-npm-dependencies.ts
```diff
@@ -0,0 +1,203 @@
...
+  return null;
+}
+
+function collectPeerAndOptionalDependencies(
+  workspaceRoot: string,
+  sourceProject: ProjectGraphProjectNode,
+  projectGraph: ProjectGraph,
+  npmDeps: Record<string, string>
+): void {
+  const packageJson = readPackageJson(sourceProject, workspaceRoot);
```
meeroslav, line 155: What would be the purpose of the function?

Whatever I add manually as a `peer` or `optional` dependency in `package.json` will be propagated back to the linter rule as a "detected" dependency?

packages/js/src/utils/find-npm-dependencies.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+function collectHelperDependencies(
+  workspaceRoot: string,
+  sourceProject: ProjectGraphProjectNode,
+  projectGraph: ProjectGraph,
+  npmDeps: Record<string, string>
+): void {
+  for (const target of Object.values(sourceProject.data.targets)) {
+    if (target.executor === '@nx/js:tsc' && target.options?.tsConfig) {
```
meeroslav, line 181: Knowing which targets we look for can simplify this.

They might have an executor with `@nx/js:tsc` but they are using something else for the `build`.

meeroslav (changes requested):
Left some comments

packages/eslint-plugin/src/rules/dependency-checks.ts
jaysoo: Didn't seem necessary since the return  was void, but I see it's the case in the other functions too. I'll update.

packages/js/src/utils/find-npm-dependencies.ts
jaysoo: I was getting a lot of optional dependencies and peer dependencies being incorrectly picked up as missing from `package.json`.

e.g.

```
> nx lint nx
  34:3  error  The "nx" uses the following packages, but they are missing from the "dependencies":
    - @nx/nx-win32-x64-msvc
    - @nx/nx-win32-arm64-msvc
    - @nx/nx-darwin-x64
    - @nx/nx-darwin-arm64
    - @nx/nx-freebsd-x64
    - @nx/nx-linux-x64-musl
    - @nx/nx-linux-x64-gnu
    - @nx/nx-linux-arm64-musl
    - @nx/nx-linux-arm64-gnu
    - @nx/nx-linux-arm-gnueabihf  @nx/dependency-checks
```

It could be another problem though so I can take a closer look.

jaysoo: Okay, it was due to `getAllDependencies` not also picking up `optionalDependencies`. Updated that function instead.

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -120,32 +127,19 @@ export default createESLintRule<Options, MessageIds>({
...
-    // find all dependencies for the project
-    const npmDeps = findProjectsNpmDependencies(
+    const npmDependencies = findNpmDependencies(
+      workspaceRoot,
       sourceProject,
       projectGraph,
-      buildTarget,
-      rootPackageJson,
+      projectFileMap,
+      buildTarget, // TODO: What if child library has a build target different from the parent?
```
meeroslav, line 137: We would have to list all "build" targets. But it's very unlikely that nomenclature diverges between projects

packages/eslint-plugin/src/rules/dependency-checks.ts
meeroslav: Thanks

packages/eslint-plugin/src/rules/dependency-checks.ts
meeroslav: We assume that there is just a single build target. This needs to be fixed, but we can do this is a followup PR

meeroslav (approved):

ndcunningham (approved):

PR #18067, opened 2023-07-11
docs(misc): fix misleading typo in docs for Incremental Builds

EthanML:
## Current Behavior

Opening section of the docs for "Incremental Builds" read as follows:

> [...] We **cannot** test and lint it independently, but the only way to build it is by building some application using it (in this case myapp). [...]

## Expected Behavior

The statement is incorrect, and is intended to say (emphasis added):

> [...] We **can** test and lint it independently, but the only way to build it is by building some application using it (in this case myapp). [...]

Very minor but it had me scratching my head for a minute 😄

### DIFF
--- a/docs/shared/incremental-builds.md
+++ b/docs/shared/incremental-builds.md
@@ -8,7 +8,7 @@ nx g @nx/react:app myapp
 nx g @nx/react:lib mylib
 ```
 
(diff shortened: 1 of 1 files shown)


isaacplmann (approved):
Thanks!

PR #18071, opened 2023-07-11
docs(js): add docs about @nx/js:tsc batch execution

leosvelperez:
Docs changed:

- https://nx-dev-git-fork-leosvelperez-docs-incremental-build-0394f7-nrwl.vercel.app/more-concepts/incremental-builds#use-the-batch-implementation
- https://nx-dev-git-fork-leosvelperez-docs-incremental-build-0394f7-nrwl.vercel.app/packages/js/executors/tsc

## Current Behavior

There's no mention in the docs about the new `@nx/js:tsc` batch implementation.

## Expected Behavior

The `@nx/js:tsc` batch implementation should be mentioned in the incremental builds docs and in the `@nx/js:tsc` executor examples.

## Related Issue(s)

Fixes #


packages/js/docs/tsc-examples.md
```diff
@@ -71,5 +71,35 @@ npx nx build ts-lib --external=all
 npx nx build ts-lib --external=none
 ```
 
+{% /tab %}
+{% tab label="Batch mode execution" %}
+
+The `@nx/js:tsc` executor supports running multiple tasks in a single process. When running in batch mode, the executor uses the [TypeScript APIs for incremental builds](https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript). This results in a much faster build time when compared to the default implementation (the bigger the task graph to run, the more the performance improvements).
+
+> **Note:** Executing tasks in batch mode is an experimental feature.
```
jaysoo, line 79: Can we use  `{% callout %}` instead so the warning is more visible?

jaysoo (approved):

isaacplmann (approved):
I think this is fine for now.  We should make a follow up PR to rewrite the incremental builds concept guide to stay more on the conceptual level and then we could link to 3 separate recipes:

1. Set up incremental builds with webpack (React)
2. Set up incremental builds with Angular
3. Set up incremental builds with typescript

Recipe 1 could be created from parts of the concept guide.
Recipe 2 is already written.
Your new section would go in recipe 3.

PR #18082, opened 2023-07-12
feat(react): interaction tests story generator

mandarini:
React generator for Storybook interaction tests


packages/react/src/generators/storybook-configuration/configuration.spec.ts
```diff
@@ -1,3 +1,4 @@
+// TODO(katerina): remove Cypress for Nx 17?
```
jaysoo, line 1: Add `TODO(v18)` so we can pick this up later, but yes we should remove it.

jaysoo (approved):

jaysoo:
Approved with one comment.

isaacplmann (approved):
Docs changes look fine

PR #18091, opened 2023-07-12
feat(js): create package.json with proper defaults

jaysoo:
This PR in combination with the new `@nx/dependency-checks` lint rule will allow users to start with good defaults in their source `package.json` files, while the lint rule helps catch/fix problems with dependencies.

**Notes:**
- Running `nx g @nx/js:lib` will create package.json with the correct `type`, `main`, `module`, and `typings` fields
- When users choose tsc or swc then they will get cjs format
- When users choose esbuild, vite, rollup then they will get both cjs and esm

**TODO:**
- [x] Decide on CJS vs ESM for different bundlers
- [x] unit tests
- [x] e2e tests

**Follow-up**
We're in discussion with package and tooling maintainers on the best way to handle `exports`, `imports`, and other considerations for crafting a `package.json` file. Once we have a better understanding on different use cases we will make more feature/doc changes to support maintainers.


e2e/js/src/js-packaging.test.ts
```diff
@@ -0,0 +1,171 @@
+import {
```
jaysoo, line 1: This was a rename from `js-bundling.test.ts` with additional tests, not sure why it wasn't marked as such.

packages/js/src/generators/library/library.ts
```diff
@@ -619,5 +627,77 @@ function createProjectTsConfigJson(tree: Tree, options: NormalizedSchema) {
...
+        type: 'commonjs',
+        main: './src/index.js',
+        typings: './src/index.d.ts',
+      };
+    case 'rollup':
+      return {
+        type: 'commonjs',
+        main: './index.cjs',
+        module: './index.js',
+        // typings is missing for rollup currently
```
jaysoo, line 674: There is another task to support typings here.

packages/js/src/generators/library/library.ts
```diff
@@ -619,5 +627,77 @@ function createProjectTsConfigJson(tree: Tree, options: NormalizedSchema) {
...
+    default: {
+      // In other cases (vite, rollup, esbuild), helpers are bundled so no need to add them as a dependency.
+      return {};
+    }
+  }
+}
+
+type EntryField = string | { [key: string]: EntryField };
+
+function determineEntryFields(
```
jaysoo, line 653: We'll also include `exports` field and other things in a follow-up PR. We just need to figure out how we want to handle that first.

mandarini (approved):

ndcunningham (approved):

e2e/linter/src/linter.test.ts
```diff
@@ -500,9 +492,12 @@ describe('Linter', () => {
           {
             "dependencies": {
               "@nx/devkit": "${nxVersion}",
+              "tslib": "^2.3.0",
```
meeroslav, line 495: This is not correct, we need to use the above deleted `tsLibVersion` from `dependencies` (not `devDependencies`).

meeroslav (approved):
Left one minor change request, but everything else is ok.

PR #18097, opened 2023-07-13
chore(core): move isNxGenerator out of workspaces

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/adapter/ngcli-adapter.ts
+++ b/packages/nx/src/adapter/ngcli-adapter.ts
@@ -48,7 +48,6 @@ import {
   toOldFormat,
 } from './angular-json';
 import { normalizeExecutorSchema } from '../command-line/run/executor-utils';
-import { Workspaces } from '../config/workspaces';
 import {
   CustomHasher,
   Executor,
@@ -1033,8 +1032,6 @@ async function getWrappedWorkspaceNodeModulesArchitectHost(
   } = await import('@angular-devkit/architect/node');
 
   class WrappedWorkspaceNodeModulesArchitectHost extends AngularWorkspaceNodeModulesArchitectHost {
-    private workspaces = new Workspaces(this.root);
-
     constructor(private workspace, private root) {
       super(workspace, root);
     }
@@ -1064,7 +1061,7 @@ async function getWrappedWorkspaceNodeModulesArchitectHost(
       const { json: packageJson, path: packageJsonPath } =
         readPluginPackageJson(
           nodeModule,
-          this.workspaces['resolvePaths'].bind(this.workspaces)()
+          this.root ? [this.root, __dirname] : [__dirname]
         );
       const executorsFile = packageJson.executors ?? packageJson.builders;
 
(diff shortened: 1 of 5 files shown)


packages/nx/src/command-line/generate/generator-utils.ts
```diff
@@ -0,0 +1,156 @@
...
+  resolvedCollectionName: string;
+  normalizedGeneratorName: string;
+  schema: any;
+  implementationFactory: () => Generator<unknown>;
+  isNgCompat: boolean;
+  isNxGenerator: boolean;
+  /**
+   * @deprecated(v16): This will be removed in v16, use generatorConfiguration.aliases
+   */
+  aliases: string[];
```
FrozenPandaz, line 29: I think we can remove this?

packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -123,7 +120,14 @@ export async function scheduleTarget(
...
           nodeModule,
-          this.workspaces['resolvePaths'].bind(this.workspaces)()
+          this.root
+            ? [
+                this.root,
+                __dirname,
+                process.cwd(),
+                ...getNxRequirePaths(this.root),
+              ]
+            : [__dirname, process.cwd(), ...getNxRequirePaths()]
```
FrozenPandaz, line 130: What's going on here? This doesn't seem like the same as the `resolvePaths` method in `Workspaces`.

`Workspaces` was originally instantiated with `this.root` here.

So the logic should be the following `this.root ? [this.root, __dirname] : [__dirname]` right?

I mean `__dirname` is not exactly the same... but i'm not sure if it matters :thinking: 

Whether we resolve from `packages/nx/src/config` (`__dirname` of `Workspaces`) or `packages/nx/src/adapter` (`__dirname`, of `ngcli-adapter`), I don't think makes much of a difference.

FrozenPandaz (changes requested):

packages/nx/src/adapter/ngcli-adapter.ts
xiongemi: how about `this.root ? getNxRequirePaths(this.root) : getNxRequirePaths()`? if not set, readPluginPackageJson will use getNxRequirePaths as default is not passed in

packages/nx/src/command-line/run/executor-utils.ts
```diff
@@ -96,9 +96,7 @@ function readExecutorsJson(
 } {
   const { json: packageJson, path: packageJsonPath } = readPluginPackageJson(
     nodeModule,
-    root
-      ? [root, __dirname, process.cwd(), ...getNxRequirePaths()]
-      : [__dirname, process.cwd(), ...getNxRequirePaths()]
+    root ? [root, __dirname] : [__dirname]
```
FrozenPandaz, line 99: This one should not have changed.

FrozenPandaz (changes requested):

packages/nx/src/command-line/run/executor-utils.ts
xiongemi: revert

FrozenPandaz (approved):

PR #18116, opened 2023-07-14
fix(core): check if socketMessenger is null before using

Cammisuli:
## Current Behavior

In certain scenarios in Nx Console, the daemon gets reset while a request was about to be sent through with the socketmessenger. When the daemon resets, SocketMessenger is set to null, and we get `TypeError: Cannot read properties of null ("sendMessage")`

## Expected Behavior

We now use the `?` operator to make sure we don't call sendMessage on a null messenger

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/daemon/client/client.ts
+++ b/packages/nx/src/daemon/client/client.ts
@@ -37,6 +37,12 @@ export type ChangedFile = {
   type: 'create' | 'update' | 'delete';
 };
 
+enum DaemonStatus {
+  CONNECTING,
+  DISCONNECTED,
+  CONNECTED,
+}
+
 export class DaemonClient {
   constructor(private readonly nxJson: NxJsonConfiguration) {
     this.reset();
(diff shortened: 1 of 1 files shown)


packages/nx/src/daemon/client/client.ts
```diff
@@ -165,7 +165,7 @@ export class DaemonClient {
     );
 
     await this.queue.sendToQueue(() =>
-      messenger.sendMessage({ type: 'REGISTER_FILE_WATCHER', config })
+      messenger?.sendMessage({ type: 'REGISTER_FILE_WATCHER', config })
```
FrozenPandaz, line 168: Instantiate the `messenger` inside of this callback.

packages/nx/src/daemon/client/client.ts
```diff
@@ -295,7 +295,7 @@ export class DaemonClient {
       this.currentResolve = resolve;
       this.currentReject = reject;
 
-      this.socketMessenger.sendMessage(message);
+      this.socketMessenger?.sendMessage(message);
```
FrozenPandaz, line 298: The `this._connected = true` is in the wrong place.

It should come after the server has started in the background.

Then it will happen synchronously with this call and `this.socketMessenger` will always be defined for this.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

alokeutpal (approved):

PR #18118, opened 2023-07-14
docs(core): add stack example recipes

isaacplmann:
- Adds a new Pill component in the docs
- Creates an [Add Lit recipe](https://nx-dev-git-fork-isaacplmann-docs-add-express-nrwl.vercel.app/showcase/add-stack/add-lit)

Screenshot 2023-07-31 at 3 09 09 PM

I'll work on the Express overview on a separate PR so we can unblock people waiting on this PR.


juristr (changes requested):
changes as discussed on the call :)

docs/generated/manifests/menus.json
```diff
@@ -3521,6 +3521,23 @@
         "id": "showcase",
         "isExternal": false,
         "children": [
+          {
+            "name": "Add a New Project",
```
juristr, line 3525: For some reason I cannot collapse this in the side-bar.

Also a question that came to my mind: how do these "Add a New Project" differ from the example repos? Couldn't this be a more elaborate page of the examples ones, like https://nx-dev-git-fork-isaacplmann-docs-add-express-nrwl.vercel.app/showcase/example-repos/apollo-react?

https://github.com/nrwl/nx-recipes/tree/main/lit is also in the `nx-recipes`, so would that also be shown in the "Example Repos" section?

docs/shared/recipes/add-stack/add-lit.md
```diff
@@ -0,0 +1,206 @@
...
+
+The code for this example is available on Github:
+
+{% github-repository url="https://github.com/nrwl/nx-recipes/tree/main/lit" /%}
+
+**Supported Features**
+
+Because we are not using an Nx plugin for Lit, there are few items we'll have to configure manually. We'll have to configure our own build system. There are no pre-created Lit-specific code generators. And we'll have to take care of updating any framework dependencies as needed.
+
+{% pill url="/core-features/run-tasks" %}✅ Run Tasks{% /pill %}
```
juristr, line 11: in the last call we had (while u were on vacation) we briefly saw this and weren't sure how useful it is...kinda the concern I had previously. I understand the attempt to re-surface some of these features but I'm wondering about the maintainability implications.

Like most of the time all of them would be ticked apart from the "Use Task Executors" and "Use Code Generators" and "Automtae Updating Framework" probably...basically whenever it is not generated by either one of our plugins or a community plugin.

docs/shared/recipes/add-stack/add-lit.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Install Lit and Other Dependencies
+
+Install all the dependencies we need:
+
+{% tabs %}
+{%tab label="npm"%}
+
+```shell
+npm i --save-dev @nx/node lit http-server
```
juristr, line 31: why do we need `@nx/node`? Doesn't it  just use `@nx/esbuild`?

juristr (changes requested):

docs/generated/manifests/menus.json
isaacplmann: These should be more polished and are all similar in style.  The example repos could be anything.  They could be showing how multiple technologies can be used together.  They could be showing how a particular complex set up works.

All the "Add a New Project" recipes will be the same style and show people just what they need to know to get started with a single new stack.

docs/shared/recipes/add-stack/add-lit.md
isaacplmann: Yes, a lot of them will be similar, but people aren't going to be reading all of the recipes back to back.  They only care about their particular stack.  Having a lot of check marks is a good thing so we can show that most of the Nx functionality works.  We do need to be clear about which parts of Nx will not be available for their tech stack though.

The whole goal of this effort is to make it clear to people that even though we don't have an official plugin for a stack, they can get a lot of value out of using Nx with it.  That's what these checkmarks emphasize.

docs/shared/recipes/add-stack/add-lit.md
isaacplmann: I'm using the `@nx/node:app` generator to scaffold out the project.  `@nx/esbuild` doesn't have a generator.

I intentionally went that route instead of building up a minimally set up from scratch, because we want to give people simple instructions that will set them up for success.  The goal for these recipes is not to show them the bare minimum files that are needed to get Nx to work with their framework.

`@nx/node:app` gives them unit test and e2e test scaffolding out of the box.

docs/generated/manifests/menus.json
isaacplmann: Also, it looks like you can't collapse the section of a page you're currently looking at.  I don't know if that's intentional or not.

juristr: > These should be more polished and are all similar in style. The example repos could be anything. They could be showing how multiple technologies can be used together. They could be showing how a particular complex set up works.

I'm just skeptical whether to have both vs just going with "Examples" and have the mixed there, where some could be more elaborate compared to others. Because in this case we would need to have the Lit also in the "Example Repos", or at least I'd expect to see it there.

> Also, it looks like you can't collapse the section of a page you're currently looking at. I don't know if that's intentional or not.

Could be..

isaacplmann: ok, I moved the recipe under "Example Repos". It seems weird to have "Showcase" with only one item under it that is also a group, but it isn't a big deal.  Maybe "Example Repos" should be a top level item then.

juristr (approved):

PR #18127, opened 2023-07-14
chore(core): move readNxJson out of Workspaces

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/config/configuration.ts
+++ b/packages/nx/src/config/configuration.ts
@@ -2,10 +2,7 @@ import { Workspaces } from './workspaces';
 import { workspaceRoot } from '../utils/workspace-root';
 import { NxJsonConfiguration } from './nx-json';
 import { ProjectsConfigurations } from './workspace-json-project-json';
-
-export function readNxJson(): NxJsonConfiguration {
-  return new Workspaces(workspaceRoot).readNxJson();
-}
+import { readNxJson } from './nx-json';
 
 // TODO(vsavkin): Remove after Nx 16 is out
 /**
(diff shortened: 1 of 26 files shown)


packages/nx/src/project-graph/file-utils.ts
```diff
@@ -155,10 +154,3 @@ export function readPackageJson(): any {
 }
 // Original Exports
 export { FileData };
-
-// TODO(v16): Remove these exports
-export {
-  readNxJson,
-  readAllWorkspaceConfiguration,
-  workspaceLayout,
-} from '../config/configuration';
```
FrozenPandaz, line 164: Sorry, we actually have to keep these. `@nrwl/devkit@15.4.0` depends on them. https://unpkg.com/@nrwl/devkit@15.4.0/index.js

We'll have to keep these until 17. Could you please change the TODO?

FrozenPandaz (changes requested):

packages/nx/src/config/configuration.ts
```diff
@@ -26,3 +23,5 @@ export function workspaceLayout(): { appsDir: string; libsDir: string } {
     libsDir: nxJson.workspaceLayout?.libsDir ?? 'libs',
   };
 }
+
+export { readNxJson } from './nx-json';
```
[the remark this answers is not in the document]
xiongemi: actually, we need it otherwise the Lerna e2e tests would fail. it is used in https://github.com/lerna/lerna/blob/main/libs/commands/run/src/index.ts#L308

packages/nx/src/project-graph/file-utils.ts
xiongemi: added back

mandarini (approved):
`js` changes look ok

FrozenPandaz (approved):

PR #18132, opened 2023-07-15
fix(angular): removed additionalProperties flag from customWebpackConfig in schema

zerodi:
## Current Behavior
If set additional property in customWebpackConfig, project.json becomes invalid.

## Expected Behavior
Support additional properties in customWebpackConfig.

## Cause
Some other packages uses customWebpackConfig for configuration. This flag makes json invalid if application using configuration which needed by other package, e.g. single-spa-angular ('cause in this block it's need additional property libraryName for normal work)

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/builders/webpack-browser/schema.json
+++ b/packages/angular/src/builders/webpack-browser/schema.json
@@ -455,8 +455,7 @@
           "type": "string"
         }
       },
-      "x-priority": "important",
-      "additionalProperties": false
+      "x-priority": "important"
     },
     "indexFileTransformer": {
       "description": "Path to transformer function to transform the index.html",


leosvelperez:
Thanks for the contribution!

The `single-spa-angular` is meant to be used with the `@angular-builders/custom-webpack` builders as you can see in their docs: https://single-spa.js.org/docs/ecosystem-angular.html#use-custom-webpack. The options you're trying to set are not supported by the `@nx/angular:webpack-browser` executor, and that's what the schema correctly validates.

You can instead ask in the [`single-spa-angular` repo](https://github.com/single-spa/single-spa-angular) for supporting this. I can see there's a lookup for the `libraryName` in the old `workspace.json` (https://github.com/single-spa/single-spa-angular/blob/master/libs/single-spa-angular/webpack/index.ts#L210). They might need to update their supported version of Nx and account for the new configuration.

PR #17759, opened 2023-06-23
fix(misc): update semver dependency

FrozenPandaz:
## Current Behavior

Dependency on `semver` is at `7.3.4` which has [a ReDoS vulnerability](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw)

## Expected Behavior

Dependency on `semver` is updated to `7.5.3` which has the vulnerability resolved.

## Related Issue(s)

Fixes #

PR #18140, opened 2023-07-17
fix(misc): use caret for semver version

LukasTy:
## Current Behavior

Currently, the `semver` package has a locked version, whereas most of the other dependencies seem to be prefixed with either caret or a tilde to allow library consumers to update the version they use.
Am I missing something, but having this dependency locked to this specific version does not seem to be required for any specific reason?

## Expected Behavior

Having such a common package not locked to a specific version would avoid the requirement of having multiple dependencies for even the same major version as can be observed here:
- https://github.com/mui/mui-x/pull/9640/files
- https://github.com/mui/mui-x/pull/9692/files

Related PR: #17759

### DIFF
--- a/package.json
+++ b/package.json
@@ -264,7 +264,7 @@
     "rxjs": "^7.8.0",
     "sass": "1.55.0",
     "sass-loader": "^12.2.0",
-    "semver": "7.5.3",
+    "semver": "^7.5.3",
     "source-map": "0.7.3",
     "source-map-loader": "^3.0.0",
     "source-map-support": "0.5.19",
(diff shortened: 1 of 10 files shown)


AgentEnder (approved):

package.json
```diff
@@ -371,4 +371,3 @@
     ]
   }
 }
-
```
LukasTy, line 374: Let me know if this change is problematic.
It has been added by my IDE when saving this file.
Some of the other packages `json` files have the ending empty line, but some do not. So, I wasn't sure how to approach it. 🤷

LukasTy:
@AgentEnder I've synched the branch with master and also changed other instances of direct `semver` dep to be used with a caret. 😉

mandarini (approved):

Coly010 (approved):

PR #18141, opened 2023-07-17
feat(nx-dev): when feedback button is pressed hide buttons

mandarini:
## Current Behavior

A user clicks the feedback button and there's no indication of whether the feedback went through.

## Expected Behavior

There should be some sort of indication that the feedback went through.

Step 1 - Click button:
Screenshot 2023-08-02 at 1 25 24 PM

Step 2 - Buttons disappear - "Thank you for your feedback" is displayed
Screenshot 2023-08-02 at 1 25 28 PM

## Notes

I think this can be considered a temporary solution while we're still working with this existing UI. When we switch to a more designed, chat-like interface, maybe this can be integrated in there?

### DIFF
--- a/nx-dev/feature-ai/src/lib/feature-ai.tsx
+++ b/nx-dev/feature-ai/src/lib/feature-ai.tsx
@@ -10,13 +10,13 @@ export function FeatureAi(): JSX.Element {
   const [error, setError] = useState(null);
   const [query, setSearchTerm] = useState('');
   const [loading, setLoading] = useState(false);
+  const [feedbackSent, setFeedbackSent] = useState<boolean>(false);
 
   const warning = `
   {% callout type="warning" title="Always double check!" %}
   This feature is still in Alpha.
   The results may not be accurate, so please always double check with our documentation.
 {% /callout %}
-
   `;
 
   const handleSubmit = async () => {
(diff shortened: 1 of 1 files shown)


MaxKless:
Works well but the toast is on the complete opposite of the screen from the buttons. Even though I know the toast exists I missed it on a large screen :P Maybe it could be centered?

FrozenPandaz:
Instead of a toast, I'd replace the feedback buttons after they are pressed. We don't want them to be pressed twice so we should just remove them.

mandarini:
@FrozenPandaz I replaced the toast with your suggestion which makes more sense! :)

nx-dev/feature-ai/src/lib/feature-ai.tsx
```diff
@@ -73,43 +86,43 @@ export function FeatureAi(): JSX.Element {
...
+              </Button>
+            </div>
+          )}
+          {feedbackSent && (
+            <p>
+              <span role="img" aria-label="check">
+                ✅
+              </span>{' '}
+              Thank you for your feedback!
+            </p>
```
FrozenPandaz, line 124: I think we can still make this prettier but it's low priority and can be done in a follow up.

FrozenPandaz (approved):

isaacplmann (approved):

PR #18163, opened 2023-07-18
feat(angular): add playwright to e2eTestRunner option

barbados-clemens:
> Note: this shouldn't be in a released until playwright plugin is published.

Blocked By: https://github.com/nrwl/nx/pull/18190 and https://github.com/nrwl/nx/pull/18202

## Current Behavior

angular app generator only support cypress|none.

## Expected Behavior

angular app generator supports playwright|cypress|none

## Related Issue(s)

Fixes #


packages/angular/src/generators/application/lib/add-e2e.ts
```diff
@@ -25,6 +29,24 @@ export async function addE2e(tree: Tree, options: NormalizedSchema) {
       skipPackageJson: options.skipPackageJson,
       skipFormat: true,
     });
+  } else if (options.e2eTestRunner === 'playwright') {
+    addProjectConfiguration(tree, options.e2eProjectName, {
```
xiongemi, line 33: would it make playwright a dev dependency of angular regardless what e2eTestRunner developers choose. for cypress, i saw some places use ensurePackage to import from @nx/cypress library, like:
```
  const { cypressProjectGenerator } = ensurePackage<
    typeof import('@nx/cypress')
  >('@nx/cypress', nxVersion);
```
should we do the same for playwright?

barbados-clemens: good point. I'll refactor to use `ensurePackage`

xiongemi (approved):

xiongemi:
the ci failed on depcheck, should we add `@nx/playwright` to scripts/depcheck/missing.ts for all?

xiongemi:
also, in angular init packages/angular/src/generators/init/init.ts, there is an import from @nx/cypress directly, since it is removed from package.json of angular, i think that is why e2e are failing

packages/angular/src/generators/init/init.ts
```diff
@@ -21,6 +20,7 @@ import {
 } from '../utils/version-utils';
 import type { PackageVersions } from '../../utils/backward-compatible-versions';
 import { Schema } from './schema';
+import { nxVersion } from '@nx/vite';
```
xiongemi, line 23: the import path of nxVersion should not from pakcage

packages/angular/src/generators/application/schema.json
```diff
@@ -105,7 +105,7 @@
     },
     "e2eTestRunner": {
       "type": "string",
-      "enum": ["cypress", "none"],
+      "enum": ["cypress", "playwright", "none"],
```
jaysoo, line 108: We should prompt for this:

```
      "x-prompt": "Which E2E test runner would you like to use?",
```

e2e/angular-core/src/ng-add.test.ts
```diff
@@ -438,8 +438,8 @@ describe('convert Angular CLI workspace to an Nx workspace', () => {
     // add other projects
     const app1 = uniq('app1');
     const lib1 = uniq('lib1');
-    runCommand(`ng g @schematics/angular:application ${app1}`);
-    runCommand(`ng g @schematics/angular:library ${lib1}`);
+    runCommand(`ng g @schematics/angular:application ${app1} --no-interactive`);
+    runCommand(`ng g @schematics/angular:library ${lib1} --no-interactive`);
```
barbados-clemens, line 442: while testing I kept getting prompted for adding routing/using stand alone components from this spec file. added to prevent tests from prompting user input for local testing (I assume ng cli turns off prompting for CI, which is why this has been working?)

Coly010: Yeah I think CI forces `--no-interactive`

Coly010 (approved):

isaacplmann (approved):

AgentEnder (approved):

packages/angular/project.json
```diff
@@ -74,5 +74,5 @@
     },
     "lint": {}
   },
-  "implicitDependencies": ["workspace", "cypress", "jest"]
+  "implicitDependencies": ["workspace", "playwright", "cypress", "jest"]
```
FrozenPandaz, line 77: See if these implicitDependencies are really needed.

packages/angular/src/generators/init/init.ts
```diff
@@ -198,9 +198,20 @@ async function addE2ETestRunner(
 ): Promise<GeneratorCallback> {
   switch (options.e2eTestRunner) {
     case E2eTestRunner.Cypress:
+      const { cypressInitGenerator } = ensurePackage<
```
FrozenPandaz, line 201: Remove this because cypress is installed

FrozenPandaz (approved):

PR #18164, opened 2023-07-18
fix(core): generate npm preset correctly

jaysoo:
This PR fixes a bug where `--preset=npm` is being overwritten by recent changes to the prompts.  Users will get `--preset=ts` instead.

## Current Behavior
`--preset=npm` is being overwritten

## Expected Behavior
`--preset=npm` should work

## Related Issue(s)

Fixes #


juristr (approved):
Yep, looks good now 😃

FrozenPandaz (approved):

PR #18172, opened 2023-07-18
chore(core): nx plugin submission nx-gcp-cache

davidgarciab:
# Community Plugin Submission

## nx-gcp-cache

[nx-gcp-cache](https://github.com/davidgarciab/nx-gcp-cache) plugin provides a distributed remote cache using Google Cloud Storage.

It solves the problem of having a distributed remote cache using a self-hosted solution instead of Nx Cloud.

The plugin provides a generator to update the nx.json file and changes the default runner to use the plugin, prompting the user to provide the bucket name.


davidgarciab:
@isaacplmann I have added E2E test and it's running in CI.
You can check it [here](https://github.com/davidgarciab/nx-gcp-cache/actions)

Let me know if you need anything else to include the plugin.
Thanks!

isaacplmann (approved):

PR #18173, opened 2023-07-19
chore(core): nx plugin submission nx-sst

berenddeboer:
# Community Plugin Submission

## {replace with plugin's name}

The SST plugin generates an [SST](https://sst.dev/) stack, and supports executing all SST cli functions inside an nx monorepo. It supports picking the jest or vitest runner.


FrozenPandaz (approved):
Good work! This looks really cool!

PR #18178, opened 2023-07-19
chore(misc): update nxcloud CI config

rarmatei:
- updates the NxCloud runner version to fix a bug related to how we create executions
- updates the CI config to use executions envs rather than manual run groups

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ jobs:
       - run:
           name: Set dynamic nx run variable
           command: |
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -22,7 +22,8 @@
           "test-native",
           "sitemap",
           "build-storybook",
-          "copy-docs"
+          "copy-docs",
+          "random-target-to-affect-everything"
```
meeroslav, line 26: This should be removed

meeroslav (changes requested):
Left one comment. The rest looks good

meeroslav (approved):

meeroslav:
Not sure if this is intended - prefixing the command with `os`:

image

Either way, it's not consistent.

PR #18187, opened 2023-07-19
fix(nx-dev): fix project config reference headlines

meeroslav:
This fixes Algolia being unable to rank the headlines higher because they are parsed as italic text.

## Related Issue(s)

Fixes #


juristr (approved):
thanks

PR #18190, opened 2023-07-19
feat(testing): add options for frameworks in playwright configuration…

barbados-clemens:
… generator

## Current Behavior

missing QoL improments for playwright plugin to make integration with other generators easier
- no way to configure web server
- missing executor outputs
- isolated test-results directory
- vscode extension recommendation

## Expected Behavior

adding the above

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/playwright/src/generators/configuration/files/playwright.config.ts.template
+++ b/packages/playwright/src/generators/configuration/files/playwright.config.ts.template
@@ -11,6 +11,8 @@ import { defineConfig, devices } from '@playwright/test';
  */
 export default defineConfig({
   testDir: './<%= directory %>',
+  /* When updating this path, make sure to update the outputs in the project.json */
+  outputDir: '<%= offsetFromRoot %>dist/.playwright/<%= projectRoot %>/test-output',
   /* Run tests in files in parallel */
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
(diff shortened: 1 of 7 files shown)


packages/playwright/src/generators/configuration/files/playwright.config.ts.template
```diff
@@ -11,6 +11,7 @@ import { defineConfig, devices } from '@playwright/test';
  */
 export default defineConfig({
   testDir: './<%= directory %>',
+  outputDir: '<%= offsetFromRoot %>dist/playwright/<%= projectRoot %>/test-output',
```
jaysoo, line 14: Add a note that if this path changes, then `project.json` should update too so outputs are cached.

Also, maybe an edge case, but if `dist/playwright` is already used by another project then there is a conflict. Maybe `dist/.playwright/<%= projectRoot %>` or `dist/<%= projectRoot%>-playwright` would be better?

jaysoo (approved):

packages/playwright/src/generators/configuration/files/playwright.config.ts.template
barbados-clemens: it should be fine since each project will be isolated by it's project path. it's the same we have configured for cypress
```
<offsetFromRoot>/cypress/<projectPath>/<screenshots|videos>
```

PR #18192, opened 2023-07-19
fix(core): normalizing key of patched packages with its actual resolved version and not the one from the key

### LINKED ISSUE #17873: Lockfile generated by `@nx/next:build` is invalid
### Current Behavior

When using Yarn 3 in a Nx workspace, running `@nx/next:build` with `generateLockfile` generates a lockfile that contains an empty package name:

```
"":
  version: 5.1.6
  resolution: "typescript@patch:typescript@npm%3A5.1.6#~builtin<compat/typescript>::version=5.1.6&hash=5da071"
  bin:
    tsc: bin/tsc
    tsserver: bin/tsserver
  checksum: f53bfe97f7c8b2b6d23cf572750d4e7d1e0c5fff1c36d859d0ec84556a827b8785077bc27676bf7e71fae538e517c3ecc0f37e7f593be913d884805d931bc8be
  languageName: node
  linkType: hard
```

### Expected Behavior

The generated lockfile should be valid. 

### GitHub Repo

https://github.com/ganlhi/nx-next-lockfile-issue

### Steps to Reproduce

1. Run `yarn install` to install dependencies with Yarn 3
2. Run `nx run my-app:build:production`
3. Look at `dist/packages/my-app/yarn.lock`

### Nx Report

```shell

   Node   : 18.13.0
   OS     : darwin-arm64
   yarn   : 3.6.0

   nx                 : 16.4.1
   @nx/js             : 16.4.1
   @nx/jest           : 16.4.1
   @nx/linter         : 16.4.1
   @nx/workspace      : 16.4.1
   @nx/cypress        : 16.4.1
   @nx/devkit         : 16.4.1
   @nx/eslint-plugin  : 16.4.1
   @nx/next           : 16.4.1
   @nx/react          : 16.4.1
   @nrwl/tao          : 16.4.1
   @nx/web            : 16.4.1
   typescript         : 5.1.6
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

veimox:
## Current Behavior
As described in #17873

My understanding of the sitatuion is that it is an edge case. The issue can be replicated in [the repo](https://github.com/ganlhi/nx-next-lockfile-issue) from @ganlhi where the following happens:
1. In the `package.json` there is the definition ["typescript": "~5.1.3"](https://github.com/ganlhi/nx-next-lockfile-issue/blob/main/package.json#L38) but in the lock file it is resolved with [5.1.6](https://github.com/ganlhi/nx-next-lockfile-issue/blob/main/yarn.lock#L10688-L10689)
2. When [generating the keys](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/lock-file/yarn-parser.ts#L355) for the `yarn.lock` it uses, not the version from the string defined by the user (5.1.3), but the actual version resolved by `yarn` (5.1.6)
3. Then we go an prepare the objects to be pruned. For the specific case of the packages that have `patch` in `berry` we use its nomalised key to prune the lock file . [This normalised key](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/lock-file/yarn-parser.ts#L408) uses the version from the string, and not the resolved from `yarn`, which then in [this bit](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/lock-file/yarn-parser.ts#L412) is trigerred as a deletion because it does not match the version.
4. This effectively [deletes the key](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/lock-file/yarn-parser.ts#L414) giving the issue described in #17873 with an empty `""` key for the value of the `typescript` package. It can seen better in the screenshot

image 

This PR solves the issue but calculating the `normalizedKey` by using the resolved version of `yarn` and fixes the issue.

That said, perhaps there is a better way of fixing this because I do not have in-depth knowledge on the matter. Hopefully will give some info to the dev that can fix it 😄 

## Expected Behavior
It generates a valid `yarn.lock` 

## Related Issue(s)
#17873
(Maybe) #17853 done by @meeroslav 

Fixes #17873

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/yarn-parser.ts
+++ b/packages/nx/src/plugins/js/lock-file/yarn-parser.ts
@@ -398,20 +398,21 @@ function mapSnapshots(
   });
 
   // remove keys that match version ranges that have been pruned away
-  snapshotMap.forEach((keysSet) => {
-    for (const key of keysSet.values()) {
+  snapshotMap.forEach((snapshotValue, snapshotKey) => {
+    for (const key of snapshotValue.values()) {
       const packageName = key.slice(0, key.indexOf('@', 1));
       let normalizedKey = key;
       if (isBerry && key.includes('@patch:') && key.includes('#')) {
         normalizedKey = key
           .slice(0, key.indexOf('#'))
-          .replace(`@patch:${packageName}@`, '@npm:');
+          .replace(`@patch:${packageName}@`, '@npm:')
+          .replace(/:.*$/u, ':' + snapshotKey.version);
       }
       if (
         !existingKeys.get(packageName) ||
         !existingKeys.get(packageName).has(normalizedKey)
       ) {
-        keysSet.delete(key);
+        snapshotValue.delete(key);
       }
     }
   });


veimox:
@meeroslav @AgentEnder  any feedback on this PR or the blocking issue?

meeroslav:
Sorry for waiting. I will look into it as soon as possible.

meeroslav (approved):

meeroslav:
Thank you for this PR!

PR #18196, opened 2023-07-19
chore(repo): update nx to 16.6.0-beta.4

FrozenPandaz:
## Current Behavior

This repo uses Nx `16.6.0-beta.2`

## Expected Behavior

This repo uses Nx `16.6.0-beta.4`

## Related Issue(s)

Fixes #


.eslintrc.json
```diff
@@ -47,6 +47,7 @@
           "error",
           {
             "enforceBuildableLibDependency": true,
+            "checkDynamicDependenciesExceptions": [".*"],
```
meeroslav, line 50: This is practically disabling the rule. I would rather:
- fix what should be fixed (e.g. spec files) and
- use comment toggles to switch off others

meeroslav (approved):
I don't like this rule disabling, but we can fix it in a followup PR.

PR #18197, opened 2023-07-19
fix(js): generate ts standalone project with better import path

jaysoo:
This PR fixes an issue with standalone TS projects (`npx create-nx-workspace --preset=ts-standalone`) where the `name` in `package.json` is invalid.

## Current Behavior
The name is something like `@acme/.`

## Expected Behavior
The name should be the name of the folder (e.g. `acme`).

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/storybook/src/storybook-nested.test.ts
+++ b/e2e/storybook/src/storybook-nested.test.ts
@@ -79,7 +79,7 @@ describe('Storybook generators and executors for standalone workspaces - using R
       writeFileSync(
         tmpProjPath(`src/app/test-button.tsx`),
         `
-          import { MyLib } from '@${appName}/my-lib';
+          import { MyLib } from 'my-lib';
 
           export function TestButton() {
             return (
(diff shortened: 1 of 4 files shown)


ndcunningham (approved):

e2e/storybook/src/storybook-nested.test.ts
```diff
@@ -79,7 +79,7 @@ describe('Storybook generators and executors for standalone workspaces - using R
       writeFileSync(
         tmpProjPath(`src/app/test-button.tsx`),
         `
-          import { MyLib } from '@${appName}/my-lib';
+          import { MyLib } from 'my-lib';
```
jaysoo, line 82: Since the standalone app isn't scoped by default now (just the app's name in `package.json`), the library does not default to a scope either.

AgentEnder (approved):

FrozenPandaz (approved):

PR #18200, opened 2023-07-19
feat(react): add playwright to e2eTestRunner option

xiongemi:
## Related Issue(s)

Fixes #


packages/react/src/generators/application/schema.json
```diff
@@ -117,7 +117,7 @@
     },
     "e2eTestRunner": {
       "type": "string",
-      "enum": ["cypress", "none"],
+      "enum": ["cypress", "playwright", "none"],
```
jaysoo, line 120: We should prompt for this:

```
      "x-prompt": "Which E2E test runner would you like to use?",
```

xiongemi: changed.

packages/react/src/generators/application/lib/add-e2e.ts
```diff
@@ -0,0 +1,61 @@
...
+import { webStaticServeGenerator } from '@nx/web';
+
+import { nxVersion } from '../../../utils/versions';
+import { NormalizedSchema } from '../schema';
+
+export async function addE2e(
+  tree: Tree,
+  options: NormalizedSchema
+): Promise<GeneratorCallback> {
+  if (options.e2eTestRunner === 'none') {
```
ndcunningham, line 18: small nitpick here I would make this a switch statement instead of an `if else` but that's a personal choice

ndcunningham (approved):

packages/react/src/generators/application/lib/add-e2e.ts
xiongemi: changed

isaacplmann (approved):

AgentEnder (approved):
eslintrc looks good

PR #18214, opened 2023-07-20
fix(nx-dev): make sure all redirects are used

meeroslav:
## Related Issue(s)

Fixes #


nx-dev/nx-dev/redirect-rules.js
```diff
@@ -192,6 +192,29 @@ const guideUrls = {
     '/cypress/generators/migrate-to-cypress-11',
 };
 
+const langSpecificGuideUrls = ['/l/r', '/l/a', '/l/n'].reduce(
+  (acc, cur) => ({
+    ...acc,
+    [`${cur}/core-concepts/configuration`]: '/reference/project-configuration',
```
juristr, line 198: why do we need the language specific URL redirects?

meeroslav: We might not need them since we have:
```
'/(l|latest|p|previous)/(a|angular|r|react|n|node)/:path*': '/:path*',
```

Let me test that quickly

juristr (approved):

PR #18229, opened 2023-07-20
chore(core): move relativeCwd out of workspaces

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/executors/read-target-options.ts
+++ b/packages/devkit/src/executors/read-target-options.ts
@@ -3,6 +3,7 @@ import type { ExecutorContext } from 'nx/src/config/misc-interfaces';
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { combineOptionsForExecutor } from 'nx/src/utils/params';
 import { requireNx } from '../../nx';
+import { relative } from 'path';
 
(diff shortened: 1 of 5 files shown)


leosvelperez (approved):

packages/nx/src/config/relative-cwd.ts
```diff
@@ -0,0 +1,5 @@
+import { relative } from 'path';
+
+export function relativeCwd(cwd: string, root: string): string | null {
+  return relative(root, cwd).replace(/\\/g, '/') || null;
```
FrozenPandaz, line 4: `relativeCwd` is basically the same as `relative` now.

Do you know when it actually matters that this is normalized? For the usages above, I don't think it is important. For this function, I feel like we can just inline this where it is used rather than keeping it a separate function.

FrozenPandaz (changes requested):

packages/nx/src/config/relative-cwd.ts
xiongemi: i trace the code, relative path will be passed to convertSmartDefaultsIntoNamedParams, which does normalizing there. so we don't need to normalize it here.

FrozenPandaz (approved):

PR #18245, opened 2023-07-21
feat(core): add monorepo generator to convert from standalone projects

jaysoo:
This PR adds a new `@nx/workspace:monorepo` generator that converts standalone projects to a monorepo.

For example:

```shelll
npx create-nx-workspace react-app
cd react-app
nx g monorepo
```

Then users will see:

```treeview
.
├── apps
│   ├── e2e
│   │   ├── cypress.config.ts
│   │   ├── project.json
│   │   ├── src
│   │   │   ├── ...
│   │   │   └── support
│   │   └── tsconfig.json
│   └── react2
│       ├── index.html
│       ├── project.json
│       ├── public
│       │   └── favicon.ico
│       ├── src
│       │   ├── app
│       │   ├── assets
│       │   ├── main.tsx
│       │   └── styles.css
│       ├── tsconfig.app.json
│       ├── tsconfig.json
│       ├── tsconfig.spec.json
│       └── vite.config.ts
├── nx.json
├── README.md
├── package.json
└── tsconfig.base.json
```

https://github.com/nrwl/nx/assets/53559/4487746d-e0c5-4d0f-90a1-cb784dc96381

## Test Cases

1. Convert a clean standalone project (i.e. one from `create-nx-workspace` without any changes). Test with React, Next.js, Angular, Node servers, TS.
2. Convert a standalone project with libs already added.
3. Convert a standalone project with other apps added.
4. Convert a standalone project with some apps and libs already in `apps` and `libs` directories.
5. Convert a standalone TS project 'my-lib', but `packages/my-lib` or `libs/my-lib` already exists.
6. Convert a monorepo (idempotency)

## Notes

- It's hard to move all root project files to their new location under `apps`since there are workspace files such as `nx.json`, or even custom files that users create that may or may not belong to the root project. The safest way to handle  it is to whitelist the root files we do move, and leave the rest for users to handle.
- The existing `e2e` project is problematic since it is not easy to replace `"e2e"` string in configuration files with `apps/e2e` without messing options like `"target": "e2e"` that should be left untouched. We  may need to handle these cases specifically.

### DIFF
--- a/docs/generated/packages-metadata.json
+++ b/docs/generated/packages-metadata.json
@@ -2843,6 +2843,15 @@
         "type": "generator"
       },
       {
+        "description": "Convert a Nx project to a monorepo.",
+        "file": "generated/packages/workspace/generators/convert-to-monorepo.json",
+        "hidden": false,
+        "name": "convert-to-monorepo",
+        "originalFilePath": "/packages/workspace/src/generators/convert-to-monorepo/schema.json",
+        "path": "workspace/generators/convert-to-monorepo",
+        "type": "generator"
+      },
+      {
         "description": "Create a workspace.",
         "file": "generated/packages/workspace/generators/new.json",
         "hidden": true,
--- a/docs/generated/packages/workspace/generators/convert-to-monorepo.json
+++ b/docs/generated/packages/workspace/generators/convert-to-monorepo.json
@@ -1,0 +1,27 @@
+{
+  "name": "convert-to-monorepo",
+  "factory": "./src/generators/convert-to-monorepo/convert-to-monorepo",
+  "schema": {
+    "$schema": "http://json-schema.org/schema",
+    "$id": "NxWorkspaceConvertToMonorepo",
+    "cli": "nx",
+    "title": "Nx Convert to Monorepo",
+    "description": "Convert an Nx project to a monorepo.",
+    "type": "object",
+    "examples": [
+      {
+        "command": "nx g @nx/workspace:monorepo",
+        "description": "Convert an Nx standalone project to a monorepo."
+      }
+    ],
+    "properties": {},
+    "required": [],
+    "presets": []
+  },
+  "description": "Convert a Nx project to a monorepo.",
+  "implementation": "/packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.ts",
+  "aliases": [],
+  "hidden": false,
+  "path": "/packages/workspace/src/generators/convert-to-monorepo/schema.json",
+  "type": "generator"
+}
--- a/packages/jest/src/generators/init/init.ts
+++ b/packages/jest/src/generators/init/init.ts
@@ -4,6 +4,7 @@ import {
   GeneratorCallback,
   getProjects,
   readNxJson,
+  readProjectConfiguration,
   removeDependenciesFromPackageJson,
   runTasksInSerial,
   stripIndents,
@@ -100,10 +101,12 @@ function createJestConfig(tree: Tree, options: NormalizedSchema) {
       const isProjectConfig = jestTarget?.options?.jestConfig === rootJestPath;
       // if root project doesn't have jest target, there's nothing to migrate
       if (isProjectConfig) {
-        const jestAppConfig = `jest.config.app.${options.js ? 'js' : 'ts'}`;
+        const jestProjectConfig = `jest.config.${
+          rootProjectConfig.projectType === 'application' ? 'app' : 'lib'
+        }.${options.js ? 'js' : 'ts'}`;
 
-        tree.rename(rootJestPath, jestAppConfig);
-        jestTarget.options.jestConfig = jestAppConfig;
+        tree.rename(rootJestPath, jestProjectConfig);
+        jestTarget.options.jestConfig = jestProjectConfig;
         updateProjectConfiguration(tree, rootProject, rootProjectConfig);
       }
       // generate new global config as it was move to project config or is missing
--- a/packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.ts
+++ b/packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.ts
@@ -1,0 +1,50 @@
+import {
+  convertNxGenerator,
+  getProjects,
+  joinPathFragments,
+  ProjectConfiguration,
+  readNxJson,
+  Tree,
+  updateNxJson,
+} from '@nx/devkit';
+import { moveGenerator } from '../move/move';
+
+export async function monorepoGenerator(tree: Tree, options: {}) {
+  const projects = getProjects(tree);
+
+  const nxJson = readNxJson(tree);
+  updateNxJson(tree, nxJson);
+
+  let rootProject: ProjectConfiguration;
+  const projectsToMove: ProjectConfiguration[] = [];
+
+  // Need to determine libs vs packages directory base on the type of root project.
+  for (const [, project] of projects) {
+    if (project.root === '.') rootProject = project;
+    projectsToMove.push(project);
+  }
+
+  // Currently, Nx only handles apps+libs or packages. You cannot mix and match them.
+  // If the standalone project is an app (React, Angular, etc), then use apps+libs.
+  // Otherwise, for TS standalone (lib), use packages.
+  const isRootProjectApp = rootProject.projectType === 'application';
+  const appsDir = isRootProjectApp ? 'apps' : 'packages';
+  const libsDir = isRootProjectApp ? 'libs' : 'packages';
+
+  for (const project of projectsToMove) {
+    await moveGenerator(tree, {
+      projectName: project.name,
+      newProjectName: project.name,
+      destination:
+        project.projectType === 'application'
+          ? joinPathFragments(appsDir, project.name)
+          : joinPathFragments(libsDir, project.name),
+      destinationRelativeToRoot: true,
+      updateImportPath: project.projectType === 'library',
+    });
+  }
+}
+
+export default monorepoGenerator;
+
+export const monorepoSchematic = convertNxGenerator(monorepoGenerator);
--- a/packages/workspace/src/generators/convert-to-monorepo/schema.json
+++ b/packages/workspace/src/generators/convert-to-monorepo/schema.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "$id": "NxWorkspaceConvertToMonorepo",
+  "cli": "nx",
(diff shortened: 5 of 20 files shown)


JamesHenry (changes requested):
Copying over my slack reply just so it doesn't get lost:

> Nice! We have a precedent for conversion generators starting with convert to make it clear it’s mutating existing stuff as its focus, so maybe convert-to-monorepo is preferable?

> As Victor notes it’s going to be run very rarely so arguably clarity can be weighted above brevity because it’s not an ongoing source of friction to type?

> I would worry `nx generate @nx/workspace:monorepo` could easily be confused with create-nx-workspace

docs/generated/packages/workspace/generators/convert-to-monorepo.json
```diff
@@ -0,0 +1,39 @@
...
+      {
+        "command": "nx g @nx/workspace:monorepo --appsDir=apps --libsDir=packages",
+        "description": "Convert to a monorepo with apps and libs directories."
+      }
+    ],
+    "properties": {
+      "appsDir": {
+        "type": "string",
+        "description": "The directory where apps are placed.",
+        "default": "apps"
```
juristr, line 21: should we `x-prompt` here?

docs/generated/packages/workspace/generators/convert-to-monorepo.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "convert-to-monorepo",
+  "factory": "./src/generators/convert-to-monorepo/convert-to-monorepo",
+  "schema": {
+    "$schema": "http://json-schema.org/schema",
+    "$id": "NxWorkspaceConvertToMonorepo",
+    "cli": "nx",
+    "title": "Nx Convert to Monorepo",
+    "description": "Convert a Nx project to a monorepo with apps and libs folders.",
```
juristr, line 9: "Convert an Nx standalone project to a monorepo"

I wouldn't even mention "apps" and "libs" tbh

juristr (changes requested):

packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.ts
```diff
@@ -0,0 +1,86 @@
...
+  await maybeExtractJestConfigBase(tree);
+
+  const nxJson = readNxJson(tree);
+  nxJson.workspaceLayout = {
+    // normalize paths without trailing slash
+    appsDir: joinPathFragments(options.appsDir),
+    libsDir: joinPathFragments(options.libsDir),
+  };
+
+  updateNxJson(tree, nxJson);
```
FrozenPandaz, line 25: Just add the directories and it will have the same effect. Do not write to nx.json.

This can move after the move.

1. If we move an app, we also have to add the libs dir that the user wants
2. If we move a lib, we don't actually have to do anything because moving the project will create the libs dir

packages/workspace/src/generators/convert-to-monorepo/schema.json
```diff
@@ -0,0 +1,30 @@
...
+      "description": "Convert an Nx standalone project to a monorepo."
+    }
+  ],
+  "properties": {
+    "appsDir": {
+      "type": "string",
+      "description": "The directory where apps are placed.",
+      "default": "apps",
+      "x-prompt": "Where do you want to place your apps?"
+    },
```
FrozenPandaz, line 20: Let's always pick `apps` if the root project is an app. Do not prompt.

packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  for (const [_name, project] of projects) {
+    maybeExtractEslintConfigIfRootProject(tree, project);
+    await moveGenerator(tree, {
+      projectName: project.name,
+      destination: project.name,
```
FrozenPandaz, line 31: Pass `destinationRelativeToRoot` so that the destination is taken as is.

This turns into `join(options.appsDir | options.libsDir, project.name)`

packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.ts
```diff
@@ -0,0 +1,86 @@
...
+  Tree,
+  updateNxJson,
+} from '@nx/devkit';
+import { moveGenerator } from '../move/move';
+import { Schema } from './schema';
+
+export async function monorepoGenerator(tree: Tree, options: Schema) {
+  const projects = getProjects(tree);
+  mayebExtractTsConfigBase(tree);
+  await maybeExtractJestConfigBase(tree);
```
FrozenPandaz, line 16: All of the extract, should go into `move`. Regardless of if we are moving from a monorepo or not, if a user moves the root app, we should do this stuff.

This could be refactored more later.

packages/workspace/src/generators/convert-to-monorepo/schema.d.ts
```diff
@@ -0,0 +1,4 @@
+export interface Schema {
+  appsDir: string;
```
FrozenPandaz, line 2: This gets removed.

packages/workspace/src/generators/move/lib/move-project-files.ts
```diff
@@ -12,14 +12,63 @@ export function moveProjectFiles(
...
     // This is a rename but Angular Devkit isn't capable of writing to a file after it's renamed so this is a workaround
-    const content = tree.read(file);
     const relativeFromOriginalSource = relative(project.root, file);
     const newFilePath = join(
       schema.relativeToRootDestination,
       relativeFromOriginalSource
     );
+
+    // Prevents moving the same file infinitely for root projects
+    if (seen.has(file)) return;
```
FrozenPandaz, line 68: This isn't needed anymore after the whitelist.

packages/workspace/src/generators/move/lib/update-project-root-files.ts
```diff
@@ -15,18 +25,74 @@ export function updateProjectRootFiles(
...
       path.join(workspaceRoot, schema.relativeToRootDestination),
       workspaceRoot
     )
     .split(path.sep)
     .join('/');
-  const oldRelativeRoot = path
-    .relative(path.join(workspaceRoot, project.root), workspaceRoot)
-    .split(path.sep)
-    .join('/');
+  const oldRelativeRoot =
```
FrozenPandaz, line 89: This can be reverted.

packages/workspace/src/generators/move/lib/update-project-root-files.ts
```diff
@@ -39,13 +105,7 @@ export function updateProjectRootFiles(
...
   );
   for (const file of tree.children(schema.relativeToRootDestination)) {
-    const ext = extname(file);
-    if (!allowedExt.includes(ext)) {
-      continue;
-    }
-    if (file === '.eslintrc.json') {
-      continue;
-    }
+    if (!shouldUpdate(file)) continue;
```
FrozenPandaz, line 108: This should go back to how it was.

packages/workspace/src/generators/move/move.spec.ts
```diff
@@ -67,4 +67,95 @@ describe('move', () => {
...
+    expect(tree.exists('libs/my-lib/tsconfig.lib.json')).toBeTruthy();
+    expect(tree.exists('libs/my-lib/tsconfig.spec.json')).toBeTruthy();
+    expect(tree.exists('libs/my-lib/.eslintrc.json')).toBeTruthy();
+    expect(tree.exists('libs/my-lib/src/index.ts')).toBeTruthy();
+
+    // Test that other libs and workspace files are not moved.
+    expect(tree.exists('package.json')).toBeTruthy();
+    expect(tree.exists('README.md')).toBeTruthy();
+    expect(tree.exists('.gitignore')).toBeTruthy();
+    expect(tree.exists('other-lib/index.ts')).toBeTruthy();
```
FrozenPandaz, line 159: We should test that root config files are made

packages/workspace/src/generators/move/move.ts
```diff
@@ -1,6 +1,8 @@
 import {
   convertNxGenerator,
+  createProjectGraphAsync,
   formatFiles,
+  ProjectGraph,
```
FrozenPandaz, line 5: Unneeded imports

FrozenPandaz (changes requested):

packages/workspace/src/generators/convert-to-monorepo/schema.json
```diff
@@ -0,0 +1,30 @@
...
+    "libsDir": {
+      "type": "string",
+      "alias": "packagesDir",
+      "description": "The directory where libs/packages are placed.",
+      "default": "libs",
+      "x-prompt": "Where do you want to place your libs/packages?"
+    }
```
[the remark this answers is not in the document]
jaysoo: Turns out, we're restricted to apps-libs vs packages anyway, so removing all the prompts. i.e. you cannot have apps-packages.

packages/jest/src/generators/init/init.ts
```diff
@@ -100,10 +101,12 @@ function createJestConfig(tree: Tree, options: NormalizedSchema) {
       const isProjectConfig = jestTarget?.options?.jestConfig === rootJestPath;
       // if root project doesn't have jest target, there's nothing to migrate
       if (isProjectConfig) {
-        const jestAppConfig = `jest.config.app.${options.js ? 'js' : 'ts'}`;
+        const jestProjectConfig = `jest.config.${
+          rootProjectConfig.projectType === 'application' ? 'app' : 'lib'
+        }.${options.js ? 'js' : 'ts'}`;
```
barbados-clemens, line 106: question (non-blocking): Is there a case where the root project isn't an application?

barbados-clemens (approved):
just a question

packages/jest/src/generators/init/init.ts
jaysoo: Yes, when TS standalone is used the root project is a TS library.

FrozenPandaz (approved):

docs/generated/packages-metadata.json
```diff
@@ -2842,6 +2842,15 @@
         "path": "workspace/generators/remove",
         "type": "generator"
       },
+      {
+        "description": "Convert a Nx project to a monorepo.",
```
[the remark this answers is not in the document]
FrozenPandaz: @ndcunningham Can you make this change in a follow up PR please?

FrozenPandaz (approved):

PR #18260, opened 2023-07-23
feat(linter): Support --print-config feature of ESLint

tonyganchev:
The change introduces a new option to the linter plugin schema: `printConfig`, accepting a file name for which the plugin to output its applicable ESLint configuration.

## Current Behavior
There is no support for `--print-config`.

## Expected Behavior
When `--printConfig` is passed on the command line as a parameter to `nx lint ...` or `printConfig` is specified in the _project.json_ the linter will show the effective ESLint configuration applicable to the specified file and not proceed with linting the project. Operation would complete successfully.

## Related Issue(s)
none

Fixes #

### DIFF
--- a/packages/linter/src/executors/eslint/lint.impl.spec.ts
+++ b/packages/linter/src/executors/eslint/lint.impl.spec.ts
@@ -25,6 +25,9 @@ class MockESLint {
   loadFormatter = mockLoadFormatter;
   isPathIgnored = mockIsPathIgnored;
   lintFiles = mockLintFiles;
+  calculateConfigForFile(file: string) {
+    return { file: file };
+  }
 }
 
 const mockResolveAndInstantiateESLint = jest.fn().mockReturnValue(
@@ -65,6 +68,7 @@ function createValidRunBuilderOptions(
     rulesdir: [],
     resolvePluginsRelativeTo: null,
     reportUnusedDisableDirectives: null,
+    printConfig: null,
     ...additionalOptions,
   };
 }
@@ -618,5 +622,31 @@ Please see https://nx.dev/guides/eslint for full guidance on how to resolve this
       mockContext
     );
     expect(fs.writeFileSync).not.toHaveBeenCalled();
+  });
+
+  it('should print the ESLint configuration for a specified file if printConfiguration is specified', async () => {
+    setupMocks();
+    jest.spyOn(console, 'log');
+    const result = await lintExecutor(
+      createValidRunBuilderOptions({
+        lintFilePatterns: [],
+        eslintConfig: './.eslintrc.json',
+        fix: true,
+        cache: true,
+        cacheLocation: 'cacheLocation1',
+        format: 'stylish',
+        force: false,
+        silent: false,
+        ignorePath: null,
+        maxWarnings: null,
+        outputFile: null,
+        quiet: false,
+        reportUnusedDisableDirectives: null,
+        printConfig: 'test-source.ts',
+      }),
+      mockContext
+    );
+    expect(console.log).toHaveBeenCalledWith('{\n "file": "test-source.ts"\n}');
+    expect(result).toEqual({ success: true });
   });
 });
--- a/packages/linter/src/executors/eslint/lint.impl.ts
+++ b/packages/linter/src/executors/eslint/lint.impl.ts
@@ -39,6 +39,8 @@ export default async function run(
     ? join(options.cacheLocation, projectName)
     : undefined;
 
+  const { printConfig, ...normalizedOptions } = options;
+
   /**
    * Until ESLint v9 is released and the new so called flat config is the default
    * we only want to support it if the user has explicitly opted into it by converting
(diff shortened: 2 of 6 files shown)


meeroslav:
Hi @tonyganchev,

Thank you for creating this PR. 

The purpose of `eslint` executor is to enable us to control the linting and achieve `nx` features such as orchestration, task running, caching, affected graph, etc. We call `eslint CLI` directly when invoking the linting.

Printing out configuration doesn't need any of that and you can easily run it as `npx eslint --print-config name-of-the-file`.

Why would you need this as a parameter in the executor?

tonyganchev:
> Hi @tonyganchev,
> 
> Thank you for creating this PR.
> 
> The purpose of `eslint` executor is to enable us to control the linting and achieve `nx` features such as orchestration, task running, caching, affected graph, etc. We call `eslint CLI` directly when invoking the linting.
> 
> Printing out configuration doesn't need any of that and you can easily run it as `npx eslint --print-config name-of-the-file`.
> 
> Why would you need this as a parameter in the executor?

@meeroslav looking at the code I was under the impression that the nx plugin is invoking the ESLint API directly in `resolveAndInstantiateESLint()` of _eslint-utils.ts_.

In any case I have been hitting issues where running eslint on the command line yields different results from running nx lint e.g. some files don't seem to be selected in nx-orchestrated lint. I needed a way to determine if nx was modifying the ESLint configuration / ruleset or there was a different issue. Turns out the issue was incorrect file selection in the _lint_ target definition in the _project.json_ which I could only spot through manual inspection i.e. looked at the config and miraculously I spotted it :)

Providing access to all capabilities of the orchestrated tools should be critical in helping developers troubleshoot integration issues, learn the nx tooling and therefore trust it.

Best
Tony

meeroslav:
Thank you, @tonyganchev, for clarifying that. It makes more sense now, why one would need this flag supported in our executor.

And yes, you are right; we use API, not CLI (that was a typo on my side, sorry for the confusion).

packages/linter/src/executors/eslint/lint.impl.ts
```diff
@@ -39,6 +39,9 @@ export default async function run(
     ? join(options.cacheLocation, projectName)
     : undefined;
 
+  const printConfigForFile = options.printConfig;
```
meeroslav, line 42: We can maybe use spread instead of mutating the original options

```suggestion
  const { printConfig, ...normalizedOptions } = options.printConfig;
```

packages/linter/src/executors/eslint/lint.impl.ts
```diff
@@ -63,6 +66,23 @@ export default async function run(
...
+      return {
+        success: true,
+      };
+    } catch (err) {
+      console.error(err);
+      return {
+        success: false,
+      };
+    }
+  }
```
meeroslav, line 84: We don't need to mutate the `options` above, just use it here. Since all of the branches return we are safe.

```suggestion
  if (options.printConfig) {
    try {
      const fileConfig = await eslint.calculateConfigForFile(
        options.printConfig
      );
      console.log(JSON.stringify(fileConfig, null, ' '));
      return {
        success: true,
      };
    } catch (err) {
      console.error(err);
      return {
        success: false,
      };
    }
  }
```

packages/linter/src/executors/eslint/lint.impl.ts
```diff
@@ -39,6 +39,9 @@ export default async function run(
     ? join(options.cacheLocation, projectName)
     : undefined;
 
+  const printConfigForFile = options.printConfig;
+  delete options.printConfig;
+
```
meeroslav, line 44: Remove this object mutating. We can use `options.printConfig` directly in the condition below.

```suggestion
```

packages/linter/src/executors/eslint/lint.impl.spec.ts
```diff
@@ -65,6 +65,7 @@ function createValidRunBuilderOptions(
     rulesdir: [],
     resolvePluginsRelativeTo: null,
     reportUnusedDisableDirectives: null,
+    printConfig: null,
```
meeroslav, line 68: Let's add a test that tests this functionality.
We can mock the `calculateConfigForFile` method.

meeroslav (changes requested):

meeroslav:
Can you also add an E2E test, please, to https://github.com/tonyganchev/nx/blob/topic/tonyganchev/linter-print-config/e2e/linter/src/linter.test.ts that would test this flag?

tonyganchev:
> Can you also add an E2E test, please, to https://github.com/tonyganchev/nx/blob/topic/tonyganchev/linter-print-config/e2e/linter/src/linter.test.ts that would test this flag?

Certainly. The latest version now includes a test for this.

packages/linter/src/executors/eslint/lint.impl.spec.ts
tonyganchev: Done :)

meeroslav (approved):
Looks good to me. Thanks!

PR #18274, opened 2023-07-24
fix(nx-dev): fix official plugin badge background

meeroslav:
https://nx-dev-git-fork-meeroslav-fix-official-nrwl.vercel.app/plugins/registry

Badge now works in both light and dark mode

## Related Issue(s)

Fixes #


juristr (approved):
I'm fine with it, but I let @bcabanes weight in here as well. Maybe we could have it a little less intense on the dark theme?

PR #18152, opened 2023-07-18
docs(core): add qwik stack recipe

Coly010:
- Add a new recipe for Qwik

PR #18277, opened 2023-07-24
docs(core): add astro recipe

barbados-clemens:
## Current Behavior

no astro content

## Expected Behavior

add docs page showing the astro example from nx-recipes

blocked by https://github.com/nrwl/nx/pull/18118 and should go in after https://github.com/nrwl/nx/pull/18152 since that PR adds the doc pages
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/recipes/add-stack/add-astro.md
+++ b/docs/shared/recipes/add-stack/add-astro.md
@@ -1,0 +1,149 @@
+# Add an Astro Project
+
+The code for this example is available on GitHub:
+
+{% github-repository url="https://github.com/nrwl/nx-recipes/tree/main/astro-standalone" /%}
+
+**Supported Features**
+
+Because we are not using an Nx plugin for Astro, there are few items we'll have to configure manually. We'll have to configure our own build system. There are no pre-created Astro-specific code generators. And we'll have to take care of updating any framework dependencies as needed.
+
+{% pill url="/core-features/run-tasks" %}✅ Run Tasks{% /pill %}
+{% pill url="/core-features/cache-task-results" %}✅ Cache Task Results{% /pill %}
+{% pill url="/core-features/remote-cache" %}✅ Remote Caching{% /pill %}
+{% pill url="/core-features/explore-graph" %}✅ Explore the Graph{% /pill %}
+{% pill url="/core-features/distribute-task-execution" %}✅ Distribute Task Execution{% /pill %}
+{% pill url="/core-features/integrate-with-editors" %}✅ Integrate with Editors{% /pill %}
+{% pill url="/core-features/automate-updating-dependencies" %}✅ Automate Updating Nx{% /pill %}
+{% pill url="/recipes/enforce-module-boundaries" %}✅ Enforce Project Boundaries{% /pill %}
+{% pill url="/core-features/plugin-features/use-task-executors" %}🚫 Use Task Executors{% /pill %}
+{% pill url="/core-features/plugin-features/use-code-generators" %}🚫 Use Code Generators{% /pill %}
+{% pill url="/core-features/automate-updating-dependencies" %}🚫 Automate Updating Framework Dependencies{% /pill %}
+
+## Create an astro app
+
+```shell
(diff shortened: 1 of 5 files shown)


docs/shared/recipes/add-stack/add-astro.md
```diff
@@ -0,0 +1,130 @@
...
+
+   - Enabled computation caching!
+   - Learn more at https://nx.dev/recipes/adopting-nx/adding-to-existing-project.
+```
+
+You can add a task as cacheable after the fact by updating the `cacheableOperations` in the `nx.json` file. Learn more about [caching task results](/recipes/adopting-nx/adding-to-existing-project#installing-nx-on-a-non-monorepo-project) or [how caching works](https://nx.dev/core-features/cache-task-results).
+
+## Running Tasks
+
+Because Nx [understand package.json scripts](/reference/project-configuration#project-configuration), You can run the predefined scripts via Nx.
```
isaacplmann, line 56:
```suggestion
Because Nx [understands package.json scripts](/reference/project-configuration#project-configuration), You can run the predefined scripts via Nx.
```

docs/shared/recipes/add-stack/add-astro.md
```diff
@@ -0,0 +1,130 @@
...
+{
+  "scripts": {
+    "e2e": "nx exec -- playwright test"
+  }
+}
+```
+
+Now when running `npm run e2e` Nx will be able to check if there is a cache hit or not.
+
+If you plan to only run with Nx cli, then you can omit `nx exec` if so desired. The safest way is to always include it in the package.json script.
```
isaacplmann, line 76:
```suggestion
If you plan to only run tasks with the Nx CLI, then you can omit `nx exec`. The safest way is to always include it in the package.json script.
```

docs/shared/recipes/add-stack/add-astro.md
```diff
@@ -0,0 +1,130 @@
...
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "@myrepo/ui/*": ["ui/src/*"]
+    }
+  }
+}
+```
+
+This allows import in our `.astro` files like so
```
isaacplmann, line 124:
```suggestion
This allows imports in our `.astro` files like so
```

docs/shared/recipes/add-stack/add-astro.md
```diff
@@ -0,0 +1,130 @@
...
+CREATE ui/src/lib/ui.ts
+CREATE ui/tsconfig.lib.json
+CREATE ui/project.json
+CREATE ui/.eslintrc.json
+UPDATE tsconfig.json
+```
+
+If you plan to import `.astro` files within `.ts` files, then you can install the [`@astrojs/ts-plugin`](https://www.npmjs.com/package/@astrojs/ts-plugin).
+
+An easier option is to allow importing the files via the root `tsconfig.json` instead of a `index.ts` file.
```
isaacplmann, line 110: Does `@nx/js:lib` not add a `tsconfig.base.json` file?  I would expect it to create one and add the path automatically.

isaacplmann (changes requested):
Some comments

docs/shared/recipes/add-stack/add-astro.md
barbados-clemens: it adds the path to the root tsconfg.json which is from astro. but it adds the path as 
```json
"paths": {
   "@org/lib": ["lib/src/index.ts"]
}
```
which can cause issues if you import and rexport .astro files.

So you can just change it to allow importing the files directly without having to reexport and use the plugin from astro if you don't want to.

```json
"paths": {
   "@org/lib/*": ["lib/src/*"]
}
```

isaacplmann: Ok, it wasn't clear that you were modifying the path that already existed.  Can you add a sentence about that?

isaacplmann (changes requested):

isaacplmann (approved):

PR #18282, opened 2023-07-24
feat(testing): update generated Playwright test and config

jaysoo:
This PR updates the generated Playwright setup so that `baseURL` is set to the passed in `webServerAddress` (or else use a default of `http://localhost:3000`). Also updates default test to check for "welcome" message that we use for our app generators.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/playwright/src/generators/configuration/files/playwright.config.ts.template
+++ b/packages/playwright/src/generators/configuration/files/playwright.config.ts.template
@@ -3,6 +3,9 @@ import { nxE2EPreset } from '@nx/playwright/preset';
 <% if(!webServerCommand || !webServerAddress) { %>// eslint-disable-next-line @typescript-eslint/no-unused-vars <% } %>
 import { workspaceRoot } from '@nx/devkit';
 
+// For CI, you may want to set BASE_URL to the deployed application.
+const baseURL = process.env['BASE_URL'] || '<% if(webServerAddress) {%><%= webServerAddress %><% } else {%>http://localhost:3000<% } %>';
+
 /**
  * Read environment variables from file.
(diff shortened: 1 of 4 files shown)


packages/playwright/src/generators/configuration/files/playwright.config.ts.template
```diff
@@ -13,8 +16,11 @@ import { workspaceRoot } from '@nx/devkit';
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
-  ...nxE2EPreset(__filename, { testDir: './<%= directory %>' }),
-  /* Run your local dev server before starting the tests */<% if(webServerCommand && webServerAddress) {%> 
+  ...nxE2EPreset(__filename, { testDir: './<>' }),
+  /* Run your local dev server before starting the tests */<% if(webServerCommand && webServerAddress) {%>
+  use: {
+    baseURL,
```
barbados-clemens, line 22: by default playwright sets
```ts
    use: {
      /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
      trace: 'on-first-retry',
    },
```
unsure if we also want to set it here since it'll override the preset that sets it.

barbados-clemens (commented):
just a comment on the `use` property settings.

packages/playwright/src/generators/configuration/files/playwright.config.ts.template
jaysoo: Moved it to the playwright config to we can make `BASE_URL` more clear.

barbados-clemens (approved):

PR #18287, opened 2023-07-24
feat(nx-dev): github star widget

juristr:
## Expected Behavior

Adds a widget under the TOC to promote our Github stars

image

## Related Issue(s)

Fixes #

### DIFF
--- a/nx-dev/nx-dev/lib/githubStars.api.ts
+++ b/nx-dev/nx-dev/lib/githubStars.api.ts
@@ -1,0 +1,33 @@
+import { Octokit } from 'octokit';
+
+let cachedGithubStarCountPromise: null | Promise<number> = null;
+
+export async function fetchGithubStarCount() {
+  if (cachedGithubStarCountPromise !== null) {
+    // If the promise is in the cache, return it directly
+    return cachedGithubStarCountPromise;
+  }
+
+  cachedGithubStarCountPromise = (async () => {
+    try {
+      const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
+      const responseData = await octokit.request('GET /repos/{owner}/{repo}', {
+        owner: 'nrwl',
+        repo: 'nx',
+        headers: {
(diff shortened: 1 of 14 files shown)


nx-dev/nx-dev/lib/githubStars.api.ts
```diff
@@ -0,0 +1,18 @@
+import { Octokit } from 'octokit';
+
+export async function fetchGithubStarCount() {
+  try {
+    const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
+    const responseData = await octokit.request('GET /repos/{owner}/{repo}', {
+      owner: 'nrwl',
```
meeroslav, line 7: Let's make this generic so we can reuse it for third-party packages.
Have `owner` and `repo` be input parameters.

nx-dev/nx-dev/pages/recipes/[...segments].tsx
```diff
@@ -96,6 +100,9 @@ export const getStaticProps: GetStaticProps = async ({
     return {
       props: {
         document,
+        widgetData: {
+          githubStarsCount: await fetchGithubStarCount(),
```
Comment (meeroslav) [inline nx-dev/nx-dev/pages/recipes/[...segments].tsx:104]: At some point, let's improve this to be a serverless function run daily that would update the statically generated codebase.

meeroslav (approved):
Left some improvement notes, but generally it's ok

nx-dev/nx-dev/pages/recipes/[...segments].tsx
Comment (juristr) [reply nx-dev/nx-dev/pages/recipes/[...segments].tsx]: yeah I'm not sure it matters. Since we show rounded to 100 stars it won't matter I think even if it shows less on the website than it is for real on GH. We have enough updates to the docs and rebuilds of the website I think.

nx-dev/nx-dev/lib/githubStars.api.ts
juristr: Let's do that in follow-up PRs once we tackle the community plugin part

PR #18296, opened 2023-07-25
feat(vite): include env files in sharedGlobals for vite projects

gerarts:
Vite will take env vars from envDir https://vitejs.dev/config/shared-options.html#envdir which defaults to `root`, which defaults to `process.cwd()`, which in the case of an nx project is the `{workspaceRoot}`.

This PR adds `'{workspaceRoot}/.env?(.*)'` to the `namedInputs.sharedGlobals` in the vite init generator.

From the [vite docs on env files](https://vitejs.dev/guide/env-and-mode.html#env-files):

> ```
> .env                # loaded in all cases
> .env.local          # loaded in all cases, ignored by git
> .env.[mode]         # only loaded in specified mode
> .env.[mode].local   # only loaded in specified mode, ignored by git
> ```

The `'{workspaceRoot}/.env?(.*)'` pattern will work for all patterns listed above, but not for example with a filed named `.envrc`.

## Current Behavior

`.env` files are not watched for affected detection even though vite depends on these

## Expected Behavior

`.env` files should be considered as inputs when using vite

### DIFF
--- a/packages/vite/src/generators/init/__snapshots__/init.spec.ts.snap
+++ b/packages/vite/src/generators/init/__snapshots__/init.spec.ts.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`@nx/vite:init dependencies for package.json should add required packages 1`] = `
+exports[`@nx/vite:init vite basic setup should add required packages 1`] = `
 {
   "dependencies": {
     "existing": "1.0.0",
--- a/packages/vite/src/generators/init/init.spec.ts
+++ b/packages/vite/src/generators/init/init.spec.ts
@@ -22,7 +22,6 @@ import { initGenerator } from './init';
 
 describe('@nx/vite:init', () => {
   let tree: Tree;
-
   beforeEach(() => {
     tree = createTreeWithEmptyWorkspace();
     projectGraph = {
@@ -31,7 +30,7 @@ describe('@nx/vite:init', () => {
     };
   });
 
-  describe('dependencies for package.json', () => {
+  describe('vite basic setup', () => {
     it('should add required packages', async () => {
       const existing = 'existing';
       const existingVersion = '1.0.0';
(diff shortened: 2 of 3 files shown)


gerarts:
Rebased. Is CircleCI failing a me-problem?

FrozenPandaz (changes requested):

Those files indeed should be in the `inputs` but I don't think it's necessary to add them to every repo since we don't generate those files. I think when adding the file, the user can also add the input.

gerarts:
@FrozenPandaz - I disagree.

Adding the `.env` files in the root of an Nx project works out of the box in Vite because the default configuration finds them in the `{workspaceRoot}` (see description above).

So for a regular user who follows the Vite docs it appears to work, and it will work though builds and everything, until someone tries to ONLY update an env var in an `.env` file, and suddenly we get a bad cache hit.

In a sense, even though these files are not generated by Nx, they are in-scope from the perspective of a Vite project, which I think should warrant them being added to the `inputs` by default.

It is not clear at al that someone should add these files to be watched manually. I think adding them would be good DX in this case.

There is a real risk of the wrong variables from `.env` files ending up in the wrong deployments in the current situation, because you will just get a cache-hit on Nx cloud with a build that contains pre-release values, even though you might be building for production (if you don't know to add the file to the inputs).

mandarini:
Hi there @gerarts ! Thank you for this PR! We have changed a lot since you first submitted it, so could you please change your implementation a bit? Can you please add the inputs in the `vite` plugin?

https://github.com/nrwl/nx/blob/master/packages/vite/src/plugins/plugin.ts#L125

I mean, in the above part of the codebase? Let me know if you need any help!

jaysoo:
I don't think the PR fixes the issue in general for a few reasons.

1. If the user sets `envDir` to something else, then the inputs will not match.
2. The default `cwd` for Nx is no longer the workspace root by default, as we now run Vite CLI from project root.
3. If a value in env file changes, but doesn't affect the task, we don't want the cache to break. It's better to use [Nx's env file support](https://nx.dev/recipes/tips-n-tricks/define-environment-variables), and set `{ env: 'SOME_VAR' }` as an input as it is more precise.

I understand the issue, but I think it's best that users set `inputs` to what's applicable to their workspace.

jaysoo:
I don't think the PR fixes the issue in general for a few reasons.

1. If the user sets `envDir` to something else, then the inputs will not match.
2. The default `cwd` for Nx is no longer the workspace root by default, as we now run Vite CLI from project root.
3. If a value in env file changes, but doesn't affect the task, we don't want the cache to break. It's better to use [Nx's env file support](https://nx.dev/recipes/tips-n-tricks/define-environment-variables), and set `{ env: 'SOME_VAR' }` as an input as it is more precise.

I understand the issue, but I think it's best that users set `inputs` to what's applicable to their workspace.

PR #18313, opened 2023-07-26
feat(linter): add error and warning statistics

SimeonC:
Outputs the error and warning count statistics including how many can be auto-fixed, something I found useful before - especially when changing eslint configs which can generate a lot of errors.

I have noticed that this is only useful in 2 cases;
* Sending the output to a file so eslint statistics aren't shown
* a lot of errors/warnings are being output and the output is truncated in the console

## Current Behavior
Linter outputs only whether there are 1 or more errors or warnings.
```
Lint warnings found in the listed files.

Lint errors found in the listed files.
```

## Expected Behavior
We should get the error/warning count and also the number of fixable error/warnings.

```
8 Lint warnings found in the listed files.

2306 Lint errors found in the listed files. (740 are fixable)
```

### DIFF
--- a/e2e/angular-core/src/ng-add.test.ts
+++ b/e2e/angular-core/src/ng-add.test.ts
@@ -411,14 +411,14 @@ describe('convert Angular CLI workspace to an Nx workspace', () => {
 
     let output = runCLI(`lint ${project}`);
     expect(output).toContain(`> nx run ${project}:lint`);
-    expect(output).toContain('All files pass linting.');
+    expect(output).toContain('All files pass linting');
     expect(output).toContain(
       `Successfully ran target lint for project ${project}`
     );
 
     output = runCLI(`lint ${project}`);
     expect(output).toContain(`> nx run ${project}:lint  [local cache]`);
-    expect(output).toContain('All files pass linting.');
+    expect(output).toContain('All files pass linting');
     expect(output).toContain(
       `Successfully ran target lint for project ${project}`
     );
--- a/e2e/expo/src/expo.test.ts
+++ b/e2e/expo/src/expo.test.ts
@@ -61,10 +61,10 @@ describe('expo', () => {
     expectTestsPass(await runCLIAsync(`test ${libName}`));
 
     const appLintResults = await runCLIAsync(`lint ${appName}`);
-    expect(appLintResults.combinedOutput).toContain('All files pass linting.');
+    expect(appLintResults.combinedOutput).toContain('All files pass linting');
 
     const libLintResults = await runCLIAsync(`lint ${libName}`);
-    expect(libLintResults.combinedOutput).toContain('All files pass linting.');
+    expect(libLintResults.combinedOutput).toContain('All files pass linting');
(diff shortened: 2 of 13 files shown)


JamesHenry:
I'm so sorry for the delay in you receiving a review here @SimeonC! This seems like a good change to me, would it be possible for you to fix up the conflicts please? I'll keep an eye on this one so we can get it merged straight away

SimeonC:
@JamesHenry I admit I had kinda forgotten about this. Looks like I'm a bit fortunate this time and the "conflicts" are just that git has too many commits and a rebase worked.

JamesHenry (changes requested):
Thanks!

Sorry to be a pain, but can we align it a bit more with ESLint's own output and mention the --fix flag, as well as not have the `.` before the parens

Their logic looks like this:

```js
 output += chalk[summaryColor].bold([
                "  ", fixableErrorCount, pluralize(" error", fixableErrorCount), " and ",
                fixableWarningCount, pluralize(" warning", fixableWarningCount),
                " potentially fixable with the `--fix` option.\n"
            ].join(""));
```
from https://github.com/eslint/eslint/blob/98b5ab406bac6279eadd84e8a5fd5a01fc586ff1/lib/cli-engine/formatters/stylish.js#L91

It doesn't have to be exactly that (e.g. we don't have the chalk dep in linter I don't think, but ```potentially fixable with the `--fix` option``` is preferable language please

SimeonC:
@JamesHenry Thanks for the feedback, I went with updating to match the stylish formatted output as much as possible

SimeonC:
Oh, I see that the success message is used in a lot of other tests. @JamesHenry would you prefer I revert the success message change or update the tests to use the stylish formatted success message?

JamesHenry:
@SimeonC I think you can just do find an replace on `toContain('All files pass linting.')` with `toContain('All files pass linting')` (i.e. just remove the period at the end, I think that's the only difference)

SimeonC:
@JamesHenry Thanks, did that.

JamesHenry:
@SimeonC did you push with `--no-verify`? You have failed CI with a check that would have run automatically prepush

SimeonC:
Yea I did. My git client was failing with something like “unexpected error” and no further information. I’ll fix it up tomorrow as I thought it’d probably be OK and was running out of time, my bad 😅

Haven’t used a push hook before, we’ve always used a precommit hook for that so was unexpected.

JamesHenry:
It's your git commit message, you just need to do a commit --amend and add a valid scope to the commit

image

The check is way too expensive (it does a lot of things) to apply to a precommit hook, that is why it is done prepush

SimeonC:
Thanks, my git client just shows "Some refs failed to push", when running `git push` from the command line I get the correct errors and could fix it.

JamesHenry (approved):
Thanks again @SimeonC!

jaysoo (approved):

FrozenPandaz (approved):

PR #18316, opened 2023-07-26
feat(linter): support path wildcards in enforce-module-boundaries autofix

dearlordylord:
…ofix

**What it solves**: supports wildcards from https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping in enforce-module-boundaries `--fix`

Original discussion and bug report https://nrwlcommunity.slack.com/archives/CMFKWPU6Q/p1690292877751699

## Current Behavior

only basic case seems to be handled, the one that is generated by default generators

## Expected Behavior

More advanced aliasing https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping, including `paths: {"@currents/graph-internals/*": ["libs/graph-internals/src/lib/*"],} `, is desirable

## Related Issue(s)

None


packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
```diff
@@ -358,14 +358,20 @@ export default createESLintRule<Options, MessageIds>({
                         dirname(fileName),
                         dirname(importPath)
                       );
+                      const maybeTagRelative = (path: string): string =>
```
meeroslav, line 361: Why was this needed? Wasn't `relative(...)` above already returning a relative path?

packages/eslint-plugin/src/utils/ast-utils.ts
```diff
@@ -27,9 +27,43 @@ function tryReadBaseJson() {
...
+    importScope: string
+  ): string[] => {
+    // TODO check and warn that the entries of paths[importScope] have no wildcards; that'd be user misconfiguration
+    if (paths[importScope]) return paths[importScope];
+    const WILDCARD_TS_SPECIAL_CASE = '*';
+    // accommodate wildcards (it's not glob) https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping
+    return getOneOrNone(
+      Object.entries(paths)
+        .map(([src, dests]) => {
+          const isWildcard = src.endsWith(WILDCARD_TS_SPECIAL_CASE);
```
meeroslav, line 48: Again, please use `*` directly. It makes the code more readable.

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
Firfi: it's relative but when there's only one level of relativity (`./`), relative() would cut `./` making i.e. expected `./my/module` into `my/module`: it goes into imports like this then `import { some } from 'my/module'` which makes resolver think it's an alias or a node lib, whereas what we really need is `import { some } from './my/module'` to reflect that we import relatively to the same dir.

packages/eslint-plugin/src/utils/ast-utils.ts
```diff
@@ -81,6 +115,29 @@ function hasMemberExport(exportedMember, filePath) {
 }
 
 export function getRelativeImportPath(exportedMember, filePath, basePath) {
+  const TSX_SUFFIX = '.tsx';
```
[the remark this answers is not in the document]
Firfi: Got it. I'll change it back soon. Although I think that encapsulating extension doesn't harm readability, I respect the project code guidelines.

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
meeroslav: Can we name this something like `ensureRelativeSignature` or something along those lines. This name doesn't really explain what the function does.

packages/eslint-plugin/src/utils/ast-utils.ts
```diff
@@ -27,9 +27,43 @@ function tryReadBaseJson() {
  */
 export function getBarrelEntryPointByImportScope(
```
meeroslav, line 28: Can we make this simpler?

The `getOneOrNone` seems to be doing a bit more than that since it also returns the result of the function. Additionally, if we are using generics, then they should extend from something other than `any`.

We want to collect all the paths where the alias is a wildcard and matches the import, and then append the delta to the mapped path.

Something like this would work:
```ts
const result = new Set<string>; // set ensures there are no duplicates

Object.entries(paths).forEach((alias, targets) => {
  if (!alias.endsWith('*')) {
    return;
  }
  const strippedAlias = alias.slice(0, -1); // remove asterisk
  if (!importScope.startsWith(strippedAlias)) {
    return;
  }
  const dynamicPart = importScope.slice(strippedAlias.length);
  targets.forEach(target => {
    result.add(target.replace('*', dynamicPart)); // add interpolated value
  })
})

return Array.from(result);
```

I think this is cleaner and more readable, but most importantly - using primitive functions over higher-order functions (map, filter, flat) improves performance significantly.

packages/eslint-plugin/src/utils/ast-utils.ts
```diff
@@ -81,6 +115,29 @@ function hasMemberExport(exportedMember, filePath) {
 }
 
 export function getRelativeImportPath(exportedMember, filePath, basePath) {
+  const TSX_SUFFIX = '.tsx';
+  const TS_SUFFIX = '.ts';
+  const JSX_SUFFIX = '.jsx';
+  const JS_SUFFIX = '.js';
+  const stats = lstatSync(filePath, { throwIfNoEntry: false }); // TODO nodejs bug: can return undefined despite contract https://nodejs.org/api/fs.html#fslstatsyncpath-options
+  if (!stats) {
+    const suffixesToTry = [TSX_SUFFIX, TS_SUFFIX, JSX_SUFFIX, JS_SUFFIX];
```
meeroslav, line 124: Please replace this recursive logic with:
```ts
if (lstatSync(filePath).isDirectory()) {
  // existing unchanged code
} else if (!lstatSync(filePath, { throwIfNoEntry: false })){  // not folder, but not full file either
  // try to find an extension that exists
	const ext = ['.ts', '.tsx', '.js', '.jsx'].find(ext => lstatSync(filePath + ext, { throwIfNoEntry: false }));
  if (ext) {
    filePath += ext;
  }
}

// the rest of the logic unchanged
```

meeroslav (changes requested):
The reasoning is correct, but code is unnecessarily complex. Check my comments and try to make it simpler.

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
Firfi: `ensureRelativeForm` sounds good, I didn't come up with a better name. Renamed!

packages/eslint-plugin/src/utils/ast-utils.ts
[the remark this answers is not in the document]
Firfi: Inlined

packages/eslint-plugin/src/utils/ast-utils.ts
Firfi: Inlined

packages/eslint-plugin/src/utils/ast-utils.ts
Firfi: Used your implementation and validated it against the older code. Looks correct. I took liberty to convert `.forEach(` into `for ... of ...` to be in line with removing higher-order functions from the computation.

packages/eslint-plugin/src/utils/ast-utils.ts
Firfi: Replaced

Firfi:
Thank you for your suggestions @meeroslav , appreciated it! I've updated the PR accordingly.

packages/eslint-plugin/src/utils/ast-utils.ts
meeroslav: The `.forEach` is not a higher-order function and is slightly faster than `for...of` and comparable with plain `for`. But this is not a deal breaker so let's just leave it.

meeroslav (approved):
Looks good! Thank you for the PR and for making the necessary changes.

PR #18317, opened 2023-07-26
fix(core): eslint dep check rule should check prereleases correctly

Coly010:
## Current Behavior

The dep check rule does not check pre-releases correctly 

This surfaced with #17879
You can see the error here: https://staging.nx.app/runs/21ueB3TkiN/task/angular%3lint
The package versions are here: https://github.com/nrwl/nx/blob/4bae74c060bc29336072aba2a0e4c3a4132fb999/packages/angular/package.json#L71

## Expected Behavior

Pre-release version should be picked up correctly

## Related Issue(s)

Fixes #


meeroslav (approved):
Thank you! LGTM

PR #18318, opened 2023-07-26
feat(bundling): generate and bundle type definitions in rollup executor

leosvelperez:
## Current Behavior

The `rollup` executor doesn't generate type definitions for `swc`. The type definitions generated when using the `tsc` and `babel` compilers are not bundled.

## Expected Behavior

The `rollup` executor should generate and bundle type definitions for all compilers.

## Related Issue(s)

Fixes #


jaysoo (approved):

leosvelperez:
@jaysoo I fixed the runtime issue causing the react-core e2e tests to fail, but the react-native and expo e2e tests are still failing even though they are not reported as such. I'm blocking this from merging until that's addressed.

PR #18319, opened 2023-07-26
feat(js): add generateExportsField and additionalEntryPoints options to update exports field when building with tsc/swc

jaysoo:
This PR adds the `generateExportsField` and `additionalEntryPoints` options to `@nx/tsc` and `@nx/swc` executors so we can generate the right fields for either CJS or ESM.

Also fixes an issue with standalone TS projects, where building with SWC doesn't generate a `dist` output that is 1:1 with `src`.

## Usage

```shell
nx g @nx/js:lib tsclib --bundler=tsc
nx g @nx/js:lib swclib --bundler=swc

nx build tsclib --exports
nx build swclib --exports
```

This will generate `package.json` as follows for both libs:

```json5
  "type": "commonjs",
  // ...
  "exports": {
    "./package.json": "./package.json", // always provide package.json import/require
    ".": "./src/index.js"
  }
```

If you update the project to include other entry points, say:

```json5
{
  "build": {
    "executor": "@nx/js:tsc",
     "options": {
      //...
      "additionalEntryPoints: ["{projectRoot}/src/foo.ts"]
    }
  }
  //...
}
```

Then the output `package.json` will show:

```json5
  "type": "commonjs",
  // ...
  "exports": {
    ".": "./src/index.js",
    "./foo": "./src/foo.js"
  }
```

### DIFF
--- a/packages/js/src/executors/tsc/tsc.impl.ts
+++ b/packages/js/src/executors/tsc/tsc.impl.ts
@@ -1,3 +1,5 @@
+import * as ts from 'typescript';
+import { sync as globSync } from 'fast-glob';
 import { ExecutorContext } from '@nx/devkit';
 import type { TypeScriptCompilationOptions } from '@nx/workspace/src/utilities/typescript/compilation';
 import { CopyAssetsHandler } from '../../utils/assets/copy-assets-handler';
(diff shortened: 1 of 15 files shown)


packages/jest/src/generators/init/init.spec.ts
```diff
@@ -287,7 +287,7 @@ module.exports = {
 `
       );
       await jestInitGenerator(tree, { js: true, rootProject: false });
-      expect(tree.exists('jest.config.app.js')).toBeTruthy();
+      expect(tree.exists('jest.config.lib.js')).toBeTruthy();
```
barbados-clemens, line 290: What caused this to switch? I'm not sure I understand the change here.

packages/js/src/executors/tsc/tsc.impl.ts
leosvelperez, line 1: We'll need to replicate these changes in the batch executor as well.

jaysoo: done

packages/js/src/executors/tsc/tsc.impl.ts
```diff
@@ -16,6 +18,26 @@ import { ExecutorOptions, NormalizedExecutorOptions } from '../../utils/schema';
...
+  absolutePathToTsConfig: string
+): 'cjs' | 'esm' {
+  const tsConfig = readTsConfig(absolutePathToTsConfig);
+  switch (tsConfig.options.module) {
+    case ts.ModuleKind.ES2015:
+    case ts.ModuleKind.ES2020:
+    case ts.ModuleKind.ES2022:
+    case ts.ModuleKind.ESNext:
+      return 'esm';
+    case ts.ModuleKind.CommonJS:
```
ndcunningham, line 33: A small nitpick
why not just `default` instead of adding these?

ndcunningham (approved):

mandarini (approved):

PR #18326, opened 2023-07-26
fix(core): shell script invocation should use yarn for berry

meeroslav:
For yarn PnP running `yarn exec` fails since there is no `.bin` in node_modules/

## Related Issue(s)

Fixes #


packages/nx/src/utils/package-manager.ts
```diff
@@ -75,7 +75,7 @@ export function getPackageManagerCommand(
         add: useBerry ? 'yarn add' : 'yarn add -W',
         addDev: useBerry ? 'yarn add -D' : 'yarn add -D -W',
         rm: 'yarn remove',
-        exec: useBerry ? 'yarn exec' : 'yarn',
+        exec: 'yarn',
```
FrozenPandaz, line 78: In `create-nx-workspace` this is set to `npx`... can you fix that one in a follow up PR as well please?

FrozenPandaz (approved):

PR #18340, opened 2023-07-27
fix(repo): fix silent yarn run on e2e for berry

meeroslav:
Yarn berry has no `--silent` flag and would fail the command.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/utils/get-env-info.ts
+++ b/e2e/utils/get-env-info.ts
@@ -76,16 +76,16 @@ export function getNpmMajorVersion(): string {
 export function getYarnMajorVersion(path: string): string {
   try {
     // this fails if path is not yet created
-    const [npmMajorVersion] = execSync(`yarn -v`, {
+    const [yarnMajorVersion] = execSync(`yarn -v`, {
       cwd: path,
       encoding: 'utf-8',
     }).split('.');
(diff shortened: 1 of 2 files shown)


e2e/utils/get-env-info.ts
```diff
@@ -73,6 +73,22 @@ export function getNpmMajorVersion(): string {
...
+      cwd: path,
+      encoding: 'utf-8',
+    }).split('.');
+    return npmMajorVersion;
+  } catch {
+    const [npmMajorVersion] = execSync(`yarn -v`, { encoding: 'utf-8' }).split(
+      '.'
+    );
+    return npmMajorVersion;
+  }
```
barbados-clemens, line 89: this would be `yarnMajorVersion` yeah?
```suggestion
    const [yarnMajorVersion] = execSync(`yarn -v`, {
      cwd: path,
      encoding: 'utf-8',
    }).split('.');
    return npmMajorVersion;
  } catch {
    const [yarnMajorVersion] = execSync(`yarn -v`, { encoding: 'utf-8' }).split(
      '.'
    );
    return yarnMajorVersion;
  }
```

barbados-clemens (approved):

PR #18366, opened 2023-07-28
fix(react-native): update generated jest config to be in sync w/cache

barbados-clemens:
## Current Behavior

Jest config generates coverage reports in the project root, which does not match executor outputs nor the other existing project outputs

jest config doesn't transform common deps

## Expected Behavior

jest config is updated to support the above 2

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react-native/src/utils/add-jest.ts
+++ b/packages/react-native/src/utils/add-jest.ts
@@ -1,4 +1,4 @@
-import { Tree } from '@nx/devkit';
+import { Tree, offsetFromRoot } from '@nx/devkit';
 import { configurationGenerator } from '@nx/jest';
 
 export async function addJest(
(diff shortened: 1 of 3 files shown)


packages/react-native/src/utils/add-jest.ts
```diff
@@ -32,9 +32,13 @@ export async function addJest(
   resolver: '@nx/jest/plugins/resolver',
   moduleFileExtensions: ['ts', 'js', 'html', 'tsx', 'jsx'],
   setupFilesAfterEnv: ['<rootDir>/test-setup.${js ? 'js' : 'ts'}'],
+  transformIgnorePatterns: ['node_modules/(?!react-native|jest-runner)/'],
```
xiongemi, line 35: by default, react native jest preset already has a transformIgnorePatterns like:
https://github.com/facebook/react-native/blob/main/packages/react-native/jest-preset.js#L24
   ` 'node_modules/(?!((jest-)?react-native|@react-native(-community)?)/)',`
is this still needed?

barbados-clemens: in that case probably not. I assume it's just extra deps that were added that ended up needing the patterns without realize it. I'll remove. 👍

xiongemi (approved):

packages/react-native/src/utils/add-jest.ts
AlaaInflyter: @barbados-clemens Oups, i'm a bit late !
This is still needed apparently.
Easy to reproduce: create a bare project, add lib, launch tests, and you'll get something like 

> Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

PR #18380, opened 2023-07-30
docs(storybook): use ESM for Storybook configs

benmccann:
## Current Behavior
The docs currently show CJS for the Storybook configs

## Expected Behavior
When you run `npx sb init` to add Storybook to your project, it creates ESM configs by default. Since users will have ESM config we should use ESM in the docs. Another reason to use ESM is that it is built into the language and more future looking as opposed to the legacy CJS format

## Related Issue(s)
N/A

### DIFF
--- a/docs/generated/packages/storybook/documents/custom-builder-configs.md
+++ b/docs/generated/packages/storybook/documents/custom-builder-configs.md
@@ -47,7 +47,7 @@ async viteFinal(config, { configType }) {
 In the `viteFinal` case, you would have to import the `mergeConfig` function from `vite`. So, on the top of your root `.storybook/main.js|ts` file, you would have to add:
 
 ```ts {% fileName=".storybook/main.js" %}
-const { mergeConfig } = require('vite');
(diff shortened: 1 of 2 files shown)


mandarini (approved):

mandarini:
@benmccann thanks for the contribution. Your commit had some issues, but I made the fixes myself to expedite the process.

In your original commit you had made the changes to:
```
docs/generated/packages/storybook/documents/custom-builder-configs.md
```
instead of
```
docs/shared/packages/storybook/custom-builder-configs.md
```

Please take a look at our ["contributing to docs" section](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md#changing-generated-api-documentation). The changes need to be made in the `.md` file that is under `docs/shared` and then you need to run `pnpm documentation` so that it generates the file again. Don't worry, I understand that it may not be that straightforward, I'm just pointing you to the right direction, for your next contribution to our docs! :D :D :D

Thanks again!

Issue #18177, opened 2023-07-19
Adding filters in a command `nx show projects`


Tom910:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md))

## Description

Hi, In last versions of NX, NX started to write a warning 
```
>  NX  Use `nx show projects --affected`, `nx affected --graph -t build` or `nx graph --affected` depending on which best suits your use case. The `print-affected` command will be removed in Nx 18.
```
when I use the command `nx print-affected --type=app --select=projects`. I tried to migrate on `nx show projects --affected` but I haven't found functionality which can help me filter result

`nx show projects --affected` return list of all affected projects, but I need to filter by projectType. My repository has a release process which we run based on this information and we only want to run release for applications

## Motivation

It helps to migrate from `nx print-affected --type=app --select=projects` to `nx show projects --affected`

## Suggested Implementation

Add functionality in `nx show projects` that will filter projects by some parameters. For example:
```
nx show projects --type=app
nx show projects --tags="tag-1"
```

PR #18381, opened 2023-07-30
docs(core): fix yargs documentation of "nx show"

Chanki-Min:
## Current Behavior

`nx show --help` displays non-available flag `--patterns`, and do not announce `--execlude` option

## Expected Behavior

`nx show --help` should display all proper-working argument list with valid example commands

## Related Issue(s)

https://github.com/nrwl/nx/issues/18177

Fixes #

#18177

### DIFF
--- a/packages/nx/src/command-line/show/command-object.ts
+++ b/packages/nx/src/command-line/show/command-object.ts
@@ -63,6 +63,12 @@ const showProjectsCommand: CommandModule<NxShowArgs, ShowProjectsOptions> = {
         type: 'boolean',
         description: 'Show only affected projects',
       })
+      .option('exclude', {
+        type: 'string',
+        alias: ['e'],
+        description: 'Exclude projects that match a given pattern.',
+        coerce: parseCSV,
+      })
       .option('projects', {
         type: 'string',
         alias: ['p'],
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/show/command-object.ts
```diff
@@ -63,6 +63,12 @@ const showProjectsCommand: CommandModule<NxShowArgs, ShowProjectsOptions> = {
         type: 'boolean',
         description: 'Show only affected projects',
       })
+      .option('exclude', {
+        type: 'string',
+        alias: ['e'],
+        description: 'Exclude projects that match a given pattern.',
+        coerce: parseCSV,
+      })
```
FrozenPandaz, line 71: This is under `nx show projects --help` and should not be added here. If the wording is unclear, it should be updated here: https://github.com/nrwl/nx/blob/master/packages/nx/src/command-line/yargs-utils/shared-options.ts#L5

Also, please do not add an alias. It will possibly conflict with aliases in other executors.

FrozenPandaz (dismissed):

packages/nx/src/command-line/show/command-object.ts
Chanki-Min: Good morning (at least in my country) @FrozenPandaz. Removed `exclude` option with your review. Thank you!

AgentEnder (approved):

PR #18386, opened 2023-07-31
docs update video generators

juristr:
- feat(nx-dev): update youtube cmp to allow caption
- docs(core): update video for generators

https://nx-dev-git-fork-juristr-docs-update-video-generators-nrwl.vercel.app/extending-nx/recipes/local-generators


nx-dev/ui-markdoc/src/lib/tags/youtube.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+    },
+    caption: {
+      // Added caption attribute here
+      type: 'String',
+      required: false, // Not required since it's optional
+    },
+  },
+};
+
+export function YouTube(props: any): JSX.Element {
```
meeroslav, line 26: Let's type those props

nx-dev/ui-markdoc/src/lib/tags/youtube.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <div className="text-center">
+      {' '}
+      {/* Center alignment applied to the container */}
+      <iframe
+        {...props}
```
meeroslav, line 32: We have just 4 props and we explicitly set 2 of them already. Let's make the remaining two explicit as well

meeroslav (changes requested):
Let's two minor notes

meeroslav (approved):

PR #18394, opened 2023-07-31
fix(linter): ensure package detection works in PnP

meeroslav:
## Related Issue(s)

Fixes #


packages/eslint-plugin/src/utils/config-utils.ts
```diff
@@ -6,8 +8,16 @@ import { accessSync } from 'fs';
  * @returns
  */
 export function packageExists(name: string): boolean {
```
FrozenPandaz, line 10: Why does this function not use `require.resolve` to see if the package exists?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #18395, opened 2023-07-31
fix(core): fix schema links

meeroslav:
This issue is blocking the flat config since config reports error

## Current Behavior
Schema links point to non-existent pages

## Expected Behavior
Schema links should point to active pages

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/schemas/nx-schema.json
+++ b/packages/nx/schemas/nx-schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema",
-  "$id": "https://nx.dev/reference/project-configuration#nxjson",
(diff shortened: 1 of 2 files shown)


packages/nx/schemas/nx-schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema",
-  "$id": "https://nx.dev/core-concepts/configuration#nxjson",
+  "$id": "https://nx.dev/reference/project-configuration#nxjson",
```
FrozenPandaz, line 3:
```suggestion
  "$id": "https://nx.dev/reference/nx-json",
```

FrozenPandaz (approved):

PR #18398, opened 2023-07-31
feat(core): make processing typescript dependencies in rust the default

FrozenPandaz:
## Current Behavior

Processing of typescript dependencies defaults to the typescript implementation.

## Expected Behavior

Processing of typescript dependencies defaults to the faster, more accurate Rust implementation.

## Related Issue(s)

Related to #17976

### DIFF
--- a/packages/nx/src/native/plugins/js/ts_import_locators.rs
+++ b/packages/nx/src/native/plugins/js/ts_import_locators.rs
@@ -1,7 +1,6 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
 use std::path::Path;
-use std::rc::Rc;
 use std::sync::Arc;
 use std::time::Instant;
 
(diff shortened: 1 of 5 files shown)


packages/nx/src/native/plugins/js/ts_import_locators.rs
```diff
@@ -589,6 +602,55 @@ mod find_imports {
...
+  nx-ignore-next-line
+  */
+import 'a4'; import 'a5';
+
+/* prettier-ignore */ /* nx-ignore-next-line */
+import 'a4'; import 'a5';
+
+/* nx-ignore-next-line */ /* prettier-ignore */
+import 'a4'; import 'a5';
+
```
Cammisuli, line 636: Adding this as a test case fails :frown:
```
// nx-ignore-next-line 
// prettier-ignore
import 'a6';
```

Cammisuli (changes requested):

packages/nx/src/native/plugins/js/ts_import_locators.rs
FrozenPandaz: That makes sense to me because the import is not on the next line following the comment.

Switching the order of the comments will work.

This doesn't work using the typescript code either.

Cammisuli (approved):

PR #18400, opened 2023-07-31
feat(nx-dev): update terminal experience

juristr:
- docs(nx-dev): add video and terminal output cmps
- feat(nx-dev): update terminal markdown embeds

https://nx-dev-git-fork-juristr-docs-update-terminal-experience-nrwl.vercel.app/getting-started/intro

---

Improves the "appearance" of fences that want to show terminal commands, basically whenever we use `shell` as a fence language, you now get something like this:

image

In addition you can now also have a 

```
 {% terminal-video src="....mp4" /%}
```

which displays a terminal shell just as in our code-fences, but plays basically a dynamic terminal output (which is a recorded mp4 in the end)

https://github.com/nrwl/nx/assets/542458/ddcc750e-5918-44d0-9f10-95c4eef565c3

If the video cannot be played, it'll show the provided alt text, such as

image

Follow-up PRs will use that feature in some of the docs updates


nx-dev/ui-markdoc/src/lib/nodes/fence.schema.tsx
```diff
@@ -9,6 +9,7 @@ export const fence: Schema = {
     command: { type: 'String', default: '' },
     path: { type: 'String', default: '~/workspace' },
     process: { type: 'Boolean', render: false, default: true },
+    enableCopy: { type: 'Boolean', default: true },
```
bcabanes, line 12: `isMessageBelow` seems to be missing

nx-dev/ui-markdoc/src/lib/tags/terminal-video.component.tsx
```diff
@@ -0,0 +1,29 @@
+import { VideoLoop } from './video-loop.component';
+import { Schema } from '@markdoc/markdoc';
+import { TerminalShellWrapper } from '../nodes/fences/terminal-shell.component';
+
+export const terminalVideo: Schema = {
+  render: 'TerminalVideo',
+  attributes: {
```
bcabanes, line 7: `isMessageBelow` seems to be missing

nx-dev/ui-markdoc/src/lib/nodes/fence.schema.tsx
juristr: no this is not something we pass in via the options, but this is being calculated in the fences.component based on the `@nrwl/` to `@nx/..` stuff

nx-dev/ui-markdoc/src/lib/tags/terminal-video.component.tsx
juristr: don't think I need it here, so I'll just pass it ahead to the wrapper where it is needed

bcabanes (approved):

PR #18403, opened 2023-07-31
feat(testing): Vite 5 compatibility

benmccann:
## Current Behavior
Currently doing a `require` of `vite`

## Expected Behavior
Do an `import` of `vite` to use the ESM version as the CJS version is likely to be dropped or cause a warning in the next release


benmccann:
@FrozenPandaz @barbados-clemens @mandarini would you be able to give this simple 1-liner a quick look? Thanks!

mandarini (approved):

mandarini:
@benmccann thanks for the PR! I approved, but still needs approval from a code owner!

packages/cypress/src/plugins/preprocessor-vite.ts
```diff
@@ -74,7 +74,7 @@ function vitePreprocessor(userConfigPath?: string): CypressPreprocessor {
 
     cache.set(filePath, outputPath);
 
-    const { build } = require('vite');
+    const { build } = await import('vite');
```
barbados-clemens, line 77: This doesn't change to using a `require` when building so I don't think this will solve any issues when vite 5 is releases (plus there is a bunch of other work for vite 5 support Nx will have to do).

This is the built code.
```js
const { build } = yield Promise.resolve().then(() => require('vite'));
```

We had this issue with Vitest since it is ESM only.
https://github.com/nrwl/nx/blob/8be2a268b7c110243273d3111a077a891575df54/packages/vite/src/executors/test/vitest.impl.ts#L59-L61


```suggestion   
    const { build } = await (Function(
    'return import("vite")'
  )() as Promise<typeof import('vitest')>);
```

Also as mentioned this change only partially supports vite usage with cypress. There is also component testing side that would need to be updated in the same fashion.

https://github.com/nrwl/nx/blob/8be2a268b7c110243273d3111a077a891575df54/packages/react/plugins/component-testing/index.ts#L86-L87

benmccann: Ah, thanks! I'm more familiar with building with Rollup where `await import` is not transformed

barbados-clemens (approved):

packages/cypress/src/plugins/preprocessor-vite.ts
```diff
@@ -74,7 +74,9 @@ function vitePreprocessor(userConfigPath?: string): CypressPreprocessor {
 
     cache.set(filePath, outputPath);
 
-    const { build } = require('vite');
+    const { build } = await (Function(
+    'return import("vite")'
+  )() as Promise<typeof import('vite')>);
```
benmccann, line 79:
```suggestion
    const { build } = await (Function('return import("vite")')() as Promise<typeof import('vite')>);
```

barbados-clemens:
Looks like a format error. you can run 
`nx format` and it'll clear that up for you.

barbados-clemens:
Hi @benmccann our pipeline validates commit message so they meet a specific format, which is what is current failing on the PR.

https://staging.nx.app/runs/m96AKc6LUX

You'll want to rename the commits to match the described format.

benmccann:
Yeah, I don't know how to do that since some of the commits are further back in history and commits from master have been merged in the middle. I'll just close this and open a clean PR.

benmccann:
New PR here: https://github.com/nrwl/nx/pull/18535

PR #18410, opened 2023-08-01
fix(core): provide taskGraph for every task calculation

Cammisuli:
## Current Behavior

When using the daemon - the taskGraph was not updated while recalculating the inputs for the task hashes.  

## Expected Behavior

The taskGraph is now passed and used for every call to the `hashTask` function, so the latest version of the taskGraph is used. 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/affected/print-affected.ts
+++ b/packages/nx/src/command-line/affected/print-affected.ts
@@ -76,7 +76,6 @@ async function createTasks(
     {},
     [],
     projectGraph,
-    taskGraph,
     nxJson,
     {},
     fileHasher
@@ -85,7 +84,7 @@ async function createTasks(
   const tasks = Object.values(taskGraph.tasks);
 
   await Promise.all(
-    tasks.map((t) => hashTask(hasher, projectGraph, {} as any, t))
+    tasks.map((t) => hashTask(hasher, projectGraph, taskGraph, t))
   );
 
   return tasks.map((task) => ({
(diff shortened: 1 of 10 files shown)


packages/nx/src/hasher/task-hasher.ts
```diff
@@ -59,34 +59,25 @@ export interface Hash {
 }
 
 export interface TaskHasher {
-  hashTask(task: Task): Promise<Hash>;
-  hashTasks(tasks: Task[]): Promise<Hash[]>;
+  hashTask(task: Task, taskGraph?: TaskGraph): Promise<Hash>;
+  hashTasks(tasks: Task[], taskGraph?: TaskGraph): Promise<Hash[]>;
```
FrozenPandaz, line 63: Why is it optional here?

packages/nx/src/command-line/affected/print-affected.ts
```diff
@@ -85,7 +84,7 @@ async function createTasks(
   const tasks = Object.values(taskGraph.tasks);
 
   await Promise.all(
-    tasks.map((t) => hashTask(hasher, projectGraph, {} as any, t))
```
FrozenPandaz, line 88: What was this object being passed for?

FrozenPandaz (changes requested):

packages/nx/src/command-line/affected/print-affected.ts
Cammisuli: it was used for custom hashers 🤷🏼

FrozenPandaz (approved):

PR #18413, opened 2023-08-01
fix(core): esbuild generateLockfile according to docs

dearlordylord:
## Current Behavior

esbuild:esbuild executor will generate package.json

## Expected Behavior

esbuild:esbuild executor will generate package.json together with a lockfile according to documentation https://nx.dev/packages/esbuild/executors/esbuild#generatepackagejson


mandarini (approved):
Thank you for the contribution @Firfi !

mandarini (changes requested):
Can we also add a e2e check? I'll show you where exactly!

Maybe add a line to check below here:
https://github.com/nrwl/nx/blob/master/e2e/esbuild/src/esbuild.test.ts#L54

like:
```
checkFilesExist(`dist/libs/${myPkg}/package-lock.json`);
```

mandarini:
Thank you @Firfi ! I'll take over from here, don't worry!

mandarini (approved):

PR #18425, opened 2023-08-01
fix(core): add verbose logging for CNW

barbados-clemens:
## Current Behavior

when CNW errors out in the sub process its hard to know what happened in some instances.

## Expected Behavior

using NX_VERBOSE_LOGGING=true will log out the sub commands being ran during CNW.
also log error file contents if in CI
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/create-nx-workspace/src/utils/child-process-utils.ts
+++ b/packages/create-nx-workspace/src/utils/child-process-utils.ts
@@ -2,11 +2,14 @@ import { spawn, exec } from 'child_process';
 import { writeFileSync } from 'fs';
 import { join } from 'path';
 import { CreateNxWorkspaceError } from './error-utils';
+import { output } from './output';
+import { isCI } from './ci/is-ci';
 
 /**
  * Use spawn only for interactive shells
  */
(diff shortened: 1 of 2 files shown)


packages/create-nx-workspace/src/utils/child-process-utils.ts
```diff
@@ -27,13 +30,21 @@ export function spawnAndWait(command: string, args: string[], cwd: string) {
...
 export function execAndWait(command: string, cwd: string) {
+  output.verbose('Running:', command, 'in', cwd);
   return new Promise<{ code: number; stdout: string }>((res, rej) => {
     exec(
       command,
       { cwd, env: { ...process.env, NX_DAEMON: 'false' } },
       (error, stdout, stderr) => {
         if (error) {
           const logFile = join(cwd, 'error.log');
+          if (isCI()) {
```
FrozenPandaz, line 41: I don't think we should check this within `create-nx-workspace`. I think people should be able to pass `--verbose` if they are in CI.

FrozenPandaz (changes requested):

PR #18427, opened 2023-08-01
perf(nx): improve performance by utilizing hardware resources with `Promise.all` 🚀

sanjaiyan-dev:
This pull request aims to enhance the performance of the existing functionality by leveraging hardware resources effectively. The main improvement comes from refactoring the code to utilize `Promise.all`  for concurrent execution of tasks.

## Current Behavior
Currently, the function responsible for performing specific tasks runs sequentially, leading to suboptimal performance and underutilization of hardware resources. As a result, the execution time for these tasks is longer than desired, causing delays and reduced responsiveness for users.

## Expected Behavior
With the changes introduced in this pull request, the function will be refactored to utilize `Promise.all` for concurrent execution of tasks. This enhancement will enable parallel processing of tasks, effectively utilizing available hardware resources to improve performance significantly. As a result, the execution time for these tasks will be reduced, leading to a more responsive and efficient system.

---

I apologize in advance for any mistakes I may have made during this process. Please feel free to provide any feedback or suggestions for further improvement. I'm open to making necessary adjustments to ensure the success and quality of this enhancement. Thank you for your review and consideration.

### DIFF
--- a/nx-dev/nx-dev/pages/changelog.tsx
+++ b/nx-dev/nx-dev/pages/changelog.tsx
@@ -97,9 +97,13 @@ export async function getStaticProps(): Promise<{ props: ChangeLogProps }> {
   const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
 
   // do 2 fetches of 100 records, which should be enough releases to display
+  const [githubRelease1, githubRelease2] = await Promise.all([
+    fetchGithubRelease(octokit, 1),
+    fetchGithubRelease(octokit, 2),
+  ]);
   const githubReleases = [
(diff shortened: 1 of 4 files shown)


AgentEnder (changes requested):
Hey @sanjaiyan-dev, thanks for the PR. We appreciate the work that went into it, but are going to close this one out as 2 of the 4 changes aren't impacting users at all.

Additionally, the changes to swc.impl.ts could result in files being copied in different orders which can cause issues if they would collide, and the changes in cache.ts should be a small improvement but would add complexity to error logs if something was to go wrong.

PR #18444, opened 2023-08-02
fix(detox): fix detox to work with yarn pnp

xiongemi:
## Related Issue(s)

Fixes #


meeroslav (approved):
Looks good. I think using `path.join` would be fine as is.

Please test it on a default generated repo to make sure it works.

ndcunningham (approved):

PR #18460, opened 2023-08-03
docs(nx-dev): add recipe for yarn pnp

meeroslav:
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/recipes/yarn-pnp.md
+++ b/docs/shared/recipes/yarn-pnp.md
@@ -1,0 +1,85 @@
+# Using Yarn Plug'n'Play
+
+Plug'n'Play (PnP) is an innovative installation strategy for Node that tries to solve the challenges of using `node_modules` for storing installed packages:
+
+- slow installation process
+- slow node runtime cold start
+- expensive packages diffing when adding new packages
+- no ability to restrict access and enforce hoisting/nesting
+
+Instead of using `node_modules` folder, `PnP` creates maps that take care of mapping package names to hoisted versions and creates dependency edges between packages. The packages are kept in `zip` archives which makes caching and restoring faster.
+
+Read more about `PnP` on [Yarn's official docs](https://yarnpkg.com/features/pnp).
+
+## Switching to Yarn v2+ (aka Berry)
+
+When using `create-nx-workspace` with optional `--pm=yarn`, we will use the default yarn version set in your global `.yarnrc.yml` file (usually in the root of your user folder). To check what is your set yarn version, run:
+
+```shell
+ yarn --version
+```
+
+If version is in `1.x.x` range, the workspace will be created with yarn classic. To migrate to modern version of Yarn, also known as `Berry`, after the workspace has been created, run:
+
+```shell {% path="~/workspace" %}
+ yarn set version stable
+```
+
+This command will update `.yarnrc.yml` with path to local yarn bin of the correct version and will set the `packageManager` property in the root `package.json`:
+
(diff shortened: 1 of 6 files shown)


docs/shared/recipes/yarn-pnp.md
```diff
@@ -0,0 +1,85 @@
...
+node_modules/
+└── .cache
+    └──  nx
+```
+
+## Dealing with inaccessible dependencies
+
+When using Yarn Berry, Nx will set node linker to backwards compatible `node-modules` by default. This is to ensure your repository works out-of-the-box.
+
+Some packages come with broken dependency requirements so using strict `PnP` can lead to broken builds. This results in error similar to one below:
```
isaacplmann, line 69:
```suggestion
Some packages come with broken dependency requirements so using strict `PnP` can lead to broken builds. This results in errors similar to the one below:
```

docs/shared/recipes/yarn-pnp.md
```diff
@@ -0,0 +1,85 @@
...
+
+```shell {% path="~/workspace" command="yarn nx build my-project" %}
+Error: [BABEL]: @babel/plugin-transform-react-jsx tried to access @babel/core (a peer dependency) but it isn't provided by your application; this makes the require call ambiguous and unsound.
+
+Required package: @babel/core
+```
+
+The problem above is something we need to fix by ensuring the requested peer dependency exists as is available:
```
isaacplmann, line 77:
```suggestion
The problem above is something we need to fix by ensuring the requested peer dependency exists and is available:
```

docs/shared/recipes/yarn-pnp.md
```diff
@@ -0,0 +1,85 @@
...
+
+```shell {% path="~/workspace" %}
+ yarn add -D @babel/core
+```
+
+Sometimes, simply adding a missing package doesn't work. In those cases the only thing we can do is to contact the author of package causing the issues and notify them of a missing dependency requirement.
```
isaacplmann, line 83:
```suggestion
Sometimes, simply adding a missing package doesn't work. In those cases the only thing we can do is to contact the author of the package causing the issues and notify them of the missing dependency requirement.
```

docs/shared/recipes/yarn-pnp.md
```diff
@@ -0,0 +1,85 @@
...
+
+The alternative is to switch back to using `node-modules` node linker.
```
isaacplmann, line 85:
```suggestion
The alternative is to switch back to using the `node-modules` node linker.
```

isaacplmann (changes requested):
Grammar fixes.  Looks good otherwise

docs/shared/recipes/yarn-pnp.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Switching to PnP
+
+Once we are on modern version of Yarn, we can use the following command to switch to `PnP`:
+
+```shell {% path="~/workspace" %}
+ yarn config set nodeLinker pnp
+```
+
+Your `.yarnrc.yml` will now have `nodeLinked` set to a proper method:
```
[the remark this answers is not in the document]
isaacplmann: This typo needs to be fixed

isaacplmann:
There are 4 hidden comments that didn't get applied

meeroslav:
All fixed (+ a few more that my Grammarly found)

isaacplmann (approved):

PR #18462, opened 2023-08-03
fix(react): add babel/core when babel/preset-react is preset

meeroslav:
## Current Behavior
Our app generators add `@babel/preset-react` which has dependencies with peer deps to `@babel/core`, but this fails with `PnP` because the `@babel/core` ends up being nested.

## Expected Behavior
The `@babel/core` needs to be added to package.json when using `babel`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/migrations.json
+++ b/packages/react/migrations.json
@@ -98,6 +98,12 @@
     },
     "remove-types-react-router-dom": {
       "cli": "nx",
+      "version": "16.7.0-beta.2",
+      "description": "Add @babel/core to package.json if @babel/preset-react is present",
+      "implementation": "./src/migrations/update-16-7-0/add-babel-core"
+    },
+    "add-babel-core": {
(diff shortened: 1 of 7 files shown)


ndcunningham (approved):

packages/react/migrations.json
```diff
@@ -97,6 +97,12 @@
       "implementation": "./src/migrations/update-16-2-0-remove-package/update-16-2-0-remove-package"
     },
     "remove-types-react-router-dom": {
+      "cli": "nx",
+      "version": "16.7.0-beta.2",
+      "description": "Add @babel/core to package.json if @babel/preset-react is present",
+      "implementation": "./src/migrations/update-16-7-0/add-babel-core"
+    },
+    "add-babel-core": {
```
ndcunningham, line 105: This section looks like it is under the wrong migration

PR #18472, opened 2023-08-03
chore(repo): fix nx package inputs

vsavkin:
## Related Issue(s)

Fixes #

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -58,13 +58,7 @@
   "targetDefaults": {
     "build": {
       "dependsOn": ["build-base", "build-native"],
-      "inputs": [
-        "production",
-        "^production",
-        {
-          "dependentTasksOutputFiles": "**/*.node"
-        }
-      ]
+      "inputs": ["production", "^production"]
     },
     "build-native": {
       "inputs": ["native"]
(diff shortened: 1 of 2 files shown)


Cammisuli (approved):

nx.json
```diff
@@ -58,13 +58,7 @@
   "targetDefaults": {
     "build": {
       "dependsOn": ["build-base", "build-native"],
-      "inputs": [
-        "production",
-        "^production",
-        {
-          "dependentTasksOutputFiles": "**/*.node"
-        }
```
FrozenPandaz, line 66: What is wrong with having it here?

FrozenPandaz (changes requested):

nx.json
Cammisuli: It invalidates the build target cache for every platform. The dependentTaskOutputFiles need to only apply to the nx:build target, because that's the only one who actually depends on the *.node file. 

For example, here's `nx build react` on macos (without having the linux cached node binary) https://staging.nx.app/runs/St0XGtbAWo

The react build target should've been a remote hit. But build-native is correct as a miss.

PR #18477, opened 2023-08-03
fix(bundling): fallback to manual file resolution if tsconfig-paths fails

### LINKED ISSUE #18459: `buildLibsFromSource: false` doesn't work with Vite executors when `package.json` has no `exports` field
### Current Behavior

Suppose an app `my-app` uses the `@nx/vite:dev-server` executor and depends on a lib `my-lib`. Then when the `buildLibsFromSource` option of the executor is set to `false`, `my-lib` is still built from source within `my-app`.

First of all, this fails because of https://github.com/nrwl/nx/issues/18458, but even after patching that, it still doesn't work.

The reason is that when the new `vite-tsconfig-paths` plugin tries to find the bundled module in the build artifacts of the lib, it first checks for an `exports` field in `package.json`, then for `module` and `main`, see here: https://github.com/nrwl/nx/blob/50a145bb64edcf76985184a9a83d1449dcd4f3d9/packages/vite/plugins/nx-tsconfig-paths.plugin.ts#L36
When there is no `exports` field (which is currently the default for a Rollup build, for example), the check for `exports` throws and the plugin falls back to using the default `tsconfig`, not the `tsconfig.generated.json` pointing at the build artifacts.

### Expected Behavior

`my-lib` is not built from source, but its build artifact is used instead.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Patch `@nx/vite` to fix https://github.com/nrwl/nx/issues/18458 by replacing `context.targetName` with `'build'` in `executor-utils`
2. Create a Vite app and a lib, where the app depends on the lib
3. Set `"buildLibsFromSource": false` for the `serve` target of the app
4. Build the lib and serve the app
5. Open the app in a browser and observe that the lib is built from its sources

### Nx Report

```shell
Node   : 20.2.0
OS     : win32-x64

@nx/express        : 16.6.0
@nx/node           : 16.6.0
@nx/plugin         : 16.6.0
@nx/react          : 16.6.0
@nx/rollup         : 16.6.0
@nx/storybook      : 16.6.0
@nrwl/tao          : 16.6.0
@nx/vite           : 16.6.0
@nx/web            : 16.6.0
typescript         : 5.1.6
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [X] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

barbados-clemens:
## Current Behavior

nxViteTsPaths plugin assumes a file can be resolved via package.json fields,
when there isn't always a package.json to resolve from in a mapped tsconfig path.

causing vite to not be able to understand some cjs deps see #18459 

## Expected Behavior

able to resolve files from paths that don't contain package.json

also includes options to plugin for debugging, selecting package.json fields, and extenions for file resolving.

## Related Issue(s)

Fixes #18459

### DIFF
--- a/packages/vite/plugins/nx-tsconfig-paths.plugin.ts
+++ b/packages/vite/plugins/nx-tsconfig-paths.plugin.ts
@@ -1,11 +1,49 @@
 import { stripIndents, workspaceRoot } from '@nx/devkit';
 import { existsSync } from 'node:fs';
 import { relative, join, resolve } from 'node:path';
-import { loadConfig, createMatchPath, MatchPath } from 'tsconfig-paths';
+import {
+  loadConfig,
+  createMatchPath,
+  MatchPath,
+  ConfigLoaderSuccessResult,
+} from 'tsconfig-paths';
 
-export function nxViteTsPaths() {
+export interface nxViteTsPathsOptions {
+  /**
+   * Enable debug logging
+   * @default false
+   **/
+  debug?: boolean;
+  /**
+   * export fields in package.json to use for resolving
+   * @default [['exports', '.', 'import'], 'module', 'main']
+   *
+   * fallback resolution will use ['main', 'module']
+   **/
+  mainFields?: (string | string[])[];
+  /**
+   * extensions to check when resolving files when package.json resolution fails
+   * @default ['.ts', '.tsx', '.js', '.jsx', '.json', '.mjs', '.cjs']
+   **/
+  extensions?: string[];
+}
+
+export function nxViteTsPaths(options: nxViteTsPathsOptions = {}) {
   let matchTsPathEsm: MatchPath;
   let matchTsPathFallback: MatchPath | undefined;
+  let tsConfigPathsEsm: ConfigLoaderSuccessResult;
+  let tsConfigPathsFallback: ConfigLoaderSuccessResult;
+
+  options.extensions ??= [
+    '.ts',
+    '.tsx',
+    '.js',
+    '.jsx',
+    '.json',
+    '.mjs',
+    '.cjs',
+  ];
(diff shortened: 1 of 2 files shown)


barbados-clemens:
Community slack convo for reference: https://nrwlcommunity.slack.com/archives/CMFKWPU6Q/p1691087332889809

and example repo where this fix is needed
https://github.com/luishcastroc/blog

fsgreco:
I think maybe my issue is related to this (?) https://github.com/nrwl/nx/issues/18492

mandarini (approved):

packages/vite/plugins/nx-tsconfig-paths.plugin.ts
```diff
@@ -31,59 +69,102 @@ There should at least be a tsconfig.base.json or tsconfig.json in the root of th
...
+
+      const normalizedImport = alias.replace(/\/\*$/, '');
+
+      if (importPath.startsWith(normalizedImport)) {
+        const path = (tsconfig.absoluteBaseUrl, paths[0].replace(/\/\*$/, ''));
+        resolvedFile = findFile(importPath.replace(normalizedImport, path));
+      }
+    }
+
+    return resolvedFile;
```
jaysoo, line 159: Do we need a `resolvedFile || null` here since the `null` indicates to use other resolveId?

jaysoo (approved):

packages/vite/plugins/nx-tsconfig-paths.plugin.ts
barbados-clemens: Yeah I'll default to null when we don't find one, so the file keeps processing 👍

PR #18478, opened 2023-08-03
fix(react): remove hard-coded typings inside tsconfig and add them to the app

ndcunningham:
## Current Behavior

When you create a react app inside `tsconfig.app.json` and `tsconfig.spec.json` there are hardcoded references to typings inside `node_modules`.

If you use Yarn PnP these hardcoded references would fail since `node_modules` do not exist.

## Expected Behavior

This PR removes the references from `tsconfig` and adds them directly to the app.
That way there would be no need to have hardcoded references and they would be automatically picked up during bundling without the need to anything additional.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/src/generators/library/files/common/tsconfig.lib.json__tmpl__
+++ b/packages/react/src/generators/library/files/common/tsconfig.lib.json__tmpl__
@@ -4,11 +4,6 @@
     "outDir": "<%= offsetFromRoot %>dist/out-tsc",
     "types": ["node"]
   },
-  "files": [<% if (style === 'styled-jsx') { %>
-    "<%= offsetFromRoot %>node_modules/@nx/react/typings/styled-jsx.d.ts",<% } %>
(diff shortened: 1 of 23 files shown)


mandarini (approved):

packages/react/src/generators/library/files/common/tsconfig.lib.json__tmpl__
```diff
@@ -6,8 +6,6 @@
   },
   "files": [<% if (style === 'styled-jsx') { %>
     "<%= offsetFromRoot %>node_modules/@nx/react/typings/styled-jsx.d.ts",<% } %>
```
meeroslav, line 8: The styled-jsx should be removed as well, right?

meeroslav (approved):
Looks good to me.

There is one broken nextjs test that needs to be fixed and I left one comment.

ndcunningham:
Closing in favor of: https://github.com/nrwl/nx/pull/18501

PR #18484, opened 2023-08-04
fix(linter): fix messages on deps checks

meeroslav:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -57,8 +57,8 @@ export default createESLintRule<Options, MessageIds>({
       },
     ],
     messages: {
-      missingDependency: `The "{{projectName}}" uses the following packages, but they are missing from the "{{section}}":{{packageNames}}`,
-      obsoleteDependency: `The "{{packageName}}" package is not used by "{{projectName}}".`,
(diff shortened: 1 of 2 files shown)


FrozenPandaz (approved):

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -57,8 +57,8 @@ export default createESLintRule<Options, MessageIds>({
       },
     ],
     messages: {
-      missingDependency: `The "{{projectName}}" uses the following packages, but they are missing from the "{{section}}":{{packageNames}}`,
-      obsoleteDependency: `The "{{packageName}}" package is not used by "{{projectName}}".`,
+      missingDependency: `The "{{projectName}}" project uses the following packages, but they are missing from the "{{section}}":{{packageNames}}`,
+      obsoleteDependency: `The "{{packageName}}" package is not used by "{{projectName}}" project.`,
```
JamesHenry, line 61: This feels like a bit of a regression to me to be honest.

I think it reads more naturally as either:

(1) by "{{projectName}}"
(2) by the "{{projectName}}" project

I would vote for (2) as it matches up with "The "{{projectName}}" project" at the start of `missingDependency`.

The current sentence sounds a little Slavic :)

PR #18486, opened 2023-08-04
chore(testing): try catch the pm command detection

mandarini:
In CI (observed in vite-ecosystem-ci) when trying to do `yarn -v` it says `Usage Error: This project is configured to use pnpm`. So adding a try catch saves us from this error.

### DIFF
--- a/e2e/utils/command-utils.ts
+++ b/e2e/utils/command-utils.ts
@@ -131,7 +131,7 @@ export function getPackageManagerCommand({
   return {
     npm: {
       createWorkspace: `npx ${
-        +npmMajorVersion >= 7 ? '--yes' : ''
+        npmMajorVersion && +npmMajorVersion >= 7 ? '--yes' : ''
       } create-nx-workspace@${publishedVersion}`,
       run: (script: string, args: string) => `npm run ${script} -- ${args}`,
       runNx: `npx nx`,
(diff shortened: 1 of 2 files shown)


e2e/utils/command-utils.ts
```diff
@@ -120,8 +120,8 @@ export function getPackageManagerCommand({
   list: string;
   runLerna: string;
 } {
-  const npmMajorVersion = getNpmMajorVersion();
-  const yarnMajorVersion = getYarnMajorVersion(path);
+  const npmMajorVersion = getNpmMajorVersion() ?? -1;
+  const yarnMajorVersion = getYarnMajorVersion(path) ?? -1;
```
mandarini, line 124: tbh I just added the `?? -1` because I was not sure what's the best approach here. Like, maybe if no major version detected, do not return the `npm` object ?

meeroslav: I don't think it matters, because if npm or yarn are not available, your install would fail if that package manager was detected based on the lock file.

meeroslav: I would leave those two lines as they were

meeroslav (approved):
Looks ok, apart from the unnecessary `?? -1`

e2e/utils/command-utils.ts
mandarini: I added a check below! :D

PR #18513, opened 2023-08-07
feat(angular): update module federation node

Coly010:
## Current Behavior

We're currently multiple versions behind the latest @module-federation/node dep

## Expected Behavior

Update to latest dep

## Related Issue(s)

Fixes #


packages/angular/src/utils/versions.ts
```diff
@@ -14,7 +14,7 @@ export const corsVersion = '~2.8.5';
 export const typesCorsVersion = '~2.8.5';
 export const expressVersion = '~4.18.2';
 export const typesExpressVersion = '4.17.14';
-export const moduleFederationNodeVersion = '~0.10.1';
+export const moduleFederationNodeVersion = '~1.0.5';
```
leosvelperez, line 17: Could this version be propagated to older versions of Angular? Or is it dependent on the Angular version? If we can use it with 14 and 15, could you update the version for those Angular versions as well?

Coly010: Yes it should be, good point

leosvelperez (approved):

PR #18535, opened 2023-08-08
feat(testing): Vite 5 compatibility

benmccann:
## Current Behavior
Currently doing a `require` of `vite`

## Expected Behavior
Do an `import` of `vite` to use the ESM version as the CJS version is likely to be dropped or cause a warning in the next release


barbados-clemens (approved):

benmccann:
thank you for the help!

PR #18536, opened 2023-08-08
feat(nx-dev): split devkit reference page

isaacplmann:
- Splits the devkit reference page into many pages (one page per object)
- Some types that were previously listed only as `Object` are now actually documented
- No redirects needed since the entry point pages remain the same
- The pages for individual types and functions are not listed in `map.json` so we don't have to manually update that file any time the code changes.  The link checker was updated to account for that

Question:
The `ngcli_adapter` items are nested under `ngcli_adapter` in the breadcrumbs.  Should I do the same thing for devkit?  This will add a redirect, but that's not a big deal.  I only hesitated to do that because the breadcrumb would be `devkit > documents > devkit > someFunction` which looks weird.


FrozenPandaz:
https://nx-dev-git-fork-isaacplmann-docs-split-devkit-reference-nrwl.vercel.app/packages/devkit/documents/ExecutorContext#target

Is it possible to get the type information in the table of contents as well? I think it would be helpful to see all the types in a single fold. Most of them are hopefully self explanatory once the reader knows the type so they don't have to click to find out the type of the property.

juristr (approved):

juristr (approved):
Much better than before. I agree with Jason's comment, maybe enriching the TOC with some further info could be helpful.

Also, I get this is ref docs, but I'm wondering whether it would make sense to categorize the TOC not based on the type (which I understand can be done automatically right now), but rather categorize by when you'd need them. Like working with files, project config, templating,... Just a thought 🤔

isaacplmann:
I'll merge this and make a follow up PR with improvements

PR #18550, opened 2023-08-09
docs: update unkown local cache with double sentences

Jefiozie:
At the moment there is a double sentence in this part of the documentation. I've removed it so that it is better readble.


FrozenPandaz (approved):

PR #18557, opened 2023-08-09
fix(testing): fix playwrigtht test with skipInstall

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/playwright/src/executors/playwright/playwright.ts
+++ b/packages/playwright/src/executors/playwright/playwright.ts
@@ -89,11 +89,15 @@ export async function playwrightExecutor(
   });
 }
 
+const nxOptions = ['skipInstall'];
 function createArgs(opts: PlaywrightExecutorSchema): string[] {
   const args: string[] = [];
 
   for (const key in opts) {
     const value = opts[key];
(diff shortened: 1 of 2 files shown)


packages/playwright/src/executors/playwright/playwright.ts
```diff
@@ -89,11 +89,15 @@ export async function playwrightExecutor(
...
 
+const nxOptions = ['skipInstall'];
 function createArgs(opts: PlaywrightExecutorSchema): string[] {
   const args: string[] = [];
 
   for (const key in opts) {
     const value = opts[key];
+    if (nxOptions.includes(key)) {
+      continue;
+    }
```
barbados-clemens, line 100: let's pass this as an option in the function instead and just set the defaults

```suggestion
function createArgs(opts: PlaywrightExecutorSchema, exclude = ['skipInstall']): string[] {
  const args: string[] = [];

  for (const key in opts) {
    const value = opts[key];
    if (exclude.includes(key)) {
      continue;
    }
```

barbados-clemens (commented):
I just into this myself, but you beat me to it in a PR 😆

PR #18563, opened 2023-08-10
feat(nx-dev): store user feedback in supabase too

mandarini:
Store user feedback in supabase as well as ga. In GA we just store super simple things (good/bad feedback and user queries). In Supabase we store all the information.

@bcabanes @isaacplmann I thought of this as a better way to store the info we need, since GA has its limitations.

### DIFF
--- a/nx-dev/data-access-ai/src/lib/data-access-ai.ts
+++ b/nx-dev/data-access-ai/src/lib/data-access-ai.ts
@@ -1,7 +1,11 @@
 // based on:
 // https://github.com/supabase-community/nextjs-openai-doc-search/blob/main/pages/api/vector-search.ts
 
-import { createClient } from '@supabase/supabase-js';
+import {
+  PostgrestSingleResponse,
+  SupabaseClient,
+  createClient,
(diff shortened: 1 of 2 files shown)


nx-dev/data-access-ai/src/lib/data-access-ai.ts
```diff
@@ -54,6 +60,14 @@ export async function nxDevDataAccessAi(
   sources: { heading: string; url: string }[];
   sourcesMarkdown: string;
 }> {
+  if (!supabaseClient) {
+    console.log('Creating supabase client');
```
isaacplmann, line 64: This is probably not supposed to be here.

isaacplmann (changes requested):
An extra console.log.

nx-dev/data-access-ai/src/lib/data-access-ai.ts
mandarini: Fixed!

isaacplmann (approved):

PR #18593, opened 2023-08-11
feat(nx-cloud): add styles and cleanup to ai feed container

bcabanes:
It adds some styling to the AI page, feed and prompts component when querying the bot. There is still some issues we would need to address before releasing.

The whole communication between the feed and the api needs to be refactored to handle metadata. We basically need 3 things:
- `chatItem` like object
- `metadata` like object
- link from `chatItem` to `metadata` with correlation id so we can split things and send relevant data.

We should also break the `data-access-ai` library into dedicated library _(e.g:the Supabase client should not be in there)_.

UI is rudimentary but kind of works for now.


FrozenPandaz:
image
 I get the feedback buttons before any answer is given. It should be attached to answers.
 
image
I would like if the disclaimer stayed with the answer.

FrozenPandaz (approved):

isaacplmann (approved):

PR #18601, opened 2023-08-12
fix(linter): Normalize paths when in getSourceFilePath

jogelin:
## Current Behavior
When running the linting rule `@nx/dependency-checks` for a project on a Windows machine, it doesn't fail even if the list of packages is not valid.

The `projectPath` provided to the utility function `getSourceFilePath` in `packages/eslint-plugin/src/utils/runtime-lint-utils.ts` is not normalized:

image

## Expected Behavior
The linting rule `@nx/dependency-checks` should work in the same way on Windows or UNIX paths.

## Fixes
To prevent future issues, I decided to make the `getSourceFilePath` compatible with any type of path instead of fixing the `@nx/dependency-checks` code.

Tests are covering that utility function `@nx/dependency-checks`

### DIFF
--- a/packages/eslint-plugin/src/utils/runtime-lint-utils.spec.ts
+++ b/packages/eslint-plugin/src/utils/runtime-lint-utils.spec.ts
@@ -9,6 +9,7 @@ import {
   DepConstraint,
   findConstraintsFor,
   findTransitiveExternalDependencies,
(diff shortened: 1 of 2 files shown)


meeroslav (approved):
LGTM!

PR #18607, opened 2023-08-14
feat(angular): use helper to determine project name and root directory in project generators

leosvelperez:
Updates the Angular plugin project generators (application, host, remote and library) to use the helper to determine the project name and root.

For more context on the changes, see: https://github.com/nrwl/nx/pull/18420

## Related Issue(s)

Fixes #


e2e/angular-core/src/projects.test.ts
```diff
@@ -350,4 +350,55 @@ describe('Angular Projects', () => {
...
+    // check build works
+    expect(runCLI(`build ${scopedLib}`)).toContain(
+      `Successfully ran target build for project ${scopedLib}`
+    );
+    // check tests pass
+    const testResult = runCLI(`test ${scopedLib}`);
+    expect(testResult).toContain(
+      `Successfully ran target test for project ${scopedLib}`
+    );
+  }, 500_000);
```
FrozenPandaz, line 403: I don't think we need specific tests for this in every single suite.

Could we instead change the generate commands to use the `as-provided` style?

FrozenPandaz (dismissed):

packages/angular/src/generators/library/lib/normalize-options.ts
```diff
@@ -74,13 +63,14 @@ export function normalizeOptions(host: Tree, schema: Schema): NormalizedSchema {
...
-    projectDirectory: fullProjectDirectory,
     modulePath,
     parsedTags,
     fileName,
     importPath,
     ngCliSchematicLibRoot,
-    standaloneComponentName: `${names(name).className}Component`,
+    standaloneComponentName: `${
+      names(projectNames.projectSimpleName).className
+    }Component`,
```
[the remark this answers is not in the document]
leosvelperez: Not sure what you mean. The result is no different. The passing of the unchanged tests proves that.

packages/angular/src/generators/library/library.spec.ts
```diff
@@ -571,7 +571,7 @@ describe('lib', () => {
     it('should accept numbers in the path', async () => {
       await runLibraryGeneratorWithOpts({ directory: 'src/1-api' });
 
-      expect(readProjectConfiguration(tree, 'src-api-my-lib').root).toEqual(
+      expect(readProjectConfiguration(tree, 'src-1-api-my-lib').root).toEqual(
```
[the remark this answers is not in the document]
leosvelperez: We used to have a restriction on numbers which no longer applies. This applied only to Angular projects, but it doesn't anymore.

e2e/angular-core/src/projects.test.ts
leosvelperez: As discussed offline, we'll keep this way for the time being. We need to check the correctness of both formats. In a follow-up PR, we'll switch all tests to use the new format and have a test to validate the legacy one (the opposite of what we have in this PR).

packages/angular/src/generators/library/library.spec.ts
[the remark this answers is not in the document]
leosvelperez: As discussed offline, we'll keep this way for the time being. We need to check the correctness of both formats. In a follow-up PR, we'll switch all tests to use the new format and have a test to validate the legacy one.

packages/angular/src/generators/application/lib/normalized-schema.ts
```diff
@@ -11,6 +11,4 @@ export interface NormalizedSchema extends Schema {
   e2eProjectName: string;
   e2eProjectRoot: string;
   parsedTags: string[];
-  ngCliSchematicAppRoot: string;
-  ngCliSchematicE2ERoot: string;
```
leosvelperez, line 15: These were no longer needed. It's a leftover of invoking the Angular CLI schematic, which we removed some time ago.

packages/angular/src/generators/application/lib/add-e2e.ts
```diff
@@ -10,19 +10,18 @@ import {
 } from '@nx/devkit';
 import { nxVersion } from '../../../utils/versions';
 import type { NormalizedSchema } from './normalized-schema';
-import { removeScaffoldedE2e } from './remove-scaffolded-e2e';
 import { cypressProjectGenerator } from '@nx/cypress';
 
 export async function addE2e(tree: Tree, options: NormalizedSchema) {
-  removeScaffoldedE2e(tree, options, options.ngCliSchematicE2ERoot);
```
leosvelperez, line 17: This was no longer needed. It's a leftover of invoking the Angular CLI schematic, which we removed some time ago.

Coly010 (approved):

meeroslav (approved):
Docs changes look ok

AgentEnder (approved):

PR #18622, opened 2023-08-14
feat(core): replace dotenv with dotenv-expand

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/bin/nx.ts
+++ b/packages/nx/bin/nx.ts
@@ -5,6 +5,7 @@ import {
 } from '../src/utils/find-workspace-root';
 import * as chalk from 'chalk';
 import { config as loadDotEnvFile } from 'dotenv';
+import { expand } from 'dotenv-expand';
 import { initLocal } from './init-local';
 import { output } from '../src/utils/output';
 import {
@@ -109,9 +110,10 @@ function main() {
  */
 function loadDotEnvFiles() {
   for (const file of ['.local.env', '.env.local', '.env']) {
-    loadDotEnvFile({
+    const myEnv = loadDotEnvFile({
       path: file,
     });
+    expand(myEnv);
   }
 }
 
--- a/e2e/nx-misc/src/extras.test.ts
+++ b/e2e/nx-misc/src/extras.test.ts
@@ -273,4 +273,51 @@ describe('Extra Nx Misc Tests', () => {
       expect(output).not.toContain('Installed');
     });
   });
+
+  describe('Env File', () => {
+    it('should have the right env', () => {
+      const appName = uniq('app');
+      runCLI(
+        `generate @nx/react:app ${appName} --style=css --bundler=webpack --no-interactive`
+      );
+      updateFile(
+        '.env',
+        `FIRSTNAME="firstname"
+  LASTNAME="lastname"
+  NX_USERNAME=$FIRSTNAME $LASTNAME`
+      );
+      updateFile(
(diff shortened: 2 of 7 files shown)


packages/webpack/package.json
```diff
@@ -38,7 +38,7 @@
     "copy-webpack-plugin": "^10.2.4",
     "css-loader": "^6.4.0",
     "css-minimizer-webpack-plugin": "^5.0.0",
-    "dotenv": "~16.3.1",
```
FrozenPandaz, line 41: I thought we still need dotenv

packages/vite/package.json
```diff
@@ -31,7 +31,7 @@
   "dependencies": {
     "@phenomnomnominal/tsquery": "~5.0.1",
     "@swc/helpers": "~0.5.0",
-    "dotenv": "~16.3.1",
```
FrozenPandaz, line 34: I thought we still need dotenv

packages/storybook/package.json
```diff
@@ -30,7 +30,7 @@
     "migrations": "./migrations.json"
   },
   "dependencies": {
-    "dotenv": "~16.3.1",
```
FrozenPandaz, line 33: I thought we still need dotenv

packages/webpack/src/executors/webpack/webpack.impl.ts
```diff
@@ -1,4 +1,4 @@
-import 'dotenv/config';
+import 'dotenv-expand/config';
```
FrozenPandaz, line 1: What does this do that the `forked-process-task-runner` doesn't already do?

packages/nx/bin/nx.ts
```diff
@@ -31,7 +32,8 @@ function main() {
   require('nx/src/utils/perf-logging');
 
   performance.mark('loading dotenv files:start');
-  loadDotEnvFiles();
```
FrozenPandaz, line 34: I don't think this has to change? :thinking: This is our own function

packages/nx/bin/nx.ts
```diff
@@ -109,9 +111,10 @@ function main() {
  */
 function loadDotEnvFiles() {
   for (const file of ['.env', '.local.env', '.env.local']) {
-    loadDotEnvFile({
+    var myEnv = loadDotEnvFile({
```
FrozenPandaz, line 114:
```suggestion
    let env = loadDotEnvFile({
```

FrozenPandaz (changes requested):
There should be 1 more important usage in `forked-process-task-runner`

packages/webpack/package.json
xiongemi: create a separate pr for this, no loner needed in the package.json

packages/vite/package.json
xiongemi: create a separate pr for this, no loner needed in the package.json

packages/nx/bin/nx.ts
xiongemi: typo, changed back

packages/nx/bin/nx.ts
xiongemi: changed

packages/storybook/package.json
xiongemi: create a separate pr for this, no loner needed in the package.json

FrozenPandaz (approved):

PR #18634, opened 2023-08-15
feat(bundling): update esbuild version

Coly010:
## Current Behavior

esbuild version being used is a few versions behind the latest version

## Expected Behavior

upgrade esbuild to latest version

## Related Issue(s)

Fixes

### DIFF
--- a/packages/angular/migrations.json
+++ b/packages/angular/migrations.json
@@ -1345,6 +1345,15 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "16.8.0": {
+      "version": "16.8.0-beta.2",
+      "packages": {
+        "esbuild": {
+          "version": "^0.19.2",
+          "alwaysAddToPackageJson": false
+        }
+      }
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     "dotenv-expand": "^10.0.0",
     "ejs": "^3.1.7",
     "enhanced-resolve": "^5.8.3",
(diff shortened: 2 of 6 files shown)


packages/angular/migrations.json
```diff
@@ -1303,6 +1303,10 @@
         "ng-packagr": {
           "version": "~16.2.0",
           "alwaysAddToPackageJson": false
+        },
+        "esbuild": {
+          "version": "^0.19.2",
+          "alwaysAddToPackageJson": false
         }
```
leosvelperez, line 1310: This should go in a separate package update targeting a future version. Those who are already in 16.7.0 wouldn't get this package update.

leosvelperez (changes requested):

packages/angular/migrations.json
Coly010: oh yes, good catch, this PR had been created before the release of 16.7.0

packages/angular/migrations.json
```diff
@@ -1305,6 +1305,15 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "16.8.0": {
+      "version": "16.8.0-beta.0",
```
leosvelperez, line 1310: Maybe check with @FrozenPandaz if this is intended to be released with the next minor or the next patch version.

leosvelperez (approved):

jaysoo (approved):

AgentEnder (approved):

PR #18636, opened 2023-08-15
fix(core): handle symlinks on debian for watch events

### LINKED ISSUE #18218: Native file watcher ignores files when .gitignore is present
### Current Behavior

In a pnpm workspace where package `app` depends on package `lib` and a `.gitignore` file is present, updates to files in the `lib` package are being ignored by the native file watcher whereas updates to files in the `app` package are correctly reported.

This results in false cache hits, because the nx daemon is never notified about the file update and thus never updates the hash of that file.

As soon as the `.gitignore` file is deleted and the file watcher restarted, changes to all files are being reported correctly.

I stripped the reproduction repo down to the bare minimum so it's easier to see the actual issue. Unfortunately the native file watcher is a rust binary and there is no easy way to debug it.

This is currently an issue on WSL2. On Windows it is working correctly and I haven't been able to test on a Mac yet.

### Expected Behavior

All file updates are being picked up by the file watcher.

### GitHub Repo

https://github.com/mzoellner/nx-watcher-repro

### Steps to Reproduce

# Install packages

- Run `pnpm install`

# Run watcher

- Run `pnpm run watcher`

# Broken state

- Modify `lib.ts` --> No Output in console
- Modify `app.ts` --> Output in console

# Fix it

- Delete empty .gitignore
- Restart watcher
- Modify `lib.ts` --> Output in console

### Nx Report

```shell
Node   : 18.16.0
OS     : linux-x64
pnpm   : 8.6.6

nx          : 16.5.3
@nrwl/tao   : 16.5.3
typescript  : 5.1.6
```

### Failure Logs

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [X] Other (Please specify)

### Additional Information
EDIT:
- It does not work on WSL2
- It works on Windows and Mac

Cammisuli:
## Current Behavior
When running a yarn/pnpm workspace dependant libraries are symlinked in node_modules. On debian, when watch events are triggered from these dependent libraries, the symlink path is used (ie, `workspace/pkgs/a/node_modules/pkgs/b/lib.ts`) instead of their real path (`workspace/pkgs/b/lib.ts`). Because we implicity ignore all events from node_modules, these events never get pass the filter and we are never notified. 

## Expected Behavior
Paths are canonicalize to their real paths (only on debian)

## Related Issue(s)

Fixes #18218

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1353,6 +1353,7 @@ dependencies = [
  "napi-build",
  "napi-derive",
  "once_cell",
+ "os_type",
  "rayon",
  "regex",
  "swc_common",
@@ -1377,6 +1378,15 @@ name = "once_cell"
 version = "1.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dd8b5dd2ae5ed71462c540258bedcb51965123ad7e7ccf4b9a8cafaa4a63576d"
+
+[[package]]
+name = "os_type"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e24d44c0eea30167516ed8f6daca4b5e3eebcde1bde1e4e6e08b809fb02c7ba5"
+dependencies = [
+ "regex",
+]
 
 [[package]]
 name = "overload"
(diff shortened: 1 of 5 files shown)


FrozenPandaz:
We will also need this for arch apparently, but we can do it as a followup

FrozenPandaz (approved):

PR #18647, opened 2023-08-16
feat(repo): add pm version field to issue template

meeroslav:
When `nx report` fails, we don't know which package manager (or version) the user is using. This is often essential to debugging why things are not working.

## Related Issue(s)

Fixes #


FrozenPandaz:
Why does `nx report` fail? Can we handle it more gracefully?

Cammisuli (approved):

PR #18671, opened 2023-08-17
fix(core): ensure preinstall is only run once on repo

meeroslav:
## Current Behavior
The preinstall step runs every time we call `installPackagesTask` (on every generator and migration that added packages)

## Expected Behavior
The preinstall step should only run once, during the CNW to ensure install is run with the proper pm version.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/tasks/install-packages-task.ts
+++ b/packages/devkit/src/tasks/install-packages-task.ts
@@ -4,7 +4,6 @@ import { requireNx } from '../../nx';
 
 import type { Tree } from 'nx/src/generators/tree';
 import type { PackageManager } from 'nx/src/utils/package-manager';
-import { existsSync } from 'fs';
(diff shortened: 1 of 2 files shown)


packages/devkit/src/tasks/install-packages-task.ts
```diff
@@ -42,7 +43,9 @@ export function installPackagesTask(
       cwd: join(tree.root, cwd),
       stdio: process.env.NX_GENERATE_QUIET === 'true' ? 'ignore' : 'inherit',
     };
-    if (pmc.preInstall) {
+    // only run preinstall if install hasn't been run yet
+    // this only happens during CNW when preset is adding its dependencies
```
FrozenPandaz, line 47: We can move this step out of devkit and into the `new` generator.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #18680, opened 2023-08-17
fix(js): fail silently on missing .modules.yaml file

fahslaj:
## Current Behavior

Nx throws an error if `.modules.yaml` is missing in `node_modules`.

## Expected Behavior

Nx fails silently if `.modules.yaml` is missing in `node_modules`.

## Related Issue(s)

Fixes #
https://github.com/lerna/lerna/issues/3807


meeroslav (approved):
Should we have a logged message when this happens though?

Otherwise I'm fine with swallowing the error.

AgentEnder (approved):

meeroslav (changes requested):
Looking now at the code we already catch all errors so this doesn't solve anything https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/lock-file/lock-file.ts#L104-L112

PR #18708, opened 2023-08-18
feat(core): add NX_IGNORE_CYCLES environment variable

timfisher:
- Add NX_IGNORE_CYCLES environment variable to make it easier to run build commands if there are multiple commands where you need to ignore circular dependency errors, to save having to pass it through each time.


Cammisuli:
You can rebase/merge master and it should fix those storybook e2es.

AgentEnder (approved):

isaacplmann (approved):

PR #18714, opened 2023-08-18
docs(core): reduce repetitive configuration recipe

isaacplmann:
Adds a recipe - Reduce Repetitive Configuration

### DIFF
--- a/docs/shared/recipes/reduce-repetitive-configuration.md
+++ b/docs/shared/recipes/reduce-repetitive-configuration.md
@@ -1,0 +1,255 @@
+# Reduce Repetitive Configuration
+
+Nx can help you dramatically reduce the lines of configuration code that you need to maintain.
+
+Lets say you have three libraries in your repository - `lib1`, `lib2` and `lib3`. The folder structure looks like this:
+
+```treeview
+repo/
+├── libs/
+│   └── lib1/
+│   │   ├── tsconfig.lib.json
+│   │   └── project.json
+│   └── lib2/
+│   │   ├── tsconfig.lib.json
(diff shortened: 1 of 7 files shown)


docs/shared/recipes/reduce-repetitive-configuration.md
```diff
@@ -0,0 +1,255 @@
...
+│       ├── tsconfig.lib.json
+│       └── project.json
+└── nx.json
+```
+
+## Initial Configuration Settings
+
+All three libraries have a similar project configuration. Here is what their `project.json` files look like:
+
+```json {% fileName="libs/lib1/project.json" %}
```
juristr, line 26: Maybe we can use the "tabs" component here? Just to reduce the overall length of the doc and make it easier to parse. You could also highlight which sections we are talking about in particular. Because you can really only reduce the `options` prop here. Using the new highlight feature you could emphasize that in the code potentially.

docs/shared/recipes/reduce-repetitive-configuration.md
```diff
@@ -0,0 +1,255 @@
...
+          "ci": true,
+          "codeCoverage": true
+        }
+      }
+    }
+  }
+}
+```
+
+```json {% fileName="libs/lib1/project.json" %}
```
juristr, line 200: Again, would use the tabs and add a short sentence before this code blocks to mention that you can now reduce the `project.json` configs of the various projects as follows (or something similar)...just to add some separation in between.

juristr (changes requested):

juristr (approved):

PR #18736, opened 2023-08-21
fix(angular): add @angular/pwa to migrations #18731

### LINKED ISSUE #18731: Upgrade @angular/pwa in migrate command
## Description
When I run `yarn nx migrate latest` in a _Nx_ workspace containing an _Angular_ project, all dependencies are updated in the `package.json` except `@angular/pwa`.

## Motivation
It is often forgotten that `@angular/pwa` also needs to be updated to be compatible with `@angular/core`.

Coly010:
## Current Behavior

We have missed migrating the `@angular/pwa` package when migrating angular packages

## Expected Behavior

We include the `@angular/pwa` package when migrating angular packages

## Related Issue(s)

Fixes #18731

### DIFF
--- a/packages/angular/migrations.json
+++ b/packages/angular/migrations.json
@@ -325,6 +325,10 @@
           "version": "~14.0.0",
           "alwaysAddToPackageJson": false
         },
+        "@angular/pwa": {
+          "version": "~14.0.0",
+          "alwaysAddToPackageJson": false
+        },
         "@angular/core": {
           "version": "~14.0.0",
           "alwaysAddToPackageJson": true
@@ -542,6 +546,10 @@
           "version": "~14.1.0",
           "alwaysAddToPackageJson": false
         },
+        "@angular/pwa": {
+          "version": "~14.1.0",
+          "alwaysAddToPackageJson": false
+        },
         "@schematics/angular": {
           "version": "~14.1.0",
           "alwaysAddToPackageJson": false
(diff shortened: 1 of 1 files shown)


packages/angular/migrations.json
```diff
@@ -1305,6 +1341,19 @@
...
+      "x-prompt": "Do you want to update the Angular version to v16.2?",
+      "requires": {
+        "@angular/core": ">=16.1.0 <16.2.0"
+      },
+      "packages": {
+        "@angular/pwa": {
+          "version": "~16.2.0",
+          "alwaysAddToPackageJson": false
+        }
+      }
```
leosvelperez, line 1356: We can't do this. We already have a package update group for Angular 16.2. We must not duplicate the prompt. Folks already in 16.2 can get a prompt asking them again if they want to update to 16.2.

Please include this package update in that one. We already missed all these updates in previous versions. Given the lack of previous feedback, the usage should be pretty low and/or anyone who had that issue already manually updated it. Anyone in the future coming to the latest version will get the update correctly.

leosvelperez (changes requested):

packages/angular/migrations.json
Coly010: That's fine, i wasn't sure on the best way to handle it for this recent version.

It makes sense to me to add it to the previous one, and we can advise to use `--from --to` syntax with migrate

FrozenPandaz (approved):

leosvelperez (approved):

PR #18747, opened 2023-08-21
feat(nx-dev): move openai call to edge function

mandarini:
Moved the OpenAI API calls into a Next.js Edge function, so that the API key is not revealed to the client!

test it here: https://nx-dev-git-fork-mandarini-feat-move-openai-call-to-edge-nrwl.vercel.app/ai

### DIFF
--- a/nx-dev/data-access-ai/src/lib/data-access-ai.ts
+++ b/nx-dev/data-access-ai/src/lib/data-access-ai.ts
@@ -7,13 +7,7 @@ import {
   createClient,
 } from '@supabase/supabase-js';
 import GPT3Tokenizer from 'gpt3-tokenizer';
-import {
-  Configuration,
-  OpenAIApi,
-  CreateModerationResponse,
-  CreateEmbeddingResponse,
-  CreateCompletionResponseUsage,
-} from 'openai';
+import { CreateEmbeddingResponse, CreateCompletionResponseUsage } from 'openai';
 import {
   ApplicationError,
   ChatItem,
@@ -23,6 +17,7 @@ import {
   getListOfSources,
   getMessageFromResponse,
   initializeChat,
+  openAiCall,
   sanitizeLinksInResponse,
   toMarkdownList,
 } from './utils';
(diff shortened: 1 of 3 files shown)


nx-dev/data-access-ai/src/lib/data-access-ai.ts
```diff
@@ -72,18 +61,26 @@ export async function queryAi(
...
     if (!query) {
       throw new UserError('Missing query in request data');
     }
 
     // Moderate the content to comply with OpenAI T&C
     const sanitizedQuery = query.trim();
-    const moderationResponse: CreateModerationResponse = await openai
-      .createModeration({ input: sanitizedQuery })
-      .then((res) => res.data);
+    const moderationResponseObj = await fetch('/api/openai-handler', {
```
ndcunningham, line 72: These API calls could be in a helper function that you call here it's small nit-pick though

ndcunningham (dismissed):

ndcunningham (commented):
```
{
    "code": "42501",
    "details": null,
    "hint": null,
    "message": "new row violates row-level security policy for table \"user_queries\""
}
```
I got this error when running the preview. 
Is this something we should handle?

ndcunningham (commented):
Do we need to handle supabase calls as well?
Screenshot 2023-08-23 at 9 32 09 AM

nx-dev/data-access-ai/src/lib/data-access-ai.ts
mandarini: Done! :D

mandarini:
@ndcunningham RE: supabase: I don't think we need to handle the Supabase stuff, because @FrozenPandaz added the anon key there, which is ok to be revealed?

ndcunningham (approved):

nx-dev/nx-dev/pages/api/openai-handler.ts
FrozenPandaz, line 1: Change this handler to query-ai-handler

FrozenPandaz (approved):

nx-dev/nx-dev/pages/api/openai-handler.ts
mandarini: Will do this in followup PR

nx-dev/data-access-ai/src/lib/data-access-ai.ts
```diff
@@ -72,18 +62,20 @@ export async function queryAi(
   }
```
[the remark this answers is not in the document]
mandarini: Will do this in followup PR

isaacplmann (approved):

PR #18769, opened 2023-08-22
fix(linter): enable analyzeSourceFile as needed when generating new lint project

### LINKED ISSUE #18704: @nx/dependency-checks doesn't find dependencies used by library
### Current Behavior

After creating new publishable library with `@nx/js` generator, bundled with `tsc` (checked also with `rollup` bundler). The automatically added rule `@nx/dependency-checks` doesn't seem to work.

There are 2 unexpected behaviors of this rule:
- It doesn't seem to notice that used `axios` dependency is missing in `package.json` file of the project
- After manually adding it there it complains about unused dependency in `package.json`
 
 I have created a minial repro here - https://github.com/rafal2228/nx-dependency-checks-min-repro

### Expected Behavior

- Using dependency within the library should notify user that it's missing in `package.json`
- Adding dependency manually to `package.json` should not trigger an error

### GitHub Repo

https://github.com/rafal2228/nx-dependency-checks-min-repro

### Steps to Reproduce

1. Pull provided repro
2. Try running `lint` task of `test` package

### Nx Report

```shell

   Node   : 18.16.0
   OS     : linux-x64
   npm    : 9.5.1
   
   nx (global)        : 16.7.1
   nx                 : 16.7.1
   @nx/js             : 16.7.1
   @nx/jest           : 16.7.1
   @nx/linter         : 16.7.1
   @nx/workspace      : 16.7.1
   @nx/devkit         : 16.7.1
   @nx/eslint-plugin  : 16.7.1
   @nrwl/tao          : 16.7.1
   typescript         : 5.1.6
```

### Failure Logs

```shell
5:5  error  The "axios" package is not used by "test" project  @nx/dependency-checks

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

Lint errors found in the listed files.


 —————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Ran target lint for project test (1s)
 
    ✖    1/1 failed
    ✔    0/1 succeeded [0 read from cache]
 

 *  The terminal process "/usr/bin/zsh '-c', 'npx nx run test:lint'" terminated with exit code: 1. 
 *  Terminal will be reused by tasks, press any key to close it.
```

### Operating System

- [X] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

jaysoo:
This PR ensures that JS sources are analyzed when generating a new project that requires `package.json` lint checks.

## Current Behavior
Users coming from package-based repos (e.g. `nx.json` extends `nx/presets/npm.json`), then adding a JS lib will not be able to use the `@nx/dependency-checks` lint rule. This is because we don't pick up dependencies from JS source files.

## Expected Behavior
When adding a new project that uses `@nx/dependency-checks` lint rule, we should enable source file analysis.

## Related Issue(s)

Fixes #18704

### DIFF
--- a/packages/linter/src/generators/lint-project/lint-project.ts
+++ b/packages/linter/src/generators/lint-project/lint-project.ts
@@ -1,8 +1,14 @@
-import type { ProjectConfiguration, Tree } from '@nx/devkit';
+import type {
+  NxJsonConfiguration,
+  ProjectConfiguration,
+  Tree,
+} from '@nx/devkit';
 import {
   formatFiles,
   offsetFromRoot,
+  readJson,
   readProjectConfiguration,
+  updateJson,
   updateProjectConfiguration,
   writeJson,
 } from '@nx/devkit';
@@ -108,6 +114,19 @@ export async function lintProjectGenerator(
     );
   }
 
+  // Buildable libs need source analysis enabled for linting `package.json`.
+  if (
+    isBuildableLibraryProject(projectConfig) &&
+    !isJsAnalyzeSourceFilesEnabled(tree)
+  ) {
+    updateJson(tree, 'nx.json', (json) => {
+      json.pluginsConfig ??= {};
+      json.pluginsConfig['@nx/js'] ??= {};
+      json.pluginsConfig['@nx/js'].analyzeSourceFiles = true;
+      return json;
+    });
+  }
+
   updateProjectConfiguration(tree, options.project, projectConfig);
 
   if (!options.skipFormat) {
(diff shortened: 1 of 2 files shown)


packages/linter/src/generators/lint-project/lint-project.ts
```diff
@@ -177,6 +195,21 @@ function isBuildableLibraryProject(
...
+  };
+
+  if (nxJson.extends === 'nx/presets/npm.json') {
+    // npm preset turns off analyzeSourceFiles, but check if user has overridden it.
+    return jsPluginConfig?.analyzeSourceFiles ?? false;
+  }
+
+  // By default the JS plugin enables analyzeSourceFiles as long as user did not override it.
+  return jsPluginConfig?.analyzeSourceFiles ?? true;
+}
```
meeroslav, line 211: This can be shortened to
```suggestion
  return jsPluginConfig?.analyzeSourceFiles ?? nxJson.extends !== 'nx/presets/npm.json'
```

meeroslav (approved):
Left one minor suggestion but otherwise looks good to me

PR #18770, opened 2023-08-22
fix(react): normalize remote name and directory correctly when using new project root format

leosvelperez:
## Related Issue(s)

Fixes #


packages/react/src/generators/host/lib/normalize-remote.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (options.projectNameAndRootFormat === 'derived' || !options.directory) {
+    return options.directory;
+  }
+
+  /**
+   * With the `as-provided` format, the provided directory would be the root
+   * of the host application. Append the remote name to the host parent
+   * directory to get the remote directory.
+   */
+  return joinPathFragments(dirname(options.directory), remote);
```
jaysoo, line 38: I think more readable as: `joinPathFragments(options.directory, '..', remote)`?

jaysoo (approved):

PR #18796, opened 2023-08-23
feat(core): add migration script to escape $ in env

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.spec.ts
+++ b/packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.spec.ts
@@ -1,0 +1,132 @@
+import { createTreeWithEmptyWorkspace } from '../../generators/testing-utils/create-tree-with-empty-workspace';
+import { addProjectConfiguration } from '../../generators/utils/project-configuration';
+import escapeDollarSignEnvVariables from './escape-dollar-sign-env-variables';
+
+describe('escape $ in env variables', () => {
+  let tree;
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+
+  it('should escape $ in env variables in .env file', () => {
+    tree.write(
+      '.env',
+      `dollar=$
+NX_SOME_VAR=$ABC`
+    );
+    escapeDollarSignEnvVariables(tree);
+    expect(tree.read('.env', 'utf-8')).toEqual(`dollar=\\$
+NX_SOME_VAR=\\$ABC`);
+  });
+
+  it('should escape $ env variables in .env file under project', () => {
+    addProjectConfiguration(tree, 'my-app', {
+      root: 'apps/my-app',
+    });
+    addProjectConfiguration(tree, 'my-app2', {
+      root: 'apps/my-app2',
+    });
+    tree.write(
+      'apps/my-app/.env',
+      `dollar=$
+NX_SOME_VAR=$ABC`
+    );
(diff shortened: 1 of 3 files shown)


packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  it('should escape $ in env variables in .env file', () => {
+    tree.write(
+      '.env',
+      `dollar=$
+NX_SOME_VAR=$ABC`
+    );
+    escapeDollarSignEnvVariables(tree);
+    expect(tree.read('.env', 'utf-8')).toEqual(`dollar=\$
+NX_SOME_VAR=\$ABC`);
```
FrozenPandaz, line 19: You need to escape the `\` in all of these checks.

```suggestion
    expect(tree.read('.env', 'utf-8')).toEqual(`dollar=\\$
NX_SOME_VAR=\\$ABC`);
```

packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.ts
```diff
@@ -0,0 +1,63 @@
...
+            `.env.${configurationName}`,
+            `.${configurationName}.env`
+          );
+        }
+      }
+    }
+
+    for (const envFile of envFiles) {
+      parseEnvFile(tree, `${configuration.root}/${envFile}`); // parse at project root
+      parseEnvFile(tree, envFile); // parse at workspace root
```
FrozenPandaz, line 35: Instead of calling it twice here for each item in the loop, let's just call it once and push the items above.

packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.ts
```diff
@@ -0,0 +1,63 @@
...
+ * - '.env', '.local.env', '.env.local'
+ * - .env.[target-name], .[target-name].env
+ * - .env.[target-name].[configuration-name], .[target-name].[configuration-name].env
+ * - .env.[configuration-name], .[configuration-name].env
+ * at each project root and workspace root
+ * @param tree
+ */
+export default async function escapeDollarSignEnvVariables(tree: Tree) {
+  for (const [_, configuration] of getProjects(tree).entries()) {
+    const envFiles = ['.env', '.local.env', '.env.local'];
```
FrozenPandaz, line 16: These should be pushed once for the root.

FrozenPandaz (changes requested):

packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.ts
xiongemi: changed

packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.ts
xiongemi: changed

packages/nx/src/migrations/update-16-8-0/escape-dollar-sign-env-variables.spec.ts
xiongemi: changed

FrozenPandaz (approved):

PR #18822, opened 2023-08-24
fix(webpack): use new watcher for webpack build coordination plugin

FrozenPandaz:
## Current Behavior

The webpack incremental builds build all of the libs once for every file that changes which can be a lot of files when switching branches.

Also, the webpack incremental builds use `chokidar` to watch for file changes.

## Expected Behavior

The webpack incremental builds only build all of the libs once when many files change at once such as when changing branches.

Also, the webpack incremental builds use the logic from the `nx watch` command and use the daemon file watcher instead of chokidar

## Related Issue(s)

Fixes #


packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
```diff
@@ -42,46 +54,46 @@ export class WebpackNxBuildCoordinationPlugin {
...
+      try {
+        const cp = exec(this.buildCmd);
+
+        cp.stdout.pipe(process.stdout);
+        cp.stderr.pipe(process.stderr);
+        cp.on('exit', () => {
+          res();
+        });
+        cp.on('error', () => {
+          res();
```
Coly010, line 67: Should this be `rej`

FrozenPandaz: Probably, but the original logic noop'd for errors so this basically does the same.

Coly010: No problem!

Cammisuli (approved):

packages/nx/src/command-line/watch/watch.ts
```diff
@@ -32,9 +32,10 @@ class BatchCommandRunner {
...
-    private command: string,
-    private projectNameEnv: string = DEFAULT_PROJECT_NAME_ENV,
-    private fileChangesEnv: string = DEFAULT_FILE_CHANGES_ENV
+    private callback: (
+      projects: Set<string>,
+      files: Set<string>
+    ) => Promise<unknown>
   ) {}
 
   enqueue(projectNames: string[], fileChanges: ChangedFile[]) {
```
jaysoo, line 41: At this point `enqueue` is just `process`, can we just move the body into here?

packages/nx/src/command-line/watch/watch.ts
```diff
@@ -48,60 +49,77 @@ class BatchCommandRunner {
     return this.process();
   }
 
-  async process() {
+  private async process() {
```
jaysoo, line 52: At this point `enqueue` is just `process`, can we move the body into `enqueue` and remove `process`?

jaysoo (approved):

packages/nx/src/command-line/watch/watch.ts
FrozenPandaz: Github hid it but `enqueue` also adds projects and files into pending

PR #18827, opened 2023-08-24
fix(nextjs): improve e2e test by reusing the project

ndcunningham:
What has been done:

1. Split `e2e/next/` into `e2x/next-core/` and `e2e/next-extensions`
2. Inside `next-core/` all tests that require an isolated workspace has been split into their own test file.
3. Update CODEOWNERS to support `e2e/next-core/` and `e2e/next-extensions/`


jaysoo (changes requested):
Some tests won't pass with shared workspace (using `beforeAll`). Can we move those into a separate file or `describe` so they can have their own `beforeAll` scope?

Also we should separate into `next-core` and `next-extensions`. Storybook, component tests, styles, etc. can all go into `next-extensions`.

jaysoo (approved):

FrozenPandaz (approved):

PR #18841, opened 2023-08-25
docs(core): add fastify recipe

isaacplmann:
[Adds a Fastify recipe](https://nx-dev-git-fork-isaacplmann-docs-add-fastify-nrwl.vercel.app/showcase/example-repos/add-fastify)

### DIFF
--- a/docs/generated/manifests/menus.json
+++ b/docs/generated/manifests/menus.json
@@ -4160,6 +4160,14 @@
                 "disableCollapsible": false
               },
               {
+                "name": "Add a Fastify Project",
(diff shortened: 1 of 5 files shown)


juristr (approved):
Just two minor things:

For these tech stacks where we have an option also directly in the `create-nx-workspace`, I feel we should mention that as well. So not just adding it to a new stack, but folks that come through the intro page might also just want to start with a new fastify application.

image

Another thing, maybe we can embed the video somewhere at the top of this recipe:

https://youtu.be/LHLW0b4fr2w?si=TovdohOwEV467lYL

PR #18843, opened 2023-08-25
docs(core): add deno recipe

isaacplmann:
Creates an [Add Deno recipe]()
Moves the 2 deno deploy recipes into a new Deno section in the recipes area along with the Add Deno recipe


juristr (approved):
Potential video we could include. Now this is a longer walkthrough so it might not need to be front and center but more towards the end in the "links" 🤔. 

https://youtu.be/Um8xXR54upQ?si=JLrudcCDR6b3vYKU

PR #18846, opened 2023-08-25
fix(js): add a production-only mode to finding npm dependencies for package.json handilng

jaysoo:
Currently, if `vite.config.ts` or other similar build configuration exists, and user runs `nx lint` with `@nx/dependency-checks`, it will error out on dev-only dependencies. This is because we correctly mark those `vite.config.ts` files as production build files (i.e. they affect the hashing/caching of the build target), but for detecting dependencies it is incorrect.

This PR excludes these configuration files from being read when linting with `@nx/dependency-checks`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/utils/find-npm-dependencies.ts
+++ b/packages/js/src/utils/find-npm-dependencies.ts
@@ -27,6 +27,7 @@ export function findNpmDependencies(
   buildTarget: string,
   options: {
     includeTransitiveDependencies?: boolean;
+    productionOnly?: boolean;
   } = {}
 ): Record<string, string> {
   let seen: null | Set<string> = null;
@@ -48,6 +49,7 @@ export function findNpmDependencies(
       projectGraph,
       projectFileMap,
       buildTarget,
+      options?.productionOnly,
       collectedDeps
     );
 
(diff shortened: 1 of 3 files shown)


mandarini (approved):

packages/js/src/utils/find-npm-dependencies.ts
```diff
@@ -101,12 +104,20 @@ function collectDependenciesFromFileMap(
...
   );
 
+  const excludedFilesRegex = productionOnly
+    ? // For prod-only deps, exclude known build config files as these are not used in prod runtime. e.g. vite.config.ts.
+      new RegExp(
+        joinPathFragments(
+          sourceProject.data.root,
+          // If users customized their names such that they don't match this pattern
+          // then they need to manually add "nx-ignore-next-line" to their build config files.
+          '(package.json|(esbuild|rollup|vite|webpack).config.[jt]s)'
```
meeroslav, line 114: I'm wondering if instead of listing known files we should instead provide a field where the user would specify all the ignored files:
```JSON
{
  "@nx/dependency-checks": [
    "error",
    {
      "buildTargets": ["build", "custom-build"], // add non standard build target names
      "ignoredFiles" ["esbuild.config.js"]
    }
  ]
}
```

We could also support Regex like in the code base.

The default value can be:
```JSONC
[
	 "^(package.json|(esbuild|rollup|vite|webpack).config.[jt]s)$"
]
```

meeroslav (changes requested):
I would make this more flexible by exposing it as an option to user, with suggested regex being the default value.

jaysoo:
Closing this in favour of this https://github.com/nrwl/nx/pull/18863

PR #18863, opened 2023-08-28
feat(linter): add option to ignore files based on pattern

jaysoo:
This PR adds a `ignoredFiles` option to the `@nx/dependency-checks` rule. Useful for build config files that import dev-only dependencies.

The `@nx/js:lib` generator that adds the checks for `package.json` will also ignore build configuration files such that packges imported in them will not count towards production dependencies.

e.g. `vite.config.ts` imports `vite`, which is not a production dependency so should not be reported as a missing dependency.

## Related Issue(s)

Fixes #


ndcunningham (approved):

meeroslav (approved):
Looks good but please add migration as part of this PR (can be done separately as well)

jaysoo:
Doing some local testing before merging.

PR #18867, opened 2023-08-28
docs(core): reference convert-to-monorepo

isaacplmann:
Adds a reference to the `convert-to-monorepo` in the standalone to integrated recipe


juristr (approved):
LGTM

PR #18882, opened 2023-08-29
feat(linter): migrate projects using dep-check lint rule to ignore build config files

jaysoo:
This PR adds a migration to add `ignoredFiles` option to `@nx/dependency-checks` for projects using vite, esbuild, and rollup. We need to ignore config files such as `vite.config.ts` so the packages imported in them are not marked as production dependencies by the lint rule.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.ts
+++ b/packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.ts
@@ -1,0 +1,74 @@
+import { getProjects, Tree } from '@nx/devkit';
+import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';
+import {
+  findEslintFile,
+  lintConfigHasOverride,
+  updateOverrideInLintConfig,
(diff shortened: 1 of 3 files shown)


packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.ts
```diff
@@ -0,0 +1,74 @@
...
+          true
+        )
+      ) {
+        updateOverrideInLintConfig(
+          tree,
+          project.root,
+          (o) => !!o.rules?.['@nx/dependency-checks'],
+          (o) => {
+            const value = o.rules['@nx/dependency-checks'];
+            let ruleSeverity: 0 | 1 | 2 | 'error' | 'warn' | 'off';
```
meeroslav, line 37: This is no longer correct. The current type doesn't support numbers

jaysoo: It was giving me type errors if I don't include them. I can cast as any as well.

meeroslav (approved):

PR #18883, opened 2023-08-29
fix(webpack): WebpackNxBuildCoordinationPlugin cancels inflight builds on new changes detected

Coly010:
## Current Behavior

When changes occur quickly in succession, builds are queued, and processed synchronously.
An action such as changing between two branches quickly could lead to a long wait time as the previous build must be completed before the second build occurs.

Webpack also continues to build the application before the dependent projects have finished building.

## Expected Behavior

Webpack should wait for the dependent projects to finish building before attempting to build the application.

When a build is currently in progress, if more changes are detected, the initial build should be cancelled, allowing the queue to start to process the next items.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
+++ b/packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
@@ -6,6 +6,7 @@ import { output } from 'nx/src/utils/output';
 
 export class WebpackNxBuildCoordinationPlugin {
   private currentlyRunning: 'none' | 'nx-build' | 'webpack-build' = 'none';
+  private buildCmdProcess: ReturnType<typeof exec> | null = null;
 
   constructor(private readonly buildCmd: string, skipInitialBuild?: boolean) {
     if (!skipInitialBuild) {
@@ -30,19 +31,7 @@ export class WebpackNxBuildCoordinationPlugin {
   }
 
   async startWatchingBuildableLibs() {
-    const unregisterFileWatcher = await createFileWatcher(
-      () => this.buildChangedProjects(),
-      () => {
-        output.error({
-          title: 'Watch connection closed',
-          bodyLines: [
-            'The daemon has closed the connection to this watch process.',
-            'Please restart your watch command.',
-          ],
-        });
-        process.exit(1);
-      }
-    );
+    const unregisterFileWatcher = await this.createFileWatcher();
 
     process.on('exit', () => {
       unregisterFileWatcher();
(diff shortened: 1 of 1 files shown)


packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
```diff
@@ -56,44 +45,55 @@ export class WebpackNxBuildCoordinationPlugin {
...
         });
         // eslint-disable-next-line no-empty
       } catch (e) {
         res();
-      } finally {
-        this.currentlyRunning = 'none';
       }
+    }).then(() => {
+      this.currentlyRunning = 'none';
+      this.buildCmdProcess = null;
```
Coly010, line 64: This ensures webpack will wait until the build of dependent projects has completed before continuing the webpack build

packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
```diff
@@ -56,44 +45,55 @@ export class WebpackNxBuildCoordinationPlugin {
     this.currentlyRunning = 'nx-build';
     return new Promise<void>((res) => {
```
jaysoo, line 46: Would changing this to `return await new Promise ...` work without having to put `this.currentlyRunning = 'none';` in the `.then()` callback? I think previously the promise is not being awaited, thus the `finally` is called before the promise resolves.

Generally returning promises should always be awaited on, or else you get problems like this, and also stacktraces are messed up.

packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
```diff
@@ -56,44 +45,55 @@ export class WebpackNxBuildCoordinationPlugin {
...
+              'Please restart your watch command.',
+            ],
+          });
+          process.exit(1);
+        }
+
+        if (this.buildCmdProcess) {
+          this.buildCmdProcess.kill(2);
+          this.buildCmdProcess = null;
+        }
```
Coly010, line 89: If new changes are detected, kill the current process running the build allowing the Promise to resolve, allowing the next item in the queue to be processed.

Consideration: An API for the `BatchFunctionRunner` to clear the queue before queuing a new build. There could be a potential issue where we have 3 builds queued, we detect changes, and we only cancel the first build, but there are still builds queued. 
However, I haven't been able to produce this scenario in testing.

jaysoo (approved):
Approved, but can you check just changing the `return new Promise` to `return await new Promise`?

packages/webpack/src/plugins/webpack-nx-build-coordination-plugin.ts
Coly010: Yeah it makes no difference. 

We can add the `await` anyway if you think it would be better to do so, but the `finally` is happening for the `try-catch-finally` block _within_ the Promise.

The code inside that promise is all non-blocking code, so it all runs through, including the `finally`.

jaysoo: Huh, that's surprising, the `finally` should fire after promise resolves normally, even after the `.then()`. Maybe it's the way the code is being transpiled by tsc? Either case, I guess it's fine in `.then()` if it's working.

jaysoo (approved):

PR #18885, opened 2023-08-29
feat(testing): support cypress v13

barbados-clemens:
- fix(testing): omit indexHtmlFile option for cy >12.17.0
- feat(testing): support cypress v13

## Current Behavior

currently nx/cypress only allows upt to v12 to be installed

## Expected Behavior

nx/cypress supports v13

## Related Issue(s)

Blocked by: #18839


barbados-clemens:
CircleCI is seeing some sha it can't checkout which is probably from me branching not off master and messing up a rebase.

Closing and reopening correctly branching. See #18899

PR #18891, opened 2023-08-29
fix(node): pass projectNameAndRootFormat to js lib generator from nest lib generator

jaysoo:
Chain the `projectNameAndRootFormat` option from `@nx/nest:lib` to `@nx/js:lib`, otherwise the generator errors out when `as-provided` is used. Also updates unit tests to use `projectNameAndRootFormat: 'as-provided'`.

## Related Issue(s)

Fixes #


ndcunningham (approved):
LGTM

PR #18892, opened 2023-08-29
fix(nextjs): Deprecate adding NX_ vars to Webpack Define

### LINKED ISSUE #17940: NX_ environment variables are undefined on a client in Next.js development
### Current Behavior

Environment variables that are prefixed with `NX_` are undefined during development in Next.js on a client. They work as expected on the server's side. It works correctly in production mode though. Both a client and server show correct environment variables.

### Expected Behavior

Environment variables that are prefixed with `NX_` should be defined on a client during development.

### GitHub Repo

https://github.com/skoob13/nx-next-dev-server-bug

### Steps to Reproduce

1. Install dependencies (yarn)
2. Run the development server using `yarn nx run ui:serve`
3. Open `localhost:4200` in a browser
4. Observe that the server outputs the environment variable correctly and the client outputs undefined in the browser's console.

### Nx Report

```shell
Node   : 18.16.0
   OS     : darwin-arm64
   yarn   : 1.22.19

   nx                 : 16.4.3
   @nx/js             : 16.4.3
   @nx/jest           : 16.4.3
   @nx/linter         : 16.4.3
   @nx/workspace      : 16.4.3
   @nx/cypress        : 16.4.3
   @nx/devkit         : 16.4.3
   @nx/eslint-plugin  : 16.4.3
   @nx/next           : 16.4.3
   @nx/react          : 16.4.3
   @nrwl/tao          : 16.4.3
   @nx/web            : 16.4.3
   typescript         : 5.1.6
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I've tested different Next.js versions and environment variables don't work on a client on any 13.4.* version. It works correctly on 13.3.* versions though.

As a temporary solution it's possible to use `env` in next.config.js defining prefixed with `NX_` variables manually.

ndcunningham:
When we added NX_ vars to Webpack Define it would be available to both the client and server. This does not follow Next.js' concept to only add environment variables prefixed with `NEXT_PUBLIC_` to the client.

Now if a `NX_` var was found and it is not a Nx defined variable the console will warn with:
```shell
Warning, in NX 18 environment variables starting with NX_ will not be available in the browser, and currently will not work with @nx/next:server executor.
Please rename the following environment variables: NX_VAR_FOUND using Next.js' built-in support for environment variables. Reference https://nextjs.org/docs/pages/api-reference/next-config-js/env
```

After Nx 18, to get access to environment variables declared in the `.env` file consumers should use the `NEXT_` or `NEXT_PUBLIC_` prefix.

closes: #17940

### DIFF
--- a/packages/next/plugins/with-nx.ts
+++ b/packages/next/plugins/with-nx.ts
@@ -8,6 +8,36 @@ import type { NextBuildBuilderOptions } from '../src/utils/types';
 import type { DependentBuildableProjectNode } from '@nx/js/src/utils/buildable-libs-utils';
 import type { ProjectGraph, ProjectGraphProjectNode, Target } from '@nx/devkit';
 
+const baseNXEnvironmentVariables = [
+  'NX_BASE',
+  'NX_CACHE_DIRECTORY',
+  'NX_CACHE_PROJECT_GRAPH',
+  'NX_DAEMON',
+  'NX_DEFAULT_PROJECT',
+  'NX_HEAD',
+  'NX_PERF_LOGGING',
+  'NX_PROFILE',
+  'NX_PROJECT_GRAPH_CACHE_DIRECTORY',
+  'NX_PROJECT_GRAPH_MAX_WORKERS',
+  'NX_RUNNER',
+  'NX_SKIP_NX_CACHE',
+  'NX_TASKS_RUNNER',
+  'NX_TASKS_RUNNER_DYNAMIC_OUTPUT',
+  'NX_VERBOSE_LOGGING',
+  'NX_DRY_RUN',
+  'NX_INTERACTIVE',
+  'NX_GENERATE_QUIET',
+  'NX_PREFER_TS_NODE',
+  'NX_TASK_TARGET_PROJECT',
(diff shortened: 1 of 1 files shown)


packages/next/plugins/with-nx.ts
```diff
@@ -387,6 +425,15 @@ function addNxEnvVariables(config: any) {
       .forEach(
         ([name, value]) => (maybeDefinePlugin.definitions[name] = value)
       );
+
+    const vars = getNonBaseVariables(env);
+    if (vars.length > 0) {
+      console.warn(
+        `Warning, in NX 18 environment variables starting with NX_ will not be available in the browser, and currently will not work with @nx/next:server executor.\nPlease rename the following environment variables: ${vars.join(
```
jaysoo, line 432: `NX 18` -> `Nx 18` and add a `// TODO(v18)` here

jaysoo (approved):
One comment, otherwise LGTM

PR #18895, opened 2023-08-29
feat(core): support for 3rd party package managers

### LINKED ISSUE #15622: Allow any 3rd party package manager

## Description
Currently NX supports 3 package managers, npm, yarn and pnpm. This is great however, I think we should allow for any package manager but add a warning to the console log saying something like "You are using a unknown / unsupported package manager" 

## Motivation
I think they some great tools out there and NX should be agnostic as possible and focus on what it does the best. This would allow community to build great community plugins. 

I've been looking at bun support and it would allow use to create workspaces with bun install rather than npm first.

## Suggested Implementation
Currently you work on a switch statement, If you try and use one you get the following error
` Argument: packageManager, Given: "bun", Choices: "npm", "pnpm", "yarn"`

I think it should assume the package manager has an install method and  attempt to use package name install 

## Alternate Implementations
Could add another argument like

`npx create-nx-workspace@latest --packageManager bun --packageInstaller install --ignoreScripts`

Jordan-Hall:
## Current Behavior
Currently you are limited to only the packages provided by NX repo and it does not allow for extending without NX support.

Major change is installPackagesTask now returns a promise instead of a void.

## Expected Behavior
You can now create a plugin that manages package. Such as deno, bun and china package managers

## Related Issue(s)
Fixes #15622

### DIFF
--- a/docs/generated/devkit/NxJsonConfiguration.md
+++ b/docs/generated/devkit/NxJsonConfiguration.md
@@ -51,11 +51,11 @@ Default generator collection. It is used when no collection is provided.
 
 #### Type declaration
 
-| Name                  | Type                                                      | Description                                                            |
-| :-------------------- | :-------------------------------------------------------- | :--------------------------------------------------------------------- |
-| `defaultCollection?`  | `string`                                                  | **`Deprecated`** - defaultCollection is deprecated and will be removed |
-| `defaultProjectName?` | `string`                                                  | -                                                                      |
-| `packageManager?`     | [`PackageManager`](../../devkit/documents/PackageManager) | -                                                                      |
+| Name                  | Type     | Description                                                            |
+| :-------------------- | :------- | :--------------------------------------------------------------------- |
+| `defaultCollection?`  | `string` | **`Deprecated`** - defaultCollection is deprecated and will be removed |
+| `defaultProjectName?` | `string` | -                                                                      |
+| `packageManager?`     | `string` | -                                                                      |
 
 ---
 
--- a/docs/generated/devkit/NxPluginV2.md
+++ b/docs/generated/devkit/NxPluginV2.md
@@ -6,8 +6,9 @@ A plugin for Nx which creates nodes and dependencies for the [ProjectGraph](../.
 
 #### Type declaration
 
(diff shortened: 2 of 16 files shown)


michaelangeloio:
+1 on this

ffMathy:
I can't believe no one has looked at this yet 😒

@Jordan-Hall do you think it'll help them merge it if you solved the conflicts?

I hope we'll even hear from the team.

Jordan-Hall:
@ffMathy there was some discussion on the community slack/discord about this. @AgentEnder  has some ideas but it's not high on priority for the team. Don't think it was just ignore the team really does engaged with community

ffMathy:
Could those ideas be shared so we can contribute, and they only have to press the merge button? 😅

Jordan-Hall:
Craigory:

> Yeah, its not something that we've thoroughly discussed. Its not something that super makes sense given that there's not a ton of PM's out and about, and very *very*  few with even 1% of the market share
> Additionally, things that are set on this incorrectly could very easily cause errors on commands like `nx migrate` or when running generators that install packages, which would then get reported as issues and so on
> Further, there's a lot to do with package managers in regards to how we parse lock files to add nodes and deps to the graph... so if these new package managers add a new lockfile type then that needs to be handled as well
> 

Miro:

> We plan to move the entire package manager (including lock file logic) to `@nx/js` package and have it entirely optional, so that would open the door to any new package managers.
> 
> But as craigory mentioned, the new package manager's logic would have to include several tricky implementations - namely lock file parsing and pruning. Also, given the reach these package managers now have making changes to Nx core to support this would have low priority.
> 
> It's an interesting requirement, though. The PR is going in the right direction, but there's way more ground that needs to be covered to support something like this.


I would consider helping with move to js package. That's a good idea tbh. But that requires alot of planning and architecture design which team likely not decided

ffMathy:
But, for now, given that Bun's implementation is identical to Yarn, couldn't we just make an easy fix temporarily while we wait for the rest?

I get the impression that Bun's package manager won't be less than 1% market share for very long.

jaysoo:
Closing this PR as it requires additional design discussion. Added this [new discussion](https://github.com/nrwl/nx/discussions/26358) to keep track of the conversation.

PR #18899, opened 2023-08-29
feat(testing): support cypress v13

barbados-clemens:
## Current Behavior

Nx does not supports cypress v13

## Expected Behavior

Nx supports v13 w/migration

## Related Issue(s)

Fixes #


jordanpowell88 (approved):
LGTM FWIW

packages/cypress/plugins/cypress-preset.ts
```diff
@@ -7,7 +7,6 @@ import vitePreprocessor from '../src/plugins/preprocessor-vite';
 interface BaseCypressPreset {
   videosFolder: string;
   screenshotsFolder: string;
-  video: boolean;
```
ndcunningham, line 10: would this be breaking if you used to use the `video` property?

barbados-clemens: Talked with Jack and we are standardizing on what cypress is doing which is defaulting video off for v13. people can easily turn it back on in their respective cypress.config.ts.

ndcunningham (approved):

isaacplmann (approved):

packages/cypress/src/migrations/update-16-8-0/cypress-13.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+  let tree: Tree;
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace({ layout: 'apps-libs' });
+  });
+
+  it('should update deps to cypress v13', async () => {
+    setup(tree, { name: 'my-app' });
+
+    await updateToCypress13(tree);
+    console.log(readJson(tree, 'package.json'));
```
leosvelperez, line 15:
```suggestion
```

packages/cypress/src/migrations/update-16-8-0/cypress-13.ts
```diff
@@ -0,0 +1,250 @@
...
+    json.devDependencies ??= {};
+    json.devDependencies.cypress = '^13.0.0';
+    return json;
+  });
+
+  await formatFiles(tree);
+
+  return () => {
+    installPackagesTask(tree);
+  };
```
leosvelperez, line 62: You can remove this. The `migrate` command ignores it. It doesn't execute any migration generator callbacks. Instead, it checks whether the `package.json` has changed after all migrations have run, and if so, it performs an install.

leosvelperez (approved):
Just left some non-blocking comments.

FrozenPandaz (approved):

PR #18903, opened 2023-08-30
fix(nx): correct the message for the removal of a project that is still a dependency of other projects.

Teebo:
fix(nx): correct the message for the removal of a project that is still a dependency of other projects.

## Current Behavior

The current message is:  XX is still **depended** on by the following projects:
## Expected Behavior

I think the correct message should be XX is still a dependency of the following projects:
## Related Issue(s)

Fixes #


FrozenPandaz (approved):
That's more clear. Thank you for your contribution! LGTM 🎉

Teebo:
Awesome! Thank you

PR #18914, opened 2023-08-30
chore(core): remove chokidar as a dependency

Cammisuli:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/ng-package.json
+++ b/packages/angular/ng-package.json
@@ -14,7 +14,6 @@
     "@phenomnomnominal/tsquery",
     "@typescript-eslint/",
     "chalk",
-    "chokidar",
     "ignore",
     "minimatch",
     "rxjs-for-await",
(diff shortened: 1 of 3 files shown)


packages/angular/ng-package.json
```diff
@@ -14,7 +14,6 @@
     "@phenomnomnominal/tsquery",
     "@typescript-eslint/",
     "chalk",
-    "chokidar",
```
FrozenPandaz, line 17: Do you know why this was here? Should we also remove `ignore` from here?

FrozenPandaz (approved):

packages/angular/ng-package.json
Cammisuli: the angular package used to have its own file-server that had chokidar and ignore. I can remove it too so that its cleaner.

PR #18919, opened 2023-08-30
feat(core): ensure @nx/js plugin is installed for all JS workspaces

jaysoo:
When running `create-nx-workspace` and choosing either packaged-based or TS integrated monorepo, the user does not get `@nx/js` plugin, which is wrong. In the future, code analysis for TS/JS files will be moved to `@nx/js` from `nx` core, so this plugin has to exist. Furthermore, not installing `@nx/js` automatically means users cannot generate JS libs even though that was the whole point of choosing the TS/JS stack.

This PR also deprecates the `Core` and `Empty` presets. They have both been deprecated for over a year, and we just missed removing them. They are now scheduled for removal in Nx 18.

Screenshot 2023-08-30 at 3 02 26 PM

Lastly, the `@nx/workspace:remove` changes [here](https://github.com/nrwl/nx/commit/70d3728457f6ecf1c764ffef14e4a317e4d941bb) is causing e2e test failure. Not sure how it passed originally, but this is fixed in this PR as well.

### DIFF
--- a/e2e/nx-misc/src/workspace.test.ts
+++ b/e2e/nx-misc/src/workspace.test.ts
@@ -759,7 +759,7 @@ describe('Workspace Tests', () => {
 
       expect(error).toBeDefined();
       expect(error.stdout.toString()).toContain(
-        `${lib1} is still depended on by the following projects`
+        `${lib1} is still a dependency of the following projects`
       );
       expect(error.stdout.toString()).toContain(lib2);
 
--- a/packages/workspace/src/generators/new/generate-preset.ts
+++ b/packages/workspace/src/generators/new/generate-preset.ts
@@ -16,14 +16,6 @@ import * as yargsParser from 'yargs-parser';
 import { spawn, SpawnOptions } from 'child_process';
 
 export function addPresetDependencies(host: Tree, options: NormalizedSchema) {
-  if (
-    options.preset === Preset.Apps ||
-    options.preset === Preset.Core ||
-    options.preset === Preset.Empty ||
-    options.preset === Preset.NPM
-  ) {
-    return;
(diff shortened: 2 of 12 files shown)


packages/workspace/src/generators/new/generate-preset.ts
```diff
@@ -16,12 +16,7 @@ import * as yargsParser from 'yargs-parser';
 import { spawn, SpawnOptions } from 'child_process';
 
 export function addPresetDependencies(host: Tree, options: NormalizedSchema) {
-  if (
-    options.preset === Preset.Apps ||
-    options.preset === Preset.Core ||
-    options.preset === Preset.Empty ||
-    options.preset === Preset.NPM
-  ) {
+  if (options.preset === Preset.Core || options.preset === Preset.Empty) {
```
FrozenPandaz, line 19: If I'm not mistaken, I think core is the same as npm and empty is the same as apps.

If so, these should also move down there. 

Separately, we should clean these up.

FrozenPandaz (changes requested):

e2e/nx-misc/src/workspace.test.ts
```diff
@@ -759,7 +759,7 @@ describe('Workspace Tests', () => {
 
       expect(error).toBeDefined();
       expect(error.stdout.toString()).toContain(
-        `${lib1} is still depended on by the following projects`
```
jaysoo, line 762: This is failing due to changes here https://github.com/nrwl/nx/commit/70d3728457f6ecf1c764ffef14e4a317e4d941bb

FrozenPandaz (approved):

PR #18927, opened 2023-08-31
cleanup(testing): switch cypress tests to use project name and root as-provided format by default

leosvelperez:
## Related Issue(s)

Fixes #


barbados-clemens (approved):
Thanks for picking this up!

PR #18943, opened 2023-08-31
feat(nx-dev): improve link text for ai docs

mandarini:
Create a `longer_heading` option to be used in case of duplicate sources heading.

* Use `longer_heading` if heading is `null`
* Use `longer_heading` if heading is duplicate
* Stop when AI starts generating `### Sources` of its own.

### DIFF
--- a/tools/documentation/create-embeddings/src/main.mts
+++ b/tools/documentation/create-embeddings/src/main.mts
@@ -307,13 +307,19 @@ async function generateEmbeddings() {
 
           const [responseData] = embeddingResponse.data;
 
+          const longer_heading = createLongerHeading(heading, url_partial);
+
           const { error: insertPageSectionError, data: pageSection } =
             await supabaseClient
               .from('nods_page_section')
               .insert({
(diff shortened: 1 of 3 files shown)


tools/documentation/create-embeddings/src/main.mts
```diff
@@ -155,7 +155,7 @@ async function generateEmbeddings() {
     type: 'boolean',
   }).argv;
 
-  const shouldRefresh = argv.refresh;
+  const shouldRefresh = argv.refresh ?? true;
```
ndcunningham, line 158: `!!argv.refresh` nit-pick

mandarini: oooops that should not be there at all!

jaysoo (approved):
I'd prefer the variable casing is consistent. So `url_partial` should be `urlPartial` or `partialUrl`. Otherwise LGTM.

isaacplmann (approved):

PR #18961, opened 2023-09-01
feat(nx-dev): get the correct query for ai feedback

mandarini:
## Current Behavior

We have a list of messages, and each one has an index, which is generated by the loop that renders the messages. The messages are Queries (by the user) and Responses (by the assistant).

The 👍 / 👎 feedback buttons are "attached" to the Response message. So, whenever a user clicks on the feedback button, we know that they clicked on the message with index `X`. Using that index `X` we assume that the Query that resulted in that Response is the message with index `X-1`. Since this logic is based on an assumption that the messages list will always be consistent and "correct" and sorted, we think that it is brittle.

## Expected Behavior

Now that we use the `useChat` from `ai/react` package, we get back a list of messages, and each message has its own `id`. So, when we get a Response, we tie the `id` of the Response with the Query that gave that Response. That way, we can be sure that when a user clicks the feedback button, their feedback will be for the correct Query.

@bcabanes

### DIFF
--- a/nx-dev/feature-ai/src/lib/feed-container.tsx
+++ b/nx-dev/feature-ai/src/lib/feed-container.tsx
@@ -4,10 +4,11 @@ import { ErrorMessage } from './error-message';
 import { Feed } from './feed/feed';
 import { LoadingState } from './loading-state';
 import { Prompt } from './prompt';
-import { ChatItem, extractLinksFromSourcesSection } from '@nx/nx-dev/util-ai';
+import { getQueryFromUid, storeQueryForUid } from '@nx/nx-dev/util-ai';
 import { Message, useChat } from 'ai/react';
 
-const assistantWelcome: ChatItem = {
+const assistantWelcome: Message = {
+  id: 'first-custom-message',
   role: 'assistant',
   content:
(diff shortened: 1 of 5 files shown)


nx-dev/feature-ai/src/lib/feed-container.tsx
```diff
@@ -16,7 +17,8 @@ const assistantWelcome: ChatItem = {
 export function FeedContainer(): JSX.Element {
   const [error, setError] = useState<Error | null>(null);
   const [startedReply, setStartedReply] = useState(false);
-  const [sources, setSources] = useState<string[]>([]);
+  // Do we want to keep the sources in GA?
+  // const [sources, setSources] = useState<string[]>([]);
```
FrozenPandaz, line 21: Mm.. I don't think we need to. At least not for feedback.

FrozenPandaz (approved):

nx-dev/feature-ai/src/lib/feed-container.tsx
mandarini: Ok! I will remove!

mandarini: Done

isaacplmann (approved):

PR #18975, opened 2023-09-01
chore(core): remove fast-glob from core nx

Cammisuli:
## Current Behavior
There was a usage of fast-glob in the daemon code. 

## Expected Behavior

fast-glob is removed for the native function

## Related Issue(s)

Fixes #


packages/nx/src/daemon/server/outputs-tracking.ts
```diff
@@ -59,30 +59,10 @@ export async function outputsHashesMatch(_outputs: string[], hash: string) {
 }
 
 async function normalizeOutputs(outputs: string[]) {
```
FrozenPandaz, line 61: This doesn't seem async anymore. Can we clean this up in a followup PR?

FrozenPandaz (approved):

PR #18996, opened 2023-09-05
feat(webpack): move module federation utils to webpack

Coly010:
## Current Behavior

Our Module Federation utilities currently live in `@nx/devkit`. They were placed here for convenience during the previous effort on MF.

## Expected Behavior

Our Module Federation utilities will live in `@nx/webpack` which makes more sense as we currently, and for the foreseeable future, only support Webpack Module Federation.
This will also remove them from the `@nx/devkit` Public API, usage of which has been deprecated already.

## Related Issue(s)

Fixes #


jaysoo (approved):

AgentEnder (approved):

FrozenPandaz (changes requested):

packages/webpack/src/utils/module-federation/secondary-entry-points.ts
```diff
@@ -2,12 +2,11 @@ import type { WorkspaceLibrary } from './models';
...
 import { existsSync, lstatSync, readdirSync } from 'fs';
-
+import { readJsonFile, joinPathFragments, workspaceRoot } from '@nx/devkit';
 import type { PackageJson } from 'nx/src/utils/package-json';
-import { requireNx } from '../../../nx';
+import { requireNx } from '@nx/devkit/nx';
 
-let { readJsonFile, joinPathFragments, workspaceRoot, readModulePackageJson } =
-  requireNx();
+let { readModulePackageJson } = requireNx();
```
FrozenPandaz, line 9: You changed other imports other than this. Can you change this too in a followup please?

This should come from nx directly.

FrozenPandaz (approved):

PR #18998, opened 2023-09-05
feat(angular): switch default to typescript configuration for module federation

Coly010:
## Current Behavior

We currently generate JS config files for webpack and module federation with the host, remote and setup-mf generators

## Expected Behavior

We should generate TS config files

## Related Issue(s)

Fixes #


packages/angular/src/builders/utilities/wait-for-port.ts
```diff
@@ -0,0 +1,40 @@
+import * as net from 'net';
+
+export function waitForPortOpen(
```
jaysoo, line 3: Should we add `@nx/web` as a dep of `@nx/angular`? We're already installed it for module federation for `@nx/web:file-server`. It's a light package.

packages/angular/src/builders/module-federation-dev-server/module-federation-dev-server.impl.ts
```diff
@@ -107,25 +115,71 @@ export function executeModuleFederationDevServerBuilder(
...
+    }
+    try {
+      await Promise.all(
+        [...remotePorts.values()].map((port) =>
+          // Allow 20 minutes for each remote to start, which is plenty of time but we can tweak it later if needed.
+          // Most remotes should start in under 1 minute.
+          waitForPortOpen(port, {
+            retries: 480,
+            retryDelay: 2500,
+            host: 'localhost',
```
jaysoo, line 165: This goes for the React version as well:

- Is there a potential issue if 'localhost' is started on `::1` (ipv6) but the server is `127.0.0.1` (ipv4)?
- If the project config sets the hostname to something else other than `localhost` this wouldn't work

docs/generated/packages/angular/generators/remote.json
```diff
@@ -163,6 +163,11 @@
         "description": "Whether to configure SSR for the remote application to be consumed by a host application using SSR.",
         "type": "boolean",
         "default": false
+      },
+      "typescriptConfiguration": {
```
meeroslav, line 167: Why is this not called `useTypescriptConfiguration`? 

It feels like the field should point to the configuration file, but it's actually just a boolean flag.

I guess this applies to other fields above (ssr, standalone) as well. If we want to match the same style we could name this one `typescript`.

packages/angular/src/generators/remote/schema.d.ts
```diff
@@ -28,4 +28,5 @@ export interface Schema {
   skipFormat?: boolean;
   standalone?: boolean;
   ssr?: boolean;
+  typescriptConfiguration?: boolean;
```
meeroslav, line 31: Why is this not called `useTypescriptConfiguration`?

It feels like the field should point to the configuration file, but it's just a boolean flag.

I guess this applies to other fields above (`ssr`, `standalone`) as well. If we want to match the same style, we could name this one `typescript`.

packages/angular/src/generators/remote/schema.json
```diff
@@ -166,6 +166,11 @@
       "description": "Whether to configure SSR for the remote application to be consumed by a host application using SSR.",
       "type": "boolean",
       "default": false
+    },
+    "typescriptConfiguration": {
```
meeroslav, line 170: Same as above

packages/angular/src/generators/setup-mf/schema.d.ts
```diff
@@ -14,6 +14,7 @@ export interface Schema {
   prefix?: string;
   standalone?: boolean;
   skipE2E?: boolean;
+  typescriptConfiguration?: boolean;
```
meeroslav, line 17: Same as above

packages/angular/src/generators/setup-mf/schema.json
```diff
@@ -73,6 +73,11 @@
       "type": "boolean",
       "description": "Whether the application is a standalone application. _Note: This is only supported in Angular versions >= 14.1.0_",
       "default": false
+    },
+    "typescriptConfiguration": {
```
meeroslav, line 77: Same as above

meeroslav (commented):
Can we rename `typescriptConfiguration`. It feels misleading.

packages/angular/src/generators/remote/schema.d.ts
Coly010: It's to match the flag on the react generator which is also `typescriptConfiguration`.

This was previously `typescript`, however, in react land it's more confusing because it could be an indication to use TS with your react app, rather than specifically about the configuration file.

packages/angular/src/builders/module-federation-dev-server/module-federation-dev-server.impl.ts
Coly010: This isn't needed anymore thanks to fixing the TS issue by using `registerTsProject`

meeroslav (approved):

mandarini (approved):

packages/angular/src/generators/host/host.ts
```diff
@@ -23,14 +23,19 @@ export async function host(tree: Tree, options: Schema) {
...
-  if (lt(installedAngularVersionInfo.version, '14.1.0') && options.standalone) {
+  if (lt(installedAngularVersionInfo.version, '14.1.0') && schema.standalone) {
     throw new Error(stripIndents`The "standalone" option is only supported in Angular >= 14.1.0. You are currently using ${installedAngularVersionInfo.version}.
     You can resolve this error by removing the "standalone" option or by migrating to Angular 14.1.0.`);
   }
 
+  const { typescriptConfiguration, ...options }: Schema = {
+    ...schema,
+    typescriptConfiguration: schema.typescriptConfiguration ?? true,
+  };
```
leosvelperez, line 37: nit:
```suggestion
  const { typescriptConfiguration = true, ...options }: Schema = schema;
```

packages/angular/src/generators/remote/files/base-ts/src/main.server.ts__tmpl__
leosvelperez, line 1: Same thing about this file.

packages/angular/src/generators/remote/remote.ts
```diff
@@ -21,14 +21,19 @@ export async function remote(tree: Tree, options: Schema) {
...
-  if (lt(installedAngularVersionInfo.version, '14.1.0') && options.standalone) {
+  if (lt(installedAngularVersionInfo.version, '14.1.0') && schema.standalone) {
     throw new Error(stripIndents`The "standalone" option is only supported in Angular >= 14.1.0. You are currently using ${installedAngularVersionInfo.version}.
     You can resolve this error by removing the "standalone" option or by migrating to Angular 14.1.0.`);
   }
 
+  const { typescriptConfiguration, ...options }: Schema = {
+    ...schema,
+    typescriptConfiguration: schema.typescriptConfiguration ?? true,
+  };
```
leosvelperez, line 35: nit:
```suggestion
  const { typescriptConfiguration = true, ...options }: Schema = schema;
```

packages/angular/src/generators/remote/files/base-ts/src/main.server.ts__tmpl__
Coly010: yeah good point

leosvelperez (approved):

PR #19024, opened 2023-09-06
fix(misc): loader-utils Dependabot Alert

mgithubmessier:
incrementing loader utils so that dependabot alerts are addressed

## Current Behavior
- 13.10.x has a critical dependabot alert around loader-utils versions under <1.4.2

## Expected Behavior
- 13.10.x no longer has a critical dependabot alert around loader-utils since it will be version 1.4.2

## Related Issue(s)
- https://github.com/nrwl/nx/issues/19025

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
     "less": "3.12.2",
     "less-loader": "^10.1.0",
     "license-webpack-plugin": "^4.0.2",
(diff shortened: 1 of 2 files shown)


FrozenPandaz:

I released a `14.8.9` which has this issue fixed.

PR #19026, opened 2023-09-06
fix(linter): handle non-JSON eslintrc files when updating overrides

jaysoo:
The new ESLint overrides update logic only works if the config is:
1. `.esilntrc.json` -- we assume config is JSON when updating old config format
2. `eslint.config.js` -- we support new flat config format, but it is different from existing `.eslintrc.js` format 

For other formats like `.eslintrc.js`, `.eslintrc.yaml`, etc., we should ignore updating the overrides entry since they are not in the expected format .

## Related Issue(s)

Fixes #19020

Issue #19020, opened 2023-09-06
Migration update-16-8-0-add-ignored-files is failing


C0ZEN:
### Current Behavior

Migrating from Nx `16.7` to `16.8`, and running the migration [update-16-8-0-add-ignored-files](https://github.com/nrwl/nx/blob/master/packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.ts) is not working.

The log error is:

```
 >  NX   Failed to run update-16-8-0-add-ignored-files from @nx/linter. This workspace is NOT up to date!
 >  NX   Cannot find apps/{project-name}/.eslintrc.json
```

Note that I don't have an `.eslintrc.json`, but instead a `.eslintrc.js`.
Normally, it should be covered inside the [eslint-file.ts](https://github.com/nrwl/nx/blob/af53c7cfb21c0ebc7be2dab33b712f8d10bc81f7/packages/linter/src/generators/utils/eslint-file.ts#L28-L29), but there is only coverage for the `json` file for this migration 🤔 .

### Expected Behavior

There is no error and the migration run successfully.

### Steps to Reproduce

1.

### Nx Report

```shell
Node   : 18.17.1
   OS     : darwin-arm64
   npm    : 9.6.7
   
   nx                 : 16.8.0
   @nx/js             : 16.8.0
   @nx/linter         : 16.8.0
   @nx/workspace      : 16.8.0
   @nx/cypress        : 16.8.0
   @nx/devkit         : 16.8.0
   @nx/eslint-plugin  : 16.8.0
   @nx/react          : 16.8.0
   @nx/storybook      : 16.8.0
   @nrwl/tao          : 16.8.0
   @nx/vite           : 16.8.0
   @nx/web            : 16.8.0
   nx-cloud           : 16.3.0
   typescript         : 5.1.3
   ---------------------------------------
   Community plugins:
   @c0zen/nx-sonarqube : 2.6.0
   nx-stylelint        : 15.0.0
```

### Failure Logs

```shell
>  NX   Failed to run update-16-8-0-add-ignored-files from @nx/linter. This workspace is NOT up to date!
>  NX   Cannot find apps/{project-name}/.eslintrc.json
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

C0ZEN:
@jaysoo FYI, since you created this migration 🙏🏻

jaysoo:
[...]

PR #19035, opened 2023-09-06
fix(linter): ensure config manipulations are run only if config is supported

meeroslav:
## Current Behavior
We do ad hoc checks for existence of eslint config when we detect error

## Expected Behavior
We should use central `isEslintConfigSupported` function, which will help us to potentially add support for more config types and we would need to change just one function.

E.g. migrating from yaml and classic js is fairly trivial after support of json.

## Related Issue(s)

Related to https://github.com/nrwl/nx/pull/19026 and https://github.com/nrwl/nx/issues/19020

### DIFF
--- a/packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.spec.ts
+++ b/packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.spec.ts
@@ -8,6 +8,7 @@ describe('update-16-8-0-add-ignored-files migration', () => {
 
   beforeEach(() => {
     tree = createTreeWithEmptyWorkspace();
+    tree.write('.eslintrc.json', '{}');
   });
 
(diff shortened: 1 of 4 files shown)


jaysoo (approved):

packages/linter/src/migrations/update-16-8-0-add-ignored-files/update-16-8-0-add-ignored-files.spec.ts
```diff
@@ -8,6 +8,7 @@ describe('update-16-8-0-add-ignored-files migration', () => {
 
   beforeEach(() => {
     tree = createTreeWithEmptyWorkspace();
+    tree.write('.eslintrc.json', '{}');
```
meeroslav, line 11: This is necessary for `isEslintConfigSupported` to work. Normally, the root eslintrc would be there

JamesHenry (approved):

PR #18100, opened 2023-07-13
docs(angular): document PNPM workaround for ngcc

daniel-shuy:
## Current Behavior

Currently, https://nx.dev/recipes/other/setup-incremental-builds-angular#requirements still warns that `ngcc` doesn't support PNPM.

## Expected Behavior

While `ngcc` doesn't support PNPM out-of-the-box, PNPM has since added a [node-linker=hoisted](https://pnpm.io/npmrc#node-linker) mode that creates flat `node_modules` without symlinks, similar to the `node_modules` created by NPM, allowing `ngcc` to work with PNPM (angular/angular#50735).

Unfortunately, using `node-link=hoisted` breaks `@parcel/watcher` that is used by NX (#16319 and parcel-bundler/watcher#142). A workaround is to explicitly declare `node-gyp-build` in `package.json`, e.g.
```jsonc
{
  ...
  "devDependencies": {
    ...
    "node-gyp-build": "4.6.0",
    ...
  }
  ...
}
```

This PR documents the workarounds to use `ngcc` with PNPM.

p.s. Huge thanks to @DzmVasileusky for discovering the workarounds.

PR #19041, opened 2023-09-06
docs(angular): document PNPM workaround for ngcc

isaacplmann:
Fix from #18100

## Current Behavior

Currently, https://nx.dev/recipes/other/setup-incremental-builds-angular#requirements still warns that `ngcc` doesn't support PNPM.

## Expected Behavior

While `ngcc` doesn't support PNPM out-of-the-box, PNPM has since added a [node-linker=hoisted](https://pnpm.io/npmrc#node-linker) mode that creates flat `node_modules` without symlinks, similar to the `node_modules` created by NPM, allowing `ngcc` to work with PNPM (angular/angular#50735).

Unfortunately, using `node-link=hoisted` breaks `@parcel/watcher` that is used by NX (#16319 and parcel-bundler/watcher#142). A workaround is to explicitly declare `node-gyp-build` in `package.json`, e.g.
```jsonc
{
  ...
  "devDependencies": {
    ...
    "node-gyp-build": "4.6.0",
    ...
  }
  ...
}
```

This PR documents the workarounds to use `ngcc` with PNPM.

p.s. Huge thanks to @DzmVasileusky for discovering the workarounds.


juristr (approved):
Should we have this documented in other places too? Like in general for using PNPM with Angular?

PR #19063, opened 2023-09-07
feat(react): Add versioning for workspace libraries

ndcunningham:
This PR grants users the ability to add versioning for their workspace libraries.

At build time inside `withModuleFederation` get each workspace library's version by:

1. Looking at the parent's `package.json` to get the version range
2. If not found in the root we look at the library's `package.json` to get the fixed version
3. If neither is found, we do not do any versioning and the functionality remains as is currently.

If the need arises where you do not want this functionality (since it is the current default) you can override it by using the `shared` API inside of the `module-federation.config.js`.

When your code is built you can check the build artifact's `remoteEntry.js` for your `host`/`remote` and you should see your workspace library registered with a version.

### DIFF
--- a/e2e/react-core/src/react-module-federation.test.ts
+++ b/e2e/react-core/src/react-module-federation.test.ts
@@ -2,12 +2,16 @@ import { stripIndents } from '@nx/devkit';
 import {
   checkFilesExist,
   cleanupProject,
+  killProcessAndPorts,
   newProject,
   readJson,
   runCLI,
   runCLIAsync,
+  runCommandUntil,
+  runE2ETests,
   uniq,
   updateFile,
+  updateJson,
 } from '@nx/e2e/utils';
 import { join } from 'path';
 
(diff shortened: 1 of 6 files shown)


Coly010:
I'm not sure this is the correct approach.

When we run the build of a remote or host, the `ModuleFederationPlugin` will read the `package.json` of the workspace libraries, if it exists, and set the version in its `register` function to that version. If it doesn't exist it sets it to `0.0.0`.

We do set `requiredVersion: false` in the `shareWorkspaceLibraries` config. Instead, we should potentially switch this to be

```ts
requiredVersion: workspaceLib.packageJson?.version ?? false
``` 

Then the user can define a range of versions of the library that the app is compatible with in their `shared` function in the `module-federation.config` if they desire, otherwise, we'll have it pinned to the version of the library at the time the app was built:


```
shared: (libraryName, libraryConfig) => {
  if(libraryName === '@acme/mylib') {
     return {
         ...libraryConfig,
        requiredVersion: "~1.0.0"
  } 
}
```

If their workspace library does not have a package.json file, they can add one with just `"version": "1.0.0"` .

I tested this out locally, and it does work. If the `remote` and the `host` have different versions set in their `shared` function, then they just use their own version, otherwise, it uses a shared version between them.

packages/webpack/src/utils/module-federation/share.ts
```diff
@@ -66,14 +76,59 @@ export function shareWorkspaceLibraries(
...
           ...libraries,
-          [library.name]: { requiredVersion: false, eager },
-        }),
-        {} as Record<string, SharedLibraryConfig>
-      ),
+          [library.name]: {
+            ...(version
+              ? {
+                  requiredVersion: version,
+                  singleton: false,
```
Coly010, line 123: I think we still want this to be true, otherwise each remote will use its own copy of the library, which can be breaking for shared libraries that have shared state between remotes. 

Unless the intention is to force the users to declare explicitly in their module-federation.config which libraries should be singletons.

ndcunningham: yea that's right thanks.

e2e/react-core/src/react-module-federation.test.ts
```diff
@@ -138,6 +140,164 @@ describe('React Module Federation', () => {
...
+      `;
+    });
+
+    const libbuildOutput = runCLI(`run ${remote}:build:production`);
+    expect(libbuildOutput).toContain('Successfully ran target build');
+
+    // read the remoteEntry.js file and check that the version is 0.0.2
+    const remoteEntry = readFile(`dist/${remote}/remoteEntry.js`);
+    pattern = new RegExp(`register\\("@acme/${lib}",\\s*"0\\.0\\.2"`, 'g');
+    expect(remoteEntry).toMatch(pattern);
```
jaysoo, line 298: Would be better to run check the apps using a e2e suite, so there are better guarantees that the apps actually do run and load the right versions. i.e. when the remote is loaded from host, it serves its own version.

packages/webpack/src/utils/module-federation/share.ts
```diff
@@ -11,20 +11,30 @@ import {
...
  * share Nx Workspace Libraries between Hosts and Remotes.
  *
- * @param libraries - The Nx Workspace Libraries to share
+ * @param workspaceLibs - The Nx Workspace Libraries to share
  * @param tsConfigPath - The path to TS Config File that contains the Path Mappings for the Libraries
  */
 export function shareWorkspaceLibraries(
-  libraries: WorkspaceLibrary[],
+  workspaceLibs: WorkspaceLibrary[],
+  project: ProjectGraphProjectNode['data'],
```
jaysoo, line 34: Better to change this to `projectRoot: string` I think. More flexible, and not tied to the project node specifically. Also, can it be made required? Not sure if there is a case where it doesn't need to be passed in.

packages/webpack/src/utils/module-federation/share.ts
```diff
@@ -66,14 +76,59 @@ export function shareWorkspaceLibraries(
...
-        (libraries, library) => ({
+    getLibraries: (eager?: boolean): Record<string, SharedLibraryConfig> => {
+      let pkgJson = null;
+      if (
+        project?.root &&
+        existsSync(
+          joinPathFragments(workspaceRoot, project.root, 'package.json')
+        )
+      ) {
+        pkgJson = readJsonFile(
```
jaysoo, line 87: This can fail if the project does not have `package.json`. We should check if it exists first.

packages/webpack/src/utils/module-federation/share.ts
```diff
@@ -66,14 +76,59 @@ export function shareWorkspaceLibraries(
...
+          joinPathFragments(workspaceRoot, project.root, 'package.json')
+        );
+      }
+      return pathMappings.reduce((libraries, library) => {
+        // Check to see if the library version ranged is declared in the root package.json
+        let version =
+          pkgJson?.dependencies?.[library.name] ??
+          pkgJson?.devDependencies?.[library.name] ??
```
jaysoo, line 95: I wouldn't bother with the `devDependencies`. That's a mistake if they use it. Our `@nx/dependency-checks` lint rule will not support this either.

packages/webpack/src/utils/module-federation/share.ts
```diff
@@ -66,14 +76,59 @@ export function shareWorkspaceLibraries(
...
+        existsSync(
+          joinPathFragments(workspaceRoot, project.root, 'package.json')
+        )
+      ) {
+        pkgJson = readJsonFile(
+          joinPathFragments(workspaceRoot, project.root, 'package.json')
+        );
+      }
+      return pathMappings.reduce((libraries, library) => {
+        // Check to see if the library version ranged is declared in the root package.json
```
jaysoo, line 92: `root package.json` sounds like the workspace's `package.json`, maybe just say `project's pacakge.json` or leave the comment off?

packages/webpack/src/utils/module-federation/share.ts
```diff
@@ -66,14 +76,59 @@ export function shareWorkspaceLibraries(
...
+          );
+
+          const workspacePkgJsonPath = join(workspaceLib.root, 'package.json');
+          if (existsSync(workspacePkgJsonPath)) {
+            const workspacePackageJson: PackageJson =
+              readJsonFile(workspacePkgJsonPath);
+
+            if (!workspacePackageJson) {
+              logger.warn(
+                `Could not find a version for "${library.name}" in the root "package.json"`
```
jaysoo, line 109: I don't think we should print a warning. If the user wants to use singletons without requiredVersion then I think they should be able to. Warnings should let the user know how to remedy the situation, but in this case I think it's a legit use-case.

ndcunningham: This should not be there anymore will be removed

packages/webpack/src/utils/module-federation/share.ts
ndcunningham: This has already been checked above line 81-85
```
if(projectRoot && existsSync(joinPathFragments(workspaceRoot, projectRoot, 'package.json'))
```

packages/webpack/src/utils/module-federation/share.ts
ndcunningham: Will update this thanks

packages/webpack/src/utils/module-federation/share.spec.ts
```diff
@@ -371,6 +373,80 @@ describe('MF Share Utils', () => {
       ).not.toThrow();
     });
   });
+
+  it('should using shared library version from root package.json if available', () => {
```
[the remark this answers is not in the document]
ndcunningham: This is probably a cleanup task. If we are changing over to use this it would make sense to update all the unit test inside `share.spec.ts`

Coly010 (approved):
LGTM! 🎉 

in a follow up task we should add or change the e2e to build the shell, then change the version of the lib, then build the remote.

then we should read the main.js file of the shell to find the version the library is being registered with by MFP, and check it is at the earlier version, and then we should look for the same line in the remote’s remoteEntry.mjs and check its version matches the new version.

jaysoo (approved):

PR #19086, opened 2023-09-08
docs(nx-dev): use correct package name

elevenpassin:
Small correction with executor example

## Current Behavior

The first example about Executors mentions how executors are associated with their target tasks. 

image
firefox_Tm43MVLndQ

In the explanation, the name of the package is mentioned in docs as "@nx/web" but in the JSON example, it's "@nx/webpack".

## Expected Behavior

I was checking the [docs for @nx/webpack](https://nx.dev/packages/webpack/executors/webpack) and it looks like the name should be `@nx/webpack`

## Related Issue(s)

N / A


isaacplmann (approved):

isaacplmann:
Thanks!

PR #19102, opened 2023-09-11
chore(repo): update nightly node versions

meeroslav:
## Current

Node v16 expires soon and v19 has been dead for a while.

Meanwhile, we don't test for v20.

## Expected
Nightly should run on v18 and v20

## Related Issue(s)

Fixes #

### DIFF
--- a/.github/workflows/e2e-matrix.yml
+++ b/.github/workflows/e2e-matrix.yml
@@ -24,15 +24,12 @@ jobs:
           - ubuntu-latest
           - macos-latest
         node_version:
-          - 19
+          - 20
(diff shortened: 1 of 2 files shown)


FrozenPandaz (changes requested):
Can we test 16 and 20?

We still support 16.

I think if something breaks in 18... Hopefully it's caught in 16 or 20 :crossed_fingers:

meeroslav:
We can replace 19 with 20, but leave 16. I'll make that change

FrozenPandaz (approved):

PR #19113, opened 2023-09-11
feat(core): add bun package manager

Jordan-Hall:
Bun uses yarn lock for it's binary file. Running the binary will produce the content of a yarn lock file (v1)

Other option is to use the -y command on add and install. This will create a yarn lock file and then createLockFile can just modify the yarn.lock file instead?

### DIFF
--- a/packages/nx/src/utils/package-manager.ts
+++ b/packages/nx/src/utils/package-manager.ts
@@ -12,7 +12,7 @@ import { readNxJson } from '../config/configuration';
 
 const execAsync = promisify(exec);
 
-export type PackageManager = 'yarn' | 'pnpm' | 'npm';
+export type PackageManager = 'yarn' | 'pnpm' | 'npm' | 'bun';
 
 export interface PackageManagerCommands {
   preInstall?: string;
@@ -37,6 +37,8 @@ export function detectPackageManager(dir: string = ''): PackageManager {
       ? 'yarn'
       : existsSync(join(dir, 'pnpm-lock.yaml'))
       ? 'pnpm'
+      : existsSync(join(dir, 'bun.lockb'))
+      ? 'bun'
       : 'npm')
   );
 }
@@ -111,6 +113,18 @@ export function getPackageManagerCommand(
         exec: 'npx',
         run: (script: string, args: string) => `npm run ${script} -- ${args}`,
         list: 'npm ls',
+      };
+    },
+    bun: () => {
+      return {
+        install: 'bun install',
+        ciInstall: 'bun install --frozen-lockfile',
+        add: 'bun add',
+        addDev: 'bun add -D',
+        rm: 'bun remove',
+        exec: 'bunx',
+        run: (script: string, args: string) => `bun ${script} ${args}`,
+        list: 'bun pm ls',
       };
     },
   };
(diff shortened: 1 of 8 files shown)


alfaproject:
Oh yay, support for bun as package manager would be VERY much appreciated!

packages/nx/src/utils/package-manager.ts
```diff
@@ -113,6 +115,19 @@ export function getPackageManagerCommand(
...
+      return {
+        install: 'bun install',
+        ciInstall: 'bun install --frozen-lockfile',
+        add: 'bun add',
+        addDev: 'bun add -D',
+        rm: 'bun remove',
+        exec: 'bunx',
+        run: (script: string, args: string) => `bun ${script} ${args}`,
+        // list not aviable yet. But cant see wherte its used
+        list: '',
```
DaSchTour, line 128: Isn't this `bun pm ls`?

Jordan-Hall: Thank you! I'll  make the change wasn't aware of that command i did go looking for it

Jordan-Hall:
I believe this is complete now. Using the same changes as a patch file currently with no issues

packages/nx/src/utils/package-manager.ts
```diff
@@ -113,6 +115,19 @@ export function getPackageManagerCommand(
...
+    bun: () => {
+      return {
+        install: 'bun install',
+        ciInstall: 'bun install --frozen-lockfile',
+        add: 'bun add',
+        addDev: 'bun add -D',
+        rm: 'bun remove',
+        exec: 'bunx',
+        run: (script: string, args: string) => `bun ${script} ${args}`,
+        // list not aviable yet. But cant see wherte its used
```
alfaproject, line 127: Outdated comment

packages/nx/src/utils/package-manager.ts
```diff
@@ -314,12 +329,16 @@ export async function packageRegistryView(
...
+  if (pm === 'yarn' || pm === 'bun') {
     /**
      * yarn has `yarn info` but it behaves differently than (p)npm,
      * which makes it's usage unreliable
      *
      * @see https://github.com/nrwl/nx/pull/9667#discussion_r842553994
+     *
+     * inregards to bun. Bun is highly modelled after yarn v1
+     * bun info is a "subcommand reserved for future use by Bun."
+     * currently nothing like npm view/list exists
```
alfaproject, line 341: list exists

Jordan-Hall: I've changed the comment. The behaviour completely different

Jordan-Hall: Added a better comment here npm view and bun pm ls works completely differently. They are likley adding a bun pm view command :) - it was mentioned on discord by the team

Jordan-Hall: Is this new commend any better?

alfaproject: Yup view is not supported. I was only talking about list. LGTM

Jordan-Hall:
Bun 1.0.2 fixes the event loop so now appears to be working (unless they break anything else by mistake)

Jordan-Hall:
I'm closing this PR. Bun is amazing no doubt it maybe great. The team doing amazing things however its still way too young and running of hype. 

Bun package manager is playing especially on NX repo. Using node as runtime 1.0.2 hasn't produce dis folder but claiming it ran ok which will be miss leading.

PR #19018, opened 2023-09-06
feat(vue): init, app, component and lib generators

mandarini:
* Added `init` generator
* Added `app` generator
* Added `lib` generator
* Added `component` generator
* Created `vue` plugin overview page in docs (under nx.dev/packages/vue)

 @jaysoo 

TODO:
- [x] Adjust `vite` generator to accept vue as uiFramework
- [x] See if there's anything extra needed for the target generation (if `vite` accepts `vue` as framework)
- [x] component generator
- [x] refine available options
- [x] make sure all works as expected
- [x] correctly configure tests
- [x] Correctly configure styles

PR #19130, opened 2023-09-12
feat(vue): init, app, component and lib generators

jaysoo:
Same as https://github.com/nrwl/nx/pull/19018

* Added `init` generator
* Added `app` generator
* Added `lib` generator
* Added `component` generator
* Created `vue` plugin overview page in docs (under nx.dev/packages/vue)
* Added `e2e` targets
* nx/vue e2e tests
* Vue preset in `create-nx-workspace`
Screenshot 2023-09-13 at 10 29 31 AM


mandarini (approved):

JamesHenry (changes requested):
I would have instinctively expected to see a `.vue` file override within ESLint configs to only apply vue-specific lint rules to vue files, but that doesn't seem to be the case here?

JamesHenry:
@mandarini @jaysoo and I discussed my linting concerns and they are going to make a few tweaks

isaacplmann (approved):

JamesHenry (approved):
All good from a linting perspective now 👍

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -76,6 +76,32 @@ async function createPreset(tree: Tree, options: Schema) {
...
+      '/vue');
+
+    return vueApplicationGenerator(tree, {
+      name: options.name,
+      directory: '.',
+      projectNameAndRootFormat: 'as-provided',
+      style: options.style,
+      linter: options.linter,
+      rootProject: true,
+      e2eTestRunner: options.e2eTestRunner ?? 'cypress',
```
jaysoo, line 102: Note: Should leave as a follow-up, but we should probably prompt for cypress vs none vs playwright across the board (angular, react, vue). Playwright is faster, and requires less configuration compared to cypress. It's weird that we force cypress still when running CNW.

packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -62,6 +62,16 @@ interface AngularArguments extends BaseArguments {
   e2eTestRunner: 'none' | 'cypress' | 'playwright';
 }
 
+interface VueArguments extends BaseArguments {
+  stack: 'vue';
+  workspaceType: 'standalone' | 'integrated';
+  appName: string;
+  // framework: 'none' | 'nuxt';
+  style: string;
+  // nextAppDir: boolean;
```
jaysoo, line 71: We'll never need this `nextAppDir` since it's specific to Next.js.

packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -62,6 +62,16 @@ interface AngularArguments extends BaseArguments {
   e2eTestRunner: 'none' | 'cypress' | 'playwright';
 }
 
+interface VueArguments extends BaseArguments {
+  stack: 'vue';
+  workspaceType: 'standalone' | 'integrated';
+  appName: string;
+  // framework: 'none' | 'nuxt';
```
jaysoo, line 69: we can leave this off until we integrate `@nx/nuxt` into CNW

packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -589,6 +608,74 @@ async function determineReactOptions(
...
 
+async function determineVueOptions(
+  parsedArgs: yargs.Arguments<VueArguments>
+): Promise<Partial<Arguments>> {
+  let preset: Preset;
+  let style: undefined | string = undefined;
+  let appName: string;
+  let e2eTestRunner: undefined | 'none' | 'cypress' | 'playwright' = undefined;
+
+  if (parsedArgs.preset && parsedArgs.preset !== Preset.Vue) {
```
jaysoo, line 619: I don't think we need `Preset.Vue`. Historically, `--preset=react` and `--preset=angular` were the monorepo presets, which we used everywhere (tutorials, video, etc.), so we had to keep those around.

Since Vue is new, we can just go with `--preset=vue-standalone` and `--preset=vue-monorepo`, which match the new angular/react presets as well.

packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -589,6 +608,74 @@ async function determineReactOptions(
...
+    preset = parsedArgs.preset;
+    if (preset === Preset.VueStandalone || preset === Preset.VueMonorepo) {
```
jaysoo, line 621: `appName` needs to be prompted for `VueMonorepo`, or else it'll have the same name as the workspace.

FrozenPandaz (approved):

PR #19141, opened 2023-09-13
feat(vue): storybook configuration generator for vue

mandarini:
Storybook generators for `@nx/vue`

- `@nx/vue:storybook-configuration`
- `@nx/vue:stories`


isaacplmann (approved):
Docs look good

packages/vue/src/generators/stories/lib/__snapshots__/component-story.spec.ts.snap
```diff
@@ -0,0 +1,115 @@
...
+
+export const Heading: Story = {
+  args: {
+    name:  'name',
+    displayAge:  false,
+    age:  0,
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    expect(canvas.getByText(/Welcome to testUiLib!/gi)).toBeTruthy();
```
yannbf, line 36: Could you add an await expression? If people use Addon-interactions, the fact you await these steps makes them debuggable in the addon panel.

```suggestion
    await expect(canvas.getByText(/Welcome to testUiLib!/gi)).toBeTruthy();
```

packages/vue/src/generators/stories/lib/__snapshots__/component-story.spec.ts.snap
```diff
@@ -0,0 +1,115 @@
...
+import { within } from '@storybook/testing-library';
+import { expect } from '@storybook/jest';
+
+
+const meta: Meta<typeof testUiLib> = {
+  component: testUiLib,
+  title: 'testUiLib',
+};
+export default meta;
+type Story = StoryObj<typeof testUiLib>;
```
yannbf, line 16: If you generate this in Typescript 4.9 and higher, you can refer to meta directly. It makes the Story type more precise, as it takes into account the args which are defined in the meta too:
```suggestion
type Story = StoryObj<typeof meta>;
```


This makes types more strict when users use Typescript 4.9 and higher, with the `satisfies` operator e.g.:

```ts
const meta = {
  title: 'Example/Page',
  component: MyPage,
  render: () => ({
    components: { MyPage },
    template: '<my-page />',
  }),
  parameters: {
    layout: 'fullscreen',
  },
  tags: ['autodocs'],
} satisfies Meta<typeof MyPage>;

export default meta;
type Story = StoryObj<typeof meta>;
```

mandarini: ohhh  interesting, thanks! Is this for Vue only, or for all frameworks? i mean should i make this adjustment to React + Angular generators too?

mandarini: Also, do you think I should change the template to be from the plain 

```
const meta: Meta<typeof <%= componentName %>> = {
  component: <%= componentName %>,
  title: '<%= componentName %>',
};
export default meta;
type Story = StoryObj<typeof meta>;
```

that we now have to be the fuller one that you posted above:

```
const meta = {
  title: '<%= componentName %>',
  component:  <%= componentName %>,
  render: () => ({
    components: {  <%= componentName %> },
    template: '<my-page />',
  }),
  parameters: {
    layout: 'fullscreen',
  },
  tags: ['autodocs'],
} satisfies Meta<typeof  <%= componentName %>>;

export default meta;
type Story = StoryObj<typeof meta>;
```

packages/vue/src/generators/stories/lib/component-story.ts
```diff
@@ -0,0 +1,60 @@
...
+  const componentDirectory = componentFilePath.replace(
+    componentFilePath.slice(componentFilePath.lastIndexOf('/')),
+    ''
+  );
+
+  const componentFileName = componentFilePath
+    .slice(componentFilePath.lastIndexOf('/') + 1)
+    .replace('.vue', '');
+
+  const name = componentFileName;
```
jaysoo, line 40: This variable seems unnecessary?

packages/vue/src/generators/stories/lib/component-story.ts
```diff
@@ -0,0 +1,60 @@
...
+) {
+  if (!tsModule) {
+    tsModule = ensureTypescript();
+  }
+  const proj = getProjects(host).get(project);
+  const sourceRoot = proj.sourceRoot;
+
+  const componentFilePath = joinPathFragments(sourceRoot, componentPath);
+
+  const componentDirectory = componentFilePath.replace(
```
jaysoo, line 31: Both `componentDirectory` and `componentFileName` can be computed with `parse` from `node:path`.

Like

```
const { name: componentFileName, dir: componentDirectory } = parse(componentFilePath);
```

packages/vue/src/generators/stories/lib/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export function createDefautPropsObject(propsObject: {
+  [key: string]: string;
+}): {
+  name: string;
+  defaultValue: any;
+}[] {
+  const props = [];
+  for (const key in propsObject) {
+    if (Object.prototype.hasOwnProperty.call(propsObject, key)) {
```
jaysoo, line 20: Would `for (const key of Object.keys(propsObject))` work better? Just need to enumerable keys from the object right? That way we don't need to check `hasOwnProperty`.

packages/vue/src/generators/stories/stories.lib.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+      appTree.read('test-ui-lib/src/components/test-ui-lib.stories.ts', 'utf-8')
+    ).toMatchSnapshot();
+    expect(
+      appTree.read(
+        'test-ui-lib/src/components/another-cmp/another-cmp.stories.ts',
+        'utf-8'
+      )
+    ).toMatchSnapshot();
+
+    const packageJson = JSON.parse(appTree.read('package.json', 'utf-8'));
```
jaysoo, line 54: `readJson` would be more succinct.

jaysoo (approved):

packages/vue/src/generators/stories/lib/__snapshots__/component-story.spec.ts.snap
yannbf: I'd say so! But @kasperpeulen can tell better as he worked on these types. WDYT Kasper?

kasperpeulen: Yes, for TS 4.9+ using satisfies should be preferred.
However, you don't need a render function anymore in SB 8.0. So I would leave that out.

FrozenPandaz (approved):

PR #19150, opened 2023-09-13
feat(nx-dev): add "new chat" button to AI page

jaysoo:
This PR:

- Adds a `New chat` button to the AI page. It will reset the messages to `[]` so AI will not remember current messages.
- Adds a `Stop generating` button so users can stop the response from streaming in the UI.
- Adds a `Regenerate` button if users press on `Stop generating`.
- Makes the thumbs up/down buttons larger so they are easier to press on mobile.
- Puts the prompt at the bottom so it's consistent with other chat experiences.
- Update URL to `/ai-chat` from `/ai` in case we want the latter in the future for other AI features.

Edit: Updated to include expanded scope

### DIFF
--- a/nx-dev/feature-ai/src/lib/error-message.tsx
+++ b/nx-dev/feature-ai/src/lib/error-message.tsx
@@ -1,9 +1,10 @@
+import { type JSX, memo } from 'react';
 import {
-  XCircleIcon,
   ExclamationTriangleIcon,
+  XCircleIcon,
 } from '@heroicons/react/24/outline';
 
-export function ErrorMessage({ error }: { error: any }): JSX.Element {
+function ErrorMessage({ error }: { error: any }): JSX.Element {
   try {
     if (error.message) {
       error = JSON.parse(error.message);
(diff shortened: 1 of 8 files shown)


FrozenPandaz:
I think "New Chat" down there might be confusing CTA.

I understand you were trying to convey we are starting a new chat now. But a major side effect is that this current chat gets blown away. Someone clicks "New Chat" not realizing their old chat will go away.

I think the following would be better.
1. Move the button to the right. User will type something, and have different ways of submitting it.
2. I think the iconography should be something more "Reset"
3. The behavior I expect, is for the screen to clear. However, I think people will have already typed something. Let's remember what they typed and keep it in the input.

Can we do that? :eyes:

jaysoo:
> Move the button to the right. User will type something, and have different ways of submitting it.

I think having a different way to submitting (i.e. two submit buttons) is confusing too, the user wouldn't know which is the right button when asking questions.

I'm playing around with showing a button above the prompt only when the response has stopped.

mandarini (approved):
Love the new UI!

isaacplmann (approved):

PR #19170, opened 2023-09-14
feat(react-native): add useTransformReactJSX babel config to generators

Tobbb:
## Current Behavior

Jest tests require import React from 'react' even though  this is not required is the component after react 17 

## Expected Behavior

you don't have to **import React from 'react**' in your jest tests. should be more consistent with modern react development
## Related Issue(s)

Fixes #


mandarini (approved):

mandarini:
after confirming with @xiongemi , I am merging this!

PR #19173, opened 2023-09-14
docs(core): angular-monorepo tutorial

isaacplmann:
Adds an angular-monorepo tutorial


docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+```bash
+nx test angular-store # runs the tests for angular-store
+nx lint inventory # runs the linter on inventory
+nx e2e angular-store-e2e # runs e2e tests for the angular-store
+```
+
+More conveniently, we can also run tasks in parallel using the following syntax:
+
+```shell
+nx run-many -t test
```
juristr, line 638: I think there should be multiple targets, right? like `test lint e2e`

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+          }
+        ]
+      }
+    },
+    ...
+  ]
+}
+```
+
+To test it, go to your `libs/products/src/lib/products.module.ts` file and import the `OrderListComponent` from the `orders` project:
```
juristr, line 934: when we fix the standalone lib generation we need to update this one here too

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+
+    ✔    6/7 succeeded [0 read from cache]
+
+    ✖    1/7 targets failed, including the following:
+         - nx run products:lint
+```
+
+If you have the ESLint plugin installed in your IDE you should immediately see an error:
+
+![ESLint module boundary error](/shared/images/tutorial-angular-standalone/angular-standalone-module-boundaries.png)
```
juristr, line 989: this image doesn't seem to load for some reason. seems to be from the angular-standalone tutorial. That shouldn't be a problem, but maybe we should move it to some shared location so it doesn't get deleted sometime in the future by accident.

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+- Learn about popular generators such as [how to setup Tailwind](/recipes/angular/using-tailwind-css-with-angular-projects)
+- Learn how to [migrate your existing Angular CLI repo to Nx](/recipes/angular/migration/angular)
+- [Setup Storybook for our shared UI library](/recipes/storybook/overview-angular)
+- [Speed up CI: Run only tasks for project that got changed](/core-features/run-tasks#run-tasks-affected-by-a-pr)
+- [Speed up CI: Share your cache](/core-features/remote-cache)
+- [Speed up CI: Distribute your tasks across machines](/core-features/distribute-task-execution)
+
+Also, make sure you
+
+- [Join the Nx community Slack](https://go.nrwl.io/join-slack) to ask questions and find out the latest news about Nx.
```
juristr, line 1031: shouldn't be Slack, but Discord now => URL is: http://go.nx.dev/community

(can you fix that in the React monorepo one as well, I missed it there: https://nx.dev/getting-started/tutorials/react-monorepo-tutorial#next-steps)

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+You can choose `github`, `circleci`, `azure`, `bitbucket-pipelines`, or `gitlab` for the `ci` flag.
+{% /callout %}
+
+This will create a default CI configuration that sets up Nx Cloud to [use distributed task execution](/core-features/distribute-task-execution). This automatically runs all tasks on separate machines in parallel wherever possible, without requiring you to manually coordinate copying the output from one machine to another.
+
+## Next Steps
+
+Here's some more things you can dive into next:
+
+- Learn more about the [underlying mental model of Nx](/concepts/mental-model)
```
juristr, line 1021: I'd also add a link to https://nx.dev/concepts/more-concepts/nx-and-angular

juristr (changes requested):
added some comments

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+import { RouterModule } from '@angular/router';
+import { AppComponent } from './app.component';
+import { appRoutes } from './app.routes';
+import { NxWelcomeComponent } from './nx-welcome.component';
+
+@NgModule({
+  declarations: [AppComponent, NxWelcomeComponent],
+  imports: [
+    BrowserModule,
+    RouterModule.forRoot(appRoutes, { initialNavigation: 'enabledBlocking' }),
```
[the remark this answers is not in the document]
isaacplmann: nice idea, but that would be a separate PR

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1048 @@
...
+Note, this tutorial sets up a repo with applications and libraries in their own subfolders. If you are looking for an Angular standalone app setup then check out our [Angular standalone app tutorial](/getting-started/tutorials/angular-standalone-tutorial).
+{% /callout %}
+
+## Nx CLI vs. Angular CLI
+
+Nx evolved from being an extension of the Angular CLI to a [fully standalone CLI working with multiple frameworks](/getting-started/why-nx#how-does-nx-work). As a result, adopting Nx as an Angular user is relatively straightforward. Your existing code, including builders and schematics, will still work as before, but you'll also have access to all the benefits Nx offers.
+
+Advantages of Nx over the Angular CLI:
+
+- [Split a large angular.json into multiple project.json files](/concepts/more-concepts/nx-and-angular#projectjson-vs-angularjson)
```
juristr, line 27: I feel like the advantages should point more towards the monorepo features given we have a monorepo tutorial. like the fact that with Nx you can create monorepos which you cannot really do with the Angular CLI (I mean, it works, but it's not made for that use case); scaling in terms of structuring (project.json files, libs etc), also the whole speed aspect like caching etc, ability to efficiently run multiple commands in a monorepo + CI scaling etc; and clearly modern tools which are integrated in a way to work efficiently in a monorepo setup which otherwise you need to figure out by yourself..

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1034 @@
...
+- better scalability in your teams by allowing different teams to work on separate libraries
+
+### Creating Local Libraries
+
+<!-- {% video-link link="https://youtu.be/OQ-Zc5tcxJE?t=1041" /%} -->
+
+Let's assume our domain areas include `products`, `orders` and some more generic design system components, called `ui`. We can generate a new library for each of these areas using the Angular library generator:
+
+```
+nx g @nx/angular:library products --directory=libs/products
```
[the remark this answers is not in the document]
juristr: 🤔 why's that? I wouldn't go with `NgModules` at all; that's why I mentioned generating these with `--standalone`. It is more future-proof, and I think if you start a new project today you wouldn't go with `NgModules`, which is what our tutorials should reflect. It was the reason I chose that in the Angular single-project workspace tutorial as well.

(means we should also update how we generate the workspace initially, like answer "yes" to the question about standalone there, s.t. the Angular app gets set up properly)

juristr (changes requested):

docs/shared/angular-tutorial/angular-monorepo.md
```diff
@@ -0,0 +1,1032 @@
...
+          }
+        ]
+      }
+    },
+    ...
+  ]
+}
+```
+
+To test it, go to your `libs/products/src/lib/products.module.ts` file and import the `OrderListComponent` from the `orders` project:
```
juristr, line 922: I believe the `products.module.ts` is no more there if we generate the library in standalone mode, right?

juristr (changes requested):

juristr (approved):

PR #19192, opened 2023-09-15
feat(react): Support loading remotes via library: var

ndcunningham:
This PR updates `withModuleFederation` plugin for React apps to now support `library: { type: 'var' }` when loading remotes.

With this change, users can now decorate their hosts and remotes with `library: { type: 'var'}`, enabling the remote to be loaded via script.

To enable this type of loading you would need to update your host and remote similar to this:
```ts
// host module-federation.config.js
module.exports = {
  library: { type: 'var', name: 'host' },
  //... Other host config
};
```

```ts
// remote module-federation.config.js
module.exports = {
  library: { type: 'var', name: 'remote' },
  //... Other remote config
};

```

### DIFF
--- a/e2e/react-core/src/react-module-federation.test.ts
+++ b/e2e/react-core/src/react-module-federation.test.ts
@@ -282,6 +282,98 @@ describe('React Module Federation', () => {
     }
   }, 500_000);
 
+  it('should support host and remote with library type var', async () => {
+    const shell = uniq('shell');
+    const remote = uniq('remote');
+
+    runCLI(
+      `generate @nx/react:host ${shell} --remotes=${remote} --project-name-and-root-format=as-provided --no-interactive`
+    );
+
+    // update host and remote to use library type var
+    updateFile(
+      `${shell}/module-federation.config.ts`,
+      stripIndents`
+      import { ModuleFederationConfig } from '@nx/webpack';
+
+      const config: ModuleFederationConfig = {
+        name: '${shell}',
+        library: { type: 'var', name: '${shell}' },
+        remotes: ['${remote}'],
+      };
+
+      export default config;
+      `
+    );
+
+    updateFile(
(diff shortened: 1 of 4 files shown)


e2e/react-core/src/react-module-federation.test.ts
```diff
@@ -282,6 +282,63 @@ describe('React Module Federation', () => {
     }
   }, 500_000);
 
+  it('should support host and remote with library type var', async () => {
```
Coly010, line 285: Can this test run serve and check that we can navigate correctly to the remote from the host? 

Building should be fine regardless, but we'd get more value from the test if we can ensure it will navigate correctly

ndcunningham: yea I will update the e2e to include navigation

e2e/react-core/src/react-module-federation.test.ts
```diff
@@ -282,6 +282,84 @@ describe('React Module Federation', () => {
...
+    
+      it('should display welcome message', () => {
+        getGreeting().contains('Welcome ${shell}');
+        
+      });
+    
+      it('should navigate to /about from /', () => {
+        cy.get('a').contains('${remote}').click();
+        cy.url().should('include', '/${remote}');
+        getGreeting().contains('Welcome ${remote}');
```
Coly010, line 341: Seems to be an error with this function, do you need to import it?

Coly010 (approved):
Left one comment, but overall looks good!!

Can we also make sure this is added to Angular?

jaysoo (approved):

ndcunningham:
> Left one comment, but overall looks good!!
> 
> Can we also make sure this is added to Angular?

We could add a task for angular all that would be needed is to update the `with-module-federation` and add a e2e

PR #19195, opened 2023-09-15
feat(testing): update playwright version

jaysoo:
Update Playwright and CI image to re-enable previous build errors.

Tested locally with `cimg/rust:1.72.0-browsers` and Playwright 1.38.0.

## Related Issue(s)

Fixes #


jaysoo:
Still not passing :(

I tried the same image in Docker locally and everything works just fine. Does not appear to be a bug in Playwright 1.38.0 but something specific to our CI runs.

PR #19196, opened 2023-09-15
feat(nx-dev): improve auto-scrolling so it does not interfere with users reading the content

jaysoo:
This PR improves auto-scrolling, so we check if the user is close to the bottom of the content before scrolling them down. It prevents problems where users can lose the position of where they are reading.

Drastically improves the mobile experience as well where the screen no longer jumps around all the time when typing or reading.

**Before:**

https://github.com/nrwl/nx/assets/53559/a380cc7e-16a2-4111-a9ea-d7cbff8ca0cc

**After:**

https://github.com/nrwl/nx/assets/53559/0fcbf1a0-ea00-459a-bc52-ba6e58b74721

### DIFF
--- a/nx-dev/feature-ai/src/lib/feed-container.tsx
+++ b/nx-dev/feature-ai/src/lib/feed-container.tsx
@@ -4,7 +4,6 @@ import {
   type JSX,
   RefObject,
   useEffect,
-  useMemo,
   useRef,
   useState,
 } from 'react';
@@ -27,8 +26,6 @@ export function FeedContainer(): JSX.Element {
   const [error, setError] = useState<Error | null>(null);
   const [startedReply, setStartedReply] = useState(false);
   const [isStopped, setStopped] = useState(false);
-
-  const feedContainer: RefObject<HTMLDivElement> | undefined = useRef(null);
 
   const {
     messages,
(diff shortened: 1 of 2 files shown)


nx-dev/feature-ai/src/lib/feed-container.tsx
```diff
@@ -57,14 +54,28 @@ export function FeedContainer(): JSX.Element {
...
-      const elements =
-        feedContainer.current.getElementsByClassName('feed-item');
-      elements[elements.length - 1].scrollIntoView({ behavior: 'smooth' });
+    if (!scrollableWrapperRef.current) return;
+    const el = scrollableWrapperRef.current;
+    let shouldScroll = false;
+    if (messages.length > currentMessagesLength.current) {
+      currentMessagesLength.current = messages.length;
+      shouldScroll = true;
+    } else if (el.scrollTop + el.clientHeight + 50 >= el.scrollHeight) {
```
ndcunningham, line 75: Why 50?

isaacplmann (approved):

nx-dev/feature-ai/src/lib/feed-container.tsx
jaysoo: If it's exactly the `scrollHeight` then it's hard for user to scroll to the bottom fast enough and re-enable auto-scroll. The 50 pixels (about two lines of text) seem to work well enough on both desktop and mobile.

PR #19214, opened 2023-09-18
fix(js): update generated .swcrc file to align with @swc/core@1.3.85

jaysoo:
This PR fixes an issue with `.swcrc` using options that are now invalid in `@swc/core@1.3.85`. When `module.type` is `es6`, the `strict` and `noInterop` options cannot be set.

The point is to transpile the library with interop between CJS and ESM, and tests show that the resulting output are the same regardless of these options (even in `@swc/core@1.3.84`). The following is the expected output, and removing the options don't affect it.

```
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

function foo() {
    return 'foo';
}

exports.foo = foo;

```

See: https://github.com/swc-project/swc/commit/7e8d72d

## Related Issue(s)

Fixes #


ndcunningham (approved):

packages/js/migrations.json
```diff
@@ -47,6 +47,12 @@
       "version": "16.6.0-beta.0",
       "description": "Explicitly set 'updateBuildableProjectDepsInPackageJson' to 'true' in targets that rely on that value as the default.",
       "factory": "./src/migrations/update-16-6-0/explicitly-set-projects-to-update-buildable-deps"
+    },
+    "16-8-2-update-swcrc": {
```
FrozenPandaz, line 51: We also need a migration for package.json update right?

And we also update the version on `js` `versions.ts`

FrozenPandaz (changes requested):

packages/js/migrations.json
jaysoo: Sorta, the version range already installed the latest, which is where the error came from. But we can update it more explicitly.

FrozenPandaz (approved):

PR #19216, opened 2023-09-18
feat(bundling): update vite and esbuild to 4.4.9 and 0.19.3 respectively

jaysoo:
This PR updates Vite and ebuild to their latest versions. We had already updated esbuild versions in `peerDependencies` of `@nx/angular` and `@nx/esbuild`, but did not update the value in `versions.ts` file.

There are no breaking changes that affect our usages.

## Related Issue(s)

Fixes #


jaysoo:
There is deprecated to c8 coverage provider. Need a follow-up to make sure this works.

PR #19218, opened 2023-09-18
feat(repo): remove projectNameAndRoot nx.json option

FrozenPandaz:
## Current Behavior

When creating a new workspace, the behavior when generating projects changed to the new `as-provided` project name and root format.

There's a configuration option which disables the prompt.

## Expected Behavior

When creating a new workspace, there are prompts just like for existing workspaces for the name and root of new projects.

There is no configuration to disable the prompt.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/generators/project-name-and-root-utils.ts
+++ b/packages/devkit/src/generators/project-name-and-root-utils.ts
@@ -77,17 +77,9 @@ export async function determineProjectNameAndRootOptions(
 > {
   validateName(options.name, options.projectNameAndRootFormat);
   const formats = getProjectNameAndRootFormats(tree, options);
-  const configuredDefault = getDefaultProjectNameAndRootFormat(tree);
-
-  if (configuredDefault === 'derived') {
-    logger.warn(
-      deprecationWarning + '\n' + getExample(options.callingGenerator, formats)
-    );
-  }
 
   const format =
     options.projectNameAndRootFormat ??
-    configuredDefault ??
     (await determineFormat(tree, formats, options.callingGenerator));
 
   return {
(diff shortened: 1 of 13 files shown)


packages/devkit/src/generators/project-name-and-root-utils.ts
```diff
@@ -175,38 +167,11 @@ async function determineFormat(
...
-      setProjectNameAndRootFormatDefault(tree);
-    } else {
-      logger.warn(deprecationWarning);
-    }
-  } else {
-    const example = getExample(callingGenerator, formats);
-    logger.warn(deprecationWarning + '\n' + example);
-  }
+  const example = getExample(callingGenerator, formats);
+  logger.warn(deprecationWarning + '\n' + example);
```
leosvelperez, line 171: This should only be shown when `result === 'derived'`, isn't it?

packages/nx/src/migrations/update-16-9-0/remove-project-name-and-root-format.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  updateJson(tree, 'nx.json', (nxJson) => {
+    if (!nxJson.workspaceLayout) {
+      return nxJson;
+    }
+
+    delete nxJson?.workspaceLayout?.projectNameAndRootFormat;
+
+    if (Object.keys(nxJson?.workspaceLayout).length === 0) {
```
leosvelperez, line 17: nit: we already checked `nxJson.workspaceLayout` is defined in line 11.
```suggestion
    delete nxJson.workspaceLayout.projectNameAndRootFormat;

    if (Object.keys(nxJson.workspaceLayout).length === 0) {
```

leosvelperez (changes requested):

AgentEnder (approved):
Seems good to me. Feel free to address Leo's nit first.

packages/nx/src/migrations/update-16-9-0/remove-project-name-and-root-format.ts
FrozenPandaz: This is to delete the entire object from the config if it becomes an empty object.

leosvelperez (approved):

PR #19219, opened 2023-09-18
fix(js): node executor now correctly kills tasks when exiting

LonguCodes:
Previously, the node executor od @nx/js did not kill spawned processes due to not awaiting finishing of killing and infinite loop.
This fix allows for seemless running nx in docker containers with support for signals

## Current Behavior
When killing nx running `@nx/js:node`, child processes are not killed

## Expected Behavior
Child processes should be killed


konrad-szychowiak:
I've been looking for this as well!

jaysoo (approved):

PR #19239, opened 2023-09-19
fix(core): set SWC_NODE_PROJECT before registering `@swc-node/register`

Cammisuli:
## Current Behavior
swc does not use the `tsconfig.base.json` file in integrated repos for use in in-process transpilation

## Expected Behavior
Set `SWC_NODE_PROJECT` to `tsconfig.base.json` if it exists. (Otherwise swc will just use tsconfig.json anyway)

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/plugins/js/utils/register.ts
+++ b/packages/nx/src/plugins/js/utils/register.ts
@@ -1,6 +1,8 @@
 import { dirname, join } from 'path';
 import type { CompilerOptions } from 'typescript';
 import { logger, NX_PREFIX, stripIndent } from '../../../utils/logger';
+import { existsSync } from 'fs';
+import { workspaceRoot } from '../../../utils/workspace-root';
 
 const swcNodeInstalled = packageIsInstalled('@swc-node/register');
 const tsNodeInstalled = packageIsInstalled('ts-node/register');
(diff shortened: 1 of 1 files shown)


jaysoo (approved):

packages/nx/src/plugins/js/utils/register.ts
```diff
@@ -44,6 +46,10 @@ export function getSwcTranspiler(
   const register = require('@swc-node/register/register')
     .register as ISwcRegister;
 
+  if (existsSync(join(workspaceRoot, 'tsconfig.base.json'))) {
+    process.env.SWC_NODE_PROJECT = 'tsconfig.base.json';
```
FrozenPandaz, line 50: Shouldn't this be the full path?

FrozenPandaz (changes requested):

packages/nx/src/plugins/js/utils/register.ts
Cammisuli: No, it works without because swc does a `resolve(tsconfigPath)` https://github.com/swc-project/swc-node/blob/9f674cd67091192b1fe62befd13cf4b61a6377e5/packages/register/read-default-tsconfig.ts#L26

AgentEnder (approved):

FrozenPandaz (approved):

PR #19246, opened 2023-09-19
feat(angular): support `~` path for webpack-dev-server ssl

titan-ron:
## Current Behavior
Today we cannot use `~` as a value on `project.json`/`nx.json` when using `@nx/angular:webpack-dev-server` such as:

```json
"serve": {
  "executor": "@nx/angular:webpack-dev-server",
  "options": {
    ...
    "ssl": true,
    "sslCert": "~/path/to/cert",
    "sslKey": "~/path/to/key"
  }
  ...
}
```

## Expected Behavior
Handle `~` at the `webpack-dev-server` executor.

## Fixes
This PR is intend to replace the value of `~` with `homedir()` value (from `node:os`) when using `webpack-dev-server` executor.


packages/angular/src/builders/webpack-dev-server/lib/normalize-options.ts
```diff
@@ -1,12 +1,17 @@
...
   return {
     host: 'localhost',
     port: 4200,
     liveReload: true,
     open: false,
-    ssl: false,
     ...schema,
+    ssl: ssl ?? false,
+    ...(sslCert ? { sslCert: sslCert.replace('~', homedir()) } : {}),
+    ...(sslKey ? { sslKey: sslKey.replace('~', homedir()) } : {}),
```
titan-ron, line 15: Maybe it should be on when it starts with `~`?

Coly010: Let me discuss this PR internally. 

Usually, we only handle files that are within the workspace, specifying the home directory would involve working with files outside the workspace.

titan-ron: Thanks! Please let me know 🙏🏼 

I will elaborate on our use case:
We are working with HTTPS and each developer generates his own SSL certificate on his machine, then, it is the same for everyone to use `~` (same for other repositories as well and not only the workspace).

What we are doing now, is some kind of overriding the `webpack-dev-server` but it's making some nonstandard scenarios to handle with + maintain.

Supporting the `~` will let us finally use the `nx serve` command properly as a standard and without any inheritance/overriding it :)

Moxita33 (approved):

titan-ron:
Hi guys, @Coly010 @Moxita33

Anything we can do to promote it? Anything to change/update in the PR?

WDYT?

Coly010:
Hi! I'm sorry it has been a while since getting back to this. 

I'm going to close this PR. 
After discussing internally, this feature requires more thought and potential design as it opens the door for potential issues for users. 

Feel free to open a discussion so that it can be discussed further: https://github.com/nrwl/nx/discussions/new/choose

PR #19252, opened 2023-09-20
feat(react): add support for transitive serve of remotes

Coly010:
## Current Behavior

We do not currently support serving remotes from remote applications that are also hosts automatically.

## Expected Behavior

Add transitive serve support for remotes.

This means that if we have a host, that depends on `remoteA` and `remoteA` depends on `remoteB`, we will now also serve `remoteB` when we run `nx serve host`.

The code here is subject to change, especially when we merge parallelized startup of remotes support to Angular.

Another thing to note is that the `checkPortIsActive` _does_ introduce a slowdown to startup when we have remotes that are also hosts, as they are checking to see if their dependent remotes are already built and served.

I'm open to suggestions on how to reduce this slowdown.

cc @jaysoo

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/angular/executors/module-federation-dev-server.json
+++ b/docs/generated/packages/angular/executors/module-federation-dev-server.json
@@ -112,6 +112,16 @@
       "pathToManifestFile": {
         "type": "string",
         "description": "Path to a Module Federation manifest file (e.g. `my/path/to/module-federation.manifest.json`) containing the dynamic remote applications relative to the workspace root."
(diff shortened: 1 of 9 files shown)


jaysoo:
I'm not sure if this is the way to go because, like you said, we are looking into merging static remotes into a single serve. Maybe we can build out the list of remotes in one go from the top-level host?

meeroslav (approved):
Docs look ok

PR #19267, opened 2023-09-21
feat(release): add 'git-tag' currentVersionResolver and conventional commits support

fahslaj:
- Adds `git-tag` as a `currentVersionResolver` option for `nx release version`. This option will cause the current version for the release group to be taken from the most recent matching release tag, as configured with `releaseTagPattern`.

- Adds support for using conventional commits to determine the `specifier` for the version bump. In this case, the `nx release version` command will look at git tags to determine which git ref to use as a base. Commits since the base will be parsed and any commits that touched files within any lib in the release group will be used to help determine the next version. Right now, only `fix(...)` and `feat(...)` are recognized, but breaking changes are recognized (`fix(...)!: ...`). If the previous version (from the git tag) is determined to be a prerelease, the bump type will always be `prerelease`. The user will need to promote prereleases to regular releases manually by passing an explicit specifier, which will override that obtained from conventional commits standards.

- Uses `npm config get registry` to determine the default registry before falling back on https://registry.npmjs.org

- Moves the specifier prompting into the version generator. This allows the current version determined in the generator to be used to determine the specifier when using conventional commits.

### DIFF
--- a/docs/generated/packages/js/generators/release-version.json
+++ b/docs/generated/packages/js/generators/release-version.json
@@ -20,7 +20,17 @@
       },
       "specifier": {
         "type": "string",
-        "description": "Exact version or semver keyword to apply to the selected release group. NOTE: This should be set on the release group level, not the project level."
+        "description": "Exact version or semver keyword to apply to the selected release group. Overrides specifierSource."
+      },
+      "releaseGroupName": {
+        "type": "string",
+        "description": "The name of the release group being versioned in the current execution."
+      },
+      "specifierSource": {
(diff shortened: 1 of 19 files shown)


JamesHenry:
Setting to draft while we merge prerequisite PR and also figure out a couple of things

JamesHenry (approved):
Played with this locally today and found a couple things which we've patched and LGTM

juristr (approved):

mandarini (approved):

PR #19271, opened 2023-09-21
feat(core): allow using environment variable to override parallel arg

igilham:
It would be useful to be able to set the default parallelism via the environment in CI to avoid having to edit scripts that use the `--parallel` argument.

I've not been able to run the tests because the devContainer doesn't work and the build instructions do not allow me to run the build without errors. I'm relying on your CI to see if this works.

Husky also forced me to commit a load of files I didn't touch. I'm not familiar with the usual workflow for this repo so I'm trusting the automation to do the right thing.

## Current Behavior

Nx can be called with the `--parallel` argument with an optional value, which can be a boolean or number. This is used to set the level of parallelism used in the build. The default is hard-coded to 3.

## Expected Behavior

Allow the user to specify `NX_PARALLEL` in the environment variables to override the default parallelism value.

## Related Issue(s)

Fixes #


igilham:
Implemented elsewhere in version 17.0.0

Issue #18641, opened 2023-08-15
Support for @typescript-eslint/parser@v6


lucavb:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md))

## Description

Currently the eslint-package contains this [line](https://github.com/nrwl/nx/blob/master/packages/eslint-plugin/package.json#L27) which prevents us from migrating to [typescript-eslint@^6](https://github.com/typescript-eslint/typescript-eslint). I guess this might make it a breaking change for nx as well so this will only be included in version 17? 

I also tried to find another issue targeting tihs but I was unable to find one. 

## Motivation

Well, pretty much the next major version with improvements being done by them. 

## Suggested Implementation

Updating the referenced line in the package.json and creating a migration.

k3nsei:
Just add this to your package.json
```
"overrides": {
  "@nx/eslint-plugin": {
    "@typescript-eslint/eslint-plugin": "^6.4.0",
    "@typescript-eslint/parser": "^6.4.0",
    "eslint-config-prettier": "^9.0.0"
  },
}
`

meeroslav:
It's already on our roadmap.

@k3nsei, it's not that simple, as it requires some code migrations.

k3nsei:
I don't know that works for me and allowed me to have all packages to their latest versions. I don't saying that this should be solution for NX. I just quick issue solver for @lucavb

meeroslav:
Sure, thank you @k3nsei

JamesHenry:
The main thing we will need to do on the Nx side is update the module resolution configuration for TS: https://github.com/typescript-eslint/typescript-eslint/issues/7284

This is a more significant change than just bumping packages, and we will need to make sure we think through the ramifications for all Nx users

eddiemonge:
Any progress on this? I know people still want to support Node versions older than 18, but 16 was end of life'd a month ago. Node 10 was EOL 3 and a half years ago. If people want to stay on dead Node versions, then they can lock their deps to versions that support that.

meeroslav:
It's not a matter of node version. Our codebase dropped support for pre Node 16 versions some time ago. It's rather the changes this brings. 

Switching to `nodenext` means all CommonJS packages will cease to work and their import will have to be hacked. 

We will likely do this once v17 is out, due to the scope of the changes needed (check draft PR here https://github.com/nrwl/nx/pull/19368)

PR #19272, opened 2023-09-21
feat(linter): update peer dependencies

stefan-schweiger:
## Current Behavior
Currently NX is using outdated versions of `@typescript-eslint/parser` and `eslint-config-prettier` which prevents you from updating other packages which might depend on current versions.

## Expected Behavior
Use latest versions.

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -24,8 +24,8 @@
   },
   "homepage": "https://nx.dev",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


JamesHenry:
Thanks @stefan-schweiger but this is not the simple change you think it is and it has ramifications across multiple areas. Please kindly follow along with https://github.com/nrwl/nx/issues/18641

@meeroslav will be working on this very soon on our side

PR #19286, opened 2023-09-21
feat(react): Add federate-module generator

ndcunningham:
This PR contains

A federate module generator which should do a few things:

Command
```shell
nx generate @nx/react:federate-module Carosel --path=carosel/src/lib/carosel --remote=ui
```

1. It will generate/update a remote adding a path to a module provided to the generator
2. Update if existing the `module-federation.config.js` to add a new entry to the `exposes` property for the new module

### DIFF
--- a/packages/react/src/generators/federate-module/federate-module.spec.ts
+++ b/packages/react/src/generators/federate-module/federate-module.spec.ts
@@ -1,0 +1,101 @@
+import { Tree, getProjects } from '@nx/devkit';
+import { Schema } from './schema';
+import { Schema as remoteSchma } from '../remote/schema';
+import { federateModuleGenerator } from './federate-module';
+import { createTreeWithEmptyWorkspace } from 'nx/src/devkit-testing-exports';
+import { Linter } from '@nx/linter';
+import { remoteGeneratorInternal } from '../remote/remote';
+
+describe('federate-module', () => {
+  let tree: Tree;
+  let schema: Schema = {
+    name: 'my-federated-module',
+    remote: 'my-remote',
+    path: 'my-remote/src/my-federated-module.ts',
+    style: 'css',
+  };
+
+  beforeAll(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+  describe('no remote', () => {
+    it('should generate a remote and e2e', async () => {
+      await federateModuleGenerator(tree, schema);
+
+      const projects = getProjects(tree);
+
+      expect(projects.get('my-remote').root).toEqual('my-remote');
+      expect(projects.get('my-remote-e2e').root).toEqual('my-remote-e2e');
+    });
+
+    it('should contain an entry for the new path for module federation', async () => {
+      await federateModuleGenerator(tree, schema);
+
+      expect(tree.exists('my-remote/module-federation.config.js')).toBe(true);
+
+      const content = tree.read(
+        'my-remote/module-federation.config.js',
+        'utf-8'
+      );
+      expect(content).toContain(
+        `'./my-federated-module': 'my-remote/src/my-federated-module.ts'`
+      );
+
+      const tsconfig = JSON.parse(tree.read('tsconfig.base.json', 'utf-8'));
+      expect(
+        tsconfig.compilerOptions.paths['my-remote/my-federated-module']
+      ).toEqual(['my-remote/src/my-federated-module.ts']);
+    });
+  });
+
+  describe('with remote', () => {
+    let remoteSchema: remoteSchma = {
+      name: 'my-remote',
+      e2eTestRunner: 'none',
+      skipFormat: false,
+      linter: Linter.EsLint,
+      style: 'css',
+      unitTestRunner: 'none',
+    };
+
+    beforeEach(async () => {
+      remoteSchema.name = uniq('remote');
+      await remoteGeneratorInternal(tree, remoteSchema);
+    });
+
+    it('should append the new path to the module federation config', async () => {
+      let content = tree.read(
+        `${remoteSchema.name}/module-federation.config.js`,
+        'utf-8'
+      );
+
+      expect(content).not.toContain(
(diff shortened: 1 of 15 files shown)


isaacplmann (approved):

packages/react/src/generators/federate-module/federate-module.spec.ts
```diff
@@ -0,0 +1,6 @@
+// TODO
+describe('federate-module', () => {
```
Coly010, line 2: Can we write some unit tests? That would be great!

packages/react/src/generators/federate-module/lib/utils.ts
```diff
@@ -0,0 +1,268 @@
...
+ * TODO (nicholas): remove this before merge
+ * @param tree The workspace tree
+ * @param remoteName The name of the remote to check
+ * @returns Remote ProjectConfig if it exists, false otherwise
+ */
+export async function checkRemoteExists(tree: Tree, remoteName: string) {
+  const remote = await getRemote(remoteName);
+  if (!remote) return false;
+  let hasModuleFederationConfig = false;
+  if (tree.exists(join(remote.root, 'module-federation.config.ts'))) {
```
Coly010, line 66: We should use `joinPathFragments` instead of `join` in generator code to account for windows. Our virtual FS uses unix style paths, but on windows, join will create windows style paths.

packages/react/src/generators/federate-module/lib/utils.ts
```diff
@@ -0,0 +1,268 @@
...
+    hasModuleFederationConfig = true;
+  } else {
+    hasModuleFederationConfig = tree.exists(
+      join(remote.root, 'module-federation.config.js')
```
Coly010, line 70: Same as above

Coly010 (changes requested):
Looks great!

i think there are a couple of things missing.

When we add the federated module, we should also add a path alias to the root tsconfig file following the same format.

`”remote-app/MyModule”: [“./path/to/file”]`

The only other things I’d love to have relate to testing.

1. can we test an integrated repo, where the `lib` is a separate project? The remote could be in apps/ but the file we want to federate is in libs/ - perhaps we should update the description on the `path` flag in the schema to mention workspace root.
You can see an example of where I’ve done this in an Angular app: https://github.com/Coly010/mf-federated-service-example/blob/main/apps/auth-remote/module-federation.config.js

2. It would be great if our e2es tested federating a path, and then having that path be consumed in the host, and ensuring that build works and when served, the content of the federated module renders correctly.

i think the e2es you have currently which check if the files exist etc could become unit tests.

but overall this is awesome! 🎉 great work

packages/react/src/generators/federate-module/schema.json
```diff
@@ -0,0 +1,116 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "cli": "nx",
+  "$id": "NxReactFederateModule",
+  "title": "Federate Module",
+  "description": "Create a federated module, which can be loaded by a remote host.",
+  "examples": [
+    {
+      "command": "nx g federate-module MyModule --path=./src/component/my-cmp.ts --remote=my-remote-app",
+      "description": "Create a federated module called my-remote-app, that exposes my-cmp from ./src/component/my-cmp.ts. called Mymodule."
```
Coly010, line 10:
```suggestion
      "description": "Create a federated module called my-remote-app, that exposes my-cmp from ./src/component/my-cmp.ts. called MyModule."
```

Coly010 (approved):
One typo, but otherwise this looks great! 🎉 

This is an exciting feature to have!

packages/react/src/generators/federate-module/schema.json
```diff
@@ -0,0 +1,116 @@
...
+  "properties": {
+    "name": {
+      "description": "The name of the module.",
+      "type": "string",
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      },
+      "x-prompt": "What name would you like to use for the module?",
+      "pattern": "^[a-zA-Z][^:]*$"
```
xiongemi, line 23: should it have ` "x-priority": "important"`?

packages/react/src/generators/federate-module/federate-module.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    tasks.push(remoteGenerator);
+  }
+
+  const projectRoot = remote ? remote.root : remoteRoot;
+  const remoteName = remote ? remote.name : projectName;
+
+  // add path to exposes property
+  addPathToExposes(tree, projectRoot, schema.name, schema.path);
```
xiongemi, line 49: just want to double check, remote could be optional, but path should be required in the schema, right?

packages/react/src/generators/federate-module/lib/utils.ts
```diff
@@ -0,0 +1,157 @@
+import { Tree, getProjects, joinPathFragments } from '@nx/devkit';
+
+import { ensureTypescript } from '@nx/js/src/utils/typescript/ensure-typescript';
+import * as ts from 'typescript';
+
+let tsModule: typeof import('typescript');
+
+if (!tsModule) {
+  tsModule = ensureTypescript();
+}
```
xiongemi, line 10: i think the import on the top `import * as ts from 'typescript';` already assumes typescript is loaded, then no need to call `ensureTypescript()`. i think we could remove tsModule variable and use ts directly.

ndcunningham: If it is not loaded wouldn't it fail?
I will switch it to use types from the import `import * as from `typescript` and rename the other functional calls to use `tsModule` instead

packages/react/src/generators/federate-module/federate-module.ts
ndcunningham: They are both required. I added them to `required: []` property inside `schema.json`

xiongemi (approved):

PR #19299, opened 2023-09-22
fix(core): do not attempt to publish private npm packages

JamesHenry:
## Current Behavior

We attempt to publish all packages in our npm implementation of a release-publish executor.

## Expected Behavior

We attempt to publish all _public_ packages in our npm implementation of a release-publish executor, but skip any marked as private.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/release/src/private-js-packages.test.ts
+++ b/e2e/release/src/private-js-packages.test.ts
@@ -14,6 +14,7 @@ expect.addSnapshotSerializer({
       str
         // Remove all output unique to specific projects to ensure deterministic snapshots
         .replace(tmpProjPath(), '')
+        .replace('/private', '')
         .replaceAll(/public-pkg-\d+/g, '{public-project-name}')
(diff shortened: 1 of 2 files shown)


fahslaj (approved):

e2e/release/src/private-js-packages.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        .map((r) => r.trim())
+        .join('\n')
+    );
+  },
+  test(val: string) {
+    return val != null && typeof val === 'string';
+  },
+});
+
+describe('nx release - private JS packages', () => {
```
FrozenPandaz, line 40: Is it possible to keep this in 1 file so that we don't have to create a new workspace for this test?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

jaysoo (approved):

PR #19338, opened 2023-09-26
fix(nx): fix Sentry example in plugin docs

lforst:
I noticed that the current Sentry example in the plugin documentation is a bit wrong. This PR fixes that.

### DIFF
--- a/docs/shared/packages/next/next-config-setup.md
+++ b/docs/shared/packages/next/next-config-setup.md
@@ -125,7 +125,8 @@ module.exports = async (phase, context) => {
 
   // If you have plugins that has to be added after Nx you can do that here.
   // For example, Sentry needs to be added last.
-  updatedConfig = require('@sentry/nextjs')(updatedConfig, { silent: true });
+  const { withSentryConfig } = require('@sentry/nextjs');
+  updatedConfig = withSentryConfig(updatedConfig, { silent: true });
 
   return updatedConfig;
 };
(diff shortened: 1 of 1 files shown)


docs/shared/packages/next/next-config-setup.md
```diff
@@ -125,7 +125,8 @@ module.exports = async (phase, context) => {
 
   // If you have plugins that has to be added after Nx you can do that here.
   // For example, Sentry needs to be added last.
-  updatedConfig = require('@sentry/nextjs')(updatedConfig, { silent: true });
+  const { withSentryConfig } = require('@sentry/nextjs');
```
lforst, line 128: I would have put this line in the top scope but I assume you probably don't want that.

docs/shared/packages/next/next-config-setup.md
```diff
@@ -125,7 +125,8 @@ module.exports = async (phase, context) => {
...
+  updatedConfig = withSentryConfig(updatedConfig, { silent: true });
```
lforst, line 129: I would also suggest turning `silent` off because it mutes useful debugging information but I will leave that up to you.

lforst:
Gentle bump :) Seems like this was helpful for at least one user https://github.com/getsentry/sentry-javascript/issues/8982#issuecomment-1760347908

docs/shared/packages/next/next-config-setup.md
lforst: Actually, I went ahead and removed the silent because the log output is actually too useful to be silenced by default.

lforst:
gentle bump :)

lforst:
smol lil gentle bump

lforst:
:(

dfr-exnaton:
Thanks a lot for this - also helped in our case...

juristr:
Fixing it in this PR: https://github.com/nrwl/nx/pull/20413

lforst:
@juristr Thank you! <3

PR #19342, opened 2023-09-26
fix(core): add quotes to args with spaces

abejfehr:
## Current Behavior

Running a command and providing additional arguments fails if the arguments contain spaces, even if they're properly surrounded in quotes.

For example:

```
nx build-ios --simulator="iPhone 14 Pro"
```

fails, because the executed command is actually:

```
<command> --simulator=iPhone 14 Pro
```

which isn't correct.

## Expected Behavior

I would've expected the original quotes be preserved, like:

```
<command> --simulator="iPhone 14 Pro"
```

but apparently that's not possible. Node doesn't have the quotes in the `process.argv` object, and apparently that's because [the shell itself removes them](https://stackoverflow.com/questions/10830418/how-to-prevent-losing-double-quotes-in-argv).

So instead, I might expect that args are automatically wrapped in quotes if deemed necessary, like:

```
<command> "--simulator=iPhone 14 Pro"
```

which does have the desired behaviour.

## Related Issue(s)

I didn't make an issue yet unfortunately 😅 let me know if this is necessary

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -299,11 +299,17 @@ export function interpolateArgsIntoCommand(
     return command.replace(regex, (_, group: string) => opts.parsedArgs[group]);
   } else if (forwardAllArgs) {
     return `${command}${
-      opts.__unparsed__.length > 0 ? ' ' + opts.__unparsed__.join(' ') : ''
+      opts.__unparsed__.length > 0
+        ? ' ' + opts.__unparsed__.map(wrapInQuotesIfNeeded).join(' ')
+        : ''
     }`;
   } else {
     return command;
   }
+}
+
+function wrapInQuotesIfNeeded(arg: string) {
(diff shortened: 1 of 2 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -299,13 +299,19 @@ export function interpolateArgsIntoCommand(
     return command.replace(regex, (_, group: string) => opts.parsedArgs[group]);
   } else if (forwardAllArgs) {
     return `${command}${
-      opts.__unparsed__.length > 0 ? ' ' + opts.__unparsed__.join(' ') : ''
+      opts.__unparsed__.length > 0
```
abejfehr, line 302: This change was unrelated, it was made by the formatter

abejfehr:
I'll just close this for now

abejfehr:
**Edit:** Whoops, this isn't actually the PR that I thought I was closing 😅 I wish I could reopen it, but I won't put in the effort right now. Maybe I'll make an issue

PR #19346, opened 2023-09-26
fix(react): use forked process for running devRemotes

Coly010:
## Current Behavior

When running `--devRemotes` with TS configuration for Module Federation, we hit a transpilation error.

## Expected Behavior

We should not hit a transpilation error.

This does require doing a refresh on the browser to see updates

## Related Issue(s)

Fixes #


Coly010:
Closed in favour of: https://github.com/nrwl/nx/pull/19387

PR #19357, opened 2023-09-27
fix(core): generate permissions on github ci workflow

JamesHenry:
## Current Behavior

Folks need to add permissions themselves to make runs on the main branch work correctly with nx-set-shas (used within the shared CI workflow).

## Expected Behavior

Appropriate permissions are already generated by the generator

## Related Issue(s)

Fixes #


meeroslav (approved):
Thank you!

PR #19364, opened 2023-09-28
fix(react-native): fix start packager script so it can locate metro config

chiubaka:
Patches the Start Packager script so that it's able to properly locate the metro config

## Current Behavior

For many iOS-related build tasks, the current script opens a terminal window with an error message stating that the metro config could not be found. Tasks requiring metro will still run if metro is running separately elsewhere, but may fail if the terminal launched by the script is the only one running.

## Expected Behavior

Launched terminal window successfully loads Metro and build processes all run normally.


kevindice:
+1 I am seeing this issue when running `test-ios` to run e2e tests on a React Native app using Detox.

xiongemi:
i don't think this pr is still valid. a lot of changes have made to the pbxproj and it should be fixed.

kevindice:
We have just started metro in a separate shell and that's worked just fine for us.

PR #19387, opened 2023-09-29
fix(react): transpile only the config files then disable tsnode service

Coly010:
## Current Behavior

Dev Remotes cause both a TS error with redeclared const in a block-scope and type-checking running out of memory.

## Expected Behavior

Dev remotes with a TS Config should work without issue

## Related Issue(s)

Fixes #


leosvelperez (approved):

e2e/react-core/src/react-module-federation.test.ts
```diff
@@ -141,6 +141,17 @@ describe('React Module Federation', () => {
     // check default generated host is built successfully
     const buildOutput = runCLI(`run ${shell}:build:development`);
     expect(buildOutput).toContain('Successfully ran target build');
+
+    // check serves devRemotes ok
+    const shellProcess = await runCommandUntil(
+      `serve ${shell} --devRemotes=${remote} --verbose`,
+      (output) => {
+        return output.includes(
+          `All remotes started, server ready at http://localhost:4200`
```
meeroslav, line 150: Not a blocking request, but can we try to keep the ports interpolated in these tests and have them somewhere explicitly in the const?

This helps with future port overlap issues.

graph/client-e2e/project.json
```diff
@@ -4,7 +4,17 @@
   "sourceRoot": "graph/client-e2e/src",
   "projectType": "application",
   "targets": {
+    "pre-e2e": {
+      "inputs": ["default", "^production", "{workspaceRoot}/jest.preset.js"],
+      "executor": "@nx/jest:jest",
```
meeroslav, line 9: This could have been just a simple `node ...` script ran by `run-commands`. There was no need for a unit test.

meeroslav (approved):
Left some comments, but it's not critical. We can clean it up afterwards.

graph/client-e2e/project.json
Coly010: Unless we just call the `pnpm install cypress` command manually, this needs to be in a unit test, because the `ensureCypressInstallation` helper relies on `jest`.

We also probably cant just use `pnpm` install as we'd need to know the current package manager. Perhaps a script could find the PM and then install it? Would that be preferable?

image

mandarini (approved):

e2e/react-core/src/react-module-federation.test.ts
Coly010: The host and remote generators do not have an option for specifying the port, this is a potential enhancement we should add to those generators

ndcunningham (approved):

packages/react/src/executors/module-federation-dev-server/module-federation-dev-server.impl.ts
```diff
@@ -53,13 +53,20 @@ function getModuleFederationConfig(
...
-    tsNodeRegister(moduleFederationConfigPathTS, tsconfigPath);
+    tsNodeService = tsNodeRegister(moduleFederationConfigPathTS, tsconfigPath, {
+      transpileOnly: true,
+    });
     moduleFederationConfigPath = moduleFederationConfigPathTS;
   }
 
   try {
     const config = require(moduleFederationConfigPath);
+    if (tsNodeService) {
```
jaysoo, line 67: Should we call `registerTsProject` instead from `@nx/js` plugin, which as a clean-up function that calls `.enabled(false)` for `ts-node`?

Coly010: I actually think `getTranspiler` might be better from `nx`

graph/client-e2e/project.json
```diff
@@ -4,7 +4,17 @@
   "sourceRoot": "graph/client-e2e/src",
   "projectType": "application",
   "targets": {
+    "pre-e2e": {
```
FrozenPandaz, line 7: This stuff is separate right? Can we sort this out outside of this PR?

packages/js/src/utils/typescript/tsnode-register.ts
```diff
@@ -1,15 +1,22 @@
...
   if (!file?.endsWith('.ts')) return;
   // Register TS compiler lazily
-  require('ts-node').register({
+  const tsNodeService = require('ts-node').register({
     project: tsConfig,
     compilerOptions: {
       module: 'CommonJS',
       types: ['node'],
     },
+    ...tsNodeOptions,
```
FrozenPandaz, line 16: Do we need to pass `transpileOnly`? I'm not sure if the `swc` version of this type checks.. maybe we should just always `transpileOnly` 🤔 

This also seems kind of separate from this change so I don't know if it should be done in this PR.

FrozenPandaz (changes requested):

packages/angular/src/builders/utilities/module-federation.ts
```diff
@@ -3,6 +3,7 @@ import { join } from 'path';
 import { existsSync, readFileSync } from 'fs';
 import { logger } from '@nx/devkit';
 import { tsNodeRegister } from '@nx/js/src/utils/typescript/tsnode-register';
+import { registerTsProject } from 'nx/src/utils/register';
```
FrozenPandaz, line 6: This should come from `@nx/js/src/internal`

packages/angular/src/builders/utilities/webpack.ts
```diff
@@ -1,5 +1,6 @@
 import { merge } from 'webpack-merge';
 import { tsNodeRegister } from '@nx/js/src/utils/typescript/tsnode-register';
+import { registerTsProject } from 'nx/src/utils/register';
```
FrozenPandaz, line 3: This should come from `@nx/js/src/internal`

FrozenPandaz (changes requested):
Looks much better, only the import should be be fixed.

FrozenPandaz (approved):

jaysoo (approved):

PR #19388, opened 2023-09-30
fix(nx): update mental model doc

dylanmoz:
This confused me a bit since there's three tasks in the image above this copy


meeroslav:
This documentation is actually correct. The `two` refers to two `test` tasks - for `lib` and for `app`.

meeroslav:
This documentation is actually correct. The `two` refers to two `test` tasks - for `lib` and for `app`.

Issue #19400, opened 2023-10-02
Add a configuration option to disable dotenv-expand if desired


benpsnyder:
## Description
Please add a flag in `nx.json` for disabling the use of `dotenv-expand` for more advanced use cases. Example:
```json
{
  "env": {
    "expand": "false"
  }
}
```

## What we're solving for
We don't want `$` signs to get expanded in the root `.env` sometimes

## Motivation
Many of our repos are multi-language, using the `@nx-dotnet/core` package for .NET 8 projects for example. Sometimes, we might have a little Rust in there too :)

All of our projects use a `Makefile` with documented commands. Here is a common setup:

(notice this line specifically `bunx --bun dotenv-cli -e .config/environment/._shared_pre.env -e .config/environment/.$(environment).env -e .config/environment/._shared_post.env -- bun --bun ./tmp/dotenv-expand-transpiler.ts $(DOTENV_EXPAND_PREFIXES) ./ .env false >&2`)

```mk
#!make
MAKEFLAGS += --silent --job=10

################# SETUP START ##################

REPO_ROOT := $(CURDIR)
SHELL := /bin/bash

nprocs ?= $(shell node -e 'console.log(require("os").cpus().length - 2)')
environment ?= development
DOTENV_EXPAND_PREFIXES ?= SNYDER_,NX_,ASPNETCORE_,DOTNET_,NODE_,PRODUCTION,WORKING_ENV
ifneq ($(wildcard pnpm-lock.yaml),)
_ := $(shell bunx --bun download-cli --out ./tmp https://raw.githubusercontent.com/snydertechnologies/bun-utils/main/nx/dotenv-expand-transpiler.ts)
_ := $(shell bunx --bun rimraf .env package-lock.json yarn.lock >&2)
_ := $(shell bunx --bun dotenv-cli -e .config/environment/._shared_pre.env -e .config/environment/.$(environment).env -e .config/environment/._shared_post.env -- bun --bun ./tmp/dotenv-expand-transpiler.ts $(DOTENV_EXPAND_PREFIXES) ./ .env false >&2)
else
$(error Please 'pnpm install'...)
endif
include .env
export $(shell sed 's/=.*//' .env)

# trick variables
noop =
comma := ,
space = $(noop) $(noop)

copy.env.dist:
	@if [ "$(environment)" != "development" ]; then \
		find ./dist/ -type d -name "$(DotnetVersion)" -exec cp ./.env {} \; ;\
[...]

PR #19401, opened 2023-10-02
feat(core): add NX_ENV_EXPAND environment variable to disable expansion if desired

benpsnyder:
## Current Behavior
Nx always expands environment variables using `dotenv-expand`.

## Expected Behavior
The user can disable expansion of variables by setting `NX_ENV_EXPAND=false`

## Related Issue(s)
#19400


Cammisuli:
Sorry! I've been busy working on something that's taking more time than I would have liked 😅 

I haven't forgot about this, it's on my todos!

packages/nx/src/tasks-runner/task-env.ts
```diff
@@ -124,7 +124,10 @@ function getNxEnvVariablesForTask(
 
 function loadDotEnvFilesForTask(
   task: Task,
-  environmentVariables: NodeJS.ProcessEnv
+  environmentVariables: NodeJS.ProcessEnv,
+  envExpand = process.env.NX_ENV_EXPAND !== undefined
```
Cammisuli, line 128: this doesn't need to be a argument to this function. You can remove it and use the logic suggested in the if branch

packages/nx/src/tasks-runner/task-env.ts
```diff
@@ -172,13 +175,20 @@ function loadDotEnvFilesForTask(
...
       override: false,
     });
-    environmentVariables = {
-      ...expand({
-        ...myEnv,
-        ignoreProcessEnv: true, // Do not override existing env variables as we load
-      }).parsed,
-      ...environmentVariables,
-    };
+    if (envExpand) {
```
Cammisuli, line 178:
```suggestion
    if (process.env.NX_ENV_EXPAND !== 'false') {
```

Cammisuli (changes requested):
see comments

SmoshySmosh:
I'm really looking forward to this feature!

Riron:
If I am not mistaking, the same check should be applied [here](https://github.com/nrwl/nx/blob/master/packages/nx/bin/nx.ts#L119). Otherwise, NX will be comparing expanded & non expanded envs when unloading envs during tasks run [here](https://github.com/nrwl/nx/blob/master/packages/nx/src/tasks-runner/task-env.ts#L192).

PR #19413, opened 2023-10-02
docs(angular): typo in dynamic-module-federation-with-angular

Formatted:
artifact spelling


Coly010 (approved):
You're right, we should stick to US English, this is my bad from being from UK 😝

isaacplmann (approved):

PR #19075, opened 2023-09-08
fix(core): prettier 3 shouldn't cause errors due to v8 compile cache

guysenpai:
## Current Behavior

After fix #19042 the errors remain

## Expected Behavior

Theses errors should no longer appear

## Related Issue(s)

Fixes #18721

PR #19416, opened 2023-10-03
fix(core): prettier getSupportInfo use promises

d0whc3r:
In prettier 3 the `getSupportInfo` method now is a promise

## Current Behavior

Now, when you try to run "format:write" it fails, and jumps into the "catch", it uses "all files".

You can't specify the files you want to "format" (if you use format:write in lint-staged this will format all other files, not only staged ones)

## Expected Behavior

Only the desired files in format:write should be formated

## Related Issue(s)

https://github.com/nrwl/nx/pull/19075

Fixes #

Add await in `getSupportInfo` method

### DIFF
--- a/packages/nx/src/command-line/format/format.ts
+++ b/packages/nx/src/command-line/format/format.ts
@@ -81,9 +81,8 @@ async function getPatterns(
 
     const p = parseFiles(args);
 
-    const supportedExtensions = (await prettier
-      .getSupportInfo())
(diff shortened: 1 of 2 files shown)


d0whc3r:
wtf??

image

is that the error?

AgentEnder (approved):

AgentEnder:
This fix has already been made in main, so I'm going to close out this PR.

PR #19425, opened 2023-10-03
feat(nuxt): nuxt init and app generator

mandarini:
App generator for Nuxt

`nx g @nx/nuxt:app`

* Small todo: what's up with the FOUC on first load?
   Fixed it: It was the `className`. In Vue it's just `class`, not `className`.

MOVED HERE: https://github.com/nrwl/nx/pull/19509


packages/nuxt/src/utils/create-ts-config.ts
```diff
@@ -0,0 +1,63 @@
...
+  options: {
+    strict?: boolean;
+    style?: string;
+    rootProject?: boolean;
+    unitTestRunner?: string;
+  },
+  relativePathToRootTsConfig: string
+) {
+  const json = {
+    compilerOptions: {
```
mandarini, line 17: Here, we put all these, or do we just create a `tsconfig.json` file that has:
```
{
  // https://nuxt.com/docs/guide/concepts/typescript
  "extends": "./.nuxt/tsconfig.json"
}

```

Like an original generated `nuxt` app has? I mean, are we missing something? I copied the compilerOptions from there, fwiw

mandarini: made the change

mandarini:
Closing, in favor of: https://github.com/nrwl/nx/pull/19509

PR #19426, opened 2023-10-03
fix(core): cache runtime hashes with the env as part of the key

Cammisuli:
## Current Behavior

The daemon caches the runtime input results using only the input as the key

## Expected Behavior

The daemon caches runtime inputs using the input + a stringified version of the ENV. 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/hasher/task-hasher.ts
+++ b/packages/nx/src/hasher/task-hasher.ts
@@ -744,7 +744,9 @@ class TaskHasherImpl {
   }
 
   private async hashRuntime(runtime: string): Promise<PartialHash> {
-    const mapKey = `runtime:${runtime}`;
+    const env = getHashEnv();
+    const env_key = JSON.stringify(env);
+    const mapKey = `runtime:${runtime}-${env_key}`;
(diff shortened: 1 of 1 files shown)


packages/nx/src/hasher/task-hasher.ts
```diff
@@ -744,15 +744,17 @@ class TaskHasherImpl {
   }
 
   private async hashRuntime(runtime: string): Promise<PartialHash> {
-    const mapKey = `runtime:${runtime}`;
+    let env = getHashEnv();
```
FrozenPandaz, line 747: This can be `const` right?

packages/nx/src/hasher/task-hasher.ts
```diff
@@ -744,15 +744,17 @@ class TaskHasherImpl {
...
+    let env_key = JSON.stringify(env);
```
FrozenPandaz, line 748: This can also be `const`

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #19431, opened 2023-10-03
docs(core): add migrate to monorepo sections

isaacplmann:
Migrate to monorepo sections for angular standalone tutorial and react standalone tutorial

### DIFF
--- a/docs/shared/angular-standalone-tutorial/angular-standalone.md
+++ b/docs/shared/angular-standalone-tutorial/angular-standalone.md
@@ -813,7 +813,9 @@ Learn more about how to [enforce module boundaries](/core-features/enforce-modul
 
 ## Migrating to a Monorepo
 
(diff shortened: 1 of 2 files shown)


docs/shared/angular-standalone-tutorial/angular-standalone.md
```diff
@@ -811,6 +811,10 @@ If you have the ESLint plugin installed in your IDE you should immediately see a
 
 Learn more about how to [enforce module boundaries](/core-features/enforce-module-boundaries).
 
+## Migrating to a Monorepo
+
+If you want to move `myngapp` out of the root, you can run the [`convert-to-monorepo` generator](/nx-api/workspace/generators/convert-to-monorepo) or [manually move the configuration files](/recipes/tips-n-tricks/standalone-to-integrated).
```
juristr, line 816: just a little comment here. Can we change the "out of the root" which sounds a bit weird. Someone doesn't really want to move "out of the root" but rather expand their current workspace with an additional application etc and therefore transition to a monorepo.

We should also mention the full monorepo tutorial and link it where we have it already

juristr (changes requested):

juristr (approved):

PR #19441, opened 2023-10-04
feat(react): remove stylus option from generators

jaysoo:
This PR removes the Stylus option from React and Next.js apps. They've been deprecated since Nx 15, and will no longer be an option starting in Nx 17. Note that existing Stylus setup will continue to work until Nx 18, and we will provide guidance on switching to Sass before then.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -52,8 +52,6 @@
     "sass-loader": "^12.2.0",
     "source-map-loader": "^3.0.0",
     "style-loader": "^3.3.0",
-    "stylus": "^0.59.0",
-    "stylus-loader": "^7.1.0",
     "terser-webpack-plugin": "^5.3.3",
     "ts-loader": "^9.3.1",
     "tsconfig-paths-webpack-plugin": "4.0.0",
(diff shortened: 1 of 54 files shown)


mandarini (approved):

isaacplmann (approved):

packages/webpack/package.json
```diff
@@ -52,7 +52,6 @@
     "sass-loader": "^12.2.0",
     "source-map-loader": "^3.0.0",
     "style-loader": "^3.3.0",
-    "stylus": "^0.59.0",
     "stylus-loader": "^7.1.0",
```
JamesHenry, line 55: Shouldn't this be removed as well?

JamesHenry (changes requested):

xiongemi (approved):

packages/webpack/package.json
jaysoo: Removed

jaysoo: Ah, I removed it from the root `package.json`. This one is here for now until Nx v18. This PR removes the ability to generate new project with Stylus. The `stylus-loader` is still used for now for projects that have not migrated.

jaysoo: removed

JamesHenry (approved):

FrozenPandaz (approved):

PR #19449, opened 2023-10-04
fix(nx-dev): video cards formatting

isaacplmann:
Fixes video card formatting

[Old cards](https://nx.dev/concepts/integrated-vs-package-based):
Screenshot 2023-10-04 at 4 02 06 PM

[New cards](https://nx-dev-git-fork-isaacplmann-nx-dev-video-card-formatting-nrwl.vercel.app/concepts/integrated-vs-package-based):
Screenshot 2023-10-04 at 4 01 12 PM

### DIFF
--- a/nx-dev/ui-markdoc/src/lib/tags/cards.component.tsx
+++ b/nx-dev/ui-markdoc/src/lib/tags/cards.component.tsx
@@ -131,25 +131,29 @@ export function Card({
   const hasYoutubeId = !!youtubeRegex ? youtubeRegex[1] : '';
 
   return (
-    <div
+    <a
       key={title}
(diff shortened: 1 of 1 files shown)


juristr (approved):
Much better, thanks. There's still an issue where the grid doesn't expand the full size, but that's a problem of the grid component/config rather than the video cards themselves. So we might address it separately.

Basically right now it has 4 grid cols, while only 3 items are in the list:
image

We should change it s.t. it properly fills the entire space
image

We need to set the various sm, md and lg props accordingly. I think I messed this up when adjusting the grid component for the docs entry page tiles which might have broken some of the defaults.

juristr:
Tbh 🤔 I think we could even have just the thumbnails without the title & description. Would probably also look better. Might be worth checking in with @ZackDeRose though as he's working on some "video gallery" component to embed YT vids in our docs.

PR #19452, opened 2023-10-04
feat(react): use JS webpack config files for module federation

jaysoo:
TS config files are having issues right now so let's default to JS files until we patch them.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/react-core/src/react-module-federation.test.ts
+++ b/e2e/react-core/src/react-module-federation.test.ts
@@ -2,6 +2,7 @@ import { stripIndents } from '@nx/devkit';
 import {
   checkFilesExist,
   cleanupProject,
+  killPorts,
   killProcessAndPorts,
   newProject,
   readJson,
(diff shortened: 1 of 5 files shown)


e2e/react-core/src/react-module-federation.test.ts
```diff
@@ -22,7 +23,7 @@ describe('React Module Federation', () => {
 
   afterAll(() => cleanupProject());
 
-  it('should generate host and remote apps', async () => {
+  it('xxxx should generate host and remote apps', async () => {
```
ndcunningham, line 26: xxx here

e2e/react-core/src/react-module-federation.test.ts
```diff
@@ -294,7 +290,7 @@ describe('React Module Federation', () => {
     }
   }, 500_000);
 
-  it('should support host and remote with library type var', async () => {
+  it('xxxx should support host and remote with library type var', async () => {
```
ndcunningham, line 293: same here

ndcunningham (changes requested):

FrozenPandaz (approved):

PR #19510, opened 2023-10-09
fix(linter): check if circular dependency resolves to a separate angular entrypoint

skrtheboss:
Instead of checking for secondary entrypoint, check if for different entrypoint to allow importing main entrypoint in secondary entrypoint. Like @angular/core/testing imports from @angular/core.

## Current Behavior

Importing main entrypoint in secondary entrypoint is not allowed.

## Expected Behavior

Importing main entrypoint in secondary entrypoint should work.

## Related Issue(s)

Fixes #


meeroslav:
Hi, thank you for the PR. This should already be solved by https://github.com/nrwl/nx/pull/20859.

It was release in the latest beta - [17.3.0-beta.1](https://github.com/nrwl/nx/releases/tag/17.3.0-beta.1)

PR #19523, opened 2023-10-10
feat(angular): add federate-module generator

Coly010:
## Current Behavior

We currently only support generating remotes with a route being exposed by the remote.

## Expected Behavior

We should add a generate that allows you to federate any code via a remote.
We should allow for this to happen in a consistent manner and provide type-safety via tsconfig paths.

This PR will add a `federate-module` generator that can

1. Add an exposes property to an existing remote to federate all the exports at a given path
2. Add a new remote if one does not exist
3. Attach the remote to the host application if --host is provided
4. Update the root tsconfig to add the federated module's remote path to the tsconfig path aliases

Usage of the command is:

```
nx g @nx/angular:federate-module IsEven --path=libs/is-even/src/index.ts --remote=myremote --host=myshell
```

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/manifests/menus.json
+++ b/docs/generated/manifests/menus.json
@@ -6080,6 +6080,14 @@
                 "disableCollapsible": false
               },
               {
+                "id": "federate-module",
+                "path": "/nx-api/angular/generators/federate-module",
+                "name": "federate-module",
+                "children": [],
+                "isExternal": false,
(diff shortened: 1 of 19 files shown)


isaacplmann (approved):

jaysoo (approved):

ndcunningham (approved):
LGTM

PR #19527, opened 2023-10-10
feat(angular): support generating artifacts using options as provided

leosvelperez:
## Current Behavior

Artifact generators (e.g. `component`) options are more complex than they should be and are not consistent across the different Nx first-party plugins. These are some of the current problems:

- Different generators have different option names for the same thing.
- The `directory`/`path` option is expected to be relative to the workspace root for some generators, and for others, it’s expected to be relative to the project source root.
- Devs need to specify many options to generate an artifact in a specific location: `name`, `project`, `directory`/`path`, `flat`.

Apart from those, there are also some automatic decisions made by these generators, like converting the provided names to kebab-case. While this can enforce certain standards, it can be limiting and might not fit every use case.

## Expected Behavior

- Similar to the changes done to the project generators, we'll have two formats: `as-provided` and `derived`. In most cases, we'll prompt the user for the format they want to use.
- We’ll deprecate the following options, which will only be considered for the `derived` format:
    - `project`
    - `flat`
- We’ll standardize the `path` and `directory` options name to be `directory`. We'll keep backward compat for `path`.
- We’ll standardize the `directory` option to be relative to the cwd.
- We’ll standardize the support for providing a path as the artifact name. It will be a shorthand of passing name & directory options.

## Related Issue(s)

Fixes #


packages/devkit/src/generators/artifact-name-and-directory-utils.spec.ts
```diff
@@ -0,0 +1,863 @@
...
+          baseName: 'my-component.ts',
+          name: 'my-component',
+          path: 'apps/app1/src/app/MyComponent/my-component.ts',
+        },
+        project: 'app1',
+        nameAndDirectoryFormat: 'derived',
+      });
+    });
+
+    it('should support receiving a different file extension', async () => {
```
jaysoo, line 701: Do we need to support file extensions? The component generators for both Angular and React only need to know what the `fileName` is (e.g. `my-component` vs `MyComponent`). The extensions are in the templates passed to `generateFiles`.

I don't think we need to expand the `file` entry, it can just be a string like:

```json
{
  name: 'myComponent',
  directory: apps/app1/src/app/MyComponent',
  file: 'my-component'
  project: 'app1',
  nameAndDirectoryFormat: 'derived',
}
```

The rest can be computed from those two values.

---

One small thing, I think `artifactName` should be `artifactType`, same way that existing `determineProjectNameAndRootOptions` function takes in `projectType`. And then the returned `name` should be `artifactName`, so the two functions are consistent.

```json
{
  artifactName: 'myComponent',
  directory: apps/app1/src/app/MyComponent',
  file: 'my-component'
  project: 'app1',
  nameAndDirectoryFormat: 'derived',
}
```

leosvelperez: Good points. I agree and by doing that we could even leave the handling of the suffix out of the helper. That's something the consumer can compose with the information returned.

packages/devkit/src/generators/artifact-name-and-directory-utils.ts
```diff
@@ -0,0 +1,404 @@
...
+function getAsProvidedOptions(
+  tree: Tree,
+  options: ArtifactGenerationOptions
+): NameAndDirectoryOptions {
+  const relativeCwd = getRelativeCwd();
+
+  const asProvidedDirectory = options.directory
+    ? joinPathFragments(relativeCwd, options.directory)
+    : relativeCwd;
+  const asProvidedProject = findProjectFromPath(tree, asProvidedDirectory);
```
FrozenPandaz, line 229: Use the existing utils for this

packages/devkit/src/generators/artifact-name-and-directory-utils.ts
```diff
@@ -0,0 +1,404 @@
...
+}
+
+async function determineFormat(
+  formats: NameAndDirectoryFormats,
+  options: ArtifactGenerationOptions
+): Promise<NameAndDirectoryFormat> {
+  if (!formats.derived) {
+    return 'as-provided';
+  }
+  if (!formats['as-provided']) {
```
FrozenPandaz, line 80: There is always a `as-provided` format

packages/devkit/src/generators/artifact-name-and-directory-utils.ts
```diff
@@ -0,0 +1,404 @@
...
+    const deprecationMessage = getDeprecationMessage(options, formats);
+    logger.warn(deprecationMessage);
+
+    return 'derived';
+  }
+
+  if (process.env.NX_INTERACTIVE !== 'true' || !isTTY()) {
+    return 'derived';
```
FrozenPandaz, line 88: Show the deprecation message

packages/angular/src/generators/component/schema.json
```diff
@@ -4,31 +4,43 @@
...
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      },
+      "x-prompt": "What name would you like to use for the component?"
+    },
+    "directory": {
+      "type": "string",
+      "description": "The directory at which to create the component file. When `--nameAndDirectoryFormat=as-provided`, it will be relative to the current working directory. Otherwise, it will be relative to the workspace root.",
+      "aliases": ["dir"],
```
FrozenPandaz, line 22: Add path as an alias, not deprecated

packages/angular/src/generators/component/schema.json
```diff
@@ -4,31 +4,43 @@
...
+      "x-priority": "important"
+    },
+    "nameAndDirectoryFormat": {
+      "description": "Whether to generate the component in the directory as provided, relative to the current working directory and ignoring the project (`as-provided`) or generate it using the project and directory relative to the workspace root (`derived`).",
+      "type": "string",
+      "enum": ["as-provided", "derived"]
+    },
     "path": {
```
FrozenPandaz, line 30: Remove in favor of directory

FrozenPandaz (changes requested):

packages/devkit/src/generators/artifact-name-and-directory-utils.spec.ts
leosvelperez: As discussed offline, the helper needs the file extension. We'll simplify some bits of the return type and rename what was suggested.

packages/nx/src/utils/params.ts
```diff
@@ -633,6 +633,8 @@ export async function combineOptionsForGenerator(
     schema,
     false
   );
+
+  warnDeprecations(combined, schema);
```
leosvelperez, line 637: This was being executed after the smart defaults were applied. That was wrong because it would warn about a deprecated option with a smart default (e.g. `project`). The order of operations after this change matches the order of those in the Angular CLI.

FrozenPandaz (approved):

PR #19548, opened 2023-10-10
feat(linter): update typescript-eslint to v6

### LINKED ISSUE #18641: Support for @typescript-eslint/parser@v6

meeroslav:
This PR should make an isolated change to `eslint-plugin` project to enable update to typescript-eslint v6

The lint tasks will fail unless https://github.com/nrwl/nx/pull/19549 is released or the [issue in typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/issues/7723#issuecomment-1756234676) is resolved.

The tests still fail (occasionally) locally with:
Screenshot 2023-10-31 at 21 53 27

## Related Issue(s)

Fixes #18641


pnpm-lock.yaml
```diff
@@ -10355,17 +10426,17 @@ packages:
       - supports-color
     dev: true
 
-  /@typescript-eslint/typescript-estree@5.60.1(typescript@5.1.3):
-    resolution: {integrity: sha512-hkX70J9+2M2ZT6fhti5Q2FoU9zb+GeZK2SLP1WZlvUDqdMbEKhexZODD1WodNRyO8eS+4nScvT0dts8IdaBzfw==}
+  /@typescript-eslint/typescript-estree@5.62.0(typescript@5.1.3):
```
bradzacher, line 10429: this could be causing problems

pnpm-lock.yaml
```diff
@@ -10324,16 +10395,16 @@ packages:
...
     resolution: {integrity: sha512-87NVngcbVXUahrRTqIK27gD2t5Cu1yuCXxbLcFtCzZGlfyVWWh8mLHkoxzjsB6DDNnvdL+fW8MiwPEJyGJQDgQ==}
     engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
     dev: true
 
+  /@typescript-eslint/types@6.7.5:
+    resolution: {integrity: sha512-WboQBlOXtdj1tDFPyIthpKrUb+kZf2VroLZhxKa/VlwLlLyqv/PwUNgL30BlTVZV1Wu4Asu2mMYPqarSO4L5ZQ==}
+    engines: {node: ^16.0.0 || >=18.0.0}
+    dev: true
+
   /@typescript-eslint/typescript-estree@5.59.2(typescript@5.1.3):
```
bradzacher, line 10408: this could be causing problems

pnpm-lock.yaml
```diff
@@ -10417,18 +10488,18 @@ packages:
       - typescript
     dev: true
 
-  /@typescript-eslint/utils@5.60.1(eslint@8.46.0)(typescript@5.1.3):
-    resolution: {integrity: sha512-tiJ7FFdFQOWssFa3gqb94Ilexyw0JVxj6vBzaSpfN/8IhoKkDuSAenUKvsSHw2A/TMpJb26izIszTXaqygkvpQ==}
+  /@typescript-eslint/utils@5.62.0(eslint@8.46.0)(typescript@5.1.3):
```
bradzacher, line 10491: this could be causing problems

pnpm-lock.yaml
```diff
@@ -10462,23 +10532,23 @@ packages:
     engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
```
bradzacher, line 10532: this could be causing problems

pnpm-lock.yaml
```diff
@@ -10462,23 +10532,23 @@ packages:
...
-  /@typescript-eslint/visitor-keys@5.60.1:
-    resolution: {integrity: sha512-xEYIxKcultP6E/RMKqube11pGjXH1DCo60mQoWhVYyKfLkwbIVVjYxmOenNMxILx0TjCujPTjjnTIVzm09TXIw==}
-    engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
-    dependencies:
-      '@typescript-eslint/types': 5.60.1
-      eslint-visitor-keys: 3.3.0
+      eslint-visitor-keys: 3.4.3
     dev: true
 
   /@typescript-eslint/visitor-keys@5.62.0:
```
bradzacher, line 10538: this could be causing problems

bradzacher (commented):
I don't fully understand the pnpm lockfile format but I pointed out a number of instances of v5 of our packages being installed which could cause issues depending on how ESLint resolves things.

pnpm-lock.yaml
meeroslav: This is normal.

We have dev dependencies to some packages that depend on the older `@typescript-eslint/*@v5.x.x` but that doesn't influence the main hoisted version.

This is not related to `pnpm` only. All the package managers keep hierarchical information to ensure version clashes are avoided.

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -353,4 +353,4 @@ export default createESLintRule<Options, MessageIds>({
       },
     };
   },
-});
+}) as any;
```
JamesHenry, line 356: How come this needed an `as any`?

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
```diff
@@ -712,4 +712,4 @@ export default createESLintRule<Options, MessageIds>({
       },
     };
   },
-});
+}) as any;
```
JamesHenry, line 715: How come this needed an `as any`?

JamesHenry (changes requested):
Do we not need a migration to add the node16 module resolution to users' workspaces?

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -40,7 +40,7 @@ export default createESLintRule<Options, MessageIds>({
     type: 'suggestion',
     docs: {
       description: `Checks dependencies in project's package.json for version mismatches`,
-      recommended: 'error',
+      recommended: 'recommended',
```
FrozenPandaz, line 43: Should we add migrations here as a followup?

FrozenPandaz (approved):

packages/eslint-plugin/src/rules/dependency-checks.ts
meeroslav: That was a working version. However, the latest `@typescript-eslint/utils` is no longer inferred so we need to specify it explicitly. Likely, it's something that should be fixed on `@typescript-eslint` side?

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
meeroslav: That was a working version. However, the latest `@typescript-eslint/utils` is no longer inferred so we need to specify it explicitly. Likely, it's something that should be fixed on `@typescript-eslint` side?

packages/eslint-plugin/src/rules/dependency-checks.ts
meeroslav: Yes

packages/eslint/src/generators/workspace-rules-project/__snapshots__/workspace-rules-project.spec.ts.snap
```diff
@@ -87,9 +88,6 @@ export default {
   },
   moduleFileExtensions: ['ts', 'js', 'html'],
   coverageDirectory: '../../coverage/tools/eslint-rules',
-  moduleNameMapper: {
-    '@eslint/eslintrc': '@eslint/eslintrc/dist/eslintrc-universal.cjs',
```
JamesHenry, line 91: We will need a migration for this too if it is no longer something we want to apply? (I'm afraid I can't remember why this was needed originally)

packages/eslint/src/generators/workspace-rules-project/workspace-rules-project.ts
```diff
@@ -104,16 +107,6 @@ export async function lintWorkspaceRulesProjectGenerator(
   // Add swc dependencies
   addSwcRegisterDependencies(tree);
 
-  // Add extra config to the jest.config.ts file to allow ESLint 8 exports mapping to work with jest
-  addPropertyToJestConfig(
```
JamesHenry, line 108: Again, because we've always been generating this we will need a migration for existing workspaces to remove it

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
JamesHenry: I've written to you on slack about this

jaysoo (approved):

FrozenPandaz (approved):

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -4,15 +4,15 @@ import { AST } from 'jsonc-eslint-parser';
...
-
-import { createESLintRule } from '../utils/create-eslint-rule';
 import { readProjectGraph } from '../utils/project-graph-utils';
 import { findProject, getSourceFilePath } from '../utils/runtime-lint-utils';
 import {
   getAllDependencies,
   getPackageJson,
   getProductionDependencies,
 } from '../utils/package-json-utils';
+import type { RuleModule } from '@typescript-eslint/utils/ts-eslint';
```
JamesHenry, line 14: I think this unused

packages/eslint-plugin/src/rules/enforce-module-boundaries.ts
```diff
@@ -41,8 +41,9 @@ import {
   getBarrelEntryPointProjectNode,
   getRelativeImportPath,
 } from '../utils/ast-utils';
-import { createESLintRule } from '../utils/create-eslint-rule';
 import { readProjectGraph } from '../utils/project-graph-utils';
+import type { RuleModule } from '@typescript-eslint/utils/ts-eslint';
```
JamesHenry, line 45: I think this is unused

JamesHenry (approved):

PR #19564, opened 2023-10-11
docs(core): ci with Nx intro and restructure

isaacplmann:
- Restructure Nx Cloud docs
- Write CI with Nx intro


juristr (approved):
Let's ship it

PR #19571, opened 2023-10-11
Add functions to copy pnpm and packageManager sections

### LINKED ISSUE #18402: GeneratePackageJson: package.json with overrides or patches generate incorrect lockfile
### Current Behavior

`GeneratePackageJson` will not copy and trim `overrides` / `patchedDependencies` to the generated lockfile (or `package.json`)

### Expected Behavior

The expectation is that the generate `package.json` and `pnpm-lock.yaml` would be correct and `pnpm install --frozen-lockfile` does not error.

### GitHub Repo

_No response_

### Steps to Reproduce

1. create e.g. NestJS app in monorepo with the following `package.json`:

```json
{
  "dependencies": {
    "@nestjs/axios": "3.0.0"
  },
  "pnpm": {
    "overrides": {
      "axios": "0.27.2"
    }
  }
}
```

2. Add an app with a build target that contains `generatePackageJson: true`.
3. Build the app.
4. Unexpected 1: pnpm overrides (and patches) are missing from the generated `package.json` and `pnpm-lock.yaml`.
5. Unexpected 2: Running `pnpm install --frozen-lockfile` fails.

### Nx Report

```shell
Node : 18.17.0
   OS   : win32 x64
   pnpm : 8.6.10

   nx                      : 15.9.4
   @nrwl/js                : 15.9.4
   @nrwl/jest              : 15.9.4
   @nrwl/linter            : 15.9.4
   @nrwl/workspace         : 15.9.4
   @nrwl/cli               : 15.9.4
   @nrwl/devkit            : 15.9.4
   @nrwl/eslint-plugin-nx  : 15.9.4
   @nrwl/nest              : 15.9.4
   @nrwl/node              : 15.9.4
   @nrwl/tao               : 15.9.4
   @nrwl/webpack           : 15.9.4
   typescript              : 5.1.6
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [X] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

riceboyler:
## Current Behavior
`pnpm` and `packageManager` sections in package.json aren't copied when a new package.json is created, meaning pnpm patchedDependencies are ignored.

## Expected Behavior
`pnpm` and `packageManager` sections should be carried over to any newly generated package.json files so as to easily support `patchedDependencies`.

## Related Issue(s)
https://github.com/nrwl/nx/issues/18402

Fixes #18402

### DIFF
--- a/packages/nx/src/plugins/js/package-json/create-package-json.spec.ts
+++ b/packages/nx/src/plugins/js/package-json/create-package-json.spec.ts
@@ -3,7 +3,11 @@ import * as fs from 'fs';
 import * as configModule from '../../../config/configuration';
 import { DependencyType, ProjectGraph } from '../../../config/project-graph';
 import * as hashModule from '../../../hasher/task-hasher';
-import { copyPackageManagerSection, copyPnpmSection, createPackageJson } from './create-package-json';
+import {
+  copyPackageManagerSection,
+  copyPnpmSection,
+  createPackageJson,
+} from './create-package-json';
 import * as fileutilsModule from '../../../utils/fileutils';
 import { PackageJson } from '../../../utils/package-json';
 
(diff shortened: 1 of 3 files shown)


jaysoo:
Hi @riceboyler Thanks for submitting this PR. There are few problems here:

1. yarn also has patch support, so we need to make sure that works the same way
2. the entire `pnpm` entry probably shouldn't be copied, there are hooks that may not be valid in dist -- we need to handle these separately if needed
3. files in `patches` are not copied to dist, so even if the entries are in `package.json` they will not work

Let me take a look today and provide a patch that solves all of these issues.

PR #19597, opened 2023-10-13
feat(graph): display expanded task inputs

MaxKless:
This PR adds a list of inputs per task to the graph UI:
image
image
It also lists dependencies' & external inputs in separate sections:
image

### Implementation
We've expanded the `nx graph` command. `graph.ts - getExpandedTaskInputs` will use the new `HashPlanner` to calculate all files that contribute to a project's inputs. Because expanding inputs for all projects is potentially time-consuming, we try to do it lazily.
- When showing the graph via `nx graph`, we have a live server started in `graph.ts`. When queried, this calculcates the expanded inputs for only one task.
- When showing the graph in Nx Console, dev mode or e2es, we don't have a live server. That means we need to eagerly calculate all inputs and store them, just like with the project & task graph.

This lazy loading sets the task input request apart from the project & task graph requests made in the graph UI. Those are made once when loading a route. By contrast, the individual task inputs are loaded when opening the task tooltip. This required us to break out of the loader pattern somewhat to get the task inputs.

We could adapt this approach to load all task inputs per route, which would probably result in a lot of unused computation (because we don't expect folks to actually click on all the different task inputs). It would, however, be more in line with how we're currently doing things and result in cleaner code.


packages/nx/src/command-line/graph/graph.ts
```diff
@@ -752,6 +834,130 @@ function createTaskId(
...
+  }
+  return {};
+}
+
+function expandInputs(
+  inputs: string[],
+  project: ProjectGraphProjectNode,
+  allWorkspaceFiles: FileData[],
+  depGraphClientResponse: ProjectGraphClientResponse
+): Record<string, string[]> {
```
Cammisuli, line 862: This should be replaced by the rust functionality that is going to be built out soon. But for now it's fine to have here. I'll replace this when the new function is ready.

Cammisuli (approved):
We're going to have a follow up PR to replace the `expandInputs` function with a rust one to match the task hasher. 

We also should take a look at the actual performance of the task view in the graph because it's quite bloated and slow. But we'll do that in a future PR as well.

PR #19602, opened 2023-10-13
feat(nx-plugin): add as provided prompt for executor generator

FrozenPandaz:
## Current Behavior

Creating executors always go under a project in an executors directory such as `libs/my-plugin/src/executors/executor.ts`

## Expected Behavior

Executors are created where the `cwd` is so the following works:
`cd packages/my-package/nx-integrations/executors/build && nx g executor build` will generate `packages/my-package/nx-integrations/executors/build/executor.ts`

There is a prompt to use the old way sitll.

## Related Issue(s)

Fixes #


AgentEnder (approved):
Looks good to me 🎉

PR #19606, opened 2023-10-13
fix(nx-dev): fix internal link checker

isaacplmann:
Fixes an issue with the internal link checker and linking to docs for private packages


scripts/documentation/internal-link-checker.ts
```diff
@@ -83,38 +83,6 @@ function readSiteMapLinks(filePath: string): string[] {
...
 }
 
-/**
- * This function checks if a link is for a private package.
- * When link is for a private package, it is not included in the sitemap.
- * However, some shared docs might be written for this private package during development.
- * @param link e.g. /nx-api/vite/generators/configuration
- * @returns true if the link is for a private package or NODE_ENV is not development, false otherwise.
- */
-function checkLinkIsForPrivatePackage(link: string) {
```
isaacplmann, line 93: This function doesn't actually help.

1. Documentation inside a private package is not checked for broken links because it doesn't get generated into the `/doc` folder.
2. Links from normal docs to docs for a private package should throw an error, but this check was silencing that warning for some reason.

nx-dev/nx-dev/project.json
```diff
@@ -18,6 +18,12 @@
         ],
         "parallel": false
       },
+      "inputs": [
+        "production",
+        "^production",
+        "{workspaceRoot}/scripts/tsconfig.scripts.json",
+        "{workspaceRoot}/scripts/documentation/internal-link-checker.ts"
+      ],
```
isaacplmann, line 26: Updated inputs because the link checker was being cached even though I had change the script code

FrozenPandaz (approved):

PR #19608, opened 2023-10-13
feat(react): support generating components using the path as provided

jaysoo:
This PR adds consistent directory support to React, Next.js, React Native, and Expo generators.

Starting in v17, users can create components and other artifact by provide a path from the CWD.

```bash
# from workspace root
nx g component libs/ui/src/button # generates libs/ui/src/button.tsx under the ui project

# from libs/ui
nx g component src/button.tsx # generates the same as above since we are in the project dir

# other artifacts
nx g hook libs/my-hooks/use-foo/use-foo # react hook under my-hooks project
nx g redux libs/cart-state/cart # redux slice "cart" under cart-state project
```

## Affected generators
- `@nx/react:component`
- `@nx/react:hook`
- `@nx/react:redux`
- `@nx/next:component`
- `@nx/next:page`
- `@nx/react-native:component`
- `@nx/expo:component`


packages/react/src/generators/redux/schema.d.ts
leosvelperez, line 1: We need to add the deprecated JSDoc for the deprecated properties here to notify devs using the generator programmatically.

packages/react/src/generators/redux/schema.json
```diff
@@ -21,16 +21,18 @@
...
-      "x-priority": "important"
+      }
     },
     "directory": {
       "type": "string",
       "alias": "dir",
       "default": "",
-      "description": "The name of the folder used to contain/group the generated Redux files.",
-      "x-priority": "important"
+      "description": "The directory at which to create the Redux files. When `--nameAndDirectoryFormat=as-provided`, it will be relative to the current working directory. Otherwise, it will be relative to the project root."
```
leosvelperez, line 30: Should `--directory` be important?

packages/react/src/generators/redux/schema.json
```diff
@@ -21,16 +21,18 @@
       "alias": "p",
       "$default": {
         "$source": "projectName"
-      },
-      "x-prompt": "What is the name of the project for this slice?",
-      "x-priority": "important"
+      }
```
leosvelperez, line 24: We need to deprecate `--project`.

leosvelperez (changes requested):

mandarini (approved):

xiongemi (approved):

packages/react/docs/component-examples.md
```diff
@@ -0,0 +1,22 @@
+## Examples
+
+{% tabs %}
+{% tab label="Simple Component" %}
+
+Create a component named `my-component` under the `libs/ui` project:
+
+```bash
```
ndcunningham, line 8: should this be `shell`?

jaysoo: yes

ndcunningham (approved):

PR #19609, opened 2023-10-13
feat(vue): support generating components using the path as provided

jaysoo:
This PR updates `@nx/vue:component` to support a component path provided as is.

e.g.

```
# this generates at libs/ui/src/button.vue, the path belongs to a project
nx g component libs/ui/src/button
```


packages/vue/src/generators/component/schema.json
```diff
@@ -22,9 +22,7 @@
       "alias": "p",
       "$default": {
         "$source": "projectName"
-      },
-      "x-prompt": "What is the name of the project for this component?",
-      "x-priority": "important"
+      }
```
leosvelperez, line 25: We need to deprecate `--project`.

packages/vue/src/generators/component/schema.d.ts
leosvelperez, line 1: We need to add the deprecated JSDoc for the deprecated properties here to notify devs using the generator programmatically.

packages/vue/src/generators/component/schema.json
leosvelperez, line 1: We also need to deprecate `--pascalCaseFiles` and `--pascalCaseDirectory`.

leosvelperez, line 1: Let's also update the description for `--directory` to match the one we're adding for other generators using both formats.

leosvelperez (dismissed):

mandarini (approved):

FrozenPandaz (approved):

PR #19645, opened 2023-10-16
fix(vue): turn multi-word component lint rule off since there is always one for restricted HTML elements

jaysoo:
This lint rule can cause issues for newly generated libs, where `nx g lib products` will have a lint error since `Products.vue` is just one word. The line rule is supposed to guard against conflicts with HTML names, but there is a reserved names lint rule for that purpose.

Also, update the default component name for libs to be less verbose. e.g. `nx g lib products` generates `Products` component, not `ProductsProducts`.

## Related Issue(s)

Fixes #


ndcunningham (approved):

mandarini (approved):

mandarini:
There are more things that should be removed that were specifically introduced to accommodate that rule, I'll add to your PR: https://github.com/nrwl/nx/pull/19507

packages/vue/src/generators/library/library.ts
```diff
@@ -58,7 +58,7 @@ export async function libraryGenerator(tree: Tree, schema: Schema) {
   if (options.component) {
     tasks.push(
       await componentGenerator(tree, {
-        name: options.fileName,
```
mandarini, line 61: Not sure this needs to be removed: https://github.com/nrwl/nx/blame/master/packages/vue/src/generators/library/library.ts#L61

It's been there all along, it's not the one causing the issues, I think.

mandarini: But no harm in removing it either, it seems.

jaysoo: It's just changing the component name from a prefixed version like `ProductProduct` to `Product`.

PR #19653, opened 2023-10-17
feat(storybook): remove component-story generator - integrate into stories

mandarini:
Remove the `@nx/*:component-story` generator.

It was meant for internal use, and its use is no longer needed in a way that justifies a generator. Its code has been moved under the `*:stories` generator, where it's the only place its used. For `vue` and for `nuxt` the generator was not created at all.

Adding a `do-not-merge` and a `breaking-changes` label, to make sure this is an ok change (it is ok on the storybook, react, angular sides).


mandarini:
Closing this, since [there are 4 ppl who use these generators](https://github.com/search?q=%40nx%2Freact%3Acomponent-story&type=code). But still, let's mark them as internal, I guess.

PR #19661, opened 2023-10-17
feat(storybook): add missing react dependency for nextjs

mandarini:
Added missing react dependency.

This PR fixes [this issue](https://staging.nx.app/runs/NRgfljDoua), but it should not be an issue in the first place, because it seems that it's not happening on Nx `next` version, so when I generate a new workspace with `@next` version and I test it out, Storybook builds successfully without that dependency. When I run the e2e tests on latest `master`, and then I also try to `build-storybook` on the generated project, like manually, the tests only pass (storybook only builds) if `react-refresh` is installed.


mandarini:
Closing until further investigation

PR #19667, opened 2023-10-17
fix(vue): remove unitTestRunner option from component generator

jaysoo:
This prompt was for adding `import {describe, ...} from 'vitest` into our spec files, but we already include `vitest/globals` in `tsconfig.config.spec.json`, so there is no reason to prompt. We don't have this option for any other component/artifact generator, so it's weird to have it here.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/vue/src/generators/component/schema.d.ts
+++ b/packages/vue/src/generators/component/schema.d.ts
@@ -8,7 +8,6 @@ export interface Schema {
   fileName?: string;
   inSourceTests?: boolean;
   skipFormat?: boolean;
-  unitTestRunner?: 'jest' | 'vitest' | 'none';
   nameAndDirectoryFormat?: 'as-provided' | 'derived';
 
   /**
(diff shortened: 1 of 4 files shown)


packages/vue/src/generators/component/schema.d.ts
```diff
@@ -8,7 +8,6 @@ export interface Schema {
   fileName?: string;
   inSourceTests?: boolean;
   skipFormat?: boolean;
-  unitTestRunner?: 'jest' | 'vitest' | 'none';
```
FrozenPandaz, line 11: Shouldn't we still have `none`?

FrozenPandaz (changes requested):

packages/vue/src/generators/component/schema.d.ts
jaysoo: No, component generator should never have this in the first place.

FrozenPandaz (approved):

ndcunningham (approved):

PR #19671, opened 2023-10-17
fix(js): remove deprecated babel transform

jaysoo:
`@babel/plugin-proposal-class-properties` is deprecated and the new transform is already in `@babel/preset-env` so it is not necessary. This PR removes it and deprecates the `classProperties.loose` option in `@nx/js/babel` preset in favour of the broader `loose` option.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/.eslintrc.json
+++ b/packages/js/.eslintrc.json
@@ -45,7 +45,6 @@
               // require.resolve is used for these packages
               "source-map-support",
               "@babel/core",
-              "@babel/plugin-proposal-class-properties",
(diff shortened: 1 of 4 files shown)


mandarini (approved):

meeroslav (approved):

Michsior14:
Aren't  `class-properties` available on `es2022` only? So all executors that uses older versions (e.g. rollup supports: `es2015`, `es2020` and `esnext`) won't work anymore?

PR #19679, opened 2023-10-18
fix(core): remove more -d aliases

MaxKless:
I missed some in my last PR


FrozenPandaz (changes requested):
These are executors.. They don't have a dryRun so they're okay?

MaxKless:
yeah sorry i should drink coffee first

PR #19694, opened 2023-10-18
feat(module-federation): add remote configuration override

### LINKED ISSUE #19693: Override MFE remote configuration

## Description

The Angular MFE server passes the configuration to the remotes which cannot be overridden.

## Motivation

If this can be changed there are no need to create empty configuration in the remotes.

## Suggested Implementation

Allow the `devRemotes` to be an object and override the configuration.

csutorasr:
## Current Behavior

The configuration of the served MFE always passed to the remotes. If a new configuration is needed to skip one remote (e.g. `serve:skip-remote1`) but `remote2` then a configuration called `skip-remote1` is needed in the `remote2`.

## Expected Behavior

Add an ability to override the configuration and the empty configurations in the remotes can be deleted.

## Related Issue(s)

Fixes #19693

### DIFF
--- a/docs/generated/packages/angular/executors/module-federation-dev-server.json
+++ b/docs/generated/packages/angular/executors/module-federation-dev-server.json
@@ -105,7 +105,20 @@
       },
       "devRemotes": {
         "type": "array",
-        "items": { "type": "string" },
+        "items": {
+          "oneOf": [
+            { "type": "string" },
+            {
+              "type": "object",
+              "properties": {
+                "remoteName": { "type": "string" },
+                "configuration": { "type": "string" }
+              },
+              "required": ["remoteName"],
+              "additionalProperties": false
+            }
+          ]
+        },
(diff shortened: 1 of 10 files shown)


Coly010:
Hey, thanks for the contribution 🎉

However, our Module Federation support is undergoing some rework currently and we'll need to discuss internally if this is something we want to add.

I'll update you when we come to a decision

csutorasr:
@Coly010 Do you have any updates/progress here? Sorry for disturbing you, I am just curious and would like to use this feature.

xiongemi:
not sure it would scale well for multiple remotes. i am going to close it for now

csutorasr:
@xiongemi Why would not it scale for multiple remotes? Currently this feature is there. If you want to replace the configuration for one, then it adds the possibility. If you want to include any remotes than those has to be listed, so it makes no difference in those situations. This change would just add better configurability.

Coly010 (approved):

Coly010 (approved):

PR #19696, opened 2023-10-18
fix(react): update usage of deprecated "project" option in generators

jaysoo:
This PR removes usages of `project` option after normalization through `determineArtifactNameAndDirectoryOptions` function. It will be undefined in the future, so we need to use the inferred project name instead.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/next/src/generators/page/page.ts
+++ b/packages/next/src/generators/page/page.ts
@@ -27,13 +27,11 @@ export async function pageGeneratorInternal(host: Tree, schema: Schema) {
   const options = await normalizeOptions(host, schema);
   const componentTask = await reactComponentGenerator(host, {
     ...options,
-    project: schema.project,
-    pascalCaseFiles: false,
+    nameAndDirectoryFormat: 'as-provided', // already determined the directory so use as is
     export: false,
     classComponent: false,
     routing: false,
     skipTests: !options.withTests,
-    flat: !!options.flat,
     skipFormat: true,
   });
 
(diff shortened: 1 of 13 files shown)


packages/react/src/generators/redux/schema.d.ts
```diff
@@ -1,5 +1,8 @@
 export interface Schema {
   name: string;
+  /**
+   * @deprecated Provide the `directory` option instead and use the `as-provided` format. The project will be determined from the directory provided. It will be removed in Nx v18.
+   */
   project: string;
```
ndcunningham, line 6: should this also be nullable?

packages/next/src/generators/page/page.ts
```diff
@@ -27,13 +27,12 @@ export async function pageGeneratorInternal(host: Tree, schema: Schema) {
   const options = await normalizeOptions(host, schema);
   const componentTask = await reactComponentGenerator(host, {
     ...options,
-    project: schema.project,
-    pascalCaseFiles: false,
+    nameAndDirectoryFormat: 'as-provided', // already determined the directory so use as is
+    project: undefined,
```
ndcunningham, line 31: this is marked as `undefined` because we are using `projectName` instead and `project` will be removed in Nx 18 right?

jaysoo: yeah exactly, just ignore it

ndcunningham (approved):

packages/next/src/generators/page/page.ts
jaysoo: Yeah, I updated to make them optional. I can probably remove this line.

FrozenPandaz (approved):

PR #19701, opened 2023-10-18
feat(nextjs): Add experimental https option

naofel1:
## Current Behavior
Currently, the @nx/next:server executor does not support an experimentalHttps flag. Users who want to use the --experimental-https feature with their Next.js applications don't have an out-of-the-box option with Nx.

## Expected Behavior
With the changes in this PR, users can use the experimentalHttps option in their project.json to enable or disable the experimental HTTPS feature for their Next.js server when using Nx.

### DIFF
--- a/docs/generated/packages/vite/executors/build.json
+++ b/docs/generated/packages/vite/executors/build.json
@@ -26,6 +26,12 @@
         "description": "Skip type-checking via TypeScript. Skipping type-checking speeds up the build but type errors are not caught.",
         "default": false
       },
+      "tsConfig": {
+        "type": "string",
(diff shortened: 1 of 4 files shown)


naofel1:
@leosvelperez

onemeangrilledcheese:
One way to get this working for now for local development is just adding `next dev --experimental-https` in the app for which you want to enable SSL flag, and then running that Nx command from the project root. E.g.: `nx run <name-of-app>:dev` if your `dev` command in  `<name-of-app>` under "scripts" is set to ``next dev --experimental-https`.

mandarini:
Hi there @naofel1 ! Thank you SO much for your PR! I rebased with `master` since we already have that feature in `master`, hence the PR was autoclosed. 

THANK YOU SO MUCH for your time!

PR #19704, opened 2023-10-18
feat(core): support changelog customizations and project changelogs

JamesHenry:
## Current Behavior

`release` config is not included in the JSON schema

AND

- It's not possible to customize generated changelogs in any way
- It's not possible to have project level changelogs
- It's not possible to set `version` config at the root of release config to be shared across all groups
- It's not possible to set `createRelease` or the tag pattern to use for releases in config
- It's not possible to create github releases per project

## Expected Behavior

`release` config is included in the JSON schema

AND

As of this PR, you can now:

- Fully customize changelogs via providing a custom renderer implementation which takes parsed commits and other metadata and returns the contents
- Tweak the renderer's output via `renderOptions` (added one `includeAuthors` option for now for our built in renderer)
- Set `version` config at the root of release config to be shared across all groups
- Set `createRelease`, `file` and `releaseTagPattern` at both the root and group level of config
- Create github releases per project (this will be much more applicable once we support independently versioned projects)

NOTE: There are breaking changes to the supported CLI arguments for `nx release changelog`:
- `--file`, `--tag-version-prefix ` and `--create-release` no longer exist (replaced by config options)
- `--interactive` is no longer a boolean, it takes in an enum string of either `all`, `workspace` or `projects` to control which projects you want to run interactively

image

^ custom `releaseTagPattern` per group and project level github releases in action on this branch


e2e/release/src/release.test.ts
```diff
@@ -122,10 +123,10 @@ describe('nx release', () => {
     const changelogOutput = runCLI(`release changelog 999.9.9`);
     expect(changelogOutput).toMatchInlineSnapshot(`
 
-      >  NX   Generating a CHANGELOG.md entry for v999.9.9
+      >  NX   Generating an entry in CHANGELOG.md for v999.9.9
```
JamesHenry, line 126: I switched round the structure of the log, because now that the path is fully customizable it could end up being reasonably long (depending on the level of nesting within the tree) and for longer paths it reads better this new way round

e2e/release/src/release.test.ts
```diff
@@ -544,6 +545,98 @@ describe('nx release', () => {
...
+      + ### ❤️  Thank You
+      +
+      + - Test
+      +
+      ## 999.9.9
+
+
+
+
+      >  NX   Previewing a Github release and an entry in {project-name}/CHANGELOG.md for {project-name}@v1000.0.0-next.0
```
JamesHenry, line 605: Just a reminder that the `{project-name}` here is normalized by the e2e snapshot serializer. This produces the correct project name in real usage

FrozenPandaz (approved):

PR #19715, opened 2023-10-18
docs(core): add nx console troubshooting guide

MaxKless:
This PR adds a troubleshooting guide for Nx Console to nx.dev.
We've been having a lot of issues submitted about VSCode + nvm so it's good to have a place to point folks to and also one they might find through google.
This will also be linked to from an Nx Console feature I'm building to make it even more discoverable.

image

### DIFF
--- a/docs/shared/recipes/console-troubleshooting.md
+++ b/docs/shared/recipes/console-troubleshooting.md
@@ -1,0 +1,17 @@
+# Nx Console Troubleshooting
+
+## VSCode + nvm Issues
+
+VSCode loads a version of Node when it starts. It can use versions set via [`nvm`](https://github.com/nvm-sh/nvm) but there's some caveats.
+
+- If you've installed Node outside of `nvm` (for example using their installer or `brew` on Mac), VSCode will always use that version. You can check by running `nvm list` and looking for a `system` alias. To enable VSCode to pick up your `nvm` version, make sure to uninstall it.
+- VSCode will load the `default` alias from `nvm` at startup. You can set it by running `nvm alias default [version]`. To make sure VSCode can load the correct value needs to be set in your OS' default terminal. Setting it in a VSCode-integrated terminal won't persist after it's closed. Similarly, setting it in a third-party app like iTerm won't influence VSCode by default.
+- VSCode only loads the `default` version when the app is first started. This means that in order to change it, you need to close all VSCode windows and restart the app - running `Reload Window` won't work.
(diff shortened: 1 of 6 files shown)


docs/shared/recipes/console-troubleshooting.md
```diff
@@ -0,0 +1,17 @@
+# Nx Console Troubleshooting
+
+## VSCode + nvm Issues
+
+VSCode loads a version of Node when it starts. It can use versions set via [`nvm`](https://github.com/nvm-sh/nvm) but there's some caveats.
```
isaacplmann, line 5:
```suggestion
VSCode loads a version of Node when it starts. It can use versions set via [`nvm`](https://github.com/nvm-sh/nvm) but there are some caveats.
```

docs/shared/recipes/console-troubleshooting.md
```diff
@@ -0,0 +1,17 @@
...
+
+- If you've installed Node outside of `nvm` (for example using their installer or `brew` on Mac), VSCode will always use that version. You can check by running `nvm list` and looking for a `system` alias. To enable VSCode to pick up your `nvm` version, make sure to uninstall it.
```
isaacplmann, line 7:
```suggestion
- If you've installed Node outside of `nvm` (for example using the Node installer or `brew` on Mac), VSCode will always use that version. You can check by running `nvm list` and looking for a `system` alias. To enable VSCode to pick up your `nvm` version, make sure to uninstall the version of Node that was installed outside of `nvm`.
```

docs/shared/recipes/console-troubleshooting.md
```diff
@@ -0,0 +1,17 @@
...
+- VSCode will load the `default` alias from `nvm` at startup. You can set it by running `nvm alias default [version]`. To make sure VSCode can load the correct value needs to be set in your OS' default terminal. Setting it in a VSCode-integrated terminal won't persist after it's closed. Similarly, setting it in a third-party app like iTerm won't influence VSCode by default.
```
isaacplmann, line 8:
```suggestion
- VSCode will load the `default` alias from `nvm` at startup. You can set it by running `nvm alias default [version]`. The `default` alias needs to be set in your OS' default terminal for VSCode to pick it up. Setting it in a VSCode-integrated terminal won't persist after it's closed. Similarly, setting it in a third-party app like iTerm won't influence VSCode by default.
```

isaacplmann (changes requested):
Small changes

isaacplmann (approved):

PR #19721, opened 2023-10-18
fix(react): fix creating a react library with cwd

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/src/generators/library/library.ts
+++ b/packages/react/src/generators/library/library.ts
@@ -8,6 +8,7 @@ import {
   Tree,
   updateJson,
 } from '@nx/devkit';
+import { getRelativeCwd } from '@nx/devkit/src/generators/artifact-name-and-directory-utils';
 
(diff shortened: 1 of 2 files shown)


mandarini (approved):

jaysoo (approved):

packages/react/src/generators/library/library.ts
```diff
@@ -171,9 +173,15 @@ export async function libraryGeneratorInternal(host: Tree, schema: Schema) {
   }
 
   if (options.component) {
+    const relativeCwd = getRelativeCwd();
```
jaysoo, line 176: I think we can just inline the logic of `getRelativeCwd` here if needed?

jaysoo: Disregard, the other utils are also exported like `getCwd`.

FrozenPandaz (approved):

PR #19726, opened 2023-10-19
fix(js): use workspace rootDir while compiling TS files

### LINKED ISSUE #11583: library importing other library using wildcard path mapping fails with "is not under 'rootDir'"
## Current Behavior
When one library imports from another, I get `.../packages/lib1/src/lib/lib1.ts' is not under 'rootDir' 'packages/lib2'`
Note that this doesn't happen if an app imports the library in the same way - presumably because an app uses executor `@nrwl/node:webpack`  rather the a library's `@nrwl/js:tsc`.

## Expected Behavior
library should import. This is a new project, no migration, therefore it's not the same like https://github.com/nrwl/nx/issues/10785#issuecomment-1159017136.

## Steps to Reproduce
1. Start from scratch:
```
npx create-nx-workspace badlibs --preset=ts --nxCloud==false
cd badlibs
nx generate @nrwl/js:library --name=lib1 --buildable
nx generate @nrwl/js:library --name=lib2 --buildable
```
2. Add the following to `tsconfig.base.json`:
```
    "paths": {
      "lib1/*": ["packages/lib1/*"]
    }
```
3. Add the following to `packages\lib2\src\lib\lib2.ts`:
```
import { lib1 } from 'lib1/src/lib/lib1';
```
4. Build all by running: `nx affected:build -all --skip-nx-cache`

Result:
```
    ×  nx run lib2:build
       Compiling TypeScript files for project "lib2"...
packages/lib2/src/lib/lib2.ts:1:22 - error TS6059: File 'C:/mono/badlibs/packages/lib1/src/lib/lib1.ts' is not under 'rootDir' 'packages/lib2'. 'rootDir' is expected to contain all source files.

       1 import { lib1 } from 'lib1/src/lib/lib1';
```

### Environment
```
   Node : 14.20.0
   OS   : win32 x64
   npm  : 6.14.17

   nx : 14.5.6
   @nrwl/angular : Not Found
   @nrwl/cypress : Not Found
   @nrwl/detox : Not Found
   @nrwl/devkit : 14.5.6
   @nrwl/eslint-plugin-nx : 14.5.6
   @nrwl/express : Not Found
   @nrwl/jest : 14.5.6
   @nrwl/js : 14.5.6
   @nrwl/linter : 14.5.6
   @nrwl/nest : Not Found
   @nrwl/next : Not Found
   @nrwl/node : Not Found
   @nrwl/nx-cloud : Not Found
   @nrwl/nx-plugin : Not Found
   @nrwl/react : Not Found
   @nrwl/schematics : Not Found
   @nrwl/storybook : Not Found
   @nrwl/web : Not Found
   @nrwl/workspace : 14.5.6
   typescript : 4.7.4
   Local workspace plugins:
   Community plugins:
```

kopach:
Fixes #11583

Tested this approach for some time already on real project using [pnpm's patch mechanism](https://pnpm.io/cli/patch)

### DIFF
--- a/packages/workspace/src/utilities/typescript/compilation.ts
+++ b/packages/workspace/src/utilities/typescript/compilation.ts
@@ -145,7 +145,7 @@ function getNormalizedTsConfig(options: TypeScriptCompilationOptions) {
   const tsConfig = readTsConfig(options.tsConfig);
   tsConfig.options.outDir = options.outputPath;
   tsConfig.options.noEmitOnError = true;
-  tsConfig.options.rootDir = options.rootDir;
+  tsConfig.options.rootDir = '.';
   if (tsConfig.options.incremental && !tsConfig.options.tsBuildInfoFile) {
     tsConfig.options.tsBuildInfoFile = joinPathFragments(
       options.outputPath,


packages/workspace/src/utilities/typescript/compilation.ts
```diff
@@ -145,7 +145,7 @@ function getNormalizedTsConfig(options: TypeScriptCompilationOptions) {
   const tsConfig = readTsConfig(options.tsConfig);
   tsConfig.options.outDir = options.outputPath;
   tsConfig.options.noEmitOnError = true;
-  tsConfig.options.rootDir = options.rootDir;
+  tsConfig.options.rootDir = '.';
```
kopach, line 148: the fix is based on conversation from here: https://github.com/nrwl/nx/issues/11583
The only problem is that `options.root` (as suggested in comments) are always `undefined`. Tested and 
```
tsConfig.options.rootDir = undefined;
```
Works as well. Decided to use `'.'` instead of `undefined` as similar approach is used in several other places in codebase

kopach:
hi @FrozenPandaz, could you, please, advice how can I handle the CI failing issue? I'ts not clear for me what exactly failing and why

mandarini:
@leosvelperez I am asking for your review, because I think you were the one working on the "file is not under `rootDir`" issues.

leosvelperez:
hey @kopach! Thanks for taking the time to contribute, and apologies for the delay in reviewing it.

I'm going to close this since the current implementation is working as expected. All projects are built with the `rootDir` set to the project root or to what the user provides in the `@nx/js:tsc` executor options. The linked issue occurs due to a manual misconfiguration. I closed that issue by explaining this and how to solve it: https://github.com/nrwl/nx/issues/11583#issuecomment-1917550914.

Other issues could result in the same error, but they are also solved differently and without the need to change the implementation.

PR #19727, opened 2023-10-19
feat(react-native): run pod commands with bundler

### LINKED ISSUE #19698: @nx/react-native:pod-install cannot be run with bundler

## Description
I propose the addition of a feature to the project that would allow users to run the Cocoapods `pod install` command via Bundler, using the syntax `bundle exec pod install` as opposed to the current implementation which requires running `pod install` directly. This implementation would encapsulate the Cocoapods environment within a Bundler context, which can provide a more consistent and controlled environment for dependency management.

## Motivation
Running `pod install` via Bundler using `bundle exec pod install` can offer a number of advantages over running `pod install` directly. These advantages include:
1. **Dependency Isolation:** Bundler provides a consistent environment for projects by tracking and installing the exact gems and versions that are needed.
2. **Avoiding Version Conflicts:** Bundler prevents environment pollution by ensuring that the gems you need are present in the development, staging, and production stages.

## Suggested Implementation
1. Introduce a new configuration option within the project settings to allow users to specify their preference for running `pod install` via Bundler.
2. Update the relevant scripts and documentation to check for this configuration option and run `pod install` commands via `bundle exec` where specified.

## Alternate Implementations
N/A

ste7en:
closed #19698

## Current Behavior

`pod install` command is run with the local installed version of cocoapods. This could lead to dependency misalignment or inconsistencies.

## Expected Behavior

Allow the possibility to handle the cocoapods version by Gem and run the `pod install` command with `bundle exec` to ensure everyone in a team use the same package versions.

## Related Issue(s)

Fixes #19698

### DIFF
--- a/packages/react-native/src/utils/pod-install-task.ts
+++ b/packages/react-native/src/utils/pod-install-task.ts
@@ -28,10 +28,12 @@ export function runPodInstall(
     buildFolder?: string;
     repoUpdate?: boolean;
     deployment?: boolean;
+    useBundler?: boolean;
   } = {
     buildFolder: './build',
     repoUpdate: false,
     deployment: false,
+    useBundler: false,
   }
 ): GeneratorCallback {
   return () => {
@@ -57,10 +59,12 @@ export function podInstall(
     buildFolder?: string;
     repoUpdate?: boolean;
     deployment?: boolean;
+    useBundler?: boolean;
   } = {
     buildFolder: './build',
     repoUpdate: false,
     deployment: false,
+    useBundler: false,
   }
 ) {
   try {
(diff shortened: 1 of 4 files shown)


packages/react-native/src/utils/pod-install-task.ts
```diff
@@ -70,15 +74,15 @@ export function podInstall(
...
-      `pod install ${options.repoUpdate ? '--repo-update' : ''} ${
-        options.deployment ? '--deployment' : ''
-      }`,
-      {
-        cwd: iosDirectory,
-        stdio: 'inherit',
-      }
-    );
+    const podCommand = [
+      options.useBundler ? 'bundle exec' : 'pod install',
```
xiongemi, line 78: question: should the command be `bundle exec pod install`? from the logic, if the useBundler is true, it only contains `bundle exec`.
should it be:
```
options.useBundler ? 'bundle exec' : '',
'pod install',
```

ste7en: Yeah, my mistake. Fixing it

xiongemi (approved):

PR #19732, opened 2023-10-19
fix(linter): add migration for eslint config prettier

mandarini:
## Current Behavior

No migration existed. User `nx migrate`s to 17, `npm i` gives version mismatch error because `eslint-config-prettier` is still on v8.

## Expected Behavior

Update `eslint-config-prettier` when `nx migrate 17`.

Related PR: https://github.com/nrwl/nx/pull/19232


JamesHenry (approved):

packages/eslint/migrations.json
```diff
@@ -74,6 +74,14 @@
           "version": "^5.60.1"
         }
       }
+    },
+    "17.0.0": {
+      "version": "17.0.0-beta.0",
```
mandarini, line 79: I'll do the same here @FrozenPandaz

FrozenPandaz, line 79: Can you change this to `17.0.0-rc.2` please?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #19733, opened 2023-10-19
docs(linter): use correct plugin name in migrating-to-flat-eslint.md

benkroeger:
the command in the docs used `@nx/eslint` as plugin name instead of `@nx/linter`

## Current Behavior

n.a.

## Related Issue(s)

n.a.


isaacplmann:
Thanks for the PR.  We accidentally merged a docs change too early.  What is currently on nx.dev is actually accurate for Nx 17, which should be released later today.

benkroeger:
ok. cool. never mind then

Issue #19457, opened 2023-10-05
Dependency on "@parcel/watcher": 2.0.4 breaks docker buildx on apple silicon.


LeunensMichiel:
### Current Behavior

We have recently introduced NX to our monorepo. We build our application with docker like this:

```
docker buildx build \
	--tag my-app \
	--file packages/my-app/Dockerfile \
	.
```
The other devs don't have an issue as they're on windows, but I got this:

At the yarn install step in Dockerfile:
```
@parcel/watcher@npm:2.0.4 must be built because it never has been before or the last one failed
58.74 ➤ │ @parcel/watcher@npm:2.0.4 couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-3b5244bb/build.log)
58.74 ➤ │ nx@npm:16.10.0 [72f3e] must be built because it never has been before or the last one failed
```

When investigating further, it seems @parcel/watcher@2.0.4 has some issues with ARM/Apple Silicon. 

I fixed the error by adding 
```
"resolutions": {
    "@parcel/watcher": "2.3.0"
  },
```
in our package.json. This fixes the docker build.

### Expected Behavior

Don't fail at `yarn` in Dockerfile.

### Steps to Reproduce

1. Build dockerfile with yarn and NX (16.10.0)

### Nx Report

```shell
\
```

### Package Manager Version

yarn v3.6.4

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

dbjorge:
Dupe of https://github.com/nrwl/nx/issues/13973

Cammisuli:
fixed by #19751

LeunensMichiel:
Epic, thanks!

PR #19751, opened 2023-10-20
feat(core): remove parcel/watcher

Cammisuli:
## Current Behavior
Current daemon implementation has both a rust watcher and a parcel/watcher implementation. 

## Expected Behavior
Only the rust watcher should be used as it handles nested gitignores better

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -33,7 +33,6 @@
   },
   "homepage": "https://nx.dev",
   "dependencies": {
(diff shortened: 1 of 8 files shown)


dbjorge:
This should incidentally fix https://github.com/nrwl/nx/issues/19457 and https://github.com/nrwl/nx/issues/13973

FrozenPandaz (approved):

PR #19757, opened 2023-10-20
fix(core): do not filter updated files when the native watcher is in use

Cammisuli:
## Related Issue(s)

Fixes #


FrozenPandaz (approved):

MillerSvt:
@Cammisuli Can you publish patch with this fix for 16.x version?

PR #19766, opened 2023-10-20
feat(misc): remove usages of @nx/cypress:cypress-project internally

jaysoo:
Replace with call to `@nx/cypress:configuration`, consistent with how we setup Playwright projects for apps. The deprecated generator will be removed in Nx 18.

## Changes

- Internal usage of `@nx/cypress:cypress-project` generator has been replaced with `addProjectConfiguration` + `@nx/cypress:configuration`
- `@nx/storybook:cypress-project` generator updated to support `projectNameAndRootFormat` -- this was missed previously
- The default `tsconfig.json` file for the e2e project has a different `include` config. Previously, we scoped to `src/**/*.ts`, but now it is `**/*.ts`, and we also include JSX files for React/Vue.

## Follow-up
- The `@nx/cypress:configuration` generator does not support `as-provided` paths, so we need a follow-up PR to address this issue.
- The Angular e2e migrator uses the deprecated generator as well, but it isn't as straight-forward so we need to update this PR or have a follow-up PR to update it.


packages/storybook/src/generators/cypress-project/cypress-project.ts
```diff
@@ -30,31 +28,61 @@ export interface CypressConfigureSchema {
...
   skipFormat?: boolean;
+  projectNameAndRootFormat?: 'as-provided' | 'derived';
 }
 
 export async function cypressProjectGenerator(
   tree: Tree,
   schema: CypressConfigureSchema
 ) {
+  return await cypressProjectGeneratorInternal(tree, {
+    projectNameAndRootFormat: 'derived',
```
mandarini, line 39: we're keeping this hard-coded then?

mandarini (approved):

packages/storybook/src/generators/cypress-project/cypress-project.ts
jaysoo: Yeah, this is going away in v18 anyway, there's no point to change it.

packages/angular/src/generators/application/__snapshots__/application.spec.ts.snap
```diff
@@ -623,9 +628,12 @@ exports[`app --project-name-and-root-format=derived should generate correctly wh
...
   "extends": "../../tsconfig.base.json",
   "include": [
-    "src/**/*.ts",
-    "src/**/*.js",
+    "**/*.ts",
+    "**/*.js",
     "cypress.config.ts",
+    "**/*.cy.ts",
+    "**/*.cy.js",
+    "**/*.d.ts",
```
jaysoo, line 636: Wouldn't `**/*.ts` already include `**/*.cy.ts` and `**/*.d.ts` files? CC: @barbados-clemens do you know why it was set up this way?

AgentEnder (approved):
Files I own (under /cypress or /web) look good.

leosvelperez (approved):

leosvelperez:
PR updating the Angular e2e migrator: https://github.com/nrwl/nx/pull/19816.

packages/angular/src/generators/application/lib/add-e2e.ts
```diff
@@ -10,23 +10,27 @@ import {
...
+      projectType: 'application',
+      root: options.e2eProjectRoot,
+      sourceRoot: joinPathFragments(options.e2eProjectRoot, 'src'),
+      targets: {},
+      tags: [],
+      implicitDependencies: [options.name],
+    });
+    await configurationGenerator(tree, {
+      project: options.e2eProjectName,
       directory: options.e2eProjectRoot,
```
leosvelperez, line 29: I think this is wrong. The `directory` option in the `configurationGenerator` is meant to be and quote "A directory where the project is placed relative from the project root". The same applies to the rest of generators updated in this PR providing the project root as the directory.
```suggestion
```

leosvelperez (changes requested):

packages/angular/src/generators/application/__snapshots__/application.spec.ts.snap
jaysoo: Actually, I understand now. The component tests could be nested in `src` (default), so we're scoping to include those in addition to the project sources.

leosvelperez (approved):

PR #19827, opened 2023-10-24
docs(core): run tasks in parallel recipe

isaacplmann:
Split parallel section into its own recipe


juristr (approved):
pushed a commit to fix an issue with the main header formatting.

PR #19828, opened 2023-10-24
docs(core): change cache location recipe

isaacplmann:
Change cache location recipe


docs/generated/manifests/menus.json
```diff
@@ -2187,6 +2187,14 @@
                 "children": [],
                 "disableCollapsible": false
               },
+              {
+                "name": "Change Cache Location",
+                "path": "/recipes/tips-n-tricks/change-cache-location",
```
juristr, line 2192: can we move this to the new "tasks & caching" section? Whenever possible we shouldn't add stuff into the "tips-n-tricks"...that's like a sink hole for everything that hasn't a proper place. which sometimes makes sense, but only really if we don't have a better option

juristr (changes requested):

juristr (approved):

PR #19830, opened 2023-10-24
docs(core): installation recipes

isaacplmann:
Splits out 2 new recipes:
- Update global installation
- Install Nx in a non-javascript repo


juristr (changes requested):
Hmm... 🤔 I don't agree with this move. It is basically reverting one of my prior PRs to move them to the installation page. Why would we want to have a separate page?

some thoughts:
- if I search installation instructions, I'll click on the installation page. All that info is not really relevant anywhere else or linked from any other page, so I'd rather have a single one that can be discovered easily than spreading it out
- the recipes are not really tips-n-tricks, it's just installation instructions
- both new recipes are also not linked from the install page

We can def improve the current installation page, renaming from "self-managed" to "Non-JS Repos" which I like better and emphasizing more that we prefer a non-global installation because it usually leads to fewer issues (although it is really a preference thing IMHO).

juristr (approved):

PR #19840, opened 2023-10-25
feat(testing): add cypress create nodes plugin

FrozenPandaz:
## Current Behavior

There is no plugin to create projects based on `cypress.config.ts` files.

## Expected Behavior

There is a plugin which will add `cypress` targets/projects to the Nx project graph. This allows you to run `cypress` for `cypress.config.ts` files in the repo without creating a `project.json` file.

To use the plugin, add the following to `nx.json`:
```json
  "plugins": [
    {
      "plugin": "@nx/cypress/plugin",
      "options": {
        "targetName": "e2e",
        "componentTestingTargetName": "component-test"
      }
    }
  ],
  ```
  
The plugin will analyze `cypress.config.ts` files and create targets for those projects without needing to specify the targets in `project.json`.

This means, adding `nx` to a workspace that already uses `cypress` is as easy as running `nx init` and then adding the config above to `nx.json`.

The target will be configured with caching setup. :tada: 

For existing Nx workspaces, the cypress preset has been updated to start the dev server (via nx) within the Cypress config itself. This means that running `cypress run` will start the dev server :tada: without having to run the tests via `nx e2e myapp-e2e`. This means that editor plugins that run Cypress should work and the approach is more idiomatic with Cypress altogether.

The config looks like this:

```ts
import { nxE2EPreset } from '@nx/cypress/plugins/cypress-preset';

import { defineConfig } from 'cypress';

export default defineConfig({
  e2e: {
    ...nxE2EPreset(__filename, {
      cypressDir: 'src',
      devServerTargets: {
        default: 'react-test:serve:development',
        production: 'react-test:serve:production',
      },
      ciDevServerTarget: 'react-test:serve-static',
    }),
  },
});
```

The `devServerTargets` from above become configurations onto the target configured by `@nx/cypress/plugin`.

You could try the above with `NX_PCV3=true npx --yes create-nx-workspace@latest`

This is not the default yet but there is a migration slated for `17.2.0` at the moment and that would also be when this becomes the default.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/cypress/src/generators/configuration/configuration.ts
+++ b/packages/cypress/src/generators/configuration/configuration.ts
@@ -1,11 +1,14 @@
 import {
   addDependenciesToPackageJson,
+  createProjectGraphAsync,
   formatFiles,
   generateFiles,
   GeneratorCallback,
   joinPathFragments,
   offsetFromRoot,
   parseTargetString,
+  ProjectGraph,
+  readNxJson,
   readProjectConfiguration,
   runTasksInSerial,
   toJS,
@@ -52,11 +55,20 @@ export async function configurationGenerator(
   if (!installedCypressVersion()) {
     tasks.push(await cypressInitGenerator(tree, opts));
   }
+  const projectGraph = await createProjectGraphAsync();
+  const nxJson = readNxJson(tree);
+  const hasPlugin = nxJson.plugins?.some((p) =>
+    typeof p === 'string'
+      ? p === '@nx/cypress/plugin'
+      : p.plugin === '@nx/cypress/plugin'
+  );
   if (opts.bundler === 'vite') {
     tasks.push(addDependenciesToPackageJson(tree, {}, { vite: viteVersion }));
   }
-  await addFiles(tree, opts);
-  addTarget(tree, opts);
+  await addFiles(tree, opts, projectGraph, hasPlugin);
+  if (!hasPlugin) {
+    addTarget(tree, opts);
+  }
 
   const linterTask = await addLinterToCyProject(tree, {
     ...opts,
@@ -89,18 +101,29 @@ In this case you need to provide a devServerTarget,'<projectName>:<targetName>[:
 
   options.directory ??= 'src';
 
+  const devServerTarget =
+    options.devServerTarget ??
+    (projectConfig.targets.serve ? `${options.project}:serve` : undefined);
+
+  if (!options.baseUrl && !devServerTarget) {
+    throw new Error('Either baseUrl or devServerTarget must be provided');
+  }
+
   return {
     ...options,
     bundler: options.bundler ?? 'webpack',
     rootProject: options.rootProject ?? projectConfig.root === '.',
     linter: options.linter ?? Linter.EsLint,
-    devServerTarget:
-      options.devServerTarget ??
-      (projectConfig.targets.serve ? `${options.project}:serve` : undefined),
+    devServerTarget,
   };
 }
 
-async function addFiles(tree: Tree, options: NormalizedSchema) {
+async function addFiles(
+  tree: Tree,
+  options: NormalizedSchema,
+  projectGraph: ProjectGraph,
+  hasPlugin: boolean
+) {
   const projectConfig = readProjectConfiguration(tree, options.project);
   const cyVersion = installedCypressVersion();
   const filesToUse = cyVersion && cyVersion < 10 ? 'v9' : 'v10';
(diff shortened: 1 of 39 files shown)


packages/cypress/src/generators/configuration/configuration.ts
```diff
@@ -52,11 +54,19 @@ export async function configurationGenerator(
...
+      ? p === '@nx/cypress/plugin'
+      : p.plugin === '@nx/cypress/plugin'
+  );
   if (opts.bundler === 'vite') {
     tasks.push(addDependenciesToPackageJson(tree, {}, { vite: viteVersion }));
   }
-  await addFiles(tree, opts);
-  addTarget(tree, opts);
+  await addFiles(tree, opts, hasPlugin);
+  if (!hasPlugin) {
```
jaysoo, line 67: Should we mark this for deprecation in Nx 18 or 19?

FrozenPandaz: Hm.. probably? But not yet.. let's see how it goes first.

packages/cypress/src/utils/config.ts
```diff
@@ -29,27 +32,48 @@ export async function addDefaultE2EConfig(
...
-        ? `nxE2EPreset(__filename, { cypressDir: '${options.directory}', bundler: 'vite' })`
-        : `nxE2EPreset(__filename, { cypressDir: '${options.directory}' })`;
+        ? `nxE2EPreset(__filename, { cypressDir: '${options.cypressDir}', bundler: 'vite' })`
+        : `nxE2EPreset(__filename, { cypressDir: '${options.cypressDir}' })`;
 
     updatedConfigContents = tsquery.replace(
       cyConfigContents,
       `${TS_QUERY_EXPORT_CONFIG_PREFIX} ObjectLiteralExpression:first-child`,
       (node: ObjectLiteralExpression) => {
+        let baseUrlContents = baseUrl ? `,\nbaseUrl: '${baseUrl}'` : '';
```
jaysoo, line 42: Is there a case where `baseUrl` is falsy? Seems like it should always be defined?

packages/cypress/src/migrations/update-17-2-0/add-nx-metadata.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+          },
+        },
+      },
+    });
+    await update(tree);
+
+    expect(tree.read('e2e/cypress.config.ts', 'utf-8')).toMatchSnapshot();
+  });
+
+  it('should nto add nx metadata for if there are none to add', async () => {
```
jaysoo, line 74: `s/nto/not`

packages/cypress/src/migrations/update-17-2-0/add-nx-metadata.ts
```diff
@@ -0,0 +1,71 @@
...
+    const e2eTarget: TargetConfiguration<CypressExecutorOptions> =
+      projectConfig.targets?.e2e;
+
+    if (!e2eTarget || e2eTarget.executor !== '@nx/cypress:cypress') {
+      continue;
+    }
+
+    const nxMetadata: NxCypressMetadata = {};
+
+    nxMetadata.devServerTarget = e2eTarget.options?.devServerTarget;
```
jaysoo, line 43: I'm wondering if having these three versions of `devServerTarget` makes sense or not. It's fixed to what we generate for users, but they customize the setup.

Maybe something like:

```js
export const nx: NxCypressMetadata = {
  devServerTarget: process.env.NX_CYPRESS_DEV_SERVER_TARGET ?? 'myapp:serve'
}
```

We can also skip the whole `nx` metadata, and document that users can change the target with `NX_CYPRESS_DEV_SERVER_TARGET` or something. Similar thing for ports, I think expecting `PORT` to be set is pretty standard

Or something like this if we are keen to generate separate `configuration`.

```js
export const nx: NxCypressMetadata = {
  devServerTarget: 'myapp:serve',
  configurations: {
    production: 'myapp:serve:production',
    ci: 'myapp:static-serve',
  }
}
```

So it's more explicit about generating support for `--configuration=production` and `--configuration=ci`, and is able to support additional executor options if we want to.

packages/cypress/src/migrations/update-17-2-0/add-nx-cypress-plugin.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  const rootMappings = createProjectRootMappingsFromProjectConfigurations(proj);
+
+  replaceProjectConfigurationsWithPlugin(
+    tree,
+    rootMappings,
+    '@nx/cypress/plugin',
+    createNodes,
+    {
+      targetName: 'e2e',
```
mandarini, line 18: I don't think this is safe, because someone may have named their target `katerina-e2e`. I think here we should use the `forEachExecutorOptions` to get the target names?

mandarini: so, if a project has a target defined as such
```
   "e2e-katerina": {
      "executor": "@nx/cypress:cypress",
      "options": {
...
      },
      "configurations": {
  ...
      }
    },
```

our migrator would create a new `e2e` node correctly, but it would not remove the `e2e-katerina` from `project.json`.

mandarini: there's an issue though if one has multiple cypress projects, each one with different target names. for a migration to work as intended, all target names of all projects should be the same 🤔

mandarini: I don't think this makes sense either, I'll suggest a way to handle this in new PR, and see if we care for such complexity.

mandarini: maybe I'm just overthinking this

packages/cypress/src/utils/config.ts
FrozenPandaz: It is undefined when you provide a devServerTarget. In those cases, we grab the baseUrl from the devServerTarget

jaysoo (approved):

AgentEnder (approved):

packages/cypress/src/migrations/update-17-2-0/add-nx-cypress-plugin.ts
FrozenPandaz: Let's land this as is for now and discuss this more. We might be able to figure out the target name but let's not block this change.

PR #19847, opened 2023-10-25
docs(core): Using --directory=libs/is-odd instead of --directory=packages/is-odd

Yoann-TYT:
## Current Behavior

Using --directory=libs/is-odd instead of --directory=packages/is-odd

## Expected Behavior

We need to use the parameter --directory=packages/is-odd

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/npm-tutorial/integrated.md
+++ b/docs/shared/npm-tutorial/integrated.md
@@ -19,11 +19,11 @@ The file structure should look like this:
 
 ```treeview
 myorg/
-├── packages/
-├── tools/
-├── nx.json
-├── package.json
 ├── README.md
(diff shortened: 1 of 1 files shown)


mandarini:
Hi there @Yoann-TYT ! Thank you for your PR, and for the time and effort you put into this!

I will be closing this PR for the following reason: the `--directory` flag provides the user with the flexibility and options to name the `directory` where they want their project to live. In this example we chose the directory `libs`. If one prefers, they can use the directory `packages` or anything else they may want. It's just an example, so there's no right or wrong.

I am sorry for closing your PR! Thank you very very much for contributing!

PR #19866, opened 2023-10-26
fix(nx-plugin): move tests from 'tests' to 'src' directory

beeman:
## Current Behavior

When creating a plugin `my-plugin`, the tests in the `my-plugin-e2e` project are generated in `my-plugin-e2e/tests/my-plugin.spec.ts`. 

However, the file `tsconfig.spec.json` does not include the files in the `tests` directory but the `src` directory, as seen below:

```json
{
  "extends": "./tsconfig.json",
  "compilerOptions": { "outDir": "../dist/out-tsc", "module": "commonjs", "types": [ "jest", "node" ] },
  "include": [ "jest.config.ts", "src/**/*.test.ts", "src/**/*.spec.ts", "src/**/*.d.ts" ]
}
```

The issue with this is that you can't import code from other libraries in the `-e2e` project, for example an `e2e-util` lib.

This patch changes where these specs are generated and updates the unit test that verifies the location.

From looking at the `nrwl/nx` repo itself, it looks like this is the structure that's being used, the actual specs live in the `src/` directory.

## Expected Behavior

The code should live in a place where it's included so you can import from other libraries.

## Related Issue(s)

Fixes #
- I didn't create an issue but confirmed on [Discord](https://discord.com/channels/1143497901675401286/1166447781058183179) that this was indeed undesired behavior.


FrozenPandaz (approved):
I think it's better in `src`. Thank you for the contribution! :tada:

PR #19871, opened 2023-10-26
fix(vue): use default coverageProvider in @nx/vue:library generator when adding vitest

benkroeger:
## Current Behavior

When generating a library with @nx/vue, invokes the vitestGenerator with `coverageProvider: 'c8'`.
c8 has been deprecated. The new default is v8 (which is already used in @nx/vite:vitest)

## Expected Behavior

c8 shouldn't be used as coverageProvider. Instead use the default from @nx/vite:vitest

## Related Issue(s)

n.a.


mandarini:
Hey sorry I've been having my eye on this for some time, and I made this PR instead because there were more things to be addressed: https://github.com/nrwl/nx/pull/20156

thank you so much for contributing in any case!

PR #19873, opened 2023-10-26
fix(vue): do not add @vue/tsconfig dependency

benkroeger:
## Current Behavior

@nx/vue:init generator adds a dev dependency for @vue/tsconfig but this package is never used.

## Expected Behavior

do not install @vue/tsconfig when it's not used

## Related Issue(s)
n.a.


ndcunningham (changes requested):
Hey, thanks for contributing!

However, you need to do some additional updates.
references 
- `packages/vue/src/generators/init`

You need to update test snapshots for `init.spec` and `library.spec`

mandarini (approved):

jaysoo (approved):

ndcunningham (approved):

PR #19898, opened 2023-10-27
fix(angular): support scoped remote names for federation

Coly010:
## Current Behavior

We currently error if someone tries to use a remote name such as '@scope/remote'

## Expected Behavior

We should not error if someone tries to use a remote name with this

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/utils/mf/utils.ts
+++ b/packages/angular/src/utils/mf/utils.ts
@@ -89,15 +89,18 @@ export async function getModuleFederationConfig(
     projectGraph = await createProjectGraphAsync();
   }
 
-  if (!projectGraph.nodes[mfConfig.name]?.data) {
+  const NPM_SCOPE_REGEX = /^@(\w|\w-\w)+\/(?!\/)/;
+  const mfConfigName = mfConfig.name.replace(NPM_SCOPE_REGEX, '');
+
+  if (!projectGraph.nodes[mfConfigName]?.data) {
     throw Error(
-      `Cannot find project "${mfConfig.name}". Check that the name is correct in module-federation.config.js`
+      `Cannot find project "${mfConfigName}". Check that the name is correct in module-federation.config.js`
     );
   }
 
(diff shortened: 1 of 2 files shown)


ndcunningham (approved):

packages/angular/src/utils/mf/utils.ts
```diff
@@ -89,15 +89,18 @@ export async function getModuleFederationConfig(
     projectGraph = await createProjectGraphAsync();
   }
 
-  if (!projectGraph.nodes[mfConfig.name]?.data) {
+  const NPM_SCOPE_REGEX = new RegExp(/^@{1}(\w)+\/{1}(?!\/)/);
```
leosvelperez, line 92: nit: you can simplify this:
```suggestion
  const NPM_SCOPE_REGEX = /^@\w+\/(?!\/)/;
```

packages/webpack/src/utils/module-federation/get-remotes-for-host.ts
```diff
@@ -42,6 +42,8 @@ function extractRemoteProjectsFromConfig(
   const staticRemotes =
     config.remotes?.map((r) => (Array.isArray(r) ? r[0] : r)) ?? [];
   remotes.push(...staticRemotes);
+  const NPM_SCOPE_REGEX = new RegExp(/^@{1}(\w)+\/{1}(?!\/)/);
```
leosvelperez, line 45: nit: you can simplify this:
```suggestion
  const NPM_SCOPE_REGEX = /^@\w+\/(?!\/)/;
```

leosvelperez (approved):

packages/angular/src/utils/mf/utils.ts
Coly010: Yeah, that's simpler, also made me realise i missed a use case "@my-name/" which this slight modification to what you have suggested solves `^@(\w|\w-\w)+\/(?!\/)`

PR #19904, opened 2023-10-27
feat(angular): rename browserTarget to buildTarget in dev-server executors

leosvelperez:
Deprecates the `browserTarget` option (deprecated in Angular v17) in favor of `buildTarget`.

Our executors will only expose the `buildTarget` option in Nx v18 and they internally normalize the option according to the Angular version (Nx v18 will still support versions of Angular expecting "browserTarget"). That way, we have a unified single option for our executors public APIs, while we ensure that it works with the versions of Angular we support.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/migrations/update-17-2-0/browser-target-to-build-target.ts
+++ b/packages/angular/src/migrations/update-17-2-0/browser-target-to-build-target.ts
@@ -1,0 +1,75 @@
+import {
+  formatFiles,
+  readJson,
+  readProjectConfiguration,
+  updateProjectConfiguration,
+  writeJson,
+  type NxJsonConfiguration,
+  type Tree,
+} from '@nx/devkit';
+import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';
+
+export const executors = [
+  '@angular-devkit/build-angular:dev-server',
+  '@angular-devkit/build-angular:extract-i18n',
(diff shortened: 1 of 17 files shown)


leosvelperez:
The failing test is being addressed by https://github.com/nrwl/nx/pull/19925.

packages/angular/src/migrations/update-17-2-0/browser-target-to-build-target.ts
```diff
@@ -0,0 +1,75 @@
...
+        ? projectConfiguration.targets[target].configurations[configuration]
+        : projectConfiguration.targets[target].options;
+
+      updateConfig(config);
+
+      updateProjectConfiguration(tree, project, projectConfiguration);
+    });
+
+    // update options from nx.json target defaults
+    const nxJson = readJson<NxJsonConfiguration>(tree, 'nx.json');
```
Coly010, line 37: Is there not a `readNxJson(tree)` util that could be used here?

Coly010 (approved):
One nitpick, looks good

yjaaidi:
This is exciting!! 

It would be nice if we could generate new apps using the new `@angular-devkit/build-angular:application` builder instead of the `@angular-devkit/build-angular:browser-esbuild` which is a compatibility one that will be dropped later.

Of course, this could come later in another version.

PR #19910, opened 2023-10-27
feat(core): change exec to run adhoc tasks

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/cli/exec.md
+++ b/docs/generated/cli/exec.md
@@ -17,11 +17,13 @@ Install `nx` globally to invoke the command directly using `nx`, or use `npx nx`
 
 ## Options
 
-### configuration
+### all
 
-Type: `string`
+Type: `boolean`
 
-This is the configuration to use when performing tasks on projects
+Default: `true`
+
+[deprecated] `run-many` runs all targets on all projects in the workspace if no projects are provided. This option is no longer required.
 
 ### exclude
 
@@ -34,6 +36,12 @@ Exclude certain projects from being processed
 Type: `string`
 
 Show the task graph of the command. Pass a file path to save the graph data instead of viewing it in the browser.
+
+### help
+
+Type: `boolean`
+
+Show help
 
 ### nxBail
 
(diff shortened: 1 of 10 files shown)


docs/generated/cli/exec.md
```diff
@@ -55,7 +63,7 @@ Ignore cycles in the task graph
 
 Type: `string`
 
-Choices: [dynamic, static, stream, stream-without-prefixes, compact]
+Choices: [dynamic, static, stream, stream-without-prefixes]
```
FrozenPandaz, line 66: Can this be removed from this? It doesn't get read.

packages/nx/src/commands-runner/command-graph-utils.ts
```diff
@@ -0,0 +1,65 @@
+import { CommandGraph } from './command-graph';
+
+function _findCycle(
```
FrozenPandaz, line 3: Make the task graph utils typings more generic and use the same functions

packages/nx/src/commands-runner/remove-projects-from-command-grpah.ts
```diff
@@ -0,0 +1,22 @@
+import { CommandGraph } from './command-graph';
+
+export function removeProjectsFromCommandGraph(
```
FrozenPandaz, line 3: Make the task graph utils typings more generic and use the same functions

packages/nx/src/commands-runner/create-command-graph.ts
```diff
@@ -0,0 +1,51 @@
...
+  nxArgs: NxArgs
+) {
+  const dependencies: Record<string, string[]> = {};
+  for (const projectName of projectNames) {
+    if (projectGraph.dependencies[projectName].length >= 1) {
+      dependencies[projectName] = [
+        ...new Set(
+          projectGraph.dependencies[projectName].map(
+            (projectDep) => projectDep.target
+          )
```
FrozenPandaz, line 18: Filter out projects that are not workspace nodes because they don't have commands associated with them.

FrozenPandaz (changes requested):

docs/generated/cli/exec.md
xiongemi: removed

packages/nx/src/commands-runner/command-graph-utils.ts
xiongemi: changed

packages/nx/src/commands-runner/remove-projects-from-command-grpah.ts
xiongemi: changed the original function

FrozenPandaz (approved):
Looks good. :+1:

PR #19926, opened 2023-10-30
fix(angular): allow numbers following a dash in application prefix

leosvelperez:
For reference: https://github.com/angular/angular-cli/commit/c5827d412f821727bbb8c6f885b35b717d607a82#diff-21e22640a309b86842754f526a37c110b6fefd87a11f0bec50f5921035489783

## Related Issue(s)

Fixes #


Coly010 (approved):

Coly010:
I think there is an e2e test that checks failure of `my-1-app` in CNW that you may need to look at

PR #19937, opened 2023-10-30
feat(angular): generate applications with the esbuild-based application builder

leosvelperez:
- Add prompt to CNW for `--bundler` for Angular applications.
- Update Angular app generator to default to `esbuild` for the bundler.
- Generate build target for apps with the `@angular-devkit/build-angular:application` builder instead of the `@angular-devkit/build-angular:browser-esbuild`.

## Related Issue(s)

Fixes #


Coly010 (commented):
LGTM

PR #19939, opened 2023-10-30
fix: remove `@` character from templates for compat with Angular cont…

yjaaidi:
…rol flow

This fixes `error NG5002: Incomplete block "nx". If you meant to write the @ character, you should use the "&#64;" HTML entity instead.` error on freshly generated Angular apps with routing.

## Related Issue(s)

Fixes #


yjaaidi:
Oops! Committed from GitHub directly with invalid commit message.
Will fix tomorrow

leosvelperez:
hey @yjaaidi! Thanks for the contribution!

I'm going to close this since, AFAIK, this is not an issue in the currently supported versions of Angular. The new control flow syntax will only be available in Angular v17. We're tracking Angular v17 support in a separate branch where this is already fixed.

PR #19942, opened 2023-10-30
chore(devkit): add util for migrating to a plugin

FrozenPandaz:
## Current Behavior

There is no migration for adding plugins

## Expected Behavior

There is a util to write migrations for adding plugins

## Related Issue(s)

Fixes #


AgentEnder (approved):
Looks good, thanks 🎉

PR #19972, opened 2023-11-01
feat(vite): Use app/lib tsconfig for buildable paths resolution

pulecka:
This change uses actual project TypeScript config - i.e. `tsconfig.app.json` or `tsconfig.lib.json` - to generate temporary `tsconfig` file when building with `"buildLibsFromSource": false` option.

This is needed when users would want to use the generated `tsconfig` in their Vite configuration - e.g. for generating type declarations.

Also exposed the generated `tsconfig` via `NX_TSCONFIG_PATH` env variable. This way users trying to use the generated config don't have to know & depend on specific path which feels like an internal implementation detail.

So now user can use `vite-plugin-dts` or `@rollup/plugin-typescript` to generate type declarations during build:

```
export default defineConfig({
  plugins: [
    nxViteTsPaths(),
    {
      ...typescript({
        tsconfig: process.env.NX_TSCONFIG_PATH ?? join(__dirname, 'tsconfig.lib.json'),
        noForceEmit: true,
      }),
      apply: 'build',
    },
  ],
})
```

## Current Behavior
Generated temporary `tsconfig` file is based on project `tsconfig.json` file which is shared in typical NX setups for tests & build. And as such it is missing some options that are relevant only for build - e.g. `emitDeclarationOnly`.

### DIFF
--- a/packages/vite/src/utils/executor-utils.ts
+++ b/packages/vite/src/utils/executor-utils.ts
@@ -1,5 +1,5 @@
 import { printDiagnostics, runTypeCheck } from '@nx/js';
-import { join, resolve } from 'path';
+import { join } from 'path';
 import { ViteBuildExecutorOptions } from '../executors/build/schema';
 import { ExecutorContext } from '@nx/devkit';
 import { ViteDevServerExecutorOptions } from '../executors/dev-server/schema';
@@ -7,6 +7,7 @@ import {
   calculateProjectBuildableDependencies,
   createTmpTsConfig,
 } from '@nx/js/src/utils/buildable-libs-utils';
+import { getProjectTsConfigPath } from './options-utils';
 
 export async function validateTypes(opts: {
   workspaceRoot: string;
@@ -31,7 +32,7 @@ export function createBuildableTsConfig(
   options: ViteBuildExecutorOptions | ViteDevServerExecutorOptions,
   context: ExecutorContext
 ) {
-  const tsConfig = resolve(projectRoot, 'tsconfig.json');
+  const tsConfig = getProjectTsConfigPath(projectRoot);
   options['buildLibsFromSource'] ??= true;
 
   if (!options['buildLibsFromSource']) {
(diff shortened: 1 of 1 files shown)


packages/vite/src/utils/executor-utils.ts
```diff
@@ -45,7 +46,14 @@ export function createBuildableTsConfig(
...
-    createTmpTsConfig(tsConfig, context.root, projectRoot, dependencies);
+    // This tsconfig is used via the Vite ts paths plugin.
+    // It can be also used by other user-defined Vite plugins (e.g. for creating type declaration files).
+    const tmpTsConfigPath = createTmpTsConfig(
+      tsConfig,
+      context.root,
+      projectRoot,
+      dependencies
+    );
+    process.env.NX_TSCONFIG_PATH = tmpTsConfigPath;
```
belaczek, line 57: When `options.buildLibsFromSource === true`, `process.env.NX_TSCONFIG_PATH` would be undefined. Is that desired?

pulecka: it's expected - `process.env.NX_TSCONFIG_PATH` would also be undefined during `test` or `serve` in case of app.
so the consumer would have to always check it like:
```
const tsconfigPath = process.env.NX_TSCONFIG_PATH ?? join(__dirname, 'tsconfig.lib.json')
```

whether it's desired I'm not that certain - it seems to me to be better for consumers than doing something like:
```
const tmpTsconfigPath = join(
  workspaceRoot,
  'tmp',
  projectRootFromWorkspaceRoot,
  'tsconfig.generated.json'
)
const tsconfigPath = existsSync(tmpTsconfigPath) ? tmpTsconfigPath : join(__dirname, 'tsconfig.lib.json')
```

both of these are leaking implementation details, but the first one is a little less leaky imho

mandarini (approved):
Ok sure I'll approve this! Thanks for the PR!

mandarini:
@pulecka I cannot force-push on your branch. Can you please check the "allow commits by maintainers" or whatever is that setting?

mandarini:
@belaczek you either need to rebase with `master` to resolve conflicts, or give me push rights on your branch and I can do the rebase! then we can merge!

pulecka:
@mandarini @belaczek rebased with latest master.

PR #19987, opened 2023-11-01
feat(module-federation): use nx run-many to build static remotes in parallel

Coly010:
- feat(angular): use nx run-many to build remotes in parallel
- feat(react): use nx run-many to build remotes in parallel

## Current Behavior

We currently spin up a new forked process for every build of a static remote. This impacts resources and memory consumption when there are a lot of remotes in the workspace.

## Expected Behavior

We should use nx run-many instead to manage parallel builds of static remotes allowing for users to find the perfect number of parallel processes that they can use to build static remotes locally.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/angular/executors/module-federation-dev-server.json
+++ b/docs/generated/packages/angular/executors/module-federation-dev-server.json
@@ -122,6 +122,10 @@
         "description": "Whether the host that is running this executor is the first in the project tree to do so.",
         "default": true,
         "x-priority": "internal"
+      },
+      "parallel": {
+        "type": "number",
+        "description": "Max number of parallel processes for building static remotes"
       }
     },
     "additionalProperties": false,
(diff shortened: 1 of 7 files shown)


ndcunningham (commented):
Looks fine just not sure what `ANSII_CODE_REGEX` means.

ndcunningham (approved):
Looks fine just not sure what `ANSII_CODE_REGEX` means.

Coly010:
It's a regex to find the color codes that output certain strings in the terminal in different colors or as bold text. 

They aren't being stripped without that regex and therefore matching strings fails because

`Successfully ran target build` is actually `[32mSuccessfully ran target \u001b[1mbuild`

ndcunningham:
> It's a regex to find the codes that output certain strings in the terminal in different colors or as bold text.
> 
> They aren't being stripped without that regex and therefore matching strings fails because
> 
> `Successfully ran target build` is actually `[32mSuccessfully ran target \u001b[1mbuild` it just gets rendered correctly in the terminal but the raw output contains them, so doing any kind of equality or lookup fails.

jaysoo (approved):

PR #19993, opened 2023-11-02
feat(testing): add migration for moving test target defaults

FrozenPandaz:
## Current Behavior

There is no migration for the changes in #19963. New projects will be created without the settings without a `targetDefault` set.

## Expected Behavior

There is a migration to move the target defaults and options to the target defaults. New projects will still have those settings set. A lot of configuration will be removed from `project.json` files in the process. :tada: 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/migrations/update-17-1-0/move-options-to-target-defaults.ts
+++ b/packages/jest/src/migrations/update-17-1-0/move-options-to-target-defaults.ts
@@ -1,0 +1,185 @@
+import {
+  createProjectGraphAsync,
+  formatFiles,
+  getProjects,
+  ProjectConfiguration,
+  ProjectGraphProjectNode,
+  readNxJson,
+  TargetConfiguration,
+  TargetDefaults,
+  Tree,
+  updateNxJson,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import { JestExecutorOptions } from '../../executors/jest/schema';
+import {
+  forEachExecutorOptions,
+  forEachExecutorOptionsInGraph,
+} from '@nx/devkit/src/generators/executor-options-utils';
+import { readTargetDefaultsForTarget } from 'nx/src/project-graph/utils/project-configuration-utils';
+
+export default async function update(tree: Tree) {
+  const nxJson = readNxJson(tree);
+
+  // Don't override anything if there are already target defaults for jest
+  if (nxJson.targetDefaults?.['@nx/jest:jest']) {
+    return;
+  }
+
+  nxJson.targetDefaults ??= {};
+
+  /**
+   * A set of targets which does not use any other executors
+   */
+  const jestTargets = new Set<string>();
+
+  const graph = await createProjectGraphAsync();
+
+  forEachExecutorOptionsInGraph(
+    graph,
+    '@nx/jest:jest',
+    (value, proj, targetName) => {
+      jestTargets.add(targetName);
+    }
+  );
+
+  // Workspace does not use jest?
+  if (jestTargets.size === 0) {
+    return;
+  }
+  // Use the project graph so targets which are inferred are considered
+  const projects = graph.nodes;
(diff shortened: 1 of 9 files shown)


packages/vite/src/migrations/update-17-1-0/move-target-defaults.ts
```diff
@@ -0,0 +1,89 @@
...
+  updateNxJson,
+} from '@nx/devkit';
+import { forEachExecutorOptionsInGraph } from '@nx/devkit/src/generators/executor-options-utils';
+import { VitestExecutorOptions } from '../../executors/test/schema';
+import { readTargetDefaultsForTarget } from 'nx/src/project-graph/utils/project-configuration-utils';
+
+export default async function update(tree: Tree) {
+  const nxJson = readNxJson(tree);
+
+  // Don't override anything if there are already target defaults for jest
```
jaysoo, line 17: Nit: `s/jest/vitest/`

packages/jest/src/migrations/update-17-1-0/move-options-to-target-defaults.ts
```diff
@@ -0,0 +1,164 @@
...
+  jestDefaults.configurations.ci.ci ??= true;
+  jestDefaults.configurations.ci.codeCoverage ??= true;
+
+  // Cleanup old target defaults
+  for (const [targetDefaultKey, targetDefault] of Object.entries(
+    nxJson.targetDefaults
+  )) {
+    // If every target in every project does not use the target default anymore, it can be removed.
+    if (
+      Object.values(projects).every((p) =>
```
jaysoo, line 84: It's a migration so not too worried about maintenance, but I find the long, nested `everyone` hard to grok. Would it better to use nested for-loops?

```js
for (const [targetDefaultKey, targetDefault] of Object.entries(
  nxJson.targetDefaults
)) {
  let isDefaultUsed = false;
  for (const p of Object.values(projects)) {
    for (const targetName of Object.keys(p.data?.targets ?? {})) {
      const defaults = readTargetDefaultsForTarget(
        targetName,
        nxJson.targetDefaults,
        projectMap.get(p.name).targets?.[targetName]?.executor
      );
      if (defaults === targetDefault) {
        isDefaultUsed = true;
        break;
      }
    }
  }
  if (!isDefaultUsed) {
    delete nxJson.targetDefaults[targetDefaultKey];
  }
}
```

packages/jest/src/migrations/update-17-1-0/move-options-to-target-defaults.ts
```diff
@@ -0,0 +1,164 @@
...
+    '@nx/jest:jest',
+    (value, proj, targetName, configuration) => {
+      const projConfig = projectMap.get(proj);
+
+      if (!configuration) {
+        // Options
+        if (value.passWithNoTests === jestDefaults.options.passWithNoTests) {
+          delete projConfig.targets[targetName].options.passWithNoTests;
+        } else {
+          projConfig.targets[targetName].options.passWithNoTests ??= false;
```
jaysoo, line 113: This could have been a different value as set in the defaults we deleted in the above loop?

```js
for (const [targetDefaultKey, targetDefault] of Object.entries(
    nxJson.targetDefaults
  )) {
  //...
}
```

Like, they could have this previously in `nx.json`:

```
{
  "targetDefaults": {
    "test": {
      "options": {
        "passWithNoTests": true
      }
  }
}
```

After this migration, it would be set to `false`, since the project did not define it previously in `project.json`, but we deleted the defaults in `nx.json`.

packages/jest/src/migrations/update-17-1-0/move-options-to-target-defaults.ts
FrozenPandaz: Yes, I agree.

AgentEnder (approved):

PR #20004, opened 2023-11-02
fix(linter): fix update to using module node16

FrozenPandaz:
## Current Behavior

The migration for adding "module": "node16" is broken because `moduleResolution` does not match with `module`.

## Expected Behavior

`module` should match with `moduleResolution` and the migration is adjusted.

## Related Issue(s)

Fixes #

### DIFF
--- a/tools/eslint-rules/tsconfig.spec.json
+++ b/tools/eslint-rules/tsconfig.spec.json
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
-    "module": "commonjs",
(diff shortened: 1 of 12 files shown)


tools/eslint-rules/tsconfig.spec.json
```diff
@@ -2,7 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
-    "module": "commonjs",
+    "module": "node16",
```
leosvelperez, line 5: We can remove this and it will be inherited from the extended `tsconfig.json`.

jaysoo (approved):

PR #20006, opened 2023-11-02
feat(module-federation): use single file-server for static remotes

Coly010:
- feat(angular): use a single file server for static remotes
- feat(module-federation): ensure static remote file server doesn't conflict with dev remote port

## Current Behavior

We currently spin up a forked process for every static remote to run a file-server for the remote.

## Expected Behavior

We should use a single file-server for all the remotes

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/src/executors/module-federation-dev-server/module-federation-dev-server.impl.ts
+++ b/packages/react/src/executors/module-federation-dev-server/module-federation-dev-server.impl.ts
@@ -50,7 +50,6 @@ function startStaticRemotesFileServer(
   context: ExecutorContext,
   options: ModuleFederationDevServerOptions
 ) {
-  const mappedLocationOfRemotes: Record<string, string> = {};
   let shouldMoveToCommonLocation = false;
   let commonOutputDirectory: string;
   for (const app of remotes.staticRemotes) {
@@ -63,15 +62,7 @@ function startStaticRemotesFileServer(
     } else if (commonOutputDirectory !== directoryOfOutputPath) {
       shouldMoveToCommonLocation = true;
     }
-
-    mappedLocationOfRemotes[app] = `http${options.ssl ? 's' : ''}://${
-      options.host
-    }:${options.staticRemotesPort}/${app}`;
   }
-
-  process.env.NX_MF_DEV_SERVER_STATIC_REMOTES = JSON.stringify(
-    mappedLocationOfRemotes
-  );
 
   if (shouldMoveToCommonLocation) {
     commonOutputDirectory = join(workspaceRoot, 'tmp/static-remotes');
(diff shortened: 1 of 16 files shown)


ndcunningham (approved):

packages/react/src/executors/module-federation-dev-server/module-federation-dev-server.impl.ts
```diff
@@ -37,89 +41,71 @@ function getBuildOptions(buildTarget: string, context: ExecutorContext) {
...
-    options.devRemotes,
-    options.skipRemotes,
-    moduleFederationConfig,
+  const staticRemotesIter = fileServerExecutor(
     {
-      projectName: context.projectName,
-      projectGraph: context.projectGraph,
-      root: context.root,
-    }
+      skipInitialRun: true,
```
jaysoo, line 82: Instead of a new option, can we make `buildTarget` optional in `file-server`, and skip watching/running build in that case?

1. If `buildTarget` is not present, `staticFilePath` _must_ be present, or else throw error
2. If `watch` is used but `buildTarget` is not present, then throw error

This makes `file-server` able to run any path, detached from any build process.

Coly010: We can, I was more apprehensive about making a large change to file-server as it's used for static serve of apps for cypress e2es etc too and it seemed like a high risk change to make.

But if this is preferable, I can do that.

packages/react/src/executors/module-federation-dev-server/module-federation-dev-server.impl.ts
```diff
@@ -147,52 +133,139 @@ export default async function* moduleFederationDevServer(
...
     );
     staticProcess.stdout.on('data', (data) => {
-      if (isCollectingStaticRemoteOutput) {
-        outWithErr.push(data.toString());
-      } else {
-        outWithErr = null;
+      const ANSII_CODE_REGEX =
+        /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g;
+      const stdoutString = data.toString().replace(ANSII_CODE_REGEX, '');
+      if (stdoutString.includes('Successfully ran target build')) {
```
jaysoo, line 183: Maybe something for later since this logic of checking `stdout` is in existing code. Maybe Nx should be sending a message when parent spawns it with IPC. Like `process.send && process.send('nx.run.success')` or something.

The text may change, but as long as we have good e2e coverage it's okay for now.

jaysoo (approved):

PR #20018, opened 2023-11-02
fix(linter): do not expect project to have flat eslint config

meeroslav:
## Current Behavior
We expect that each project has `eslint.config.js` when in flat config.

## Expected Behavior
Project can use root `eslint.config.js`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint/src/executors/lint/lint.impl.ts
+++ b/packages/eslint/src/executors/lint/lint.impl.ts
@@ -51,12 +51,6 @@ export default async function run(
     joinPathFragments(workspaceRoot, 'eslint.config.js')
   );
 
-  if (!eslintConfigPath && hasFlatConfig) {
-    const projectRoot =
-      context.projectsConfigurations.projects[context.projectName].root;
-    eslintConfigPath = joinPathFragments(projectRoot, 'eslint.config.js');
-  }
-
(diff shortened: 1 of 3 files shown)


packages/eslint/src/executors/lint/lint.impl.ts
```diff
@@ -51,12 +51,6 @@ export default async function run(
     joinPathFragments(workspaceRoot, 'eslint.config.js')
   );
 
-  if (!eslintConfigPath && hasFlatConfig) {
-    const projectRoot =
-      context.projectsConfigurations.projects[context.projectName].root;
-    eslintConfigPath = joinPathFragments(projectRoot, 'eslint.config.js');
```
JamesHenry, line 57: Just checking - will `FlatESLint` resolve the closest `eslint.config.js` file automatically if we don't pass it explicitly (like the original ESLint class did)? Do we have an e2e case that covers this removal/flat config in general?

meeroslav: It's a valid question. I tested it on an isolated repo, but I'll add an E2E test to have this officially verified.

JamesHenry (approved):

PR #20024, opened 2023-11-03
feat(module-federation): Add react support for  dynamic federation

ndcunningham:
## Current Behavior

Currently, we only support static federation or promise based modules.

## Expected Behavior

With this change, you will be able to load remotes via a manifest file (`module-federation.manifest.json`).

The host generator has been updated to behave similarly to `@nx/angular` that has the `--dynamic` flag.

```shell
nx generate @nx/react:host acme --remotes=nx --dynamic=true  --styles=css
```
This configures the host to dynamically load remotes via the manifest file.

### DIFF
--- a/packages/react/src/generators/remote/schema.json
+++ b/packages/react/src/generators/remote/schema.json
@@ -28,6 +28,12 @@
       "type": "string",
       "enum": ["as-provided", "derived"]
     },
+    "dynamic": {
+      "type": "boolean",
+      "description": "Should the host application use dynamic federation?",
+      "default": false,
+      "x-priority": "internal"
+    },
     "style": {
       "description": "The file extension to be used for style files.",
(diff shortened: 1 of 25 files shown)


packages/react/src/generators/remote/schema.json
```diff
@@ -28,6 +28,12 @@
       "type": "string",
       "enum": ["as-provided", "derived"]
     },
+    "dynamic": {
+      "type": "boolean",
+      "description": "Should the host application use dynamic federation?",
+      "default": false,
+      "x-priority": "internal"
```
ndcunningham, line 35: This _could_ be left open if the user generates a new remote to an existing host.

mandarini (approved):

Coly010 (approved):

PR #20025, opened 2023-11-03
cleanup(angular): fix vercel deployment

leosvelperez:
- Removes the `pnpm` overrides, which are causing an issue with the Vercel deployment.
- Convert `import('@nguniversal/builders')` to `require('@nguniversal/builders')` to avoid type-checking issues (previously solved by the `pnpm` overrides). It's safe to do so because it's only executed at runtime, and the API won't change anymore because there won't be new versions of the package.

## Related Issue(s)

Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -369,10 +369,5 @@
       "check-codeowners",
       "documentation"
     ]
-  },
-  "pnpm": {
-    "overrides": {
-      "@angular-devkit/core": "$@angular-devkit/core"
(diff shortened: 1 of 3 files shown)


leosvelperez:
Current e2e failing tests will be fixed by https://github.com/nrwl/nx/pull/20028.

Coly010 (approved):

PR #20036, opened 2023-11-03
feat(angular): block router initial navigation only for ssr when opting-out of hydration

leosvelperez:
- Stop generating all apps with `initialNavigation: 'enabledBlocking'` (NgModule) or `withEnabledBlockingInitialNavigation` (standalone)
- Set the above only when setting up ssr and if using Angular v17, when opting out of hydration

## Related Issue(s)

Fixes #


Coly010:
LGTM

PR #20043, opened 2023-11-03
docs(core): nx cloud features section

isaacplmann:
Moves remote caching, DTE and affected from Nx into a new Nx Cloud features section
Moves private cloud recipes into a dedicated recipes section
Moves the private cloud intro into the Nx Cloud features section

Screenshot 2023-11-03 at 4 00 20 PM


docs/shared/using-nx/affected.md
```diff
@@ -1,27 +1,122 @@
-# How Affected Works
+# Run Only Tasks Affected by a PR
 
-{% callout type="note" title="First thing!" %}
-Before reading this guide, [check out the mental model guide](/concepts/mental-model). It will help you understand how computation caching fits into the rest of Nx.
-{% /callout %}
+s
```
juristr, line 3: Typo

docs/shared/using-nx/affected.md
```diff
@@ -1,27 +1,122 @@
...
+When you run `nx test app1`, you are telling Nx to run the `app1:test` task plus all the tasks it depends on.
```
juristr, line 4: Do we really need these first set of paragraphs? I would almost just start with "As you workspace grows,..." and then mentioning maybe "retesting, rebuilding, re-linting..." basically emphasize it is not just for testing.

Then right after "How does this work?", I'd place in a terminal window showing `nx affected -t <task>` or an alternate syntax. Basically to have the actual command right there. And then proceed with the explanations

docs/shared/using-nx/affected.md
```diff
@@ -1,27 +1,122 @@
-# How Affected Works
+# Run Only Tasks Affected by a PR
```
juristr, line 1: Overall for this one. Can we also make the headings more descriptive? 

Right now they are not really useful
image

Maybe something like

- How Affected Works
- Visualizing Affected Projects (with the Graph)
- Determining the base SHA to compare with (or something similar)
- Ignoring Files from Affected Commands
- Not using Git

juristr: Some more generic comments because I cannot comment directly on the line since it didn't change in this PR 🤷‍♂️.

- In the current "CI" section I'd also link to the various ways we have to find the last successful SHA etc.
- In the "Not Using Git" I'd mention that Nx uses Git to determine the affected files, otherwise it might be confusing. Or maybe we should mention that in the "How it works" section. I don't think we did that.

juristr (changes requested):
some small changes

docs/shared/using-nx/affected.md
isaacplmann: Each CI provider has a different way of finding the last successful SHA, so I put in a generic statement about the importance of comparing to the last successful commit.

juristr (approved):

docs/shared/using-nx/affected.md
AgentEnder: It would probably be worth linking the [nx-orb](https://github.com/nrwl/nx-orb) and [set-shas](https://github.com/nrwl/nx-set-shas) right?

AgentEnder (approved):

docs/shared/using-nx/affected.md
isaacplmann: ok, I can mention them.  They're specific to Circle CI and Github Actions though, so people who aren't using those CI providers won't know what to do

PR #20045, opened 2023-11-03
feat(testing): add jest create-nodes plugin

FrozenPandaz:
## Current Behavior

Jest targets are inferred based off of `jest.config.ts` files.

`jest.config.ts` files will add targets to projects which run the `jest` cli. 

Inputs and outputs are automatically determined based on the config in `jest.config.js` so there is no need to provide additional config in most cases.

## Expected Behavior

Jest targets are added to `project.json` which run the `@nx/jest:jest` executor.

## Related Issue(s)

Fixes #


AgentEnder (approved):

FrozenPandaz:
CI Failures are incorrectly being reported. The test did indeed pass.

PR #20046, opened 2023-11-03
fix(core): attach github release to the current branch

FrozenPandaz:
## Current Behavior

When creating a release, it always happens off of the current commit of `master`.

## Expected Behavior

Releases can be created off of different branches such as patch branches.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/release/changelog.ts
+++ b/packages/nx/src/command-line/release/changelog.ts
@@ -26,6 +26,7 @@ import {
   getGitDiff,
   getLatestGitTagForPattern,
   parseCommits,
+  getCommitHash,
 } from './utils/git';
 import {
   GithubRelease,
@@ -38,6 +39,7 @@ import {
 import { launchEditor } from './utils/launch-editor';
 import { parseChangelogMarkdown } from './utils/markdown';
 import { printChanges, printDiff } from './utils/print-changes';
+import { handleErrors } from '../../utils/params';
 
 class ReleaseVersion {
   rawVersion: string;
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/release/changelog.ts
```diff
@@ -420,6 +421,7 @@ async function generateChangelogForWorkspace(
         {
           version: releaseVersion.gitTag,
           prerelease: releaseVersion.isPrerelease,
+          target_commitish: await getCurrentCommit(),
```
JamesHenry, line 424: I don't think this should always be the current commit, it should be based on the resolved value for `--to`/`args.to` right? Because a custom end range could be being passed in. I also imagine  `target_commitish` would support a value of `HEAD` (the default of `args.to`) but we'd need to double check

JamesHenry (changes requested):

packages/nx/src/command-line/release/changelog.ts
FrozenPandaz: That's fair, I moved the value to the `createRelease` only. Existing releases will not override the existing `target_commitish`

JamesHenry: Ah but even though it's likely a `--to` earlier than the current HEAD suggests an update to an existing release, it doesn't necessarily mean that right? You might want to cut a new release but only go up to a certain commit. So I think you were on the right track with the initial fix, it's just we want to pass through the value of `args.to` to use for `target_commitish` in all cases

FrozenPandaz: Hm I see, so I passed it through now. I still didn't send it in for updating the existing release though. I'm not sure if it makes sense to update the commit on the existing release because... the release has already been done at least for the Nx repo.

JamesHenry: Pinged you on slack

JamesHenry (approved):

PR #19689, opened 2023-10-18
feat(angular): support angular v17

leosvelperez:
## Main changes

- Add support for Angular v17.0.0.
- Switch `--standalone` default to `true` in generators
- Generate v17 applications with the new `application` builder when choosing the `esbuild` bundler
	- Add CNW prompt to choose the bundler
- Support new `@angular/ssr` and ssr builders under `@angular-devkit/build-angular` for v17
	- Add CNW prompt to choose whether to generate the Angular app with SSR
	- Add prompt for SSR to `application` generator
	- Default `hydration` to `true` for v17
- Drop support for Angular v14

## Current Behavior

Angular v17 is not supported.

## Expected Behavior

Angular v17 is supported.

## Related Issue(s)

Fixes #

PR #20048, opened 2023-11-04
fix(angular): use escape character for @ symbol in sample app

yharaskrik:
## Current Behavior

Will not work with v17 control flow

## Related Issue(s)

Fixes #


LayZeeDK:
Sample error in generated Angular 17 app

```shell
Error: apps/my-app/src/app/nx-welcome.component.ts:786:6 - error NG5002: Incomplete block "nx". If you meant to write the @ character, you should use the "&#64;" HTML entity instead.

786 nx g @nx/angular:lib ui
```

AgentEnder (approved):

leosvelperez:
Hey @yharaskrik!

Thanks for the contribution! The support for Angular v17 is being tracked here https://github.com/nrwl/nx/pull/19689 and this change is already included there. I'm going to close this one in favor of that PR.

PR #20086, opened 2023-11-07
feat(vite): nodes for build, serve, test, preview targets

mandarini:
A plugin (`@nx/vite/plugin`) which will add Vite targets/projects to the Nx project graph. This allows you to run Vite for projects that have a `vite.config.ts` file without creating a `project.json` file.

## To-do's

- [x] Read outputs from `vite.config.ts`
- [x] Figure out why it does not like `.js`
- [x] Figure out `cannot use import outside module` on pure `.ts` vite apps
- [x] Vitest: `passWithNoTests` should be moved to `targetDefaults` like `jest`
- [x] Vitest: `reportsDirectory` shouldn't be passed in the executor options but rather in the config
- [x] node for `static-serve`

Once [this PR](https://github.com/nrwl/nx/pull/20355) is merged, we can add the cypress web server config.

## How it works

On `@nx/vite:init` (which is called through the `@nx/vite:configuration` generator, and the `@nx/react:app|lib` and `@nx/web:app|lib` generators) this is added in the `nx.json`:
```
...
plugins: [
  {
    plugin: '@nx/vite/plugin',
    options: {
      buildTargetName: 'build',
      previewTargetName: 'preview',
      testTargetName: 'test',
      serveTargetName: 'serve',
    },
  },
  ...
]
```
so that you can run `nx build vite`, `nx test vite` etc, without manually defining these targets in your projects' `project.json`.

The plugin generates these targets with bare minimum (no extra) options configuration. Any options you need for your vite app, you can add in your project's `vite.config.ts`.

## Docs changes

Simple docs page: https://nx-dev-git-fork-mandarini-feat-vite-nodes-nrwl.vercel.app/nx-api/vite/documents/vite-nodes-plugin (to be discussed?) 

### FIXED: Why it does not like `js`
```
 >  NX   require() of ES Module /Users/fileas/Projects/nrwl/test-nx-workspaces/vitenodes/rlibjs/vite.config.js from /Users/fileas/Projects/nrwl/test-nx-workspaces/vitenodes/node_modules/@nx/vite/src/plugins/plugin.js not supported.

   Instead change the require of vite.config.js in /Users/fileas/Projects/nrwl/test-nx-workspaces/vitenodes/node_modules/@nx/vite/src/plugins/plugin.js to a dynamic import() which is available in all CommonJS modules.
```

#### Solutions

1. This can be solved once the [createNodes function handles `async`](https://github.com/nrwl/nx/pull/20195).
2. This can be solved now using `readFileSync` and then AST parsing.

### FIXED:  `cannot use import outside module` on pure `.ts` vite apps

It does not like the loaded `vite.config.ts`.

#### Solutions 

1. This will be solved once we use `loadConfigFromFile` function, [when `async` is ready](https://github.com/nrwl/nx/pull/20195)
2. This can be solved now using `readFileSync` and then AST parsing.


Coly010 (approved):

packages/vite/src/plugins/plugin.ts
```diff
@@ -0,0 +1,294 @@
...
+];
+
+function buildViteTargets(
+  configFilePath: string,
+  projectRoot: string,
+  options: VitePluginOptions,
+  context: CreateNodesContext
+) {
+  // Is this needed for some reason?
+  getViteConfig(configFilePath, context);
```
jaysoo, line 65: I think you will need to find the output directory. Right now it is hardcoded as `joinPathFragments('dist', projectRoot)` but the user could change it.

jaysoo (approved):

jaysoo (changes requested):
We should handle reading output from config. This is blocked on making `CreateNodes` handle async.

packages/vite/src/migrations/update-17-2-0/add-vite-plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+export default async function update(tree: Tree) {
+  const proj = Object.fromEntries(getProjects(tree).entries());
+
+  const rootMappings = createProjectRootMappingsFromProjectConfigurations(proj);
+
+  // add function here to remove
+  // passWithNoTests: true,
+  // reportsDirectory: '../../coverage/apps/my-app',
+
+  await replaceProjectConfigurationsWithPlugin(
```
FrozenPandaz, line 16: This won't do much so we should remove it for now.

packages/vite/src/migrations/update-17-2-0/add-vite-plugin.ts
```diff
@@ -0,0 +1,31 @@
...
+import { replaceProjectConfigurationsWithPlugin } from '@nx/devkit/src/utils/replace-project-configuration-with-plugin';
+
+export default async function update(tree: Tree) {
+  const proj = Object.fromEntries(getProjects(tree).entries());
+
+  const rootMappings = createProjectRootMappingsFromProjectConfigurations(proj);
+
+  // add function here to remove
+  // passWithNoTests: true,
+  // reportsDirectory: '../../coverage/apps/my-app',
```
FrozenPandaz, line 14: Do we need a migration for this?

FrozenPandaz (changes requested):

packages/vite/src/generators/init/lib/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  nxJson.targetDefaults ??= {};
+  nxJson.targetDefaults['@nx/vite:test'] ??= {};
+  nxJson.targetDefaults['@nx/vite:test'].cache ??= true;
+  nxJson.targetDefaults['@nx/vite:test'].inputs ??= [
+    'default',
+    productionFileSet ? '^production' : '^default',
+  ];
+  nxJson.targetDefaults['@nx/vite:test'].options ??= {
```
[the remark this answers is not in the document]
mandarini: Yes, I agree, let's not add the target defaults.

I am not sure I understand the meaning of (3). We should not configure jest/vitest at all?

isaacplmann (approved):

jaysoo (approved):

FrozenPandaz (approved):

PR #20091, opened 2023-11-07
fix(react): import path info on federated module

raketeFlo:
## Current Behavior
When generating a federated module the logger shows an example import in the terminal. However, the import path is wrong. 
Screenshot 2023-11-07 at 14 16 51

## Expected Behavior
The import path for the newly generated module should be displayed correctly.
`import { MyComponent } from 'dashboard/widgetA';`
`import('dashboard/widgetA').then((m) => m.widgetA);`


ndcunningham (approved):
LGTM thanks!

PR #20130, opened 2023-11-08
fix(angular): use esm2022 for secondary entrypoints in ng-packagr-lite executor

skrtheboss:
## Related Issue(s)

Fixes #


atsjo:
still can't build my angular project with incremental esbuild and a library with secondary entrypoints... I guess this is the fix for that, as this is supposed to be fixed in ng-packagr now...

packages/angular/src/executors/ng-packagr-lite/ng-packagr-adjustments/ng-package/entry-point/write-package.transform.ts
```diff
@@ -135,7 +135,7 @@ export const nxWritePackageTransform = (options: NgPackagrOptions) =>
       await writeFile(
         path.join(ngEntryPoint.destinationPath, 'package.json'),
         JSON.stringify(
-          { module: relativeUnixFromDestPath(destinationFiles.fesm2022) },
+          { module: relativeUnixFromDestPath(destinationFiles.esm2022) }, // esm2022 is used since ng-packager-lite does not provide fesm2022
```
leosvelperez, line 138: We also need to account for the differences between the supported Angular versions. We don't need the comment. We already have a comment at the top of the files to mention the overall changes.
```suggestion
          {
            module: relativeUnixFromDestPath(
              angularVersion.major < 16
                ? (destinationFiles as any).esm2020
                : destinationFiles.esm2022
            ),
          },
```

leosvelperez (changes requested):
Thanks for the contribution!

This is a good catch. I left a comment with a suggested change. We need to account for the differences between the supported Angular versions.

leosvelperez (approved):

PR #20137, opened 2023-11-08
chore(node): fix e2e tests and add implicit dependency on @nx/webpack

jaysoo:
Fixed failing test since the optimized output has changed, and was not updated in previous PR (due to missing implicit dependency).

## Related Issue(s)

Fixes #


e2e/node/src/node-webpack.test.ts
```diff
@@ -35,7 +35,7 @@ describe('Node Applications + webpack', () => {
       function foo(x: string) {
         return "foo " + x;
       };
-      console.log(foo("bar")); 
+      console.log(foo("bar")); a
```
FrozenPandaz, line 38: Stray character

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #20145, opened 2023-11-09
feat(nuxt): nodes for build, serve, test targets

mandarini:
A plugin (`@nx/nuxt/plugin`) which will add Nuxt targets/projects to the Nx project graph. This allows you to run Nuxt for projects that have a `nuxt.config.ts` file without creating a `project.json` file.

## To-do's

- [x] Read outputs from config ([when `async` is ready](https://github.com/nrwl/nx/pull/20195))
- [x] test node
- [x] Look into correct dir resolution - what outputs are indeed needed
- [x] Decide on what to do with the docs

Once [this PR](https://github.com/nrwl/nx/pull/20355) is merged, we can add the cypress web server config.

## How it works

On `@nx/nuxt:init` (which is called through the `@nx/nuxt:application` generator this is added in the `nx.json`:
```
...
plugins: [
  {
    plugin: '@nx/nuxt/plugin',
    options: {
      buildTargetName: 'build',
      serveTargetName: 'serve',
      testTargetName: 'test'
    },
  },
  ...
]
```
so that you can run `nx build my-app`, `nx test my-app` etc, without manually defining these targets in your projects' `project.json`.

The plugin generates these targets with bare minimum (no extra) options configuration. Any options you need for your nuxt app, you can add in your project's `nuxt.config.ts`.


mandarini:
Todo:
- [ ] Read outputs from config

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -125,18 +127,23 @@ function buildTarget(
...
     '@nx/nuxt:build'
   );
 
-  const targetConfig: TargetConfiguration<Partial<NuxtBuildExecutorOptions>> = {
-    executor: '@nx/nuxt:build',
-    defaultConfiguration: 'production',
+  const targetConfig: TargetConfiguration = {
+    executor: 'nx:run-commands',
+    options: {
+      command: 'npx nuxi build',
```
meeroslav, line 142: You should use `detectPackageManager` + `getPackageManagerCommand` instead of hardcoded `npx`. This would fail with `PnP`

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -155,10 +162,13 @@ function buildTarget(
...
 
-function serveTarget() {
-  const targetConfig: TargetConfiguration<Partial<NuxtServeExecutorOptions>> = {
-    executor: '@nx/nuxt:serve',
-    defaultConfiguration: 'development',
+function serveTarget(projectRoot: string) {
+  const targetConfig: TargetConfiguration = {
+    executor: 'nx:run-commands',
+    options: {
+      command: 'npx nuxi dev',
```
meeroslav, line 169: You should use `detectPackageManager` + `getPackageManagerCommand` instead of hardcoded `npx`. This would fail with `PnP`

docs/shared/packages/nuxt/nuxt-plugin.md
```diff
@@ -50,3 +50,87 @@ pnpm install -D @nx/nuxt
 ```shell
 nx g @nx/nuxt:app my-app
 ```
+
+## Add Nx to existing Nuxt applications
+
+Using the `@nx/nuxt/plugin`, you can adds Nuxt targets to Nx. This allows you to run Nuxt for projects that have a `nuxt.config.ts` file via Nx with caching and everything setup.
```
jaysoo, line 56: I think we need a doc to explain the plugin system and point to it here. Something @isaacplmann can help with?

The mental model is adding a plugin allows the commands to run through Nx, with caching, distribution, etc.

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -0,0 +1,284 @@
...
+  projectRoot: string
+) {
+  const targetDefaults = readTargetDefaultsForTarget(
+    options.testTargetName,
+    context.nxJsonConfiguration.targetDefaults,
+    '@nx/vite:test'
+  );
+
+  const targetConfig: TargetConfiguration = {
+    command: `${
```
jaysoo, line 189: Can just run `vitest` directly. Nx already wraps it.

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -0,0 +1,284 @@
...
+    [inputName: string]: any[];
+  },
+  outputs: string[],
+  options: NuxtPluginOptions,
+  projectRoot: string
+) {
+  const targetDefaults = readTargetDefaultsForTarget(
+    options.testTargetName,
+    context.nxJsonConfiguration.targetDefaults,
+    '@nx/vite:test'
```
jaysoo, line 185: We don't need read from `@nx/vite:test` for defaults I think, since we're not using the executor here.

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -0,0 +1,284 @@
...
+) {
+  const targetDefaults = readTargetDefaultsForTarget(
+    'build',
+    context.nxJsonConfiguration.targetDefaults,
+    '@nx/nuxt:build'
+  );
+
+  const targetConfig: TargetConfiguration = {
+    command: `${
+      getPackageManagerCommand(detectPackageManager()).exec
```
jaysoo, line 138: Can just run `nuxi` directly.

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -0,0 +1,284 @@
...
+        ? ['production', '^production']
+        : ['default', '^default'];
+  }
+
+  return targetConfig;
+}
+
+function serveTarget() {
+  const targetConfig: TargetConfiguration = {
+    executor: '@nx/nuxt:serve',
```
jaysoo, line 166: Can we run `nuxi` directly here?

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  if (targetDefaults?.cache === undefined) {
+    targetConfig.cache = true;
+  }
+
+  if (targetDefaults?.inputs === undefined) {
+    targetConfig.inputs =
+      'production' in namedInputs
+        ? ['production', '^production']
```
jaysoo, line 157: We should also add `nuxi` as an external dep here since we're running it for build.

See: https://github.com/nrwl/nx/blob/master/packages/cypress/src/plugins/plugin.ts#L338

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -0,0 +1,284 @@
...
+    if (
+      !siblingFiles.includes('package.json') &&
+      !siblingFiles.includes('project.json')
+    ) {
+      return {};
+    }
+
+    options = normalizeOptions(options);
+
+    const hash = calculateHashForCreateNodes(projectRoot, options, context);
```
jaysoo, line 68: Also add the lockfile into the hash.

See: https://github.com/nrwl/nx/blob/master/packages/cypress/src/plugins/plugin.ts#L75

jaysoo (approved):

isaacplmann (approved):

FrozenPandaz (approved):

PR #20146, opened 2023-11-09
feat(angular): update component generator to use a single string `styles` or `styleUrl` property

leosvelperez:
## Current Behavior

Components are always generated with `styles` or `styleUrls` properties with an array value.

## Expected Behavior

For Angular v17, components should be generated with `styles` or `styleUrl` properties with a single string value.

## Related Issue(s)

Fixes #


leosvelperez:
The failing E2E tests are caused by an issue with `jest-preset-angular` being addressed by https://github.com/thymikee/jest-preset-angular/pull/2186. This PR is blocked until a new release of `jest-preset-angular` with the fix comes out.

Coly010 (approved):

PR #20155, opened 2023-11-09
fix(release): catch 403 when publishing an existing version

fahslaj:
## Current Behavior

Publishing fails when the existing version of the package already exists on npm.

## Expected Behavior

Publishing continues when the existing version of the package already exists on npm.


packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -117,12 +124,7 @@ export default async function runExecutor(
       }
 
       console.error('npm publish error:');
-      if (stdoutData.error.summary) {
-        console.error(stdoutData.error.summary);
-      }
-      if (stdoutData.error.detail) {
-        console.error(stdoutData.error.detail);
-      }
+      console.error(JSON.stringify(stdoutData, null, 2));
```
FrozenPandaz, line 127: Maybe keep the current one but then print this when verbose is true.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

JamesHenry (approved):

PR #20156, opened 2023-11-09
fix(vite): vitest should use v8 as default

mandarini:
c8 is deprecated


jaysoo (approved):

mandarini:
Ah, those flaky tests <3

Issue #19989, opened 2023-11-02
NX 17 does not update package.json with internal buildable dependency versions during build


mta-trackunit:
### Current Behavior

NX 16 did update package.json on build using updateBuildableProjectDepsInPackageJson, so dist folder of the lib would have updated dependencies.

NX 17 does not update internal dependency versions on build so using the new lint the versions ends up with x: "*" 
updateBuildableProjectDepsInPackageJson has been removed but I cannot find any details on how it is supposed to work, how do i end up with a build with the right versions of dependencies and not "*".

I also tried checkout nx repo and build nx.js - it also does not update dependencies however its using "file://path_to_lib"

### Expected Behavior

It should work as in NX 16 so a build ends up with the package.json that can be published to npm.

### Steps to Reproduce

1. If I follow this guide --> https://nx.dev/getting-started/tutorials/integrated-repo-tutorial 
   _(also setup the lint to --fix the versions)_ 
    once you are done its package.json should look like this: 
    ```
    {
      "name": "@myorg/is-odd",
      "version": "1.2.3",
      "dependencies": {
        "tslib": "^2.3.0",
        "@myorg/is-even": "*"
      },
      "type": "commonjs",
      "main": "./src/index.js",
      "typings": "./src/index.d.ts"
    }
    ```
2. RUN
```
nx run is-odd:build          
```
3. Check the output in dist/libs/is-odd/package.json
```
{
  "name": "@myorg/is-odd",
  "version": "0.0.1",
  "dependencies": {
    "tslib": "^2.3.0",
    "@myorg/is-even": "*"
  },
  "type": "commonjs",
  "main": "./src/index.js",
  "typings": "./src/index.d.ts"
}

```

### Nx Report

```shell
cat dist/libs/is-odd/package.json

{
  "name": "@myorg/is-odd",
  "version": "0.0.1",
  "dependencies": {
    "tslib": "^2.3.0",
    "@myorg/is-even": "*"
  },
  "type": "commonjs",
  "main": "./src/index.js",
  "typings": "./src/index.d.ts"
}
```

### Package Manager Version

Node v18.17.0 yarn 1.22.19

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

jaysoo:
To get some clarifications:

1. You set the version in `package.json` of each lib (committed to git).
[...]

PR #20157, opened 2023-11-09
feat(linter): add option for @nx/dependency-checks to update workspace dependencies using local file paths

jaysoo:
This PR adds `useLocalPathsForWorkspaceDependencies` option to `@nx/dependency-checks` lint rule so that when using `--fix`, it can use a local path (e.g. `file:../foobar`) rather than the current `*` value.

Also, when `useLocalPathsForWorkspaceDependencies` is `false` or `undefined`, the value is read from the source/committed `package.json` file of the dependency rather than using `*`. This way users can use the committed versions if that is how they manage releases, or set the version to `*` if desired.

Another benefit is that committed versions are now also checked to ensure version ranges match their dependencies' versions.

## Related Issue(s)

Related #19989


meeroslav (approved):

mandarini (approved):

BenSawary:
@jaysoo your build is broken. Can you please check this out, to get your feature with the next Nx release :)

Issue #19841, opened 2023-10-25
Expected a JavaScript module


Cpt-Falcon:
### Current Behavior

After upgrading from nx 16 to nx 17 and running the two migration commands, my federated modules, which were working before, are now seeing this error when going to cloudfront

Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.

### Expected Behavior

Should work the same as NX 16

### Steps to Reproduce

1. Setup a federated module

Tomorrow I will try to create an example repo

### Nx Report

```shell
Node   : 18.15.0
   OS     : win32-x64
   pnpm   : 8.9.0

   nx                 : 17.0.1
   @nx/js             : 17.0.1
   @nx/jest           : 17.0.1
   @nx/linter         : 17.0.1
   @nx/eslint         : 17.0.1
   @nx/workspace      : 17.0.1
   @nx/cypress        : 17.0.1
   @nx/devkit         : 17.0.1
   @nx/eslint-plugin  : 17.0.1
   @nx/react          : 17.0.1
   @nrwl/tao          : 17.0.1
   @nx/web            : 17.0.1
   @nx/webpack        : 17.0.1
   typescript         : 5.2.2
```

### Operating System

- [ ] macOS
- [X] Linux
- [X] Windows
- [ ] Other (Please specify)

Cpt-Falcon:
I actually fixed the issue; for anyone else seeing this, all you have to do is specify remoteEntry.js in your webpack files remotes section explicitly. Previously I didn't need to specify the remoteEntry.js, but now you do apparently.

rom1vtle:
Hello, same probleme

So we need to setup in webpack.prod.config.js ?
and add /remoteEntry.mjs ?*

```
 remotes: [
    ['myremoteapp', '//myremoteapp.domain.com/**remoteEntry.mjs**'],
```

Cpt-Falcon:
@rom1vtle That's correct, I had to do this, but i'm not sure what the exact equivalent would be in angular:
```

const { composePlugins, withNx } = require('@nx/webpack');
const { withReact } = require('@nx/react');
const { withModuleFederation } = require('@nx/react/module-federation');

const baseConfig = require('./module-federation.config');

const prodConfig = {
    ...baseConfig,
    remotes: [
        ['module1', '//module1.cloudfront.net/remoteEntry.js'],
[...]

PR #20177, opened 2023-11-10
feat(module-federation): add comment to generated module federation config explaining usage of external remotes

Coly010:
## Current Behavior

We do not showcase any method for using remotes that are external to the repo

## Expected Behavior

Have a comment in the generated module federation config that indicates how to use external remotes

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/setup-mf/files/ts-webpack/module-federation.config.ts__tmpl__
+++ b/packages/angular/src/generators/setup-mf/files/ts-webpack/module-federation.config.ts__tmpl__
@@ -2,6 +2,18 @@ import { ModuleFederationConfig } from '@nx/webpack';
 
 const config: ModuleFederationConfig = {
   name: '<%= name %>',<% if(type === 'host') { %>
+  /**
+   * To use a remote that does not exist in your current Nx Workspace
+   * You can use the tuple-syntax to define your remote
+   *
+   * remotes: [['my-external-remote', 'https://nx-angular-remote.netlify.app']]
+   *
+   * You _may_ need to add a `remotes.d.ts` file to your `src/` folder declaring the external remote for tsc, with the
+   * following content:
+   *
+   * declare module 'my-external-remote';
+   *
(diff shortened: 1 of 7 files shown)


packages/angular/src/generators/setup-mf/files/ts-webpack/module-federation.config.ts__tmpl__
```diff
@@ -2,6 +2,18 @@ import { ModuleFederationConfig } from '@nx/webpack';
 
 const config: ModuleFederationConfig = {
   name: '<%= name %>',<% if(type === 'host') { %>
+  /**
+   * To use a remote that does not exist in your current Nx Workspace
+   * You can use the tuple-syntax to define your remote
+   *
+   * remotes: [['my-external-remote', 'https://nx-angular-remote.netlify.app']]
```
minijus, line 9: Should it point to `remoteEntry.mjs` directly to better address the confusion/misunderstanding as in this issue https://github.com/nrwl/nx/issues/19841
```suggestion
   * remotes: [['my-external-remote', 'https://nx-angular-remote.netlify.app/remoteEntry.mjs']]
```

Coly010: Our utils should add the appropriate default file. I think it's fine to stick to that

minijus: Strange, that is not what we observed after migrating to nx v17. Had to add remoteEntry.mjs suffix manually to all remotes on custom host.

Coly010: That’s interesting. I tried this out locally before I made the PR and it worked 🤔 id be interested in seeing a reproduction

jaysoo (approved):

PR #20188, opened 2023-11-10
feat(testing): distribute cypress tests for ci

FrozenPandaz:
## Current Behavior

Cypress projects are run as a suite in CI.

## Expected Behavior

Cypress tasks are configured for each each spec file and then linked by a parent task. The task to run is the parent task, called `e2e-ci` by default.
graph (12)

What this allows Nx to do is run these tasks in parallel distributed across multiple machines via DTE.

Ultimately, this means that cypress suites are run across multiple machines in less time AND each spec file is cached independently.

This PR also contains other changes too

Cypress changes

1. A cache is added to the node construction so that it doesn't need to run unless something in the project changes. This isn't bullet proof but should help a lot for the performance issues. This might need to be tweaked further after testing.

Devkit

1. There is a new internal function in devkit which allows other plugins to easily hash a project's files so that they too can cache their target construction.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/nx/executors/noop.json
+++ b/docs/generated/packages/nx/executors/noop.json
@@ -9,7 +9,7 @@
     "cli": "nx",
     "outputCapture": "pipe",
     "properties": {},
-    "additionalProperties": false,
+    "additionalProperties": true,
     "presets": []
   },
   "description": "An executor that does nothing",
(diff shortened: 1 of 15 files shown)


jaysoo (approved):

jaysoo (approved):

AgentEnder (approved):

jogelin:
@FrozenPandaz do you plan to do the same for playwright as it was done here https://github.com/nrwl/tic-tac-toe-playwright/tree/e2e-sharding ?

PR #20193, opened 2023-11-10
feat(nextjs): Add support for create nodes for nextjs

ndcunningham:
## Current Behavior

There is no plugin to run Nextjs projects (a project containing `next.config.js`)
## Expected Behavior

Using this plugin `@nx/next/plugin` inside a fresh will create targets (`build`, `start` & `dev`) to `nx.json`. 
The user should now be able to run these targets without configuring a `project.json`.

### How to Use
Add this to your `nx.json` plugins
```json
  "plugins": [
    {
      "plugin": "@nx/next/plugin",
      "options": {
        "buildTargetName": "build", // <--- These names you can changed to your liking
        "startTargetName": "start",
        "devTargetName": "dev",
      }
    }
  ]
```

For new projects when you create a Nextjs project this will already be set up for you after passing `NX_PCV3=true`.
```shell
NX_PCV3=true nx generate @nx/next:app ${appName} --project-name-and-root-format=as-provided --no-interactive
```

## NB
`next export` has been deprecated in favor of:
```js
const nextConfig = {
//....... other configs

  output: 'export'
};
```

### DIFF
--- a/e2e/next-core/src/next-webpack.test.ts
+++ b/e2e/next-core/src/next-webpack.test.ts
@@ -94,6 +94,6 @@ describe('Next.js Webpack', () => {
     expect(() => {
       runCLI(`build ${appName}`);
     }).not.toThrow();
-    checkFilesExist(`apps/${appName}/.next/build-manifest.json`);
+    checkFilesExist(`dist/apps/${appName}/.next/build-manifest.json`);
   }, 300_000);
 });
--- a/packages/next/plugins/with-nx.ts
+++ b/packages/next/plugins/with-nx.ts
@@ -5,12 +5,11 @@
 import type { NextConfig } from 'next';
 import type { NextConfigFn } from '../src/utils/config';
 import type { NextBuildBuilderOptions } from '../src/utils/types';
-import type { DependentBuildableProjectNode } from '@nx/js/src/utils/buildable-libs-utils';
-import type {
-  ExecutorContext,
-  ProjectGraph,
-  ProjectGraphProjectNode,
-  Target,
+import {
+  type ExecutorContext,
+  type ProjectGraph,
+  type ProjectGraphProjectNode,
+  type Target,
 } from '@nx/devkit';
 
 const baseNXEnvironmentVariables = [
@@ -48,6 +47,7 @@ const baseNXEnvironmentVariables = [
   'NX_MAPPINGS',
   'NX_FILE_TO_RUN',
   'NX_NEXT_PUBLIC_DIR',
+  'NX_CYPRESS_COMPONENT_TEST',
 ];
 
 export interface WithNxOptions extends NextConfig {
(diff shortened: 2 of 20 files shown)


e2e/next-core/src/next-webpack.test.ts
```diff
@@ -94,6 +94,6 @@ describe('Next.js Webpack', () => {
     expect(() => {
       runCLI(`build ${appName}`);
     }).not.toThrow();
-    checkFilesExist(`apps/${appName}/.next/build-manifest.json`);
+    checkFilesExist(`dist/apps/${appName}/.next/build-manifest.json`);
```
ndcunningham, line 97: `build` only creates production artifacts so this test case was wrong before.

packages/next/src/plugins/plugin.ts
```diff
@@ -0,0 +1,229 @@
...
+    if (
+      !siblingFiles.includes('package.json') &&
+      !siblingFiles.includes('project.json')
+    ) {
+      return {};
+    }
+
+    options = normalizeOptions(options);
+
+    const hash = calculateHashForCreateNodes(projectRoot, options, context);
```
jaysoo, line 70: We also need to take lockfile into account. See https://github.com/nrwl/nx/blob/master/packages/cypress/src/plugins/plugin.ts#L74-L76

packages/next/src/plugins/plugin.ts
```diff
@@ -0,0 +1,229 @@
...
+  options.buildTargetName ??= 'build';
+  options.devTargetName ??= 'dev';
+  options.startTargetName ??= 'start';
+  return options;
+}
+
+function getInputs(
+  namedInputs: NxJsonConfiguration['namedInputs']
+): TargetConfiguration['inputs'] {
+  return [
```
jaysoo, line 221: We realized that this will prevent `externalDependencies` from being added to inputs for build, since Nx generates with a default `build` entry in `nx.json`.

See Katerina's fix here https://github.com/nrwl/nx/pull/20528/files#diff-e21b84bc7a86a657a55ca6be0f7f8dbe1c2a3ff3475357e0de168fc775c987bbR174-R177

jaysoo:
Added some comments. I think the rest looks fine.

jaysoo:
We'll need to check what's still needed from `withNx` and what features require the full Nx context (task name, graph, etc.). If we can simplify the new setup we should, and leave legacy behavior for existing projects.

ndcunningham:
> We'll need to check what's still needed from `withNx` and what features require the full Nx context (task name, graph, etc.). If we can simplify the new setup we should, and leave legacy behavior for existing projects.

Updated:
- Now we don't generate `.nx-helpers` nor `next.config.js` inside `dist`
- For those we will need nx context (_Docker_) will need `require` @nx/next` as a prod dependency

packages/next/plugins/with-nx.ts
```diff
@@ -237,6 +234,24 @@ function withNx(
           nextConfig.distDir && nextConfig.distDir !== '.next'
             ? joinPathFragments(outputDir, nextConfig.distDir)
             : joinPathFragments(outputDir, '.next');
+      } else {
+        // if we don't have outputPath set, we are probably running a build using run-commands.
+        // So we need to read distDir from the project's next.config.js. and generate the dist relative to workspace root.
+        // Here we assume the cwd is the projectRoot called from plugins @nx/next/plugin PCV3.
+
+        const nextJSProjectRoot = findNextConfigPath(projectDirectory);
```
jaysoo, line 242: I wonder if we want to do this, without our executor the dist folder would just contain `dist/<app>/.next` right? The `public`, `next.config.js`, `package.json` are missing?

If we only write `.next` to dist, perhaps it's better to leave it as the default `<project-src>/.next`? That way the source folder is deployable, otherwise the `dist` directory isn't deployable anyway.

jaysoo (approved):

PR #20214, opened 2023-11-13
feat(angular): support application builder for cypress component testing

Pascalmh:
## Current Behavior
Using the Application Builder as the executor the "component-test" target fails using cypress because the `main` property in the `options`-object was renamed to `browser`.

## Expected Behavior
Use the `browser` property when the `main` property is missing.

See: https://angular.io/guide/esbuild#using-the-application-builder

### DIFF
--- a/packages/angular/plugins/component-testing.ts
+++ b/packages/angular/plugins/component-testing.ts
@@ -73,7 +73,7 @@ ${e.stack ? e.stack : e}`
   const buildTarget = getBuildableTarget(ctContext);
 
   if (!buildTarget.project && !graph.nodes?.[buildTarget.project]?.data) {
-    throw new Error(stripIndents`Unable to find project configuration for build target. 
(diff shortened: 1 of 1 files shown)


leosvelperez (approved):

packages/angular/plugins/component-testing.ts
```diff
@@ -322,6 +322,7 @@ function withSchemaDefaults(options: any): BrowserBuilderSchema {
   options.outputHashing ??= 'none';
   options.progress ??= true;
   options.scripts ??= [];
+  options.main = options.main ??= options.browser;
```
alfaproject, line 325:
```js
  options.main ??= options.browser;
```

Pascalmh: Good catch, removed the check here: #20382

I like the co-location of the checks that throw errors more then having only a single condition/branch.

PR #20224, opened 2023-11-13
fix(angular): serve host app with --skipRemotes='*'

sokolovstas:
set `--all=false` for static remotes building in `module-federation-dev-server`

## Current Behavior
When I run command `nx serve web-app --skipRemotes='*'` serve failed because I don't have remotes to build. `run-many` exit with code not 0

## Expected Behavior
When I run command `nx serve web-app --skipRemotes='*'` no remotes builded and only host served


Coly010 (commented):
I’m not sure this is the correct approach to solving this issue. 

I’ll have a think about it and respond when I have a better approach.

PR #20243, opened 2023-11-14
feat(webpack): add plugin to automatically configure build and serve targets

jaysoo:
This PR adds support for projects using `webpack.config.js` files to have their `build`, `serve`, and `preview` targets automatically configured by `@nx/webpack/plugin`.

Just need to add to your `nx.json` file:

```json5
{
  //...
  plugins: ['@nx/webpack/plugin']
}
```

The plugin will support both `composePlugins` and `NxWebpackPlugin` usages. The former requires a new `isNxWebpackComposablePlugin` function that checks if a `nxWebpackComposablePlugin` exists on the exported function. Otherwise, the function could be a normal [webpack config function](https://webpack.js.org/configuration/configuration-types/#exporting-a-function).

When `NX_PCV3=true`, Wwe'll also generate projects with options set in webpack config files using `NxWebpackPlugin`, rather than in `project.json`.

e.g.

```js
const { NxWebpackPlugin } = require('@nx/webpack');
const { NxReactWebpackPlugin } = require('@nx/react');

// Standard webpack config format, no Nx-specific logic
module.exports =  {
  plugins: [
    new NxWebpackPlugin({
      compiler: 'babel',
      index: './src/index.html',
      main: './src/main.tsx',
      tsConfig: './tsconfig.app.json',
      baseHref: '/',
      assets: ['./src/favicon.ico', './src/assets'],
      styles: ['./src/styles.css'],
    }),
    NxReactWebpackPlugin({
      // Uncomment this line if you don't want to use SVGR
      // See: https://react-svgr.com/
      // svgr: false
    }),
  ],
};
```

If `NX_PCV3=true` is not set, then everything behaves as normal.

**Note:** We use relative paths (e.g. `main: './src/main.tsx'`), or absolute paths (e.g. `output: { path: join(__dirname, '...') }`). Previously inside targets, we used paths relative to workspace root, which doesn't make sense in the new setup. You can still use paths from the workspace root to make migration easier.

## Summary of changes

- Add `@nx/webpack/plugin` that uses `CreateNodes`  to infer build, serve, and preview targets
- Support standard webpack configuration, even the functional format that exports a `(env, args) -> config` [function](webpack.js.org/configuration/configuration-types/#exporting-a-function)
- Option defaults are moved from `schema.json` of webpack build executor to plugins (e.g. `withNx` and `NxWebpackPlugin`)
- Our webpack plugins can load in both the "reading" phase of Nx (e.g. during `CreateNodes`), and during actual task run (e.g. webpack build) -- need to read out options like `config.output.path` to determine what's cached
- Our composable webpack plugins (e.g. `withNx`) works with normal webpack CLI
- Add `webpack-cli` as a workspace dep in `@nx/webpack:init` generator
- `withWeb` is merged into `withNx`, and a deprecation message is show to user when used -- this should not impact most users, since `withWeb` is not used for React
- Generated apps will have options, such as `main` and `tsConfig`, configured in `webpack.config.js` rather than `project.json` -- existing apps will continue to function as before

##  TODO
- [x] Cache targets so we don't slow down project graph generation
- [x] e2e tests to cover PCv3 cases


mandarini (approved):

packages/webpack/src/executors/dev-server/lib/get-dev-server-config.ts
```diff
@@ -47,11 +25,14 @@ function getDevServerPartial(
   }
 
   const config: WebpackDevServerConfiguration = {
-    host: options.host,
-    port: options.port,
+    // TODO(wip): These should be read from actual config
+    host: serveOptions.host ?? 'localhost',
```
jaysoo, line 29: remove these

packages/webpack/src/plugins/plugin.ts
```diff
@@ -0,0 +1,185 @@
...
+    context.nxJsonConfiguration.targetDefaults
+  );
+
+  if (buildTargetDefaults?.cache === undefined) {
+    targets[options.buildTargetName].cache = true;
+  }
+
+  if (buildTargetDefaults?.inputs === undefined) {
+    targets[options.buildTargetName].inputs =
+      'production' in namedInputs
```
FrozenPandaz, line 143: This should now have an input of the `webpack-cli`.

https://github.com/nrwl/nx/blob/a57122d47c0b9d2ae94facd5a865645093067342/packages/cypress/src/plugins/plugin.ts#L338C41-L338C41

FrozenPandaz (changes requested):

FrozenPandaz (approved):

ndcunningham (approved):

PR #20250, opened 2023-11-15
docs(core): nx-cloud concepts section

isaacplmann:
[Creates a Concepts section under Nx Cloud.](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts)

No content changes, just moving pages around.

- Nx Cloud -> Account was split between Concepts and Recipes
- Improve Worst Case CI Times was moved from Nx Concepts to Cloud Concepts
- Nx Cloud -> Account -> Users was deleted
- Moved Nx Cloud Workflows under features instead of having it in the Intro section


nx-dev/nx-dev/redirect-rules.js
```diff
@@ -437,6 +437,13 @@ const nxCloudUrls = {
     '/nx-cloud/recipes/on-premise/auth-saml-managed',
   '/nx-cloud/private-cloud/advanced-config':
     '/nx-cloud/recipes/on-premise/advanced-config',
+  '/concepts/dte': '/nx-cloud/concepts/dte',
+  '/nx-cloud/intro/nx-cloud-workflows': '/nx-cloud/concepts/nx-cloud-workflows',
```
juristr, line 441: I think it should be `/nx-cloud/features/nx-cloud-workflows`.

docs/generated/manifests/cloud.json
```diff
@@ -40,16 +30,6 @@
...
-    "id": "nx-cloud-workflows",
-    "name": "Nx Cloud Workflows",
-    "description": "",
-    "file": "nx-cloud/intro/nx-cloud-workflows",
-    "itemList": [],
-    "isExternal": false,
-    "path": "/nx-cloud/intro/nx-cloud-workflows",
-    "tags": []
-  },
   "/nx-cloud/features": {
```
juristr, line 33: I'd call this section "Core Features" just like we do on the Nx section.

juristr (changes requested):
Can we improve the [Improve Worst Case CI Times](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts/dte)
- I'd change the title to a concept page that speaks about "Improving CI Times" or "Improving CI Speed"
- The structure could be
	- Running tasks serially, like `nx build app1`, `nx build app2`, `nx test...` etc
	- Running tasks in parallel with `run-many` and `--parallel` => but obviously process limits, doesn't scale etc..
	- Affected commands
	- Manual binning
	- Distribution with Nx Cloud
- I think we can still keep the pros/cons, but I'd not have it as section titles, but rather as a table at the end of each of the sections; so we can explain the concept and then have the pros/cons table for a quick overview
- the idea I'd want to convey is that these things can be applied incrementally;

We need to improve the [Security Scenarios](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts/scenarios) as discussed already previously; we can also do that in a subsequent PR

[End to End Encryption](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts/encryption) also needs more info. Like setting an encryption key: is that even something we'd recommend? I'd probably rather want to have an encryption key as part of an env variable..

docs/generated/manifests/cloud.json
isaacplmann: In the Nx section, we call it Core Features to distinguish them from Plugin Features.  With the line being blurred between plugin features and core features with project config v3, I think we should start to move away from the Core/Plugin distinction and eventually flatten the Nx feature list to all be under a single Features list.

That's why I think the Nx Cloud/CI features should be called Features and not Core Features

isaacplmann:
> Can we improve the [Improve Worst Case CI Times](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts/dte)
> 
> * I'd change the title to a concept page that speaks about "Improving CI Times" or "Improving CI Speed"
> * The structure could be
>   
>   * Running tasks serially, like `nx build app1`, `nx build app2`, `nx test...` etc
>   * Running tasks in parallel with `run-many` and `--parallel` => but obviously process limits, doesn't scale etc..
>   * Affected commands
>   * Manual binning
>   * Distribution with Nx Cloud
> * I think we can still keep the pros/cons, but I'd not have it as section titles, but rather as a table at the end of each of the sections; so we can explain the concept and then have the pros/cons table for a quick overview
> * the idea I'd want to convey is that these things can be applied incrementally;
> 
> We need to improve the [Security Scenarios](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts/scenarios) as discussed already previously; we can also do that in a subsequent PR
> 
> [End to End Encryption](https://nx-dev-git-fork-isaacplmann-docs-ci-concepts-nrwl.vercel.app/nx-cloud/concepts/encryption) also needs more info. Like setting an encryption key: is that even something we'd recommend? I'd probably rather want to have an encryption key as part of an env variable..

Yes, those pages need to be updated, but this PR is just for moving things around, not changing content.

nx-dev/nx-dev/redirect-rules.js
isaacplmann: Good catch.  Fixed.

juristr (approved):

PR #20260, opened 2023-11-15
feat(web): add verbose logging to waitForPortOpen function

jaysoo:
When we call `waitForPortOpen` in our executors (e.g. React/Angular module federation executors), the timeout could be triggered by an error code other than what we expect (`ECONNREFUSED`, `ECONNRESET`). This change allows easier debugging when the timeout occurs unexpectedly.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/web/src/utils/wait-for-port-open.ts
+++ b/packages/web/src/utils/wait-for-port-open.ts
@@ -1,10 +1,12 @@
+import { logger } from '@nx/devkit';
 import * as net from 'net';
 
 export function waitForPortOpen(
   port: number,
   options: { host?: string; retries?: number; retryDelay?: number } = {}
 ): Promise<void> {
-  const allowedErrorCodes = ['ECONNREFUSED', 'ECONNRESET'];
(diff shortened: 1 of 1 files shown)


ndcunningham (approved):

packages/web/src/utils/wait-for-port-open.ts
```diff
@@ -23,6 +25,11 @@ export function waitForPortOpen(
 
       client.once('error', (err) => {
         if (retries === 0 || !allowedErrorCodes.includes(err['code'])) {
+          if (process.env['NX_VERBOSE_LOGGING'] === 'true') {
```
ndcunningham, line 28: does `--verbose` set this to true?

Coly010: It should yeah

jaysoo: yep :)

Coly010 (approved):

PR #20264, opened 2023-11-15
feat(linter): add create-nodes plugin

meeroslav:
## Related Issue(s)

Fixes #


packages/eslint/src/plugins/plugin.ts
```diff
@@ -0,0 +1,104 @@
...
+    context.nxJsonConfiguration.targetDefaults,
+    '@nx/eslint:lint'
+  );
+
+  const isRootProject = projectRoot === '.';
+
+  const targets: Record<string, TargetConfiguration> = {};
+
+  const command = isFlatConfig(rootEslintConfigFile)
+    ? `ESLINT_USE_FLAT_CONFIG=true eslint`
```
meeroslav, line 73: This would not work for Windows so we need to find a better way of setting ENV vars

packages/eslint/src/plugins/plugin.ts
```diff
@@ -0,0 +1,128 @@
...
+  targets[options.targetName] = {
+    ...baseTargetConfig,
+    cache: targetDefaults?.cache ?? true,
+    inputs: targetDefaults?.inputs ?? [
+      'default',
+      `{workspaceRoot}/${rootEslintConfigFile}`,
+      '{workspaceRoot}/tools/eslint-rules/**/*',
+      { externalDependencies: ['eslint'] },
+    ],
+    outputs: targetDefaults?.outputs ?? ['{options.outputFile}'],
```
meeroslav, line 114: Remove this

packages/eslint/src/plugins/plugin.ts
```diff
@@ -0,0 +1,128 @@
...
+      return false;
+    }
+    return true;
+  }
+  // if it has an eslint config it's lintable
+  if (siblingFiles.some((f) => ESLINT_CONFIG_FILENAMES.includes(f))) {
+    return true;
+  }
+  // check whether the root has an eslint config
+  return readdirSync(workspaceRoot).some((f) =>
```
meeroslav, line 73: Go up the tree to find all parent configs

packages/eslint/package.json
```diff
@@ -27,16 +27,37 @@
     "requirements": {},
     "migrations": "./migrations.json"
   },
+  "exports": {
```
FrozenPandaz, line 30: How come this was needed?

packages/eslint/package.json
```diff
@@ -27,16 +27,37 @@
...
+    "./executors": "./executors.js",
+    "./src/executors/*/schema.json": "./src/executors/*/schema.json",
+    "./src/executors/*.impl": "./src/executors/*.impl.js",
+    "./src/utils/flat-config": "./src/utils/flat-config.js",
+    "./plugin": "./plugin.js"
+  },
   "executors": "./executors.json",
   "generators": "./generators.json",
+  "nx-migrations": {
+    "migrations": "./migrations.json"
```
FrozenPandaz, line 51: We have migrations above.

FrozenPandaz (changes requested):

packages/eslint/package.json
meeroslav: I based it off of @AgentEnder's PR assuming that is intended direction.

I will revert this block entirely

FrozenPandaz (approved):

PR #20308, opened 2023-11-17
docs(misc): remove callout to use @nx/angular:move generator

Pascalmh:
@nx/angular:move is deprecated and will be removed in  Nx v18


meeroslav (approved):

leosvelperez:
@Pascalmh thanks for the contribution! Could you rebase and edit the commit message so it passes our commit checks? A correct commit message would be `docs(misc): remove callout to use @nx/angular:move generator`.

Pascalmh:
> @Pascalmh thanks for the contribution! Could you rebase and edit the commit message so it passes our commit checks?

Done ✅

PR #20313, opened 2023-11-17
feat(linter): default lintFilePatterns to {projectRoot}

meeroslav:
This PR simplifies the `@nx/eslint:lint` executor options by using following defaults:

- `lintFilePatterns` is not required any more
- default value of `lintFilePatterns` is `{projectRoot}`
- default executor config will be generated without `lintFilePatterns` (unless it's standalone, see below)
- lint project generator will automatically set `./src` to root/standalone projects unless `lintFilePatterns` is passed via options
- our internal `@nx/...` generators now don't need to pass any lint patterns when triggering the lint project generator
- migration will convert all existing patterns to `{projectRoot}` if there are other non project paths otherwise remove the lint file pattern completely

## Related Issue(s)

Fixes #


docs/shared/node-server-tutorial/affected-project-graph.json
```diff
@@ -324,7 +322,7 @@
             "inputs": ["default", "{workspaceRoot}/.eslintrc.json"],
             "executor": "@nrwl/linter:eslint",
             "outputs": ["{options.outputFile}"],
-            "options": { "lintFilePatterns": ["./**/*.ts"] },
+            "options": { "lintFilePatterns": ["./src"] },
```
JamesHenry, line 325: The answer may come later in this PR, but how come this is a special case and still has a config?

nx.json
```diff
@@ -102,16 +102,7 @@
...
-          "{projectRoot}/**/*.{ts,tsx,js,jsx}",
-          "{projectRoot}/**/executors/**/schema.json",
-          "{projectRoot}/**/generators/**/schema.json",
-          "{projectRoot}/generators.json",
-          "{projectRoot}/executors.json",
-          "{projectRoot}/package.json",
-          "{projectRoot}/project.json",
-          "{projectRoot}/migrations.json"
-        ]
+        "lintFilePatterns": ["{projectRoot}"]
```
JamesHenry, line 105: Can't this be removed?

packages/eslint/src/generators/lint-project/lint-project.ts
```diff
@@ -71,19 +59,29 @@ export async function lintProjectGenerator(
...
     executor: '@nx/eslint:lint',
     outputs: ['{options.outputFile}'],
-    options: {
-      lintFilePatterns: lintFilePatterns,
-    },
   };
 
+  let lintFilePatterns = options.eslintFilePatterns;
+  if (!lintFilePatterns && options.rootProject) {
+    lintFilePatterns = ['./src'];
```
JamesHenry, line 69: Ok so this is the answer to my previous question... but is this correct?

In the case of other projects we lint other files like the package.json and project.json by default, but this wouldn't do that?

packages/eslint/src/migrations/update-17-2-0/simplify-eslint-patterns.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+describe('simplify-eslint-patterns migration', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+    tree.write('.eslintrc.json', '{}');
+  });
+
+  it('should remove pattern in matches default', async () => {
```
JamesHenry, line 14:
```suggestion
  it('should remove pattern if matches default', async () => {
```

packages/eslint/src/migrations/update-17-2-0/simplify-eslint-patterns.ts
```diff
@@ -0,0 +1,67 @@
...
+export default async function update(tree: Tree) {
+  const projects = getProjects(tree);
+
+  for (const [projectName, projectConfiguration] of projects) {
+    let needsUpdate = false;
+
+    for (const [targetName, targetConfig] of Object.entries(
+      projectConfiguration.targets ?? {}
+    )) {
+      // this will be in a broken state after the package is globally renamed
```
JamesHenry, line 18: It's not 100% clear to me what this comment means sorry?

Are you about to globally rename the package?

packages/next/src/executors/build/lib/test-fixtures/ensure-exts/baz.json
JamesHenry, line 1: What is this file related to?

JamesHenry (changes requested):
Please see comments inline

docs/shared/node-server-tutorial/affected-project-graph.json
meeroslav: It's mentioned in the description of the PR:
> lint project generator will automatically set ./src to root/standalone projects unless lintFilePatterns is passed via options

Standard nested projects get `{projectRoot}` as a lint file pattern. This wouldn't work for root/standalone as it's root is `.` which would cause all files in the monorepo to be linter. That's why for those projects we use `./src` as a lint file pattern.

packages/eslint/src/generators/lint-project/lint-project.ts
meeroslav: This is how it has been working so far. You can always override this behavior by explicitly including those files via `lintFilePatterns`. Unfortunately, it's the limitation of the standalone projects.

packages/eslint/src/migrations/update-17-2-0/simplify-eslint-patterns.ts
meeroslav: What this means is after this command is run:
```
await replaceNrwlPackageWithNxPackage(tree, '@nx/linter', '@nx/eslint');
```
the executor is now set to `@nx/eslint:eslint` which is an invalid name so we need to change it to `@nx/eslint:lint`

How would you rephrase this?

packages/next/src/executors/build/lib/test-fixtures/ensure-exts/baz.json
meeroslav: This is a fixture, that fails after the change due to invalid json file.

Once we changed the lint file pattern to include all files it lints also the `json` files (thanks to included `jsonc-eslint-parser`)

packages/eslint/src/migrations/update-17-2-0/simplify-eslint-patterns.spec.ts
meeroslav: Thanks, pushing the fix.

packages/eslint/src/generators/lint-project/lint-project.ts
JamesHenry: But we could include project.json and package.json explicitly in the patterns couldn't we?

JamesHenry: I understand not including all files, but we could include ones explicitly that we know to be important?

meeroslav: This is the default generator that never includes the `package.json` by default unless it's a buildable lib (which case is handled underneath on line 76). 

Users can always subsequently add any paths they want, and this only simplifies the default generation.

meeroslav: Moreover, having a standalone buildable library with `@nx/dependency-checks` (use case where we do want to test package.json) does not make too much sense since the root `package.json` would have to include a lot of additional dependencies which would not be part of the package so there would have to be a lot of `ignores` set in place.

JamesHenry: Ok sounds good, resolved 👍

JamesHenry (approved):

FrozenPandaz (approved):

juristr (approved):

mandarini (approved):

isaacplmann (approved):

jaysoo (approved):

PR #20315, opened 2023-11-17
fix(release): default to only publishing libs when no config is defined

fahslaj:
## Current Behavior

`nx release version` fails in a newly created nx repo with any `apps`.

## Expected Behavior

`nx release version` succeeds in a newly created nx repo with any `apps`.

## Related Issue(s)

Fixes #


JamesHenry (changes requested):
Feedback on slack

JamesHenry (approved):

PR #20316, opened 2023-11-17
feat(release): update dist-tags when publishing a package version that already exists

fahslaj:
## Current Behavior

`nx release` tries to publish a package whose version already exists in the remote repository.

## Expected Behavior

`nx release` recognizes if a package version has already been published and intelligently skips it OR updates its dist tag if necessary.

## Related Issue(s)

Fixes #


packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -72,6 +76,100 @@ export default async function runExecutor(
...
+      env: processEnv(true),
+      cwd: packageRoot,
+      stdio: ['ignore', 'pipe', 'pipe'],
+    });
+
+    const resultJson = JSON.parse(result.toString());
+    const distTags = resultJson['dist-tags'] || {};
+    if (distTags[tag] === currentVersion) {
+      console.warn(
+        `Skipped ${packageTxt}, as v${currentVersion} already exists in ${registry} with tag "${tag}"`
```
JamesHenry, line 92:
```suggestion
        `Skipped ${packageTxt} because v${currentVersion} already exists in ${registry} with tag "${tag}"`
```

It's a minor thing but I just think in this context "as" is a little more ambiguous and it made me pause when scanning (and removed the comma https://www.grammarly.com/blog/comma-before-because/)

packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -72,6 +76,100 @@ export default async function runExecutor(
...
+
+          if (context.isVerbose) {
+            console.error('npm dist-tag add stdout:');
+            console.error(JSON.stringify(stdoutData, null, 2));
+          }
+          return {
+            success: false,
+          };
+        } catch (err) {
+          // npm v9 onwards seems to guarantee stdout will be well formed JSON when --json is used, so maybe we need to
```
JamesHenry, line 143: You can remove this comment (and the original one you copied it from) now I think, we only support node 18+ in nx 17

packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -72,6 +76,100 @@ export default async function runExecutor(
...
+          );
+          return {
+            success: false,
+          };
+        }
+      }
+    }
+  } catch (err) {
+    const stdoutData = JSON.parse(err.stdout?.toString() || '{}');
+    // If the error is that the package doesn't exist, then we can ignore it
```
JamesHenry, line 157:
```suggestion
    // If the error is that the package doesn't exist, then we can ignore it because we will be publishing it for the first time in the next step
```

This comment currently is more of a "what", which the code itself already explains, it would be more useful to include the "why". Hopefully my inference here is correct, but please feel free to reword my suggestion regardless

e2e/release/src/repeat-publish.test.ts
JamesHenry, line 1: Let's please avoid the overhead of a dedicated e2e workspace just for this functionality, we can instead cover it by adding it to the main existing release e2e tests. In some cases the workflow or usage is complex or unique enough to warrant a separate file, and therefore workspace, but here I don't think that's the case

JamesHenry (changes requested):

JamesHenry (approved):

PR #20375, opened 2023-11-22
feat(js): add js create nodes plugin to register @nx/js:tsc build targets

leosvelperez:
Adds support for projects using TypeScript config files to have their `build` target automatically configured by `@nx/js/plugin`. If another build tool configuration file is found at the project root, the target won't be configured.

## Related Issue(s)

Fixes #


packages/js/src/plugins/plugin.ts
```diff
@@ -0,0 +1,379 @@
...
+  const packageJsonPath = join(workspaceRoot, projectRoot, 'package.json');
+  if (
+    !existsSync(packageJsonPath) ||
+    readProjectPackageJson(workspaceRoot, projectRoot).private
+  ) {
+    return false;
+  }
+
+  // TODO(leo): is there a better way?
+  const blacklistedConfigFiles = [
```
Coly010, line 233: Can we add remix config files to this also?

JamesHenry:
Set to draft pending updates we discussed

leosvelperez:
This needs to be revisited and updated. I'm closing it until it gets picked up again.

PR #20384, opened 2023-11-23
docs(misc): document NX_BUILDABLE_LIBRARIES_TASK_GRAPH environment variable

leosvelperez:
## Related Issue(s)

Fixes #


docs/shared/reference/environment-variables.md
```diff
@@ -2,30 +2,31 @@
...
+| NX_PREFER_TS_NODE                 | boolean | If set to `true`, Nx will use ts-node for local execution of plugins even if `@swc-node/register` is installed.                                                                                                                        |
+| NX_IGNORE_CYCLES                  | boolean | If set to `true`, Nx will ignore errors created by a task graph circular dependency. Can be overriden on the command line with `--nxIgnoreCycles`                                                                                      |
+| NX_BATCH_MODE                     | boolean | If set to `true`, Nx will run task(s) in batches for executors which support batches.                                                                                                                                                  |
+| NX_BUILDABLE_LIBRARIES_TASK_GRAPH | boolean | If set to `true`, Nx will use the task graph to identify the dependent buildable libraries' targets instead of relying on finding dependent libraries with targets with the same name.                                                 |
```
isaacplmann, line 29: Why is this not the default?

isaacplmann (approved):

docs/shared/reference/environment-variables.md
leosvelperez: We'll first advise to try it out in a couple of reported issues to gather feedback and ensure we don't break existing setups. We'll switch the default soon.

leosvelperez: I'll hold this off. I'm gonna wait to see if some folks reply in the issues I pinged about using this. Based on the feedback, we might switch the default, in which case, I'll rename the variable to mean the opposite. We'll keep for a while as a escape hatch and once we're confident it works, we'll remove it.

PR #20429, opened 2023-11-27
docs(core): reduce waste in CI

isaacplmann:
Adds a new concept page: [Reduce Waste in CI](https://nx-dev-git-fork-isaacplmann-docs-faster-ci-nrwl.vercel.app/nx-cloud/concepts/reduce-waste)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -82,6 +82,16 @@ Your content goes here.
 {% /callout %}
 ```
 
+#### Disclosure
+
+A disclosure can be used for less important information that is initially collapsed.
+
+```markdown
+{% disclosure title="string" %}
+Your content goes here.
+{% /disclosure %}
+```
+
 #### Cards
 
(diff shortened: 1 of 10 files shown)


juristr (approved):

docs/README.md
```diff
@@ -82,6 +82,16 @@ Your content goes here.
 {% /callout %}
 ```
 
+#### Disclosure
```
juristr, line 85: Minor comment, but can't we just add a "collapsible" option to the existing callouts?

isaacplmann: Yes, I debated whether I should do it that way.  I'd be happy to reimplement it as the same component, but I did as a separate component because the original callout component markup would have to be rearranged to account for the collapsible bar across the top.

I was erring on the side of repeating code so that different components that happen to look similar can evolve independently.  These two components are very similar and could definitely be implemented as the same component though.

PR #20432, opened 2023-11-27
fix(module-federation): fix static remote port determination

bigcakes:
## Current Behavior
If a staticRemotesPort is not passed in to the module federation dev server and a remote's port is less than the default staticRemotesPort, undefined is return and module federation dev server fails to start

## Expected Behavior
Module federation dev server starts with a calculated staticRemotesPort if it is not passed


Coly010 (approved):

ndcunningham (approved):
LGTM,  can you rebase to fix the conflicts?
thanks!

PR #20442, opened 2023-11-28
fix: nestjs lib generator pass skipPackageJson flag to js lib generator

ckaeslin:
## Current Behavior

Using the nestjs library generator with skipPackageJson: true adds ts-node to package.json.

## Expected Behavior

skipPackageJson: true in nestjs lib generator should not add anything to package.json

## Related Issue(s)

Fixes #

Pass the skipPackageJson flag to the js lib generator so that package.json is not adjusted if the flag is set to true in the nest lib generator.


MikeFear (approved):

ckaeslin:
@jaysoo this tiny PR would help us to fix our generator, is there a chance to get this merged in the near future?

mandarini (approved):

jaysoo (approved):

PR #20466, opened 2023-11-29
feat(graph): add project details view

MaxKless:
This adds a very rudimentary Project Details view to the graph app, available under `/project-details/:projectName`.
You can open this view by calling `nx show project project1 --web`.

By hovering over individual properties, the source map will be used to display which plugin & file resulted in this property being set.

Disclaimer: the view is ugly and will be improved in the future. More interactivity with the rest of the graph app will also be added.


isaacplmann (approved):

graph/project-details/src/lib/nx-welcome.tsx
philipjfulcher, line 1: todo: this file should be deleted

graph/client/src/app/routes.tsx
```diff
@@ -174,4 +210,13 @@ export const releaseRoutes: RouteObject[] = [
       ...childRoutes,
     ],
   },
+  {
+    path: 'project-details/:projectName',
+    id: 'selectedProjectDetails',
+    element: <ProjectDetails />,
+    loader: async ({ request, params }) => {
```
philipjfulcher, line 217: suggestion: this route should be feature-flagged in release so that user can't reach it unintentionally. 

You can feature flag the route in the loader by adding a check for the `showExperimentalFeatures` property in the environment:

```typescript
    const environmentConfig = getEnvironmentConfig();

    if (!environmentConfig.appConfig.showExperimentalFeatures) {
        return redirect(`/projects`);
    }
```

philipjfulcher (approved):
One small nitpick, one suggestion. Otherwise, it's a good starting point.

graph/client/src/app/routes.tsx
MaxKless: that would mean we can only look at the view through the dev server and break `nx --show`...
 With the hidden `--web` flag, I think chances are already pretty low of anyone stumbling over it.

packages/nx/src/command-line/graph/graph.ts
```diff
@@ -749,6 +785,10 @@ async function createExpandedTaskInputResponse(
   return response;
 }
 
+function createSourceMapsResponse() {
+  return readSourceMaps();
```
FrozenPandaz, line 789: This reads it from the file? Can we make a new function which will build a graph and return the source map?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #20471, opened 2023-11-29
feat(core): introduce workspace file archive

Cammisuli:
## Current Behavior

When we run Nx without the daemon, we constantly read and hash all files in the workspace. 

## Expected Behavior

When we read the workspace the first time, we will then save the hashes in the `.nx/cache` folder. This archive will then be read on subsequent runs comparing the modified times, and only hashing files that have changed or do not exist in the archive

## Related Issue(s)

Fixes #


packages/nx/src/lib.rs
```diff
@@ -1,3 +1,10 @@
+#[cfg(not(target_env = "msvc"))]
+use tikv_jemallocator::Jemalloc;
+
+#[cfg(not(target_env = "msvc"))]
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
```
Cammisuli, line 6: Only macos/linux - this essentially optimizes allocations, and increases performance for large allocs from vecs/hashmaps

FrozenPandaz (approved):

PR #20510, opened 2023-11-30
fix(linter): check if eslint plugin exists before adding target

mandarini:
only check `NX_PCV3` on `init` generator. In other generators check if plugin exists.


meeroslav (changes requested):
I don't think this should be fixed on init project but rather on init function.

If plugin driven target is requested but it's missing, we should create it.

meeroslav (approved):

PR #20524, opened 2023-12-01
fix(core): unloading expanded envs

Riron:
## Current Behavior
When an env is expandable, it is not being unloaded before loading a task environment variables. Hence the initial value (from `bin/nx.ts`) is kept, and overwrites are not beings used when running the task.
This is because the `unloadDotEnvFiles()` function is comparing expanded variables with non expanded variables.

## Expected Behavior
Every `.env`, `.env.local` and `.local.env` should be unloaded before loading the appropriate envs for a task.

### DIFF
--- a/packages/nx/src/tasks-runner/task-env.ts
+++ b/packages/nx/src/tasks-runner/task-env.ts
@@ -187,7 +187,8 @@ function loadDotEnvFilesForTask(
 function unloadDotEnvFiles(environmentVariables: NodeJS.ProcessEnv) {
   const unloadDotEnvFile = (filename: string) => {
     let parsedDotEnvFile: NodeJS.ProcessEnv = {};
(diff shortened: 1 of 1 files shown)


Cammisuli:
Lets wait for https://github.com/nrwl/nx/pull/19401 to merge in, then we can add the same check in here with:
```
if (process.env.NX_EXPAND_ENV !== 'false') {
	/// expand logic
}
```

Riron:
> Lets wait for #19401 to merge in, then we can add the same check in here with:
> 
> ```
> if (process.env.NX_EXPAND_ENV !== 'false') {
> 	/// expand logic
> }
> ```

If #19401 takes too long to get merged, should we consider merging this one first and instead adding the condition directly in #19401 ?
This is quite a pain point for us :)

xiongemi:
there is another pr https://github.com/nrwl/nx/pull/22585 that also tries to fix the same issue. i am going to close this one.

PR #20535, opened 2023-12-01
fix(release): add overall nx release command

fahslaj:
## Current Behavior

Versioning, changelog generation, and publishing must be done in distinct steps.

## Expected Behavior

Versioning, changelog generation, and publishing can be done in the same step by running:
`nx release [specifier]`

The command will prompt the user for confirmation before publishing if `--dry-run` is not set. This prompt can  be overridden with the `--no` and `--yes` flags, or their aliases, `-n` and `-y`.


docs/generated/packages/nx/documents/release.md
```diff
@@ -57,6 +57,114 @@ Show version number
 
 ## Subcommands
 
+### $0
+
+Create a version and release for the workspace, then generate a changelog
+
+```shell
+nx release $0 [specifier]
+```
```
fahslaj, line 66: The generated docs aren't correct for this case. Any suggestions on how to do a manual override?

docs/generated/cli/release.md
```diff
@@ -57,6 +57,114 @@ Show version number
 
 ## Subcommands
 
+### $0
+
+Create a version and release for the workspace, then generate a changelog
+
+```shell
+nx release $0 [specifier]
+```
```
fahslaj, line 66: The generated docs aren't correct for this case. Any suggestions on how to do a manual override?

packages/nx/src/command-line/release/meta.ts
```diff
@@ -0,0 +1,66 @@
...
+  args: VersionOptions & ChangelogOptions
+): Promise<NxReleaseVersionResult | number> {
+  const nxReleaseConfig = await readNxReleaseConfig(args);
+
+  const versionResult: NxReleaseVersionResult | 1 =
+    await releaseVersionCLIHandler({
+      ...args,
+      // if enabled, committing and tagging will be handled by the changelog
+      // command, so we should only stage the changes in the version command
+      stageChanges: args.gitCommit || nxReleaseConfig.git?.commit,
```
fahslaj, line 22: This logic is the only reason I need to parse the nx config below. If we can remove this, then we can leave the config parsing to the inner commands.

packages/nx/src/command-line/release/command-object.ts
```diff
@@ -116,32 +121,51 @@ export const yargsReleaseCommand: CommandModule<
...
+          'Automatically answer yes to the confirmation prompt for publishing',
+      })
+      .option('no', {
+        type: 'boolean',
+        alias: 'n',
+        description:
+          'Automatically answer no to the confirmation prompt for publishing',
+      })
+      .check((argv) => {
+        if (argv.yes && argv.no) {
```
JamesHenry, line 145: I think we want to error if they are both set at all, not just if both truthy

packages/nx/src/command-line/release/meta.ts
```diff
@@ -0,0 +1,102 @@
...
+  args: MetaOptions
+): Promise<NxReleaseVersionResult | number> {
+  const nxReleaseConfig = await readNxReleaseConfig(args);
+
+  const versionResult: NxReleaseVersionResult = await releaseVersion({
+    ...args,
+    // if enabled, committing and tagging will be handled by the changelog
+    // command, so we should only stage the changes in the version command
+    stageChanges: args.gitCommit || nxReleaseConfig.git?.commit,
+    gitCommit: false,
```
JamesHenry, line 28: I think we need some more config validation before proceeding.

If they have explicitly put `version.git.commit` and/or `version.git.tag` in their config, we should error before attempting to run the meta command (rather than just setting things to false here like you have) because it doesn't make sense.

There are other such permutations as well when it comes to changelog.

Therefore I think the simplest and clearest thing would be, any `version.git` or `changelog.git` config set on the user side is incompatible with the meta command and we hard error up front explaining that.

Root level `git` is naturally fine (and you are already appropriately leveraging it on line 27)

packages/nx/src/command-line/release/meta.ts
```diff
@@ -0,0 +1,102 @@
...
+    console.log('Skipped publishing packages.');
+  }
+
+  return versionResult;
+}
+
+async function promptForPublish(): Promise<boolean> {
+  console.log('\n');
+
+  const reply = await prompt<{ confirmation: 'yes' | 'no' }>([
```
JamesHenry, line 58: How come we didn't go for the native ConfirmPrompt? https://www.npmjs.com/package/enquirer#confirm-prompt

packages/nx/src/command-line/release/meta.ts
```diff
@@ -0,0 +1,102 @@
...
+      ],
+    },
+  ]);
+
+  console.log('\n');
+
+  return reply.confirmation === 'yes';
+}
+
+async function readNxReleaseConfig(
```
JamesHenry, line 81: This isn't consistent with the other three commands, let's just inline the statements to match them

JamesHenry (changes requested):
Having a `--no` option is definitely more unusual, but I get why you've added it.

The alternative would be forcing the user to use `version` and `changelog` separately if they wanted to programmatically prevent publishing, which as we know comes with different trade offs, so I think on balance having the `--no` makes sense.

I can't help thinking, though, that maybe `nx release --no-publish` reads better. In that case however we obviously lose the symmetry with the `--yes` option.

Maybe we could add `--publish`/`--no-publish` as aliases of `--yes/--no` so that all the options are available and folks can use whatever seems most intuitive to them?

e2e/release/src/independent-projects.test.ts
```diff
@@ -648,4 +648,58 @@ describe('nx release - independent projects', () => {
       `);
     });
   });
+
+  describe('meta', () => {
```
JamesHenry, line 652: Let's move away from meta and just call it release

packages/nx/src/command-line/release/release.ts
```diff
@@ -0,0 +1,110 @@
...
+  args: ReleaseOptions
+): Promise<NxReleaseVersionResult | number> {
+  const projectGraph = await createProjectGraphAsync({ exitOnError: true });
+  const nxJson = readNxJson();
+
+  if (args.verbose) {
+    process.env.NX_VERBOSE_LOGGING = 'true';
+  }
+
+  if (
```
JamesHenry, line 29: I think this is unnecessarily complex and granular. We can check for `Object.keys(version.git).length` and have the `jsonConfigErrorPath` be `version.git` (and same for changelog)

JamesHenry:
```ts
const hasVersionGitConfig =
  Object.keys(nxJson.release?.version?.git ?? {}).length > 0;
const hasChangeLogGitConfig =
  Object.keys(nxJson.release?.changelog?.git ?? {}).length > 0;

if (hasVersionGitConfig || hasChangeLogGitConfig) {
  const jsonConfigErrorPath = hasVersionGitConfig
    ? ["release", "version", "git"]
    : ["release", "changelog", "git"];
}
```

JamesHenry (approved):

docs/shared/reference/nx-json.md
```diff
@@ -259,3 +277,181 @@ As of Nx 17, if you only use one tasks runner, you can specify these properties
...
+For a full reference of the available options for the `@nx/js:release-version` generator, see the [release version generator reference](/nx-api/js/generators/release-version).
+
+#### Changelog
+
+The `changelog` property configures the changelog phase of the release process. It is used to generate a changelog for your projects, and commit it to your repository.
+
+There are two types of possible changelog that can be generated:
+
+- **Workspace Changelog**: A changelog that contains all changes across all projects in your workspace. This is not applicable when releasing projects independently.
+
```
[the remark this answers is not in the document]
JamesHenry: Release groups are intentionally not mentioned anywhere yet

JamesHenry (approved):

isaacplmann (approved):
We can merge this as is, but I'll move a lot of the `Release` section of the nx.json docs into a separate page in a future PR.

PR #20540, opened 2023-12-01
chore(core): add nightly canary release

JamesHenry:
## Current Behavior

Nx is not published automatically from master.

## Expected Behavior

Nx is automatically published from master once per day, at around 3am UTC, under the `canary` tag on npm.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,15 +1,18 @@
 name: publish
-env:
-  DEBUG: napi:*
-  NX_RUN_GROUP: ${{ github.run_id }}-${{ github.run_attempt }}
-  NPM_CONFIG_PROVENANCE: true
+
 on:
+  schedule:
+    - cron: "0 3 * * *" # Once per day, at 3am UTC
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -1,15 +1,18 @@
 name: publish
-env:
-  DEBUG: napi:*
-  NX_RUN_GROUP: ${{ github.run_id }}-${{ github.run_attempt }}
-  NPM_CONFIG_PROVENANCE: true
+
 on:
+  schedule:
+    - cron: "0 3 * * *" # Once per day, at 3am UTC
```
FrozenPandaz, line 5: Can we have this for weekdays only? 

Run Tuesday - Saturday (Picks up changes from Monday - Friday).

```suggestion
    - cron: "0 3 * * 2-6" # Tuesdays - Saturdays, at 3am UTC
```

I don't think we're going to land many changes over the weekends we'll just end up with the same version with different dates.

FrozenPandaz (approved):

PR #20542, opened 2023-12-01
feat(core): have plugins generate target defaults based on their exec…

FrozenPandaz:
…utors

## Related Issue(s)

Fixes #


FrozenPandaz:
We found an alternate solution here

PR #20562, opened 2023-12-04
feat(storybook): nodes plugin

mandarini:
A plugin (`@nx/storybook/plugin`) which will add Storybook targets/projects to the Nx project graph. This allows you to run Storybook for projects that have a `.storybook/main.ts` file without creating a `project.json` file.

## To-do's

- [x] READ CONFIG AND SEE IF IT'S ANGULAR - USE ANGULAR EXECUTORS IF IT IS
- [x] Figure out what to do with being unable to read outputs and port from config
- [x] Figure out what to do with being unable to set outputs and port in config
- [x] figure out why it looks for name
- [x] don't add targets if `NX_PCV3` is true
- [x] Test it thoroughly
- [x] Figure out how to specify `@storybook/nextjs` as framework (if no executor is present in project.json) (use graph??)
- [x] Make angular work - right now it does not find targets: Needs this to work: https://github.com/nrwl/nx/pull/21019

## How it works

On `@nx/storybook:init` this is added in the `nx.json`:
```
...
plugins: [
  {
    plugin: '@nx/storybook/plugin',
    options: {
      buildStorybookTargetName = 'build-storybook',
      serveStorybookTargetName = 'storybook',
      testStorybookTargetName = 'test-storybook',
      staticStorybookTargetName = 'static-storybook'
    },
  },
  ...
]
```
so that you can run `nx storybook my-app`, `nx build-storybook my-app` etc, without manually defining these targets in your projects' `project.json`.

The plugin generates these targets with bare minimum (no extra) options configuration. Any options you need for your Storybook app, you can pass through the CLI as normal.

### DIFF
--- a/packages/storybook/src/generators/configuration/lib/util-functions.ts
+++ b/packages/storybook/src/generators/configuration/lib/util-functions.ts
@@ -33,7 +33,7 @@ import { useFlatConfig } from '@nx/eslint/src/utils/flat-config';
 
 const DEFAULT_PORT = 4400;
 
-export function addStorybookTask(
+export function addStorybookTarget(
   tree: Tree,
   projectName: string,
   uiFramework: UiFramework,
@@ -82,7 +82,7 @@ export function addStorybookTask(
   updateProjectConfiguration(tree, projectName, projectConfig);
 }
 
-export function addAngularStorybookTask(
+export function addAngularStorybookTarget(
   tree: Tree,
   projectName: string,
   interactionTests: boolean
@@ -696,18 +696,32 @@ export async function getE2EProjectName(
   return e2eProject;
 }
 
-export function getViteConfigFilePath(
+export function findViteConfig(
   tree: Tree,
-  projectRoot: string,
-  configFile?: string
+  projectRoot: string
 ): string | undefined {
-  return configFile && tree.exists(configFile)
-    ? configFile
-    : tree.exists(joinPathFragments(`${projectRoot}/vite.config.ts`))
-    ? joinPathFragments(`${projectRoot}/vite.config.ts`)
-    : tree.exists(joinPathFragments(`${projectRoot}/vite.config.js`))
-    ? joinPathFragments(`${projectRoot}/vite.config.js`)
-    : undefined;
+  const allowsExt = ['js', 'mjs', 'ts', 'cjs', 'mts', 'cts'];
+
+  for (const ext of allowsExt) {
+    const viteConfigPath = joinPathFragments(projectRoot, `vite.config.${ext}`);
+    if (tree.exists(viteConfigPath)) {
+      return viteConfigPath;
+    }
+  }
+}
+
+export function findNextConfig(
+  tree: Tree,
+  projectRoot: string
+): string | undefined {
+  const allowsExt = ['js', 'mjs', 'cjs'];
+
+  for (const ext of allowsExt) {
+    const nextConfigPath = joinPathFragments(projectRoot, `next.config.${ext}`);
+    if (tree.exists(nextConfigPath)) {
+      return nextConfigPath;
+    }
+  }
 }
 
 export function renameAndMoveOldTsConfig(
--- a/packages/storybook/src/plugins/plugin.ts
+++ b/packages/storybook/src/plugins/plugin.ts
@@ -1,0 +1,375 @@
+import {
+  CreateDependencies,
+  CreateNodes,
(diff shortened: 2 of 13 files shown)


packages/storybook/src/generators/configuration/lib/util-functions.ts
```diff
@@ -696,18 +696,32 @@ export async function getE2EProjectName(
...
+      return viteConfigPath;
+    }
+  }
+}
+
+export function findNextConfig(
+  tree: Tree,
+  projectRoot: string
+): string | undefined {
+  const allowsExt = ['js', 'mjs', 'ts', 'cjs', 'mts', 'cts'];
```
ndcunningham, line 717: next config can only be `.js`, `.cjs` or `.mjs`

ndcunningham (approved):

packages/storybook/src/plugins/plugin.ts
```diff
@@ -0,0 +1,378 @@
...
+            'storybook',
+            '@storybook/angular',
+            '@storybook/test-runner',
+          ],
+        },
+      ],
+    };
+  } else {
+    targetConfig = {
+      command: `storybook build --config-dir ${dirname(
```
jaysoo, line 189: We need to install `@storybook/cli` for this to work?

packages/storybook/src/plugins/plugin.ts
```diff
@@ -0,0 +1,378 @@
...
+) {
+  if (frameworkIsAngular) {
+    return {
+      executor: '@storybook/angular:start-storybook',
+      options: {
+        configDir: `${dirname(configFilePath)}`,
+        browserTarget: `${projectName}:build-storybook`,
+        compodoc: false,
+      },
+      inputs: [
```
jaysoo, line 221: I don't think serve needs input since it's not cached.

packages/storybook/src/plugins/plugin.ts
```diff
@@ -0,0 +1,370 @@
...
+  workspaceRoot,
+  writeJsonFile,
+} from '@nx/devkit';
+import { dirname, isAbsolute, join, relative } from 'path';
+import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
+import { existsSync, readFileSync, readdirSync } from 'fs';
+import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
+import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
+import { getLockFileName } from '@nx/js';
+import { tsquery } from '@phenomnomnominal/tsquery';
```
jaysoo, line 19: I wonder if `tsquery` is the way to go, or can we import the file like we do in other plugins. Is there a reason querying is preferred?

mandarini: requiring it gives a 
```
import type { StorybookConfig } from '@storybook/angular';
   ^^^^^^
SyntaxError: Cannot use import statement outside a module
```
error.

AST is not that much of a performance concern in this case:
```
Time for '@nx/storybook/plugin:createNodes' 162.18487500026822
Time for '@nx/storybook/plugin:createNodes:apps/ngapp1/.storybook/main.ts' 0.4941249992698431
Time for '@nx/storybook/plugin:createNodes:apps/rapp1/.storybook/main.ts' 0.4677499998360872
Time for '@nx/storybook/plugin:createNodes:ngapp2/.storybook/main.ts' 0.37250000052154064
Time for '@nx/storybook/plugin:createNodes:rapp3/.storybook/main.ts' 0.39054099842906
Time for '@nx/storybook/plugin:createNodes:rapp4/.storybook/main.ts' 0.33058400079607964

my logs just for getStorybookConfig
getStorybookConfig took 5.05587499961257 milliseconds
getStorybookConfig took 1.6850829999893904 milliseconds
getStorybookConfig took 0.6668329983949661 milliseconds
getStorybookConfig took 0.5654579997062683 milliseconds
getStorybookConfig took 0.5120829995721579 milliseconds

for cypress, which requires the module
Time for '@nx/cypress/plugin:createNodes' 158.9198329988867
Time for '@nx/cypress/plugin:createNodes:rapp3-e2e/cypress.config.ts' 0.40424999967217445
Time for '@nx/cypress/plugin:createNodes:rapp4-e2e/cypress.config.ts' 0.30654199980199337
```

jaysoo (approved):

PR #20616, opened 2023-12-06
fix(testing): add webServerCommands/ciWebServerCommands option for cy…

FrozenPandaz:
…press configuration

## Current Behavior

The cypress generator looks at the project configuration in the tree to determine if a `serve` and `serve-static` target exist. For PCv3, the tree doesn't have the project configuration.

## Expected Behavior

There is an option which is used over looking at the project configuration in the tree. Apps should pass this in.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/src/generators/application/lib/add-e2e.ts
+++ b/packages/react/src/generators/application/lib/add-e2e.ts
@@ -18,10 +18,10 @@ export async function addE2e(
 ): Promise<GeneratorCallback> {
   switch (options.e2eTestRunner) {
     case 'cypress': {
-      if (
-        (options.bundler === 'webpack' && !hasWebpackPlugin(tree)) ||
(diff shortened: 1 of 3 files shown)


packages/react/src/generators/application/lib/add-e2e.ts
```diff
@@ -52,6 +52,15 @@ export async function addE2e(
         baseUrl: 'http://localhost:4200',
         jsx: true,
         rootProject: options.rootProject,
+        webServerCommands: hasNxBuildPlugin
+          ? {
+              default: `nx run ${options.projectName}:serve`,
+              production: `nx run ${options.projectName}:serve:production`,
```
jaysoo, line 58: Production configuration may not be there since the "production" server may be through `preview` target.

jaysoo (approved):

PR #20626, opened 2023-12-07
feat(vue): add nuxt as cnw vue framework

mandarini:
Just like for React we have the `framework` option for `Nextjs`, add the `framework` option for Vue to choose between `none` and `nuxt` when creating a new workspace.

```
npx create-nx-workspace my-org

 >  NX   Let's create a new workspace [https://nx.dev/getting-started/intro]

✔ Which stack do you want to use? · vue
? What framework would you like to use? …
None               I only want vue.
Nuxt       [  https://nuxt.com/  ]
```

Screenshot 2023-12-07 at 4 44 05 PM

### DIFF
--- a/docs/generated/cli/create-nx-workspace.md
+++ b/docs/generated/cli/create-nx-workspace.md
@@ -139,7 +139,7 @@ Package manager to use
 
 Type: `string`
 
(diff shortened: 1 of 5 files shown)


JamesHenry (changes requested):
Coming at this a little cold but shouldn't we have some coverage for when the framework is nuxt?

jaysoo (approved):

JamesHenry (approved):

PR #20639, opened 2023-12-07
feat(nx-dev): new year challenge

isaacplmann:
Adds a /tips page for the [Nx New Year Tips](https://nx-dev-git-fork-isaacplmann-docs-new-year-challenge-nrwl.vercel.app/tips)


meeroslav (approved):
A couple of minor improvements but otherwise LGTM:
General:
- When card is opaque, try to make full date or just day bolded
- When card is shown, whap date and heading - heading first, then date underneath to emphasize the title even more and detach it from the video/thumbnail
Dark mode:
- Make text full white to increase contrast
- Container background should also be darker blue - you can try to swap it with cards background color - have container be dark blue and cards ligher blue
- Icons for January 15th still need to be modified, but we can do that later

PR #20656, opened 2023-12-08
feat(nuxt): make nuxt public

mandarini:
Switch nuxt to public


jaysoo (approved):

isaacplmann (approved):

cheindl:
What about making actually also the npm package public? https://www.npmjs.com/package/@nx/nuxt seems to be still a private package.

PR #20666, opened 2023-12-08
chore(node): decrease e2e times

FrozenPandaz:
## Current Behavior

https://staging.nx.app/runs/qLbKKji3E7/task/e2e-node%3Ae2e

When we create new workspaces in e2e tests, every single package is installed.

## Expected Behavior

I made the Node e2e tests go from 30 minutes to 8 minutes :smile:

https://staging.nx.app/runs/c2BBFGVuvq/task/e2e-node%3Ae2e

I added a new `packages` option for the `newProject` function to allow for you to specify exactly which Nx plugins you want to install.

```ts
newProject({
  packages: ['@nx/node']
});
```

Currently, we install every single @nx package in a new workspace before tests run. This takes a long time and the tests probably only need a few of the packages.

This along with making sure not too many workspaces are created during the tests should make the e2e tests much faster.

There will be a warning in the e2e logs if the packages option is not passed which means all packages are installed. Please pass the packages to get rid of the warnings.

### Bonus

You can set `NX_E2E_EDITOR` to open the workspace in the editor of your choice.

Example:

`NX_E2E_EDITOR=code nx e2e e2e-cypress`

Be careful if too many workspaces are created.. a lot of editors will open... but you should probably try to refactor the tests so not as many workspaces are opened.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/node/src/node-esbuild.test.ts
+++ b/e2e/node/src/node-esbuild.test.ts
@@ -6,20 +6,21 @@ import {
   promisifiedTreeKill,
   readFile,
   runCLI,
-  runCLIAsync,
   runCommandUntil,
   setMaxWorkers,
-  tmpProjPath,
   uniq,
   updateFile,
 } from '@nx/e2e/utils';
-import { execSync } from 'child_process';
 import { join } from 'path';
 
 describe('Node Applications + esbuild', () => {
-  beforeEach(() => newProject());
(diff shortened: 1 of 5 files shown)


ndcunningham (approved):
LGTM

AgentEnder (approved):

PR #20698, opened 2023-12-11
docs(misc): adjust Nx version number in pkg json mentions

juristr:
we should probably have a simple mechanism to auto-update these mentions 🤔


isaacplmann (approved):
I think it is better to not have it automated so a person has to actually look at the docs and make sure they're still valid

PR #20702, opened 2023-12-11
fix(webpack): add standardWebpackConfigFunction option when users opts for a standard config function

jaysoo:
This PR fixes an issue when we added support for standard [webpack function as config](webpack.js.org/configuration/configuration-types/#exporting-a-function).

Any existing projects that doesn't call `composePlugins` on the top-level export will now be treated as a standard function, which will fail the build.

The fix is to add a new option `standardWebpackConfigFunction`, so users have to opt into this setting. There should not be an effect on PCv3 since those projects do not run through our executor anyway, this new option is for the specific case of PCv3 + `@nx/webpack:webpack`.

## Related Issue(s)

Fixes #


mandarini (approved):

ndcunningham (approved):

barrymichaeldoyle:
This should be in the release notes somewhere :) I spent the day trying to find out why my project was failing after updating. Turns out one of my `webpack,config.prod.ts` files wasn't using `composePlugins`. It's been working all year and since update it started failing.

PR #20721, opened 2023-12-12
chore(angular): split module-federation tests into their own project

Coly010:
## Current Behavior

Module Federation tests for angular take a significant amount of time on their own.
Having them as part of a project with other test suites increases the overall test time.

## Expected Behavior

Split the Module Federation tests into their own project to allow it to be run in parallel

Doing so shows a ~50% decrease in time to run angular e2es.
**Before**
image

**After**
image

## Related Issue(s)

Fixes #


jaysoo (approved):

leosvelperez (changes requested):
The new e2e project must be added to the nightly CI pipelines:

- `.github/workflows/e2e-matrix.yml`
- `.github/workflows/e2e-windows.yml`

.github/workflows/e2e-matrix.yml
```diff
@@ -214,6 +217,8 @@ jobs:
             project: e2e-angular-core
           - node_version: 16
             project: e2e-angular-extensions
+          - node_version: 16
+            project: e2e-angular-module-federation
```
leosvelperez, line 221: It should also be added for `node_version: 20`

.github/workflows/e2e-windows.yml
```diff
@@ -159,6 +162,8 @@ jobs:
             project: e2e-angular-core
           - node_version: 16
             project: e2e-angular-extensions
+          - node_version: 16
+            project: e2e-angular-module-federation
```
leosvelperez, line 166: It should also be added for `node_version: 20`

leosvelperez (changes requested):

.github/workflows/e2e-matrix.yml
Coly010: it is? right? 
image

.github/workflows/e2e-windows.yml
Coly010: it is? right? 
image

.github/workflows/e2e-matrix.yml
leosvelperez: That's correctly added to the matrix, but we exclude the other angular e2e projects from running with node 16 and 20. The new one need to match that.

leosvelperez (approved):

AgentEnder (approved):

PR #20772, opened 2023-12-14
feat(core): add gradle plugin

xiongemi:
## Related Issue(s)

Fixes #


packages/gradle/src/plugin/nodes.ts
```diff
@@ -0,0 +1,164 @@
...
+      >;
+
+      const targets =
+        targetsCache[hash] ??
+        createGradleTargets(tasks, projectRoot, options, context, outputDirs);
+      calculatedTargets[hash] = targets;
+
+      return {
+        projects: {
+          [projectName]: {
```
khalilou88, line 111: I think this should be` [projectRoot]: {....}`?

xiongemi: changed

PR #20778, opened 2023-12-14
fix(nextjs): enhance page generator to work when --project is not supplied

ndcunningham:


jaysoo (approved):

packages/next/src/generators/page/page.ts
```diff
@@ -60,11 +61,26 @@ async function normalizeOptions(host: Tree, options: Schema) {
...
+        artifactType: 'page',
+        callingGenerator: '@nx/next:page',
+        name: options.name,
+        directory: options.directory,
+      });
+
+    const project = readProjectConfiguration(host, determinedProjectName);
+
+    // app/ is a reserved folder in nextjs so it is safe to check it's existence
+    isAppRouter = host.exists(`${project.root}/app`);
```
jaysoo, line 81: I think this could also be `src/app` since Next.js supports both.

ndcunningham: Sure, I'll check both paths.

PR #20782, opened 2023-12-15
fix(linter): fix workspace-rule naming with flat config

### LINKED ISSUE #19754: Nx eslint workspace-rule dosn't work with eslint flat config
### Current Behavior

Currently, none of the workspace rules are detected in eslint flat config. I've tried to setup from scratch and generate the workspace rule after the flat config, and before the migration, none works.

### Expected Behavior

The workspace rule is detected and run

### GitHub Repo

https://github.com/beaussan/repro-nx-workspace-rule-with-flat-config

### Steps to Reproduce

- create a workspace, typescript, node, single server, vanilla. Any other will work since it's an issue with the `@nx/eslint` package.
- `nx run-many -t lint --all` => ✅
- `nx generate @nx/eslint:convert-to-flat-config` => ✅
- `nx run-many -t lint --all` => ✅
- `nx generate @nx/eslint:workspace-rule --name "demoRule"` => ✅
- `nx run-many -t lint --all` => ❌

Fails with `TypeError: Key "rules": Key "@nx/workspace/demoRule": Could not find plugin "@nx/workspace".`.

### Nx Report

```shell

Node   : 18.18.1
OS     : linux-x64
yarn   : 1.22.18

nx                 : 17.0.0
@nx/js             : 17.0.0
@nx/jest           : 17.0.0
@nx/linter         : 17.0.0
@nx/eslint         : 17.0.0
@nx/workspace      : 17.0.0
@nx/devkit         : 17.0.0
@nx/esbuild        : 17.0.0
@nx/eslint-plugin  : 17.0.0
@nx/node           : 17.0.0
@nrwl/tao          : 17.0.0
typescript         : 5.1.6
```

### Failure Logs

```shell
TypeError: Key "rules": Key "@nx/workspace/demoRule": Could not find plugin "@nx/workspace".
at throwRuleNotFoundError (/workdir/repro-nx-workspace-rule-with-flat-config/node_modules/eslint/lib/config/rule-validator.js:65:11)
at RuleValidator.validate (/workdir/repro-nx-workspace-rule-with-flat-config/node_modules/eslint/lib/config/rule-validator.js:127:17)
at FlatConfigArray.[finalizeConfig] (/workdir/repro-nx-workspace-rule-with-flat-config/node_modules/eslint/lib/config/flat-config-array.js:231:23)
at FlatConfigArray.getConfig (/workdir/repro-nx-workspace-rule-with-flat-config/node_modules/@humanwhocodes/config-array/api.js:969:55)
at FlatConfigArray.isFileIgnored (/workdir/repro-nx-workspace-rule-with-flat-config/node_modules/@humanwhocodes/config-array/api.js:993:15)
at /workdir/repro-nx-workspace-rule-with-flat-config/node_modules/eslint/lib/eslint/eslint-helpers.js:312:49
at Array.reduce (<anonymous>)
at entryFilter (/workdir/repro-nx-workspace-rule-with-flat-config/node_modules/eslint/lib/eslint/eslint-helpers.js:299:28)
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I also tried with a non flat config, and run the generator, and the result is the same: workspace rules don't work with flat config

meeroslav:
## Related Issue(s)

Fixes #19754


packages/eslint-plugin/src/resolve-workspace-rules.ts
```diff
@@ -25,7 +25,7 @@ export const workspaceRules = ((): ESLintRules => {
     // Apply the namespace to the resolved rules
     const namespacedRules: ESLintRules = {};
     for (const [ruleName, ruleConfig] of Object.entries(rules as ESLintRules)) {
-      namespacedRules[`${WORKSPACE_RULE_NAMESPACE}/${ruleName}`] = ruleConfig;
+      namespacedRules[`${WORKSPACE_RULE_PREFIX}-${ruleName}`] = ruleConfig;
```
JamesHenry, line 28: Let's add it under both for now? That would make this change a lot less disruptive overall and I don't think cause any issues for either the majority of users using `.eslintrc.json`, nor the minority using flat config?

JamesHenry (changes requested):

packages/eslint-plugin/src/resolve-workspace-rules.ts
meeroslav: We do have migration in place so I don't see this being any more disruptive than other code mods we do.

The failure in flat-config is correct - @nx/workspace/my-rule suggests there is a plugin called @nx/eslint-plugin-workspace with a rule named my-rule and config verification rightfully fails.

So far this is only happening for fat config but future consolidations could break this for standard config in between two patches. I would rather migrate to safe config now than once it breaks.

We anyway need to migrate existing flat config repos so why not do it for all of them?

JamesHenry: I didn't say not to do the migration, I only said to register it under both namespaces. I think we should do the migration but is there any downside to keeping the old namespace for now for folks who don't/can't run the migration straight away?

JamesHenry (approved):

PR #20786, opened 2023-12-15
fix(module-federation): support buildable libs

### LINKED ISSUE #20491: When compiling the library bundle,when use mf the singleton pattern for the service does not take effect.
### Current Behavior

When compiling the library bundle, the singleton mode of the service does not work. When not compiling, the result is normal. I am not sure if it is an angular problem or an nx problem.

### Expected Behavior

result normal

### GitHub Repo

https://github.com/EnochGao/nx-test

### Steps to Reproduce

1. npm install
2. nx serve child
3. nx serve host
4. visit localhost://4200 

### Nx Report

```shell
Node   : 20.9.0
   OS     : linux-x64
   npm    : 10.1.0
   
   nx                 : 17.2.0-beta.10
   @nx/js             : 17.2.0-beta.10
   @nx/jest           : 17.2.0-beta.10
   @nx/linter         : 17.2.0-beta.10
   @nx/eslint         : 17.2.0-beta.10
   @nx/workspace      : 17.2.0-beta.10
   @nx/angular        : 17.2.0-beta.10
   @nx/cypress        : 17.2.0-beta.10
   @nx/devkit         : 17.2.0-beta.10
   @nx/eslint-plugin  : 17.2.0-beta.10
   @nrwl/tao          : 17.2.0-beta.10
   @nx/web            : 17.2.0-beta.10
   @nx/webpack        : 17.2.0-beta.10
   typescript         : 5.2.2
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Coly010:
## Current Behavior

Buildable libs are not consumed from dist when using MF when creating shared deps

## Expected Behavior

Buildable libs should be consumed from dist when creating shared deps

## Related Issue(s)

Fixes #20491

### DIFF
--- a/packages/webpack/src/utils/module-federation/dependencies.ts
+++ b/packages/webpack/src/utils/module-federation/dependencies.ts
@@ -1,6 +1,11 @@
 import type { ProjectGraph } from '@nx/devkit';
 import type { WorkspaceLibrary } from './models';
 import { readTsPathMappings } from './typescript';
+import {
+  getOutputsForTargetAndConfiguration,
+  parseTargetString,
+} from '@nx/devkit';
+import { interpolate } from 'nx/src/tasks-runner/utils';
 
 export function getDependentPackagesForProject(
   projectGraph: ProjectGraph,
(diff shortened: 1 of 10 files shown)


isaacplmann (approved):

packages/webpack/src/utils/module-federation/dependencies.ts
```diff
@@ -57,12 +62,32 @@ function getLibraryImportPath(
   library: string,
   projectGraph: ProjectGraph
 ): string | undefined {
+  let buildLibsFromSource = true;
+  if (process.env.NX_BUILD_LIBS_FROM_SOURCE) {
```
jaysoo, line 66: We can find this info from `NX_TASK_TARGET_PROJECT` and `NX_TASK_TARGET_TARGET`? Do we need the extra env var from executor?

Users won't be able to override with CLI options like `--buildLibsFromSource=false` but I think that's fine.

Coly010: I used a separate one because of cases where a new process isn't spawned, those env vars match the parent process (the host) and not the child (the remote).

jaysoo: Hmm, interesting. Maybe something to clean-up later if we can move `buildLibsFromSource` to a normal webpack plugin then we wouldn't need all the env vars to be passed like this.

jaysoo (approved):

PR #20794, opened 2023-12-15
chore(nx-dev): update nx-dev banner

bcabanes:


isaacplmann (changes requested):
The added s has a space on the left

isaacplmann (approved):

PR #20798, opened 2023-12-15
fix(release): changelog renderer should prefer breaking change explanation text

JamesHenry:
## Current Behavior

We always repeat the breaking change commit message and do not recognize breaking changes without the `!` being present.

## Expected Behavior

We recognize breaking changes with only `BREAKING CHANGE:` present (in accordance with the spec https://www.conventionalcommits.org/en/v1.0.0/), and we preferentially render the explanation text after `BREAKING CHANGE:` in the relevant section of the changelog. 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/changelog-renderer/index.ts
+++ b/packages/nx/changelog-renderer/index.ts
@@ -143,7 +143,16 @@ const defaultChangelogRenderer: ChangelogRenderer = async ({
           const line = formatCommit(commit, repoSlug);
           markdownLines.push(line);
           if (commit.isBreaking) {
-            breakingChanges.push(line);
+            const breakingChangeExplanation = extractBreakingChangeExplanation(
+              commit.body
+            );
+            breakingChanges.push(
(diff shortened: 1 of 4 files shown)


packages/nx/changelog-renderer/index.ts
```diff
@@ -294,11 +312,37 @@ function groupBy(items: any[], key: string) {
 function formatCommit(commit: GitCommit, repoSlug?: RepoSlug): string {
   let commitLine =
     '- ' +
-    (commit.scope ? `**${commit.scope.trim()}:** ` : '') +
     (commit.isBreaking ? '⚠️  ' : '') +
+    (commit.scope ? `**${commit.scope.trim()}:** ` : '') +
```
JamesHenry, line 316: Moved these around it looks better when they line up.

Real example of the "after" of this PR from rxjs repo:

image

AgentEnder (approved):

PR #20809, opened 2023-12-17
fix(core): normalize version prefix to avoid "v" twice

guiseek:
Since the new version has the v prefix, it was duplicated in the message when a workspace is being created
Like this: " >  NX   Creating your vv17.2.5 workspace."

## Related Issue(s)

Fixes #


JamesHenry:
Sorry for the confusion here @guiseek. The publish of v17.2.5 was accidental and it means that that version has made its way into the package.json files themselves.

Therefore this PR isn't really the "fix" for that, it shouldn't have been possible in the first place.

This PR https://github.com/nrwl/nx/pull/20815 explains how it was possible and is the fix for the issue. We will cut 17.2.6 very soon with this fix in to ensure it doesn't happen again.

PR #20872, opened 2023-12-20
feat(core): use Nx plugins and inferred targets when running "nx init"

jaysoo:
The goal is to change `nx init` so it:

1. Automatically detects Nx plugins to install
2. Configures Nx plugins so projects will have inferred targets 
3. Makes minimal changes to the existing project. Only updates to `package.json`, lockfile, and `nx.json` changes. No `project.json`, `tsconfig.base.json`, etc.

Screenshot 2024-01-09 at 3 52 38 PM

Note: These changes are hidden behind `NX_PCV3` env var for now.

## Changes

- Fork logic in `nx init` so when `NX_PCV3=true` then we will use the new logic, otherwise it's business as usual.
- For Angular CLI projects (e.g. `angular.json`) we run the existing logic, since `@nx/angular` doesn't support PCv3 yet.
- Expose `globSync` as a function that is callable from JS side, and remove dependency on `glob` package.
- ~~Fix `@nx/react:init` and `@nx/next:init` generators to remove default values for `e2eTestRunner` and `unitTestRunner`. When none are passed, Cypress and Jest will not be initialized.~~
- ~~Expose `@nx/eslint:init` generator so we can call it from `nx init` (and eventually `nx add`).~~
- ~~Update `@nx/next/plugin` PCv3 plugin to respect existing package scripts when inferring targets. This behavior can be controlled via `usePackageScripts` option for the plugin (default to `true`).~~
- ~~Fix `@nx/next/plugin` to use correct output path for root projects. e.g. `{projectRoot}/.next` not `{workspaceRoot}/./.next` (which does not work)~~

Edit: The changes unrelated to `nx init` will be in other PRs.

### Use `--useDotNxInstallation` flag to install to `.nx` folder

We no longer prompt for which folder to install Nx to, and just always install `.nx` to the current directory. Previously, if users choose to create a nested folder we would run `npx create-nx-workspace <new-folder>`, but this doesn't make sense if `nx init` it meant to add Nx to existing projects.

https://app.warp.dev/block/sb5dhH9o4mgKtLrj8s3DP8

### Angular CLI project (standalone)

This is unchanged from current behaviour.

https://app.warp.dev/block/JDtqLLkxsRW2XWfeYA4bXm

### Angular CLI project (integrated)

This is unchanged from current behaviour.

https://app.warp.dev/block/7HjIsOamgd8OaBzzHNonZw

### All other cases (not Angular, not `.nx`)

The `nx init` command currently errors out for projects at the root, which will be fixed soon in another PR.
https://app.warp.dev/block/01WeQ9xP98jDijRJKlftNL


packages/nx/src/command-line/init/implementation/add-nx-to-npm-repo.ts
```diff
@@ -33,7 +33,7 @@ export async function addNxToNpmRepo(options: Options) {
   if (options.interactive && scripts.length > 0) {
     output.log({
       title:
-        '🧑‍🔧 Please answer the following questions about the scripts found in your package.json in order to generate task runner configuration',
+        '🧑‍🔧 Please answer the following questions about the scripts found in your package.json in order tozRenerate task runner configuration',
```
mandarini, line 36: Remove this change?

packages/eslint/src/plugins/plugin.ts
```diff
@@ -31,7 +32,11 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
       return {};
     }
 
-    return {
+    if (projectRoot === '.') {
+    } else {
+    }
```
mandarini, line 37: is this a work-in-progress, or just it was accidentally left there?

packages/next/src/plugins/plugin.ts
```diff
@@ -4,22 +4,24 @@ import {
...
 
 import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
-import { type NextConfig } from 'next';
 import { PHASE_PRODUCTION_BUILD } from 'next/constants';
 import { getLockFileName } from '@nx/js';
 
 export interface NextPluginOptions {
+  // TODO(jack): the design of this option will need to be revisited, but we do want to respect the user's package scripts.
+  usePackageScripts?: boolean;
```
mandarini, line 24: should this be added in all plugins?

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,184 @@
...
+} from './implementation/utils';
+import { prompt } from 'enquirer';
+import { execSync } from 'child_process';
+
+export interface InitArgs {
+  addE2e: boolean;
+  force: boolean;
+  integrated: boolean;
+  interactive: boolean;
+  vite: boolean;
```
mandarini, line 26: why so specific?

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+    generateDotNxSetup(version);
+    // invokes the wrapper, thus invoking the initial installation process
+    runNxSync('');
+  }
+}
+
+async function detectPlugins(options: InitArgs): Promise<undefined | string[]> {
+  const files = ['package.json'].concat(
+    sync('{apps,packages,libs}/**/*/package.json')
```
mandarini, line 100: why not just a general glob under root dir? excluding `node_modules`? i mean to say, someone may have any type of setup (any dir name), no?

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,176 @@
...
+  runInstall,
+  updateGitIgnore,
+} from './implementation/utils';
+import { prompt } from 'enquirer';
+import { execSync } from 'child_process';
+
+export interface InitArgs {
+  addE2e: boolean;
+  force: boolean;
+  integrated: boolean;
```
mandarini: these are not used, is there a purpose they are there?

packages/nx/src/command-line/init/init-v2.ts
mandarini: also, it's not used, is there a purpose it's there?

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,176 @@
...
+        'Setting Nx up installation in `.nx`. You can run nx commands like: `./nx.bat --help`'
+      );
+    }
+    generateDotNxSetup(version);
+    // invokes the wrapper, thus invoking the initial installation process
+    runNxSync('');
+  }
+}
+
+async function detectPlugins(options: InitArgs): Promise<undefined | string[]> {
```
mandarini: `options` is not used, is there a reason it's there?

packages/next/src/plugins/plugin.ts
mandarini: i added it in all plugins as an option for the time being

packages/eslint/src/plugins/plugin.ts
```diff
@@ -43,6 +48,15 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
...
       },
     };
+    // For root projects, the name is not inferred from root package.json, so we need to manually set it.
+    // TODO(jack): We should handle this in core and remove this workaround.
+    if (projectRoot === '.') {
+      result.projects[projectRoot]['name'] = buildProjectName(
+        projectRoot,
+        context.workspaceRoot
+      );
+    }
```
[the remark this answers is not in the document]
mandarini: it should be fixed in core

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,171 @@
...
+      detectedPlugins.add('@nx/jest');
+    }
+    if (deps['cypress']) {
+      detectedPlugins.add('@nx/cypress');
+    }
+    if (deps['playwright']) {
+      detectedPlugins.add('@nx/playwright');
+    }
+    // Frameworks
+    if (deps['@nestjs/core'] && deps['@nestjs/cli']) {
```
leosvelperez, line 138: Do we have a PCV3 `createNodes` plugin for Nest?

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,171 @@
...
+import { prompt } from 'enquirer';
+import { execSync } from 'child_process';
+
+export interface InitArgs {
+  interactive: boolean;
+  nxCloud?: boolean;
+  useDotNxInstallation?: boolean;
+}
+
+export async function initHandler(options: InitArgs) {
```
leosvelperez, line 27: If this is run in an Angular CLI workspace, it wouldn't work because we don't have the plugin yet. This might also be the case with other plugins.

Should this handle that scenario and error? Or reroute it to the old handler?

packages/nx/src/command-line/init/init-v2.ts
jaysoo: Won't be ready, I'll remove for now.

mandarini (approved):

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,171 @@
...
+    generateDotNxSetup(version);
+    // invokes the wrapper, thus invoking the initial installation process
+    runNxSync('');
+  }
+}
+
+async function detectPlugins(): Promise<undefined | string[]> {
+  const files = ['package.json'].concat(
+    sync(`**/*/package.json`, {
+      ignore: ['**/node_modules/**', '**/build/**', '**/dist/**', '**/tmp/**'],
```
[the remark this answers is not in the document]
jaysoo: Spoke to Jon and we'll switch to the Rust version of glob that should handle the `.gitignore` file already.

packages/nx/src/command-line/init/init-v2.ts
jaysoo: Solved by using the Rust glob util instead, which will respect `.gitignore` files.

packages/nx/src/command-line/init/init-v2.ts
jaysoo: Added the custom Angular logic back for now.

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,176 @@
...
+    updateGitIgnore(repoRoot);
+
+    // TODO(jack): Remove this Angular logic once `@nx/plugin` is compatible with PCv3.
+    let plugins: string[];
+    if (existsSync('angular.json')) {
+      plugins = [];
+      await addNxToAngularCliRepo({
+        ...options,
+        integrated: !!options.integrated,
+      });
```
leosvelperez, line 51: We should do this as early as possible, in line 36, and we also need to exit after this. Otherwise, we'll duplicate some operations. The `addNxToAngularCliRepo` function already includes all the steps performed here unrelated to the new logic adding plugins (e.g. `createNxJsonFile`, `updateGitIgnore`, setting up Nx Cloud, etc.).

mandarini (approved):

packages/nx/src/command-line/init/init-v2.ts
jaysoo: Moved the `.nx` and Angular logic to the top with early return.

leosvelperez (approved):

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -0,0 +1,175 @@
...
+    } catch {
+      // Could have malformed JSON for unit tests, etc.
+      continue;
+    }
+
+    const deps = {
+      ...packageJson.dependencies,
+      ...packageJson.devDependencies,
+    };
+
```
mandarini, line 125:
```suggestion
    if (deps['storybook']) {
      detectedPlugins.add('@nx/storybook');
    }
```

add Storybook too!

jaysoo: done

FrozenPandaz (changes requested):
Looks good. Some minor changes then I think it's good to go!

packages/nx/src/native/walker.rs
```diff
@@ -122,6 +122,15 @@ where
     receiver_thread.join().unwrap()
 }
 
+#[napi]
+fn glob_sync(globs: Vec<String>, root: String) -> Vec<String> {
```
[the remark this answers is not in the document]
jaysoo: Oh interesting, didn't know about it.

FrozenPandaz (approved):

PR #20878, opened 2023-12-21
fix(vite): dist and coverage paths for root projects

mandarini:
Account for root projects when setting build and coverage output paths.

### DIFF
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -221,13 +221,17 @@ function getOutputs(
 } {
   const { build, test } = viteConfig;
 
-  const buildOutputPath =
-    normalizeOutputPath(build?.outDir, projectRoot) ??
-    '{workspaceRoot}/dist/{projectRoot}';
+  const buildOutputPath = normalizeOutputPath(
+    build?.outDir,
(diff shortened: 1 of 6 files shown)


packages/vite/src/plugins/plugin.ts
```diff
@@ -237,16 +241,28 @@ function getOutputs(
...
-    if (outputPath.startsWith('..')) {
-      return join('{workspaceRoot}', join(projectRoot, outputPath));
+    if (isAbsolute(outputPath)) {
+      return `{workspaceRoot}/${relative(workspaceRoot, outputPath)}`;
     } else {
-      return outputPath;
+      if (outputPath.startsWith('..')) {
+        return join('{workspaceRoot}', join(projectRoot, outputPath));
+      } else {
+        if (projectRoot === '.' && outputPath.startsWith('./')) {
```
jaysoo, line 260: I think this if-else can just be:

```
join('{projectRoot}', outputPath.slice(2));
```

since joining `./foo` or `foo` gives the same result.

mandarini: ohhhhhhhh 😅

jaysoo (approved):

PR #20888, opened 2023-12-21
docs(core): fix columns

isaacplmann:
Fix columns on several pages


nx-dev/ui-markdoc/src/lib/tags/cards.component.tsx
bcabanes, line 1: Can you remove the `const calcCols` function? l68.

bcabanes (approved):

PR #20896, opened 2023-12-22
feat(react-native): upgrade react native to 0.73

xiongemi:
## Current Behavior

- This PR upgrade react-native to 0.73.1
- [x] need to wait for the expo to publish 50
- [x] also need to wait for @expo/webpack-config to support expo 50

## Expected Behavior

- react native 0.73 is going to support symlink as default, no need to manually symlink the node_modules folder
- now for the react native app, developers can run generator @nx/react-native:web-configuration to create a web app version of the native app using either webpack or vite as bundler
- now change the storybook to use the default react storybook configuration, so the storybook would be viewed on the web, with no more version conflict
- also, since web configuration is supported, now developers can also use cypress or playwright to e2e tests their apps

## Related Issue(s)

Fixes #


AgentEnder (approved):
Files I own look good 👍

Rhuansantos (approved):

mandarini (approved):

packages/expo/src/executors/build/build.impl.ts
```diff
@@ -1,6 +1,9 @@
 import { ExecutorContext, names } from '@nx/devkit';
 import { resolve as pathResolve } from 'path';
 import { ChildProcess, fork } from 'child_process';
+import { removeSync } from 'fs-extra';
```
ndcunningham, line 4: Is this used you imported it but I don't see where it is used.

ndcunningham (approved):

packages/expo/src/executors/build/build.impl.ts
xiongemi: removed

isaacplmann (approved):

PR #20975, opened 2024-01-02
fix(core): prioritize nxignore for watcher updates

### LINKED ISSUE #20942: Target cache calculation does not respect leading-slash rules in nested .gitignores on subsequent runs
### Current Behavior

With certain .gitignore structures (see Steps to Reproduce), cacheable commands will rerun if ignored files change while the daemon is running.

### Expected Behavior

If a file is ignored, it should never be considered for the cache.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Take any nx project, new or existing
2. Add an empty .gitignore file to the root of the project
3. Add a new .gitignore file anywhere inside one of the nx targets, and add an ignore rule with a leading slash
4. Run a cacheable nx command (with the daemon enabled) against the target with the .gitignore
5. Change a file ignored by the rule from step 3
6. Run the cacheable nx command again
7. The command will execute again, even though it should be a cache hit

### Nx Report

```shell

   Node   : 18.18.2
   OS     : darwin-x64
   npm    : 9.8.1
   
   nx             : 17.2.7
   lerna          : 6.6.2
   @nx/workspace  : 17.2.7
   @nx/devkit     : 17.2.7
   @nrwl/tao      : 17.2.7
   typescript     : 5.2.2
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

This bug is due to an underlying bug in `ignore-files`, which was reported here: https://github.com/watchexec/watchexec/pull/745, and will be fixed in https://github.com/watchexec/watchexec/pull/746.

### LINKED ISSUE #20945: .nxignore is not used in any non-root directory with a .gitignore
### Current Behavior

If you have a top-level .nxignore file it will not apply to any directory which contains a .gitignore file.

### Expected Behavior

The .nxignore file should apply globally, and should override all .gitignore files, regardless of file structure.

### GitHub Repo

https://github.com/thislooksfun/nx-caching-ignore-bugs

### Steps to Reproduce

1. Take any nx project, new or existing
2. Add a .nxignore file to the root of the project with some rule(s) in it
3. Add an empty .gitignore file anywhere inside one of the nx targets
4. Run a cacheable nx command (with the daemon enabled) against the target with the .gitignore
5. Change a file ignored by the rule from step 2
6. Run the cacheable nx command again
7. The command will execute again, even though it should be a cache hit

### Nx Report

```shell

   Node   : 18.18.2
   OS     : darwin-x64
   npm    : 9.8.1
   
   nx             : 17.2.7
   lerna          : 6.6.2
   @nx/workspace  : 17.2.7
   @nx/devkit     : 17.2.7
   @nrwl/tao      : 17.2.7
   typescript     : 5.2.2
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

This is caused by a couple problems coming together due to the way the nxignore is added to the ignore filter:

https://github.com/nrwl/nx/blob/8a8dbb381ea6ce50d3a5e8ba605bde56dbb22dc2/packages/nx/src/native/watch/watch_config.rs#L37-L43

#### Problem 1:

Due to a bug in the way the `ignore-files` crate currently works internally, calling `add_file` to add more ignore rules to a given path does not apply those rules to any of its children. This means that any rules added to the root of the project by the `.nxignore` file do not propagate down into any subdirectories containing their own .gitignore files. This issue will be fixed by https://github.com/watchexec/watchexec/pull/746.

#### Problem 2:

`ignore-files` assumes that all given ignore rules are hierarchical, with the closest (i.e. deepest) rules taking precedence. This means that even after https://github.com/watchexec/watchexec/pull/746 fixes problem 1 above, the .nxignore file will still not take precedence over the .gitignore tree. In order to make that work as expected, I think fixing that will require implementing some custom priority logic in the watch filterer.

Specifically, I suggest having a separate Filter instance for the `.nxignore` that is checked first. If matching against it returns `Match::None`, then it should fall back to the hierarchical `.gitignore` matcher.

Cammisuli:
## Current Behavior

When there is a gitignore file in a sub-folder, the nxignore rules do not apply correctly.

## Expected Behavior
if an .nxignore file exists, it will be used as a priority over gitignores. 

This means that if a file is allowed or blocked in the nxignore file, we do not check gitignore. If there is no match with the nxignore file, only then do we check the gitignores.

## Related Issue(s)

Fixes #20945
Fixes #20942


packages/nx/Cargo.toml
```diff
@@ -12,7 +12,7 @@ fs_extra = "1.3.0"
 globset = "0.4.10"
 hashbrown = { version = "0.14.3", features = ["rayon", "rkyv"] }
 ignore = '0.4'
-ignore-files = "1.3.0"
+ignore-files = "2.0.0"
```
thislooksfun, line 15: This should also fix https://github.com/nrwl/nx/issues/20942.

packages/nx/src/native/watch/watch_filterer.rs
```diff
@@ -19,7 +21,15 @@ impl Filterer for WatchFilterer {
...
         let event = transformed.as_ref().unwrap_or(watch_event);
 
-        if !self.inner.check_event(event, priority)? {
+        if !self
+            .nx_ignore
+            .as_ref()
+            .is_some_and(|nx_ignore| nx_ignore.check_event(event, priority).unwrap_or(false))
+        {
+            return Ok(false);
+        }
```
thislooksfun, line 30: Unfortunately, using `check_event` directly doesn't fully solve #20945. Take the following config:

```gitignore
# file: .gitignore
node_modules/

# file: inner/.gitignore
.env.local

# file: .nxignore
!.env.*
```

This config tells git to ignore the file(s) `inner/**/.env.local`, so they don't get committed, but we still want nx to re-run if it changes, since build tooling and output might differ if it relies on the environment, notably (though not exclusively) if a bundler embeds any environment variables into the output. This setup fails with this implementation, as the call to `nx_ignore.check_event` will return `true`, which will cause it to fall back on the gitignore filter.

Unfortunately in order to solve this issue we can't call `check_event` directly, as it's only a boolean state. We need to get a trinary state from the matching so we know whether it matched at all (allow/deny) or had no match (fall back on .gitignore). This will either require querying the underling `Filter` object directly, without wrapping it in a `Filterer`, or will require updating `ignore_files` again to add a helper API to `Filterer` to allow losslessly querying the underlying `Filter`.

thislooksfun (changes requested):

thislooksfun (approved):
Looks solid to me! 🎉

FrozenPandaz (approved):

PR #20988, opened 2024-01-03
fix(webpack): correctly handle data URIs with escaped quotes in style…

andreasmoller25:
This is just a copy of the fix already done in the angular-cli repository: https://github.com/angular/angular-cli/pull/23691

## Current Behavior

Double quoted url() in css arguments are parsed incorrectly
## Expected Behavior

Double quoted url() in css arguments are parsed correctly
## Related Issue(s)

No issue in repo, fixes same copied code as described in angular-cli repository https://github.com/angular/angular-cli/issues/23680


andreasmoller25:
@jaysoo Would be really grateful if this fix could be included in the next nx-release.

Best regards:
Andreas

mandarini (approved):
I think this change makes sense. Thanks

PR #20999, opened 2024-01-04
fix(release): do not set extra v on GitHub release, improve GH API error handling

### LINKED ISSUE #20993: `nx release changelog` should not have hardcoded `v` prefix for `tag` and `title` query parameters when submitting github releases via web form.
### Current Behavior

I have a monorepository, and I am using the programmatic releases API to handle the process of versioning and changelog creation, and for that I have a custom script, the exact implementation is [available here](https://github.com/virtuallyunknown/nx-changelog-issue/blob/master/release.ts#L41-L60), and it's similar to the example provided in [the documentation](https://nx.dev/core-features/manage-releases#using-the-programmatic-api-for-nx-release).

First we run `releaseVersion` to bump all versions, stage and commit all changes and creates a version tag, which uses `"releaseTagPattern": "{version}"`. This is configured in a separate `nx.json` file [here](https://github.com/virtuallyunknown/nx-changelog-issue/blob/master/nx.json#L13), because I couldn't find a way to do it via the API directly.

Next, we run `releaseChangelog` who's job is to create a GitHub release. Behind the scenes, what this function attempts to do is detect if we have `GH_TOKEN` or `GITHUB_TOKEN` set, and if we do then it will submit the GitHub release via the GitHub API. If a token is not available however, it will fall back to a web form which opens in the user's default browser. The URL of that form is pre-populated with query parameters with information from our release:

`https://github.com/${config.repo}/releases/new?tag=v${release.version}&title=v${release.version}&body=...}`

https://github.com/nrwl/nx/blob/e31c179c02f9aa2a767ed75be32be6bf53d69fda/packages/nx/src/command-line/release/utils/github.ts#L226-L233

The problem is that the query parameters for `tag` and `title` have a hardcoded `v` prefix. What that means is, when the web form opens, it will have a prefixed tag pre-selected, so if the users submits this form as is, it will create a brand new and separate tag different from our release tag, and it will submit the GitHub release under it.

So even if we were to use `"releaseTagPattern": "v{version}"` in our `nx.json` config, the web form will come with a `vvX.Y.Z` as a suggested tag, and if the user isn't mindful about this will create extra tags.

I have illustrated the issue better in the picture below.

changelog

### Expected Behavior

The web form which opens in the web browser should have pre-selected tag that matches the tag we create earlier with `releaseVersion`

### GitHub Repo

https://github.com/virtuallyunknown/nx-changelog-issue

### Steps to Reproduce

1. Clone the repository **via ssh**, cd into the directory and install all dependencies.

```bash
git clone git@github.com:virtuallyunknown/nx-changelog-issue.git
cd nx-changelog-issue
pnpm install
```

2. Make any type of modification and commit it.

```bash
touch my_new_feature.txt
git add my_new_feature.txt
git commit -m "feat: added a new feature"
```

3. Make sure there is no `GH_TOKEN` or `GITHUB_TOKEN` environment variable set, so that the `releaseChangelog` command falls back to submitting the github release via the webform instead of the API. Run the `release.ts` script.

```bash
pnpm exec tsx release.ts

# Pick the following options:
# Question #1: ✔ What kind of change is this for your packages? - select any version
# Question #2: ? Do you want to dry run these commands? (Y/n)   - select "n" (no)
#
# Because no github token is available, the API request will fail, and a webform
# with pre-populated options will open in the browser.
```

### Nx Report

```shell
Node   : 20.10.0
   OS     : linux-x64
   pnpm   : 8.13.1
   
   nx             : 17.2.8
   @nx/js         : 17.2.8
   @nx/workspace  : 17.2.8
   @nx/devkit     : 17.2.8
   @nrwl/tao      : 17.2.8
   typescript     : 5.3.3
```

### Failure Logs

_No response_

### Package Manager Version

8.13.1

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

JamesHenry:
Fixes #20993 

Improved GitHub API error handling and manual fallback DX:

image

### DIFF
--- a/packages/nx/src/command-line/release/utils/github.ts
+++ b/packages/nx/src/command-line/release/utils/github.ts
@@ -4,9 +4,11 @@
  */
 import type { AxiosRequestConfig } from 'axios';
 import * as chalk from 'chalk';
+import { prompt } from 'enquirer';
 import { execSync } from 'node:child_process';
 import { existsSync, promises as fsp } from 'node:fs';
 import { homedir } from 'node:os';
+import { output } from '../../../utils/output';
 import { joinPathFragments } from '../../../utils/path';
 import { Reference } from './git';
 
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/release/utils/github.ts
```diff
@@ -227,9 +301,9 @@ function githubNewReleaseURL(
   config: GithubRequestConfig,
   release: { version: string; body: string }
 ) {
-  return `https://github.com/${config.repo}/releases/new?tag=v${
+  return `https://github.com/${config.repo}/releases/new?tag=${
```
JamesHenry, line 304: These days release.version will already be formatted appropriately based on the user's defined `releaseTagPattern` (which may or may not include a `v`, it's up to them)

FrozenPandaz (approved):

PR #21019, opened 2024-01-05
fix(angular): support scheduling inferred angular cli builder targets

leosvelperez:
Update the Angular CLI adapter layer to support scheduling targets that are inferred. The adapter layer is updated so the host used by Angular CLI builders reads the project configurations from the project graph. This way, it can access the expanded targets.

The host used by Angular CLI schematics remains untouched and reads the project configurations from the `project.json` files. This way, Angular CLI schematics only process/modify what's in the actual configuration files.

This is needed to unblock https://github.com/nrwl/nx/pull/20562.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/adapter/ngcli-adapter.ts
+++ b/packages/nx/src/adapter/ngcli-adapter.ts
@@ -75,7 +75,7 @@ export async function createBuilderContext(
   context: ExecutorContext
 ) {
   require('./compat');
-  const fsHost = new NxScopedHost(context.root);
+  const fsHost = new NxScopedHostForBuilders(context.root);
   // the top level import is not patched because it is imported before the
   // patching happens so we require it here to use the patched version below
   const { workspaces } = require('@angular-devkit/core');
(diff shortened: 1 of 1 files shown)


packages/nx/src/adapter/ngcli-adapter.ts
```diff
@@ -613,6 +613,39 @@ export class NxScopedHost extends virtualFs.ScopedHost<any> {
...
+          workspaceConfig.projects[projectName] ??= {
+            ...graph.nodes[projectName].data,
+          };
+        }
+
+        return workspaceConfig;
+      }),
+      catchError((err) => {
+        console.log('Unable to read angular.json');
+        console.log(err);
```
Coly010, line 642: Should these be `.log` or `.error`?

It would only really result in color formatting in the terminal I believe, but it feels like we should keep it consistent.

leosvelperez: This code is consistent with what currently exists. This function overrides a parent function that currently logs the error in this way. We could definitely change it though. It makes more sense what you suggest. I'll update the other function as well.

Coly010 (approved):

Cammisuli (approved):

PR #21047, opened 2024-01-08
docs(core): combine features sections

isaacplmann:
Combines Core Features and Plugin Features into one Features section
- Rewrites Use Task Executors to [Executors and Configurations](https://nx-dev-git-fork-isaacplmann-docs-pcv3-features-nrwl.vercel.app/concepts/executors-and-configurations)
- Rewrites Use Code Generators to [Generate Code](https://nx-dev-git-fork-isaacplmann-docs-pcv3-features-nrwl.vercel.app/features/generate-code)


docs/shared/recipes/running-tasks/executors-and-configurations.md
```diff
@@ -1,18 +1,10 @@
-# Use Task Executors
+# Executors and Configurations
```
juristr, line 1: Why did we move this to a recipe? It is still kind of a concept page explaining at a higher level what an executor is and various aspects of it (config, running etc..).

docs/shared/features/enhance-tooling.md
```diff
@@ -0,0 +1,102 @@
+# Enhance Tooling
+
+Nx is able to enhance your existing tooling in a non-intrusive way so that, without modifying your tooling set-up, Nx can add:
+
+- [Fully configured caching](/features/cache-task-results)
+- [Task pipelines](/concepts/task-pipeline-configuration)
+- [Task distribution](/ci/features/distribute-task-execution)
+
+This works for any tool for which there exists an [official plugin](/plugin-registry).
```
juristr, line 9: Can we formulate this slightly different. This sounds like a "either use plugins or you don't get caching, task pipelines and task distribution", which is not entirely true.

It is the "automatic enhancing" that only works with plugins. I want it to be softer though and not be a "no-go" for folks that read the line and are like "oh I need plugins, forget it"

docs/shared/features/enhance-tooling.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Setup
+
+The following command will make Nx detect your tooling and set up the appropriate plugins:
+
+```shell
+nx init
+```
+
+This command will detect the presence of any tooling for which Nx has an [official plugin](/plugin-registry) and give you the option to install the plugin. The installed plugin will be added to your `package.json` file and an entry will be added to the `plugins` array in the `nx.json` file.
```
juristr, line 19: should we show a small excerpt of the `nx.json` here? Just so ppl get an idea?

docs/shared/features/enhance-tooling.md
```diff
@@ -0,0 +1,102 @@
+# Enhance Tooling
```
juristr, line 1: I'm missing more info on what gets enhanced and the benefit of it, like why would I want it. 

When I read this page I get some quick "you get caching, task pipelines,.." which as mentioned in the next comment is not exclusive to the plugins. I get how to set it up via `nx init`, how to use it and types of targets as well as how to modify options (where I'm not even sure I'd need the two here 🤔..but let's leave it for now).
But not really what that enhancing gives me

docs/shared/features/enhance-tooling.md
```diff
@@ -0,0 +1,102 @@
...
+- [Task distribution](/ci/features/distribute-task-execution)
+
+This works for any tool for which there exists an [official plugin](/plugin-registry).
+
+## Setup
+
+The following command will make Nx detect your tooling and set up the appropriate plugins:
+
+```shell
+nx init
```
juristr, line 16: now this is mostly for migrating to Nx on existing repos. Which for an "enhance tooling" page might make sense. I wonder though whether we should also mention the CNW option

juristr (changes requested):

docs/shared/features/enhance-tooling.md
isaacplmann: `nx init` now will detect if there are packages that you are using for which we have official plugins and give you the option to install them.  You can run it on an existing Nx repo to detect plugins that you don't have installed that would be useful.

I think getting people in the habit of running `nx init` whenever they want to add something from Nx is really good DX.

PR #21048, opened 2024-01-08
feat(angular): support esbuild middleware functions

dmitry-stepanenko:
## Current Behavior

Angular has added an ability to provide both plugins and middleware functions. However nx now only leverages the exposed plugins API.
## Expected Behavior

Should be able to configure middleware functions in the `@nx/angular:dev-server` executor.
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/angular/executors/dev-server.json
+++ b/docs/generated/packages/angular/executors/dev-server.json
@@ -109,6 +109,14 @@
         "type": "boolean",
         "description": "Read buildable libraries from source instead of building them separately. If not set, it will take the value specified in the `browserTarget` options, or it will default to `true` if it's also not set in the `browserTarget` options.",
         "x-priority": "important"
+      },
+      "esbuildMiddleware": {
+        "description": "A list of HTTP request middleware functions. _Note: this is only supported in Angular versions >= 17.0.0_.",
+        "type": "array",
+        "items": {
+          "type": "string",
+          "description": "The path to the middleware function. Relative to the workspace root."
+        }
       }
     },
     "additionalProperties": false,
--- a/packages/angular/.eslintrc.json
+++ b/packages/angular/.eslintrc.json
@@ -45,6 +45,7 @@
             "ignoredDependencies": [
               "nx",
               "eslint",
+              "vite",
               "rxjs",
               "semver",
               // These are installed by ensurePackage so missing in package.json
--- a/packages/angular/src/builders/dev-server/dev-server.impl.ts
+++ b/packages/angular/src/builders/dev-server/dev-server.impl.ts
@@ -10,8 +10,8 @@ import {
   normalizePath,
   parseTargetString,
   readCachedProjectGraph,
+  stripIndents,
   type Target,
-  targetToTargetString,
 } from '@nx/devkit';
 import { getRootTsConfigPath } from '@nx/js';
 import type { DependentBuildableProjectNode } from '@nx/js/src/utils/buildable-libs-utils';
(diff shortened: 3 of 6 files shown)


docs/generated/packages/angular/executors/dev-server.json
```diff
@@ -109,6 +109,14 @@
         "type": "boolean",
         "description": "Read buildable libraries from source instead of building them separately. If not set, it will take the value specified in the `browserTarget` options, or it will default to `true` if it's also not set in the `browserTarget` options.",
         "x-priority": "important"
+      },
+      "esbuildMiddleware": {
```
dmitry-stepanenko, line 113: was not completely sure about the property name, "middleware" seems to be too generic since this executor can be used with both webpack and esbuild

dmitry-stepanenko:
Also, should I update the docs for dev-server with a mention of `esbuildMiddleware` prop similarly to https://nx.dev/nx-api/angular/executors/browser-esbuild ?

dmitry-stepanenko:
Hi @leosvelperez, any chance this will be merged for the 17.3.0 release?

leosvelperez:
> Hi @leosvelperez, any chance this will be merged for the 17.3.0 release?

Yeah. I have it on my list for early next week. Thanks for contributing this!

packages/angular/package.json
```diff
@@ -72,7 +72,8 @@
     "@schematics/angular": ">= 15.0.0 < 18.0.0",
     "@angular-devkit/core": ">= 15.0.0 < 18.0.0",
     "rxjs": "^6.5.3 || ^7.5.0",
-    "esbuild": "^0.19.2"
+    "esbuild": "^0.19.2",
+    "vite": "^5.0.0"
```
leosvelperez, line 76: Let's not add `vite` to the plugin dependencies. We're only using a type which is only used at build time and it's stripped from the final runtime code. The actual usage of `vite` is done by the Angular builder we invoke from the `@angular-devkit/build-angular` package. That package is the one that should and does depend on it.

You can add it to the `ignoredDependencies` array in the `packages/angular/.eslintrc.json` file.

docs/generated/packages/angular/executors/dev-server.json
leosvelperez: That sounds good to me 👍🏻

leosvelperez (changes requested):

leosvelperez:
hey @dmitry-stepanenko! Thanks again for contributing this!

> Also, should I update the docs for dev-server with a mention of `esbuildMiddleware` prop similarly to https://nx.dev/nx-api/angular/executors/browser-esbuild ?

Yeah, the current `dev-server` description actually requires an overhaul. It's solely focused on the prior state where only webpack was supported. You can take a stab at that if you want, but I think it requires more changes than what this PR is for. I also want to improve the examples to be more like the ones in https://github.com/nrwl/nx/pull/20556/files#diff-f7afde405397d901e0e6d3a714023e3b06448c5976af8b894db9aede4cfc2e1c.

We can skip this for now and do it separately. Let me know if you'd like to take a stab at it. Otherwise, I'll make those changes once this is merged.

dmitry-stepanenko:
Thanks @leosvelperez. I will leave the docs for you in this case so that there's no back and forth on wording

leosvelperez (changes requested):
Sorry. One thing I forgot, we need to validate at runtime that the option was not provided when the workspace is using an Angular version that doesn't support it. Similar to what we do in https://github.com/nrwl/nx/blob/master/packages/angular/src/builders/webpack-server/webpack-server.impl.ts#L99.

dmitry-stepanenko:
Oh, my bad, should have noticed that 🙈 

Copied the error from the `browser-esbuild` executor

leosvelperez (approved):
LGTM! 🎉

meeroslav (approved):

PR #21050, opened 2024-01-08
fix(vite): only import vite dynamically

mandarini:
Only import vite dynamically

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -28,7 +28,6 @@
     "migrations": "./migrations.json"
   },
   "dependencies": {
-    "nuxi": "^3.10.0",
     "tslib": "^2.3.0",
     "@nuxt/kit": "^3.8.1",
     "@nx/devkit": "file:../devkit",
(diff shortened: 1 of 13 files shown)


ndcunningham (approved):

jaysoo (approved):

packages/cypress/src/plugins/preprocessor-vite.ts
```diff
@@ -25,7 +25,6 @@
 
 import { dirname, basename, extname } from 'path';
 import type { RollupOutput, RollupWatcher, WatcherOptions } from 'rollup';
-import type { InlineConfig } from 'vite';
```
leosvelperez, line 28: A type import doesn't affect runtime. This is only consumed at build time and stripped from the compiled output. Removing this reduces the type-safety we have and doesn't provide any benefits.

Let's revert this change to maintain type safety. We only need to lazy load the runtime code from `vite`.

leosvelperez (changes requested):

leosvelperez (approved):

PR #21052, opened 2024-01-08
feat(misc): install @nx/web if static-serve is setup as target

mandarini:
Waiting for another PR before merging

`static-serve` uses `@nx/web:file-server` executor, so we should install `@nx/web` when generating such a target.


Coly010 (approved):

jaysoo:
We'll handle this in another PR to for `--keepExistingVersions` option.

PR #21053, opened 2024-01-08
docs(core): nx replay rename

isaacplmann:
Rename Nx Cache to Nx Replay across all the docs


bcabanes (approved):

docs/generated/manifests/ci.json
```diff
@@ -131,15 +131,15 @@
     "file": "",
     "itemList": [
       {
-        "id": "remote-cache",
-        "name": "Use Remote Caching",
+        "id": "remote-replay",
+        "name": "Use Remote Replay",
```
juristr, line 135: "Remote Replay" does not make any sense to me. The "marketing term" is "Nx Replay", which replays tasks that are both, cached locally and remotely. But "Remote Replay" sounds weird to me.

Also, we should be cautious to just rename it everywhere. How would people find it searching for "Remote caching", like folks coming from other monorepo solutions?

juristr (changes requested):

docs/map.json
```diff
@@ -285,7 +285,7 @@
             },
             {
               "name": "How Caching Works",
-              "tags": ["cache-task-results"],
+              "tags": ["replay-task-results"],
```
juristr, line 288: What do we do with the title though: "How Caching Works"

Do we leave "caching" here? And how would ppl relate the "replay" feature with the functionality?

juristr:
Closing this one as discussed. We're going to fix some URL mismatches in another separate PR

PR #21054, opened 2024-01-08
fix(release): stage changes when versioning with --projects argument

fahslaj:
## Current Behavior

Nx release will not stage changes when using the `--projects` argument with `nx release version`.

## Expected Behavior

Nx release stages changes appropriately when using the `--projects` argument with `nx release version`.

## Related Issue(s)

Fixes #
#21008

### DIFF
--- a/packages/nx/src/command-line/release/version.ts
+++ b/packages/nx/src/command-line/release/version.ts
@@ -169,6 +169,28 @@ export async function releaseVersion(
 
     printAndFlushChanges(tree, !!args.dryRun);
 
+    const changedFiles = tree.listChanges().map((f) => f.path);
+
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/release/version.ts
```diff
@@ -169,6 +169,28 @@ export async function releaseVersion(
...
+      output.logSingleLine(
+        `Staging changed files with git because --stage-changes was set`
+      );
+      await gitAdd({
+        changedFiles,
+        dryRun: args.dryRun,
+        verbose: args.verbose,
+      });
+    }
+
```
fahslaj, line 193: This logic is exactly what is done below starting on line 287, but it is also needed up here for this `--projects` case

JamesHenry (approved):

PR #21073, opened 2024-01-10
feat(remix): add createNodes support for target inference

Coly010:
## Current Behavior

Remix can only be used with `project.json` files that define the required targets.

## Expected Behavior

Add `@nx/remix/plugin` to allow for inferring targets for projects where we find a `remix.config` file.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/remix/src/plugins/plugin.ts
+++ b/packages/remix/src/plugins/plugin.ts
@@ -1,0 +1,211 @@
+import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
+import {
+  type CreateDependencies,
+  type CreateNodes,
+  type CreateNodesContext,
+  type TargetConfiguration,
+  detectPackageManager,
+  readJsonFile,
+  writeJsonFile,
+} from '@nx/devkit';
+import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
+import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
+import { getLockFileName, getRootTsConfigPath } from '@nx/js';
+import { registerTsProject } from '@nx/js/src/internal';
+import { type AppConfig } from '@remix-run/dev';
+import { join, dirname } from 'path';
+import { existsSync, readdirSync } from 'fs';
+
+const cachePath = join(projectGraphCacheDirectory, 'remix.hash');
+const targetsCache = existsSync(cachePath) ? readTargetsCache() : {};
+const calculatedTargets: Record<
+  string,
+  Record<string, TargetConfiguration>
+> = {};
+
+function readTargetsCache(): Record<
+  string,
+  Record<string, TargetConfiguration>
+> {
+  return readJsonFile(cachePath);
+}
+
+function writeTargetsToCache(
+  targets: Record<string, Record<string, TargetConfiguration>>
+) {
+  writeJsonFile(cachePath, targets);
+}
+
+export const createDependencies: CreateDependencies = () => {
+  writeTargetsToCache(calculatedTargets);
+  return [];
+};
+
+export interface RemixPluginOptions {
+  buildTargetName?: string;
+  serveTargetName?: string;
+  startTargetName?: string;
+  typecheckTargetName?: string;
+}
+
+export const createNodes: CreateNodes<RemixPluginOptions> = [
+  '**/remix.config.{js,cjs,mjs}',
+  async (configFilePath, options, context) => {
+    const projectRoot = dirname(configFilePath);
+    const fullyQualifiedProjectRoot = join(context.workspaceRoot, projectRoot);
+    // Do not create a project if package.json and project.json isn't there
+    const siblingFiles = readdirSync(fullyQualifiedProjectRoot);
+    if (
+      !siblingFiles.includes('package.json') &&
+      !siblingFiles.includes('project.json')
+    ) {
+      return {};
+    }
+
+    options = normalizeOptions(options);
+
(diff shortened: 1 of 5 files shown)


packages/remix/src/plugins/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+
+export interface RemixPluginOptions {
+  buildTargetName?: string;
+  serveTargetName?: string;
+  startTargetName?: string;
+  typecheckTargetName?: string;
+}
+
+export const createNodes: CreateNodes<RemixPluginOptions> = [
+  '**/remix.config.{js,cjs}',
```
ndcunningham, line 52: Does remix support `.mjs` as well?

packages/remix/src/plugins/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+    },
+  };
+}
+
+function serveTarget(serverBuildPath: string): TargetConfiguration {
+  return {
+    executor: '@nx/remix:serve',
+    options: {
+      command: `npx remix-serve ${serverBuildPath}`,
+      port: 4200,
```
ndcunningham, line 143: Should this be hardcoded?

packages/remix/src/plugins/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+async function getServerBuildPath(
+  configFilePath: string,
+  workspaceRoot: string
+): Promise<string> {
+  const configPath = join(workspaceRoot, configFilePath);
+  let appConfig: AppConfig = {};
+  try {
+    let appConfigModule: any;
+    try {
+      appConfigModule = require(configPath);
```
ndcunningham, line 189: https://github.com/nrwl/nx/blob/master/packages/cypress/src/plugins/plugin.ts#L308
Would you be able to update it to be like this?

To handle `require` caching.

packages/remix/src/plugins/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+    } catch {
+      appConfigModule = await Function(`return import("${configPath}")`)();
```
jaysoo, line 191: We can always use `import()` and not need the try-catch here, since it works with both CJS and ESM.

packages/remix/src/plugins/plugin.ts
Coly010: I have not seen it in any of their examples or in the wild as of yet. 

They force `type: module` in the package.json which allows .js file extensions to be used.

I will test out if .mjs is supported by the current remix compiler, and if so I can update this glob

packages/remix/src/plugins/plugin.ts
Coly010: Let me double check it, but i'm pretty sure i was running into an issue with using import on the CJS config which led me to this.

packages/remix/src/plugins/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+function startTarget(
+  projectRoot: string,
+  serverBuildPath: string,
+  buildTargetName: string
+): TargetConfiguration {
+  return {
+    dependsOn: [buildTargetName],
+    command: `npx remix-serve ${serverBuildPath}`,
```
xiongemi, line 155: do you explicitly use `npx`?

packages/remix/src/plugins/plugin.ts
Coly010: Hmm... we did this to keep it consistent with the port that is used by other projects in Nx, but it may not make sense here.

cc @jaysoo

ndcunningham: Can they pass this in? with like `PORT=4200` or `--port` if it is not passed in does it default to a port?
I think the aim is not to dictate but leave it close to vanilla.

Coly010: Good point, they can pass it in so it could be removed here.

packages/remix/src/plugins/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+    executor: '@nx/remix:build',
+    options: {
+      outputPath: '{workspaceRoot}/dist',
+    },
+  };
+}
+
+function serveTarget(serverBuildPath: string): TargetConfiguration {
+  return {
+    executor: '@nx/remix:serve',
```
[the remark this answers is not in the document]
Coly010: It may need to be re-evaluated in the future but there were issues previously with just using run-commands for serving with the remix cli.

packages/remix/src/plugins/plugin.ts
Coly010: Serve has this command passed to remix’s dev option which uses it under the hood to decide how to run the dev server.

start runs a production-ready server.

packages/remix/src/plugins/plugin.ts
Coly010: It's because of Jest.

It throws 
```
TypeError [ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING]: A dynamic import callback was not specified.
```
when just using import.

packages/remix/src/plugins/plugin.ts
Coly010: .mjs is supported

packages/remix/src/plugins/plugin.ts
Coly010: I will swap it around however, because import will work and has worked in all my testing and the require should only be required for jest

jaysoo (approved):

Issue #15301, opened 2023-02-27
Nx Format The command line is too long.


Zeus-The-Cat:
### Current Behavior

Running `Nx format` using prettier.js errors with the line 
"The command line is too long." 

I'm assuming this is caused by the current implementation not taking into consideration the maximum command line limit of 8191 characters. 

### Expected Behavior

Nx format should format any project regardless of path depth. 

### Steps to Reproduce

1. Create a project with deeply nested folder structure
2. Hit the command line limit of 8191 characters 
3. format your project using Nx format and prettier.js

### Nx Report

```shell
Node : 16.16.0
   OS   : win32 x64
   npm  : 8.11.0

   nx : 13.8.1
   @nrwl/angular : 13.8.1
   @nrwl/cli : 13.8.1
   @nrwl/cypress : 13.8.1
   @nrwl/detox : undefined
   @nrwl/devkit : 13.8.1
   @nrwl/eslint-plugin-nx : 13.8.1
   @nrwl/express : undefined
   @nrwl/jest : 13.8.1
   @nrwl/js : undefined
   @nrwl/linter : 13.8.1
   @nrwl/nest : undefined
   @nrwl/next : undefined
   @nrwl/node : undefined
   @nrwl/nx-cloud : undefined
   @nrwl/react : undefined
   @nrwl/react-native : undefined
   @nrwl/schematics : undefined
   @nrwl/storybook : 13.8.1
   @nrwl/tao : 13.8.1
   @nrwl/web : undefined
   @nrwl/workspace : 13.8.1
   typescript : 4.5.5
   rxjs : 7.4.0
   ---------------------------------------
   Community plugins:
         @angular/animations: 13.2.2
         @angular/cdk: 13.2.2
         @angular/common: 13.2.2
         @angular/compiler: 13.2.2
         @angular/core: 13.2.2
         @angular/elements: 13.2.2
         @angular/forms: 13.2.2
         @angular/localize: 13.2.2
         @angular/material: 13.2.2
         @angular/material-moment-adapter: 13.2.2
         @angular/platform-browser: 13.2.2
         @angular/platform-browser-dynamic: 13.2.2
         @angular/router: 13.2.2
         @ng-bootstrap/ng-bootstrap: 7.0.0
         @ngrx/effects: 9.2.0
         @ngrx/entity: 9.2.0
         @ngrx/router-store: 9.2.0
         @ngrx/store: 9.2.0
         @ngrx/store-devtools: 9.2.0
[...]

Issue #8646, opened 2022-01-21
Jest & lint fails with the following ERROR: NX   ERROR  spawn E2BIG


dean-g:
Both my jest & lint fails with the following ERROR: NX   ERROR  spawn E2BIG

this is happening only on the CI pipeline. but the same works well in my local machine. 

```
$ yarn affected:test 

yarn run v1.22.5
$ ./node_modules/.bin/nx affected:test --base=origin/test-code --parallel --detectOpenHandles --forceExit
>  NX   ERROR  spawn E2BIG
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Cleaning up project directory and file based variables
00:01
ERROR: Job failed: exit code 1
```

Jeffrey-Zutt:
## Current Behavior
We are running into the same issue after updating from `13.4.5` to `13.4.6`.
Locally everything works just fine, but on our CI pipelines (gitlab) we're getting the following error:

```
$ npx nx affected:lint --base=origin/master
>  NX   ERROR  spawn E2BIG
```

## Expected Behavior

No E2BIG errors on `affected` commands.

## Steps to Reproduce

- Upgrade from `13.4.5` to `13.4.6`. 
- Run a `affected` command in a gitlab pipeline.

I will try to make a clean repository which reproduces this issue.

### Failure Logs
`>  NX   ERROR  spawn E2BIG`

### Environment
  NX  Report complete - copy this into the issue template
  Node : 16.13.2
  OS   : linux x64
  yarn : 1.22.15
  
  nx : 13.4.6
  @nrwl/angular : undefined
  @nrwl/cli : 13.4.6
  @nrwl/cypress : 13.4.6
  @nrwl/devkit : 13.4.6
  @nrwl/eslint-plugin-nx : 13.4.6
  @nrwl/express : 13.4.6
  @nrwl/jest : 13.4.6
  @nrwl/linter : 13.4.6
  @nrwl/nest : undefined
  @nrwl/next : undefined
  @nrwl/node : 13.4.6
  @nrwl/nx-cloud : 12.5.4
  @nrwl/react : 13.4.6
  @nrwl/react-native : undefined
  @nrwl/schematics : undefined
  @nrwl/tao : 13.4.6
  @nrwl/web : 13.4.6
  @nrwl/workspace : 13.4.6
  @nrwl/storybook : 13.4.6
  @nrwl/gatsby : undefined
  typescript : 4.5.5
[...]

PR #21074, opened 2024-01-10
fix(core): use argument length that match the actual size of the argument length

rluvaton:
## Current Behavior
Getting an argument list is too big (E2BIG) in `format:check`

## Expected Behavior
to not fail due to argument list is too big

## Related Issue(s)

- #8646 
- #15301

### DIFF
--- a/packages/nx/src/utils/chunkify.ts
+++ b/packages/nx/src/utils/chunkify.ts
@@ -1,7 +1,4 @@
-import { execSync } from 'child_process';
-
-const TERMINAL_SIZE =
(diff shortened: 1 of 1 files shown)


rluvaton:
I have no idea why the CI failed for Linux, and the macOS is something in CircleCI

AgentEnder (approved):

BurningEnlightenment:
If you could switch from `execSync` to `execFileSync` without a shell the Windows specific limit would be raised to 32k.

Issue #20688, opened 2023-12-11
Support TypeScript declarations with @nx/esbuild:esbuild


castleadmin:
- [X ] I'd be willing to implement this feature ([contributing guide](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md))

## Description
esbuild doesn't support the creation of declaration files (*.d.ts) and probably never will (see https://github.com/evanw/esbuild/issues/95).
Since declaration files are essential for published libraries,
it would be great if `@nx/esbuild:esbuild` would provide an option to output them.

## Motivation
Declaration files (*.d.ts) are essential for published libraries.

## Suggested Implementation
The commonly accepted workaround is to run a tsc build before the esbuild build that only outputs the declaration files.
This additional build step could be provided as a `@nx/esbuild:esbuild` option.

## Alternate Implementations
Implement the additional build step in the `@nx/js:library` generator for publishable libraries.

jaysoo:
@castleadmin Hey, thanks for expressing interest on this, do you have any suggestions on how you'd want to do this?

castleadmin:
Hi @jaysoo, for the suggested case I would do the following

- Introduce a new boolean valued `--declaration` option
- If `--declaration`  or the tsconfig option [declaration](https://www.typescriptlang.org/tsconfig#declaration) is true,
  then the TypeScript compiler will be used before the esbuild build call to generate the declarations
- The TypeScript compiler could be called via the command line or the @nx/js:tsc executor
- The same `tsconfig` options should be used for the esbuild and declarations build with one exception
  - The declaration build will always set the [emitDeclarationOnly](https://www.typescriptlang.org/tsconfig#emitDeclarationOnly) tsconfig option

castleadmin:
@jaysoo What do you think about the new feature? Should I start implementing a concrete solution?

GideonMax:
Hello, what are the plans on implementing this, and are there any workarounds I can do to achieve this behavior?

GideonMax:
When can we expect this feature?

castleadmin:
I should be able to implement the feature within the next 2 weeks.

ChristopherPHolder:
What is missing to complete this feature? It seems the MR is open and ready for review. 

I am happy to contribute if there is anything needed to complete the feature 👍

ThePlenkov:
Here is the way I have fixed it:
```

    "dts": {
      "executor": "@nx/js:tsc",
      "outputs": [
        "{options.outputPath}"
[...]

PR #21084, opened 2024-01-10
feat(bundling): added support for declarations (*.d.ts)

castleadmin:
## Current Behavior
esbuild doesn't support the creation of declaration files (*.d.ts) and probably never will (see https://github.com/evanw/esbuild/issues/95).
Since declaration files are essential for published libraries,
it would be great if @nx/esbuild:esbuild would provide an option to output them.

## Expected Behavior

- Introduced a new boolean valued `declaration` option for the `esbuild` executor
- If `declaration` or the tsconfig option [declaration](https://www.typescriptlang.org/tsconfig#declaration) is true,
then the TypeScript compiler is used before esbuild to generate the declarations
- The output directory structure of the declarations can be influenced by setting the TypeScript rootDir via the `declarationRootDir` option

## Related Issue(s)
https://github.com/nrwl/nx/issues/20688

### Additional Comment
Please note that the generated declaration files directory structure is affected by the tsconfig `rootDir` property.

For a library that doesn't reference other libraries inside the monorepo, the `rootDir` property can be changed freely.
If a library inside the monorepo is referenced the `rootDir` property must be set to the workspace root.

The `tsc` executor has a sophisticated check that automatically sets the `rootDir` to the workspace root if a library is referenced.
https://github.com/nrwl/nx/blob/master/packages/js/src/executors/tsc/tsc.impl.ts#L104

This check is quite complex and specific to the `tsc` executor options. Therefore, it hasn't been included inside the esbuild implementation.

The current implementation leaves it to the user to solve the edge case by removing the `declarationRootDir` option or by setting the `declarationRootDir` to `.`.

In the future, it might make sense to generalize and use the `tsc` executor check.

### DIFF
--- a/docs/generated/packages/esbuild/executors/esbuild.json
+++ b/docs/generated/packages/esbuild/executors/esbuild.json
@@ -5,7 +5,7 @@
     "version": 2,
     "outputCapture": "direct-nodejs",
     "title": "esbuild (experimental)",
-    "description": "Bundle a package for different platforms. Note: declaration (*.d.ts) file are not currently generated.",
+    "description": "Bundle a package for different platforms.",
     "cli": "nx",
     "type": "object",
     "properties": {
@@ -51,6 +51,14 @@
         "alias": "f",
         "items": { "type": "string", "enum": ["esm", "cjs"] },
         "default": ["esm"]
+      },
+      "declaration": {
+        "type": "boolean",
+        "description": "Generate declaration (*.d.ts) files for every TypeScript or JavaScript file inside your project. Should be used for libraries that are published to an npm repository."
+      },
+      "declarationRootDir": {
+        "type": "string",
+        "description": "Sets the rootDir for the declaration (*.d.ts) files."
       },
       "watch": {
         "type": "boolean",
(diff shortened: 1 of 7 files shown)


Afellman:
Any progress on this? My team is waiting for it.

mandarini:
Tested this locally, and the tests on CI seem to pass, so I am going to merge.

However, once `esbuild` is part of the crystal project, this functionality will go away, since we will be relying on the native features of esbuild.

mandarini (approved):

danielsharvey:
> Tested this locally, and the tests on CI seem to pass, so I am going to merge.

Thank you

> 
> However, once `esbuild` is part of the crystal project, this functionality will go away, since we will be relying on the native features of esbuild.

What approach would you recommend in the context of Project Crystal?

castleadmin:
An esbuild plugin could be a viable alternative to support declaration file builds.

[esbuild-plugin-d.ts](https://www.npmjs.com/package/esbuild-plugin-d.ts) is listed on the esbuild plugin page and provides the needed functionality.

Another solution that doesn't rely on third-party plugins would be to split the current implementation into 2 commands.

- Build the declaration files with the help of tsc
- Build the project with esbuild

If the clearing of the output directory and caching doesn't interfere, then a solution that uses 2 commands should work fine.

danielsharvey:
> Another solution that doesn't rely on third-party plugins would be to split the current implementation into 2 commands.
> 
> * Build the declaration files with the help of tsc
> * Build the project with esbuild
> 

👍🏻 This is what I currently do.

PR #21083, opened 2024-01-10
docs(core): improve run commands in parallel section

edbzn:
## Current Behavior

Execution of background processes involves the repetitive task of storing PIDs and running commands in the background.

## Expected Behavior

Improved doc section about running background processes in parallel, this can be useful if we want to run multiple commands with different arguments, for instance:

```bash
run_command "npx nx affected -t lint,test --parallel=3"
run_command "npx nx affected -t build --parallel=1"
```

If you like this change I can also raise a PR to modify the actual `circleci/config.yml`.

PR #21086, opened 2024-01-11
chore(repo): optimize circleci parallel run command

edbzn:
Follow up of https://github.com/nrwl/nx/pull/21083.

## Current Behavior

CircleCI config can be optimized by using a function to run processes in the background and store PIDs.

## Expected Behavior

Usage of the `run_command` makes it easier to read.


.circleci/config.yml
```diff
@@ -109,15 +109,18 @@ jobs:
...
-            pids+=($!)
+            run_command "pnpm nx-cloud record -- nx format:check --base=$NX_BASE --head=$NX_HEAD"
 
-            (pnpm nx affected --targets=lint,test,build --base=$NX_BASE --head=$NX_HEAD --parallel=3 &&
-            pnpm nx affected --target=e2e --base=$NX_BASE --head=$NX_HEAD --parallel=1) &
-            pids+=($!)
+            run_command "pnpm nx run-many -t check-imports check-commit check-lock-files check-codeowners documentation --parallel=1 --no-dte"
+
+            run_command "pnpm nx affected --targets=lint,test,build --base=$NX_BASE --head=$NX_HEAD --parallel=3"
+            run_command "pnpm nx affected --target=e2e --base=$NX_BASE --head=$NX_HEAD --parallel=1"
```
meeroslav, line 123: The E2E commands should not run in parallel to the others. Since they are very expensive, we only run them if lint, test and build passed.

```suggestion
            run_command "(pnpm nx affected --targets=lint,test,build --base=$NX_BASE --head=$NX_HEAD --parallel=3 && pnpm nx affected --target=e2e --base=$NX_BASE --head=$NX_HEAD --parallel=1)"
```

meeroslav (changes requested):

.circleci/config.yml
meeroslav: I took the liberty to update the necessary change so that we can merge it. Hope you don't mind.

meeroslav (approved):

meeroslav:
Sadly this approach doesn't work for running sequential pair, like we have at the moment (small repro visible here https://github.com/meeroslav/test-bash-parallelism/actions/runs/7581496823/job/20649211850) so we will not be merging this in.

Thank you for the effort. This is still valuable proposal for most of the folks so we will keep it in the docs.

meeroslav:
No, but feel free to play with my repo linked above. We kept getting exit status 127 when trying to use that function to run `({command} && {command})` and choosing that one to run differently makes the whole cleanup questionable.

If you find out what is wrong and fix it, I would gladly reopen the PR and merge it.

PR #21088, opened 2024-01-11
cleanup(misc): clean up init generators

leosvelperez:
This PR simplifies the tasks performed by `init` generator in the following way:

- Remove calls to other `init` generators
- Install the minimum required dependencies needed for the plugin to work
  - Most dependencies are only needed and can be installed by the generators invoking the `init` generators
- Configure the minimum required for the plugin to work
  - Most configuration changes are only needed and can be done by the generators invoking the `init` generators

The `init` generators were always internal and meant to be called by other generators to ensure the base dependencies and configuration were set up. This is changing. While they'll continue to be internal, soon they could also be invoked when running [`nx add`](https://github.com/nrwl/nx/pull/20976) and [`nx init`](https://github.com/nrwl/nx/pull/20872). In those scenarios, we only need to initialize the minimum required for the plugin to work. We wouldn't be generating anything just yet and can't make assumptions and install/configure things that might not be needed by the workspace.

## Related Issue(s)

Fixes #


Coly010 (approved):

mandarini (commented):
Just two small comments. I'm testing this out locally and then I'll approve.

packages/nuxt/src/generators/init/lib/utils.ts
```diff
@@ -1,48 +1,21 @@
...
-    devDependencies['less'] = lessVersion;
-  }
-
-  return addDependenciesToPackageJson(host, {}, devDependencies);
+  return addDependenciesToPackageJson(
+    host,
+    {},
+    {
+      '@nx/nuxt': nxVersion,
+      nuxi: nuxtVersion,
```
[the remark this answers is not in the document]
leosvelperez: The `addDependenciesToPackageJson` function is getting a new option to [not overwrite existing installed packages](https://github.com/nrwl/nx/pull/21123). When that lands, we'll update our generators with it. That will be done either separately if it's delayed or I'll amend this PR if merged in time.

packages/vue/src/generators/init/init.ts
```diff
@@ -1,67 +1,40 @@
...
-  if (schema.style === 'scss') {
-    devDependencies['sass'] = sassVersion;
-  } else if (schema.style === 'less') {
-    devDependencies['less'] = lessVersion;
-  }
+  tasks.push(removeDependenciesFromPackageJson(host, ['@nx/vue'], []));
+  tasks.push(
+    addDependenciesToPackageJson(
+      host,
+      { vue: vueVersion },
```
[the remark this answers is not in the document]
leosvelperez: Same as previous.

mandarini (approved):
Tested it locally:
- generate new workspace
- add react app
- add vue app
- add nuxt app
- add storybook

All works as expected.

packages/jest/src/generators/init/init.ts
```diff
@@ -223,23 +57,21 @@ function updateProductionFileSet(tree: Tree) {
...
 
-    nxJson.targetDefaults['@nx/jest:jest'].cache ??= true;
-    // Test targets depend on all their project's sources + production sources of dependencies
-    nxJson.targetDefaults['@nx/jest:jest'].inputs ??= [
-      'default',
-      productionFileSet ? '^production' : '^default',
-      '{workspaceRoot}/jest.preset.js',
-    ];
-  }
+  nxJson.targetDefaults['@nx/jest:jest'].cache ??= true;
```
jaysoo, line 68: How come we removed the check for `@nx/jest/plugin`? I don't think we want the targetDefaults fo PCv3.

jaysoo:
Added a question, the rest LGTM.

packages/jest/src/generators/init/init.ts
leosvelperez: The check is already done when invoking this function https://github.com/nrwl/nx/pull/21088/files#diff-cbef53cf0c3f36af94e1a3f3ad60780ad4c6f884c04f3800afe22f972cfb3efaR106-R110.

jaysoo (approved):

packages/eslint/src/generators/init/init.ts
```diff
@@ -127,22 +88,22 @@ function initEsLint(tree: Tree, options: LinterInitOptions): GeneratorCallback {
     addTargetDefaults(tree);
   }
 
-  updateVSCodeExtensions(tree);
-
-  return !options.skipPackageJson
-    ? addDependenciesToPackageJson(
+  const tasks: GeneratorCallback[] = [];
+  if (!options.skipPackageJson) {
+    tasks.push(removeDependenciesFromPackageJson(tree, ['@nx/eslint'], []));
```
meeroslav, line 93: If the purpose of removing this dependency before re-adding it was to avoid version clash if the same package is installed in `dependencies` and `devDependencies` then wouldn't we want to do the same with the other packages listed below - `eslint`, `@nx/eslint-plugin`, `@typescript-eslint/*`?

meeroslav (approved):
Left one minor comment but otherwise looks good to me.

packages/eslint/src/generators/init/init.ts
jaysoo: No, there's follow-up work to use the updated `addDependenciesToPackageJson`, the util PR is pending merge: https://github.com/nrwl/nx/pull/21123

leosvelperez: I assume that was the reason, but I'm unsure. I just left the existing behavior for this. IMHO, we shouldn't do it for any. AFAIK, the `addDependenciesToPackageJson` function handles this though maybe differently. I haven't checked.

PR #21094, opened 2024-01-11
feat(core): extend nxCloud prompt to include basic CI workflow options

meeroslav:
Enhance `connect` command by adding CI workflow generation options.

- [X] CNW implementation 
- [X] migration implementation 
- [X] add-to-* + init implementation 
- [x] add free + reversible disclaimer
- [x] docs update
- [x] ensure (partial) backwards compatibility & fix tests

## A/B options

**Do you want Nx Cloud to make your CI fast?**
*(it’s free and can be disabled any time)*
- Yes, enable Nx Cloud
- Yes, configure Nx Cloud for GitHub Actions
- Yes, configure Nx Cloud for Circle CI
- Skip for now

**Enable remote caching to make your CI fast**
*(it’s free and can be disabled any time)*
- Yes
- Yes, with GitHub Actions
- Yes, with CircleCI
- Skip for now

## Related Issue(s)

Fixes #


mandarini (approved):

packages/create-nx-workspace/src/utils/nx/ab-testing.ts
```diff
@@ -1,26 +1,54 @@
...
+      fallback: { value: 'skip', key: 'setupNxCloud' },
     },
   ],
-  nxCloudMigration: [
+  setupNxCloud: [
     {
-      code: 'make-ci-faster',
-      message: `Enable remote caching to make your CI faster?`,
+      code: 'enable-caching',
+      message: `Would you like remote caching to make your build faster`,
```
JamesHenry, line 34: The other question had a question mark

```suggestion
      message: `Would you like remote caching to make your build faster?`,
```

Open question, is “build” too specific? Should we say “tasks” instead? Or “builds, tests etc”

packages/nx/src/utils/ab-testing.ts
```diff
@@ -1,39 +1,43 @@
...
-      },
-    ],
-  };
+export type MessageOptionKey = 'yes' | 'skip';
+
+const messageOptions = {
+  setupNxCloud: [
+    {
+      code: 'enable-caching',
+      message: `Would you like remote caching to make your build faster`,
```
JamesHenry, line 9: Same as above

JamesHenry (commented):
Left some comments/questions

packages/create-nx-workspace/src/utils/nx/ab-testing.ts
meeroslav: We had these discussions in the past. I was also on `task` team, as it's more accurate but the reason for going with build is that it's historically the pain point people care about.

packages/nx/src/utils/ab-testing.ts
meeroslav: Fixed

FrozenPandaz (approved):

PR #21096, opened 2024-01-11
docs(misc): update nx replay copy

ZackDeRose:
## Current Behavior

Nx Replay seems to imply local caching is part of the product

## Expected Behavior

Nx Replay only refers to remote task caching on the Nx Cloud platform

## Related Issue(s)

Fixes #


ZackDeRose:
leaving as-is per slack conversation

PR #21107, opened 2024-01-12
feat(release): update lockfile after version command

fahslaj:
## Current Behavior

`nx release` does not update package manager lockfiles after performing a version bump. This means that the first install command after `nx release` will result in a changed lockfile.

## Expected Behavior

`nx release` updates lockfiles after performing a version bump and includes the lockfile in the release commit.


packages/js/src/generators/release-version/utils/package-manager-install.ts
```diff
@@ -0,0 +1,128 @@
+import { detectPackageManager, output } from '@nx/devkit';
+import { exec } from 'child_process';
+import { gte } from 'semver';
+
+export async function packageManagerInstall({
```
JamesHenry, line 5: Did the existing `installPackagesTask()` not work? If not we should probably look to enhance that one rather than create a fresh one

The dry run element can live outside of this as well (so that it aligns better with `installPackagesTask`), the returned function from the generator could apply that check and call `installPackagesTask` within its implementation

```ts
installCallback: () => {
  if (!dryRun) {
    installPackagesTask();
  }
 // ...etc...
}
```

packages/js/src/generators/release-version/release-version.ts
```diff
@@ -27,12 +28,13 @@ import * as ora from 'ora';
...
 import { prerelease } from 'semver';
 import { ReleaseVersionGeneratorSchema } from './schema';
+import { packageManagerInstall } from './utils/package-manager-install';
 import { resolveLocalPackageDependencies } from './utils/resolve-local-package-dependencies';
 
 export async function releaseVersionGenerator(
   tree: Tree,
   options: ReleaseVersionGeneratorSchema
-) {
+): Promise<ReleaseVersionGeneratorResult> {
```
JamesHenry, line 37: In a dedicated follow up (because it needs some consideration around trying not to break existing usage), we can maybe generalize an updated `GeneratorResult` interface.

Currently a generator can return `void | GeneratorCallback` I believe.

I think it's reasonable that any generator could return arbitrary data, as we've seen already with versioning, and as we know we still need the `GeneratorCallback`, so maybe in the future:

```ts
generator(...): void | GeneratorCallback | GeneratorResult;

interface GeneratorResult {
  callback?: GeneratorCallback;
  data?: Record<string, unknown>;
}
```

And maybe even in v18 just `generator(...): GeneratorResult;` with the two optional properties

fahslaj: We can do this in a follow up. The existing GeneratorCallback will need to be changed because it currently doesn't accept any arguments and always returns `void` or `Promise<void>`, whereas this callback needs to return data.

packages/js/src/generators/release-version/utils/package-manager-install.ts
fahslaj: The existing `installPackagesTask()` would work if I heavily modified that function to take additional arguments. I don't think we should do that, since updating the lock file is a different task than just installing packages. I think we should keep a separate `updateLockFile` function for now and move it to @nx/devkit if it's ever needed elsewhere outside of `nx release`.

packages/js/src/generators/release-version/release-version.ts
fahslaj: Something else to note is that in the case of `nx release`, the version generator is called once per release group. We don't actually want to perform the lock file update once per release group though, just once at the end of all updates. The lock file update callback can be a static property on the generator.

JamesHenry (approved):

PR #21111, opened 2024-01-12
feat(remix): use esm config file

Coly010:
## Current Behavior

We currently generate a CJS config file, and our utils for interacting with Remix config files expect a CJS config file.

## Expected Behavior

Switch to using ESM for generated apps and update utils to support both ESM and CJSx

## Related Issue(s)

Fixes #


packages/remix/src/utils/remix-config.ts
```diff
@@ -1,19 +1,62 @@
...
+    workspaceRoot,
+    getRemixConfigPath(tree, projectName)
+  );
+  const cacheKey = `${projectName}/${remixConfigPath}`;
+  let appConfig = _remixConfigCache[cacheKey];
+
+  if (!appConfig) {
+    try {
+      const importedConfig = await Function(
+        `return import("${remixConfigPath}")`
```
jaysoo, line 52: To avoid import cache, you can append a querystring to the end.

e.g. `import("${remixConfigPath}?t=${Date.now()}"`

Otherwise, changes to the file will not reload config in Nx daemon.

We should do this for other plugins as well.

jaysoo (approved):

jaysoo:
Added a comment about avoiding import cache, otherwise LGTM.

packages/remix/src/utils/remix-config.ts
Coly010: Interesting, ok!

PR #21123, opened 2024-01-12
feat(devkit): add a flag to keep existing versions when calling addDependenciesToPackageJson

jaysoo:
This PR adds a flag to `addDependenciesToPackageJson` function in devkit to prevent versions from being bumped. This will be used in init generators since using `nx init` and `nx add` should not bump versions and risk breaking existing projects.

**Example:**

If existing `package.json` already has `next@13.4.0`, then calling this function:
```ts
addDependenciesToPackageJson(
  tree,
  {
    next: '~14.0.0'
  {},
  undefined,
  true // keepExistingVersions
);
```

Will leave `next` at existing version of `13.4.0`.

## Notes

- There was discussion on converting the args to accept an options object, but this will break existing usages with signature `addDependenciesToPackageJson(tree, {...}, {...}, customPackageJsonPath)`, unless we do function overrides.

### DIFF
--- a/packages/devkit/src/utils/package-json.ts
+++ b/packages/devkit/src/utils/package-json.ts
@@ -132,13 +132,15 @@ function updateExistingDependenciesVersion(
  * @param dependencies Dependencies to be added to the dependencies section of package.json
  * @param devDependencies Dependencies to be added to the devDependencies section of package.json
  * @param packageJsonPath Path to package.json
+ * @param keepExistingVersions If true, prevents existing dependencies from being bumped to newer versions
  * @returns Callback to install dependencies only if necessary, no-op otherwise
  */
 export function addDependenciesToPackageJson(
   tree: Tree,
   dependencies: Record<string, string>,
   devDependencies: Record<string, string>,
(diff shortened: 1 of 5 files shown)


packages/devkit/src/utils/package-json.ts
```diff
@@ -227,12 +240,24 @@ function removeLowerVersions(
 ) {
   return Object.keys(incomingDeps).reduce((acc, d) => {
     if (
-      existingDeps?.[d] &&
-      !isIncomingVersionGreater(incomingDeps[d], existingDeps[d])
+      !existingDeps?.[d] ||
```
FrozenPandaz, line 243: I did not imagine this function should change at all in this change.

packages/devkit/src/utils/package-json.ts
```diff
@@ -187,6 +189,17 @@ export function addDependenciesToPackageJson(
     currentPackageJson.devDependencies
   );
```
FrozenPandaz, line 191: There are some comments above this code which should be updated

FrozenPandaz (approved):

PR #21141, opened 2024-01-15
feat(core): update ci-workflow generator

meeroslav:
This PR updates our `ci-workflow` generator templates to be:
- as minimal as possible
- cleaned up of any unnecessary stuff
- aligned between different providers
- use non-DTE by default
- have DTE with agents commented out
- additionally fixes test to properly detect package manager

> Note: Docs changes will follow after the official release of Agents

## Related Issue(s)

Fixes #


juristr (approved):

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -21,63 +20,24 @@ variables:
...
-      - script: pnpm exec nx affected --base=$(BASE_SHA) --head=$(HEAD_SHA) --target=build --parallel=3
-        displayName: Run build
-      - script: pnpm exec nx-cloud stop-all-agents
-        condition: always()
-        displayName: Stop all Nx-Cloud agents
+      - script: npm ci
+      # uncomment to enable task distribution
+      # - script: npx nx-cloud start-ci-run --distributes-on="5 linux-medium"
+      - script: npx nx-cloud record -- nx format:check --base=$(BASE_SHA) --head=$(HEAD_SHA)
+      - script: npx nx affected --base=$(BASE_SHA) --head=$(HEAD_SHA) --target=lint,test,build --parallel=3
```
JamesHenry, line 31: Let’s go with the short form -t

Victor prefers it as it’s visually cleaner but also makes sense not to emphasize target based on our recent discussion

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -87,93 +47,40 @@ pipelines:
...
-                - pnpm exec nx affected --target=lint & pnpm exec nx affected --target=test & pnpm exec nx affected --target=build
-                - pnpm exec nx-cloud stop-all-agents
-          - step: *agent
-          - step: *agent
-          - step: *agent
+                - npm ci
+                # uncomment to enable task distribution
+                # - npx nx-cloud start-ci-run --distributes-on="5 linux-medium"
+                - npx nx-cloud record -- nx format:check
+                - npx nx affected --target=lint,test,build --parallel=3
```
JamesHenry, line 54: -t

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -87,93 +47,40 @@ pipelines:
...
-          name: Run build
-          command: pnpm exec nx affected --base=$NX_BASE --head=$NX_HEAD --target=build --parallel=3
-      - run:
-          name: Stop all agents
-          command: pnpm exec nx-cloud stop-all-agents
-          when: always
+      # uncomment to enable task distribution
+      # - run: npx nx-cloud start-ci-run --distributes-on="5 linux-medium"
+      - run: npx nx-cloud record -- nx format:check --base=$NX_BASE --head=$NX_HEAD
+      - run: npx nx affected --base=$NX_BASE --head=$NX_HEAD --target=lint,test,build --parallel=3
```
JamesHenry, line 76: -t

JamesHenry: Is the passing of base and head really necessary here? It should be read from the env automatically? Is there something specific about circle that doesn’t allow it to work?

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -186,32 +93,30 @@ on:
...
+          node-version: 20
+          cache: 'npm'
+      - run: npm ci
+      - uses: nrwl/nx-set-shas@v4
+
+      # uncomment to enable task distribution
+      # - run: npx nx-cloud start-ci-run --distributes-on="5 linux-medium"
+
+      - run: npx nx-cloud record -- nx format:check
+      - run: npx nx affected -target lint,test,build --parallel=3
```
JamesHenry, line 119: -t

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -186,32 +93,30 @@ on:
...
+          fetch-depth: 0
+      # Cache node_modules
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          cache: 'npm'
+      - run: npm ci
+      - uses: nrwl/nx-set-shas@v4
+
+      # uncomment to enable task distribution
```
JamesHenry, line 115: Should we also advise people to remove the parallel 3 when they do this to avoid unnecessary limitations to throughput?

JamesHenry: (Applies in multiple locations)

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -224,81 +129,52 @@ on:
...
+      # - run: npx nx-cloud start-ci-run --distributes-on="5 linux-medium"
+
+      - run: npx nx-cloud record -- nx format:check
+      - run: npx nx affected -target lint,test,build --parallel=3
```
JamesHenry, line 155: -t

JamesHenry (changes requested):
Left comments inline on the final rendered snapshots as they were much easier to read from my phone

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
meeroslav: This only works in GitHub. Other CI providers allow passing variables between steps but do not add them to process environments. We had tested this in the past.

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
meeroslav: This strongly depends on the type of machine used, the complexity of the tasks, etc. Most of our customers still use `parallel=3`. This is something that should be documented in our recipes but not necessarily in the generated workflows.

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
JamesHenry: Ah ok thanks

JamesHenry (approved):

PR #21143, opened 2024-01-15
chore(repo): bump pnpm and nx-orb versions

meeroslav:
## Related Issue(s)

Fixes #


JamesHenry (approved):

Cammisuli:
We should update this too then

https://github.com/nrwl/nx/blob/master/.github/workflows/publish.yml#L92

PR #21146, opened 2024-01-15
feat(remix): add init generator

Coly010:
## Current Behavior

Remix plugin does not have an init generator that can be used either by other Remix generators or when initing nx in a non-Nx workspace
## Expected Behavior

Add an init generator

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/remix/src/generators/application/application.impl.ts
+++ b/packages/remix/src/generators/application/application.impl.ts
@@ -34,10 +34,16 @@ import {
   updateUnitTestConfig,
 } from './lib';
 import { NxRemixGeneratorSchema } from './schema';
+import { updateDependencies } from '../utils/update-dependencies';
+import initGenerator from '../init/init';
+import { initGenerator as jsInitGenerator } from '@nx/js';
 
 export default async function (tree: Tree, _options: NxRemixGeneratorSchema) {
   const options = await normalizeOptions(tree, _options);
-  const tasks: GeneratorCallback[] = [];
+  const tasks: GeneratorCallback[] = [
+    await initGenerator(tree, { skipFormat: true }),
+    await jsInitGenerator(tree, { skipFormat: true }),
+  ];
 
   addProjectConfiguration(tree, options.projectName, {
     root: options.projectRoot,
@@ -78,25 +84,7 @@ export default async function (tree: Tree, _options: NxRemixGeneratorSchema) {
     },
   });
 
-  const installTask = addDependenciesToPackageJson(
-    tree,
-    {
-      '@remix-run/node': remixVersion,
-      '@remix-run/react': remixVersion,
-      '@remix-run/serve': remixVersion,
-      isbot: isbotVersion,
-      react: reactVersion,
-      'react-dom': reactDomVersion,
-    },
-    {
-      '@remix-run/dev': remixVersion,
-      '@remix-run/eslint-config': remixVersion,
-      '@types/react': typesReactVersion,
-      '@types/react-dom': typesReactDomVersion,
-      eslint: eslintVersion,
-      typescript: typescriptVersion,
-    }
-  );
+  const installTask = updateDependencies(tree);
   tasks.push(installTask);
 
   const vars = {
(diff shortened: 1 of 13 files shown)


packages/remix/src/generators/init/init.ts
```diff
@@ -0,0 +1,70 @@
...
+      options: {
+        buildTargetName: 'build',
+        serveTargetName: 'serve',
+        startTargetName: 'serve',
+        typecheckTargetName: 'typecheck',
+      },
+    });
+  }
+
+  updateNxJson(tree, nxJson);
```
leosvelperez, line 42: We can skip updating `nx.json` if the plugin is already present. We could bail out early or move the `updateNxJson` call inside the current `if` statement:
```suggestion
  for (const plugin of nxJson.plugins) {
    if (
      typeof plugin === 'string'
        ? plugin === '@nx/remix/plugin'
        : plugin.plugin === '@nx/remix/plugin'
    ) {
      return;
    }
  }

  nxJson.plugins.push({
    plugin: '@nx/remix/plugin',
    options: {
      buildTargetName: 'build',
      serveTargetName: 'serve',
      startTargetName: 'serve',
      typecheckTargetName: 'typecheck',
    },
  });

  updateNxJson(tree, nxJson);
```

packages/remix/src/generators/application/application.impl.ts
```diff
@@ -78,25 +79,7 @@ export default async function (tree: Tree, _options: NxRemixGeneratorSchema) {
...
-    {
-      '@remix-run/dev': remixVersion,
-      '@remix-run/eslint-config': remixVersion,
-      '@types/react': typesReactVersion,
-      '@types/react-dom': typesReactDomVersion,
-      eslint: eslintVersion,
-      typescript: typescriptVersion,
-    }
-  );
+  const installTask = updateDependencies(tree);
```
leosvelperez, line 82: We should still invoke the plugin `init` generator. It would cover the scenario of someone installing the plugin (without using `nx add`) and generating an application.

Additionally, we invoke this function that installs additional dependencies not installed by the `init` generator.

packages/remix/src/generators/utils/update-dependencies.ts
```diff
@@ -0,0 +1,33 @@
...
+  typesReactVersion,
+} from '../../utils/versions';
+
+export function updateDependencies(tree: Tree) {
+  return addDependenciesToPackageJson(
+    tree,
+    {
+      '@remix-run/node': remixVersion,
+      '@remix-run/react': remixVersion,
+      '@remix-run/serve': remixVersion,
```
leosvelperez, line 19: Not an issue, but we could leave this one out since it will be installed by the `init` generator (that should be called by any generator needing to ensure the plugin is set up).

packages/remix/src/generators/utils/update-dependencies.ts
```diff
@@ -0,0 +1,33 @@
...
+      isbot: isbotVersion,
+      react: reactVersion,
+      'react-dom': reactDomVersion,
+    },
+    {
+      '@remix-run/dev': remixVersion,
```
leosvelperez, line 25: We could leave this one out since it should be installed by the `init` generator.

leosvelperez (changes requested):

leosvelperez (approved):

packages/remix/src/generators/init/init.ts
```diff
@@ -0,0 +1,69 @@
...
+  runTasksInSerial,
+} from '@nx/devkit';
+import { type Schema } from './schema';
+import { remixVersion } from '../../utils/versions';
+
+function addPlugin(tree) {
+  const nxJson = readNxJson(tree);
+  nxJson.plugins ??= [];
+
+  let hasNxRemixPlugin = false;
```
mandarini, line 17: maybe this not needed?

mandarini (approved):

packages/remix/src/generators/init/init.ts
Coly010: ahh, good catch, leftovers from cleanup

meeroslav (approved):

PR #21157, opened 2024-01-16
fix(angular): run function is not called in setup-ssr/application-builder

jiayisheji:
## Related Issue(s)

```base
nx generate @nx/angular:application --name=demo --bundler=esbuild --ssr=true
```

In the generated `server.ts` file, the `run` function is still not called, so `ssr` does not work properly.


leosvelperez (approved):
Thanks for the contribution! LGTM 🎉

PR #21167, opened 2024-01-16
docs(core): inferred targets

isaacplmann:
Branch containing all changes related to inferred tasks


Coly010 (approved):

mandarini (approved):
Approved, but I thought we would be adding the "inferred tasks" docs inside each plugin intro page?

meeroslav (approved):

PR #21186, opened 2024-01-17
feat(graph): decouple graph client from nx.dev <Fence> component

jaysoo:
This PR adds a new component for rendering JSON blocks inside graph client.

## Changes

- Rename `@nx/shared-ui-fence` to `@nx/nx-dev/ui-fence`
- Remove `@nx/shared-ui-selector` and inline `Selector` component in `@nx/nx-dev/ui-fence` lib since it's only used there as an internal component
- Add `@nx/graph/ui-code-block` lib to render JSON blocks for the project detail view
- Add `preact/debug` to graph client so we can use devtools


isaacplmann (approved):

MaxKless (approved):
LGTM! Some styling changes should be made imo but the scope of this PR is just clean-up so I'd merge it like this for now.

meeroslav (approved):

FrozenPandaz (approved):

PR #21204, opened 2024-01-18
feat(misc): optionally update package.json scripts in init generators

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/utils/update-package-scripts.ts
+++ b/packages/devkit/src/utils/update-package-scripts.ts
@@ -60,6 +60,10 @@ async function processProject(
     return;
   }
 
+  // exclude package.json scripts from nx
+  packageJson.nx ??= {};
+  packageJson.nx.includedScripts ??= [];
+
   for (const targetCommand of targetCommands) {
     const { command, target, configuration } = targetCommand;
     const targetCommandRegex = new RegExp(
@@ -75,6 +79,7 @@ async function processProject(
             ? `$1nx ${target} --configuration=${configuration}$3`
             : `$1nx ${target}$3`
         );
+        excludeScriptFromPackageJson(packageJson, scriptName);
       } else {
         /**
          * Parse script and command to handle the following:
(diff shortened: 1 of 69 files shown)


isaacplmann (approved):

mandarini (approved):

packages/devkit/src/utils/update-package-scripts.ts
```diff
@@ -0,0 +1,224 @@
...
+  );
+
+  const targetCommands = getInferredTargetCommands(result);
+  if (!targetCommands.length) {
+    return;
+  }
+
+  // exclude package.json scripts from nx
+  packageJson.nx ??= {};
+  packageJson.nx.includedScripts ??= [];
```
leosvelperez, line 65: @FrozenPandaz @jaysoo do we want to include all scripts by default and only remove the ones we end up replacing?

Right now, the PR excludes everything if there is no `nx` entry in the `package.json`. If there's an `nx` entry, it's respected, but replaced scripts are removed.

packages/nx/src/command-line/add/add.ts
```diff
@@ -104,7 +107,18 @@ async function initializePlugin(pkgName: string): Promise<void> {
...
 
   try {
-    await runNxAsync(`g ${pkgName}:${initGenerator}`);
+    let updatePackageScripts: boolean;
+    if (options.updatePackageScripts !== undefined) {
+      updatePackageScripts = options.updatePackageScripts;
+    } else {
+      updatePackageScripts =
+        process.env.NX_PCV3 === 'true' && coreNxPlugins.includes(pkgName);
+    }
```
leosvelperez, line 116: @jaysoo @FrozenPandaz should we ignore `--updatePackageScripts` if no `NX_PCV3=true`?

jaysoo: Let's keep it as is, we will be removing the flag later anyway.

jaysoo (approved):

FrozenPandaz (approved):
This is a good start. Let's iterate off of this.

PR #21208, opened 2024-01-18
fix(graph): take vscode light/dark theme into account

MaxKless:
getting the system theme is changed. First we'll check for `vscode-dark` or `vscode-light` classes on the body (those are set by vscode automatically). If either exists, we'll use that as the 'system' theme.
If it doesn't, we'll defer to the `prefers-color-scheme: dark` media query.

We also listen to changes on the document body.


FrozenPandaz (approved):
Fancy Migration observer

PR #21213, opened 2024-01-18
fix(expo): fix externalDependencies for expo plugin

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/detox/src/plugins/plugin.ts
+++ b/packages/detox/src/plugins/plugin.ts
@@ -17,6 +17,7 @@ import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
 
 export interface DetoxPluginOptions {
   buildTargetName?: string;
+  startTargetName?: string;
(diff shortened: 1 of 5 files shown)


packages/detox/src/plugins/plugin.ts
```diff
@@ -91,10 +92,15 @@ function buildDetoxTargets(
   const targets: Record<string, TargetConfiguration> = {
     [options.buildTargetName]: {
       command: `detox build`,
+      dependsOn: [], // it does not depend on the build the app target
```
ndcunningham, line 95: If this is not set does it mean it will depend on the `buildTarget`?

ndcunningham (approved):

packages/detox/src/plugins/plugin.ts
xiongemi: yes, actually setting dependsOn here does not do anything

PR #21218, opened 2024-01-18
fix(graph): refresh pdv periodically in watch mode

MaxKless:
Just like in the graph mode, we check whether the project graph every 5 seconds to see if anything has changed.
We compare hashes between the old and new project graph and if anything has changed, we update the PDV while keeping the expanded views & scroll position.


FrozenPandaz (changes requested):
This looks amazing!

1. Can we have `watch: true` be the default when we run `nx show project test --web`
2. Can we make the interval faster for the project details page? But keep it a bit longer for the graph?

FrozenPandaz (approved):

PR #21223, opened 2024-01-18
feat(release): global stageChanges option & changelog fixes

fahslaj:
## Current Behavior

- nx release cannot stage changes from the changelog command if committing is disabled
- nx release errors on first run because changelog can't determine --from ref
- nx release errors when all changelogs are disabled
- nx release top level git configuration can be mixed with granular git configuration

## Expected Behavior

- nx release can stage changes from the changelog command without committing
- stageChanges can be set globally in git config
- the --first-release option indicates to the changelog command that it should look up a --from ref based on commit history
- the --first-release option indicates to the publish command that it should not perform a preemptive `npm view`
- nx release no longer errors when all changelogs are disabled
- nx release top level command can only be used with top level git config
- nx release subcommands and programmatic api can only be used with granular git config


docs/generated/cli/release.md
```diff
@@ -25,6 +25,12 @@ Default: `false`
 
 Preview the changes without updating files/creating releases
 
+### first-release
+
+Type: `boolean`
+
+Indicates that this is the first release for the selected release group. If the current version cannot be determined as usual, the version on disk will be used as a fallback. This is useful when using git or the registry to determine the current version of packages, since those sources are only available after the first release. Also indicates that changelog generation should not assume a previous git tag exists and that publishing should not check for the existence the package to publish.
```
JamesHenry, line 32:
```suggestion
Indicates that this is the first release for the selected release group. If the current version cannot be determined as usual, the version on disk will be used as a fallback. This is useful when using git or the registry to determine the current version of packages, since those sources are only available after the first release. Also indicates that changelog generation should not assume a previous git tag exists and that publishing should not check for the existence of the package before running.
```

docs/generated/cli/release.md
```diff
@@ -177,7 +171,7 @@ Exact version or semver keyword to apply to the selected release group.
 
 Type: `boolean`
 
-Whether or not to stage the changes made by this command. Useful when combining this command with changelog generation.
+Whether or not to stage the changes made by this command. Assumed to be true if git-commit is true.
```
JamesHenry, line 174: I think we should be stronger with this language:


```suggestion
Whether or not to stage the changes made by this command. Always treated as true if git-commit is true.
```

JamesHenry (changes requested):
See slack comments

packages/nx/src/command-line/release/changelog.ts
```diff
@@ -209,20 +223,42 @@ export async function releaseChangelog(
...
           (
             await getLatestGitTagForPattern(releaseGroup.releaseTagPattern, {
               projectName: project.name,
             })
           )?.tag;
-        if (!fromRef) {
+
+        let commits: GitCommit[] | null = null;
+
+        if (useAutomaticFromRef) {
```
fahslaj, line 236: `if (useAutomaticFromRef && !fromRef)`

JamesHenry (approved):

rarmatei (approved):

PR #21247, opened 2024-01-19
fix(linter): fixes a bug where the random order of hashes causes wron…

RuvendeGroote:
…g cache mismatches

The hasher returns a list of hashes whose order is random on each run. This causes the overall hash of the run to be different, resulting in cache mismatches. This commit sorts the list of hashes, fixing the issue as the overall hash is matching again.

### DIFF
--- a/packages/eslint/src/executors/lint/hasher.ts
+++ b/packages/eslint/src/executors/lint/hasher.ts
@@ -50,6 +50,10 @@ export default async function run(
       hashes.push(res.details.nodes[d]);
     }
   }
+  // The hashes in res.details are in a random order causing cache mismatches even if the various hashes are equal.
+  // Sorting the hashes prevents this from happening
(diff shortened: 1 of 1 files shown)


RuvenSalamon:
This is my first attempt to contribute, I tried to follow the contribution doc.
I couldn't find out why the list of hashes is random which is the actual root cause of the issue but this fixes it regardless.
Didn't see a spec file for the file I touched so didn't attempt to include tests.

RuvenSalamon:
We have a cache mismatch in the first screenshot and no cache mismatch in the second one (post fix)

Screenshot 2024-01-19 at 18 29 20
Screenshot 2024-01-19 at 18 27 24

RuvenSalamon:
Closing, discovered it was due to nx version mismatches on our side.

PR #21276, opened 2024-01-23
fix(remix): use twStyles as import to prevent conflicts

Coly010:
## Current Behavior

When `@nx/remix:setup-tailwind` is run, we generate `styles` import in the root of the application.

This can conflict with any existing `styles` import.

## Expected Behavior

Add `tw_` prefix to import to indicate more clearly that these styles are related to Tailwind and to prevent any potential conflicts.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/remix/src/generators/setup-tailwind/__snapshots__/setup-tailwind.impl.spec.ts.snap
+++ b/packages/remix/src/generators/setup-tailwind/__snapshots__/setup-tailwind.impl.spec.ts.snap
@@ -31,8 +31,8 @@ exports[`setup-tailwind generator should add a js tailwind config to an applicat
   Scripts,
   ScrollRestoration,
 } from '@remix-run/react';
-import styles from './tailwind.css';
-export const links = () => [{ rel: 'stylesheet', href: styles }];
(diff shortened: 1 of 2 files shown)


packages/remix/src/generators/setup-tailwind/__snapshots__/setup-tailwind.impl.spec.ts.snap
```diff
@@ -31,8 +31,8 @@ exports[`setup-tailwind generator should add a js tailwind config to an applicat
   Scripts,
   ScrollRestoration,
 } from '@remix-run/react';
-import styles from './tailwind.css';
-export const links = () => [{ rel: 'stylesheet', href: styles }];
+import tw_styles from './tailwind.css';
```
jaysoo, line 34: Can we keep it mixed case since it's JS code?

Coly010: Fixed

jaysoo (approved):

PR #21277, opened 2024-01-23
docs(misc): update plugins overview pages for inferred targets

mandarini:
Update overview pages to include inferred targets docs:
- vite
- nuxt
- cypress
- webpack
- storybook
- playwright
- jest
- next
- remix

### DIFF
--- a/docs/generated/packages/cypress/documents/overview.md
+++ b/docs/generated/packages/cypress/documents/overview.md
@@ -1,3 +1,8 @@
+---
+title: Overview of the Nx Cypress Plugin
+description: The Nx Plugin for Cypress contains executors and generators that support e2e testing with Cypress. This page also explains how to configure Cypress on your Nx workspace.
+---
+
 Cypress is a test runner built for the modern web. It has a lot of great features:
 
 - Time travel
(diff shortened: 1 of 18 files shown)


docs/generated/packages/cypress/documents/overview.md
```diff
@@ -7,33 +12,73 @@ Cypress is a test runner built for the modern web. It has a lot of great feature
...
 {% /tab %}
-{% tab label="pnpm" %}
+{% tab label="Nx < 18" %}
+
+Install the `@nx/cypress` package with your package manager and then run the `init` generator.
 
 ```shell
-pnpm add -D @nx/cypress
+npm add -D @nx/cypress
+nx g @nx/cypress:init
```
leosvelperez, line 81: Are we now recommending users run the `init` generators manually? Those generators are internal, and we mark them as `hidden`. I understand we're making them more relevant with PCV3, but that's what `nx add` will do for users, and that's what we'll be recommending for Nx 18+.

Until now, we haven't required users to run the `init` generators manually. That's done automatically by the project generators. If we have decided to start recommending this, then we should remove the `hidden` attribute from these generators.

mandarini: @isaacplmann

PR #21280, opened 2024-01-23
docs(core): v18 tutorials

isaacplmann:
Updates React Standalone, React Monorepo and Vue Standalone tutorials to use PCV3.
Removes the outdated Node Standalone tutorial until we can update it.
The Angular Standalone and Angular Monorepo tutorials are left as is, until we come up with a PCV3 solution for Angular.


docs/shared/react-standalone-tutorial/react-standalone.md
```diff
@@ -120,65 +120,47 @@ Nx uses the following syntax to run tasks:
 
 ![Syntax for Running Tasks in Nx](/shared/images/run-target-syntax.svg)
 
-All targets, such as `serve`, `build`, `test` or your custom ones, are defined in the `project.json` file.
+## Identify Tasks
```
juristr, line 123: I would make this almost a subsection of the "Serving the App". It is like an evolved side-note of why one can just invoke the tasks, but it is not part of the tutorial flow really, which is why I'd make it a subsection.

Should we call this "Inferred Tasks" though?

docs/shared/react-standalone-tutorial/react-standalone.md
```diff
@@ -120,65 +120,47 @@ Nx uses the following syntax to run tasks:
...
+    // ...
+  },
+});
+```
 
-Note that Nx can pick up tasks from both, the `package.json` as well as the `project.json`. [Read more](/features/run-tasks)
+Now if you look at the project details view, the outputs for the build target will say `{projectRoot}/build/myreactapp`. This feature ensures that Nx will always be caching the correct files.
 
-{% /callout %}
+## Overriding Inferred Tasks
```
juristr, line 145: 🤔 I'd wrap this into the "Inferred Tasks" section instead of having a separate one and keep it much shorter. Basically saying you can override the config etc & more details and link to that section in the docs.

I don't want to emphasize all of this too much even before you get started with the actual tutorial. Makes it feel convoluted otherwise I think.

juristr (changes requested):

juristr (changes requested):
Just one final thing. I think we should remove the video and video link buttons for now until we have the updated video recorded.

PR #21287, opened 2024-01-23
fix(vite): PCV3 Plugin update to use resolveConfig

ndcunningham:


packages/vite/src/plugins/plugin.ts
```diff
@@ -15,7 +15,7 @@ import { existsSync, readdirSync } from 'fs';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
 import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
 import { getLockFileName } from '@nx/js';
-import { loadViteDynamicImport } from '../utils/executor-utils';
+import { resolveConfig } from 'vite';
```
mandarini, line 18: i think we want to import it dynamically? so that we don't get the warning? @jaysoo wdyt?

jaysoo: Yeah, it needs to be dynamic so we support Vite ESM, otherwise it'll break in the future (plus I think we may get a warning if going through CJS).

FrozenPandaz:
Victor got this error when running with `vite` in CI:

```
~/p/t/p/test (main|✔) $ rm -rf node_modules/ pnpm-lock.yaml && CI=true pnpm i
Progress: resolved 1, reused 0, downloaded 0, added 0
Progress: resolved 33, reused 33, downloaded 0, added 0
Progress: resolved 89, reused 81, downloaded 0, added 0
Progress: resolved 317, reused 299, downloaded 0, added 0
Progress: resolved 701, reused 649, downloaded 1, added 0
Progress: resolved 1012, reused 960, downloaded 1, added 0
 WARN  2 deprecated subdependencies found: abab@2.0.6, domexception@4.0.0
Packages: +967
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 1017, reused 966, downloaded 1, added 967, done
.../esbuild@0.19.12/node_modules/esbuild postinstall$ node install.js
.../esbuild@0.19.12/node_modules/esbuild postinstall: Done
.../cypress@13.6.3/node_modules/cypress postinstall$ node index.js --exec install
.../cypress@13.6.3/node_modules/cypress postinstall: Cypress 13.6.3 is installed in /home/jason/.cache/Cypress/13.6.3
.../cypress@13.6.3/node_modules/cypress postinstall: Done
.../node_modules/nx postinstall$ node ./bin/post-install
.../node_modules/nx postinstall: The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
.../node_modules/nx postinstall: (node:46908) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
.../node_modules/nx postinstall: (Use `node --trace-deprecation ...` to show where the warning was created)
.../node_modules/nx postinstall: ✘ [ERROR] Could not resolve "/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/nx@17.3.0_@swc-node+register@1.6.8_@swc+core@1.3.105/node_modules/nx/apps/app1/vite.config.ts"
.../node_modules/nx postinstall: failed to load config from /home/jason/projects/triage/pcv3/test/node_modules/.pnpm/nx@17.3.0_@swc-node+register@1.6.8_@swc+core@1.3.105/node_modules/nx/apps/app1/vite.config.ts
.../node_modules/nx postinstall: Unable to create nodes for apps/app1/vite.config.ts using plugin @nx/vite/plugin.
.../node_modules/nx postinstall: 	 Inner Error: Error: Build failed with 1 error:
.../node_modules/nx postinstall: error: Could not resolve "/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/nx@17.3.0_@swc-node+register@1.6.8_@swc+core@1.3.105/node_modules/nx/apps/app1/vite.config.ts"
.../node_modules/nx postinstall:     at failureErrorWithLog (/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1651:15)
.../node_modules/nx postinstall:     at /home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1059:25
.../node_modules/nx postinstall:     at runOnEndCallbacks (/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1486:45)
.../node_modules/nx postinstall:     at buildResponseToResult (/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1057:7)
.../node_modules/nx postinstall:     at /home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1086:16
.../node_modules/nx postinstall:     at responseCallbacks.<computed> (/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:704:9)
.../node_modules/nx postinstall:     at handleIncomingPacket (/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:764:9)
.../node_modules/nx postinstall:     at Socket.readFromStdout (/home/jason/projects/triage/pcv3/test/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:680:7)
.../node_modules/nx postinstall:     at Socket.emit (node:events:514:28)
.../node_modules/nx postinstall:     at addChunk (node:internal/streams/readable:545:12)
.../node_modules/nx postinstall:     at readableAddChunkPushByteMode (node:internal/streams/readable:495:3)
.../node_modules/nx postinstall:     at Socket.Readable.push (node:internal/streams/readable:375:5)
.../node_modules/nx postinstall:     at Pipe.onStreamRead (node:internal/stream_base_commons:190:23)
.../node_modules/nx postinstall: Done
```

We can fix it by passing in the absolute path of the viteConfig. Can we sneak that into this PR please? :pray:

jaysoo (approved):

jaysoo:
It's still not working for this repo https://github.com/jaysoo/vite-app-libs

I think once the config is resolved we still need to check the inputs to make sure they exist?

ndcunningham:
> It's still not working for this repo https://github.com/jaysoo/vite-app-libs
> 
> I think once the config is resolved we still need to check the inputs to make sure they exist?

There was a problem using `resolveConfig` looking through the vite code it merges other configs. So library configs would contain `build` after the resolve. Which is not expected? 

Maybe we can look into that further, But moving forward, using `loadConfigFromFile` respects the config so we can determine if the config is buildable.

packages/vite/src/plugins/plugin.ts
```diff
@@ -235,6 +240,8 @@ function getOutputs(
     'dist'
   );
 
+  const isBuildable = !!build || existsSync(join(projectRoot, 'index.html'));
```
mandarini, line 243: @jaysoo we still want that, right? Or don't we, due to what @Coly010 had said "some vite plugins like qwikVite will alter the entry points and therefore there won’t be an index.html present"?

packages/vite/src/plugins/plugin.ts
```diff
@@ -92,16 +92,20 @@ async function buildViteTargets(
   options: VitePluginOptions,
   context: CreateNodesContext
 ) {
+  const absoluteConfigFilePath = joinPathFragments(
+    context.workspaceRoot,
+    configFilePath
+  );
   const { loadConfigFromFile } = await loadViteDynamicImport();
   const viteConfig = await loadConfigFromFile(
```
Coly010, line 100: Is this still using `loadConfigFromFile` rather than `resolveConfig`?

packages/vite/src/plugins/plugin.ts
Coly010: `build` is coming from `viteConfig` but, `viteConfig` is still using `loadConfigFromFile` and i don't see any mappings to check for `rollup.build`? is there a different approach now?

Coly010: It looks like @ndcunningham has mentioned that `build` should only be truthy if the project is buildable, which vite works out from `loadConfigFromFile`. 

So I think this might be ok? 

But I'm not sure the fallback is needed. Does `vite` not already create `build` if `index.html` is present and it assumes it as an entry point by default?

jaysoo: Vite will try to find the entry point from `index.html` by default unless you or a plugin provides it. I'll double-check this logic, but from what I recall it is fine.

jaysoo (changes requested):
Qwik's config loads as 

```{
  path: '/Users/jack/projects/qwik-app/vite.config.ts',
  config: {
    plugins: [ [Array], [Object], [Object] ],
    server: { headers: [Object] },
    preview: { headers: [Object] }
  },
  dependencies: [ 'vite.config.ts' ]
}
```

We should get this to work because unlike Remix, Qwik expects users to run `vite build`, etc.

Example: https://app.warp.dev/block/xuT3vrlbiAmWrnHvoQK7TF

packages/vite/src/plugins/plugin.ts
ndcunningham: I went back to it since Qwik uses hooks for vite which would be unavailable if i used `loadFromConfigFile`

jaysoo (approved):

mandarini (approved):

PR #21290, opened 2024-01-23
feat(core): support args to be an array for command

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -47,7 +47,7 @@ export interface RunCommandsOptions extends Json {
   readyWhen?: string;
   cwd?: string;
   env?: Record<string, string>;
-  args?: string;
+  args?: string | string[];
   envFile?: string;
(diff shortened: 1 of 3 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -169,14 +168,20 @@ function normalizeOptions(
       typeof c === 'string' ? { command: c } : c
     );
   }
+
+  if (options.args && typeof options.args === 'object') {
```
FrozenPandaz, line 172:
```suggestion
  if (options.args && Array.isArray(options.args)) {
```

FrozenPandaz (approved):

packages/nx/src/executors/run-commands/run-commands.impl.ts
FrozenPandaz: Could you make this small change and then merge it?

xiongemi: changed

PR #21312, opened 2024-01-24
docs(core): add Nx to standalone

isaacplmann:
Update the Adding to existing project guide

### DIFF
--- a/docs/shared/migration/adding-to-existing-project.md
+++ b/docs/shared/migration/adding-to-existing-project.md
@@ -213,7 +213,7 @@ If you want to run one of your existing scripts with Nx, you need to tell Nx abo
 2. Add the script to `includedScripts`.
 3. Define caching settings.
 
-The `nx exec` command allows you to keep using `npm test` or `npm run test` (or other package manager's alternatives) as you're accustomed to. But still get the benefits of making those operations cacheable.  Configuring the `test` script from the example above to run with Nx would look something like this:
+The `nx exec` command allows you to keep using `npm test` or `npm run test` (or other package manager's alternatives) as you're accustomed to. But still get the benefits of making those operations cacheable. Configuring the `test` script from the example above to run with Nx would look something like this:
 
 ```json {% fileName="package.json" %}
 {
--- a/docs/shared/migration/adding-to-existing-project.md
+++ b/docs/shared/migration/adding-to-existing-project.md
@@ -26,7 +26,7 @@ Run the following command:
 npx nx@latest init
 ```
 
-This will set up Nx for you - updating the `package.json` file and creating a new `nx.json` file with Nx configuration based on your answers during the set up process. The set up process will suggest installing Nx plugins that might be useful based on your existing repository. The example below is using the `@nx/eslint` and `@nx/next` plugins to run ESLint and Next tasks with Nx:
+This will set up Nx for you - updating the `package.json` file and creating a new `nx.json` file with Nx configuration based on your answers during the set up process. The set up process will suggest installing Nx plugins that might be useful based on your existing repository. The example below is using the `@nx/eslint` and `@nx/next` plugins to run ESLint and Next.js tasks with Nx:
 
 ```json {% fileName="nx.json" %}
 {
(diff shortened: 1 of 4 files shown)


docs/shared/migration/adding-to-existing-project.md
```diff
@@ -33,110 +28,221 @@ npx nx@latest init
...
 ```
 
-{% callout type="note" title="Use Nx commands directly" %}
+{% /project-details %}
 
-Alternatively you could obviously also just switch to using `nx` for invoking the commands. Like `nx build` rather than `npm run build`.
+The project detail view lists all available tasks, the configuration values for those tasks and where those configuration values are being set.
 
-{% /callout %}
+## Run More Tasks with Nx
```
juristr, line 212: Can we find a more descriptive title for this one?

juristr (changes requested):

docs/shared/migration/adding-to-existing-project.md
```diff
@@ -33,110 +28,221 @@ npx nx@latest init
...
-  ...
-  "devDependencies": {
-    ...
-    "nx": "17.2.0"
-  }
-}
-```
-
-In addition it generates a `nx.json` based on your answers during the setup process. This includes cacheable targets as well as some initial definition of the task pipeline. Here is an example:
+- detect the tools you are using and ask if you want to install plugins for them
```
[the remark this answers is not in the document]
isaacplmann: I can make a "Why Install a Plugin" concept page.  "Why Nx" and other pages can link to it.  It isn't a feature directly, but it enables certain features.

docs/shared/migration/adding-to-existing-project.md
```diff
@@ -23,120 +18,225 @@ You can make these scripts faster by leveraging Nx's caching capabilities. For e
...
+  "name": "my-workspace",
+  ...
+  "scripts": {
+    "build": "nx build",
+    "lint": "nx lint",
+    "test": "node ./run-tests.js"
+  },
+  ...
+  "nx": {
+    "includedScripts": []
```
jaysoo, line 65: Should we explain what `includedScripts` means here or point to a page that explains it (e.g. https://nx.dev/reference/project-configuration#ignoring-packagejson-scripts)?

Since users are expected to run commands through `nx ...`, the `package.json` scripts are just making them more discoverable, but not necessary. And `ignoreScripts` just makes sure Nx doesn't use the build, test, etc. from package scripts.  I dunno how we convey that, or if we need to at all

docs/shared/migration/adding-to-existing-project.md
```diff
@@ -23,120 +18,225 @@ You can make these scripts faster by leveraging Nx's caching capabilities. For e
...
 
-## Wrapping Cacheable Scripts
+The `test` script was not recognized by any Nx plugin, so it was left as is.
 
-Nx also automatically wraps your cacheable scripts with the `nx exec` command. The main advantage here is that you can still keep using `npm start` or `npm run build` (or other package manager's alternatives) as you're accustomed to. But still get the benefits of making those operations cacheble.
+## Inferred Tasks
 
-Here's an example of a `build` and `lint` script being wrapped by Nx:
+You may have noticed that `@nx/next` provides `dev` and `start` tasks in addition to the `build` task. Those tasks were created by the `@nx/next` plugin from your existing Next configuration. To view all available tasks, open the Project Details view with Nx Console or use the terminal to launch the project details in a browser window.
```
jaysoo, line 76: We should refer to Next as "Next.js" since that's how it is referenced in official docs and other content.

docs/shared/migration/adding-to-existing-project.md
```diff
@@ -23,120 +18,225 @@ You can make these scripts faster by leveraging Nx's caching capabilities. For e
...
+      }
     }
-  }
+  ]
 }
 ```
 
-{% /tab %}
-{% tab label="Nx < 17" %}
+When Nx updates your `package.json` scripts, it looks for scripts that can be non-intrusively replaced with an Nx command that has caching automatically enabled. The `package.json` defined above would be updated to look like this:
```
jaysoo, line 52: Maybe we shouldn't say "non-intrusively", just "scripts that can be replaced with an Nx command". It might make the operation seem weirder than it should be.

PR #21330, opened 2024-01-25
feat(misc): exclude package.json scripts only if replaced while running init generators

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/utils/update-package-scripts.spec.ts
+++ b/packages/devkit/src/utils/update-package-scripts.spec.ts
@@ -235,12 +235,12 @@ describe('updatePackageScripts', () => {
     );
   });
 
-  it('should exclude all package.json scripts when none are excluded', async () => {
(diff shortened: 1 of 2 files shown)


jaysoo (approved):

jaysoo:
LGTM

jaysoo:
As explained in the task:

```
After more discussion with @james @jason and @isaac, we've decided to pivot away from this for now. Reason is:

Users will still use npm run (or yarn or pnpm) for certain scripts.

We don't know which scripts should a target in Nx (since we removed the prompt in nx init).

The reason we would want to add to includeScripts is to make is discoverable for the scripts that should be orchestrated through Nx. But we can capture this in the tutorial.

We can revisit this after launch week once we observe usages in the wild.
```

jaysoo:
Will close this for now.

PR #21338, opened 2024-01-25
fix(release): only add nx-release-publish to public packages

JamesHenry:
- Only adds the `nx-release-publish` target when `"private": true` is not set in the package.json of a JS project
- Moves handling of missing `nx-release-publish` target to the `publish` command
- Provides better error handling for missing plugins or generators within plugins for the `version` command


fahslaj (approved):

packages/nx/src/command-line/release/publish.ts
```diff
@@ -198,3 +199,22 @@ async function runPublishOnProjects(
...
+  }
+  output.error({
+    title: `Based on your config, the following projects were matched for publishing but do not have the "${requiredTargetName}" target specified:`,
+    bodyLines: [
+      ...projectsMissingTarget.map((p) => `- ${p.name}`),
+      '',
+      'There are a few possible reasons for this: (1) The projects may be private (2) You may not have an appropriate plugin (such as `@nx/js`) installed which adds the target automatically to public projects (3) You intended to configure the target manually, or exclude those projects via config in nx.json',
+    ],
+  });
+  process.exit(1);
```
FrozenPandaz, line 219: Could this have been thrown as an error? It's quite a lengthy message but since we have a programmatic API, it's weird for their process to suddenly exit after calling the function without a chance to `catch` it?

packages/nx/src/utils/package-json.ts
```diff
@@ -134,16 +134,21 @@ export function buildTargetFromScript(
...
   });
 
-  // Add implicit nx-release-publish target for all package.json files to allow for lightweight configuration for package based repos
-  if (!res['nx-release-publish']) {
+  /**
+   * Add implicit nx-release-publish target for all package.json files that are
+   * not marked as `"private": true` to allow for lightweight configuration for
+   * package based repos.
+   */
+  if (!packageJson.private && !res['nx-release-publish']) {
```
FrozenPandaz, line 151: Do we have to consider `publishConfig` with `access: public`?

I don't think so right? `private` specifies if something can be published and access is only to choose between `restricted` and `public` right? But either way, `publishConfig` cannot disable publishing.

scripts/nx-release.ts
```diff
@@ -116,36 +115,6 @@ const LARGE_BUFFER = 1024 * 1000000;
   if (options.dryRun) {
     console.warn('Not Publishing because --dryRun was passed');
   } else {
-    // If publishing locally, force all projects to not be private first
```
FrozenPandaz, line 119: This is removed because `@nx/js:` handles this?

FrozenPandaz (commented):
I didn't have the chance to test this locally yet but left some initial comments

scripts/nx-release.ts
JamesHenry: No it’s because we agreed that the only workaround for this now was to manually set the targets instead, so I went to do that and realised we don’t have any private projects. If you want me to restore it in case we ever do something like that again in the future I can

packages/nx/src/command-line/release/publish.ts
JamesHenry: Yes fair point, it would block programmatic consumers. We don’t have handleErrors in use in this command so I’ll double check the experience tomorrow

packages/nx/src/utils/package-json.ts
JamesHenry: Yes that’s right

packages/nx/src/command-line/release/publish.ts
JamesHenry: So the way I have decided to handle this is leverage the handleErrors function for the CLI usage like we do for the other subcommands, but additionally have an appreciation for whether or not the code is being run via the CLI.

This allows us to keep the CLI error messaging focused on the publish run (where there are no other config errors), instead of having the runCommand failure and then another error message afterwards just reiterating that publishing has failed.

E.g. what I have gone for 

image

vs

no appreciation for how it's being run via the CLI

image

scripts/nx-release.ts
JamesHenry: I have restored the behaviour in case we ever need it again in future, and updated the log message to more explicitly cover the fact that a target will now need to be added manually for the private packages

FrozenPandaz (approved):

PR #21340, opened 2024-01-25
feat(webpack): simplify inferred webpack-cli command

jaysoo:
This PR changes the webpack build/serve command to `webpack-cli build` and `webpack-cli serve` respectively. We don't need to pass the config file since it is standard naming. The `--node-env` option is now passed in `args` rather than command so it is easier to override.

Also, fix an issue with root projects where `{projectRoot}` is invalid for `outputs` plus removes `.cts` and `.mts` files from being picked up since webpack does work with them even if you pass `--config`.

### DIFF
--- a/packages/webpack/src/plugins/plugin.ts
+++ b/packages/webpack/src/plugins/plugin.ts
@@ -8,7 +8,7 @@ import {
   workspaceRoot,
   writeJsonFile,
 } from '@nx/devkit';
-import { basename, dirname, isAbsolute, join, relative } from 'path';
+import { dirname, isAbsolute, join, relative } from 'path';
 import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
 import { WebpackExecutorOptions } from '../executors/webpack/schema';
 import { WebDevServerOptions } from '../executors/dev-server/schema';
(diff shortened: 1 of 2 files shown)


packages/webpack/src/plugins/plugin.ts
```diff
@@ -175,12 +176,14 @@ async function createWebpackTargets(
...
-  } else {
-    return outputPath;
-  }
+  const outputPathRelativeToWorkspaceRoot =
+    outputPath && isAbsolute(outputPath)
+      ? relative(workspaceRoot, outputPath)
+      : outputPath;
+  return projectRoot === '.'
+    ? `{workspaceRoot}/${outputPathRelativeToWorkspaceRoot}`
+    : `{projectRoot}/${outputPathRelativeToWorkspaceRoot}`;
```
mandarini, line 188: is this correct?
we have it the other way round in `vite` and `nuxt`, and and if we're using `relative` maybe it should be `join()`?

jaysoo: Hmm, I was getting some error saying `{projectRoot}` is invalid. Let me check again.

mandarini (approved):

PR #21356, opened 2024-01-26
docs(misc): replace npm i with nx add

mandarini:
Replace `npm i` etc with `nx add`


isaacplmann (changes requested):
Looks good except you made changes to the Next plugin intro page.  Were those intentional?

PR #21416, opened 2024-01-30
fix(module-federation): ensure incorrect tsconfig path mappings are not used #21391

### LINKED ISSUE #21391: Module federation dev server does not calculate shared dependencies correctly when multiple dev remotes are used.
### Current Behavior

Module federation dev server does not calculate shared dependencies correctly when multiple dev remotes are used.

### Expected Behavior

Module federation dev server does calculates shared dependencies correctly when multiple dev remotes are used.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Run a module federation dev server with multiple dev remotes, note that shared dependencies are not correctly shared.

### Nx Report

```shell

   Node   : 18.17.1
   OS     : darwin-arm64
   pnpm   : 8.14.3

   nx                 : 17.2.8
   @nx/js             : 17.2.8
   @nx/jest           : 17.2.8
   @nx/eslint         : 17.2.8
   @nx/workspace      : 17.2.8
   @nx/angular        : 17.2.8
   @nx/devkit         : 17.2.8
   @nx/eslint-plugin  : 17.2.8
   @nx/node           : 17.2.8
   @nx/plugin         : 17.2.8
   @nx/react          : 17.2.8
   @nx/rollup         : 17.2.8
   @nx/storybook      : 17.2.8
   @nx/vite           : 17.2.8
   @nx/web            : 17.2.8
   @nx/webpack        : 17.2.8
   typescript         : 5.2.2
   Community plugins:
   @openadp/tools-plugin : 24.5.0
   @storybook/angular    : 7.6.10
   Local workspace plugins:
   	 @openadp/tools-plugin
```

### Failure Logs

_No response_

### Package Manager Version

pnpm 8.14.3

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

The core problem is the caching of ts path mappings in here https://github.com/nrwl/nx/blob/master/packages/webpack/src/utils/module-federation/typescript.ts#L7

That can cause the calculated path of one project to be used for a totally separate project when they are run in the same process.

Coly010:
## Current Behavior

We cache the TS Path Mappings when building MF apps. However, if multiple builds occur in a single process, an incorrect set of mappings can be used as they are cached.

## Expected Behavior

Make the cache more robust and based on the actual tsconfig path used.

## Related Issue(s)

Fixes #21391

### DIFF
--- a/packages/webpack/src/utils/module-federation/typescript.ts
+++ b/packages/webpack/src/utils/module-federation/typescript.ts
@@ -3,23 +3,31 @@ import { ParsedCommandLine } from 'typescript';
 import { dirname, join } from 'path';
 import { workspaceRoot } from '@nx/devkit';
 
-let tsConfig: ParsedCommandLine;
-let tsPathMappings: ParsedCommandLine['options']['paths'];
+let tsConfig: Map<string, ParsedCommandLine> = new Map();
+let tsPathMappings: Map<string, ParsedCommandLine['options']['paths']> =
+  new Map();
 
 export function readTsPathMappings(
   tsConfigPath: string = process.env.NX_TSCONFIG_PATH ?? getRootTsConfigPath()
 ): ParsedCommandLine['options']['paths'] {
-  if (tsPathMappings) {
-    return tsPathMappings;
+  if (tsPathMappings.has(tsConfigPath)) {
+    return tsPathMappings.get(tsConfigPath);
   }
 
-  tsConfig ??= readTsConfiguration(tsConfigPath);
-  tsPathMappings = {};
-  Object.entries(tsConfig.options?.paths ?? {}).forEach(([alias, paths]) => {
-    tsPathMappings[alias] = paths.map((path) => path.replace(/^\.\//, ''));
-  });
+  if (!tsConfig.has(tsConfigPath)) {
+    tsConfig.set(tsConfigPath, readTsConfiguration(tsConfigPath));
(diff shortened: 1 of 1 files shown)


packages/webpack/src/utils/module-federation/typescript.ts
```diff
@@ -4,22 +4,26 @@ import { dirname, join } from 'path';
...
 export function readTsPathMappings(
   tsConfigPath: string = process.env.NX_TSCONFIG_PATH ?? getRootTsConfigPath()
 ): ParsedCommandLine['options']['paths'] {
-  if (tsPathMappings) {
-    return tsPathMappings;
+  if (tsPathMappings.has(tsConfigPath)) {
+    return tsPathMappings.get(tsConfigPath);
   }
 
   tsConfig ??= readTsConfiguration(tsConfigPath);
```
statop, line 17: You need to not cache this too.

Coly010: This is caching by path now? Unless you see some issue with it?

it’ll take the path to the tsconfig from the build process and use that to build and entry in the cache

Coly010: Oh wait I see it. Yes good catch

jaysoo (approved):

PR #21433, opened 2024-01-30
fix(graph): fix expand when using external api

xiongemi:
## Current Behavior

the flow is:
- when page is initialized -> triggers useLayoutChange -> collapse and expand sections based on search params
- when external api got call -> search params change -> triggers useLayoutChange -> collapse and expand sections based on search params
- when click on the section -> going to call collapse and expand handler -> setSearchParams -> still triggers useLayoutChange, but should not do anything

remove isExpandedHandled because it prevents external api triggers the useLayoutChange

## Related Issue(s)

Fixes #

### DIFF
--- a/graph/client/src/app/external-api-impl.ts
+++ b/graph/client/src/app/external-api-impl.ts
@@ -81,10 +81,8 @@ export class ExternalApiImpl extends ExternalApi {
     }
   }
 
-  async openProjectDetails(projectName: string, targetName?: string) {
-    await this.router.navigate(
-      `/project-details/${encodeURIComponent(projectName)}`
-    );
+  openProjectDetails(projectName: string, targetName?: string) {
+    this.router.navigate(`/project-details/${encodeURIComponent(projectName)}`);
     if (targetName) {
       this.focusTargetInProjectDetails(targetName);
     }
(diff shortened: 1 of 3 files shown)


graph/client/src/app/external-api-impl.ts
```diff
@@ -81,23 +81,13 @@ export class ExternalApiImpl extends ExternalApi {
...
+  openProjectDetails(projectName: string, targetName?: string) {
+    const pathname = `/project-details/${encodeURIComponent(projectName)}`;
+    const searchParams = new URLSearchParams();
     if (targetName) {
-      this.focusTargetInProjectDetails(targetName);
+      searchParams.set('expanded', targetName);
     }
-  }
-
-  focusTargetInProjectDetails(targetName: string) {
```
MaxKless, line 93: I know this isn't used anywhere right now, but in the future we will need this for smooth handling in Nx Console. So it would be better if we kept the separate `focusTargetInProjectDetails` method

MaxKless (changes requested):

graph/client/src/app/external-api-impl.ts
xiongemi: change it back

FrozenPandaz (approved):

MaxKless (approved):
works, thank you :)

PR #21485, opened 2024-02-01
fix(linter): ensure flat config generator works for pcv3 plugin

meeroslav:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint/src/generators/convert-to-flat-config/generator.spec.ts
+++ b/packages/eslint/src/generators/convert-to-flat-config/generator.spec.ts
@@ -1,6 +1,7 @@
 import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
 import {
   NxJsonConfiguration,
+  ProjectConfiguration,
   Tree,
   addProjectConfiguration,
   readJson,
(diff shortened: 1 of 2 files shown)


packages/eslint/src/generators/convert-to-flat-config/generator.spec.ts
```diff
@@ -430,4 +431,63 @@ describe('convert-to-flat-config generator', () => {
...
+    updateJson(
+      tree,
+      'libs/test-lib/project.json',
+      (json: ProjectConfiguration) => {
+        delete json.targets.lint;
+        return json;
+      }
+    );
+
+    await convertToFlatConfigGenerator(tree, options);
```
JamesHenry, line 456: Please could we add an assertion just before that the flat config doesn't already exist? Verifying that mentally requires taking in the whole rest of the file right now

packages/eslint/src/generators/convert-to-flat-config/generator.spec.ts
```diff
@@ -430,4 +431,63 @@ describe('convert-to-flat-config generator', () => {
...
+    });
+    updateJson(
+      tree,
+      'libs/test-lib/project.json',
+      (json: ProjectConfiguration) => {
+        delete json.targets.lint;
+        return json;
+      }
+    );
+
```
JamesHenry, line 488: Same comment

packages/eslint/src/generators/convert-to-flat-config/generator.ts
```diff
@@ -102,13 +104,20 @@ function convertProjectToFlatConfig(
         }
         updateProjectConfiguration(tree, project, projectConfig);
       }
-      const nxHasLintTargets = Object.keys(nxJson.targetDefaults || {}).some(
+      const nxHasEsLintTargets = Object.keys(nxJson.targetDefaults || {}).some(
         (t) =>
           (t === '@nx/eslint:lint' ||
-            nxJson.targetDefaults[t].executor === '@nx/eslint:lint') &&
+            nxJson.targetDefaults[t].executor === '@nx/eslint:lint' ||
+            nxJson.targetDefaults[t].command.includes('eslint ')) &&
```
JamesHenry, line 111: Same comment

packages/eslint/src/generators/convert-to-flat-config/generator.ts
```diff
@@ -88,7 +88,9 @@ function convertProjectToFlatConfig(
   if (eslintFile && !eslintFile.endsWith('.js')) {
     if (projectConfig.targets) {
       const eslintTargets = Object.keys(projectConfig.targets || {}).filter(
-        (t) => projectConfig.targets[t].executor === '@nx/eslint:lint'
+        (t) =>
+          projectConfig.targets[t].executor === '@nx/eslint:lint' ||
+          projectConfig.targets[t].command.includes('eslint ')
```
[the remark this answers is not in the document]
meeroslav: I thought that having just `eslint` might have some unexpected things caught. But I guess we can solve that once it happens.

packages/eslint/src/generators/convert-to-flat-config/generator.ts
meeroslav: Same as above

JamesHenry (approved):

PR #21490, opened 2024-02-01
fix(testing): set devServerTarget for vite and webpack plugin

xiongemi:
## Related Issue(s)

Fixes #


FrozenPandaz (changes requested):
We should skip the error and then handle it later. It would be better to start from scratch again.

PR #21491, opened 2024-02-01
fix(release): default to 'patch' for --first-release with conventiona…

nkalinov:
…l commits without relevant commits or previous versions

## Current Behavior
Setting up the first release for repos that don't follow conventional commits yet (but want to) OR only have `chore:` commits (or similar "non-release" commits) will fail with:
```
> npx nx release --first-release --dry-run 
...
foo 🚫 No changes were detected using git history and the conventional commits standard.
foo 🚫 Skipping versioning "foo" as no changes were detected.
```
Wasn't immediately obvious (to me at least) what's wrong.

## Expected Behavior
I know this can be worked around by rebasing or pushing new conventional commit but I figured I'll take the opportunity to get some exposure to this awesome project and make a quick contribution.
Feel free to disagree of course if this default doesn't make sense to you.
Basically the change is to set a default `specifier` to `patch` on such first releases:
```
> npx nx release --first-release --dry-run 
...
UPDATE libs/foo/package.json [dry-run]

    "name": "foo",
-   "version": "0.0.1",
+   "version": "0.0.2",
```

## Related Issue(s)
- https://github.com/nrwl/nx/pull/21320


JamesHenry:
Thank you for the kind words and contribution @nkalinov but I don't think we want to move in this direction.

`--first-release` is just a flag that is imperatively passed by you. We are not able to offer any guarantees that it is truly the first release, and it relaxes a number of restrictions. If we behaved how you describe here it is likely to lead to users inadvertently continuing to use it with conventional-commits and continue to apply inappropriate patches, thereby negating the value/purpose of conventional commits.

The error messages seems relatively clear to me, your git history does not align with the conventional commits standard right now, so the solution is to either add a commit which does, or imperatively pass any valid specifier, such as `patch` when you invoke the command.

With hindsight and now understanding the issue better, is there anything that you feel could be improved about the error message?

I will close this in the meantime for clarity. Thanks again!

PR #21498, opened 2024-02-01
fix(core): wait for target output to be flushed before exiting

rluvaton:
## Current Behavior
When we run `nx run lint --lintFilePatterns=...` in `lint-staged` a lot of time the output of the lint is truncate, this cause the find process of error really hard

the reason it's happening after debugging is because the `run-executor.ts` file, exits immidetly after receiving the result from the target, this cause some of the terminal output written to nx to be lost

> Calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, **including I/O operations to process.stdout and process.stderr.**
>
> From [Node.js docs - `process.exit([code])`](https://nodejs.org/docs/latest-v20.x/api/process.html#processexitcode:~:text=Calling%20process.exit()%20will%20force%20the%20process%20to%20exit%20as%20quickly%20as%20possible%20even%20if%20there%20are%20still%20asynchronous%20operations%20pending%20that%20have%20not%20yet%20completed%20fully%2C%20including%20I/O%20operations%20to%20process.stdout%20and%20process.stderr.)

## Expected Behavior

To have the full output printed out


rluvaton:
The test failures are due to permission error, so it's unrelated 

If you have an idea how can I test it as it's a race condition

rluvaton:
Can someone please review the tiny PR?

@FrozenPandaz

mplewis-pge:
I am also encountering this issue. Is it possible to have this PR reviewed and merged?

jaysoo:
We'll have to be very careful about changing `process.exit()` to that poorly managed processes do not cause Nx to hang indefinitely. There is another PR that awaits for streams to finish before exiting, which is a bit safer, but we had to revert it.

PR #21509, opened 2024-02-01
fix(core): Update Nx Welcome to include nx show project

ndcunningham:
Update NxWelcome component to include `nx show project { project-name } --web` 

Sample

Web:
Screenshot 2024-02-01 at 3 57 34 PM

Mobile:
Screenshot 2024-02-01 at 3 56 10 PM


Coly010:
We should have the project name in the generator code, and it might already even be passed to `generateFiles`. I think we should use that such that it shows `nx show myapp —web`

Coly010 (approved):

jaysoo (approved):

PR #21510, opened 2024-02-02
fix(release): move github release creation to git tasks

fahslaj:
## Current Behavior

`nx release` fails when configured to create a github release because the release is created before changes are committed and tagged. This results in duplicate git tags being created and an error.

## Expected Behavior

`nx release` properly creates a github release after committing, tagging, and pushing changes. `nx release changelog` is unaffected.


JamesHenry (changes requested):
The programmatic API example docs need to updated to no longer treat the return value of releaseChangelog as a status code, and @juristr will need to update his recent blog post once this is released

JamesHenry (approved):

isaacplmann (approved):

PR #21531, opened 2024-02-02
fix(webpack): fix webpack plugin issues and e2e tests

jaysoo:
This PR fixes the two failing webpack e2e tests One test required two small fixes:

1. `NxWebpackPlugin` - Should not override user's `config.target` with `'web'`
2. `@nx/webpack:configuration`- Generator had wrong `output.path` and was nested it under `<proj>/dist/<proj>` rather than `dist/<proj>`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/webpack/src/generators/configuration/configuration.ts
+++ b/packages/webpack/src/generators/configuration/configuration.ts
@@ -1,7 +1,10 @@
 import {
   formatFiles,
+  GeneratorCallback,
   joinPathFragments,
+  offsetFromRoot,
   readProjectConfiguration,
+  runTasksInSerial,
   Tree,
   updateProjectConfiguration,
   writeJson,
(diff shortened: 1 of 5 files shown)


packages/webpack/src/generators/configuration/configuration.ts
```diff
@@ -94,14 +105,15 @@ const { join } = require('path');
 
 module.exports = {
   output: {
-    path: join(__dirname, '${buildOptions.outputPath}'),
+    path: join(__dirname, '../../${buildOptions.outputPath}'),
```
mandarini, line 108: should this be hard-coded to `../../`?

mandarini (approved):
approved, just one comment

Coly010 (approved):

PR #21533, opened 2024-02-02
fix(core): always ignore temp vite config files for the watcher

Cammisuli:
## Current Behavior

When using vite, a temporary file gets created that causes the daemon always retrigger

## Expected Behavior
Always ignore this temporary vite file.

## Related Issue(s)

Fixes #


Cammisuli:
closing for https://github.com/nrwl/nx/pull/21514

PR #21534, opened 2024-02-02
fix(nest): ensure dependencies are installed when generating nest libraries

jaysoo:
When a Node lib is generated without an app, we don't ensure that dependencies are installed. This PR adds those installs to the lib generator.

In practice, this should not happen since the workspace must have an app if it is adding a lib, but for correctness we need the ensure.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nest/src/generators/application/application.ts
+++ b/packages/nest/src/generators/application/application.ts
@@ -5,12 +5,12 @@ import { applicationGenerator as nodeApplicationGenerator } from '@nx/node';
 import { initGenerator } from '../init/init';
 import {
   createFiles,
-  ensureDependencies,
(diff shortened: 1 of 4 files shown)


e2e/node/src/node.test.ts
```diff
@@ -593,7 +593,7 @@ ${jslib}();
       );
     }, 60000);
 
-    it('should be able to generate a nest library w/ service', async () => {
+    it('xyz should be able to generate a nest library w/ service', async () => {
```
mandarini, line 596: should we remove the `xyz`?

mandarini (approved):

barbados-clemens (approved):

PR #21556, opened 2024-02-02
fix(workspace): explicitly call `complete-ci-run` in generated CI config

StalkAltan:
We will remove this for 18.1, when heartbeat should be released

## Current Behavior
Generated workflow configurations will not report status to cloud correctly

## Expected Behavior
Generated workflow configurations will report status to cloud correctly


FrozenPandaz (approved):

FrozenPandaz (changes requested):
Talked with Victor, this won't be needed soon

PR #21559, opened 2024-02-02
fix(expo): set port for e2e

xiongemi:
## Current Behavior

need to set the expo serve port number for e2e testing

## Related Issue(s)

Fixes #


packages/expo/plugins/plugin.ts
```diff
@@ -110,7 +110,7 @@ function buildExpoTargets(
       options: { cwd: projectRoot },
     },
     [options.serveTargetName]: {
-      command: `expo start --web`,
+      command: `expo start --web --port 4200`,
```
ndcunningham, line 113: I think you should change the port inside e2e to match the default port of expo

ndcunningham (approved):

PR #10538, opened 2022-05-31
fix(devkit): add options allow trailing commas for parsing json

FrozenPandaz:
## Current Behavior

`readJson` and `updateJson` do not support json files with trailing commas.

## Expected Behavior

There is now an option for both methods to support reading json files with trailing commas.

```ts
readJson(tree, 'tsconfig.json', {
  allowTrailingCommas: true
});

updateJson(tree, 'tsconfig.json', json => json, {
  allowTrailingCommas: true
});
```

## Related Issue(s)

Fixes #

PR #5859, opened 2021-06-01
feat(core): first step to common JSON behavior and common fileutils

Phillip9587:
## Current Behavior
- Nx repo contains many different ways to parse and serialize JSON (see #5631)
- Nx defines many different ffileutils

## Expected Behavior
- Nx should use a common way to parse and serialize JSON

# What has been done

I created a `utils` directory in `@nrwl/tao`. I defined a `parseJson` and a `serializeJson` method in `json.ts`. Both methods support options to customize the parse or serialization process if needed. I also switched from `strip-json-comments` to [`jsonc-parser`](https://github.com/microsoft/node-jsonc-parser) because of the options and nice api the [`jsonc-parser`](https://github.com/microsoft/node-jsonc-parser) package provides. 

The error messages due to JSON parse errors were also improved with the switch to [`jsonc-parser`](https://github.com/microsoft/node-jsonc-parser). See: https://github.com/microsoft/node-jsonc-parser/blob/ff813e9f741b2b35f8bea2dd7022a3a43156d04a/src/main.ts#L149-L166

I also create a `fileutils.ts` file in the `@nrwl/tao` utils which exports the `readJsonFile` and `writeJsonFile` methods which support the options from `parseJson` and `serializeJson`. 

All 4 methods and the options interfaces are reexported from `@nrwl/devkit` to simplify the imports and usage. 

I also started to use the methods in places where changes where needed due to the removal of the `strip-json-comments` package. 

If this PR gets merged i will open further PRs to cleanup the JSON parse an serialize behaviors and also one to cleanup the fileutils mess. 

In another step the https://eslint.org/docs/rules/no-restricted-globals rule could be enabled for `JSON.parse` and `JSON.stringify` to warn developers and get a consistent JSON parse and serialization behavior.

## Related Issue(s)
#5631

PR #21584, opened 2024-02-04
fix(devkit): respect expectComments when parsing json

llwt:
[This change](https://github.com/nrwl/nx/pull/10538/files#diff-7f0343ab5195423aded5b2ed7b98cef5fd459edbb66227eb67f0c0fa25eaf3a4L40-L50) introduced a small performance regression by removing the check for the `expectComments` flag.

## Current Behavior

Passing `expectComments` to `parseJson` does nothing.

## Expected Behavior

Passing `expectComments` to `parseJson` should bypass the optimistic `JSON.parse` since in those cases it is expected to fail.

## Related Issue(s)

* Initially introduced here: https://github.com/nrwl/nx/pull/5859/files#diff-1a98443b5c509860939d182ca80607c7150f119241722dfb7632c7145450e279R40-R45
* Regression here: https://github.com/nrwl/nx/pull/10538/files#diff-7f0343ab5195423aded5b2ed7b98cef5fd459edbb66227eb67f0c0fa25eaf3a4L40-L50


llwt:
@FrozenPandaz Do you need an issue for this? Or is a PR sufficient?

AgentEnder (approved):

PR #21593, opened 2024-02-05
fix(core): add missing prettier dependency

meeroslav:
Prettier is used in format command and there fore must be explicitly specified in the package.json

cc @jaysoo 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -55,6 +55,7 @@
     "minimatch": "9.0.3",
     "npm-run-path": "^4.0.1",
     "open": "^8.4.0",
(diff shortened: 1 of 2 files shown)


packages/nx/package.json
```diff
@@ -55,6 +55,7 @@
     "minimatch": "9.0.3",
     "npm-run-path": "^4.0.1",
     "open": "^8.4.0",
+    "prettier": "^2.6.2",
```
Coly010, line 58: Is there a reason this uses `2.6.2` when latest version of prettier is `3.2.5`?

PR #21608, opened 2024-02-05
docs(core): update nx node matrix

isaacplmann:
Include Nx 18 in the nx/node/typescript matrix

### DIFF
--- a/docs/generated/packages/workspace/documents/nx-nodejs-typescript-version-matrix.md
+++ b/docs/generated/packages/workspace/documents/nx-nodejs-typescript-version-matrix.md
@@ -11,5 +11,6 @@ and the version of NodeJS that we tested it against.
 | 13.x            | 10.x, 12.x, 14.x | ~4.6.2             |
 | 14.x            | 12.x, 14.x, 16.x | ~4.7.2             |
 | 15.x            | 14.x, 16.x, 18.x | ~5.0.0             |
-| 16.x (previous) | 16.x, 18.x, 20.x | ~5.1.0             |
-| 17.x (latest)   | 18.x, 20.x       | ~5.1.0             |
+| 16.x            | 16.x, 18.x, 20.x | ~5.1.0             |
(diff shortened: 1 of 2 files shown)


AgentEnder (approved):

docs/generated/packages/workspace/documents/nx-nodejs-typescript-version-matrix.md
```diff
@@ -11,5 +11,6 @@ and the version of NodeJS that we tested it against.
 | 13.x            | 10.x, 12.x, 14.x | ~4.6.2             |
 | 14.x            | 12.x, 14.x, 16.x | ~4.7.2             |
 | 15.x            | 14.x, 16.x, 18.x | ~5.0.0             |
-| 16.x (previous) | 16.x, 18.x, 20.x | ~5.1.0             |
```
FrozenPandaz, line 14: I actually kept 16 as previous because it's probably too early for them to move to 17 which would have more breaking changes.

The plan is to make 18 the previous version when Nx 19 comes out. 17 will be skipped over in terms of becoming previous.

FrozenPandaz (changes requested):

docs/generated/packages/workspace/documents/nx-nodejs-typescript-version-matrix.md
FrozenPandaz: Sorry :facepalm: forgive me.

It turns out it did get updated to 17 when I did the `17.3.2` release :thinking: 

Fine let's go with 17 being previous... I think it makes sense either way.... It was a bit weird when latest is 18 but previous is 16...

FrozenPandaz (changes requested):
Sorry, I was wrong about 16 remaining previous... 

Can you change it back to what you initially had please? :pray:

isaacplmann:
> Can you change it back to what you initially had please? 🙏

Done

FrozenPandaz (approved):

PR #21631, opened 2024-02-05
docs(core): remove reference to deno

Jordan-Hall:
as stated in the package https://www.npmjs.com/package/@nx/deno  The Deno plugin is deprecated and will no longer receive updates. We are committed to providing high-quality tooling to community, and we no longer have the capacity to keep this plugin updated.

I thought Deno was still supported until I noticed the code gone and package saying its unsupported. However the website is still showing it as supported


Jordan-Hall:
I don't believe the tests is actually valid anymore, i think it was only on the deno page (from looking around the docs). Happy to remove it

isaacplmann:
Thanks for the PR!  Yes, we should remove these recipes since we're not supporting the deno plugin any more.  I re-added the test that you removed since it wasn't specifically testing the deno package - just using the deno page as an example.

isaacplmann (approved):

PR #21646, opened 2024-02-06
fix(vite): project conversion generator

mandarini:
Cleanup for the `@nx/vite:configuration` generator, the ability to convert an existing project to use vite. 

The generator is outdated and needs to be updated to the new format.


packages/vite/src/utils/generator-utils.ts
```diff
@@ -86,81 +59,22 @@ export function findExistingTargetsInProject(
     '@nrwl/react-native:build-android',
     '@nrwl/react-native:bundle',
     '@nrwl/next:build',
-    '@nrwl/next:server',
     '@nrwl/js:tsc',
     '@angular-devkit/build-angular:browser',
-    '@angular-devkit/build-angular:dev-server',
+    '@angular-devkit/build-angular:browser-esbuild',
```
jaysoo, line 64: Do we still need `@angular-devkit/build-angular:dev-server`? What's the effect of moving this?

mandarini: We are not looking for `serve` executors, just `build` executors, and it's a list to identify unsupported build executors. 

in the previous example we would want to update the serve/build/test targets on `project.json`, but now we don't update any targets, so we don't care about identifying unsupported  serve targets.

we are convertings apps that have webpack.config, and libs that use js executors or rollup, or are non-buildable.

isaacplmann (approved):

jaysoo (approved):

PR #21650, opened 2024-02-06
docs(webpack): Add docs for webpack ci deployment

ndcunningham:
### Changes
Now the Project Crystal is enabled by default we need to revise our webpack documentation. 

[Preview](https://nx-dev-git-docs-ci-deployment-webpack-nrwl.vercel.app/ci/recipes/other/ci-deployment)

### DIFF
--- a/docs/shared/recipes/ci-deployment.md
+++ b/docs/shared/recipes/ci-deployment.md
@@ -6,6 +6,69 @@ Nx supports the generation of the project's `package.json` by identifying all th
 
 Additionally, we should generate pruned lock file according to the generated `package.json`. This makes the installation in the container significantly faster as we only need to install a subset of the packages.
 
+Nx offers two varieties of Webpack plugin which can be used to generate `package.json`.
+
+{% tabs %}
+
+{% tab label="Nx 18+" %}
+
+## Basic Plugin Configuration
+
+`@nx/webpack/plugin` plugin is compatible with a conventional webpack configuration setup which offers a smooth integration with the Webpack CLI.
+It is configured in the `plugins` array in `nx.json`.
+
+```json {% fileName="nx.json" %}
+{
+  "plugins": [
+    {
+      "plugin": "@nx/webpack/plugin",
+      "options": {
+        "buildTargetName": "build",
+        "serveTargetName": "serve",
+        "serveStaticTargetName": "serve-static",
+        "previewStaticTargetName": "preview"
+      }
+    }
+  ]
+}
+```
+
+Where `build`, `serve`, `serve-static` and `preview` in conjunction with your `webpack.config.js` are the names of the targets that are used to _build_, _serve_, and _preview_ the application respectively.
+
+### NxWebpackPlugin
+
+The [`NxWebpackPlugin`](/recipes/webpack/webpack-plugins#nxwebpackplugin) plugin takes a `main` entry file and produces a bundle in the output directory as defined in `output.path`. You can also pass the `index` option if it is a web app, which will handle outputting scripts and stylesheets in the output file.
+
+To generate a `package.json` we would declare it in the plugin options.
+
+```js {% fileName="apps/acme/app/webpack.config.js" %}
+const { NxWebpackPlugin } = require('@nx/webpack');
+const { join } = require('path');
+
+module.exports = {
+  output: {
+    path: join(__dirname, '../../dist/apps/acme'),
+  },
+  devServer: {
+    port: 4200,
+  },
+  plugins: [
+    new NxWebpackPlugin({
+      tsConfig: './tsconfig.app.json',
+      compiler: 'swc',
+      main: './src/main.tsx',
+      index: '.src/index.html',
+      styles: ['./src/styles.css'],
+      generatePackageJson: true,
+    }),
+  ],
+};
+```
+
+{% /tab %}
+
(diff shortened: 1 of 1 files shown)


docs/shared/recipes/ci-deployment.md
```diff
@@ -16,12 +79,95 @@ Some executors automatically generate output `package.json` and the lock file ge
...
+
+main();
+```
+
+Then to run the script, update your `package.json` to include the following:
+
+```json {% fileName="package.json" %}
+{
+  "scripts": {
+    "build": "node scripts/custom-build.js"
```
jaysoo, line 156: Maybe this should be `custom-build` rather than `build` to differentiate between build targets that could be inferred or configured in `project.json`?

docs/shared/recipes/ci-deployment.md
```diff
@@ -16,12 +79,95 @@ Some executors automatically generate output `package.json` and the lock file ge
...
+  const lockFileName = getLockFileName(pm);
+
+  writeFileSync(`${outputDir}/package.json`, packageJson);
+  writeFileSync(`${outputDir}/${lockFileName}`, lockFile, {
+    encoding: 'utf8',
+  });
+
+  //... Any additional build steps you want to run
+
+  exec(`${pm.exec} nx build`, (err, stdout, stderr) => {
```
jaysoo, line 139: I don't think we need call the `nx build` here, users can do `nx build && npm run custom-script`. Note that most `build` commands will also clear dist first, so copy definitely needs to be second.

docs/shared/recipes/ci-deployment.md
jaysoo: Or maybe just `copy-package-json` script name?

ndcunningham: it is `custom-build`

docs/shared/recipes/ci-deployment.md
ndcunningham: I wanted to show that you could run it solely from script. Since `nx build && nx copy-packagejson` is usually common knowledge.

docs/shared/recipes/ci-deployment.md
```diff
@@ -16,12 +79,87 @@ Some executors automatically generate output `package.json` and the lock file ge
 - `@nx/js:tsc`
 - `@nx/next:build`
 
-## Using a custom executor
+{% /tab %}
+{% /tabs %}
+
+## Programmtic usage
+
+If you using a custom setup that does not support the creation of a `package.json` or a lock file, you can still use Nx to generate them via The `createPackageJson` and `createLockFile` functions which are exported from `@nx/js`:
```
isaacplmann, line 87:
```suggestion
If you are using a custom setup that does not support the creation of a `package.json` or a lock file, you can still use Nx to generate them via The `createPackageJson` and `createLockFile` functions which are exported from `@nx/js`:
```

docs/shared/recipes/ci-deployment.md
```diff
@@ -42,11 +180,18 @@ export default async function buildExecutor(
...
   });
 
   // any subsequent executor code
 }
 ```
+
+{% /tab %}
+{% /tabs %}
+
+{% callout type="note" title="What about Vite ?" %}
```
isaacplmann, line 195:
```suggestion
{% callout type="note" title="What about Vite?" %}
```

isaacplmann (changes requested):
Small changes.

Why do you say "we are working on a plugin for Vite"?  `@nx/vite` already exists.  Does it not have parity with `@nx/webpack` yet?  What does it not do?

ndcunningham:
> Small changes.
> 
> Why do you say "we are working on a plugin for Vite"? `@nx/vite` already exists. Does it not have parity with `@nx/webpack` yet? What does it not do?

I think the plugin being referred to here is `NxWebpackPlugin` so it probably would be called `NxVitePlugin` which would contain some of the already existing functionality `nxViteTsPaths()` and also functionality that does not exists for vite _generate `package.json`_.

cc: @jaysoo

docs/shared/recipes/ci-deployment.md
```diff
@@ -42,11 +180,18 @@ export default async function buildExecutor(
...
 
   // any subsequent executor code
 }
 ```
+
+{% /tab %}
+{% /tabs %}
+
+{% callout type="note" title="What about Vite?" %}
+Vite is a build tool that is great for development, and we want to make sure that it is also great for production. We are working on a plugin for Vite that will have parity with the Webpack plugin. Stay tuned for updates.
```
isaacplmann, line 196:
```suggestion
Vite is a build tool that is great for development, and we want to make sure that it is also great for production. We are working on an `NxVitePlugin` plugin for Vite that will have parity with the `NxWebpackPlugin`. Stay tuned for updates.
```

isaacplmann (changes requested):

isaacplmann (approved):

isaacplmann:
> I think the plugin being referred to here is `NxWebpackPlugin` so it probably would be called `NxVitePlugin` which would contain some of the already existing functionality `nxViteTsPaths()` and also functionality that does not exists for vite _generate `package.json`_.
> 
> cc: @jaysoo

I updated the text to make this more explicit and approved the PR.

jaysoo (approved):

PR #21715, opened 2024-02-08
fix(core): cannot read property kind of undefined

maxence-lefebvre:
## Current Behavior

When running `nx g mv ...`, if an updated files exports don't include a moduleSpecifier (e.g `export { MyClass };`), the generator fails with `Cannot read property "kind" of undefined`.

## Expected Behavior

The generator should ignore those exports, as there is no need to update the module specifier if it doesn't exist.

## Related Issue(s)

I didn't open an issue.

I updated the tests, made sure the updated snapshot was correct and tried locally in my own workspace.


FrozenPandaz (approved):
Thank you for your contributions! LGTM :tada: 

Just rebased it and am letting CI run.

PR #21727, opened 2024-02-08
Update define-environment-variables.md

motdotla:
add `dotenvx` example. 

## Current Behavior

* env-cmd only

## Expected Behavior

* include dotenvx example 

## Related Issue(s)
N/A

Fixes #
N/A


isaacplmann (approved):

LuisLopez44431 (commented):
We all trying to do or best you all are doing wander full job I'm taking end other challenge I will let you know

PR #21738, opened 2024-02-09
docs(misc): adding tusky to nx cloud products

ZackDeRose:
## Related Issue(s)

Fixes #


isaacplmann (approved):

juristr (changes requested):
The icons are not aligned with what is on nx.app

image

Ideally I'd also make the tagline shorter

isaacplmann:
Good catch.  I checked that the text matched, but not the icons.  If you want the tag line to be shorter, that should also be updated on nx.app

ZackDeRose:
Updated the icons to match the ones on the marketing site.

juristr (approved):

PR #21746, opened 2024-02-09
fix(testing): jest should handle root jest.preset.cjs

Coly010:
## Current Behavior

Jest Configuration generator only generates `jest.preset.js` files, and only creates `jest.config` files that point to `jest.preset.js`.

If root `package.json` has `type: module` then the `jest.preset` needs to be `.cjs`.

If a `jest.preset.cjs` already exists, instead of using the `jest.preset.cjs` in the `jest.config` files, it will still create and try to point to `jest.preset.js`.

## Expected Behavior

Ensure Jest Configuration supports `jest.preset.js` and `jest.preset.cjs`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/utils/config/is-preset-cjs.ts
+++ b/packages/jest/src/utils/config/is-preset-cjs.ts
@@ -1,0 +1,14 @@
+import { type Tree, readJson } from '@nx/devkit';
+
+export function isPresetCjs(tree: Tree) {
+  if (tree.exists('jest.preset.cjs')) {
+    return true;
+  }
+
+  const rootPkgJson = readJson(tree, 'package.json');
+  if (rootPkgJson.type && rootPkgJson.type === 'module') {
+    return true;
+  }
+
+  return false;
+}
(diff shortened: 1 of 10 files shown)


packages/jest/src/utils/config/is-preset-cjs.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export function isPresetCjs(tree: Tree) {
+  if (tree.exists('jest.preset.cjs')) {
+    return true;
+  }
+
+  const rootPkgJson = readJson(tree, 'package.json');
+  if (rootPkgJson.type && rootPkgJson.type === 'module') {
+    return true;
+  }
```
edbzn, line 11: Not sure it's by design but not all paths return a boolean in this function, there's no default `return false` branch, which could lead to bug if used this way `isPresetCjs(tree) === false`.

Coly010: It’s a good catch, it should be updated.

leosvelperez (approved):

PR #21769, opened 2024-02-10
fix(release): skip prompt for publish when no version created

edbzn:
## Current Behavior

Running `nx release` prompt "Do you want to publish these versions?" even if no version was created during the process.

## Expected Behavior

Skip prompt and publish if no version was created.

## Related Issue(s)
N/A

### DIFF
--- a/packages/nx/src/command-line/release/release.ts
+++ b/packages/nx/src/command-line/release/release.ts
@@ -140,8 +140,17 @@ export async function release(
     }
   }
 
-  let shouldPublish = !!args.yes && !args.skipPublish;
-  const shouldPromptPublishing = !args.yes && !args.skipPublish && !args.dryRun;
+  let hasNewVersion = false;
+  // null means that all projects are versioned together but there were no changes
+  if (versionResult.workspaceVersion !== null) {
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/release/release.ts
```diff
@@ -139,9 +139,15 @@ export async function release(
       });
     }
   }
+  const hasNewVersion =
```
JamesHenry, line 142: `workspaceVersion === null; // workspace version is applicable no changes, no new version required`

`workspaceVersion === undefined; // a unified version for the workspace is not applicable (independent projects) but it says nothing about whether or not a new version is required`

```suggestion
  let hasNewVersion = false;
  // null means that all projects are versioned together but there were no changes
  if (workspaceVersion !== null) {
    hasNewVersion = Object.values(versionResult.projectsVersionData).some(
      (version) => version.newVersion !== null
    );
  }

```

JamesHenry:
Thanks @edbzn! The e2e failure should hopefully go away once you update to latest

JamesHenry (changes requested):

JamesHenry (approved):

PR #21784, opened 2024-02-12
feat(react): add tailwind as style prompt option for app gen

Coly010:
## Current Behavior

We do not prompt the user to use Tailwind as an option when generating a react app either via `nx generate` or via `create-nx-workspace`.

## Expected Behavior

Prompt the user with an option for Tailwind in both scenarios listed above.

## Related Issue(s)

Fixes #


isaacplmann (approved):

jaysoo (approved):
Can you also add an e2e test to make sure build includes the right class names in css?

PR #21805, opened 2024-02-13
fix(nextjs): Add missing e2e-ci target for cypress

ndcunningham:
## Current Behavior

Currently, we have no way of running a static server for Next.js while running `e2e` tests.
Which should help reduce time with running test suites especially on CI.

## Expected Behavior

Have targets to run static files for e2e tests.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/next/plugins/with-nx.ts
+++ b/packages/next/plugins/with-nx.ts
@@ -197,6 +197,12 @@ function withNx(
             : joinPathFragments(outputDir, '.next');
       }
 
+      // If we are running a static serve of the Next.js app, we need to change the output to 'export' and the distDir to 'out'.
+      if (process.env.NX_SERVE_STATIC_BUILD_RUNNING === 'true') {
+        nextConfig.output = 'export';
+        nextConfig.distDir = 'out';
+      }
+
       const userWebpackConfig = nextConfig.webpack;
 
       const { createWebpackConfig } = require('@nx/next/src/utils/config');
--- a/packages/next/src/plugins/plugin.ts
+++ b/packages/next/src/plugins/plugin.ts
@@ -21,6 +21,7 @@ export interface NextPluginOptions {
   buildTargetName?: string;
   devTargetName?: string;
   startTargetName?: string;
+  serveStaticTargetName?: string;
 }
 
 const cachePath = join(projectGraphCacheDirectory, 'next.hash');
@@ -62,7 +63,6 @@ export const createNodes: CreateNodes<NextPluginOptions> = [
     ) {
       return {};
     }
-
     options = normalizeOptions(options);
 
     const hash = calculateHashForCreateNodes(projectRoot, options, context, [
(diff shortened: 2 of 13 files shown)


packages/next/src/plugins/plugin.ts
```diff
@@ -152,6 +155,18 @@ function getStartTargetConfig(options: NextPluginOptions, projectRoot: string) {
   return targetConfig;
 }
 
+function getStaticServeTargetConfig(options: NextPluginOptions) {
+  const targetConfig: TargetConfiguration = {
+    executor: '@nx/web:file-server',
+    options: {
+      buildTarget: options.buildTargetName,
```
xiongemi, line 162: i am looking at the generated project.json with serve target, should the buildTarget be projectName:options.buildTargetName?

packages/next/src/generators/application/lib/add-project.ts
```diff
@@ -63,6 +63,14 @@ export function addProject(host: Tree, options: NormalizedSchema) {
         buildTarget: `${options.projectName}:build:production`,
       },
     };
+  } else {
+    // We need to add the serve-static target so that the port can be configured.
+    targets['serve-static'] = {
```
xiongemi, line 68: it goes to this condition when hasPlugin is true, don't you have the getStaticServeTargetConfig to set the serve-static target?

packages/next/plugins/with-nx.ts
```diff
@@ -197,6 +197,12 @@ function withNx(
             : joinPathFragments(outputDir, '.next');
       }
 
+      // If we are running a static serve of the Next.js app, we need to change the output to 'export' and the distDir to 'out'.
```
xiongemi, line 200: would it lead to unexpected behavior if file server executor got killed in between?

xiongemi (changes requested):

packages/next/src/plugins/plugin.ts
ndcunningham: I am not sure I understand but it follows the same principle as `serveTarget` inside the `plugin.ts`

packages/next/plugins/with-nx.ts
ndcunningham: No it doesn't change the source file and if the file serve executor is killed this is child process and would be killed as well.

xiongemi (approved):

isaacplmann (approved):

packages/next/src/generators/application/lib/add-project.ts
jaysoo: I think we should skip this when plugin is added. Mixing inferred and explicit targets is something users can do themselves, but we should not generate it that way.

packages/next/src/plugins/plugin.ts
jaysoo, line 162: We should specify the port options here as well so users can inspect it via PDV, and know how to override it.

ndcunningham: hmmm wouldn't they just override it inside `project.json` 
This is already being shown in the PDV.

Screenshot 2024-02-15 at 10 08 11 AM

jaysoo (approved):

FrozenPandaz (approved):

PR #21806, opened 2024-02-13
fix(testing): allows getJestProjects to fetch config information from inferred plugins

jamie-wk:
After updating my codebase to use the new plugin inference system, I found that I could no longer run pnpm jest at the root of my repo.

This change allows the getJestProject methods to pull from retrieveProjectConfigurations in order to find all project config files.

## Current Behavior
When running `pnpm jest` at the root of the application, `getJestProjects` returns empty if there is not an explicit `test` target in the project.json

## Expected Behavior
Running `pnpm jest` should run all valid Jest projects in the repository.

This is my first contribution to NX, happy to make any changes necessary and would love feedback!

### DIFF
--- a/packages/jest/src/utils/config/get-jest-projects.spec.ts
+++ b/packages/jest/src/utils/config/get-jest-projects.spec.ts
@@ -1,182 +1,200 @@
+import * as RetrieveWorkspaceFiles from 'nx/src/project-graph/utils/retrieve-workspace-files';
 import { getJestProjects } from './get-jest-projects';
-import * as Workspace from 'nx/src/project-graph/file-utils';
-import type { WorkspaceJsonConfiguration } from '@nx/devkit';
 
 describe('getJestProjects', () => {
-  test('single project', () => {
-    const mockedWorkspaceConfig: WorkspaceJsonConfiguration = {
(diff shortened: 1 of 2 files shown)


leosvelperez (changes requested):

These changes turn the `getJestProjects` function into an async function, so the generated `jest.config.ts` at the workspace root needs to account for that. It's currently [expecting a sync function](https://github.com/nrwl/nx/blob/master/packages/jest/src/generators/configuration/lib/create-jest-config.ts#L129-L144).

Also, these changes don't really address the issue you're trying to solve. The inferred targets for Jest don't use the `@nx/jest:jest` executor, so they wouldn't be collected because in this function, we only look for targets with that executor. Those inferred targets use the `nx:run-commands` executor, and they use the command `jest`. On top of that, users could also overwrite the command in the project configuration to provide some extra args (e.g. `jest --runInBand`).

leosvelperez:
Thanks again for contributing! I'm going to close this in favor of https://github.com/nrwl/nx/pull/21897. We're instead adding a separate function to not break existing usages.

king0120:
Sounds good! Thank you for addressing!

PR #21819, opened 2024-02-14
feat(release): prompt to create github release when no file changes

fahslaj:
## Current Behavior

Nx Release exits with a warning when changelog files and github releases are enabled and there are no changelog file changes. This means that there is no way to update a remote github release if the local changelog file is already up to date.

## Expected Behavior

When changelog files and github releases are enabled and there are no changelog file changes, the end user is prompted to see if they want to create a github release anyway. This allows users to update a remote github release regardless of the state of a local changelog file, like in the case where the remote release was updated in the github UI.


packages/nx/src/command-line/release/changelog.ts
```diff
@@ -550,7 +552,32 @@ async function applyChangesAndExit(
...
+      // no GitHub releases to create so we can just exit
+      return;
+    }
+
+    if (isCI()) {
+      output.warn({
+        title: `Skipped GitHub release creation because no changes were detected for any changelog files.`,
+      });
+      return;
+    } else {
```
JamesHenry, line 566: Nit: the else is unnecessary with the early return

fahslaj: removed

JamesHenry (approved):

PR #21825, opened 2024-02-15
fix(testing): playwright - Property 'uiPort' does not match the schema

d3473r:
Running playwright e2e tests with a specific `uiPort` throws an error, the schema is expecting a string and not a number:

nx run e2e:e2e --ui-port=8080
>  NX   Property 'uiPort' does not match the schema. '8080' should be a 'string'.

## Current Behavior
see above

## Expected Behavior

Running playwright e2e tests with a specific `uiPort` works


d3473r:
This was fixed by https://github.com/nrwl/nx/pull/22610

PR #21831, opened 2024-02-15
docs(core): explain NX_ADD_PLUGINS

isaacplmann:
Explain NX_ADD_PLUGINS in the Inferred Tasks guide


docs/shared/concepts/inferred-tasks.md
```diff
@@ -70,3 +70,19 @@ More details about how to override task configuration is available in these reci
...
+## Existing Nx Workspaces
+
+If you have an existing Nx Workspace and upgrade to Nx 18, the migration generator will automatically add `NX_ADD_PLUGINS=false` to your `.env` file. This environment variable allows you to continue to use Nx without inferred tasks. We are working on creating migrations for existing workspaces to start switching to inferred tasks, where desired.
+
+When `NX_ADD_PLUGINS` is `false`:
+
+1. A newly generated project will have all targets defined with executors - not with inferred tasks.
+2. Running `nx add @nx/some-plugin` will not create a plugin entry for `@nx/some-plugin` in the `nx.json` file. (So that plugin will not create inferred tasks.)
+
+When `NX_ADD_PLUGINS` is `true`:
```
FrozenPandaz, line 83: This is the default behavior. So existing workspaces won't need to turn it on persay. And I would also explicitly say that we are working on a migration path for people so they should not worry too much about enabling it at this time.

Workspaces that have their `.env` file gitignored should ensure that everyone sets the variable. Such as adding it to a shared copy of the `.env` file that's not checked in or via adding it to a `.bashrc` script etc.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

juristr (approved):
Works for now. 

A couple of notes though, I think this concept page is starting to get bloated and we might want to extract things into separate recipes. Like this one should at a high level explain what Inferred Tasks (and Proj Crystal) is about, why it is beneficial etc..

Things like

- View Inferred Tasks
- Overriding Inferred Task Configuration
- (Opting-out of Inferred Tasks for existing workspaces)

..would almost be a better fit for a recipe and just be linked (via tagging) from the concept page.

PR #21850, opened 2024-02-16
fix(core): generate azure pipeline config correctly

barbados-clemens:
## Current Behavior

azure pipeline throws an error with the current template saying "unexpected Value 'displayName'" from default config
WkMacM1-y5ZlUQAq_02161219

## Expected Behavior

template is generated correctly

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
+++ b/packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
@@ -30,11 +30,7 @@ jobs:
       - bash: az devops configure --defaults organization=$(System.TeamFoundationCollectionUri) project=$(System.TeamProject)
         displayName: 'Set default Azure DevOps organization and project'
       # Get last successfull commit from Azure Devops CLI
-      - displayName: 'Get last successful commit SHA'
-        condition: ne(variables['Build.Reason'], 'PullRequest')
-        env:
-          AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
-        bash: |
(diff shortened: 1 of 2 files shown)


packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -30,11 +30,7 @@ jobs:
       - bash: az devops configure --defaults organization=$(System.TeamFoundationCollectionUri) project=$(System.TeamProject)
         displayName: 'Set default Azure DevOps organization and project'
       # Get last successfull commit from Azure Devops CLI
-      - displayName: 'Get last successful commit SHA'
-        condition: ne(variables['Build.Reason'], 'PullRequest')
-        env:
-          AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
-        bash: |
+      - bash: |
```
barbados-clemens, line 33: was informed that this actually does work if the repo is hosted in GitHub and uses ADO pipelines? 

unsure how that magically works, but should still default to work in both platforms if possible

barbados-clemens:
apparently this syntax works if using a GH hosted repo in ADO but breaks for ADO repos and vice versa if the proposed change is added. 
idk what microsoft be doing out there for this to happen. 

gonna close for now, but might be worth setting up a 'ADO' and 'ADO on GH' option to have the differences so people don't have to debug why something is broken on first generate. 🤷

meeroslav:
I could not see the difference in your commit apart from reordered parameters. Can you explain what the difference was that this PR was proposing?

PR #21890, opened 2024-02-20
fix(vite): update vitest and use parseCLI

mandarini:
Use the exposed method: https://github.com/vitest-dev/vitest/pull/5248

Related: https://github.com/nrwl/nx/pull/21870


leosvelperez (approved):

packages/vite/src/executors/test/lib/utils.ts
```diff
@@ -59,7 +61,9 @@ export async function getOptions(
       ? process.cwd()
       : relative(context.cwd, joinPathFragments(context.root, projectRoot));
 
-  const normalizedExtraArgs = normalizeArgs(extraArgs);
+  const { parseCLI } = await loadVitestDynamicImport();
+
+  const normalizedExtraArgs = parseCLI(getOptionsAsArgv(options));
```
leosvelperez, line 66: CI is failing because the `parseCLI` helper expects the `vitest` as the first argument.

leosvelperez (changes requested):

isaacplmann (approved):

jaysoo (approved):

leosvelperez (approved):

AgentEnder (approved):

PR #21912, opened 2024-02-21
feat(angular): update jest-preset-angular dependency to 14.0.3

Coly010:
## Current Behavior

The minimum version we add for `jest-preset-angular` is 14.0.2.
`jest-preset-angular` has released `14.0.3` with support for signal queries.

## Expected Behavior

Update minimum version we install for `jest-preset-angular` to `14.0.3`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/migrations.json
+++ b/packages/angular/migrations.json
@@ -1641,7 +1641,7 @@
       }
     },
     "18.1.0-jest": {
-      "version": "18.1.0-beta.1",
+      "version": "18.1.0-beta.2",
       "requires": {
         "@angular-devkit/build-angular": ">=15.0.0 <18.0.0",
         "@angular/compiler-cli": ">=15.0.0 <18.0.0",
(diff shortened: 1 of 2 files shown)


packages/angular/migrations.json
```diff
@@ -1657,7 +1657,7 @@
       }
     },
     "18.1.0": {
-      "version": "18.1.0-beta.1",
+      "version": "18.1.0-beta.2",
```
leosvelperez, line 1660: We can reuse the `18.1.0-jest` above this one. That's the correct one for `jest`. This package update group is meant to be used for the Angular update and I just realized is missing the prompt and the requirements. Could you please do thas as well in this PR?

leosvelperez (changes requested):

packages/angular/migrations.json
Coly010: no problem!

packages/angular/migrations.json
```diff
@@ -1651,13 +1651,17 @@
...
+          "version": "~14.0.3",
           "alwaysAddToPackageJson": false
         }
       }
     },
     "18.1.0": {
       "version": "18.1.0-beta.1",
+      "x-prompt": "Do you want to update the Angular version to v17.2?",
+      "requires": {
+        "@angular/core": ">=17.0.0 <17.2.0"
```
leosvelperez, line 1663:
```suggestion
        "@angular/core": ">=17.1.0 <17.2.0"
```

leosvelperez (changes requested):

leosvelperez (approved):

PR #21915, opened 2024-02-21
feat(vite): add vitest.workspace.ts at root

### LINKED ISSUE #18204: vitest CLI (and other vitest tools) do not work properly for monorepos
### Current Behavior

The `vitest` CLI (e.g. `npx vitest`) does not work correctly for monorepos. The workspace configuration file is missing from automated file generation.

In contrast, for `jest`, a `jest.config.ts` file is generated in the project root that correctly configures jest. For example:

**root/jest.config.ts**
```typescript
import { getJestProjects } from '@nx/jest';

export default {
  projects: getJestProjects(),
};
```

A similar file for `vitest` is required but is missing. For example:

**vitest.workspace.ts**
```typescript
import { getVitestProjects } from '@nx/vite';

export default [
  ...getVitestProjects()
];
```

### Expected Behavior

The `vitest` CLI and other tools that support `vitest` should work with default project configuration.

### GitHub Repo

not applicable

### Steps to Reproduce

1. See details of current behavior

### Nx Report

```shell
not applicable
```

### Failure Logs

```shell
not applicable
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

mandarini:
As per: https://vitest.dev/guide/workspace.html

* No need to add `name` in `vite(st).config.ts`, it works without name, too (example repo: https://github.com/mandarini/vitest-workspace-nx)
* No need for particular helper to collect all projects that use vitest, we can use the `'**/*/vite.config.ts', '**/*/vitest.config.ts'` globs

### DIFF
--- a/packages/vite/src/generators/vitest/vitest-generator.ts
+++ b/packages/vite/src/generators/vitest/vitest-generator.ts
@@ -46,10 +46,8 @@ export async function vitestGeneratorInternal(
 ) {
   const tasks: GeneratorCallback[] = [];
 
-  const { targets, root, projectType } = readProjectConfiguration(
-    tree,
-    schema.project
-  );
+  const { root, projectType } = readProjectConfiguration(tree, schema.project);
+  const isRootProject = root === '.';
 
   tasks.push(await jsInitGenerator(tree, { ...schema, skipFormat: true }));
   const initTask = await initGenerator(tree, {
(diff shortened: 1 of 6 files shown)


packages/vite/src/generators/vitest/vitest-generator.ts
```diff
@@ -118,6 +118,12 @@ export async function vitestGeneratorInternal(
   );
   tasks.push(installCoverageProviderTask);
 
+  // Setup workspace config file (https://vitest.dev/guide/workspace.html)
+  tree.write(
```
jaysoo, line 122: Let's check if it's not a root project and generate the file at that point.

mandarini: Yes yes thanks!

jaysoo (approved):

PR #21916, opened 2024-02-21
docs(core): remove angularjs migration

isaacplmann:
Remove angularjs migration guide


Coly010 (approved):

leosvelperez (approved):

leosvelperez (changes requested):
@isaacplmann there are a few other mentions for AngularJS that we should remove:

- https://github.com/nrwl/nx/blob/master/docs/shared/packages/angular/angular-plugin.md?plain=1#L13
- https://github.com/nrwl/nx/blob/master/packages/angular/package.json#L5
- https://github.com/nrwl/nx/blob/master/packages-legacy/angular/package.json#L5
- https://github.com/nrwl/nx/blob/master/docs/shared/mental-model/large-tasks.json#L26182-L26189 (non-important)

leosvelperez (approved):

jaysoo (approved):

AgentEnder (approved):

PR #21926, opened 2024-02-22
fix(core): exponential backoff retry on cache put fail

mike-fam:
This PR fixes the same issue as #7349, but implements a more resilient and less aggressive retry strategy. I regularly get cache errors on my pipeline, because I share the same local cache (i.e. EFS) between multiple EC2 agents. This is hard to reproduce since it's a race condition on my custom set-up. Given that this change won't affect the normal performance (since it's just an edge case), I think this change would be a nice-to-have.

### DIFF
--- a/packages/nx/src/tasks-runner/cache.ts
+++ b/packages/nx/src/tasks-runner/cache.ts
@@ -277,17 +277,19 @@ export class Cache {
   private tryAndRetry<T>(fn: () => Promise<T>): Promise<T> {
     let attempts = 0;
     const baseTimeout = 100;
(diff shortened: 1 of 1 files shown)


mike-fam:
Hi @FrozenPandaz, could you please let me know if you have time to review this change, or if there's something else I need to do before getting this PR reviewed? Thanks.

FrozenPandaz (approved):

FrozenPandaz:
The changes look good. Currently, it would go a very long time before actually giving up. I decreased the amount of maximum time it will wait.

PR #21943, opened 2024-02-22
fix(testing): ensure baseUrl is not passed to playwright cli

Coly010:
## Current Behavior

Playwright does not support `--base-url` as a CLI argument.
Some of our generators have generated `baseUrl` as an option to the `@nx:/playwright:playwright` executor.
When executed, this causes an error.

## Expected Behavior

Ensure that `baseUrl` does not get added to the CLI args if it exists in the project.json
Add a migration to remove `baseUrl` if it already exists in the project.json

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/playwright/src/migrations/update-18-1-0/remove-baseUrl-from-project-json.ts
+++ b/packages/playwright/src/migrations/update-18-1-0/remove-baseUrl-from-project-json.ts
@@ -1,0 +1,65 @@
+import {
+  formatFiles,
+  readNxJson,
+  readProjectConfiguration,
+  type Tree,
+  updateNxJson,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';
+
+export default async function (tree: Tree) {
+  forEachExecutorOptions(
+    tree,
+    '@nx/playwright:playwright',
+    (options, projectName, targetName, configurationName) => {
+      if (options?.['baseUrl']) {
+        const project = readProjectConfiguration(tree, projectName);
+        if (configurationName) {
+          delete project.targets[targetName].configurations[configurationName][
+            'baseUrl'
+          ];
+        } else {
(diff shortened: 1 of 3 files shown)


packages/playwright/src/executors/playwright/playwright.impl.ts
```diff
@@ -111,6 +111,17 @@ function createArgs(
...
+      // baseUrl should not exist as dictated by the Schema
+      // However, it does exist in some existing workspaces
+      console.warn(
+        "'baseUrl' is not a valid option for @nx/playwright:playwright. Please remove this from your project.json."
+      );
+      console.warn(
+        "You can configure 'baseUrl' in the 'playwright.config.ts' file."
+      );
+      continue;
+    }
```
leosvelperez, line 124: I don't think we need to validate this here. It's on users to provide the right extra args for Playwright and those should be validated/rejected by Playwright, not us.
Existing workspaces will get the option removed by the migration.

packages/playwright/src/migrations/update-18-1-0/remove-baseUrl-from-project-json.ts
```diff
@@ -0,0 +1,56 @@
...
+export default async function (tree: Tree) {
+  const playwrightProjects: ProjectConfiguration[] = [];
+  const projects = getProjects(tree);
+  projects.forEach((project) => {
+    // Check if the project contains playwright config
+    const configPath = joinPathFragments(project.root, 'playwright.config.ts');
+    if (tree.exists(configPath)) {
+      playwrightProjects.push(project);
+    }
+  });
```
leosvelperez, line 19: We shouldn't need to filter projects by the existence of a `playwright.config.ts`. That can leave other valid config file names out. A better and more accurate discriminator would be the executor.

Here we should use the `forEachExecutorOptions` utility instead. Also, we need to account for targetDefaults in `nx.json`.

leosvelperez (changes requested):

leosvelperez (approved):

PR #22004, opened 2024-02-26
feat(release): add conventional commits configurability for version and changelog

fahslaj:
## Current Behavior

- `nx release version` will only use `feat` and `fix` commit types when determining the specifier (when `specifierSource` is `conventional-commits`). It will also always assume that fixes are patches and feats are minor bumps.
- `nx release changelog` will only include `feat`, `fix`, and `perf` changes in changelogs.

## Expected Behavior

- `nx release version` will consider user configuration when picking the specifier when `specifierSource` is `conventional-commits`.
- `nx release changelog` will consider user configuration when filtering commits to appear in the changelog. It will also consider the configured section title.


docs/shared/recipes/nx-release/customize-conventional-commit-types.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Conventional Commits Usage within Nx Release
+
+The conventional commits configuration is used in two different places within Nx Release - once in the version step for determining the version bump, and once when generating changelogs.
+
+### Determine the Version Bump
+
+When `release.version.conventionalCommits` is `true` in `nx.json`, Nx Release will use the commit messages since the last release to determine the version bump. It will look at the type of each commit and determine the highest version bump from the following list:
+
+- 'feat' -> minor
```
JamesHenry, line 15: Missing breaking changes for major and a description of how they work (and how users, in the common case, are not expected to configure major bumps to go along with types themselves, and instead should understand how breaking changes are layered on top by us via `!` or `BREAKING CHANGE` - as per the conventional commits specification)

docs/shared/recipes/nx-release/customize-conventional-commit-types.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Generate Changelog Sections
+
+Nx Release sorts changes within changelogs into sections based on the type of commit. By default, `fix`, `feat`, and `perf` commits will be included in the changelog. To customize the headers of changelog sections, include other commit types, or exclude the default commit types, see the [Configure Commit Types](#configure-commit-types) section below.
+
+See the [Nx repo](https://github.com/nrwl/nx/releases) for examples of a changelogs generated with Nx Release.
+
+## Configure Commit Types
+
+Commit types are configured in the `release.conventionalCommits.types` property in `nx.json`:
```
JamesHenry, line 41: I'd like it to be made clearer that this is an advanced case and that most folks are perfectly fine/happy with the default behavior and no configuration should be required for most people (i.e. should just be `"conventionalCommits": true`

e2e/release/src/conventional-commits-config.test.ts
```diff
@@ -0,0 +1,355 @@
...
+              },
+            },
+            // unspecified semverBump will default to "patch"
+            chore: {
+              // "changelog.hidden" defaults to true, but setting changelog: false
+              // is a shortcut for setting "changelog.hidden" to false.
+              changelog: false,
+            },
+            // unspecified semverBump will default to "patch"
+            perf: {},
```
JamesHenry, line 133: Can we use true as a shorthand for "enabled with defaults" like we do in another places?

packages/nx/schemas/nx-schema.json
```diff
@@ -636,6 +639,55 @@
...
+      "type": "object",
+      "properties": {
+        "types": {
+          "type": "object",
+          "description": "A map of commit types to their configuration. If a type is set to 'false', then it will be hidden from changelogs and will not trigger a version bump.",
+          "additionalProperties": {
+            "oneOf": [
+              {
+                "type": "boolean",
+                "enum": [false]
```
JamesHenry, line 653: See comment about true as enabled with defaults

packages/nx/src/command-line/release/config/conventional-commits.ts
```diff
@@ -0,0 +1,98 @@
...
+      },
+      fix: {
+        semverBump: 'patch',
+        changelog: {
+          title: '🩹 Fixes',
+          hidden: false,
+        },
+      },
+      perf: {
+        semverBump: 'none',
```
JamesHenry, line 21: Patch right? How would users benefit from the performance improvements otherwise?

JamesHenry (changes requested):

packages/nx/src/command-line/release/config/conventional-commits.ts
fahslaj: This is consistent with existing functionality. `perf` types weren't triggering version bumps, but showed up in changelogs - https://github.com/nrwl/nx/blob/46592d79128b14ca3e98d62fe78be6aed4633505/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts#L10

I could default this to patch if you'd like, but it would be a breaking change.

docs/shared/recipes/nx-release/customize-conventional-commit-types.md
fahslaj: I'll add something about this being an advanced use case, but I want to point out that `release.conventionalCommits` property they are setting here is different than the existing `release.version.conventionalCommits` property that is frequently set to `true` to enable it for versioning. Most users will never set `release.conventionalCommits`, as its only valid value is an advanced config option outlined in this doc. It is invalid (and meaningless) for `release.conventionalCommits` to be set to `true`.

JamesHenry (approved):

isaacplmann (changes requested):

isaacplmann (approved):

PR #22007, opened 2024-02-27
feat(graph): add error boundary error page for project details

xiongemi:
## Current Behavior

no error page for project details view when killed graph

## Expected Behavior
Screenshot 2024-03-01 at 4 21 15 PM

Screenshot 2024-03-01 at 4 21 53 PM

## Related Issue(s)

Fixes #

### DIFF
--- a/graph/client/src/app/ui-components/error-boundary.tsx
+++ b/graph/client/src/app/ui-components/error-boundary.tsx
@@ -1,0 +1,27 @@
+import { useEnvironmentConfig } from '@nx/graph/shared';
+import { ProjectDetailsHeader } from 'graph/project-details/src/lib/project-details-header';
+import { useRouteError } from 'react-router-dom';
+
+export function ErrorBoundary() {
+  let error = useRouteError()?.toString();
+  console.error(error);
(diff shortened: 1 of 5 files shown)


graph/client/src/app/ui-components/error-boundary.tsx
```diff
@@ -0,0 +1,25 @@
...
+export function ErrorBoundary() {
+  let error = useRouteError().toString();
+  console.error(error);
+  const environment = useEnvironmentConfig()?.environment;
+
+  let message = 'Failed to fetch data for graph. ';
+  if (environment === 'nx-console') {
+    message += 'Please refresh the page.';
+  } else {
+    message += 'Please rerun the command and refresh the page.';
```
MaxKless, line 13: I would update the text slightly here. Maybe 'Please restart the grpah server and refresh the page'/
Because users might not know what 'the command' is/

MaxKless (changes requested):
- small rewording
- we should take light / dark mode into consideration. Right now if you're in dark mode, the error page is rendered in light mode which is kind of jarring

Otherwise it looks good :)

MaxKless (approved):

PR #22018, opened 2024-02-27
docs(core): npm workspace tutorial

isaacplmann:
- Creates an npm workspaces tutorial
- Removes the package-based tutorial
- Updates the Fence component to allow a shell code block to specify a path without any output
- Updates `create-nx-workspace` links to tutorials


jaysoo (approved):

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,353 @@
...
+    },
+    "lint": {
+      "cache": true
+    }
+  },
+  "defaultBase": "main"
+}
+```
+
+The `dependsOn` line makes Nx run the `typecheck` task for the current project and the `build` task for any dependencies before running the current project's `build` task. Now `nx build` will run the `typecheck` task just like `npm run build` does.
```
juristr, line 242: should we clarify the `^` better?

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,353 @@
...
+
+## Infer Cache Settings
```
juristr, line 244: Can we name this differently? Something like use Nx plugins to enhance etc?

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,353 @@
...
+## Summary
+
+After following this tutorial, the repository is still using all the same tools to run tasks, but now Nx runs those tasks in a smarter way. The tasks are efficiently cached so that there is no repeated work and the cache configuration settings are automatically synced with your tooling configuration files by Nx plugins. Also, any task dependencies are automatically executed whenever needed because we configured task pipelines for the projects.
+
+The final task graph for `demo` app's `build` task looks like this:
+
+{% graph height="200px" title="Build Task Pipeline" type="task" jsonFile="shared/tutorials/npm-workspaces-build-tasks2.json" %}
+{% /graph %}
+
+## CI with Nx
```
juristr, line 317: can we link to the CI tutorial? Also link to the recipes for the various CI providers so someone following along with this tutorial would have links to continue

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,353 @@
...
+nx generate ci-workflow --ci=github
+```
+
+{% callout type="note" title="Choose your CI provider" %}
+You can choose `github`, `circleci`, `azure`, `bitbucket-pipelines`, or `gitlab` for the `ci` flag.
+{% /callout %}
+
+This will create a default CI configuration that sets up Nx Cloud to [use distributed task execution](/ci/features/distribute-task-execution). This automatically runs all tasks on separate machines in parallel wherever possible, without requiring you to manually coordinate copying the output from one machine to another.
+
+## Next Steps
```
juristr, line 341: Maybe we should have a similar section on "Releasing with Nx Release" like we have for CI?

juristr (changes requested):

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,353 @@
...
+
+## Starting Repository
+
+To get started, check out [the sample repository](https://github.com/nrwl/tuskydesigns) on your local machine:
+
+```shell
+git clone https://github.com/nrwl/tuskydesign.git
+```
+
+The repository has two packages (under `packages/buttons` and `packages/forms`) that are used in a `demo` application located in `apps/demo`. The root `package.json` has a `workspaces` property that tells NPM how to find the projects in the repository.
```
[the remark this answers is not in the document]
isaacplmann: The image you added here doesn't show up

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,353 @@
...
+
+The first time `nx build` is run, it takes about 5 seconds - just like running `npm run build`. But the second time you run `nx build`, it completes instantly and displays this message:
+
+```text
+Nx read the output from the cache instead of running the command for 3 out of 3 tasks.
+```
+
+You can see the same caching behavior working when you run `npx nx lint` or `npx nx typecheck`.
+
+## Understand Task Pipelines
```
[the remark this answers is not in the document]
isaacplmann: I updated the section titles, but the sections are unique enough that I think it is clearer to keep them separate.

docs/shared/tutorials/npm-workspaces.md
isaacplmann: Is increasing `nx release` adoption one of our company OKRs? I feel like adding another section between `CI with Nx` and the `Next Steps` will decrease the impact of the call to action.  Some people will forget about CI as the next logical step and go look at `nx release`.  I think we should keep the message consistent and focused on going to CI next.

juristr: `nx release` is a major feature that closes a gap many ppl were complaining about wrt Nx. And an `npm workspace` is often something where the next step would be to implement a release process of some of these pacakges. We don't have to have a dedicated section but I'd at least mention it in the next steps so ppl have something to follow up with.

docs/shared/tutorials/npm-workspaces.md
```diff
@@ -0,0 +1,377 @@
...
+  },
+  "defaultBase": "main"
+}
+```
+
+The `dependsOn` line makes Nx run the `typecheck` task for the current project and the `build` task for any dependencies before running the current project's `build` task. Now `nx build` will run the `typecheck` task just like `npm run build` does.
+
+## Use Nx Plugins to Enhance Vite Tasks with Caching
+
+You may remember that we defined the `outputs` property in `nx.json` when we were answering questions in the `nx init` script. The value is currently hard-coded so that if you change the output path in your `vite.config.ts`, you have to remember to also change the `outputs` array in the `build` task configuration. This is where plugins can help.
```
juristr, line 270: I missed this the first round of review, but I feel like we need to emphasize the value proposition of plugins much more here. Like that last sentence "This is where plugins can help"..but transform it into something more like:

"This is where Nx plugins help. They directly infer information from the actual tooling configs like `vite.config.ts` in this case. Nx plugins...

- automatically configure caching for you, including inputs and outputs based on the underlying tooling config
- infer tasks ....."

something along those lines. wdyt?

juristr (changes requested):

docs/shared/tutorials/npm-workspaces.md
isaacplmann: I'll make a section before the CI section about Nx release.  I agree that is particularly important for the NPM workspace scenario.  I think if release is described first, it won't take away from the ending focus on CI.

juristr (approved):

PR #22027, opened 2024-02-27
docs(cypress): add "await" to the documented extension to setupNodEvents in order for webserver to be set up properly

jaysoo:
The current example here for Vite configuration is outdated and will not work with our new setup.

https://nx.dev/nx-api/cypress/generators/cypress-project#customizing-the-vitejs-configuration

This is because our `setupNodeEvents` return a promise that must be awaited on.

The task splitting example should also show `setupNodeEvents` as it's commonly used to prepare services for API servers, etc.

https://nx.dev/nx-api/cypress/documents/overview#splitting-e2e-tasks-by-file

The full example should be:

```ts
import { defineConfig } from 'cypress';
import { nxE2EPreset } from '@nx/cypress/plugins/cypress-preset';

const config = nxE2EPreset(__filename, {
  cypressDir: 'src',
  bundler: 'vite',
  webServerCommands: {
    default: 'nx run my-project:serve',
    production: 'nx run my-project:preview',
  },
  ciWebServerCommand: 'nx run my-project:serve-static',
});

export default defineConfig({
  e2e: {
    ...config,
    baseUrl: 'http://localhost:4200',

    // If you set `setupNodeEvents`, make sure to invoke the function from `nxE2EPreset`
    // See: https://docs.cypress.io/guides/references/configuration#setupNodeEvents
    async setupNodeEvents(on, config) {
      await config.setupNodeEvents(on, config);

      // Register your listeners here
    },
  },
});
```

## Related Issue(s)

Fixes #


isaacplmann (approved):

isaacplmann (approved):

docs/shared/packages/cypress/cypress-plugin.md
```diff
@@ -81,18 +81,28 @@ To enable e2e task splitting, make sure there is a `ciWebServerCommand` property
...
+    ...config,
     baseUrl: 'http://localhost:4200',
+
+    // If you set `setupNodeEvents`, make sure to invoke the function from `nxE2EPreset`
+    // See: https://docs.cypress.io/guides/references/configuration#setupNodeEvents
+    async setupNodeEvents(on, config) {
+      await config.setupNodeEvents(on, config);
+
+      // Register your listeners here
+    },
```
leosvelperez, line 105: I don't think this is relevant for this particular section of the docs. Here we're showing an example of splitting e2e tasks, and we're particularly focused on adding the `ciWebServerCommand`. The sentences leading to this code snippet describes that:

```
To enable e2e task splitting, make sure there is a `ciWebServerCommand` property set in your `cypress.config.ts` file. It will look something like this:
```

And we explicitly highlight the line where we set that option.

I think this could go into a recipe and be linked from the plugin page. If we want to keep it on the plugin page, it should probably be a separate section/example focusing on `setupNodeEvents`.

leosvelperez (changes requested):

leosvelperez (approved):

docs/shared/packages/cypress/cypress-plugin.md
```diff
@@ -97,7 +97,9 @@ export default defineConfig({
 });
 ```
 
-Note: The `nxE2EPreset` is a collection of default settings, but is not necessary for task splitting.
```
jaysoo, line 100: This turns out to be not true since local and CI runs depend on `webServerCommands` and `ciWebServerCommand` to start and listen to server ready. There are ways users can customize the startup on their own, but by default we want them to use our web server support.

leosvelperez (approved):

PR #22044, opened 2024-02-28
feat(angular): propagate execution context into esbuild plugins as variables

dmitry-stepanenko:
## Current Behavior

Right now esbuild plugins can be only configured based on the options provided in the target configuration
## Expected Behavior

There're scenarios where it's needed to know the context the plugin is executed with. 

For example, build target can either be invoked directly or be a `buildTarget` of a serve one. In my particular scenario I need to know what target was invoked in order to apply certain logic.

It is also similar to what we have with custom webpack configurations.

This PR adds an ability to provide variables (only `<ExecutorContext>` one for now) that are assigned by the executor in the runtime.

So that if plugin is defined as follows
```json
          {
            "path": "apps/my-app/plugins/plugin2.js",
            "options": {
              "someOption": "some value",
              "myContext": "<ExecutorContext>"
            }
          }
```
myContext will get the value of the executionContext object that is available to the particular executor.


leosvelperez:
hey @dmitry-stepanenko!

For the use case you've mentioned, we provide the `NX_TASK_TARGET_TARGET` env variable.

I'd prefer not to introduce some magic strings for variables to replace at runtime unless strictly needed. If we were to do something like this, it would be a separate argument provided to the plugin wrapper function (that's what we do with webpack).

Given the Nx task runner will set the `NX_TASK_TARGET_TARGET` env variable with the target being run, there's no real need to access the executor context, so I'm going to close this for the time being. If you have use cases for this feature, please create an issue with more details and ping me to discuss it further.

dmitry-stepanenko:
Thanks @leosvelperez , I guess I can use the env variable to achieve what I need.

leosvelperez:
@dmitry-stepanenko awesome. If you do come across a use case that would require more than that, please let me know. I'd be happy to discuss it further.

PR #22079, opened 2024-02-29
fix(nextjs): Add spec files when creating a next app

ndcunningham:


e2e/next-core/src/next-appdir.test.ts
```diff
@@ -35,7 +35,7 @@ describe('Next.js App Router', () => {
         import React from 'react';
         import { ${jsLib} } from '@${proj}/${jsLib}';
 
-        export default async function Page() {
+        export default function Page() {
```
ndcunningham, line 38: Next.js by default generates functions/pages that **are not** async, for `pages/` and `app/` routing conventions. We should follow the same assumption prioritizing client-side pages

xiongemi (approved):

PR #22103, opened 2024-03-01
docs(core): nx cloud video

isaacplmann:
Adds a video to the ci intro page

### DIFF
--- a/docs/nx-cloud/intro/ci-with-nx.md
+++ b/docs/nx-cloud/intro/ci-with-nx.md
@@ -20,13 +20,13 @@ npx nx connect
 
 {% cards cols="2" lgCols="4" mdCols="4" smCols="2" %}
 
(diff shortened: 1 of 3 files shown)


juristr (changes requested):
Couple of things.

- I'd use this as the first one in the row as it explains Nx + Nx Cloud overview
- let's name it "What is Nx Cloud" ? 🤔 
- let's use the Nx Agents or Nx Cloud icon as the logo instead of the YT one. Probably the Nx Cloud icon is better; maybe we could use the Nx Agents icon for the e2e test splitting instead
- We can either drop the YT button at all or have more buttons on the same row (e.g. make them smaller)

We should also update the "What is Nx Cloud" one on the nx.dev/docs getting started page
image

juristr (approved):

PR #22107, opened 2024-03-01
fix(core): fix azure pipelines template, docs

liammccartney:
Bash steps require bash to be the first property:
https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-bash?view=azure-pipelines#properties

## Current Behavior
Using the template or copying the example from the docs creates a broken azure pipelines configuration
Screenshot 2024-03-01 at 2 23 18 PM
image

## Expected Behavior
Using the template or copying the example from the docs creates a working azure pipelines configuration


meeroslav:
Thank you for the PR. Please update the test snapshots by running `pnpm nx test workspace -u`

meeroslav (changes requested):
This PR looks good but it's missing the unit test update.
Can you please include those changes as well?

mandarini:
@liammccartney hi there, thank you for the PR! Can you please add the unit tests, as requested by @meeroslav ? Then we can merge!

PR #22111, opened 2024-03-02
fix(core): prefer project specific external deps

JamesHenry:
## Current Behavior

We only ever discover one version of an external dependency for the file-map.json. This means the `@nx/dependency-checks` lint rule can produce incorrect failures when multiple copies of a dependency exist within a workspace.

## Expected Behavior

In the file-map.json the project specific version of a package (if multiple exist) is preferred and therefore the lint rule produces accurate results.

---

For example, in a repo where the root package.json has lodash@4.0.0 (which becomes `npm:lodash` on graph) and the foo project has lodash@3.0.0:

**Before**

image

**After**

image

### DIFF
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
@@ -1,22 +1,23 @@
-import { defaultFileRead } from '../../../../project-graph/file-utils';
 import { join } from 'path';
 import { DependencyType } from '../../../../config/project-graph';
-import { parseJson } from '../../../../utils/json';
-import { joinPathFragments } from '../../../../utils/path';
 import {
   ProjectConfiguration,
   ProjectsConfigurations,
 } from '../../../../config/workspace-json-project-json';
-import { NxJsonConfiguration } from '../../../../config/nx-json';
-import { PackageJson } from '../../../../utils/package-json';
+import { defaultFileRead } from '../../../../project-graph/file-utils';
 import { CreateDependenciesContext } from '../../../../project-graph/plugins';
 import {
   RawProjectGraphDependency,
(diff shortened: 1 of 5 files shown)


packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
```diff
@@ -99,10 +119,26 @@ function processPackageJson(
...
 function readDeps(packageJson: PackageJson) {
-  return [
-    ...Object.keys(packageJson?.dependencies ?? {}),
-    ...Object.keys(packageJson?.devDependencies ?? {}),
-    ...Object.keys(packageJson?.peerDependencies ?? {}),
-    ...Object.keys(packageJson?.optionalDependencies ?? {}),
+  const deps = {};
+
+  /**
+   * We process dependencies in a rough order or increasing importance such that if a dependency is listed in multiple
```
JamesHenry, line 125:
```suggestion
   * We process dependencies in a rough order of increasing importance such that if a dependency is listed in multiple
```

Will apply once the first CI run completes to see if there are other things to address

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
```diff
@@ -80,15 +89,29 @@ function processPackageJson(
         };
         validateDependency(dependency, ctx);
         collectedDeps.push(dependency);
-      } else if (ctx.externalNodes[`npm:${d}`]) {
+        return;
+      }
+      // external/3rd party package
+      const externalWithExactVersion = `npm:${d}@${version}`;
```
JamesHenry, line 95: TODO: semver range matching

JamesHenry:
Closed in favour of https://github.com/nrwl/nx/pull/23307

PR #22173, opened 2024-03-05
fix(core): Handle Next.js returning 0 for SIGINT and SIGTERM error codes

ndcunningham:
Currently, if a process was ran using the `run-commands` executor ends due to `SIGTERM` or `SIGINT` while the process code is `0`. Nx, will continue to cache that output.

To handle this type of behaviour we can force the process to be node instead of rust. 

This will be a temporary stand-in until: https://github.com/vercel/next.js/pull/62907 is merged and promoted.


FrozenPandaz:
This was handled with #23013

PR #22182, opened 2024-03-06
feat(core): support nx-cloud as non-default task runner

StalkAltan:


packages/nx/src/utils/nx-cloud-utils.ts
```diff
@@ -1,13 +1,27 @@
 import { NxJsonConfiguration, readNxJson } from '../config/nx-json';
 
-export function isNxCloudUsed(nxJson: NxJsonConfiguration) {
-  return (
+// Returns tuple [isCloudUsed, cloudConfigurationName]
+export function isNxCloudUsed(nxJson: NxJsonConfiguration): [boolean, string] {
```
FrozenPandaz, line 4: I think this function should only return `boolean`.

1. It's weird otherwise
2. There's no real magic going on here which can't be moved into `getCloudOptions` or somewhere else.

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -499,21 +493,10 @@ export function getRunnerOptions(
     ...nxArgs,
   };
 
-  // NOTE: we don't pull from env here because the cloud package
-  // supports it within nx-cloud's implementation. We could
-  // normalize it here, and that may make more sense, but
-  // leaving it as is for now.
-  if (nxJson.nxCloudAccessToken && isCloudDefault) {
```
FrozenPandaz, line 506: This seems unrelated? should this actually have changed?

With these changes, this would...  add this info to all runners, even non-nx-cloud runners, if it's in `nx.json`

FrozenPandaz (changes requested):

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -1216,11 +1216,12 @@ async function generateMigrationsJsonAndUpdatePackageJson(
       readNxVersion(originalPackageJson);
 
     try {
+      const [isCloudUsed, _] = isNxCloudUsed(originalNxJson);
```
meeroslav, line 1219: The second parameter is not used so there is no need to define it.

```suggestion
      const [isCloudUsed] = isNxCloudUsed(originalNxJson);
```

packages/nx/src/command-line/connect/connect-to-nx-cloud.ts
```diff
@@ -48,7 +48,8 @@ export async function connectToNxCloudIfExplicitlyAsked(
 
 export async function connectToNxCloudCommand(): Promise<boolean> {
   const nxJson = readNxJson();
-  if (isNxCloudUsed(nxJson)) {
+  const [isCloudUsed, _] = isNxCloudUsed(nxJson);
```
meeroslav, line 51: The second parameter is not used so there is no need to define it.

```suggestion
      const [isCloudUsed] = isNxCloudUsed(originalNxJson);
```

packages/nx/src/tasks-runner/life-cycles/view-logs-utils.ts
```diff
@@ -5,7 +5,8 @@ import { output } from '../../utils/output';
 const VIEW_LOGS_MESSAGE = `Hint: Try "nx view-logs" to get structured, searchable errors logs in your browser.`;
 
 export function viewLogsFooterRows(failedTasks: number) {
-  if (failedTasks >= 2 && !isNxCloudUsed(readNxJson())) {
+  const [isCloudUsed, _] = isNxCloudUsed(readNxJson());
```
meeroslav, line 8: The second parameter is not used so there is no need to define it.

```suggestion
  const [isCloudUsed] = isNxCloudUsed(readNxJson());
```

packages/nx/src/command-line/connect/view-logs.ts
```diff
@@ -7,8 +7,8 @@ import { readNxJson } from '../../config/nx-json';
 import { connectExistingRepoToNxCloudPrompt } from './connect-to-nx-cloud';
 
 export async function viewLogs(): Promise<number> {
-  const cloudUsed = isNxCloudUsed(readNxJson());
-  if (cloudUsed) {
+  const [isCloudUsed, _] = isNxCloudUsed(readNxJson());
```
meeroslav, line 10: The second parameter is not used so there is no need to define it.

```suggestion
  const [isCloudUsed] = isNxCloudUsed(readNxJson());
```

meeroslav (commented):
Minor cosmetical changes + agree with @FrozenPandaz's comments

PR #22184, opened 2024-03-06
docs(core): remove @nrwl/ rescope message from codeblocks

jaysoo:
Remove the `@nrwl/` rescope message from codeblocks since new users do not need to know about this, and existing users have mostly migrated to Nx 16+.

Screenshot 2024-03-06 at 1 52 44 PM

## Related Issue(s)

Fixes #


isaacplmann (changes requested):
You should also update the `terminal-video.component.tsx` file so that it doesn't use the `isMessageBelow` prop.

isaacplmann (approved):
Approving so you can merge when you update the reference

Issue #22006, opened 2024-02-26
`@nx/vite:test` hangs after dependency upgrade, if multiple spec files


mattfysh:
### Current Behavior

On running `nx test foo` the vite test executor hangs, and you have to either CTRL+C to get out or wait for a "fetch timeout" error. I've included a reproducible shell script below for showing the problem and how it only occurs when there is more than one test file

After running the script below, the only way to fix is to manually downgrade the vite version used by vitest and vite-node, by editing the pnpm-lock.yaml file, replace all instances of `vite: 5.1.4(@types/node@20.11.20)` with `vite: 5.0.12(@types/node@20.11.20)` then run `pnpm i` - the test will begin working again

### Expected Behavior

The test should complete as expected

### Steps to Reproduce

```sh
# vitest problems

project_name="zzz"

npx create-nx-workspace@latest $project_name \
  --workspaceType integrated --ci skip --preset ts --packageManager pnpm

cd $project_name

nx g lib foo --bundler=none --unit-test-runner=vitest

nx test foo # 1 spec file - works
cp foo/src/lib/foo.spec.ts foo/src/lib/foo2.spec.ts
nx test foo # 2 spec files - works

pnpm update --latest

nx test foo # 2 spec files - fails
rm foo/src/lib/foo2.spec.ts
nx test foo # 1 spec files - works

pnpm i -D vite@5.0.12
cp foo/src/lib/foo.spec.ts foo/src/lib/foo2.spec.ts
nx test foo # 2 spec files - fails
```

### Nx Report

```shell
Node   : 21.6.2
OS     : darwin-arm64
pnpm   : 8.15.3

nx (global)        : 18.0.4
nx                 : 18.0.5
@nx/js             : 18.0.5
@nx/eslint         : 18.0.5
@nx/workspace      : 18.0.5
@nx/eslint-plugin  : 18.0.5
@nx/vite           : 18.0.5
@nx/web            : 18.0.5
typescript         : 5.3.3
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

mattfysh:
I'm going to chuck a bunch of other issues I've found in here as well, because it seems to me like the vite plugin is not being properly maintained

* without setting `watch` anywhere, it seems to be defaulting to `watch=true` which will cause all sorts of issues, e.g. snapshots won't update and teardown won't run. Despite the task ending after the first test run, the console output shows that it is running in watch mode
[...]

PR #22185, opened 2024-03-06
fix(vite): fix ignored vitest cli options (#22006)

bheftel:
## Current Behavior

Passing CLI parameters to vitest doesn't work - vitest does not get the params and they are ignored.

`nx run my-app:test --update` does not update the snapshots.

It looks like this started happening after the switch to use `parseCLI` in https://github.com/nrwl/nx/pull/21890/

`parseCLI` returns CLI parameters in an `options` object (see https://vitest.dev/advanced/api.html#parsecli), and the merged vitest config looks like this:
```typescript
{
    ...,
    options: {
        update: true
    }
}
```

Before the switch to `parseCLI`, options were included directly in the merged config, so it looked like this:
```typescript
{
    ...,
    update: true
}
```

`options` isn't a valid field for a vitest config, and the `update` flag gets ignored.

## Expected Behavior

`nx run my-app:test --update` should update the snapshots.

## Related Issue(s)

I'm linking to https://github.com/nrwl/nx/issues/22006 because this issue was noted in this comment - https://github.com/nrwl/nx/issues/22006#issuecomment-1965539423

But this isn't about the hanging test runner, so please let me know if it is better to file a new issue rather than use the existing one.

Related to #22006

### DIFF
--- a/packages/vite/src/executors/test/lib/utils.ts
+++ b/packages/vite/src/executors/test/lib/utils.ts
@@ -69,7 +69,7 @@ export async function getOptions(
   ]);
 
   const settings = {
-    ...normalizedExtraArgs,
+    ...normalizedExtraArgs.options,
     // This should not be needed as it's going to be set in vite.config.ts
     // but leaving it here in case someone did not migrate correctly
     root: resolved.config.root ?? root,


Coly010:
Fixed already by: https://github.com/nrwl/nx/pull/22355

PR #22208, opened 2024-03-07
chore(repo): cleanup macos pipeline

meeroslav:
## Related Issue(s)

Fixes #


FrozenPandaz:
What is the point of this? It just moves those steps to a different part of the file right?

meeroslav:
It's making pipeline more readable imho, and adds the same set of env params like linus run does.

PR #22231, opened 2024-03-08
fix(angular): install jsonc-eslint-parser only when @nx/dependency-checks is used

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/generators/add-linting/lib/buildable-project.ts
+++ b/packages/angular/src/generators/add-linting/lib/buildable-project.ts
@@ -1,0 +1,12 @@
+import { readProjectConfiguration, type Tree } from '@nx/devkit';
+
+export function isBuildableLibraryProject(
+  tree: Tree,
+  projectName: string
+): boolean {
+  const projectConfig = readProjectConfiguration(tree, projectName);
+
+  return (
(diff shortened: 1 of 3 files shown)


packages/angular/src/generators/add-linting/lib/buildable-project.ts
```diff
@@ -0,0 +1,12 @@
+import { readProjectConfiguration, type Tree } from '@nx/devkit';
+
+export function isBuildableLibraryProject(
+  tree: Tree,
+  projectName: string
+): boolean {
+  const projectConfig = readProjectConfiguration(tree, projectName);
+
+  return (
+    projectConfig.projectType === 'library' && !!projectConfig.targets?.build
```
Coly010, line 10: Would it be better to check for executor name?
Someone may not have named the target "build".

leosvelperez: This is an internal generator only used by the Angular plugin generators (folks could still use it, but that's like using a private API, not guaranteed to work). We already had this same check for quite a while because our generators always generate the target `build` for buildable libraries.

Coly010: no problem :)

Coly010 (approved):

PR #22284, opened 2024-03-12
docs(misc): add recipe on how to pass args to commands

leosvelperez:
- New Recipe: https://nx-dev-git-fork-leosvelperez-docs-recipe-pass-args-06908e-nrwl.vercel.app/recipes/running-tasks/pass-args-to-commands
- Updated:
  - https://nx-dev-git-fork-leosvelperez-docs-recipe-pass-args-06908e-nrwl.vercel.app/reference/nx-json#executorcommand-options
  - https://nx-dev-git-fork-leosvelperez-docs-recipe-pass-args-06908e-nrwl.vercel.app/reference/project-configuration#executorcommand-options

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/recipes/running-tasks/pass-args-to-commands.md
+++ b/docs/shared/recipes/running-tasks/pass-args-to-commands.md
@@ -157,6 +157,10 @@ To statically pass some extra args to a specific project, you can update its `pr
 {% tabs %}
 {% tab label="Setting args directly as options" %}
 
+{% callout type="note" title="Providing command args as options" %}
+Support for providing command args as options was added in **Nx v18.1.1**.
+{% /callout %}
+
 ```json {% fileName="apps/my-app/project.json" highlightLines=["5-10"] %}
 {
   "sourceRoot": "apps/my-app/src",
@@ -175,10 +179,6 @@ To statically pass some extra args to a specific project, you can update its `pr
 {% /tab %}
 
 {% tab label="Setting the \"args\" option" %}
-
-{% callout type="note" title="Providing command args as options" %}
-Support for providing command args as options was added in **Nx v18.1.1**.
-{% /callout %}
 
 ```json {% fileName="apps/my-app/project.json" highlightLines=["5-11"] %}
 {
@@ -248,22 +248,8 @@ If multiple targets with the same name run different commands (or use different 
 
 ## Pass args when running the command in the terminal
 
-To pass args in a one-off manner when running a task, you can also provide them as individual options or by providing the `args` option when running the task:
-
-{% tabs %}
-{% tab label="Providing args directly as options" %}
+To pass args in a one-off manner when running a task, you can provide them as options directly:
 
 ```shell
 nx build my-app --assetsInlineLimit=2048 --assetsDir=static/assets
 ```
-
-{% /tab %}
-
-{% tab label="Providing the \"args\" option" %}
-
-```shell
-nx build my-app --args="--assetsInlineLimit=2048 --assetsDir=static/assets"
-```
-
-{% /tab %}
-{% /tabs %}
--- a/docs/shared/recipes/running-tasks/pass-args-to-commands.md
+++ b/docs/shared/recipes/running-tasks/pass-args-to-commands.md
@@ -248,8 +248,34 @@ If multiple targets with the same name run different commands (or use different 
 
 ## Pass args when running the command in the terminal
 
-To pass args in a one-off manner when running a task, you can provide them as options directly:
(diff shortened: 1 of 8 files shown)


docs/shared/reference/nx-json.md
```diff
@@ -265,6 +265,37 @@ If you are using distributed task execution and disable caching for a given targ
 
 {% /callout %}
 
+### Executor/command options
+
+You can configure options specific to a target's executor. As an example, if your repo have projects using the `@nx/js:tsc` executor, you can provide some default options as follows:
```
isaacplmann, line 270:
```suggestion
You can configure options specific to a target's executor. As an example, if your repo has projects using the `@nx/js:tsc` executor, you can provide some default options as follows:
```

docs/shared/reference/nx-json.md
```diff
@@ -265,6 +265,37 @@ If you are using distributed task execution and disable caching for a given targ
...
+{
+  "targetDefaults": {
+    "@nx/js:tsc": {
+      "generateExportsField": true
+    }
+  }
+}
+```
+
+You can also provide defaults for [inferred targets](/concepts/inferred-tasks) or targets running a command using the `nx:run-commands` executor. As an example, if your repo has projects with a `build` inferred target running the `vite build` command, you can provide some defaults options as follows:
```
isaacplmann, line 282: This is a pretty dangerous example.  Should we even show people this?

docs/shared/reference/project-configuration.md
```diff
@@ -569,6 +569,40 @@ Additionally, when using the expanded object syntax, you can specify individual
...
+        "assetsDir": "static/assets"
+      }
+    }
+  }
+}
+```
+
+For more details on how to pass args to the underlying command see the [Pass Args to Commands recipe](/recipes/running-tasks/pass-args-to-commands).
+
+In the case of an explicit target using an executor, you need to specify the executor and the options specific to that executor as follows:
```
isaacplmann, line 591:
```suggestion
In the case of an explicit target using an executor, you can specify the executor and the options specific to that executor as follows:
```

isaacplmann (changes requested):

docs/shared/reference/nx-json.md
leosvelperez: That's a good point. It's true that it's dangerous for projects with multiple plugins inferring multiple `build` targets with different commands, but there are projects that use a single tool to build and where this wouldn't be a dangerous thing to do, and it will be as valid as setting target defaults for explicit targets/executors.

I think the danger this poses makes it more important to have an example but we need to add a caution callout to explain the possible consequences. Even if we don't show an example, users might still add a target default and wouldn't know about the possible consequences. So, having an explicit example and calling out the potentially bad consequences in some scenarios would actually be more helpful.

Wdyt?

leosvelperez: I amended the PR with the rephrased example and the caution callout. Let me know what you think.

docs/shared/reference/nx-json.md
```diff
@@ -265,6 +265,43 @@ If you are using distributed task execution and disable caching for a given targ
...
+        "assetsInlineLimit": 2048,
+        "assetsDir": "static/assets"
+      }
+    }
+  }
+}
+```
+
+{% callout type="caution" title="Be careful" %}
+If multiple targets with the same name run different commands (or with different executors), do not set options in `targetDefaults`. Different commands would accept different options, and the target defaults will apply to all targets with the same name regardless of the command they run. If you were to provide options in `targetDefaults` for them, the different commands that don't expect those options could throw.
```
isaacplmann, line 300:
```suggestion
If multiple targets with the same name run different commands (or use different executors), do not set options in `targetDefaults`. Different commands would accept different options, and the target defaults will apply to all targets with the same name regardless of the command they run. If you were to provide options in `targetDefaults` for them, the commands that don't expect those options could throw an error.
```

isaacplmann (approved):
Modified the wording of your callout a bit.

isaacplmann (changes requested):

docs/shared/recipes/running-tasks/pass-args-to-commands.md
```diff
@@ -0,0 +1,269 @@
...
+}
+```
+
+{% /tab %}
+
+{% tab label="Setting the \"args\" option" %}
+
+{% callout type="note" title="Providing command args as options" %}
+Support for providing command args as options was added in **Nx v18.1.1**.
+{% /callout %}
```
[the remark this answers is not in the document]
leosvelperez: Yeah, I had this updated locally but didn't push it before because I had other pending changes. I've now pushed the change.

isaacplmann (approved):

PR #22299, opened 2024-03-13
feat(core): add ability to add metadata to projects

FrozenPandaz:
## Current Behavior

There is no official API for adding metadata to Project Nodes

## Expected Behavior

During the `createNodes` process, plugins will be able to add metadata to Project Nodes.

2 of these fields will be handled in the future by the `nx graph` app and Nx Cloud:
1. `technologies`
  a. Type: `string[]`
  b. An array of technologies which the project utilizes
  c. ex. React, Angular, Gradle, Cypress
3. `targetGroups`
  a. Type: `Record<string, string[]>`
  b. A map of target groups which will be grouped by the project details view as well as Nx Cloud
  c. ex. `e2e-ci`

The cypress plugin has been updated to update cypress nodes

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/config/workspace-json-project-json.ts
+++ b/packages/nx/src/config/workspace-json-project-json.ts
@@ -111,6 +111,10 @@ export interface ProjectConfiguration {
       'generator' | 'generatorOptions'
     >;
   };
+  metadata?: {
+    technologies?: string[];
+    targetGroups?: Record<string, string[]>;
+  };
 }
 
 export interface TargetDependencyConfig {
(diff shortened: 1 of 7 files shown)


packages/nx/src/config/workspace-json-project-json.ts
```diff
@@ -111,6 +111,10 @@ export interface ProjectConfiguration {
       'generator' | 'generatorOptions'
     >;
   };
+  metadata?: {
+    technologies?: string[];
+    targetGroups?: Record<string, string[]>;
+  };
```
leosvelperez, line 117: Should we add this to the `project-schema.json`?

Also, did we consider additionally allowing any arbitrary metadata to be set? That could be useful for third-party plugins.

FrozenPandaz: I didn't add it to `project-schema.json` even though it works when defined in `project.json` because I'm not sure we should recommend it. At least for the metadata that we're going to consume, I think it should be mostly programmatic.

By the way, I believe we've actually always allowed for arbitrary properties on the project node.. Like it will end up on the project graph but we don't do anything with it.

xiongemi (approved):

leosvelperez (approved):

PR #22331, opened 2024-03-15
fix(release): fix default renderer resolution to be relative within t…

FrozenPandaz:
…he package

## Current Behavior

The default renderer is specified as an absolute path which can get resolved to a different version of Nx.

For the nx repo, this is the typescript code.

## Expected Behavior

The default renderer is specified as a relative path which is always within the same package release is being run with.

## Related Issue(s)

Fixes #


JamesHenry (approved):

JamesHenry (changes requested):
Need to double check snapshot changes

JamesHenry (approved):

PR #22334, opened 2024-03-15
feat(remix): add option to create-nx-workspace

Coly010:
## Current Behavior

We currently do not have an option for selecting Remix as a framework when generating a new workspace with `create-nx-workspace`.

## Expected Behavior

Add ability to select Remix when generating a new workspace with `create-nx-workspace`.

## Related Issue(s)

Fixes #


isaacplmann (approved):

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -93,6 +90,33 @@ async function createPreset(tree: Tree, options: Schema) {
...
+      '/remix/generators');
+
+    return remixApplicationGenerator(tree, {
+      name: options.name,
+      directory: '.',
+      projectNameAndRootFormat: 'as-provided',
+      linter: options.linter,
+      e2eTestRunner: options.e2eTestRunner ?? 'cypress',
+      rootProject: true,
+      unitTestRunner: 'vitest',
```
ndcunningham, line 117: Is this always `vitest` or is `jest` an option as well?

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -93,6 +90,33 @@ async function createPreset(tree: Tree, options: Schema) {
...
+    const { applicationGenerator: remixApplicationGenerator } = require('@nx' +
+      '/remix/generators');
+
+    return remixApplicationGenerator(tree, {
+      name: options.name,
+      directory: join('apps', options.name),
+      projectNameAndRootFormat: 'as-provided',
+      linter: options.linter,
+      e2eTestRunner: options.e2eTestRunner ?? 'cypress',
+      unitTestRunner: 'vitest',
```
ndcunningham, line 103: same as below

ndcunningham (approved):

packages/workspace/src/generators/preset/preset.ts
Coly010: Jest is an option; but it doesn’t get asked in the CNW flow for React applications.

not sure if that’s intentional or not.

but vitest is default and provides best experience, and as this is out a new workspace, I think it’s fine.

FrozenPandaz (approved):

PR #22371, opened 2024-03-18
feat(core): validate that outputs is an array of strings

MJez29:
## Current Behavior

I accidentally configured my target outputs as
```
outputs: "{projectRoot}/dist"
```

And was confused by the error message

```
The following outputs are invalid:
- {
- p
- r
- o
- j
...
Please run "nx repair" to repair your configuration
```

## Expected Behavior

A slightly nicer error message

```
The 'outputs' field must be an array
```

I also made it so that if you enter non-string values in the array it flags them

```
The 'outputs' field must contain only strings, but received types: [string, number, object, boolean, object, object]
```

### DIFF
--- a/packages/nx/src/tasks-runner/utils.ts
+++ b/packages/nx/src/tasks-runner/utils.ts
@@ -107,7 +107,32 @@ class InvalidOutputsError extends Error {
   }
 }
 
+function assertOutputsAreValidType(outputs: unknown) {
+  if (!Array.isArray(outputs)) {
+    throw new Error("The 'outputs' field must be an array");
+  }
+
+  const typesArray = [];
+  let hasInvalidType = false;
+  for (const output of outputs) {
+    if (typeof output !== 'string') {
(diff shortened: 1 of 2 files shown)


packages/nx/src/tasks-runner/utils.ts
```diff
@@ -107,7 +107,40 @@ class InvalidOutputsError extends Error {
...
+    }
+  }
+}
+
 export function validateOutputs(outputs: string[]) {
+  const invalidOutputsTypeError =
+    InvalidOutputsTypeError.maybeGetError(outputs);
+  if (invalidOutputsTypeError) {
+    throw invalidOutputsTypeError;
+  }
```
FrozenPandaz, line 142: Can you change this to an `assert` function please?

Something like `assertOutputsAreValidType`

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #22379, opened 2024-03-18
feat(core): add ability to scope plugins

FrozenPandaz:
## Current Behavior

There is no way to users to filter where plugins are applied. Only by the plugin author.

This means that errors thrown make the plugin unusable and more modifications to the project graph are made than desired.

## Expected Behavior

Plugins can be scoped via `include` and `exclude` which are both lists of file patterns.

```json
{
  plugins: [
    {
      plugin: '@nx/jest/plugin',
      include: ['packages/**/*'],
      exclude: ['e2e/**/*']
    }
  ]
}
```

This will scope running the plugin to processing only those config files.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/devkit/PluginConfiguration.md
+++ b/docs/generated/devkit/PluginConfiguration.md
@@ -1,3 +1,3 @@
 # Type alias: PluginConfiguration
 
-Ƭ **PluginConfiguration**: `string` \| \{ `options?`: `unknown` ; `plugin`: `string` }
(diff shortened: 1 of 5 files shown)


AgentEnder (approved):
Looks good, thanks Jason!

tukusejssirs:
Could this be used to exclude a target for a workspace-level `project.json` (see #22183)? 🤔

PR #22388, opened 2024-03-19
fix(linter): convert parser options to flat config even is parser is missing

meeroslav:
When converting to flat config and parser is missing, the parserOptions would be left unparsed which then lead to a runtime error. This PR ensures parserOptions are properly handled on their own.

## Related Issue(s)

Fixes #


JamesHenry (approved):

PaulACoroneos:
Thanks for the quick turnaround on this.

PR #22426, opened 2024-03-21
fix(core): handle undefined properties in schemas with additionalProperties

### LINKED ISSUE #22349: Invalid JSON Schema enforcement of additionalProperties
### Current Behavior

A executor schema that allows an object with arbitrary keys so long as the values match.

Typescript equivalent: `Record<{ foo: Record<string, number> }>`

```
{
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "foo": {
      "type": "object",
      "additionalProperties": {
          "type": "number"
      }
    }
  }
}
```

Running the executor throws with
```


 NX   Cannot convert undefined or null to object


TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at /workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:194:24
    at Array.find (<anonymous>)
    at validateObject (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:193:27)
    at validateProperty (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:346:9)
    at /workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:210:9
    at Array.forEach (<anonymous>)
    at validateObject (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:209:23)
    at validateOptsAgainstSchema (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:144:5)
    at combineOptionsForExecutor (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:432:5)
```

### Expected Behavior

Executor can run successfully. Schema is treated as valid.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create an executor schema that has a `"type": "object"` with `"additionalProperties"` set to anything but true.
2. Attempt to run the executor

### Nx Report

```shell
Node   : 20.11.1
OS     : linux-arm64
pnpm   : 8.15.4

nx (global)  : 18.1.1
nx           : 18.1.1
@nx/eslint   : 18.1.1
@nrwl/tao    : 18.1.1
```

### Failure Logs

```shell
NX   Cannot convert undefined or null to object


TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at /workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:194:24
    at Array.find (<anonymous>)
    at validateObject (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:193:27)
    at validateProperty (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:346:9)
    at /workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:210:9
    at Array.forEach (<anonymous>)
    at validateObject (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:209:23)
    at validateOptsAgainstSchema (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:144:5)
    at combineOptionsForExecutor (/workspace/node_modules/.pnpm/nx@18.1.1/node_modules/nx/src/utils/params.js:432:5)
```

### Package Manager Version

pnpm 8.15.4

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Offending logic is right where the error logs point us: https://github.com/nrwl/nx/blob/master/packages/nx/src/utils/params.ts#L293

If `additionalProperties` is found, the logic attempts to get the keys of `properties`. However there is no guarantee that value exists, and `Object.keys(undefined)` is an error.

The JSONSchema spec does not specify that `properties` _must_ be set in order to use `additionalProperties` (as a general rule of thumb, no fields allow/disallow others).

This logic should default `properties` to `{}` in the case where it does not exist.

Alternatively, I recommend switching to a battle tested JSON Schema validator, such as [AJV](https://ajv.js.org/guide/getting-started.html)

MaxKless:
fixes https://github.com/nrwl/nx/issues/22349

### DIFF
--- a/e2e/nx-run/src/run.test.ts
+++ b/e2e/nx-run/src/run.test.ts
@@ -190,6 +190,27 @@ describe('Nx Running Tests', () => {
         expect(output).toContain(app);
       });
     });
+
+    it('should pass env option to nx:run-commands executor', () => {
+      const mylib = uniq('mylib');
+      runCLI(`generate @nx/js:lib ${mylib}`);
+
+      updateJson(`libs/${mylib}/project.json`, (c) => {
+        c.targets['echo'] = {
+          executor: 'nx:run-commands',
+          options: {
+            command: 'node -e "console.log(process.env.ONE)"',
+            env: {
+              ONE: 'TWO',
+            },
+          },
+        };
+        return c;
+      });
+
+      const output = runCLI(`echo ${mylib}`);
+      expect(output).toContain('TWO');
+    });
   });
 
   describe('Nx Bail', () => {
--- a/packages/nx/src/tasks-runner/task-orchestrator.ts
+++ b/packages/nx/src/tasks-runner/task-orchestrator.ts
@@ -351,7 +351,7 @@ export class TaskOrchestrator {
     const temporaryOutputPath = this.cache.temporaryOutputPath(task);
     const streamOutput = shouldStreamOutput(task, this.initiatingProject);
 
-    const env = pipeOutput
+    let env = pipeOutput
       ? getEnvVariablesForTask(
           task,
           taskSpecificEnv,
@@ -403,6 +403,12 @@ export class TaskOrchestrator {
           relative(task.projectRoot ?? workspaceRoot, process.cwd()),
           process.env.NX_VERBOSE_LOGGING === 'true'
         );
+        if (combinedOptions.env) {
+          env = {
+            ...env,
+            ...combinedOptions.env,
+          };
+        }
         if (streamOutput) {
           const args = getPrintableCommandArgsForTask(task);
           output.logCommand(args.join(' '));
--- a/packages/nx/src/utils/params.ts
+++ b/packages/nx/src/utils/params.ts
@@ -306,7 +306,7 @@ export function validateObject(
   ) {
     Object.keys(opts).find((p) => {
       if (
-        Object.keys(schema.properties).indexOf(p) === -1 &&
+        Object.keys(schema.properties ?? {}).indexOf(p) === -1 &&
         (!schema.patternProperties ||
           !Object.keys(schema.patternProperties).some((pattern) =>
             new RegExp(pattern).test(p)


robertIsaac:
thanks for the fix :)

noelsoong:
i don't like the idea of using the nullish operator and giving it a default empty object.

i think the schema.additionalProperties !== undefined could be changed to 
```typescript
schema.additionalProperties != null
```

noelsoong:
nvm addditionalProperties is not .properties carry on XD

mattfysh:
This does prevent the error from appearing, but I think there is a larger issue at play here because after applying this fix the property that was triggering this error is still ignored. To reproduce:

```
npx create-nx-workspace@latest xyz \
  --workspaceType integrated --ci skip --preset ts --packageManager pnpm
cd xyz
nx g lib foo --bundler=none --unit-test-runner=none
```

*foo/project.json*
```
"printenv": {
      "options": {
        "command": "env",
        "env": {
          "ONE": "TWO"
        }
      }
    },
```

Without this fix applied, you'll see the `Cannot convert undefined or null to object`

After applying the fix, the error no longer appears, but the property that triggered the error (`env`) is not applied to the task. You can see the ONE env var is missing in the logged output

MaxKless:
thanks @mattfysh for the catch & the repro. Super helpful! 
It's an unrelated issue but I fixed it and added it to the PR :)

FrozenPandaz (approved):

robertIsaac:
hi @FrozenPandaz 
do you plan to publish it as a bugfix for nx 18.1?

PR #22434, opened 2024-03-21
fix(core): fix no plugins found for nx init without packge.json

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/init/init-v2.ts
+++ b/packages/nx/src/command-line/init/init-v2.ts
@@ -72,9 +72,9 @@ export async function initHandler(options: InitArgs): Promise<void> {
 
   output.log({ title: '🧐 Checking dependencies' });
 
-  const detectPluginsResponse = await detectPlugins();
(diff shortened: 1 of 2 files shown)


packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -49,12 +49,14 @@ export async function initHandler(options: InitArgs): Promise<void> {
       );
     }
     generateDotNxSetup(version);
-    const { plugins } = await detectPlugins();
-    plugins.forEach((plugin) => {
-      execSync(`./nx add ${plugin}`, {
-        stdio: 'inherit',
+    const detectPluginsResponse = await detectPlugins();
```
FrozenPandaz, line 52: Let's make `detectPlugins` return an empty array?

FrozenPandaz (changes requested):
There's still the issue with the `--help` being printed out.

FrozenPandaz (approved):

PR #22449, opened 2024-03-22
docs(nx-cloud): add why nx cloud page

juristr:
https://nx-dev-git-fork-juristr-docs-why-nx-cloud-page-nrwl.vercel.app/ci/intro/why-nx-cloud


docs/nx-cloud/intro/why-nx-cloud.md
```diff
@@ -0,0 +1,64 @@
...
+```yml
+- run: npx nx-cloud record -- nx format:check
+- run: npx nx affected -t lint test build e2e-ci
+```
+
+All these commands are automatically picked up by Nx Cloud, split up into smaller tasks and distributed across the specified number of machines.
+
+**Step 4: All results are played back automatically**
+
+Nx Cloud automatically plays back all results to your CI system, as if distribution never happened. You can continue doing post-processing on the results, like uploading test reports, deploying artifacts etc.
```
isaacplmann, line 57: This section is no longer talking about "Why Nx Cloud?" and it is duplicating content that already exists in several places.

I would rewrite it to talk about the mental model of writing your CI pipeline as if it were executing on your local machine, but Nx automatically makes it faster for you.  Then you can talk about how that process works at each step.

So you can keep some of this same content - but don't frame it with `Step 1`, `Step 2` etc.

- To start using cloud, there's a simple registration process. (Link to recipes)
- Nx Cloud makes distribution across machines easy. Changing the way tasks are distributed is controlled with a single command.
- The rest of the CI pipeline runs the same commands you would run on your local machine.  This allows you to think about the pipeline as if it were run locally, but Nx allows it to be efficiently parallelized when it is run in CI.
- This is made possible because Nx Cloud automatically plays back all results to your CI system, as if distribution never happened.

isaacplmann (dismissed):

docs/nx-cloud/intro/why-nx-cloud.md
```diff
@@ -0,0 +1,64 @@
...
+width="100%" /%}
+
+If CI feels challenging to you, **it might not be your fault**. It's a fundamental issue with how the current, traditional CI execution model works. Nx Cloud adopts a new **task-based** CI model which allows to overcome the limitations of slowness but also unreliability of the current VM-based CI execution. _(Dive deeper into the task based CI execution model [in this blog post](https://blog.nrwl.io/reliable-ci-a-new-execution-model-fixing-both-flakiness-and-slowness-6849fd4b4037))_
+
+Such approach has led to great results:
+
+- 30 - 70% faster CI
+- 40 - 75% reduction in CI costs
+- 9x faster e2e test execution
+- robust CI setup by automatically identifying flaky tasks (e2e tests in particular) and re-running them
```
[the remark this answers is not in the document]
juristr: I'd like to have this as part of the bullet points, but I see your point. I'll make it shorter

docs/nx-cloud/intro/why-nx-cloud.md
juristr: I'm aware there's duplication, but I think that's fine. I intentionally made it as little as possible, but I want to give the full picture here. I had only such bullet points first, but I don't want to have ppl jump again to pages etc, but at least get to see "that single line" etc. Also seeing some concrete things is easier to parse than just bullet points.

If they then go and implement Nx Cloud they'd have to dig deeper anyway, so this page is to give an overview and get the bird-view understanding of what's going on.

meeroslav (approved):

PR #22452, opened 2024-03-22
docs(nx-cloud): add basic Azure DevOps setup recipe

catfireparty:
Adds Azure DevOps to the list of supported VCS integrations and adds a recipe for the setup.

Also converted all the `png` files in that set-up folder to `webp`.

(Fixes APP-1280)


isaacplmann (approved):
Approved, but you should update the permission image's alt text

PR #22455, opened 2024-03-22
docs(nx-dev): Add feedback CTA

ndcunningham:
Adds Feedback CTA to all pages that have the table of contents

The changes include:

1. Feedback CTA and Edit this page are now listed with the table of contents.
2. The Feedback CTA will only be shown either in the Table of Contents or before the footer (Not all pages have Table of Contents)
3. The Feedback opens a Modal dialog which gives the options of opening an issue or providing a suggestion

Can be seen here:
Screenshot 2024-03-28 at 4 08 00 PM

Which opens:
Screenshot 2024-03-28 at 4 08 25 PM


isaacplmann (changes requested):
Looks good overall.  A few minor tweaks

isaacplmann (approved):

FrozenPandaz (approved):

PR #22456, opened 2024-03-22
docs(misc): improve docs around target defaults

leosvelperez:
Updated docs:

- https://nx-dev-git-fork-leosvelperez-docs-update-targets-defaults-nrwl.vercel.app/recipes/tips-n-tricks/integrated-in-package-based#projectjson
- https://nx-dev-git-fork-leosvelperez-docs-update-targets-defaults-nrwl.vercel.app/recipes/running-tasks/configure-inputs#view-the-inputs-of-a-task
- https://nx-dev-git-fork-leosvelperez-docs-update-targets-defaults-nrwl.vercel.app/recipes/running-tasks/reduce-repetitive-configuration#reduce-configuration-with-targetdefaults
- https://nx-dev-git-fork-leosvelperez-docs-update-targets-defaults-nrwl.vercel.app/reference/nx-json

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/recipes/repo-types/integrated-in-package-based.md
+++ b/docs/shared/recipes/repo-types/integrated-in-package-based.md
@@ -4,7 +4,7 @@ In a package-based repository, you've intentionally opted out of some of Nx's fe
 
 To add an integrated project:
 
-1. Install the plugin you want to use (i.e. `npm install @nx/react`)
+1. Install the plugin you want to use (i.e. `nx add @nx/react`)
 2. Generate an application or library using that plugin (i.e. `nx g @nx/react:app`)
 
 The integrated project is now ready to use. Next, we'll discuss some of the changes that were applied to your codebase.
--- a/docs/shared/recipes/repo-types/integrated-in-package-based.md
+++ b/docs/shared/recipes/repo-types/integrated-in-package-based.md
@@ -19,7 +19,7 @@ Even if the project you added only uses javascript, a `tsconfig.base.json` file 
 
 ## project.json
 
-The project itself will have a `project.json` file that defines all the tasks that can be run on the project. This includes tasks like `build`, `serve` and `test`. See [Executors and Configurations](/concepts/executors-and-configurations) for more information.
+The project itself will have [a `project.json` file](/reference/project-configuration#project-level-configuration-files) where some tasks might be explicitly defined while other [tasks will be inferred by the plugin](/concepts/inferred-tasks) used. This can include tasks like `build`, `serve` and `test`.
 
 ## Other Configuration Files
 
--- a/docs/shared/recipes/running-tasks/configure-inputs.md
+++ b/docs/shared/recipes/running-tasks/configure-inputs.md
@@ -69,7 +69,7 @@ Doing so will show a view such as the one below:
 {% project-details  jsonFile="shared/concepts/myreactapp.json"%}
 {% /project-details %}
 
-Nx Console has a button which will show a preview of this screen when a project level configuration file (`project.json` or `package.json`) is opened in the IDE.
+Nx Console has a button which will show a preview of this screen when a project level configuration file (`project.json` or `package.json`) is opened in the IDE. See more at [Nx Console Project Details View](/recipes/nx-console/console-project-details).
 
 Another way of accessing this information is to run `nx show project myreactapp --web` and the view above will be opened in a browser.
 
--- a/docs/shared/recipes/running-tasks/configure-inputs.md
+++ b/docs/shared/recipes/running-tasks/configure-inputs.md
@@ -69,7 +69,7 @@ Doing so will show a view such as the one below:
 {% project-details  jsonFile="shared/concepts/myreactapp.json"%}
 {% /project-details %}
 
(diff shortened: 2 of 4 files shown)


docs/shared/recipes/repo-types/integrated-in-package-based.md
```diff
@@ -19,7 +19,7 @@ Even if the project you added only uses javascript, a `tsconfig.base.json` file
 
 ## project.json
```
juristr, line 20: I wonder whether this page is needed at all. I have a hard time to think about who would go and read this specific one, also because with Crystal you'd expect to be able to use `nx add @nx/react` and then keep going (which we're not 100% there yet especially with generators).

This page also still contains steps for first `npm install` etc.. rather than using `nx add`.

docs/shared/recipes/running-tasks/configure-inputs.md
```diff
@@ -69,7 +69,7 @@ Doing so will show a view such as the one below:
 {% project-details  jsonFile="shared/concepts/myreactapp.json"%}
 {% /project-details %}
 
-Nx Console has a button which will show a preview of this screen when a project level configuration file (`project.json` or `package.json`) is opened in the IDE.
+Nx Console has a button which will show a preview of this screen when a project level configuration file (`project.json` or `package.json`) is opened in the IDE. See more at [Nx Console Project Details View](/recipes/nx-console/console-project-details).
```
juristr, line 72: I'd say "Read more about..."

docs/shared/reference/nx-json.md
```diff
@@ -22,10 +22,9 @@ The following is an expanded example showing all options. Your `nx.json` will li
     "production": ["!{projectRoot}/**/*.spec.tsx"]
   },
   "targetDefaults": {
-    "build": {
+    "@nx/js:tsc": {
```
juristr, line 25: tbh I never used this syntax yet. So this updates all targets using the `@nx/js:tsc` executor. Is there a way to define the target name in this config as well? So have both?

I also think it's still valuable to have the alternative version of defining just the target name like `build`, `lint` etc and then have another one that shows how to "target" a specific executor. There is one such example further down in the `Outputs` section. Maybe we could just merge those and use a similar one.

juristr (changes requested):

docs/shared/recipes/repo-types/integrated-in-package-based.md
leosvelperez: I can definitely amend the `npm install` but this PR is for a task to address some `targetDefaults` examples, not to update the full pages. Whether the page is still needed or not, that's also a separate conversation. I don't think we're there yet. Generators definitely don't support package-based repos yet.

docs/shared/reference/nx-json.md
leosvelperez: > tbh I never used this syntax yet. So this updates all targets using the `@nx/js:tsc` executor. Is there a way to define the target name in this config as well? So have both?

It will apply to all targets using that executor. It's explained here https://nx-dev-git-fork-leosvelperez-docs-update-targets-defaults-nrwl.vercel.app/reference/nx-json#target-defaults. 

You can have both, but the `targetDefaults` for an executor will override the `targetDefaults` for a target that uses the executor. They are not merged and the executor one takes precedence.

> I also think it's still valuable to have the alternative version of defining just the target name like `build`, `lint` etc and then have another one that shows how to "target" a specific executor. There is one such example further down in the `Outputs` section. Maybe we could just merge those and use a similar one.

I'm not sure we need to duplicate examples with that. We'd need to do it everywhere. We've always supported both, and most of the time we were only suggesting using target names as keys. Now with Project Crystal, we generate `targetDefaults` using executors as the key. So, we're replacing some relevant examples to show that a bit more. We're not hiding you can use targets as keys. In the link I shared above, which is in this same document, it's explained that you can use both, and as you mentioned, throughout our docs there will be examples of both. I'm not replacing all examples with executors.

juristr: Sure, I'm not saying duplicate across the board, but ppl are still familiar seeing the `targetName` config approach, in particular when it comes to defining caching related config or `dependsOn` as it is usually related to a target.
That's why I mentioned having both options as someone using a npm workspace repo looking at this would be really confused because they wouldn't even use Nx plugins etc. While someone using Nx Crystal plugins might want to see your setup there.

leosvelperez: That's fine. I still don't think it's too relevant given that most examples in this same page and in all the docs mostly use target names anyways (so users not using executors have plenty of examples showing what they have), but I also think it doesn't hurt to add it, so I added a separate target default using a target to this particular snippet 👍🏻 

This snippet is only an example showcasing the structure of an `nx.json` file. Users don't stop at this snippet and leave with all the information they need. When users go to a reference page, they look for definitions, and the definition for `targetDefaults` explains that you can use an executor or a target name as the key. It describes how they are applied, what the precedence is, and a few examples of using one or the other. So, IMHO, anyone coming to the page will still read the description of the options they're interested in, and that's what needs to show and explain anything needed.

juristr (approved):

PR #22487, opened 2024-03-25
fix(web): spa flag should correctly define redirect

### LINKED ISSUE #22418: Inferred Tasks Not Working after Upgrading Nx
### Current Behavior

I am upgrading from Nx v16.10 to latest so that I can use the inferred tasks, specifically e2e-ci. I followed the upgrade steps, but the e2e-ci inferred task will not show up. The e2e task was inferred, however

### Expected Behavior

Adding the nx/cypress plugin and using inferred tasks should show the e2e-ci tasks

### GitHub Repo

The repo is private, but I talked to @Coly010 about getting on a call to show the repo

### Steps to Reproduce

1. Start with an Nx workspace on an older version, and then upgrade it to the latest
2. Add the Cypress plugin
3. show the inferred tasks, and e2e-ci does not show

### Nx Report

```shell

Node   : 18.0.0
OS     : darwin-arm64
npm    : 8.6.0

nx                 : 18.0.8
@nx/js             : 18.0.8
@nx/jest           : 18.0.8
@nx/linter         : 18.0.8
@nx/eslint         : 18.0.8
@nx/workspace      : 18.0.8
@nx/angular        : 18.0.8
@nx/cypress        : 18.0.8
@nx/devkit         : 18.0.8
@nx/eslint-plugin  : 18.0.8
@nrwl/tao          : 18.0.8
@nx/web            : 18.0.8
@nx/webpack        : 18.0.8
typescript         : 5.3.3
```

### Failure Logs

_No response_

### Package Manager Version

npm v8.6.0

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I spoke with @Coly010 on Twitter and at ng-conf, and have shown him what's happening. He and I talked about getting together on a call to walk through the issue if needed,

Coly010:
## Current Behavior

The `@nx/web:file-server` executor does not set up `spa` handling correctly. The current solution still causes the browser to throw a 404 which can cause unexpected behavior.
Additionally, some of the packages that make use of the `@nx/web:file-server` were not setting `spa` and thus deep linking would 404.

## Expected Behavior

The `spa` flag on `@nx/web:file-server` should correctly redirect to index.html and not throw 404s.
The packages that need `spa` from `@nx/web:file-server` should define it in their target, or via their inferred targets.

## Related Issue(s)

Fixes #22418
Despite having the `ciWebServerCommand` set, the `serve-static` target needs to have `spa` working correctly.

### DIFF
--- a/packages/web/src/executors/file-server/file-server.impl.ts
+++ b/packages/web/src/executors/file-server/file-server.impl.ts
@@ -181,6 +181,7 @@ export default async function* fileServerExecutor(
     run();
   }
 
+  const port = await detectPort(options.port || 8080);
   const outputPath = getBuildTargetOutputPath(options, context);
 
   if (options.spa) {
@@ -189,6 +190,10 @@ export default async function* fileServerExecutor(
 
     // See: https://github.com/http-party/http-server#magic-files
     copyFileSync(src, dst);
+
+    // We also need to ensure the proxyUrl is set, otherwise the browser will continue to throw a 404 error
+    // This can cause unexpected behaviors and failures especially in automated test suites
+    options.proxyUrl ??= `http${options.ssl ? 's' : ''}://localhost:${port}?`;
   }
 
   const args = getHttpServerArgs(options);
@@ -205,7 +210,6 @@ export default async function* fileServerExecutor(
 
   // detect port as close to when used to prevent port being used by another process
   // when running in  parallel
-  const port = await detectPort(options.port || 8080);
   args.push(`-p=${port}`);
 
   const serve = fork(pathToHttpServer, [outputPath, ...args], {
--- a/packages/web/src/generators/static-serve/static-serve-configuration.ts
+++ b/packages/web/src/generators/static-serve/static-serve-configuration.ts
@@ -1,42 +1,54 @@
 import {
+  createProjectGraphAsync,
   logger,
   parseTargetString,
+  type ProjectGraph,
+  readCachedProjectGraph,
   readProjectConfiguration,
   stripIndents,
-  TargetConfiguration,
-  Tree,
+  type TargetConfiguration,
+  type Tree,
   updateProjectConfiguration,
 } from '@nx/devkit';
-import { Schema as FileServerExecutorSchema } from '../../executors/file-server/schema.d';
+import type { Schema as FileServerExecutorSchema } from '../../executors/file-server/schema.d';
 
 interface WebStaticServeSchema {
   buildTarget: string;
   outputPath?: string;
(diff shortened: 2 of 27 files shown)


packages/web/src/generators/static-serve/static-serve-configuration.ts
```diff
@@ -1,42 +1,54 @@
...
   options: WebStaticServeSchema
-): NormalizedWebStaticServeSchema {
-  const target = parseTargetString(options.buildTarget);
+): Promise<NormalizedWebStaticServeSchema> {
+  let projectGraph: ProjectGraph;
+  try {
+    projectGraph = readCachedProjectGraph();
+  } catch (e) {
+    projectGraph = await createProjectGraphAsync();
+  }
```
leosvelperez, line 45: We need to be careful about using the project graph in generators. If a project generator (e.g. `@nx/next:app`) invokes this generator, the project being generated is not yet in the project graph because the generation changes have not been flushed to disk yet. So, anything that relies on the project graph to get the still-in-memory project wouldn't find it.

I'm unsure if there's indeed a problem here or not, but we definitely need to check all usages of this generator in other generators and validate the above.

leosvelperez (changes requested):

packages/web/src/executors/file-server/file-server.impl.ts
```diff
@@ -189,6 +190,10 @@ export default async function* fileServerExecutor(
 
     // See: https://github.com/http-party/http-server#magic-files
     copyFileSync(src, dst);
+
+    // We also need to ensure the proxyUrl is set, otherwise the browser will continue to throw a 404 error
+    // This can cause unexpected behaviors and failures especially in automated test suites
+    options.proxyUrl ??= `http${options.ssl ? 's' : ''}://localhost:${port}?`;
```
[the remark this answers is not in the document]
Coly010: Yes, this is the only option from `http-server` along with creating a 404 page which has the index contents, which is why I didn't remove that logic. I'll add a disclaimer to the `proxyUrl` option

packages/web/src/generators/static-serve/static-serve-configuration.ts
Coly010: It is only being used by `parseTargetString` which handles the case of when the project graph does not contain the project. 

But for cases when the generator is invoked via CLI, it is more accurate.

leosvelperez (approved):

jaysoo (approved):

PR #22502, opened 2024-03-26
fix(core): combine executor properties with name `env` before passing to run commands

StalkAltan:
## Current Behavior
Nx's environment variable loading is also called `env`, which would overwrite the nx:run-commands executor `env` property value. Combine the two if an `env` property from the executor exists.

## Expected Behavior
`env` executor property is propagated properly


FrozenPandaz:
Duplicate of https://github.com/nrwl/nx/pull/22426

PR #22519, opened 2024-03-26
docs(core): update troubleshooting guide & console adjustments

MaxKless:
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/installation/troubleshoot-installation.md
+++ b/docs/shared/installation/troubleshoot-installation.md
@@ -49,9 +49,8 @@ If you're running a machine that isn't part of the list above, then Nx does not 
 
 ### Node.js Installation Issues
 
-Especially on MacOS, it's possible that Node.js is installed with the wrong architecture (`x64` instead of `arm64`). Because of the Rosetta compatibility layer, Node.js will still run but the Nx CLI won't be able to find the proper native binary for your platform.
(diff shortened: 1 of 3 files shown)


docs/shared/installation/troubleshoot-installation.md
```diff
@@ -45,3 +46,12 @@ We publish modules for the following platforms:
 - FreeBSD (x64)
 
 If you're running a machine that isn't part of the list above, then Nx does not support it at this time. [Please open an issue on GitHub](https://github.com/nrwl/nx/issues/new/choose) if you feel Nx should support that platform and we will assess what can be done, please make sure to include your platform and architecture in the issue.
+
+### Node.js Installation Issues
```
jaysoo, line 50: Reword this for clarity and conciseness. It could also happen on other platforms other than MacOS.

Maybe something like this:

```md
### Node.js Installation Issues

Ensure that the architecture of your Node.js installation matches your hardware. Run `nx report` and check that the `OS` property is correct (e.g. `darwin-arm64` for MacOS on Apple silicon). If it contains `x64` even though you're on an `arm64` chip, then something is wrong. A mismatch in architecture can lead to errors loading Nx's native binary.

Often, the culprit of the mismatch is a faulty installation of your toolchain: Homebrew (MacOS), Node.js or VSCode (if using Nx Console). You should reinstall your toolchain with the correct architecture. Run `nx report` again to validate the installation.

For issues inside VSCode or Nx Console, also refer to the [Nx Console troubleshooting docs](recipes/nx-console/console-troubleshooting)
```

leosvelperez (approved):

Issue #22531, opened 2024-03-27
@nx/esbuild:esbuild thirdParty not handled


zartinn:
### Current Behavior

When using `"thirdParty": false` third party packages from node_modules are still bundled.

### Expected Behavior

When using `"thirdParty": false` third party packages from node_modules should not be bundled.

It should be behave the same as I would set the option in esbuild `packages=external`
See documentation here: https://esbuild.github.io/api/#packages

The thirdParty property is not handled at all in the [build-esbuild-options module](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts) and therefore the packages option for esbuild is never set.

### Steps to Reproduce

1. create a simple node project
2. import and use at least one external dependency (like express)
3. build the node project with `@nx/esbuild:build` executor and the option `thirdParty: false`
4. look into the output and see that the external dependency is bundled with the code

### Nx Report

```shell
Node   : 20.11.1
OS     : darwin-arm64
npm    : 10.2.4

nx (global)        : 18.0.5
nx                 : 18.1.3
@nx/js             : 18.1.3
@nx/jest           : 18.1.3
@nx/linter         : 18.1.3
@nx/eslint         : 18.1.3
@nx/workspace      : 18.1.3
@nx/esbuild        : 18.1.3
@nx/eslint-plugin  : 18.1.3
@nx/nest           : 18.1.3
@nx/node           : 18.1.3
@nrwl/tao          : 18.1.3
typescript         : 5.3.3
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I created a PR to handle this bug:
https://github.com/nrwl/nx/pull/22532

zartinn:
Okay, it seems there is a bit more complex logic with the thirdParty options and my PR isn't sufficient enough.
I just recognized that the function: [`getExtraDependencies`](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/lib/get-extra-dependencies.ts)

doesn't return any dependencies in my case. I will dig into it deeper later

zartinn:
It was something stupid having the package-lock.json in the .gitignore (had done some testing with bun.

PR #22532, opened 2024-03-27
fix(esbuild): thirdparty option

zartinn:
## Current Behavior
When using "thirdParty": false third party packages from node_modules are still bundled.

## Expected Behavior
When using "thirdParty": false third party packages from node_modules should not be bundled.

It should be behave the same as I would set the option in esbuild packages=external
See documentation here: https://esbuild.github.io/api/#packages

The thirdParty property is not handled at all in the [build-esbuild-options module](https://github.com/nrwl/nx/blob/master/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts) and therefore the packages option for esbuild is never set.

## Related Issue(s)
[@nx/esbuild:esbuild thirdParty not handled](https://github.com/nrwl/nx/issues/22531)

### DIFF
--- a/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
+++ b/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
@@ -34,6 +34,7 @@ export function buildEsbuildOptions(
           ...options.external,
         ]
       : undefined,
+    packages: options.thirdParty ? undefined : 'external',
     minify: options.minify,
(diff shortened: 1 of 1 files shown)


zartinn:
On hold due to: https://github.com/nrwl/nx/issues/22531#issuecomment-2022169664

zartinn:
It was really something stupid like having the package-lock.json in the .gitignore (had it in there due some testing with bun)

PR #22536, opened 2024-03-27
chore(repo): bump agents docker image

leosvelperez:
Updates the Nx Agents Docker image to a newer one that supports Node 20.11.1, which is a requirement of the upcoming Angular v18.

The new image changed the PNPM installation from `@pnpm/exe` to `pnpm`, so this PR also updates the PNPM installation to match that. Without that change, the step fails, as can be seen in the following CIPE: https://staging.nx.app/cipes/6602f2ab7216aa25ee6d7123?step=c39656c5-7e40-4921-9fe0-84cb4580beeb#step-list-pane.

The following are runs with the old image and the new image with the Angular v18 upcoming changes:

- Old image: https://staging.nx.app/runs/NalwptI048
  - the projects `e2e-angular-core` and `e2e-nx-init` fail due to an invalid node engine
- New image: https://staging.nx.app/runs/l6PYvgnKY6
  - the projects `e2e-angular-core` and `e2e-nx-init` no longer fail

## Related Issue(s)

Fixes #

### DIFF
--- a/.nx/workflows/agents.yaml
+++ b/.nx/workflows/agents.yaml
@@ -1,7 +1,7 @@
 launch-templates:
   linux-medium:
     resource-class: 'docker_linux_amd64/medium+'
-    image: 'ubuntu22.04-node20.9-v3'
+    image: 'ubuntu22.04-node20.11-v3'
     env:
       GIT_AUTHOR_EMAIL: test@test.com
       GIT_AUTHOR_NAME: Test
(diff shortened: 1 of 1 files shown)


.nx/workflows/agents.yaml
```diff
@@ -30,7 +30,7 @@ launch-templates:
           sudo apt-get install -y ca-certificates lsof
       - name: Install Pnpm
         script: |
-          npm install -g @pnpm/exe@8.7.4
```
meeroslav, line 33: Remove this, the image already has pnpm installed

meeroslav (dismissed):
Small change required

FrozenPandaz (approved):

PR #22544, opened 2024-03-27
fix(webpack): resolve assets from executor options as relative to workspace root

jaysoo:
The [previous fix to relative asset paths](https://github.com/nrwl/nx/commit/4be897ac3f40d4b4d4001584d9dbd8d6ab85db96) introduced a regression when using `@nx/webpack:webpack` executor.

When options are passed from the executor, the paths are relative to the workspace root _not_ project root. Thus, we need special handling when normalizing executor options.

## Current Behavior
Executor is resolving assets relative to project root.

## Expected Behavior
Executor options should always be relative to workspace root for consistency.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/normalize-options.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/normalize-options.ts
@@ -51,14 +51,16 @@ export function normalizeOptions(
     }
     Object.assign(
       combinedPluginAndMaybeExecutorOptions,
-      buildTargetOptions,
-      options // plugin options take precedence
+      options,
+      // executor options take precedence (especially for overriding with CLI args)
+      buildTargetOptions
     );
   } else {
     Object.assign(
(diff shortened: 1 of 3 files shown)


packages/webpack/src/plugins/nx-webpack-plugin/lib/normalize-options.ts
```diff
@@ -51,14 +51,16 @@ export function normalizeOptions(
...
+      buildTargetOptions
     );
   } else {
     Object.assign(
       combinedPluginAndMaybeExecutorOptions,
-      originalTargetOptions,
-      options // plugin options take precedence
+      options,
+      // executor options take precedence (especially for overriding with CLI args)
+      originalTargetOptions
```
jaysoo, line 63: The order here was previously set wrong. The executor options should override plugin options, since users can pass CLI args, or have overrides in `configurations`.

Coly010 (approved):

ndcunningham (approved):

PR #22565, opened 2024-03-28
fix(testing): app generators should create correct e2e config at generation time

Coly010:
## Current Behavior

The current e2e app generation for various app generators has the chance of producing invalid config files due to some edge cases around Nx Crystal.

## Expected Behavior

Ensure e2e app generation creates correct config

## Related Issue(s)

Fixes #


packages/expo/src/generators/application/lib/normalize-options.ts
```diff
@@ -41,8 +45,34 @@ export async function normalizeOptions(
     ? options.tags.split(',').map((s) => s.trim())
     : [];
   const rootProject = appProjectRoot === '.';
+
+  let e2eWebServerTarget = 'serve';
```
Coly010, line 49: @xiongemi Could make sure some of the assumptions for `expo` projects are correct here?

- default `webServerTarget`
- default `port` the app gets served on

packages/react-native/src/generators/application/lib/normalize-options.ts
```diff
@@ -44,8 +48,35 @@ export async function normalizeOptions(
   const iosProjectRoot = joinPathFragments(appProjectRoot, 'ios');
   const androidProjectRoot = joinPathFragments(appProjectRoot, 'android');
   const rootProject = appProjectRoot === '.';
+
+  let e2eWebServerTarget = 'start';
```
Coly010, line 52: @xiongemi Could make sure some of the assumptions for `react-native` projects are correct here?

- default `webServerTarget`
- default `port` the app gets served on

packages/react/src/generators/application/lib/normalize-options.ts
ndcunningham, line 1: Based on this, if the `nx.json` has both plugins how will it decide which plugin the project is using currently without checking for a config?

ndcunningham (commented):
Next LGTM

packages/react/src/generators/application/lib/normalize-options.ts
ndcunningham, line 1: Based on these changes, if the `nx.json` has both plugins (`webpack` and `vite`) how will this decide which plugin the project is using currently without checking for a config?

ndcunningham (commented):
Next LGTM

packages/react/src/generators/application/lib/normalize-options.ts
ndcunningham, line 1: Based on these changes, if the `nx.json` has both plugins (`webpack` and `vite`) how will this decide which plugin the project is using currently without checking for a config?

Coly010: As this is happening during application generation, we use the option user has specified: https://github.com/nrwl/nx/pull/22565/files/6c1a16e96fe8b5cecc8aed45bd62bc985812fed6#diff-38fa881985e391e1ad6deea8885844e8ddac2374b59be50fcecff18b4fae629fR54

xiongemi (approved):

leosvelperez (approved):

ndcunningham (approved):

PR #22571, opened 2024-03-28
fix(nuxt): use loadConfigFile from devkit rather than @nuxt/kit

jaysoo:
When using `@nuxt/kit` to load `nuxt.config.ts` files, it can result in errors when used in conjunction with `@nx/vite/plugin` (and the `loadConfigFile` function from `vite`).

This PR switches to our own function from` @nx/devkit`, which we can guarantee plays nicely with `@nx/cypress`, `@nx/playwright`, and anything else that uses the same plugin.

Note: When we isolate each crystal plugin into their own worker, this issue should be resolved as well. However, this is a quicker fix with smaller impact.

## Current Behavior
Errors can occur during graph construction when combining nuxt + vitest + playwright/cypress.

## Expected Behavior
Errors should not occur during graph construction.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nuxt/src/plugins/plugin.ts
+++ b/packages/nuxt/src/plugins/plugin.ts
@@ -13,9 +13,9 @@ import { basename, dirname, isAbsolute, join, relative } from 'path';
 import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
 import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
 import { existsSync, readdirSync } from 'fs';
-import { loadNuxtKitDynamicImport } from '../utils/executor-utils';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
 import { getLockFileName } from '@nx/js';
+import { loadConfigFile } from '@nx/devkit/src/utils/config-utils';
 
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/plugins/plugin.ts
```diff
@@ -226,16 +227,10 @@ function getOutputs(
...
-  if (nuxtConfig?.buildDir && basename(nuxtConfig?.buildDir) === '.nuxt') {
-    // if buildDir exists, it will be `something/something/.nuxt`
-    // we want the "general" outputPath to be `something/something`
-    nuxtBuildDir = nuxtConfig.buildDir.replace(
-      basename(nuxtConfig.buildDir),
-      ''
-    );
-  }
-  const buildOutputPath = normalizeOutputPath(nuxtBuildDir, projectRoot);
+  const buildOutputPath = normalizeOutputPath(
```
jaysoo, line 230: We actually only want to cache the `.nuxt` folder. This is consistent with Next.js where we cache only `.next` folder. Previously, this logic was to handle caching `dist/<app>` rather than just `dist/<app>/.nuxt`, but we've since moved the default `buildDir` back to within the project root anyway. We cannot cache the entire `{projectRoot}`, which includes source code.

Coly010 (approved):
LGTM!

ndcunningham (approved):

PR #22585, opened 2024-03-29
fix(core): expand env variables on load and unload

matheo:
Env variables using other variables were not unloaded from the environment and further customizations were impossible in more specific env files.

## Current Behavior
The unloading and re-loading of env variables in [task-runner/task-env](https://github.com/nrwl/nx/blob/ef81455b6492fee266c0d6ddfc41a79dc2041f8b/packages/nx/src/tasks-runner/task-env.ts#L189) is faulty because composed variables are not unloaded because they are not expanded, just parsed and checked to be equal to the process.env value.

I'm using the `@nx-tools/nx-container` plugin to build Docker images and I'm passing some input parameters via env var like:
```
INPUT_BUILD_ARGS="API_HOST=${API_HOST}\nAPI_PORT=${API_PORT}"
```
Nx loads `.env`, `.local.env` and `.env.local` by default and I have the localhost configuration in one of them.

But when I want to build for `production` I want to get the values of `.env.production` and  debugging the issue deeply I found that my var was not unloaded because it was comparing the evaluated value in process.env (`INPUT_BUILD_ARGS="API_HOST=localhost\nAPI_PORT=3000"`) with the non-expanded value (`INPUT_BUILD_ARGS="API_HOST=${API_HOST}\nAPI_PORT=${API_PORT}"`) so it was not unloaded.

After this adjustment and the upgrade of `dotenv-expand` to not affect process.env it works successfully :)

## Expected Behavior
To have my composed variables expanded so they are correctly unloaded and overridden by a specific `.env*` file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,7 +163,7 @@
     "czg": "^1.4.0",
     "detect-port": "^1.5.1",
     "dotenv": "~16.3.1",
-    "dotenv-expand": "^10.0.0",
+    "dotenv-expand": "~11.0.0",
     "ejs": "^3.1.7",
     "enhanced-resolve": "^5.8.3",
     "esbuild": "0.19.5",
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -42,7 +42,7 @@
     "cli-spinners": "2.6.1",
     "cliui": "^8.0.1",
     "dotenv": "~16.3.1",
-    "dotenv-expand": "~10.0.0",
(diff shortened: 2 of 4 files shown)


matheo:
@AgentEnder  @Coly010 your feedback here will be appreciated :pray:

xiongemi:
create a pr https://github.com/nrwl/nx/pull/26459 to replace this.

matheo:
Thank you @xiongemi!

PR #22595, opened 2024-04-01
fix(core): not passing props of run-commands to underlying command

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -8,6 +8,7 @@ import {
   getPseudoTerminal,
   PseudoTerminal,
 } from '../../tasks-runner/pseudo-terminal';
+import { output } from '../../utils/output';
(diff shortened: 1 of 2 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -301,13 +306,17 @@ async function createProcess(
     !isParallel &&
     usePty
   ) {
+    let terminalOutput = commandConfig.command + '\r\n';
```
FrozenPandaz, line 309:
```suggestion
    let terminalOutput = '> ' + commandConfig.command + '\r\n';
```

FrozenPandaz (approved):

packages/nx/src/executors/run-commands/run-commands.impl.ts
xiongemi: changed

PR #22628, opened 2024-04-02
fix(webpack): bring back previous SVG and SVGR behavior for React projects

### LINKED ISSUE #21773: SVG default import behaviour changed for webpack
### Current Behavior

As described here - https://nx.dev/recipes/react/adding-assets-react#adding-images-fonts-and-files, the default import for image assets (including svg) should be able to be used as the `src` for an `image` tag.

This appears to have changed for svg as now the default import seems to be the ReactComponent as described here - https://nx.dev/recipes/react/adding-assets-react#adding-svgs and in fact the syntax described:

`import { ReactComponent as Logo } from './logo.svg';`

actually causes an error.

### Expected Behavior

Functionality as described here - https://nx.dev/recipes/react/adding-assets-react#adding-svgs which was working using NX 17.3.2

### GitHub Repo

_No response_

### Steps to Reproduce

1.

### Nx Report

```shell
Node   : 20.11.0
   OS     : darwin-arm64
   npm    : 10.2.4

   nx                 : 18.0.4
   @nx/js             : 18.0.4
   @nx/jest           : 18.0.4
   @nx/linter         : 18.0.4
   @nx/eslint         : 18.0.4
   @nx/workspace      : 18.0.4
   @nx/cypress        : 18.0.4
   @nx/devkit         : 18.0.4
   @nx/eslint-plugin  : 18.0.4
   @nx/react          : 18.0.4
   @nx/storybook      : 18.0.4
   @nrwl/tao          : 18.0.4
   @nx/web            : 18.0.4
   @nx/webpack        : 18.0.4
   typescript         : 5.3.3
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Image assets are exported from a shared NX library:

Example `index.ts`

`export {default as logo, ReactComponent as LogoSvg} from "./lib/images/logo.svg";`

jaysoo:
The removal of `file-loader` from React's webpack configuration caused a breaking change where importing `default` like so:

```js
import svg from './example.svg';
export App() {
  return <img src={svg} alt="" />
}
```

Results in a React component via SVGR rather than a proper asset. There is no way to make asset modules from Webpack 5 work with the existing behavior, so we should introduce the new behavior later in Nx 20 as a breaking change + a migration.

## Current Behavior
SVG default export is always a React component.

## Expected Behavior
SVG assets should work for React projects.

## Related Issue(s)

Fixes #21773


packages/react/package.json
ndcunningham, line 1: This should also affect the `withNx` plugin can you also make the changes there as well?

ndcunningham (approved):

PR #22648, opened 2024-04-03
feat(core): load native files from tmp location instead of node_modules

### LINKED ISSUE #1863: Plugin interferes with the deletion of the node_modules
## Current Behavior
It is impossible to fully remove `node_modules` while in IDE 

## Expected Behavior
Unfortunately, developers in the npm ecosystem are used to removing the `node_modules` folder in some cases.
Currently, it is problematic while in IDE because the plugin locks resources in `node_modules/@nx/`.
Such adds extra overhead on communicating that IDE must be closed in order to do such things, while theoretically can be solved one time on your side.

## Steps to Reproduce
1. `npx create-nx-workspace@latest nx1672`
2. `idea nx1672` or `code nx1672`
3. `nx reset` (to stop the daemon?)
4. `rm -fo -r node_modules` (I'm using powershell)

### Failure Logs / Images / Videos

```
rm -fo -r node_modules
Cannot remove C:\nx1672\node_modules\@nx\.nx-win32-x64-msvc-kycqiCtG\nx.win32-x64-msvc.node: Access to the path 'nx.win32-x64-msvc.node' is denied.
```

### Environment

```

   Node   : 16.14.2
   OS     : win32-x64
   npm    : 9.8.1

   nx (global)        : 16.7.2
   nx                 : 16.7.2
   @nx/js             : 16.7.2
   @nx/jest           : 16.7.2
   @nx/linter         : 16.7.2
   @nx/workspace      : 16.7.2
   @nx/angular        : 16.7.2
   @nx/cypress        : 16.7.2
   @nx/devkit         : 16.7.2
   @nx/eslint-plugin  : 16.7.2
   @nrwl/tao          : 16.7.2
   @nx/webpack        : 16.7.2
   typescript         : 5.1.6
```

- Nx Console: v18.6.0
- VSCode: 1.82.0-insider 

- Nx Console: 1.11.1
- IDEA: 2023.2

MaxKless:
## Current Behavior
Currently native code is loaded straight from the package location in `node_modules` (for example a `nx.darwin-arm64.node` file). 
This is causing issues - especially on windows - when trying to delete `node_modules` or reinstalling packages while the file is still held in memory. This often surfaces in Nx Console where the daemon is started in the background. 

## Expected Behavior
Users should be able to delete `node_modules` at will without having to worry about whether the daemon is still running & locking files in the background. 

Fixes https://github.com/nrwl/nx-console/issues/1863


packages/nx/src/native/native-file-cache-location.ts
```diff
@@ -0,0 +1,19 @@
+import { existsSync, removeSync, rmSync } from 'fs-extra';
+import { tmpdir } from 'os';
+import { join } from 'path';
+
+export const nativeFileCacheLocation = join(tmpdir(), 'nx-native-cache');
```
FrozenPandaz, line 5: We don't need a separate file for this.

This should be relative to the `cacheDir`

packages/nx/src/native/import.js
```diff
@@ -0,0 +1,73 @@
...
+  if (
+    nxPackages.includes(modulePath) ||
+    localNodeFiles.some((f) => modulePath.endsWith(f))
+  ) {
+    const nativeLocation = require.resolve(modulePath);
+    const fileName = normalize(nativeLocation).split(sep).pop();
+    const fileContents = readFileSync(nativeLocation, 'utf8');
+    const hash = createHash('md5')
+      .update(nativeLocation + fileContents)
+      .digest('hex');
```
FrozenPandaz, line 54: Let's use the nxVersion

FrozenPandaz (changes requested):

FrozenPandaz (changes requested):
The js file that napi produces is set in `packages/nx/project.json`. If you update that, then you can keep the index file the correct import path.

No more lint rule
No more merge conflicts
No more weird imports

packages/nx/src/daemon/cache.ts
```diff
@@ -43,6 +43,23 @@ export async function safelyCleanUpExistingProcess(): Promise<void> {
...
+            resolve();
+          }
+          if ((count += 1) > 200) {
+            clearInterval(interval);
+            reject(
+              `Daemon process ${daemonProcessJson.processId} didn't exit after 2 seconds.`
+            );
+          }
+        }, 10);
+      });
```
FrozenPandaz, line 62: Can you leave a comment here?

The code looks a little weird at first.

On line 50 we're actually checking if the process still exists right? And it'll error if the process does not which is actually a success for what we're doing here?

packages/nx/src/daemon/client/client.ts
```diff
@@ -26,7 +26,7 @@ import { Hash } from '../../hasher/task-hasher';
 import { Task, TaskGraph } from '../../config/task-graph';
 import { ConfigurationSourceMaps } from '../../project-graph/utils/project-configuration-utils';
 import { DaemonProjectGraphError } from '../daemon-project-graph-error';
-import { ProjectGraphError } from '../../project-graph/project-graph';
+import { ProjectGraphError } from '../../project-graph/project-graph-error';
```
FrozenPandaz, line 29: There's a `project-graph/error-types.ts` that I think it would make sense to contain this.

packages/nx/src/native/index.js
```diff
@@ -1,268 +1,71 @@
...
-        } catch (e) {
-          loadError = e
-        }
-        break
-      default:
-        throw new Error(`Unsupported architecture on macOS: ${arch}`)
+Module._load = function (request, parent, isMain) {
+  const modulePath = request;
+  if (
+    nxPackages.includes(modulePath) ||
```
FrozenPandaz, line 46: Make this a `Set` so that the check is faster.

packages/nx/src/native/index.js
```diff
@@ -1,268 +1,71 @@
...
-          } else {
-            nativeBinding = require('@nx/nx-darwin-arm64')
-          }
-        } catch (e) {
-          loadError = e
-        }
-        break
-      default:
-        throw new Error(`Unsupported architecture on macOS: ${arch}`)
+Module._load = function (request, parent, isMain) {
```
FrozenPandaz, line 43: Leave more comments here in general.

packages/nx/src/project-graph/project-graph-error.ts
```diff
@@ -0,0 +1,74 @@
+import { DaemonProjectGraphError } from '../daemon/daemon-project-graph-error';
```
FrozenPandaz, line 1: This can move into `error-types.ts`

FrozenPandaz (changes requested):

packages/nx/src/project-graph/project-graph-error.ts
MaxKless: moved DaemonProjectGraphError into `error-types.ts`

packages/nx/src/native/index.js
```diff
@@ -1,268 +1,77 @@
...
+  'nx.darwin-universal.node',
+  'nx.darwin-x64.node',
+  'nx.darwin-arm64.node',
+  'nx.freebsd-x64.node',
+  'nx.linux-x64-musl.node',
+  'nx.linux-x64-gnu.node',
+  'nx.linux-arm64-musl.node',
+  'nx.linux-arm64-gnu.node',
+  'nx.linux-arm-gnueabihf.node',
+];
```
FrozenPandaz, line 39: Followup: Can we make these lists by reading `package.json`?

FrozenPandaz (approved):

PR #22649, opened 2024-04-03
feat(core): list crystal plugins with nx report

Coly010:
## Current Behavior

When running Nx Report we have no way to tell if users are using Crystal plugins or not. 
When dealing with GH Issues, this makes it more difficult to discern if problems are executor related or crystal plugin related.

## Expected Behavior

We should list the Crystal plugins that are being used in the Nx Workspace when `nx report` is run.
This will not completely alleviate the issues with knowing whether executors are being used or not, but it will help at least.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/report/report.ts
+++ b/packages/nx/src/command-line/report/report.ts
@@ -57,6 +57,7 @@ export async function reportHandler() {
     pmVersion,
     localPlugins,
     communityPlugins,
+    registeredPlugins,
     packageVersionsWeCareAbout,
     outOfSyncPackageGroup,
     projectGraphError,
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/report/report.ts
```diff
@@ -268,6 +281,17 @@ export function findInstalledCommunityPlugins(): PackageJson[] {
       )
   );
 }
+
+export function findCrystalPluginsBeingUsed(nxJson: NxJsonConfiguration) {
```
FrozenPandaz, line 285:
```suggestion
export function findRegisteredPlugins(nxJson: NxJsonConfiguration) {
```

packages/nx/src/command-line/report/report.ts
```diff
@@ -77,6 +78,14 @@ export async function reportHandler() {
     );
   });
 
+  if (crystalPlugins.length) {
+    bodyLines.push(LINE_SEPARATOR);
+    bodyLines.push('Nx Crystal Plugins:');
```
FrozenPandaz, line 83: Please change this to `Registered Plugins`.

Crystal is just a codename, it shouldn't be shown in output.

FrozenPandaz (changes requested):

FrozenPandaz:
Can you please show the before and after of the terminal output?

Coly010:
Before
image


After
image

FrozenPandaz (approved):

PR #22679, opened 2024-04-04
feat(core): experimental wasi target

FrozenPandaz:
Originally opened by @Brooooooklyn

This pull request is trying to add `wasm32-wasi` target support for the `nx/native`

To test the build, you can run the following commands:

- `pnpm exec napi build --release --platform -c packages/nx/package.json --manifest-path packages/nx/Cargo.toml --js ./index.js -o packages/nx/src/native --target wasm32-wasi-preview1-threads`

And the `wasm` file will be built at `packages/nx/src/native/nx.wasm32-wasi.wasm`

Blocked by:
- https://github.com/Cammisuli/monodon/issues/48
- https://github.com/napi-rs/napi-rs/issues/2009

The  `pseudo_terminal` mod is excluded on the `wasm32` targets, which is as expected. 

The `watch` mod is excluded because of the upstream `gix-*` deps introduce by `ignore-files` don't support wasi target at this moment (but we can improve it).


packages/nx/src/native/index.js
```diff
@@ -1,268 +1,355 @@
...
-          loadError = e
-        }
-        break
-      case 'arm':
-        localFileExisted = existsSync(join(__dirname, 'nx.android-arm-eabi.node'))
-        try {
-          if (localFileExisted) {
-            nativeBinding = require('./nx.android-arm-eabi.node')
-          } else {
-            nativeBinding = require('@nx/nx-android-arm-eabi')
```
FrozenPandaz, line 46: These imports have been broken some how.

The prefix is specified in `package.json` here: https://github.com/nrwl/nx/blob/fc8d5ba828118300bf4c4927cdb7578c46b5b355/packages/nx/package.json#L170 but is not being picked up by napi for some reason. Could you take a look please? :eyes: I imagine it's a result of upgrading Napi to the 3.0.0 version.

FrozenPandaz: @Brooooooklyn Do you know what might be going on here?

Brooooooklyn: I'll take a look tomorrow 🥹

Brooooooklyn: @FrozenPandaz it should be `napi.packageName` now

FrozenPandaz:
Closing in favor of #22870

PR #22694, opened 2024-04-05
feat(core): improve checks for process.env.CI

luckened:
## Current Behavior
Today, nx does not disable the output styling and [WoodpeckerCI](https://woodpecker-ci.org/) doesn't go well with it, This requires us to manually add "output-style=stream" to every nx command, which solves the issue but I understand that we could have a woodpecker check to isCI function as well

reference:
https://woodpecker-ci.org/docs/usage/environment#built-in-environment-variables

## Expected Behavior
Disable styling in nx command outputs when running in WoodpeckerCI

## Related Issue(s)
I haven't found a related issue, but this PR came as a suggestion in this discord thread: https://discord.com/channels/1143497901675401286/1225808236104646706

@ edit

the title is "feat(core): add WoodpeckerCI to isCI check" but the check also applies to other CI vendors that set process.env.CI to a non-empty string

@ edit2

I've improved the PR title with feat(core): improve checks for process.env.CI since it now covers all CI players that dont'set process.env.CI directly to "true" but to other values instead

### DIFF
--- a/packages/create-nx-workspace/src/utils/ci/is-ci.ts
+++ b/packages/create-nx-workspace/src/utils/ci/is-ci.ts
@@ -1,6 +1,6 @@
 export function isCI() {
   return (
-    process.env.CI === 'true' ||
+    (process.env.CI && process.env.CI !== 'false') ||
     process.env.TF_BUILD === 'true' ||
     process.env.GITHUB_ACTIONS === 'true' ||
     process.env.BUILDKITE === 'true' ||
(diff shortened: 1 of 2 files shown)


AgentEnder (changes requested):
Thanks for the PR! I think we can do this in a bit more agnostic way.

Rather than adding an additional check for woodpecker specifically, can you update the earlier check from `process.env.CI === 'true'` to `(process.env.CI && process.env.CI !== 'false')`

luckened:
@AgentEnder thank you for the review!

I see, I'm thinking if its okay to just check if `process.env.CI` exists, as I believe process.env.CI is not "false" locally, does this make sense?

AgentEnder (approved):

PR #22729, opened 2024-04-08
feat(react-native): upgrade react native to 0.72.6

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react-native/migrations.json
+++ b/packages/react-native/migrations.json
@@ -627,6 +627,47 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "19.0.0": {
+      "version": "19.0.0-beta.7",
+      "packages": {
+        "react-native": {
+          "version": "0.73.6",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-native-community/cli-platform-android": {
+          "version": "~12.3.6",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-native/babel-preset": {
+          "version": "^0.73.21",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-native/metro-config": {
+          "version": "^0.73.5",
+          "alwaysAddToPackageJson": false
+        },
+        "react-native-web": {
+          "version": "^0.19.10",
+          "alwaysAddToPackageJson": false
+        },
+        "@testing-library/react-native": {
+          "version": "~12.4.5",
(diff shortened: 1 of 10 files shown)


jaysoo (approved):

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -252,7 +252,7 @@ async function createPreset(tree: Tree, options: Schema) {
       directory: join('apps', options.name),
```
FrozenPandaz, line 252: There are some console logs up here... can you remove those please?

FrozenPandaz (changes requested):

packages/workspace/src/generators/preset/preset.ts
xiongemi: removed

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -252,7 +252,7 @@ async function createPreset(tree: Tree, options: Schema) {
...
       projectNameAndRootFormat: 'as-provided',
       linter: options.linter,
-      e2eTestRunner: options.e2eTestRunner ?? 'detox',
+      e2eTestRunner: options.e2eTestRunner ?? 'playwright',
```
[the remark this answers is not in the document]
xiongemi: there is a pr to change the default e2e to playwright https://github.com/nrwl/nx/pull/22511, so I changed it here.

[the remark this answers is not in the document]
FrozenPandaz: Okay, but we can keep detox here for now.

packages/react-native/migrations.json
```diff
@@ -627,6 +627,47 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "19.0.0": {
+      "version": "19.0.0-beta.0",
```
FrozenPandaz, line 632: Can you please update these migration versions to `19.0.0-beta.6` since `19.0.0-beta.5` is already out please?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

Balake:
Isn't the title of this MR incorrect? Was looking at the changes for 19 beta and saw "upgrade react native to 0.72.6", but I think it's supposed to be 0.73.6.

PR #22747, opened 2024-04-09
docs(core): remove old Nx Console recipes

jaysoo:
This PR removes three pages that aren't useful.

1. [Keyboard shortcuts](https://nx.dev/recipes/nx-console/console-shortcuts) - This only applies to VS Code, and it isn't an Nx Console features. It's well documented on VS Code docs site already.
2. [Console Project Pane](https://nx.dev/recipes/nx-console/console-project-pane) - This page is basically unused. Also, `project.json` is de-emphasized with Project Crystal, and this page doesn't show much beyond what users can already see in Console.
3. [Add Dependency command](https://nx.dev/recipes/nx-console/console-add-dependency-command) - Another page that is basically unused. This only applies to VS Code, and isn't useful to document since 

## Related Issue(s)

Fixes #


isaacplmann (changes requested):
Are these recipes invalid now?

If we do need to remove them, there should be a redirect rule added

isaacplmann (approved):

PR #22763, opened 2024-04-10
fix(core): should not pass --no-color --no-parallel

### LINKED ISSUE #22321: --no-color CLI option

## Description

Today, when using commands such as `nx build`, it is not possible to disable colors in the console. 

## Motivation

Today I am running an NX build from a Husky pre-commit hook. 

When this hook finds errors on commit when committing via the UI in VS Code, the resulting hook output in the GIT output window has strange characters in it. This is a known problem with VS Code. But disabling colors in this scenario would be a fine workaround. 

## Suggested Implementation

I suggest adding a `--no-colors` option to the NX CLI to support this. This is also in line with the Angular CLI and many other CLIs with that same option.

## Alternate Implementations

It'd be great if the VS Code team would be interested in fixing the root problem to begin with. But it doesn't seem like it. The issues on it are just getting closed.

xiongemi:
## Current Behavior

- pass --no-parallel --no-color to underlying command

## Expected Behavior

- should handle --no-parallel and --no-color

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/22321

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
@@ -36,16 +36,85 @@ describe('Run Commands', () => {
     expect(readFile(f)).toEqual('123');
   });
 
-  it('should not pass --args into underlying command', async () => {
-    const result = await runCommands(
-      {
-        command: `echo`,
-        __unparsed__: ['--args=--key=123'],
-        args: '--key=123',
-      },
-      context
-    );
-    expect(result.terminalOutput.trim()).not.toContain('--args=--key=123');
+  it.each([
+    {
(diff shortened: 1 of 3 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
```diff
@@ -200,7 +276,7 @@ describe('Run Commands', () => {
           } as any,
           true
         )
-      ).toEqual('echo hello');
+      ).toEqual('echo test hello');
```
FrozenPandaz, line 279: This seems broken..

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #22767, opened 2024-04-10
docs(core): remove deprecated docs page for Nx 14 and earlier

jaysoo:
This PR removes pages that have not been relevant since Nx 14, or have been removed since Nx 17 and have almost no traffic. By removing outdated pages, it prevents searches such as `react storybook` from surfacing pages that are irrelevant.

## Related Issue(s)

Fixes #


isaacplmann (approved):
Looks good to me if @mandarini thinks the storybook pages are no longer needed.

Issue #22769, opened 2024-04-10
`@nx/eslint-pluggin` throws an error in Eslint 9


olaf-cichocki:
### Current Behavior

With Eslint 9 we get two errors: 

![image](https://github.com/nrwl/nx/assets/12986906/63f20c49-c58f-402f-aa22-44f90b227f8f)
![image](https://github.com/nrwl/nx/assets/12986906/11cea1a3-7e70-4f8c-86d9-79b47c2d0a6b)

### Expected Behavior

I believe no errors should be shown

### Steps to Reproduce

1. Update eslint to 9
2. Run lint in a package with uses `@nx/dependency-checks` rule

### Nx Report

```shell
Node   : 20.6.1
OS     : darwin-arm64
pnpm   : 8.6.10

nx                 : 18.2.2
@nx/js             : 18.2.2
@nx/linter         : 18.2.2
@nx/eslint         : 18.2.2
@nx/workspace      : 18.2.2
@nx/devkit         : 18.2.2
@nx/eslint-plugin  : 18.2.2
@nx/playwright     : 18.2.2
@nrwl/tao          : 18.2.2
typescript         : 4.7.4
```

### Failure Logs

```shell
✖  nx run @cld/npm-scripts:lint
      > @cld/npm-scripts@2.10.100 lint /Users/olaf/Repositories/cld-web/libs/dev/npm-scripts
      > eslint .


      Oops! Something went wrong! :(

      ESLint: 8.57.0

      Error: Key "plugins": Cannot redefine plugin "@typescript-eslint".
          at /Users/olaf/Repositories/cld-web/node_modules/.pnpm/@humanwhocodes+object-schema@2.0.2/node_modules/@humanwhocodes/object-schema/src/object-schema.js:250:27
          at Array.reduce (<anonymous>)
          at ObjectSchema.merge (/Users/olaf/Repositories/cld-web/node_modules/.pnpm/@humanwhocodes+object-schema@2.0.2/node_modules/@humanwhocodes/object-schema/src/object-schema.js:237:24)
          at /Users/olaf/Repositories/cld-web/node_modules/.pnpm/@humanwhocodes+config-array@0.11.14/node_modules/@humanwhocodes/config-array/api.js:935:42
          at Array.reduce (<anonymous>)
[...]

PR #22770, opened 2024-04-10
fix(nx-plugin): fix eslint 9 errors

olaf-cichocki:
## Current Behavior

With Eslint 9 we get two errors

image
image

## Expected Behavior
No unexpected errors

## Related Issue(s)
https://github.com/nrwl/nx/issues/22769

## Fixes
- fix `Cannot read properties of undefined (reading 'isJSON')` error 
- fix filename context in eslint 9

## Other
I confirmed that this would fix the issues in our monorepo by manually modifying the `dependency-checks` and confirming ti solves the error

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -96,10 +96,10 @@ export default ESLintUtils.RuleCreator(
       },
     ]
   ) {
-    if (!(context.parserServices as any).isJSON) {
+    if (!(context.sourceCode.parserServices as any).isJSON) {
       return {};
     }
-    const fileName = normalizePath(context.getFilename());
+    const fileName = normalizePath(context.filename ?? context.getFilename());
     // support only package.json
     if (!fileName.endsWith('/package.json')) {
       return {};


packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -96,10 +96,10 @@ export default ESLintUtils.RuleCreator(
       },
     ]
   ) {
-    if (!(context.parserServices as any).isJSON) {
+    if (!(context.sourceCode.parserServices as any).isJSON) {
```
olaf-cichocki, line 99: `parserServices` was moved to the `context.sourceCode` and `context.parserServices` was deprecated

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -96,10 +96,10 @@ export default ESLintUtils.RuleCreator(
...
       return {};
     }
-    const fileName = normalizePath(context.getFilename());
+    const fileName = normalizePath(context.filename ?? context.getFilename());
```
olaf-cichocki, line 102: In Eslint 9 `context.getFilename()` is now `context.filename`

JamesHenry:
Thanks a lot for working on this @olaf-cichocki but there's a bit more that we need related to v9 support. I've opened https://github.com/nrwl/nx/pull/24632 so please follow along with that

PR #22775, opened 2024-04-11
fix(vite): get tsconfig from new path including target

ghost:
## Current Behavior
Since v18.2.3 (e4c4697f69cf29c50aced1997715a7da5f680e23) the `tsconfig.generated.json` file generated if `"buildLibsFromSource": false` is placed under a path containing the current build target. The `nx-tsconfig-paths` plugin of `@nx/vite` was not updated accordingly, so now it doesn't find `tsconfig.generated.json`, effectively causing `@nx/vite` to ignore the `"buildLibsFromSource": false` setting.

## Expected Behavior
With this PR, `nx-tsconfig-paths` finds `tsconfig.generated.json` at the correct path, so `"buildLibsFromSource": false` works as it did before v18.2.3.


Coly010 (approved):
Good catch!

PR #22784, opened 2024-04-11
feat(testing): add playwright generator to convert from executors to plugin

Coly010:
## Current Behavior

The only way to migrate a project from playwright executor usage to playwright plugin usage is to do so manually.

## Expected Behavior

Add a generator that will convert projects from playwright executor usage to playwright plugin usage

### Usage

`nx g @nx/playwright:convert-to-inferred --project="myapp-e2e"` // Convert only `myapp-e2e` from executor to plugin usage
`nx g @nx/playwright:convert-to-inferred` // Migrate all projects

## Related Issue(s)

Fixes #


packages/devkit/src/generators/plugin-migrations/plugin-migration-utils.ts
```diff
@@ -0,0 +1,398 @@
...
+    tree: Tree,
+    projectRoot: string,
+    targetName: string,
+    context: CreateNodesContext
+  ) => CreateNodesResult | Promise<CreateNodesResult>,
+  executor: string,
+  createNodes: CreateNodes<T>,
+  projectOptionsTransformer: (
+    target: TargetConfiguration
+  ) => TargetConfiguration = (targetConfiguration) => targetConfiguration
```
FrozenPandaz, line 100: Update this to your new version with `tree`, etc.

packages/devkit/src/generators/plugin-migrations/plugin-migration-utils.ts
```diff
@@ -0,0 +1,398 @@
...
+      createdProject.targets[targetName];
+    delete createdTarget.command;
+    delete createdTarget.options?.cwd;
+
+    const projectConfig = readProjectConfiguration(tree, projectName);
+    let target = projectConfig.targets[targetName];
+
+    target = mergeTargetConfigurations(target, targetDefaultsForExecutor);
+    delete target.executor;
+    delete target.options?.config;
```
FrozenPandaz, line 150: Move this to the `postTargetTransformer`

packages/playwright/src/generators/convert-to-inferred/lib/migrate-executor-to-plugin.ts
```diff
@@ -0,0 +1,107 @@
...
+      tree,
+      '@nx/playwright/plugin',
+      include,
+      { targetName, ciTargetName: 'e2e-ci' }
+    );
+
+    for (const project of migratedProjects) {
+      allProjectsWithExecutor.delete(project);
+    }
+  }
```
FrozenPandaz, line 64: Move this into devkit.

FrozenPandaz (changes requested):

packages/playwright/src/generators/convert-to-inferred/convert-to-inferred.ts
```diff
@@ -0,0 +1,38 @@
...
+  let project: ProjectConfiguration;
+  if (options.project) {
+    project = readProjectConfiguration(tree, options.project);
+  }
+
+  await migrateExecutorToPlugin(tree, project?.name, project?.root);
+
+  if (!options.skipFormat) {
+    await formatFiles(tree);
+  }
```
FrozenPandaz, line 35: This should basically look like...

1. Get the graph
2. Call the devkit function
3. Format Files

packages/devkit/src/generators/plugin-migrations/plugin-migration-utils.ts
```diff
@@ -0,0 +1,521 @@
...
+      workspaceRoot: tree.root,
+      nxJsonConfiguration,
+      configFiles,
+    };
+    const projectConfigurations = await createProjectsConfig(
+      tree,
+      projectFromGraph.data.root,
+      targetName,
+      context
+    );
```
FrozenPandaz, line 233: This is weird... keep it for now... we should remove this ideally before we do too many plugins.

Let's get an initial version in.. and start using it.

packages/devkit/src/generators/plugin-migrations/plugin-migration-utils.ts
```diff
@@ -0,0 +1,521 @@
...
+    return;
+  }
+
+  nxJson.plugins.push({
+    plugin: pluginPath,
+    include,
+    options: pluginOptions,
+  });
+
+  updateNxJson(tree, nxJson);
```
FrozenPandaz, line 353: Let's change this to update the existing plugin to update includes.

At the end check again if include includes every configfile... if it does... we can then remove the include property.

We will need update the playwright init.

FrozenPandaz (changes requested):

isaacplmann (approved):

FrozenPandaz (approved):

PR #22792, opened 2024-04-12
docs(expo): fix the instructions for rejecting local cache

jaypeeig:
## Current Behavior

The current instructions are misleading, for example, NX REJECT_UNKNOWN_LOCAL_CACHE=0, which won't function if we implement it; it should be NX_REJECT_UNKNOWN_LOCAL_CACHE=0.

## Expected Behavior

The instructions should be consistent with the code implementation.

https://github.com/search?q=repo%3Anrwl%2Fnx%20REJECT_UNKNOWN_LOCAL_CACHE&type=code


leosvelperez (approved):
Thanks for contributing!

PR #22801, opened 2024-04-12
feat(angular): generate `public` directory instead of `assets`

leosvelperez:
Upstream change: https://github.com/angular/angular-cli/commit/6530aa11bed5ef67d611e8aed268bd20345cf0e6

## Related Issue(s)

Fixes #


e2e/angular-core/src/ng-add.test.ts
```diff
@@ -76,7 +76,7 @@ describe('convert Angular CLI workspace to an Nx workspace', () => {
     cleanupProject();
   });
 
-  it('should generate a workspace', () => {
+  it('hhhhhhshould generate a workspace', () => {
```
Coly010, line 79: Leftover to cleanup

Coly010 (changes requested):
We should also update the host generator (might be setup-mf) when dynamic is true to generate the module federation manifest in public directory, as well as update the bootstrap template to fetch from the correct directory

Coly010 (approved):
LGTM!

PR #22819, opened 2024-04-15
fix(js): do not default to commonjs type field in package.json

### LINKED ISSUE #19384: NX automatically adds `"type": "commonjs"` to package.json when using generators, which breaks imports with 16.8 changes
### Current Behavior

New packages created with generators have `"type": "commonjs"` in their `package.json`, which in nx@v16.8 prevents them from being imported by some projects (for example, a vanilla `create-react-app` project) if `"generateExportsField"` is also set to true.

```shell
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (1:0)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/source-map-loader/dist/cjs.js
You may need an additional loader to handle the result of these loaders.
```

**I'm able to work around this by explicitly setting my package.json's `"type": undefined` in my generator, but this feels like a pretty easy thing to miss.**

Since everything seems to work fine without `"type": "commonjs"` (I tested it in both ESM and commonjs consumers), should this simply not be added automatically?

### Expected Behavior

New packages created with generators have able to be imported into React projects as ES modules.

### GitHub Repo

_No response_

### Steps to Reproduce

- use a  generator based on `libraryGenerator` to create a package in a monorepo
- use `executor: '@nx/rollup:rollup'` and `generateExportsField: true` in the package's build
- observe `"type": "commonjs"` is added automatically to the `dist/package.json` when built
- use `npx create-react-app` to create a barebones react app
- observe that artifacts from the nx package cannot be imported due to above module error

### Nx Report

```shell

   Node   : 18.13.0
   OS     : linux-x64
   npm    : 9.7.2
   
   nx                 : 16.9.1
   @nx/js             : 16.9.1
   @nx/jest           : 16.9.1
   @nx/linter         : 16.9.1
   @nx/workspace      : 16.9.1
   @nx/devkit         : 16.9.1
   @nx/eslint-plugin  : 16.9.1
   @nx/plugin         : 16.9.1
   @nx/rollup         : 16.9.1
   @nrwl/tao          : 16.9.1
   @nx/web            : 16.9.1
   nx-cloud           : 16.4.0
   typescript         : 5.1.6
   Local workspace plugins:
         @js-sdk-contrib/workspace-plugin
```

### Failure Logs

```shell
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (1:0)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/source-map-loader/dist/cjs.js
You may need an additional loader to handle the result of these loaders.
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Workaround: https://github.com/open-feature/js-sdk-contrib/pull/596

Coly010:
## Current Behavior

The `@nx/js:library` generator will default the `packageJson.type` field to `commonjs` in the following scenarios:

 - Rollup Bundler is used
 - No bundler is used

#### Rollup Behaviour

Rollup supports generating both ESM and CJS output files. The `build` target that is created even specifies that both formats should be used.
However, the `@nx/rollup:rollup` executor will restrict the output format based on the `packageJson.type` field.

Therefore, we're always forcing CJS. This can lead to further issues when the package is built and distributed to other repos.

The `skipTypeField` option is also not respected in `commonjs` output formats.

#### No Bundler Behaviour

When no bundler is specified for the library, we still add `packageJson.type = commonjs`.
This is unnecessarily restrictive and can lead to further behavioural issues.

It is safer to not set a `type` field and allow the user to specify this themselves based on their needs.
 
## Expected Behavior

Do not default to `commonjs` for no bundler or with rollup.

## Related Issue(s)

Fixes #19384

### DIFF
--- a/packages/js/src/generators/library/library.ts
+++ b/packages/js/src/generators/library/library.ts
@@ -869,10 +869,10 @@ function determineEntryFields(
       };
     case 'rollup':
       return {
-        type: 'commonjs',
+        // Since we're publishing both formats, skip the type field.
+        // Bundlers or Node will determine the entry point to use.
         main: './index.cjs',
         module: './index.js',
-        // typings is missing for rollup currently
       };
     case 'vite':
       return {
@@ -891,9 +891,9 @@ function determineEntryFields(
       };
     default: {
       return {
-        // CJS is the safest optional for now due to lack of support from some packages
-        // also setting `type: module` results in different resolution behavior (e.g. import 'foo' no longer resolves to 'foo/index.js')
-        type: 'commonjs',
+        // Safest option is to not set a type field.
+        // Allow the user to decide which module format their library is using
+        type: undefined,
       };
     }
   }
--- a/packages/js/src/generators/library/library.spec.ts
+++ b/packages/js/src/generators/library/library.spec.ts
@@ -44,7 +44,6 @@ describe('lib', () => {
         name: '@proj/my-lib',
         private: true,
         version: '0.0.1',
-        type: 'commonjs',
         scripts: {
           build: "echo 'implement build'",
           test: "echo 'implement test'",
@@ -977,6 +976,9 @@ describe('lib', () => {
         expect(config.targets.build.options.project).toEqual(
           `my-lib/package.json`
         );
+
+        const pkgJson = readJson(tree, 'my-lib/package.json');
+        expect(pkgJson.type).not.toBeDefined();
       });
 
       it('should always set compiler to swc if bundler is rollup', async () => {
--- a/packages/rollup/src/executors/rollup/lib/update-package-json.ts
+++ b/packages/rollup/src/executors/rollup/lib/update-package-json.ts
@@ -52,7 +52,7 @@ export function updatePackageJson(
 
     packageJson.main = cjsExports['.'];
 
-    if (!hasEsmFormat) {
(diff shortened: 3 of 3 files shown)


packages/js/src/generators/library/library.ts
```diff
@@ -869,10 +869,11 @@ function determineEntryFields(
       };
     case 'rollup':
       return {
-        type: 'commonjs',
+        // Since we're publishing both formats, skip the type field.
+        // Bundlers or Node will determine the entry point to use.
         main: './index.cjs',
         module: './index.js',
-        // typings is missing for rollup currently
+        typings: './src/index.d.ts',
```
jaysoo, line 876: We don't need to set `typings` since it should automatically match based on file name. It's only needed if the paths/names are different, e.g. `lib/index.js` vs `types/index.d.ts`. Also, if `generateExportsField` is specified, the types should match on those entries not `typings`.

Coly010: Perfect, I can tidy that up

jaysoo (approved):

PR #22843, opened 2024-04-16
fix(core): fix init logging and package.json updates

FrozenPandaz:
## Current Behavior

The "final" message for the init flow is not always at the bottom of the logs. For the new init flow, it gets covered up by all the init logs. Also, for npm monorepos, all `package.json` files have `nx: {}` added which should not be the case for any `package.json` files in the case of a monorepo.

## Expected Behavior

The final message for the init flow appears at the bottom of the logs. All the `package.json` files are also not touched for a npm workspace case.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/init/implementation/utils.ts
+++ b/packages/nx/src/command-line/init/implementation/utils.ts
@@ -206,20 +206,16 @@ export function markPackageJsonAsNxProject(packageJsonPath: string) {
 
 export function printFinalMessage({
   learnMoreLink,
-  bodyLines,
 }: {
   learnMoreLink?: string;
-  bodyLines?: string[];
 }): void {
-  const normalizedBodyLines = (bodyLines ?? []).map((l) =>
-    l.startsWith('- ') ? l : `- ${l}`
-  );
+  const pmc = getPackageManagerCommand();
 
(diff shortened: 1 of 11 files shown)


packages/nx/src/command-line/init/implementation/utils.ts
```diff
@@ -214,11 +214,15 @@ export function printFinalMessage({
...
   );
+  const pmc = getPackageManagerCommand();
 
   output.success({
     title: '🎉 Done!',
     bodyLines: [
-      '- Enabled computation caching!',
+      `- Run "${pmc.exec} nx run-many --t build" to run the build target for every project in the workspace. Run it again to replay the cached computation.`,
+      `  - https://nx.dev/features/cache-task-results`,
+      `- Run "${pmc.exec} nx graph" to see the graph of the monorepo. It will show tasks that you can run with Nx.`,
```
jaysoo, line 224: `graph of the monorepo` - it may be a standalone workspace

Maybe reword to `Run "${pmc.exec} nx graph" to explore the projects and tasks in your workspace.

jaysoo (approved):

AgentEnder (approved):

PR #22887, opened 2024-04-18
chore(misc): update message in legacy readme files

jaysoo:
We no longer have plans to discontinue the legacy packages in the near future. This PR updates the message to reflect this.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages-legacy/angular/README.md
+++ b/packages-legacy/angular/README.md
@@ -4,7 +4,7 @@
 
 [Read more here](https://nx.dev/recipes/other/rescope)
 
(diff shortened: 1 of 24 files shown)


Phillip9587:
Why is this being pushed back again?

jaysoo:
> Why is this being pushed back again?

To allow more time for people to migrate off of them. There is still a gap between `@nrwl/` and `@nx/` installs. There won't be any effect on users already migrated.

FrozenPandaz (approved):

PR #22897, opened 2024-04-19
fix(gradle): should not cache buildDependents

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/plugin/nodes.ts
+++ b/packages/gradle/src/plugin/nodes.ts
@@ -185,7 +185,8 @@ function createGradleTargets(
       options: {
         cwd: projectRoot,
       },
-      cache: cacheableTaskType.has(task.type),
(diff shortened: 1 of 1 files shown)


packages/gradle/src/plugin/nodes.ts
```diff
@@ -185,7 +185,8 @@ function createGradleTargets(
       options: {
         cwd: projectRoot,
       },
-      cache: cacheableTaskType.has(task.type),
+      cache:
+        cacheableTaskType.has(task.type) && task.name !== 'buildDependents',
```
FrozenPandaz, line 189: Is there something else we can use to distinguish `buildDependents` from other `build` tasks?

FrozenPandaz (changes requested):

PR #22944, opened 2024-04-22
fix(linter): do not infer lint tasks for projects without files to lint

leosvelperez:
## Current Behavior

`@nx/eslint/plugin` infers tasks for projects without files to lint. It doesn't properly consider the ESLint ignored files configuration.

## Expected Behavior

`@nx/eslint/plugin` should not infer tasks for projects without files to lint.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint/src/plugins/plugin.ts
+++ b/packages/eslint/src/plugins/plugin.ts
@@ -4,6 +4,7 @@ import {
   CreateNodesResult,
   TargetConfiguration,
 } from '@nx/devkit';
+import type { ESLint } from 'eslint';
 import { existsSync } from 'node:fs';
 import { dirname, join } from 'node:path';
 import { combineGlobPatterns } from 'nx/src/utils/globs';
@@ -28,18 +29,30 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
     baseEsLintConfigFile,
     baseEsLintFlatConfigFile,
   ]),
-  (configFilePath, options, context) => {
+  async (configFilePath, options, context) => {
     options = normalizeOptions(options);
 
     // Ensure that configFiles are set, e2e-run fails due to them being undefined in CI (does not occur locally)
     // TODO(JamesHenry): Further troubleshoot this in CI
     (context as any).configFiles = context.configFiles ?? [];
 
-    // Create a Set of all the directories containing eslint configs
-    const eslintRoots = new Set(context.configFiles.map(dirname));
-    const configDir = dirname(configFilePath);
+    // Create a Set of all the directories containing eslint configs, and a
+    // list of globs to exclude from child projects
+    const eslintRoots = new Set();
+    const filesToExclude: string[] = [];
+    for (const configFile of context.configFiles) {
+      const configDir = dirname(configFile);
+      eslintRoots.add(configDir);
+      if (configDir !== '.' && configDir !== dirname(configFilePath)) {
+        filesToExclude.push(`${configDir}/**/*`);
+      }
+    }
 
-    const childProjectRoots = globWithWorkspaceContext(
+    const configDir = dirname(configFilePath);
+    const childProjectRoots = new Set<string>();
+    const ESLint = resolveESLintClass(isFlatConfig(configFilePath));
+
+    const projectFiles = globWithWorkspaceContext(
       context.workspaceRoot,
       [
         'project.json',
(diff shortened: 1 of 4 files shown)


packages/eslint/src/plugins/plugin.ts
```diff
@@ -28,48 +29,72 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
...
+    // Create a Set of all the directories containing eslint configs, and a
+    // list of globs to exclude from child projects
+    const eslintRoots = new Set();
+    const filesToExclude: string[] = [];
+    for (const configFile of context.configFiles) {
+      const configDir = dirname(configFile);
+      eslintRoots.add(configDir);
+      if (configDir !== '.' && configDir !== dirname(configFilePath)) {
+        filesToExclude.push(`${configDir}/**/*`);
+      }
```
FrozenPandaz, line 48: I believe there's a possibility where a parent `configDir` of the `configFilePath` will get added to the `filesToExclude` paths. This would ignore everything including the files in the directory as well.

If `configFiles` is for `libs/a/.eslintrc.json`

```ts
[
  `.eslintrc.json`, // This isn't added because it's the root...
  `libs/.eslintrc.json`, // This one does get added...
  `libs/a/`.eslintrc.json`, // This is the actual config file.
]
```

packages/eslint/src/plugins/plugin.ts
```diff
@@ -28,48 +29,72 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
...
+          continue projectFileLoop;
         }
-        return true;
-      })
-      .filter((dir) => {
-        // Ignore project roots where the project does not contain any lintable files
-        const lintableFiles = globWithWorkspaceContext(context.workspaceRoot, [
-          join(dir, `**/*.{${options.extensions.join(',')}}`),
-        ]);
-        return lintableFiles.length > 0;
```
FrozenPandaz, line 69: So this never handled if there were lintable files.. but they were all under a child config path right?

packages/eslint/src/plugins/plugin.ts
```diff
@@ -153,6 +180,9 @@ function buildEslintTargets(
           isRootProject ? '{projectRoot}/' : '{projectRoot}'
         )
       ),
+      ...(existsSync(join(workspaceRoot, projectRoot, '.eslintignore'))
+        ? ['{projectRoot}/.eslintignore']
+        : []),
```
FrozenPandaz, line 185: This includes `proj/.eslintignore` as an input?

Should it also include anything in a parent directory.. as well as anything in a child directory?

FrozenPandaz (changes requested):

packages/eslint/src/plugins/plugin.ts
leosvelperez: Good catch! I only need to ignore the nested ones. I'll amend it and add tests to cover that.

packages/eslint/src/plugins/plugin.ts
leosvelperez: Exactly. It never handled the ESLint ignored file configuration. It did glob for children with specific file extensions, which might or might not have been lintable for the project.

packages/eslint/src/plugins/plugin.ts
leosvelperez: Only the `.eslintignore` file at the project root (cwd) applies (https://eslint.org/docs/latest/use/configure/ignore-deprecated#the-eslintignore-file). Other `.eslintignore` are not used by ESLint.

packages/eslint/src/plugins/plugin.ts
```diff
@@ -28,48 +29,64 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
...
+      );
+      const eslint = new ESLint({
+        cwd: join(context.workspaceRoot, childProjectRoot),
       });
+      for (const file of lintableFiles) {
+        if (!(await eslint.isPathIgnored(join(context.workspaceRoot, file)))) {
+          childProjectRoots.add(childProjectRoot);
+          break;
+        }
+      }
```
FrozenPandaz, line 86: We should probably use a `Promise.all` here. But you can address it in a followup PR.

packages/eslint/src/plugins/plugin.ts
```diff
@@ -28,48 +29,64 @@ export const createNodes: CreateNodes<EslintPluginOptions> = [
...
-      .filter((dir) => {
-        // Ignore project roots where the project does not contain any lintable files
-        const lintableFiles = globWithWorkspaceContext(context.workspaceRoot, [
-          join(dir, `**/*.{${options.extensions.join(',')}}`),
-        ]);
-        return lintableFiles.length > 0;
+      ].map((f) => join(configDir, f)),
+      nestedEslintRootPatterns.length ? nestedEslintRootPatterns : undefined
+    );
+    for (const projectFile of projectFiles) {
```
FrozenPandaz, line 66: This loop could also be a `Promise.all`

FrozenPandaz (approved):

PR #22946, opened 2024-04-22
chore(repo): add eslint and jest plugins to use inferred targets

jaysoo:
This PR adds `@nx/eslint/plugin` and `@nx/jest/plugin` to use inferred targets for lint and unit tests. The main changes are removing targets from `project.json` and `nx.json`. There is a fix in the `patch-jest-resolver.js` file to ensure that all modules are resolved to an absolute path, which is needed for `jest.mock` to work. A couple of snapshots are updated too with `cwd` changes.

**Note:** We can now run any Jest spec file from IDE since the executor and cwd are no longer assumed.


.nxignore
```diff
@@ -1,2 +1,3 @@
 nx-dev/**/jest.config.js
 .next
+__fixtures__
```
jaysoo, line 3: This is needed because some of our tests use `__fixtures__` that contain jest config files, and we do not want to create targets/projects from them.

.eslintrc.json
```diff
@@ -28,7 +28,8 @@
     "storybook/no-uninstalled-addons": [
       "error",
       {
-        "ignore": ["@nx/react/plugins/storybook"]
+        "ignore": ["@nx/react/plugins/storybook"],
+        "packageJsonLocation": "../../package.json"
```
jaysoo, line 32: Need to point the Storybook lint rule to the root `package.json`, or else it'll pick up the package one (e.g. `packages/vue/package.json`) which results in missing dependencies being reported.

.nxignore
FrozenPandaz: Hm we don't want to `.nxignore` them... I would add them to the `exclude` array along with `e2e`

nx.json
```diff
@@ -215,6 +187,19 @@
...
+    },
+    {
+      "plugin": "@nx/eslint/plugin",
+      "options": {
+        "targetName": "lint"
+      }
+    },
+    {
+      "plugin": "@nx/jest/plugin",
+      "include": ["packages/**/*", "graph/**/*", "nx-dev/**/*", "tools/**/*"],
```
FrozenPandaz, line 199: I would just include everything except for `e2e/**/*`

packages/remix/src/generators/application/__snapshots__/application.impl.spec.ts.snap
```diff
@@ -189,7 +189,7 @@ export default defineConfig({
   },
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: 'pnpm exec nx dev test',
+    command: 'npx nx dev test',
```
FrozenPandaz, line 192: Hm... this looks different.. I think it'll fail on VSCode.. but not a blocker

FrozenPandaz (changes requested):

nx.json
jaysoo: I've tried this but it doesn't work in exclude: ` "packages/**/__fixtures__/**/*"`

packages/remix/src/generators/application/__snapshots__/application.impl.spec.ts.snap
jaysoo: It shouldn't matter, it was picking up `pnpm-lock.yaml` from the workspace root, but since cwd changed it falls back to `npx`.

packages/eslint-plugin/src/configs/react-base.ts
```diff
@@ -90,7 +89,9 @@ export default {
     'no-template-curly-in-string': 'warn',
     'no-this-before-super': 'warn',
     'no-throw-literal': 'warn',
-    'no-restricted-globals': ['error'].concat(restrictedGlobals),
+    'no-restricted-globals': ['error'].concat(
+      require('confusing-browser-globals')
```
jaysoo, line 93: Because we point to source, the transpile here can lead to errors if we use the default import (due to ESM interopt). Safer to just use `require` since this package is CJS-only.

This is an Nx repo issue only.

packages/jest/src/plugins/plugin.spec.ts
```diff
@@ -31,8 +33,11 @@ describe('@nx/jest/plugin', () => {
     });
   });
 
+  test('foo', () => {});
```
FrozenPandaz, line 36: Please remove this.

FrozenPandaz (approved):

PR #22955, opened 2024-04-23
chore(repo): use @nx/webpack/plugin to infer graph client targets

jaysoo:
This PR adds `@nx/webpack/plugin` to use inferred serve/build targets for `graph-client`. Changes are needed for the `e2e-graph-client` to work since it can no longer rely on the webpack dev-server executor to pass `baseUrl`. Instead, we'll use the `webServerCommands` feature to start the dev server.

**Note:** Since `graph-client` uses `--configuration` heavily to differentiate between static graph, nx console, etc. I choose to leave those configuration intact but move the logic of `assets` and other options in `webpack.config.js`. This also means less changes to `e2e-graph-client`. We could reconsider the setup later, but it's good enough for now.

## Related Issue(s)

Fixes #


nx.json
```diff
@@ -215,6 +212,14 @@
         "targetName": "pw-e2e",
         "ciTargetName": "e2e-ci"
       }
+    },
+    {
+      "plugin": "@nx/webpack/plugin",
+      "include": ["graph/**/*"],
```
FrozenPandaz, line 218: I think we should everything. There are no other `webpack.config.ts/js` right?

graph/client/webpack.config.js
```diff
@@ -1,20 +1,157 @@
...
+  },
+  plugins: [
+    new NxWebpackPlugin({
+      index: './src/index.html',
+      compiler: 'babel',
+      main: './src/main.tsx',
+      tsConfig: './tsconfig.app.json',
+      styles: ['./src/styles.css'],
+      scripts: [],
+      assets: getAssets(process.env.NX_TASK_TARGET_CONFIGURATION),
```
FrozenPandaz, line 33: This depends on being run within Nx? :thinking: Is there a more native webpack way to specify a configuration or arbitrary variable was used?

graph/client/project.json
```diff
@@ -31,191 +31,24 @@
...
-        },
-        "release-static": {
-          "buildTarget": "graph-client:build-client:release-static",
-          "port": 4205
-        },
-        "dev-e2e": {
-          "buildTarget": "graph-client:build-client:dev-e2e",
-          "port": 4206
-        }
+        "dev": {},
```
FrozenPandaz, line 46: Mm this is not so great but I see what's going on here.

FrozenPandaz (approved):

nx.json
jaysoo: Nope

graph/client/webpack.config.js
jaysoo: Not really. Only the mode is available, but it is strictly for `development`, `production`, `none`. The other option is to use environment variables, but it is clunky.

I think we should re-examine how `graph-client` and `e2e-graph-client` is setup. Emily can take that one as a future improvement.

PR #22963, opened 2024-04-23
fix(gradle): change gradle command to be relative path

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/plugin/nodes.ts
+++ b/packages/gradle/src/plugin/nodes.ts
@@ -8,7 +8,7 @@ import {
 } from '@nx/devkit';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
 import { existsSync } from 'node:fs';
-import { dirname, join } from 'node:path';
+import { dirname, join, relative } from 'node:path';
 import { projectGraphCacheDirectory } from 'nx/src/utils/cache-directory';
 
 import { getGradleBinaryPath } from '../utils/exec-gradle';
@@ -180,8 +180,9 @@ function createGradleTargets(
     const targetName = options?.[`${task.name}TargetName`] ?? task.name;
 
     const outputs = outputDirs.get(task.name);
+    const path = relative(projectRoot, getGradleBinaryPath());
     targets[targetName] = {
-      command: `${getGradleBinaryPath()} ${task.name}`,
+      command: `${path} ${task.name}`,
       options: {
         cwd: projectRoot,
       },
--- a/packages/gradle/src/utils/exec-gradle.ts
+++ b/packages/gradle/src/utils/exec-gradle.ts
@@ -10,14 +10,14 @@ import { join } from 'node:path';
 
 export function execGradle(
   args: string[],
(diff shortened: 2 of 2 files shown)


packages/gradle/src/utils/exec-gradle.ts
```diff
@@ -10,36 +10,38 @@ import { join } from 'node:path';
 
 export function execGradle(
   args: string[],
-  execOptions: ExecFileSyncOptionsWithBufferEncoding
+  execOptions?: ExecFileSyncOptionsWithBufferEncoding
 ) {
-  const gradleBinaryPath = getGradleBinaryPath();
+  const graldeFile = getGradleFile();
```
FrozenPandaz, line 15:
```suggestion
  const gradleFile = getGradleFile();
```

packages/gradle/src/plugin/nodes.ts
```diff
@@ -181,9 +182,9 @@ function createGradleTargets(
 
     const outputs = outputDirs.get(task.name);
     targets[targetName] = {
-      command: `${getGradleBinaryPath()} ${task.name}`,
+      command: `${getGradleFile()} ${task.name}`,
       options: {
-        cwd: projectRoot,
+        cwd: workspaceRoot,
```
FrozenPandaz, line 187: Can we keep `projectRoot` and get a relative path from `projectRoot` to `gradleFile`?

Moving to the `workspaceRoot` would break the command for nested projects.

FrozenPandaz (changes requested):

packages/gradle/src/utils/exec-gradle.ts
```diff
@@ -31,12 +31,9 @@ export function getGradleBinaryPath() {
 
 export function execGradleAsync(
   args: ReadonlyArray<string>,
-  execOptions: ExecFileOptions
+  execOptions?: ExecFileOptions
 ) {
   const gradleBinaryPath = getGradleBinaryPath();
-  if (!existsSync(gradleBinaryPath)) {
-    throw new Error('Gradle is not setup. Run "gradle init"');
-  }
```
FrozenPandaz, line 39: I don't think there was anything wrong this. Can you bring it back please?

packages/gradle/src/utils/get-gradle-report.ts
```diff
@@ -33,9 +33,7 @@ export function getGradleReport(): GradleReport {
   const gradleProjectReportStart = performance.mark(
     'gradleProjectReport:start'
   );
-  const projectReportLines = execGradle(['projectReport'], {
-    cwd: workspaceRoot,
```
FrozenPandaz, line 37: Please keep this in case the `cwd` changes.

FrozenPandaz (changes requested):

packages/gradle/src/utils/exec-gradle.ts
xiongemi: this check is a part of getGradleBinaryPath function, which got called before this line. i don't think we need this

FrozenPandaz (approved):

PR #22986, opened 2024-04-24
docs(core): restrict plugin inferred tasks

isaacplmann:
Add section about restricting plugin inferred tasks

### DIFF
--- a/docs/shared/reference/nx-json.md
+++ b/docs/shared/reference/nx-json.md
@@ -86,6 +86,24 @@ To learn about creating your own plugin read about [extending Nx](/extending-nx/
 }
 ```
 
+### Restrict Plugin Inferred Tasks
+
+You can limit which projects are processed by Nx plugins to infer tasks using the `include` and `exclude` properties in the plugin configuration object.
+
+```jsonc {% fileName="nx.json" %}
+{
+  "plugins": [
+    {
+      "plugin": "@nx/jest/plugin",
+      "include": ["packages/**/*"], // include any projects in the packages folder
+      "exclude": ["**/*-e2e/**/*"] // exclude any projects in a *-e2e folder
+    }
+  ]
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/shared/reference/nx-json.md
```diff
@@ -86,6 +86,24 @@ To learn about creating your own plugin read about [extending Nx](/extending-nx/
 }
 ```
 
+### Restrict Plugin Inferred Tasks
+
+You can limit which projects are processed by Nx plugins to infer tasks using the `include` and `exclude` properties in the plugin configuration object.
```
FrozenPandaz, line 91:
```suggestion
You can scope which files are processed by Nx plugins to infer tasks using the `include` and `exclude` properties in the plugin configuration object.
```

I also think the section above can go into a little more details about how plugins work in general.

Specifically that they analyze files. Without that preface, i feel this section is kind of out of the blue talking about files getting processed or not processed and it's confusing.

docs/shared/reference/nx-json.md
```diff
@@ -86,6 +86,24 @@ To learn about creating your own plugin read about [extending Nx](/extending-nx/
...
+    {
+      "plugin": "@nx/jest/plugin",
+      "include": ["packages/**/*"], // include any projects in the packages folder
+      "exclude": ["**/*-e2e/**/*"] // exclude any projects in a *-e2e folder
+    }
+  ]
+}
+```
+
+The `include` and `exclude` properties are each file glob patterns that are used to include or exclude the configuration file that the plugin is interpreting. In the example provided, the `@nx/jest` plugin will only infer tasks for projects where the `jest.config.ts` file path both matches the `packages/**/*` glob and does not match the `**/*-e2e/**/*` glob.
```
FrozenPandaz, line 105:
```suggestion
The `include` and `exclude` properties are each file glob patterns that are used to include or exclude the configuration file that the plugin is interpreting. In the example provided, the `@nx/jest/plugin` plugin will only infer tasks for projects where the `jest.config.ts` file path matches the `packages/**/*` glob but does not match the `**/*-e2e/**/*` glob.
```

FrozenPandaz (changes requested):

leosvelperez (approved):

FrozenPandaz (approved):

PR #23021, opened 2024-04-26
feat(webpack): change plugin import paths to speed up config loading

jaysoo:
This PR updates the import for `NxWebpackPlugin` and `NxReactWebpackPlugin` to avoid importing too many things. This change reduces config loading by about 33%.

The imports are now as follows:

```js
// Plugin to build Nx-style web/node apps
const { NxAppWebpackPlugin } = require('@nx/webpack/app-plugin');

// Plugin to bring in React HMR, SVGR, etc.
const { NxReactWebpackPlugin } = require('@nx/react/webpack-plugin');
```

The old paths from `@nx/webpack` and `@nx/react` continues to work, but are deprecated. Will do a follow-up before Nx 20 to update existing configs.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/webpack/index.ts
+++ b/packages/webpack/index.ts
@@ -1,10 +1,18 @@
 import { configurationGenerator } from './src/generators/configuration/configuration';
+import { NxAppWebpackPlugin } from './src/plugins/nx-webpack-plugin/nx-app-webpack-plugin';
+import { NxTsconfigPathsWebpackPlugin as _NxTsconfigPathsWebpackPlugin } from './src/plugins/nx-typescript-webpack-plugin/nx-tsconfig-paths-webpack-plugin';
 
 export { configurationGenerator };
 
 // Exported for backwards compatibility in case a plugin is using the old name.
 /** @deprecated Use `configurationGenerator` instead. */
 export const webpackProjectGenerator = configurationGenerator;
+
+// TODO(v20): Remove this in favor of deep imports in order to load configs faster (150-200ms faster).
+/** @deprecated Use NxAppWebpackPlugin from `@nx/webpack/app-plugin` instead. */
+export const NxWebpackPlugin = NxAppWebpackPlugin;
+/** @deprecated Use NxTsconfigPathsWebpackPlugin from `@nx/webpack/tsconfig-paths-plugin` instead. */
+export const NxTsconfigPathsWebpackPlugin = _NxTsconfigPathsWebpackPlugin;
 
 export * from './src/utils/create-copy-plugin';
 export * from './src/utils/config';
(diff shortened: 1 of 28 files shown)


packages/webpack/nx-webpack-plugin.ts
```diff
@@ -0,0 +1,2 @@
+export { NxWebpackPlugin } from './src/plugins/nx-webpack-plugin/nx-webpack-plugin';
```
FrozenPandaz, line 1: Can we tighten up the path a bit to `@nx/webpack/webpack-plugin`?

Should look like:
```
import { NxWebpackPlugin } from '@nx/webpack/webpack-plugin';
```

What does this plugin actually do? Is there a more descriptive name for this?

packages/webpack/nx-typescript-webpack-plugin.ts
```diff
@@ -0,0 +1 @@
+export { NxTsconfigPathsWebpackPlugin } from './src/plugins/nx-typescript-webpack-plugin/nx-tsconfig-paths-webpack-plugin';
```
FrozenPandaz, line 1: Can we tighten up the path a bit to `@nx/webpack/tsconfig-paths-plugin`?

Should look like:
```
import { NxTsConfigPathsPlugin } from '@nx/webpack/tsconfig-paths-plugin';
```

FrozenPandaz (changes requested):

leosvelperez (approved):

packages/webpack/nx-webpack-plugin.ts
jaysoo: The deep import path is weird because we already have `@nx/webpack/plugin` reserved for inferring targets. The `NxWebpackPlugin` handles building both web and node apps. Could be `NxAppWebpackPlugin` or whatever. It replaces what the executor used to do.

packages/webpack/index.ts
```diff
@@ -1,11 +1,19 @@
...
 
 // Exported for backwards compatibility in case a plugin is using the old name.
 /** @deprecated Use `configurationGenerator` instead. */
 export const webpackProjectGenerator = configurationGenerator;
 
+// TODO(v20): Remove this in favor of deep imports in order to load configs faster (150-200ms faster).
+/** @deprecated Use NxAppWebpackPlugin from `@nx/webpack/app-plugin` instead. */
+export const NxWebpackPlugin = NxAppWebpackPlugin;
+/** @deprecated Use NxTsconfigPathsWebpackPlugin from `@nx/webpack/tsconfig-paths-plugin` instead. */
+export const NxTsconfigPathsWebpackPlugin = _NxTsconfigPathsWebpackPlugin;
```
FrozenPandaz, line 15: Can we add a migration in a followup?

FrozenPandaz (approved):

packages/webpack/index.ts
jaysoo: Yes, planning to next week.

PR #23056, opened 2024-04-28
fix(core): add quotes around string to command

xiongemi:
## Expected Behavior
- nx run will not have quotes, but underlying command will have

Screenshot 2024-05-23 at 2 07 20 PM
Screenshot 2024-05-23 at 2 07 03 PM

## Related Issue(s)

Fixes #


FrozenPandaz (changes requested):
Please add e2e tests as well. in `run.test.ts`

FrozenPandaz (approved):

PR #23082, opened 2024-04-29
fix(graph): use class sticky for sticky header

xiongemi:
## Related Issue(s)

Fixes #


MaxKless:
Are we sure we only want to make the target group heading sticky? It looks a bit lonely IMO, we could also make the project name sticky so there's more information.
Just a thought @xiongemi  :)

FrozenPandaz (approved):

PR #23093, opened 2024-04-30
cleanup(angular): use module & moduleResolution node16

leosvelperez:
Updates the TypeScript configuration for the `@nx/angular` plugin to use `module` and `moduleResolution` set to `node16`. The upcoming Angular v18 support needs this, but we can do it ahead of time to keep the Angular 18 PR leaner.

## Related Issue(s)

Fixes #


leosvelperez:
I'm going to close this and include the changes in the Angular v18 branch so they make more sense and are tested in the context where they are actually required.

PR #23098, opened 2024-04-30
fix(bundling): resolve index files from ts paths when running esbuild without bundling

andriizavoiko:
## Current Behavior
During build process a file `tmp/<path>/main-with-require-overrides.js` is generated to handle `paths` resolutions from `tsconfig` when `esbuild` is not running in `bundle` mode.

Having following `tsconfig.json` as example
```json
{
  ...
  "compilerOptions": {
    ...
    "paths": {
      "@lib/lib1": ["libs/lib1/src/index.ts"],
      "@app1/*": ["apps/app1/src/*"],
    }
  }
}
```

We can use `lib1` in code as follows and during runtime it will automatically be resolved to `dist/apps/app1/libs/lib1/src/index.js`:
```js
// apps/app1/src/main.ts
import lib1 from '@lib/lib1';
```

Hovewer, when trying to use paths with wildcards, e.g.:
```js
// apps/app1/src/config/index.ts
const config = {};
export default config;

// apps/app1/src/main.ts
import config from '@app1/config';
```

It gets resolved to `dist/apps/app1/apps/app1/src/config.js` and `isFile` condition fails as such module doesn't exist, thus path is not replaced and runtime error is produced.

## Expected Behavior
During resolution of following code:
```js
import config from '@app1/config';
```

`_resolveFilename` should consider all possible combinations of module path - `dist/apps/app1/apps/app1/src/config.js` and `dist/apps/app1/apps/app1/src/config/index.js`


andriizavoiko:
Just found that this issue already has a PR https://github.com/nrwl/nx/pull/22054 - but with a different approach and still in draft, maybe we can close it once this once is merged (or that one)

Coly010 (approved):

PR #23109, opened 2024-05-01
fix(core): fix missing import

chabb:
## Current Behavior

The import for `performance` is missing, which causes an error to appear

## Expected Behavior

No error should occured
## Related Issue(s)

Fixes #


AgentEnder (approved):
Hey 👋 - `performance` is available on Node 16 and higher, which means that its available for all of the node versions we currently target.

I'm going to accept this PR, but for future reference you may encounter other issues if using a lower node version. See: https://nx.dev/nx-api/workspace/documents/nx-nodejs-typescript-version-matrix

chabb:
This was actually breaking with node v20. According to the doc, you need to do the import

alaingiller:
I'm using node 20.12.0 and have this problem with the last nx version when I run jest test for my nx angular application plugin.
 The problem is probably because 'perf_hooks' is imported in same places but not here and probably related to nx cypress plugin.

AgentEnder:
Here's the docs that show perf was added as a global in node v16, if you get an error about it not being defined then there is something causing Nx to be ran with a different node version than you are expecting. https://nodejs.org/docs/latest/api/globals.html#performance

You can test this locally in your terminal with `node -p "performance.now()"`.

image

image

alaingiller:
Agree, and I've it in node, and I've only one instance of node, the version 20, but with the last version of nx, I got this problem while running jest. 
command: nx run my-project:test --coverage
node: v20.12.2
Error:
```
    LoadPluginError: Could not load plugin /builds/diemobiliar/it/rwc/utils/rwc-utils-nxplugin-jslib/node_modules/nx/src/plugins/project-json/build-nodes/package-json-next-to-project-json
      at loadNxPluginAsync (../../node_modules/nx/src/project-graph/plugins/loader.js:207:15)
      at loadNxPlugin (../../node_modules/nx/src/project-graph/plugins/loader.js:180:9)
      at loadNxPlugins (../../node_modules/nx/src/project-graph/plugins/internal-api.js:46:48)
      at buildProjectGraphAndSourceMapsWithoutDaemon (../../node_modules/nx/src/project-graph/project-graph.js:75:32)
      at createProjectGraphAndSourceMapsAsync (../../node_modules/nx/src/project-graph/project-graph.js:201:25)
      at createProjectGraphAsync (../../node_modules/nx/src/project-graph/project-graph.js:190:39)
      at initEsLint (../../node_modules/@nx/eslint/src/generators/init/init.js:41:19)
      at lintInitGenerator (../../node_modules/@nx/eslint/src/generators/init/init.js:80:12)
      at lintProjectGeneratorInternal (../../node_modules/@nx/eslint/src/generators/lint-project/lint-project.js:25:22)
      at addLintingGenerator (../../node_modules/@nx/angular/src/generators/add-linting/add-linting.js:14:22)
      at addLinting (../../node_modules/@nx/angular/src/generators/library/library.js:123:5)
      at libraryGeneratorInternal (../../node_modules/@nx/angular/src/generators/library/library.js:69:5)
      at libraryGenerator (../../node_modules/@nx/angular/src/generators/library/library.js:25:12)
    Cause:
    TypeError: performance.mark is not a function
      at loadNxPluginAsync (../../node_modules/nx/src/project-graph/plugins/loader.js:198:21)
      at loadNxPlugin (../../node_modules/nx/src/project-graph/plugins/loader.js:180:9)
      at loadNxPlugins (../../node_modules/nx/src/project-graph/plugins/internal-api.js:46:48)
      at buildProjectGraphAndSourceMapsWithoutDaemon (../../node_modules/nx/src/project-graph/project-graph.js:75:32)
      at createProjectGraphAndSourceMapsAsync (../../node_modules/nx/src/project-graph/project-graph.js:201:25)
      at createProjectGraphAsync (../../node_modules/nx/src/project-graph/project-graph.js:190:39)
      at initEsLint (../../node_modules/@nx/eslint/src/generators/init/init.js:41:19)
      at lintInitGenerator (../../node_modules/@nx/eslint/src/generators/init/init.js:80:12)
      at lintProjectGeneratorInternal (../../node_modules/@nx/eslint/src/generators/lint-project/lint-project.js:25:22)
      at addLintingGenerator (../../node_modules/@nx/angular/src/generators/add-linting/add-linting.js:14:22)
      at addLinting (../../node_modules/@nx/angular/src/generators/library/library.js:123:5)
```
But with this patch, it is working

JamesHenry:
@AgentEnder @alaingiller I imagine this is a jest `testEnvironment ` issue. I imagine @alaingiller is using `jsdom` or a custom one where the global won't exist?

alaingiller:
Good point, I effectively have jest-environment-jsdom.
**BUT** This is the standard generated by `npx create-nx-plugin my-plugin` like describe in https://nx.dev/extending-nx/tutorials/create-plugin, so nothing custom.

chabb:
@JamesHenry @AgentEnder.  `npx create-nx-plugin` generates a project that uses the jest plugin under the hood

The jest plugin uses jsdom under the hood : https://github.com/jsdom/jsdom/issues/3309. Jsdom does not implement the `performance` API ( e.g https://github.com/jsdom/jsdom/issues/3309 ). But IHMO, we should not use an evironment that simulates the browser here. For a plugin, I think it makes more sense to use node as the test environment. Using this jest config should solve it:

```const nxPreset = require('@nx/jest/preset').default;
module.exports = { ...nxPreset, testEnvironment: 'node' };```

Maybe we can update the generator to use node for tests ( for plugin )

JamesHenry:
@chabb Thanks that makes sense, I'll open an issue to track

PR #23051, opened 2024-04-27
feat(release): add support for version plans

fahslaj:
This PR adds support for "version plans", which is Nx Release's way of decoupling the release process from git history. 

TODO on this PR:

- [x] Add command for creation of version plans
- [x] Test flow when using "release" top level command
- [ ] Test flow when using programmatic API to apply version plans
- [ ] Test flow for using version/changelog subcommands
- [ ] Move "firstRelease" out of general release options so that it doesn't appear for the plan command
- [ ] Do not generate version plan files with the implicit default group name (use project names instead? Or more friendly group identifier?)
- [ ] Export the `releasePlan(...)` function so that it's formally part of the programmatic api
- [ ] Implement a way to pass bump selections directly into command to bypass enquirer prompts (useful for testing and to allow users to use the `releasePlan(...)` programmatic api to overlay their own UX)
- [ ] Add e2e tests for "plan" command
- [ ] Write full PR writeup describing all version plan functionality
- [ ] Add guide for using version plans to docs

PR #23190, opened 2024-05-06
feat(release): add support for version plans

fahslaj:
This PR adds support for "version plans", which is Nx Release's way of decoupling the release process from git history. 

- Create a version plan with `nx release plan`. Specify version bumps for each project or release group as determined by nx release config, then a conventional commits style message for changelog generation.
- Plan files are contained in `.nx/version-plans` directory, and will specify any number of release groups or projects to release
- By default, plans are consumed by the changelog command. This can be overridden with the programmatic API.
- Plans can be enabled on a per-release-group basis

Other changes in this PR:
- The changelog renderer is now decoupled from the GitCommit interface. This allows changelog changes to be passed from other sources (like version plans)
- gitAdd will no longer error if passed a file that doesn't exist and isn't tracked in git
- `--firstRelease` is now an explicit option on each of the `release`, `release version`, `release changelog`, and `release publish` commands instead of all children of the top level `release` command (because it's not relevant for the `release plan` subcommand)

Outstanding tasks for version plans that this PR does NOT cover:
- Add a more friendly indicator within release plan files when using the default release group. In this PR, the indicator will be the default release group placeholder of `__default__`.
- Make the `plan` command smarter, such as the ability to opt-in for conventional commit integration to recommend packages for which to plan releases.
- Decouple version plan change messages from conventional commits standard, which will involve significant changes to how grouping messages is done in the changelog renderer.
- Docs & guides for using version plans


fahslaj:
This PR replaces #23051

e2e/release/src/circular-dependencies.test.ts
```diff
@@ -76,6 +76,12 @@ describe('nx release circular dependencies', () => {
       return json;
     });
 
+    // Generate changelog files for each package to avoid a quirk with gitAdd(...)
+    // not specifying that new changelog files would be added when in dry-run mode
+    await runCommandAsync(`touch CHANGELOG.md`);
```
JamesHenry, line 81: We'll change things so this isn't required

packages/js/src/generators/release-version/release-version.ts
```diff
@@ -434,6 +441,85 @@ To fix this you will either need to add a package.json file at that location, or
...
+                );
+                return gt(nextNewVersion, prevNewVersion)
+                  ? plan.groupVersionBump
+                  : spec;
+              }, null);
+            }
+
+            if (!specifier) {
+              specifier = null;
+              log(`🚫 No changes were detected within version plans.`);
```
JamesHenry, line 499:
```suggestion
            if (!specifier) {
                if (projectToDependencyBumps.has(projectName)) {
                    // No applicable changes to the project directly by the user, but one or more dependencies have been bumped and updateDependents is enabled
                    specifier = updateDependentsBump;
                    log(`📄 Resolved the specifier as "${specifier}" because "release.version.generatorOptions.updateDependents" is enabled`);
                } else {
                    specifier = null;
                    log(`🚫 No changes were detected within version plans.`);
                }
            }
```

packages/nx/src/command-line/release/config/version-plans.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export interface GroupVersionPlan extends VersionPlan {
+  groupVersionBump: ReleaseType;
+}
+
+export interface ProjectsVersionPlan extends VersionPlan {
+  projectVersionBumps: Record<string, ReleaseType>;
+}
+
+const versionPlansDirectory = join('.nx', 'version-plans');
```
JamesHenry, line 34: Ensure folder exists

packages/nx/release/changelog-renderer/index.ts
```diff
@@ -43,7 +42,7 @@ export type DependencyBump = {
  */
 export type ChangelogRenderer = (config: {
   projectGraph: ProjectGraph;
-  commits: GitCommit[];
```
JamesHenry, line 46: Let's pass both for now to avoid the breaking change to the renderer for now

packages/nx/release/index.ts
```diff
@@ -4,6 +4,7 @@
 export {
   release,
   releaseChangelog,
+  releasePlan,
```
JamesHenry, line 7: Let's not expose this for now

packages/nx/src/command-line/release/index.ts
```diff
@@ -2,6 +2,10 @@
  * @public
  */
 export { releaseChangelog } from './changelog';
+/**
+ * @public
+ */
+export { releasePlan } from './plan';
```
JamesHenry, line 8: Remove

JamesHenry (approved):

JamesHenry (approved):

PR #23203, opened 2024-05-06
Feat/migrate more blogs

ndcunningham:
This PR adds 4 more blogs to the `/blog` on nx.dev


isaacplmann (changes requested):
All the featured images need to be optimized.  They're bigger than they need to be and should be run through https://squoosh.app/ to turn them into webp images.

The first time I load a post, it takes 3 or 4 seconds before the feature image appears - and I'm on a fast internet connection.

isaacplmann (approved):

PR #23205, opened 2024-05-06
fix(gradle): use local gradlew instead of sdkman

xiongemi:
## Related Issue(s)

Fixes #


e2e/gradle/src/gradle.test.ts
```diff
@@ -95,11 +98,24 @@ function createGradleProject(
   type: 'kotlin' | 'groovy' = 'kotlin'
 ) {
   e2eConsoleLogger(`Using java version: ${execSync('java --version')}`);
-  e2eConsoleLogger(`Using gradle version: ${execSync('gradle --version')}`);
-  e2eConsoleLogger(execSync(`gradle help --task :init`).toString());
+  copySync(`${__dirname}/../gradlew`, `${tmpProjPath()}/gradlew`);
+  copySync(`${__dirname}/../gradlew.bat`, `${tmpProjPath()}/gradlew.bat`);
+  copySync(`${__dirname}/../gradle`, `${tmpProjPath()}/gradle`);
```
FrozenPandaz, line 103: Do we have to copy these into the tmp workspace? Doesn't `gradle init` handle creating these?

In other words, can we invoke init with the absolute path? `${__dirname}/../gradlew --init`?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #23226, opened 2024-05-07
fix(gradle): run gradle init if no settings.gradle

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/generators/init/init.ts
+++ b/packages/gradle/src/generators/init/init.ts
@@ -8,12 +8,20 @@ import {
   Tree,
   updateNxJson,
 } from '@nx/devkit';
+import { execSync } from 'child_process';
 import { nxVersion } from '../../utils/versions';
 import { InitGeneratorSchema } from './schema';
 import { hasGradlePlugin } from '../../utils/has-gradle-plugin';
 
 export async function initGenerator(tree: Tree, options: InitGeneratorSchema) {
   const tasks: GeneratorCallback[] = [];
+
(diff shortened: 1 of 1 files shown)


packages/gradle/src/generators/init/init.ts
```diff
@@ -8,13 +8,21 @@ import {
...
 import { hasGradlePlugin } from '../../utils/has-gradle-plugin';
 
 export async function initGenerator(tree: Tree, options: InitGeneratorSchema) {
   const tasks: GeneratorCallback[] = [];
 
+  if (!tree.exists('settings.gradle') && !tree.exists('settings.gradle.kts')) {
+    logger.warn(`Could not find 'settings.gradle' or 'settings.gradle.kts' file in your gradle workspace.
+A Gradle build should contain a 'settings.gradle' or 'settings.gradle.kts' file in its root directory. It may also contain a 'build.gradle' or 'build.gradle.kts' file.
+Running 'gradle init':`);
+    execSync('gradle init', { stdio: 'inherit' });
```
FrozenPandaz, line 23: Did you do this with an `execSync` here because it would get picked up later on via the Tree? :eyes:

xiongemi: yes. just needs to be called before the function addProjectReportToBuildGradle so when checking `if (tree.exists('settings.gradle.kts')) {`, it would be true

FrozenPandaz (approved):

PR #23234, opened 2024-05-08
cleanup(misc): improve check for whether stats should be recorded

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/create-nx-workspace/src/utils/nx/ab-testing.ts
+++ b/packages/create-nx-workspace/src/utils/nx/ab-testing.ts
@@ -1,4 +1,6 @@
+import { execSync } from 'node:child_process';
 import { isCI } from '../ci/is-ci';
+import { getPackageManagerCommand } from '../package-manager';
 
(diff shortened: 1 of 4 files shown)


packages/create-nx-workspace/src/utils/nx/ab-testing.ts
```diff
@@ -116,3 +116,17 @@ export async function recordStat(opts: {
     }
   }
 }
+
+function shouldRecordStats(): boolean {
+  const pm = detectPackageManager();
+  try {
+    const stdout = execSync(`${pm} config get registry`, { encoding: 'utf-8' });
```
FrozenPandaz, line 123: This is not the right command for `yarn 2+` the variable there is... `npmRegistryServer` https://yarnpkg.com/configuration/yarnrc#npmRegistryServer.

Maybe we should put this into the `pmCommands`?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #23244, opened 2024-05-08
docs(core): fix images to point to webp

jaysoo:
A few images were missed when converting to webp.

## Related Issue(s)

Fixes #


vsavkin (approved):

jaysoo:
Missed some images, will follow-up with another PR

PR #23252, opened 2024-05-08
feat(release): updateDependents generator option for versioning, support circular dependencies

### LINKED ISSUE #22268: Nx release in "projectRelationship": "independent" mode is not tagging and releasing projects that depend on projects that are getting published
### Current Behavior

`nx release` in `"projectRelationship": "independent"` mode is only tagging and releasing projects with changes. If there are projects that depend on these projects, they are not getting tagged and released.
For example, if we have 2 libraries, `is-even` and `is-odd`.  `is-odd` has a dependency on package `is-even` that is correctly identified by `dependency-checks` `eslint` rule, and stated in it's `package.json` file. When changes are committed to package `is-even` and `nx release` is triggered, only that package is tagged for release. `is-odd`'s `package.json` is getting updated, but the package is not getting released.

Our `release` configuration from `nx.json`:
```json
  "release": {
    "version": {
      "conventionalCommits": true
    },
    "changelog": {
      "projectChangelogs": true
    },
    "releaseTagPattern": "{projectName}-{version}"
  },
```

image
image

### Expected Behavior

We are expecting that with the changes made to the "root" package, `release` bubbles up to all packages that depend on that package too. In this concrete example, we are expecting that with the changes made to `is-even` package, `nx release` not only triggers and tags `is-even` package, but `is-odd` package too, as that one directly depends on `is-even`.

### GitHub Repo

https://github.com/brugi82/nx-independent-monorepo

### Steps to Reproduce

1. Clone a [repository](https://github.com/brugi82/nx-independent-monorepo) 
2. Package `is-odd` depends on package `is-even`. `is-even` has one committed change that is not released.
3. Execute `nx release --skip-publish --dry-run --verbose`
4. Only package `is-even` is getting tagged & released.

### Nx Report

```shell
nx                 : 18.0.7
@nx/js             : 18.0.7
@nx/linter         : 18.0.7
@nx/eslint         : 18.0.7
@nx/workspace      : 18.0.7
@nx/devkit         : 18.0.7
@nx/eslint-plugin  : 18.0.7
@nx/react          : 18.0.7
@nx/rollup         : 18.0.7
@nrwl/tao          : 18.0.7
@nx/vite           : 18.0.7
@nx/web            : 18.0.7
typescript         : 5.3.3
```

### Failure Logs

_No response_

### Package Manager Version

yarn 4.1.0

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I am willing to submit a PR that solves this problem.

JamesHenry:
## Current Behavior

When releasing projects independently, if a dependent project is untouched directly by the changes, it will not have its version updated and there is no way to opt into this behavior.

Additionally, circular dependencies between packages are not supported.

## Expected Behavior

When releasing projects independently, if a dependent project is untouched directly by the changes, **BY DEFAULT** it will not have its version updated, **BUT** you can opt into it always being updated via a generator option (`release.version.generatorOptions.updateDependents = auto` and you can control what kind of semver bump should be applied to the otherwise unchanged dependent project. Transitive local dependents (`A -> B -> C`) will also be updated in this scenario.

Additionally, when opted into, such version only changes will appear in the changelog under a new `Updated Dependencies` section.

Circular dependencies between packages are now supported for versioning, changelog generation and publishing.

## Related Issue(s)

Fixes #22268


packages/js/src/generators/release-version/release-version.ts
```diff
@@ -583,3 +757,58 @@ function getColor(projectName: string) {
 
   return colors[colorIndex];
 }
+
+function sortProjectsTopologically(
```
FrozenPandaz, line 761: Ensure that this handles circular dependencies. It would be a good unit test to have.

packages/nx/src/command-line/release/changelog.ts
FrozenPandaz, line 1: Ensure this does not happen when the versioning is `fixed` rather than `independent`

FrozenPandaz (changes requested):

packages/js/src/generators/release-version/release-version.ts
JamesHenry: Circular dependencies in versioning are handled and unit tested in: https://github.com/nrwl/nx/pull/23252/commits/ee11e208363fb641f69843f274e973c7d67d11ec

packages/nx/src/command-line/release/changelog.ts
JamesHenry: I added e2e coverage for all the combinations of `fixed` vs `independent`:
https://github.com/nrwl/nx/pull/23252/commits/3350b72c60fae15c00ccc1bd3eff0a3e62d71bae

That additionally led me to discover that we did not previously allow cycles for publishing at all, so this PR now officially adds support for circular dependencies across versioning, changelog generation and publishing.

I have updated the PR title and description accordingly

JamesHenry:
@FrozenPandaz Please take another look when you get chance, I think this should be good to go

diosif-cr:
@JamesHenry I've been following this bit of work and how much you've done for it, and wanted to say thank you and that I'm extremely grateful. This makes Nx feature complete for a few of our projects, without any other 3rd party plug-ins.

mpsanchis:
@JamesHenry Also my thank you in advance for the effort. I am looking forward to this feature 🙏 

Here's two quick questions for the MR, if I may:
1. I have seen that you declare `projectToDependencyBumps` at the top of the `release-version` executor, and then in line 376 it is checking `if (projectToDependencyBumps.has(projectName))` (in the case `conventional-commits`). However, if I understand correctly, the Map will not be filled with data until line 651, where the first `.set()` is called on it. Won't the Map be empty in line 376?

2. From the results of my manual tests, the behavior that you are adding is not only missing *"When releasing projects independently"*, but in any scenario. I have seen repos where there is one release group (G1) containing shared libs, and then other release groups that contain apps that use those shared libs. If there is a change to G1, I'd expect that apps that belong to other groups "Gx" are also bumped. In this case, there would be no independent releasing, but this behavior would also be expected. Your change in the doesn't seem to apply only when "independent" is set, but I just wanted to double check.

JamesHenry:
@mpsanchis 

1. I don't think you have taken the loop into account
2. Is it out of scope for this PR to change the relationships between groups

Please open/use and existing discussion or issue to discuss any specific concerns around how groups relate to one another

jbadeau:
Also watching this MR as we need it to be able to release our repos which have cross release group dependencies

```mermaid
graph TD
    subgraph Todo Group
        A[Todo App] --> B[Todo Lib]
    end
    subgraph Shared Group
        B --> C[Shared Lib]
    end
```

If I change a project in Shared Group then I would expect a bump in someo or all of the Todo Group projects depending on if independent releasing

JamesHenry:
@jbadeau See my previous comment, this PR is not the place for a discussion around changing how groups relate to one another

mpsanchis:
If any others have the same doubt that I had: it's the topological ordering of the projects that ensures that the first projects in the for loop don't have any dependencies, and then the `projectToDependencyBumps` gets filled as the iterations progress. So projects that depend on other projects appear after in the loop, and the map already contains their information.

fahslaj (approved):

FrozenPandaz (approved):

rcolwell-cb:
This option doesn't seem to produce any change in behavior for me.
`{
  "defaultBase": "no-lerna",
  "release": {
    "version": {
      "generatorOptions": {
        "updateDependents": "auto"
      },
      "conventionalCommits": true
    },
    "projects": [
      "packages/*/*"
    ],
    "releaseTagPattern": "{projectName}@{version}",
    "changelog": {
      "projectChangelogs": true
    }
  }
}`

This is a project where lerna previously was used and attempting to transition everything to nx. When I make a change to a library project, none of the other library projects are getting version bumped, just their dependency updated.

rcolwell-cb:
However they are linked through package.json in my case.

rcolwell-cb:
UPDATE packages/apricot-react/button/package.json

      "name": "@fakeorg/cool-ui-button",
-     "version": "5.1.0",
+     "version": "5.1.1",
      "main": "dist/index.js",

UPDATE packages/apricot-react/notification/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "@fakeorg/cool-ui-glyph": "^5.0.4",

UPDATE packages/apricot-react/back-to-top/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0"
+         "@fakeorg/cool-ui-button": "^5.1.1"
      }

UPDATE packages/apricot-react/date-picker/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "classnames": "^2.3.1",

UPDATE packages/apricot-react/popover/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "classnames": "^2.3.1",

UPDATE packages/apricot-react/filter/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "classnames": "^2.3.1",

UPDATE packages/apricot-react/forms/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "@fakeorg/cool-ui-glyph": "^5.0.4",

UPDATE packages/apricot-react/modal/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "@fakeorg/cool-ui-notification": "^5.0.14",

UPDATE packages/apricot-react/table/package.json

      "dependencies": {
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "@fakeorg/cool-ui-glyph": "^5.0.4",

UPDATE packages/apricot-react/all/package.json

          "@fakeorg/cool-ui-breadcrumbs": "^5.0.5",
-         "@fakeorg/cool-ui-button": "^5.1.0",
+         "@fakeorg/cool-ui-button": "^5.1.1",
          "@fakeorg/cool-ui-card": "^5.1.6",
          

The dependencies are properly updated, but not getting version bumped with this new option.

rcolwell-cb:
it's working for me now actually - still testing it with different scenarios of bumps. My nx wasn't updated after switching to beta so cleared some things and reinstalled...

PR #23253, opened 2024-05-08
chore(repo): profiling

JamesHenry:


patches/@nx__js@19.0.0-rc.1.patch
JamesHenry, line 1: This applies the same change as in `packages/js/src/plugins/jest/start-local-registry.ts` so that it is immediately available for use.

The patch (and reference to it in package.json) can be removed once a version of Nx is released that contains this PR.

e2e/utils/global-setup.ts
JamesHenry, line 1: This is not an easy diff to read, probably better to review the final state of the file in isolation

JamesHenry:
Closing in favour of https://github.com/nrwl/nx/pull/23429 to have a clean history and be more focused on what this ended up becoming

PR #23267, opened 2024-05-09
docs(nx-dev): Add 404 for unknown blog urls

ndcunningham:
This PR adds the 404 fallback if a user navigates to a specified blog that does not exist.

Currently, we are showing a 500 error.


isaacplmann:
I don't think it is working:
https://nx-dev-git-fix-nx-docs-blog-404-nrwl.vercel.app/blog/fake

isaacplmann (approved):

PR #23274, opened 2024-05-09
fix(core): fix eslint --help command

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/run/run.ts
+++ b/packages/nx/src/command-line/run/run.ts
@@ -1,3 +1,4 @@
+import { env as appendLocalEnv } from 'npm-run-path';
 import {
   combineOptionsForExecutor,
   handleErrors,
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/run/run.ts
```diff
@@ -130,7 +131,7 @@ async function printTargetRunHelpInternal(
     targetConfig.options.command
   ) {
     const command = targetConfig.options.command.split(' ')[0];
-    const helpCommand = `${command} --help`;
+    const helpCommand = `${getPackageManagerCommand().exec} ${command} --help`;
```
FrozenPandaz, line 134: This won't work when the command already has `npx eslint .` in it for some reason?

FrozenPandaz (changes requested):
The logic already exists in `nx:run-commands`.. I feel like we're getting to the point where this maybe should be handled in there?

FrozenPandaz (approved):

PR #23279, opened 2024-05-09
fix(core): Handle removing non-empty directories and non-existing paths in cache

### LINKED ISSUE #22140: Directory not empty (os error 66
### Current Behavior

We are seeing this spurious error a lot. It's unclear what the cause it, but it seems to happen especially when there are a lot of affected files (i.e. change to root package.json).

This happens only during "git commit". We are using the pre-commit package to run "nx affected --target=build".

I'm just opening this issue to see if errors see this too. It seems to be a combination of pre-commit and nx, but what exactly is unclear, as the message comes from nx.

Typical output looks like:
image

### Expected Behavior

Running nx by itself works fine.

### GitHub Repo

_No response_

### Steps to Reproduce

We don't really now.

### Nx Report

```shell

Node   : 18.19.0
OS     : darwin-arm64
pnpm   : 8.15.4

nx                 : 18.0.6
@nx/js             : 18.0.6
@nx/linter         : 18.0.6
@nx/eslint         : 18.0.6
@nx/workspace      : 18.0.6
@nx/cypress        : 18.0.6
@nx/devkit         : 18.0.6
@nx/esbuild        : 18.0.6
@nx/eslint-plugin  : 18.0.6
@nx/next           : 18.0.6
@nx/playwright     : 18.0.6
@nx/react          : 18.0.6
@nx/storybook      : 18.0.6
@nrwl/tao          : 18.0.6
@nx/vite           : 18.0.6
@nx/web            : 18.0.6
@nx/webpack        : 18.0.6
typescript         : 5.3.3
Community plugins:
@berenddeboer/nx-aws-cdk         : 2.2.2
@berenddeboer/nx-sst             : 0.13.5
@trumbitta/nx-plugin-unused-deps : 1.12.1
Local workspace plugins:
         @tc/nx-sst
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Exelord:
## Current Behavior
When running NX concurrently sometimes NX is trying to remove non-existing path (which have been already deleted by other process) or non-empty directories (which contains files created by other process but not yet removed)

To prevent race conditions, remove from fs-extra works in following way:
> Removes a file or directory. **The directory can have contents. If the path does not exist, silently does nothing.**

## Expected Behavior
Remove operations should work with non-existing paths as well as non-empty directories

## Related Issue(s)

Fixes #22140

### DIFF
--- a/packages/nx/src/native/cache/file_ops.rs
+++ b/packages/nx/src/native/cache/file_ops.rs
@@ -1,14 +1,17 @@
 use std::path::{Path, PathBuf};
 use std::{fs, io};
 
-use fs_extra::error::ErrorKind;
-
 #[napi]
 pub fn remove(src: String) -> anyhow::Result<()> {
-    fs_extra::remove_items(&[src]).map_err(|err| match err.kind {
-        ErrorKind::Io(err_kind) => anyhow::Error::new(err_kind),
-        _ => anyhow::Error::new(err),
-    })
+    fs::metadata(&src)
+        .and_then(|metadata| {
+            if metadata.is_dir() {
+                fs::remove_dir_all(src)
+            } else {
+                fs::remove_file(src)
+            }
+        })
+        .map_err(anyhow::Error::new)
 }
 
 #[napi]
--- a/packages/nx/src/tasks-runner/cache.ts
+++ b/packages/nx/src/tasks-runner/cache.ts
@@ -1,5 +1,5 @@
 import { workspaceRoot } from '../utils/workspace-root';
-import { mkdir, mkdirSync, pathExists, readFile, writeFile, remove } from 'fs-extra';
+import { mkdir, mkdirSync, pathExists, readFile, writeFile, remove as fsRemove } from 'fs-extra';
 import { join } from 'path';
 import { performance } from 'perf_hooks';
 import { DefaultTasksRunnerOptions } from './default-tasks-runner';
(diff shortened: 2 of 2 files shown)


FrozenPandaz:
Hm, this removes the usage of our rust implementation of removing files where we also use a `fs-extras` type library.

https://github.com/nrwl/nx/blob/master/packages/nx/src/native/cache/file_ops.rs#L8

The rust one proved to be much faster so we should continue using that one and address the issue there.

Thank you so much for isolating the issue!

@Exelord If you're familiar with Rust, we would appreciate if you continue looking into the issue. However, if not, @AgentEnder can take over from here.

Exelord:
Great! Please do take over. I have tried to use rust fs-extra but it does not work the way described, so it would require some native implementation which I cannot help with. While JS remove might be slower it does fix the issue, so I went with that as a quick fix.

Thank you :)

Exelord:
Small, update. As in other places, the remove errors have been ignored as OS still can throw lock errors while trying to remove files used by concurrent process. (there is no fix for that - OS intended, but the file will be scheduled for remove)

While its still using remove from fs-extra, it works now, which is faster than using rust's remove which throws.

Eastonco:
Hey guys, Would love to see this merged. My enterprise team has been dealing with this issue for awhile now and would love to see a fix :) @AgentEnder Can you assist?

Eastonco:
@AgentEnder Thanks for the rewrite! Can I bump for some review and approvals :)

Exelord:
I think one issue is still not handled. If the file won't exist anymore`remove_file` will throw an error which is not intended

Eastonco:
@Exelord @AgentEnder Are there any updates on this? This issue is still plaguing our team and we'd love to see it resolved. The current work around has been to always run with the `--skip-nx-cache` flag

HPaulson:
Reproduced in Nx 21.0.3, MacOS 15.1

PR #23307, opened 2024-05-10
fix(core): prefer project specific external deps

JamesHenry:
## Current Behavior

We only ever discover one version of an external dependency for the file-map.json. This means the `@nx/dependency-checks` lint rule can produce incorrect failures when multiple copies of a dependency exist within a workspace.

## Expected Behavior

In the file-map.json the project specific version of a package (if multiple exist) is preferred and therefore the lint rule produces accurate results.

---

For example, in a repo where the root package.json has lodash@4.0.0 (which becomes `npm:lodash` on graph) and the foo project has lodash@3.0.0:

**Before**

image

**After**

image

## Performance

`NX_ISOLATE_PLUGINS=true NX_PERF_LOGGING=true NX_DAEMON=false nx show project nx --json false`

** Before **

Time for 'build typescript dependencies' 505.52144700009376

** After **

Time for 'build typescript dependencies' 701.247584999539


packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
```diff
@@ -80,29 +109,80 @@ function processPackageJson(
...
+          const externalPackageJson = parseJson(
+            readFileSync(externalPackageJsonPath, 'utf-8')
+          );
+          externalPackageJsonCache.set(
+            externalPackageJsonPath,
+            externalPackageJson
+          );
+          externalPackageVersion = externalPackageJson.version;
+        } else {
+          externalPackageVersion = externalPackageJsonCache.get(
```
FrozenPandaz, line 136: Hm... could we keep the cache management and usage in the module scope of the function rather than passing it around?

packages/nx/src/plugins/js/project-graph/build-dependencies/build-dependencies.ts
```diff
@@ -12,8 +25,23 @@ export function buildExplicitDependencies(
 ): RawProjectGraphDependency[] {
   if (totalNumberOfFilesToProcess(ctx) === 0) return [];
 
+  const externalDependenciesCache: ExternalDependenciesCache = new Map();
   let dependencies: RawProjectGraphDependency[] = [];
 
+  /**
+   * NOTE: It is important that we process package.json files first so that accurate versions for
```
FrozenPandaz, line 32: Doing this first... probably shouldn't cause any issues.. but I think it's weird that it is important for this to be first...

The 2 parts should be independent :thinking: 

Especially because someone can have `analyzePackageJson` false? :thinking:

FrozenPandaz (changes requested):

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-project-dependencies.ts
```diff
@@ -20,17 +21,44 @@ function isRoot(
   return projects[projectName]?.root === '.';
 }
 
+function getPackageName(importExpression: string) {
```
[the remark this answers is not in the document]
JamesHenry: Actually the logic in `targetProjectLocator` currently intentionally only considers root level packages. But yes that's the location I am now refactoring

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
JamesHenry: - I didn't want to run any risks of stale caches via require caching
- Unit tests are more straightforward/direct

JamesHenry:
@FrozenPandaz Please take another look

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -1,45 +1,74 @@
...
    *
    * @param importExpr
    * @param filePath
+   * @param projectRoot
    */
-  findProjectWithImport(importExpr: string, filePath: string): string {
+  findProjectFromImport(
+    importExpr: string,
+    filePath: string,
+    projectRoot: string
```
FrozenPandaz, line 70: I don't think this should be limited to project roots. It should be able to use the `dirname` of the `filePath`?

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
```diff
@@ -80,29 +92,61 @@ function processPackageJson(
...
-          sourceFile: fileName,
-          type: DependencyType.static,
-        };
-        validateDependency(dependency, ctx);
-        collectedDeps.push(dependency);
+        continue;
       }
-    });
+
+      const targetProjectLocator = new TargetProjectLocator(
```
FrozenPandaz, line 98: We shouldn't re-create this every single `package.json`.

Create it once in `buildExplicitPackageJsonDependencies`

FrozenPandaz (changes requested):

packages/nx/src/plugins/js/utils/find-external-package-json-path.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * In many cases the package.json will be directly resolvable, so we try that first.
+ * If, however, package exports are used and the package.json is not defined, we will
+ * need to resolve the main entry point of the package and traverse upwards to find the
+ * package.json.
+ *
+ * NOTE: Unit testing this code is currently impractical as it is not possible to mock
+ * require.resolve in jest https://github.com/jestjs/jest/issues/9543
+ */
+export function findExternalPackageJsonPath(
```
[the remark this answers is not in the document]
JamesHenry: It exists in its own file so that it can be mocked in tests

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -1,45 +1,74 @@
...
-  private npmResolutionCache = new Map<string, string | null>();
 
   constructor(
     private readonly nodes: Record<string, ProjectGraphProjectNode>,
-    private readonly externalNodes: Record<string, ProjectGraphExternalNode>
+    private readonly externalNodes: Record<
+      string,
+      ProjectGraphExternalNode
+    > = {},
+    private readonly npmResolutionCache: NpmResolutionCache = new Map<
```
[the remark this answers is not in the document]
JamesHenry: I previously responded to the module level cache here: https://github.com/nrwl/nx/pull/23307#discussion_r1604868361

- I didn't want to run any risks of stale caches via require caching
- Unit tests are more straightforward/direct because the cache can easily be stubbed/mocked

packages/nx/src/plugins/js/utils/find-external-package-json-path.ts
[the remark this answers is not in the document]
FrozenPandaz: Yeah, the file

packages/nx/src/plugins/js/project-graph/build-dependencies/explicit-package-json-dependencies.ts
```diff
@@ -80,29 +92,61 @@ function processPackageJson(
...
-      } else if (ctx.externalNodes[`npm:${d}`]) {
-        const dependency: RawProjectGraphDependency = {
-          source: sourceProject,
-          target: `npm:${d}`,
-          sourceFile: fileName,
-          type: DependencyType.static,
-        };
-        validateDependency(dependency, ctx);
-        collectedDeps.push(dependency);
+        continue;
```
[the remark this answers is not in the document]
JamesHenry: I am a big believer in an "early return style", and in the benefits of avoiding unnecessary blocks and indentation when it comes to code readability and `else` is therefore less commonly needed/desirable IMO. Here we are already unavoidably nested within function > try > for and we can avoid nesting all remaining logic within another block by leveraging an early return style (in this case continue, obviously)

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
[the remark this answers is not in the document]
JamesHenry: After updating the code to apply the other changes such as sharing a single targetProjectLocator instance, I combined these ideas by applying a default cache at the module level which is still overridable via the constructor parameter for unit tests

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.spec.ts
```diff
@@ -1,776 +1,763 @@
...
 
-  it('should return null for native modules', () => {
-    const result = targetProjectLocator.findProjectWithImport(
-      'path',
-      'libs/proj/index.ts'
-    );
-    expect(result).toEqual(null);
+describe('target-project-locator', () => {
+  describe('TargetProjectLocator', () => {
+    describe('findTargetProjectWithImport', () => {
```
FrozenPandaz, line 33: Oh boy, it looks like `git` thinks this whole file has changed... Is it because of the extra nesting here?

I imagine that isn't actually the case.. Is it possible for you to change the nesting here back so that git produces a reasonable diff please?

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.spec.ts
```diff
@@ -1,776 +1,763 @@
...
+          'npm-package',
+          'libs/proj1/index.ts'
+        );
+
+        expect(result1).toEqual('npm:@ng/core');
+        expect(result2).toEqual('npm:npm-package');
+      });
+
+      // TODO: verify this test is doing what we want
+      xit('should be able to resolve paths that have similar names', () => {
```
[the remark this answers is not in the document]
FrozenPandaz: Hm, yeah I think the one under the `"without tsconfig json"` wasn't properly failing. If I run just that test, it fails :thinking: 

The `vol` isn't being reset properly so it passes piggybacking off of the previous `fs`.

We can remove this test.

FrozenPandaz (changes requested):
We will want to profile how this does on the `nx` repo to ensure there isn't a large performance regression for finding dependencies.

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.spec.ts
```diff
@@ -1,776 +1,740 @@
...
-      './libs/proj4ab/index.ts': `export const a = 4;`,
-      './libs/proj5/index.ts': `export const a = 5;`,
-      './libs/proj6/index.ts': `export const a = 6;`,
-      './libs/proj7/index.ts': `export const a = 7;`,
-      './libs/proj123/index.ts': 'export const a = 123',
-      './libs/proj1234/index.ts': 'export const a = 1234',
-      './libs/proj1234-child/index.ts': 'export const a = 12345',
-    };
-    vol.fromJSON(fsJson, '/root');
-    ctx = {
```
JamesHenry, line 442: This was unused

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.spec.ts
```diff
@@ -1,776 +1,740 @@
...
-    const result2 = targetProjectLocator.findProjectWithImport(
-      'npm-package',
-      'libs/proj1/index.ts'
-    );
-
-    expect(result1).toEqual('npm:@ng/core');
-    expect(result2).toEqual('npm:npm-package');
-  });
-
-  it('should be able to resolve paths that have similar names', () => {
```
JamesHenry, line 727: This was the test we agreed to remove

FrozenPandaz (approved):

PR #23323, opened 2024-05-12
docs(core): add changelog renderOptions to nx-schema.json

pkuczynski:
## Current Behavior
Schema does not have the documentation

## Expected Behavior
Schema should list options from `DefaultChangelogRenderOptions`


JamesHenry:
@pkuczynski The trouble with this is that the changelog renderer is swappable by design. There is no way in JSON schema (that I am aware of) to customize these values dynamically based on the value of the renderer property. If using a custom renderer there is no requirement for the options to be the same as these (again by design), so the schema would give incorrect feedback, which I think is worse than none given there is no way to override it.

For that reason I am going to close this one for now unless we can somehow can up with a reliable dynamic solution.

Many thanks!

PR #23324, opened 2024-05-12
docs(release): add configure changelogs

pkuczynski:
## Current Behavior
Documentation is missing

## Expected Behavior
Documentation should be there...

### DIFF
--- a/docs/generated/manifests/menus.json
+++ b/docs/generated/manifests/menus.json
@@ -1985,6 +1985,14 @@
                 "isExternal": false,
                 "children": [],
                 "disableCollapsible": false
+              },
+              {
+                "name": "Configure Changelog Format",
+                "path": "/recipes/nx-release/configure-changelog-format",
+                "id": "configure-changelog-format",
+                "isExternal": false,
+                "children": [],
+                "disableCollapsible": false
               }
             ],
             "disableCollapsible": false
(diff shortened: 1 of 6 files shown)


isaacplmann:
Thanks for contributing!  This content is correct, but doesn't quite fit how we write recipes for documentation.

Two things:
1. For the recipe to show up in the sidebar, you need to add it to the `map.json` file and then run `nx documentation`.  (This will also make CI pass for this PR)
2. This recipe is written like an API reference document.  Recipes are intended to be short descriptions of how to solve a specific problem.  To address this, you could add some text describing the problem you're trying to solve, list the options (like you have already) and then show an example configuration that solves the problem you listed at the beginning.

pkuczynski:
So I run the `pnpm run documentation` to update remaining files and it seemed to work. I committed those. The build still fails though and I am not sure why?

I did a bit of an update to the doc itself. I understand it does not necessarily fit _receipe_, but at the same time there is no other place to put it. And I found it quote useful myself to know about such options, which I found just by accident browsing through reported issues.

All in all, not sure what shall I do next to get it merged?

isaacplmann:
The PR was failing CI because it didn't like some of your commit messages.  I squashed your commits together and then expanded the text of your recipe

isaacplmann (approved):

isaacplmann:
Yep, thanks for catching my mistake

PR #23333, opened 2024-05-13
fix(nx-cloud): ensure generated ci workflows use dlx for nx-cloud

JamesHenry:
## Current Behavior

E.g. `pnpm exec nx-cloud` is used, which will fail to find the nx-cloud executable because node_modules have not yet been installed.

## Expected Behavior

E.g. `pnpm dlx nx-cloud` is used, which can resolve the nx-cloud executable without node_modules.

## Related Issue(s)

Fixes #


nx-dev/ui-blog/.eslintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": ["../../.eslintrc.base.json"],
```
meeroslav, line 2: Why was this changed?

.eslintrc.base.json
meeroslav, line 1: I don't think this was supposed to be deleted, let alone as part of this PR

meeroslav (approved):
The PR looks good to me, but the eslint changes should be reverted.

Approving to unlock the PR

PR #23375, opened 2024-05-14
fix(core): copy native files to tmp file location instead of .nx/cache

### LINKED ISSUE #23224: Running nx reset on Windows results in an error when attempting to delete nx-win32-x64-msvc.node
### Current Behavior

This is almost identical to this bug: https://github.com/nrwl/nx/issues/20346 but after the location of that file was moved from `node_modules` to `.nx/cache`, the command which causes the issue changed from `npm ci` to `nx reset` (which attempts to delete the cache folder, effectively the same root cause as npm ci deleting node modules).

### Expected Behavior

No crash

### GitHub Repo

_No response_

### Steps to Reproduce

See this issue: https://github.com/nrwl/nx/issues/20346, just swap out `npm ci` for `nx reset`

### Nx Report

```shell
Node   : 20.9.0
OS     : win32-x64
npm    : 10.1.0
 
nx                 : 18.3.3
@nx/js             : 18.3.3
@nx/jest           : 18.3.3
@nx/linter         : 18.3.3
@nx/eslint         : 18.3.3
@nx/workspace      : 18.3.3
@nx/devkit         : 18.3.3
@nx/esbuild        : 18.3.3
@nx/eslint-plugin  : 18.3.3
@nx/plugin         : 18.3.3
@nrwl/tao          : 18.3.3
@nx/vite           : 18.3.3
typescript         : 5.4.5
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

MaxKless:
## Current Behavior
Currently, the `.node` files required to load native code are saved in `.nx/cache`. This can cause different issues:
- users on windows sometimes experience errors during `nx reset` because it's trying to delete the entire folder and some process is still locking the file
- `@angular-eslint` users are seeing the `.nx/cache` folder in their workspace since it uses `@nx/devkit` 

## Expected Behavior
We want no errors and for noone to be bothered by the `.node` file. This is why we move the `.node` file to a tmp location outside the workspace instead of `.nx/cache`.  We still make sure to delete it during `nx reset` but throw no errors if that fails. It will simply be deleted by the next invocation of `nx reset`.

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/23224

### DIFF
--- a/packages/nx/src/native/native-file-cache-location.ts
+++ b/packages/nx/src/native/native-file-cache-location.ts
@@ -1,0 +1,10 @@
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { createHash } from 'crypto';
+import { workspaceRoot } from '../utils/workspace-root';
+
+export const nativeFileCacheLocation = join(
+  tmpdir(),
+  'nx-native-file-cache',
+  createHash('sha256').update(workspaceRoot).digest('hex')
+);
--- a/packages/nx/src/command-line/reset/reset.ts
+++ b/packages/nx/src/command-line/reset/reset.ts
@@ -5,6 +5,7 @@ import {
   projectGraphCacheDirectory,
 } from '../../utils/cache-directory';
 import { output } from '../../utils/output';
+import { nativeFileCacheLocation } from '../../native/native-file-cache-location';
 
 export async function resetHandler() {
   output.note({
@@ -13,6 +14,11 @@ export async function resetHandler() {
   });
   await daemonClient.stop();
   output.log({ title: 'Daemon Server - Stopped' });
+  try {
+    rmSync(nativeFileCacheLocation, { recursive: true, force: true });
+  } catch (e) {
+    // ignore, deleting the native file cache is not critical and can fail if another process is locking the file
+  }
   rmSync(cacheDir, { recursive: true, force: true });
   if (projectGraphCacheDirectory !== cacheDir) {
     rmSync(projectGraphCacheDirectory, { recursive: true, force: true });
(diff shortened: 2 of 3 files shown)


JamesHenry (approved):

packages/nx/src/native/native-file-cache-location.ts
```diff
@@ -0,0 +1,10 @@
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { createHash } from 'crypto';
+import { workspaceRoot } from '../utils/workspace-root';
+
+export const nativeFileCacheLocation = join(
+  tmpdir(),
+  'nx-native-file-cache',
+  createHash('sha256').update(workspaceRoot).digest('hex')
```
FrozenPandaz, line 9: Why do we need this nesting at all? Can't it be `join(tmpdir(), 'nx/bin')`?

FrozenPandaz (changes requested):

packages/nx/src/native/native-file-cache-location.ts
MaxKless: I want to keep the same behaviour as before where the files are workspace-scoped. I'm not sure if there could be funky side-effects when one repo calls nx reset and deletes the `.node` file that is currently used by another process in another workspace.

FrozenPandaz: Okay sounds good.

FrozenPandaz (approved):

PR #23380, opened 2024-05-14
docs(core): local links should not use nx.dev

isaacplmann:
Ensure that local links start with `/` instead of `https://nx.dev/`


ndcunningham:
Can you update the error message so that we can tell the difference between links that do not link to an actual url and those that are internal but start with `https://nx.dev`

ndcunningham (approved):

PR #23422, opened 2024-05-16
fix(core): remove duplicate `js-yaml` packages

Phillip9587:
## Current Behavior

`nx` package depends on both `js-yaml` and `@zkochan/js-yaml`

## Expected Behavior

`nx` should only depend on one of the both

## Related Issue(s)

Fixes #


meeroslav:
Thank you @Phillip9587 for this PR. 

The two failing lint tasks are related to a limitation in the `dependency-checks` rule. Can you please add `js-yaml` to the list of `ignoredDependencies` in the corresponding `.eslintrc.json`, and we will address that issue separately?

You also need to rebase with the latest pnpm-lock file changes.

meeroslav (changes requested):
The `dependency-checks` rule is failing with alias dependencies. This should be resolved elsewhere.

Just include the override in the `ignoredDependencies` to get this PR over the line and we will address this issue separately.

Phillip9587:
@meeroslav Done. I created #23469 to track the issue with the lint rule

meeroslav (approved):

meeroslav:
Thank you! Merging it in as soon as the build it green.

Phillip9587:
@meeroslav CI is finally green

leosvelperez (approved):

JamesHenry (approved):

nzacca:
@Phillip9587 

Hello! May I inquire why Nx is using the package from zkochan vs the official package?

We are using `js-yaml` directly in our project but this change is now causing a conflict as it expects zkochan's package as a peer dependency. I can't seem to find the actual source for the zkochan package to compare what is different to assess if it is OK to change.

Any insight for the differences would be appreciated! Thank you!

PR #23428, opened 2024-05-16
fix(web): stop the patch if idx is undefined

hazemessam:
## Current Behavior
The following patch `node_modules/@nx/web/src/migrations/update-15-5-4/update-babel-preset.js` fails and causing to stop the migration plan when migrating from Nx v15 to v16 in case of there is no `presets` defined in the `.babelrc` file.
Screenshot 2024-05-16 at 1 01 34 PM

## Expected Behavior
Should stop the patch and continue the migration plan.

### DIFF
--- a/packages/web/src/migrations/update-15-5-4/update-babel-preset.ts
+++ b/packages/web/src/migrations/update-15-5-4/update-babel-preset.ts
@@ -33,7 +33,7 @@ export default async function update(tree: Tree) {
         : p[0] === '@nrwl/web/babel'
     );
 
-    if (idx === -1) return;
+    if (idx === undefined || idx === -1) return;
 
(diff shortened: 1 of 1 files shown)


AgentEnder (approved):

FrozenPandaz:
I'll port this over to `18.3.x` rather than `16.0.x`. #23449 

You'll be able to migrate to `18.3.5` once it is released. Or you can make the modification locally and rerun the migrations.

I don't think I would release a `16.0.x` version with this fix. That's an old version, and it would be pretty hard to catch every version which people might migrate to.

But I'll patch it in the `18.3.x` branch so that anyone moving to the latest `18` version is not experiencing this issue.

PR #23429, opened 2024-05-16
chore(repo): prepopulate verdaccio storage for e2e-ci

JamesHenry:
## Current Behavior

All `e2e-ci` targets independently run a local release of all our packages to seed verdaccio with the necessary data to be able run installs within test workspaces.

## Expected Behavior

Upfront we perform a single local release of all our packages to create a valid copy of verdaccio storage data. This is then reused via the Nx Replay cache to seed the verdaccio instances of each `e2e-ci` target.

This takes the time spent over and above the raw `jest` execution for each `e2e-ci` target down from 90-110s to 5-10s.

## Related Issue(s)

Fixes #


e2e/eslint/project.json
```diff
@@ -6,18 +6,18 @@
...
       "inputs": ["e2eInputs", "^production"]
     },
     "e2e-ci--src/linter-legacy.test.ts": {
+      "dependsOn": [
+        "nx:build-native",
+        "@nx/nx-source:populate-local-registry-storage"
+      ],
       "inputs": ["e2eInputs", "^production"]
-    },
-    "run-e2e-tests": {
```
JamesHenry, line 14: Removed left over reference

e2e/remix/project.json
```diff
@@ -5,5 +5,13 @@
   "projectType": "application",
   "implicitDependencies": ["remix"],
   "// targets": "to see all targets run: nx show project e2e-remix --web",
-  "targets": {}
+  "targets": {
+    "e2e-ci--src/nx-remix.test.ts": {
```
JamesHenry, line 9: Added missing target config

e2e/utils/global-setup.ts
JamesHenry, line 1: This is not an easy diff to read, probably better to review the final state of the file in isolation

patches/@nx__js@19.0.0-rc.1.patch
JamesHenry, line 1: This applies the same change as in packages/js/src/plugins/jest/start-local-registry.ts so that it is immediately available for use.

The patch (and reference to it in package.json) can be removed once a version of Nx is released that contains this PR.

Coly010 (approved):

scripts/local-registry/populate-storage.js
```diff
@@ -0,0 +1,56 @@
...
+    const registryTeardown = await startLocalRegistry({
+      localRegistryTarget: '@nx/nx-source:local-registry',
+      verbose: isVerbose,
+      clearStorage: true,
+    });
+
+    console.log('Publishing packages to local registry to populate storage');
+    await runLocalRelease(publishVersion, isVerbose);
+
+    registryTeardown();
```
FrozenPandaz, line 22: You can do this as a followup since this is an existing issue but could you please ensure that when publishing fails, that `registryTeardown()` is called please? And I guess we have to fix it in the version which does the local release as well.

FrozenPandaz (approved):

scripts/local-registry/populate-storage.js
JamesHenry: https://github.com/nrwl/nx/pull/23470

PR #23449, opened 2024-05-16
fix(web): stop the patch if idx is undefined

FrozenPandaz:
## Current Behavior
The following patch `node_modules/@nx/web/src/migrations/update-15-5-4/update-babel-preset.js` fails and causing to stop the migration plan when migrating from Nx v15 to v16 in case of there is no `presets` defined in the `.babelrc` file.
Screenshot 2024-05-16 at 1 01 34 PM

## Expected Behavior
Should stop the patch and continue the migration plan.

Closes https://github.com/nrwl/nx/pull/23428


FrozenPandaz:
Pretty sure the e2e failures here are unrelated.

AgentEnder (approved):

PR #23478, opened 2024-05-17
feat(web): update postcss to a version without security advisory

jaysoo:
https://github.com/advisories/GHSA-7fh5-64p2-3v2j

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/migrations.json
+++ b/packages/react/migrations.json
@@ -193,6 +193,15 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "19.2.0": {
+      "version": "19.2.0-beta.8",
+      "packages": {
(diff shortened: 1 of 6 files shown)


xiongemi (approved):

packages/react/migrations.json
```diff
@@ -193,6 +193,15 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "19.0.5": {
+      "version": "19.0.5-beta.0",
```
FrozenPandaz, line 198: Can you update the version here? `19.1.1` if you think you can get this in today?

FrozenPandaz (changes requested):

Issue #9238, opened 2022-03-08
@nx/web:app "strict" option


ziacik:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/nrwl/nx/blob/master/CONTRIBUTING.md))

## Description
_@nrwl/js:lib_, _@nrwl/react:app_ and many others have a `"strict"` option, `@nrwl/web:app` should have one too.

## Motivation
To have consistency between generators, and to be able to generate strict apps, by default.

ziacik:
Don't close, please.

ziacik:
Not stale. Still no strict option in @nx/web:application generator, still generated with non-strict tsconfigs. In 16.5.5.

FrozenPandaz:
Would someone in the community like to open a PR?

ziacik:
I might try to make this one.

PR #23497, opened 2024-05-17
fix(web): Add strict mode for @nx/web

ndcunningham:
Reference: #9238

Essentially reverting: #23472


isaacplmann (approved):
Docs are fine

Issue #4159, opened 2020-11-26
Feature Request: nx format:check should provide a workaround for Prettier not supporting format diffing


jacopolanzoni:
# UPDATE FROM NX TEAM (2024-05-17)

Please read: https://github.com/nrwl/nx/issues/4159#issuecomment-2118165264

---

## Description
As `nx format:check` shows which files fail the format check but not what is actually wrong in them, I would love if the `--verbose` option could actually add that information. At the moment, unless I am doing something wrong, `nx format:check --verbose` has the same output as `nx format:check`.

## Motivation
The command would give more information about what is actually wrong in the file format. 

## Suggested Implementation
I would love if the `--verbose` option could actually add that information.

## Alternate Implementations
As an alternative, I would like even only to see either the line where the check fails, or maybe the broken rule.

konrad-garus:
It's like a slap in the face without ever even telling you _why_. Imagine a compiler telling you "Error, GFY, period".

Please add this.

TCModus:
My entire team is complaining about this, going to have to remove my formatting check from pre-commit hook because it is just confusing everyone.  The --verbose flag should be at minimum spitting out the same output of the format:write command which at least shows the command crawling through the project directories doing the formatting.

FYI for others with this issue it seems like you can run prettier using npm and get the output from running the prettier command directly instead of through Nx.

Alspaladin:
Why it's closed? My project is failing on nx format:check without any message besides the `error Command failed with exit code 1.` and --verbose adds nothing to it

eprokofev:
I don't understand "stale" logic as well. My pipeline also failed at this step without any reason. Verbose is not informative at all.

xaphod:
Yeah, reopen this issue - verbose is broken

KlausNie:
Same here

konnic:
Just ran into the same issue, would like to see this issue being reopened...

konnic:
If it helps anyone, running `nx format:write` solved issue for me.

StephenStrickland:
`nx format:check --verbose` is still broken

savvyshell:
How has this not been addressed , it's running fine locally, failing in CI with no information.

savvyshell:
```
init-commands: |
  npx nx-cloud record -- npx prettier --write .
  npx nx-cloud record -- npx nx format:write
  npx nx-cloud record -- npx prettier --check .
  npx nx-cloud record -- npx nx format:check
```

[...]

PR #23503, opened 2024-05-17
fix(core): more helpful output for format:check --verbose

JamesHenry:
Nx format is currently exclusively built on top of prettier. We leverage the `--list-different` feature of prettier to show the user which files failed the formatting check. The simple common case solution is to run `nx format:write` (powered by `prettier --write`) and commit the result.

Right now a user might want to try and learn more about their `format:check` failure by passing `--verbose` but it will not change the output in any way.

Based on the feedback in https://github.com/nrwl/nx/issues/4159 some users are expecting to see the diff output of what correct vs incorrect formatting would look like.

However, because nx format is currently exclusively built on top of prettier, it is therefore subject to its limitations. https://github.com/prettier/prettier/issues/6885 remains unresolved and so there is no native way for users to see a diff.

It is important to note that only showing the names of the files in the default `nx format:check` output is an important feature - it allows easy piping into custom scripts to perform further handling. Therefore we do not want to change the default output here.

This PR updates the case where `--verbose` is passed in order to provide more information/context about the overall problem and how to fix it.

**Before**

image

**After**

image


pujux (approved):
great!

xiongemi (approved):

PR #23508, opened 2024-05-17
fix(node): Enable e2e test

ndcunningham:
Re-enables e2e tests for node 

Docker changes to be merged first: https://github.com/nrwl/nx/pull/26365


e2e/node/src/node-server.test.ts
```diff
@@ -14,11 +14,16 @@ import {
 } from '@nx/e2e/utils';
 import { join } from 'path';
 
-xdescribe('Node Applications + webpack', () => {
+describe('Node Applications + webpack', () => {
```
FrozenPandaz, line 17: There's another `xdescribe` in this file. Could you enable that as well please?

packages/node/src/generators/setup-docker/setup-docker.ts
```diff
@@ -30,14 +28,16 @@ function normalizeOptions(
...
 
   if (
     !projectConfig ||
     !projectConfig.targets ||
     !projectConfig.targets[options.buildTarget]
   ) {
+    logger.error(
+      `Could not find the project ${options.project} or the build target ${options.buildTarget} in the workspace.`
+    );
     return;
```
FrozenPandaz, line 41: Should this throw an error?

FrozenPandaz (changes requested):

e2e/node/src/node-server.test.ts
ndcunningham: there are no `xdescribe`s in this file you can see the output here: https://staging.nx.app/runs/oaKPGjBQVJ?utm_source=pull-request&utm_medium=comment

packages/node/src/generators/setup-docker/setup-docker.ts
ndcunningham: Updated

FrozenPandaz (changes requested):

packages/node/src/generators/setup-docker/setup-docker.ts
```diff
@@ -30,15 +28,16 @@ function normalizeOptions(
 }
 
 function addDocker(tree: Tree, options: SetUpDockerOptions) {
-  // Inferred targets are only available in the project graph
-  const projectConfig = readCachedProjectConfiguration(options.project);
+  const projectConfig = readProjectConfiguration(tree, options.project);
```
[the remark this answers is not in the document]
ndcunningham: Why not?

1. The project may not exist on file as yet so reading only from the graph would be invalid.
2. The `tree` is passed in as an argument, which is probably the source of truth for generators.

The current setup should work for both crystal and non-inferred projects.

jaysoo (approved):

e2e/node/src/node-server.test.ts
```diff
@@ -14,11 +14,16 @@ import {
...
   let proj: string;
   beforeAll(() => {
     proj = newProject({
       packages: ['@nx/node'],
+      // npm has resolution for ajv some packages require ajv6 and some require ajv8 and npm resolves it to ajv6 (Error: Cannot find module 'ajv/dist/compile/codegen')
```
FrozenPandaz, line 22: We should at least add a //TODO to re-enable the tests for other package managers.

@jaysoo Do you know how we can either align these versions or mitigate this issue with the default options of npm?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #23668, opened 2024-05-20
docs(core): fix dependsOn snippet in project configuration reference

chabb:
I'm not totally sure, but by reading the docs, the intent of the first example is to say that the test target depends of the build target. Otherwise, it would mean that the build target depends of the build target, which does not make sense. Or did I misunderstood something in the doc ?


leosvelperez (approved):
Thanks for contributing!

PR #24052, opened 2024-05-22
docs(nx-cloud): Change exec to dlx for GHA Nx cloud agent command.

TheWrightDev:
Node modules are installed yet so exec causes `Command "nx-cloud" not found. ` error.

## Current Behavior
The suggested command in recommended position causes an error.

## Expected Behavior
The suggested command in recommended position would not cause an error.

## Related Issue(s)

Fixes #


StalkAltan (approved):
Thank you for the PR!

Issue #23904, opened 2024-05-21
Installation doesn't work on multi-user computers


Ten0:
### Current Behavior
While attempting to install:
```
Error: EACCES: permission denied, mkdir '/tmp/nx-native-file-cache/7922ff398abcaa4ba02c132005c3f9a29c05714432c8da5d5607e32b0f2c7602'
```
This happens because the `/tmp/nx-native-file-cache` directory is owned by the first user that happens to have tried to install `nx`.

### Expected Behavior

Doesn't use global static-path temporary files, thus allowing multiple users to have nx somewhere in their dependencies.
Also, one user shouldn't trust the file downloaded by another user, otherwise that would lead to privilege escalation security vulnerability (esp. since `/tmp` is world-writable), so the cache should not be shared between users.
Probably this can be fixed by appending the uid of the user to the path, or better yet storing to `~/.cache` (if this is actually cache).

### Steps to Reproduce

1. install the lib on one user
2. install the lib on another user

### Package Manager Version

pnpm 8.15.5

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

Cammisuli:
fixed by #24326

PR #24326, opened 2024-05-22
fix(core): use current user when hashing native file & enable setting its directory via env

MaxKless:
## Current Behavior
The native file hashing takes only the nx version & workspace path into account. 
There is also no way to control the storage location of native `.node` files used by nx.

## Expected Behavior
The native file hashing takes the nx version, workspace path & current username into account, making sure the cache isn't shared between multiple users on the same device.
It's also possible to control the exact location of the native file cache by setting the `NX_NATIVE_FILE_CACHE_DIRECTORY` environment variable if you need to move it somewhere else.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/native-file-cache-location.ts
+++ b/packages/nx/src/native/native-file-cache-location.ts
@@ -15,6 +15,10 @@ function getNativeFileCacheBaseFolder() {
   if (process.env.NX_NATIVE_FILE_CACHE_DIRECTORY) {
     return process.env.NX_NATIVE_FILE_CACHE_DIRECTORY;
   } else {
-    return join(tmpdir(), 'nx-native-file-cache');
+    const shortUserHash = createHash('sha256')
+      .update(userInfo().username)
+      .digest('hex')
+      .substring(0, 7);
+    return join(tmpdir(), `nx-native-file-cache-${shortUserHash}`);
   }
 }
--- a/packages/nx/src/native/native-file-cache-location.ts
+++ b/packages/nx/src/native/native-file-cache-location.ts
@@ -3,22 +3,15 @@ import { join } from 'path';
 import { createHash } from 'crypto';
 import { workspaceRoot } from '../utils/workspace-root';
 
-export const nativeFileCacheLocation = join(
-  getNativeFileCacheBaseFolder(),
-  createHash('sha256')
-    .update(workspaceRoot)
-    .update(userInfo().username)
-    .digest('hex')
-);
-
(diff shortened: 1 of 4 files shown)


isaacplmann (approved):

FrozenPandaz (approved):

packages/nx/src/native/native-file-cache-location.ts
```diff
@@ -1,10 +1,24 @@
...
+
+function getNativeFileCacheBaseFolder() {
+  if (process.env.NX_NATIVE_FILE_CACHE_DIRECTORY) {
+    return process.env.NX_NATIVE_FILE_CACHE_DIRECTORY;
+  } else {
+    const shortUserHash = createHash('sha256')
+      .update(userInfo().username)
+      .digest('hex')
+      .substring(0, 7);
+    return join(tmpdir(), `nx-native-file-cache-${shortUserHash}`);
```
FrozenPandaz, line 22: Do we need a `shortUserHash` here as well as in the `hash` for the child directory?

right now we have...

`/tmp/[F(username)]/[F(username, workspaceRoot]`

We should be able to have..

`/tmp/[F(username, workspaceRoot)]` right?

FrozenPandaz (dismissed):

FrozenPandaz (approved):

Ten0:
because the hash is deterministic and tmp is world-writable and regularly cleaned up, if the client assumes that the node binary is correct, isn't that a security issue?
Another user could place a trojaned binary in the path that this user would use as a cache hit, no?
(This is the attack scenario I described at https://github.com/nrwl/nx/issues/23904#issue-2308595513)

MaxKless:
if you're worried about other users on your system, you can set the location to a folder only you have access to via the environment variables

Ten0:
It looks like alternately this could be stored in `~/.cache`.

I have this as a transitive dependency that I didn't particularly ask for.

Expecting from all users that they realize that they have this dependency, inspect all of its inner workings to realize there is a security issue, and finally set an env variable to mitigate it does not seem like a sound way to handle security issues.

MaxKless:
That's a valid point. I'm hesitant to set the directory to something user-scoped because of potentially different permission handling across OS. My biggest priority is avoiding errors when using Nx because copying that file is required and if there is a permission issue, we could cause errors for everyone - not just those with hypothetical malicious other users on the same machine. 
For now the environment variable is the way to go but I see your point and will reconsider this.

PR #24481, opened 2024-05-22
fix(linter): fix js-yaml peerDependency

Phillip9587:
## Current Behavior

peerDependency was set to a alias

## Expected Behavior

peerDependency on `js-yaml`

## Related Issue(s)

#23422
#25287

### DIFF
--- a/packages/eslint/package.json
+++ b/packages/eslint/package.json
@@ -30,7 +30,7 @@
   "generators": "./generators.json",
   "executors": "./executors.json",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


FrozenPandaz:
@meeroslav Was this solved with https://github.com/nrwl/nx/pull/25999?

meeroslav:
> @meeroslav Was this solved with #25999?

Yes, as explained previously at https://github.com/nrwl/nx/pull/25999#discussion_r1613282859 we do not want to make this change right now.

If there's a need we can revisit it later.

PR #25060, opened 2024-05-22
feat(webpack): Support incremental builds via buildLibsFromSource 

ndcunningham:
This PR adds the ability for incremental builds when using the Webpack Plugin.

Instead of using the source library directly, you can now utilize the output folder by utilizing the `buildLibsFromSource` option within your webpack.config file, through `NxAppWebpackPlugin`. This means that instead of accessing `mylib/src/index.ts`, it will access `dist/mylib/index.js`.

This directly aligns with incremental builds as it ensures that the build process only recompiles the source doe that has been modified since the last build.

### DIFF
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
@@ -219,9 +219,8 @@ function applyNxDependentConfig(
   { useNormalizedEntry }: { useNormalizedEntry?: boolean } = {}
 ): void {
   const tsConfig = options.tsConfig ?? getRootTsConfigPath();
-  const plugins: WebpackPluginInstance[] = [
-    new NxTsconfigPathsWebpackPlugin({ tsConfig }),
-  ];
+  const plugins: WebpackPluginInstance[] = [];
+
   const executorContext: Partial<ExecutorContext> = {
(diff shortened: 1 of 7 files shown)


packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
```diff
@@ -230,6 +234,41 @@ function applyNxDependentConfig(
...
+        .filter((dependency) => dependency.node.type === 'lib')
+        .map((dependency) => dependency.node.name)
+        .join(',');
+
+      const buildCommand = `nx run-many --target=build --projects=${buildableDependencies}`;
+
+      plugins.push(new WebpackNxBuildCoordinationPlugin(buildCommand));
+    }
+  }
+  plugins.push(new NxTsconfigPathsWebpackPlugin({ tsConfig }));
```
Coly010, line 270: Can we move this to the `NxTsconfigPathsWebpackPlugin`?
It would prevent adding this plugin twice and is a more correct location for the logic

Coly010 (changes requested):

therk (commented):
@ndcunningham what is the easiest way to set buildLibFromSource to false without creating a custom webpack.config.js file? This change is causing failures form my project when used with cypress component tests library that is referencing Storybook Library module in the project.

PR #25999, opened 2024-05-24
fix(core): use zkochan/js-yaml directly to avoid false audit errors

meeroslav:
Some of the audit tools have been falsely flagging the alias to `@zkochan/js-yaml` as `js-yaml@0.0.7` (which has security holes) so we decided to use the package explicitly.

## Related Issue(s)

Fixes #


JamesHenry (approved):

JamesHenry (approved):

rarmatei (approved):

packages/eslint/package.json
```diff
@@ -31,7 +31,7 @@
   "generators": "./generators.json",
   "executors": "./executors.json",
   "peerDependencies": {
-    "js-yaml": "npm:@zkochan/js-yaml@0.0.7"
+    "@zkochan/js-yaml": "0.0.7"
```
Phillip9587, line 34: @meeroslav Is the peerDependency needed? The nx package also depends on `@zkochan/js-yaml` so we can move it to the dependencies

meeroslav: True, but it doesn't change the requirement need for this one. The `eslint` package only needs `js-yaml` if user needs to migrate yaml-based config to flat config.

Phillip9587: But it would eliminate the user confusion why the @zochkan/js-yaml package has to be installed instead of js-yaml. And it would allow existing usage to function without changes...

meeroslav: The only things we could do is make `eslint` depend on `js-yaml` since it has no need for Zoltan's overrides as it uses plain `load` and `dump`. But the entire usage might be completely removed once code mods from `eslint`'s team land.

On top of that we generally don't have support for `yaml` configs, so the chance of someone using converter to convert `yaml` rc to flat config are almost zero.

If there is an issue created later on we can address it.

PR #26007, opened 2024-05-24
feat(nx-cloud): get git remote info to use in cloud connect

mandarini:
CLOSED in favor of: https://github.com/nrwl/nx/pull/26262

THIS PR IS THE FIRST IN A CHAIN OF PRs. The last PR in the chain is: https://github.com/nrwl/nx/pull/26262

So if https://github.com/nrwl/nx/pull/26262 is OK, then maybe we just merge https://github.com/nrwl/nx/pull/26262?

Get info from setup remotes (if any), identify if github, if github get github slug, to be able to generate GH cloud onboarding URL.

Todo:
- [x] hide behind feature flag
- [x] create commit message for PR
- [x] add PR instructions

Example screenshots from `nx init` and `nx connect`:

Screenshot 2024-05-29 at 4 15 19 PM

Screenshot 2024-05-29 at 4 15 36 PM


packages/nx/src/command-line/connect/connect-to-nx-cloud.ts
```diff
@@ -51,12 +52,39 @@ export async function connectToNxCloudIfExplicitlyAsked(
...
+        `Unable to authenticate. Either define accessToken in nx.json or set the NX_CLOUD_ACCESS_TOKEN env variable.`
+      );
+    }
+
     output.log({
       title: '✔ This workspace already has Nx Cloud set up',
       bodyLines: [
         'If you have not done so already, connect your workspace to your Nx Cloud account:',
-        `- Login at ${getNxCloudUrl(nxJson)} to connect your repository`,
+        `- Do so by going to the following URL:
```
jaysoo, line 85: I'd reword this to be more terse.

`- Connect with Nx Cloud at ${connectCloudUrl}`

Just to reinforce what going to the URL will do.

packages/nx/src/nx-cloud/generators/connect-to-nx-cloud/connect-to-nx-cloud.ts
```diff
@@ -72,18 +73,32 @@ async function createNxCloudWorkspace(
...
+  let connectCloudUrl;
+  if (githubSlug) {
+    connectCloudUrl = `${apiUrl}/setup/connect-workspace/vcs?provider=GITHUB&selectedRepositoryName=${encodeURIComponent(
+      githubSlug
+    )}`;
+  } else {
+    connectCloudUrl = `https://cloud.nx.app/setup/connect-workspace/manual?accessToken=${token}`;
+  }
   output.note({
     title: `Your Nx Cloud workspace is public`,
```
jaysoo, line 103: We've received a lot of feedback that this message is kinda scary, and users aren't sure if they should be concerned: `Your Nx Cloud workspace is public`. 

Do we want to reword it while we're in here? Something like:

```
Your Nx Cloud workspace is ready

To claim it, connect it to your Nx Cloud account:
- Push your changes
- Connect your workspace here ${connectCloudUrl}
```

packages/nx/src/nx-cloud/utilities/git-info.ts
```diff
@@ -0,0 +1,41 @@
+import { execSync } from 'child_process';
+
+export function getGithubSlugOrNull(): string | null {
+  const gitRemote = execSync('git remote -v').toString();
+  return extractUserAndRepoFromGitHubUrl(gitRemote);
+}
+
+function extractUserAndRepoFromGitHubUrl(gitRemotes: string): string | null {
```
jaysoo, line 8: A unit test to show this working would be good.

jaysoo: Also, if we only support `origin` as remote name, should it be `git config --get remote.origin.url`?

mandarini: We don't really only support origin. Here's the thing.

If a user has `origin` set up and the `origin` is a github URL, then we'll use that. If the user does not have a remote named `origin`, then we pick the first remote in the list that has a github URL. 

Even if we do pick `origin` or in the absence of `origin` we pick the first remote with github, it still does not guarantee that it's the remote they want to set up. they could be working with multiple forks, or custom named remotes, etc. So we either pick origin or just one, and if it's the wrong one, then they'll have to change the URL themselves.

We could list their remotes and tell them "pick the one you want" but I think it's an overkill?

mandarini: (or `git remote get-url origin` :P )

mandarini: I'm adding a unit test now!

packages/nx/src/nx-cloud/utilities/git-commit.ts
```diff
@@ -0,0 +1,42 @@
+import * as chalk from 'chalk';
+import { logger } from '../../utils/logger';
+import { commitChanges } from '../../command-line/migrate/migrate';
```
FrozenPandaz, line 3: Can we move this to a different file rather than import it from `migrate`?

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -1588,7 +1588,7 @@ function commitChanges(commitMessage: string): string | null {
   return getLatestCommitSha();
 }
 
-function getLatestCommitSha(): string | null {
+export function getLatestCommitSha(): string | null {
```
FrozenPandaz, line 1591: I don't think this needs to be exported?

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -1573,7 +1573,7 @@ function getStringifiedPackageJsonDeps(root: string): string {
   }
 }
 
-function commitChanges(commitMessage: string): string | null {
+export function commitChanges(commitMessage: string): string | null {
```
FrozenPandaz, line 1576: Move this into some nx utils for git.

FrozenPandaz (changes requested):

packages/nx/src/command-line/migrate/migrate.ts
mandarini: oooops!

packages/nx/src/utils/git-utils.ts
```diff
@@ -0,0 +1,69 @@
+import { execSync } from 'child_process';
+
+export function getGithubSlugOrNull(): string | null {
+  const gitRemote = execSync('git remote -v').toString();
```
jdpearce, line 4: Could we use `git ls-remote --get-url` instead? We wouldn't have to parse anything then.

packages/nx/src/utils/git-utils.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { extractUserAndRepoFromGitHubUrl } from './git-utils';
+
+describe('extractUserAndRepoFromGitHubUrl', () => {
```
jdpearce, line 3: What about the case where the remote starts with `https`?

packages/nx/src/nx-cloud/generators/connect-to-nx-cloud/connect-to-nx-cloud.ts
```diff
@@ -72,20 +73,55 @@ async function createNxCloudWorkspace(
...
-  let origin = 'https://nx.app';
-  try {
-    origin = new URL(url).origin;
-  } catch (e) {}
+function printSuccessMessage(
+  url: string,
+  token: string,
+  installationSource: string
+) {
+  if (process.env.NX_NEW_CLOUD_ONBOARDING !== 'true') {
```
jdpearce, line 81: What's the plan to remove this?

packages/nx/src/utils/git-utils.ts
mandarini: I want to make sure we're getting the right remote, this is why I'm listing all.

packages/nx/src/nx-cloud/generators/connect-to-nx-cloud/connect-to-nx-cloud.ts
mandarini: I have no idea!

mandarini: I assume when we are ready to release the feature?

PR #26011, opened 2024-05-24
fix(linter): migrate no-extra-semi rules into user config, out of nx extendable configs

JamesHenry:
## Current Behavior

The `typescript` and `javascript` shared configs that Nx exposes via `@nx/eslint-plugin` reference `no-extra-semi` and `@typescript-eslint/no-extra-semi`, meaning they are leveraged in any user configs which extend from those shared configs.

## Expected Behavior

The upcoming v8 of typescript-eslint removes the `@typescript-eslint/no-extra-semi` rule because it has been migrated to live under the https://eslint.style umbrella for stylistic rules:

https://deploy-preview-9022--typescript-eslint.netlify.app/blog/announcing-typescript-eslint-v8-beta/

This means users will need to decide if they want to start using ESLint Stylistic to continue using these rules (which we should not force upon users, or even particularly encourage IMO) or remove the rules from their configs.

In order to make this much more explicit and straightforward, this PR removes the rules from our shared configs and migrates them into applicable user `.eslintrc.json` config.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint-plugin/migrations.json
+++ b/packages/eslint-plugin/migrations.json
@@ -14,9 +14,9 @@
     },
     "update-19-1-0-rename-no-extra-semi": {
       "cli": "nx",
-      "version": "19.1.0-beta.4",
+      "version": "19.1.0-beta.6",
       "description": "Migrate no-extra-semi rules into user config, out of nx extendable configs",
-      "implementation": "./src/migrations/update-19-1-0-rename-no-extra-semi/migrate-no-extra-semi"
+      "implementation": "./src/migrations/update-19-1-0-migrate-no-extra-semi/migrate-no-extra-semi"
     }
   },
   "packageJsonUpdates": {},
--- a/packages/eslint-plugin/migrations.json
+++ b/packages/eslint-plugin/migrations.json
@@ -11,6 +11,12 @@
       "version": "17.2.6-beta.1",
       "description": "Rename workspace rules from @nx/workspace/name to @nx/workspace-name",
       "implementation": "./src/migrations/update-17-2-6-rename-workspace-rules/rename-workspace-rules"
+    },
+    "update-19-1-0-rename-no-extra-semi": {
+      "cli": "nx",
+      "version": "19.1.0-beta.4",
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/migrations.json
```diff
@@ -11,6 +11,12 @@
       "version": "17.2.6-beta.1",
       "description": "Rename workspace rules from @nx/workspace/name to @nx/workspace-name",
       "implementation": "./src/migrations/update-17-2-6-rename-workspace-rules/rename-workspace-rules"
+    },
+    "update-19-1-0-rename-no-extra-semi": {
+      "cli": "nx",
+      "version": "19.1.0-beta.4",
```
leosvelperez, line 17: 19.1.0-beta.5 is the latest beta version released. It's not a huge deal because folks would normally only migrate to stable versions, but repos that might have migrated to it wouldn't get this migration.
```suggestion
      "version": "19.1.0-beta.6",
```

JamesHenry: Thanks, @leosvelperez, pushed as part of path fix

packages/eslint-plugin/src/migrations/update-19-1-0-migrate-no-extra-semi/migrate-no-extra-semi.ts
```diff
@@ -0,0 +1,71 @@
+import {
+  Tree,
+  formatFiles,
+  updateJson,
+  visitNotIgnoredFiles,
+} from '@nx/devkit';
+
+export default async function migrate(tree: Tree): Promise<void> {
+  visitNotIgnoredFiles(tree, '.', (path) => {
+    if (!path.endsWith('.eslintrc.json')) {
```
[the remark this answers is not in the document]
JamesHenry: No, that will only be a small subset of users at this point (who are likely to be more advanced and up to date with the latest developments anyway), and it's much, much harder to do on the dynamic JS files.

FrozenPandaz (approved):

PR #26024, opened 2024-05-24
fix(core): include dependencies when hashing nx executors

FrozenPandaz:
## Current Behavior

Dependencies of executors are not hashed when hashing a task.

## Expected Behavior

Dependencies of executors are hashed when hashing a task.

## Related Issue(s)

Fixes #


Cammisuli (approved):

Elyahou:
Can we have more details on this one ? Does it mean that inputs will automatically include external dependencies of the underlying plugin ? Is it only for official nx executors or also custom ones ?

PR #26119, opened 2024-05-26
fix(gradle): use os specific line separator for dependency parsing

svetch:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/plugin/dependencies.ts
+++ b/packages/gradle/src/plugin/dependencies.ts
@@ -12,6 +12,7 @@ import { basename } from 'node:path';
 import {
   getGradleReport,
   invalidateGradleReportCache,
+  newLineSeparator,
 } from '../utils/get-gradle-report';
 import { writeTargetsToCache } from './nodes';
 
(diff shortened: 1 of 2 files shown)


xiongemi (approved):

packages/gradle/src/plugin/dependencies.ts
```diff
@@ -89,7 +90,7 @@ function processGradleDependencies(
   context: CreateDependenciesContext
 ): Set<RawProjectGraphDependency> {
   const dependencies: Set<RawProjectGraphDependency> = new Set();
-  const lines = readFileSync(depsFile).toString().split('\n');
+  const lines = readFileSync(depsFile).toString().split(newLineSeparator);
```
Den-dp, line 93: are there any guarantees that this file will have `\r\n` on windows?
I'm asking cause depending on the nature of this file - it also might use `\n` even on Windows (for example if it lives under the git source control with configured `autocrlf=input`).
As a more robust alternative - `/\r?\n/` regex might be used

PR #26154, opened 2024-05-28
fix(release): improve error handling for npm publish

vensller:
This change improves error handling for npm publish function. When there is any error that is not handled by the catch clause, it is throwing an "cannot read summary of undefined", because the stdoutData.error is undefined. With this change, it is going to correctly throw the root error that was the cause of the problem.

## Current Behavior
When an unexpected error occurs on npm publish, it is trowing "cannot read summary of undefined" error.

## Expected Behavior
Throws the exact error that was the root cause of the problem

## Related Issue(s)

Fixes #


JamesHenry (changes requested):
Thanks @vensller - please redo your commit message to start with `fix(release):`

The scope is required, and this is not a feature, it's a fix.

You can copy what I just did for the PR title

PR #26165, opened 2024-05-28
fix(angular): update angular-eslint to v18

JamesHenry:
## Current Behavior

angular-eslint v17 is used

## Expected Behavior

angular-eslint v18 is used

## Related Issue(s)

Fixes #


packages/angular/migrations.json
```diff
@@ -1887,6 +1887,27 @@
...
+    },
+    "19.1.2": {
+      "version": "19.1.2-beta.1",
+      "requires": {
+        "eslint": "^8.57.0 || ^9.0.0",
+        "@angular/core": ">= 18.0.0 < 19.0.0"
+      },
+      "packages": {
+        "@angular-eslint/eslint-plugin": {
+          "version": "18.0.0",
```
leosvelperez, line 1899:
```suggestion
          "version": "~18.0.0",
```

packages/angular/migrations.json
```diff
@@ -1887,6 +1887,27 @@
...
+          "alwaysAddToPackageJson": false
+        },
+        "@angular-eslint/eslint-plugin-template": {
+          "version": "18.0.0",
```
leosvelperez, line 1903:
```suggestion
          "version": "~18.0.0",
```

packages/angular/migrations.json
```diff
@@ -1887,6 +1887,27 @@
...
+          "alwaysAddToPackageJson": false
+        },
+        "@angular-eslint/template-parser": {
+          "version": "18.0.0",
```
leosvelperez, line 1907:
```suggestion
          "version": "~18.0.0",
```

packages/angular/src/utils/versions.ts
```diff
@@ -16,7 +16,7 @@ export const typesExpressVersion = '4.17.14';
 export const browserSyncVersion = '^3.0.0';
 export const moduleFederationNodeVersion = '~1.0.5';
 
-export const angularEslintVersion = '~17.3.0';
+export const angularEslintVersion = '18.0.0';
```
leosvelperez, line 19:
```suggestion
export const angularEslintVersion = '~18.0.0';
```

leosvelperez (changes requested):
@JamesHenry I missed this yesterday. We should have new projects and migrations to install, at least, the latest patch version at the time of generation. I think we could even have them install the latest minor. We've always used the `~` until now, but we could start using `^` if you think it's safe to do so.

leosvelperez (approved):

FrozenPandaz (approved):

PR #26237, opened 2024-05-29
feat(nx-dev): improve the enteprise page

meeroslav:
https://nx-dev-git-docs-improve-enterprise-page-nrwl.vercel.app/enterprise

## Related Issue(s)

Fixes #


nx-dev/ui-enterprise/src/lib/download-ebook.tsx
```diff
@@ -25,9 +25,7 @@ export function DownloadEbook(): JSX.Element {
         </svg>
         <div className="mx-auto max-w-md text-center lg:mx-0 lg:flex-auto lg:py-12 lg:text-left">
           <h2 className="text-3xl font-bold tracking-tight text-slate-950 sm:text-4xl dark:text-slate-50">
-            Download
-            <br />
-            Fast CI for monorepos
+            Download our ebook
```
juristr, line 28: Question is...is it an ebook.. I think that's why it was called like this. But I'm fine with the change

juristr (approved):

PR #26247, opened 2024-05-29
cleanup(devkit): add util for processing outputs in crystal migration generators

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/generators/plugin-migrations/executor-to-plugin-migrator.ts
+++ b/packages/devkit/src/generators/plugin-migrations/executor-to-plugin-migrator.ts
@@ -29,8 +29,9 @@ import type { ConfigurationResult } from 'nx/src/project-graph/utils/project-con
 type PluginOptionsBuilder<T> = (targetName: string) => T;
 type PostTargetTransformer = (
   targetConfiguration: TargetConfiguration,
-  tree?: Tree,
-  projectDetails?: { projectName: string; root: string }
+  tree: Tree,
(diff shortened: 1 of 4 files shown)


jaysoo (approved):

Coly010 (approved):

ndcunningham (approved):

packages/devkit/src/generators/plugin-migrations/executor-to-plugin-migrator.ts
```diff
@@ -29,8 +29,9 @@ import type { ConfigurationResult } from 'nx/src/project-graph/utils/project-con
 type PluginOptionsBuilder<T> = (targetName: string) => T;
 type PostTargetTransformer = (
   targetConfiguration: TargetConfiguration,
-  tree?: Tree,
-  projectDetails?: { projectName: string; root: string }
+  tree: Tree,
+  projectDetails: { projectName: string; root: string },
+  inferredTargetConfiguration: TargetConfiguration
```
Coly010, line 34: With these not being optional now, will it break any existing migrations?

leosvelperez: It doesn't. The function is always invoked with those params provided. They are not optional. It's up to the consumer to use them or not. You can see the build and tests are still passing.

FrozenPandaz (approved):

PR #26248, opened 2024-05-29
chore(repo): add logging for main pipeline e2e targets

lourw:
Enable detailed logging on main CI pipeline job.

## Related Issue(s)

Fixes #


StalkAltan (approved):

FrozenPandaz:
Enabling this usually causes tests to fail because the test looks for output which is different when verbose is added.

lourw:
> Enabling this usually causes tests to fail because the test looks for output which is different when verbose is added.

No worries, I think I saw enough of the logging output to not need this anymore.

PR #26251, opened 2024-05-29
chore(repo): more eagerly seed the local registry storage in CI

JamesHenry:


FrozenPandaz:
This might be.. a little too eager? We should probably wait for the builds to be done first?

JamesHenry:
@FrozenPandaz I'll close this one for now pending the discussion you are having with Altan

PR #26259, opened 2024-05-30
feat(testing): add convert-to-inferred migration generator for jest

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/generators/convert-to-inferred/convert-to-inferred.ts
+++ b/packages/jest/src/generators/convert-to-inferred/convert-to-inferred.ts
@@ -1,0 +1,337 @@
+import type { Config } from '@jest/types';
+import {
+  createProjectGraphAsync,
+  formatFiles,
+  type TargetConfiguration,
+  type Tree,
+} from '@nx/devkit';
+import { migrateExecutorToPlugin } from '@nx/devkit/src/generators/plugin-migrations/executor-to-plugin-migrator';
+import {
+  processTargetOutputs,
+  toProjectRelativePath,
+} from '@nx/devkit/src/generators/plugin-migrations/plugin-migration-utils';
+import { readConfig } from 'jest-config';
+import { join, normalize, posix } from 'node:path';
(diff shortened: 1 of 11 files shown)


packages/jest/src/generators/convert-to-inferred/convert-to-inferred.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+async function postTargetTransformer(
+  target: TargetConfiguration,
+  tree: Tree,
+  projectDetails: { projectName: string; root: string },
+  inferredTarget: TargetConfiguration
+): Promise<TargetConfiguration> {
+  if (target.options) {
+    const jestConfigPath = target.options.jestConfig;
```
Coly010, line 66: Is there any chance this can be undefined and the jest executor still works?
I ask because you use this value here: https://github.com/nrwl/nx/pull/26259/files#diff-c440c2d41bff855c8035e19fe4948557afa0785f9508f9dfba1ce53cb7856442R119

Coly010 (approved):

packages/jest/src/generators/convert-to-inferred/convert-to-inferred.ts
leosvelperez: That's a required option, so it _should_ always be defined. The executor doesn't work without it. That said, I guess there's nothing stopping folks from having an unused, invalid target without it. What are we doing in those cases?

isaacplmann (approved):

ndcunningham (approved):

packages/jest/src/generators/convert-to-inferred/schema.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/schema",
+  "$id": "NxCypressConvertToInferred",
```
ndcunningham, line 3:
```suggestion
  "$id": "NxJestConvertToInferred",
```

jaysoo (approved):

Issue #26169, opened 2024-05-28
Error start remotes Nx 19.1.0


AlanGRutter:
### Current Behavior

After upgrading and migrating Nx to 19.1.0, when I try to start my module federated host application using nx serve, I get a TypeError as shown in the failure logs.

If I use the command nx serve and specify --devRemotes, it works without any problems.

### Expected Behavior

The host app should be served correctly without errors.

### Steps to Reproduce

1. I don't have a repo in which to reproduce.

### Nx Report

```shell
Node   : 20.12.2
OS     : linux-x64
yarn   : 4.2.2

nx (global)        : 19.1.0
nx                 : 19.1.0
@nx/js             : 19.1.0
@nx/jest           : 19.1.0
@nx/linter         : 19.1.0
@nx/eslint         : 19.1.0
@nx/workspace      : 19.1.0
@nx/angular        : 19.1.0
@nx/cypress        : 19.1.0
@nx/devkit         : 19.1.0
@nx/eslint-plugin  : 19.1.0
@nx/storybook      : 19.1.0
@nrwl/tao          : 19.1.0
@nx/web            : 19.1.0
@nx/webpack        : 19.1.0
typescript         : 5.4.5
---------------------------------------
Registered Plugins:
@nx/cypress/plugin
@nx/eslint/plugin
@nx/storybook/plugin
---------------------------------------
Community plugins:
@nx-tools/nx-container                : 6.0.1
@storybook/angular                    : 8.1.4
@syncfusion/ej2-angular-base          : 25.2.6
@syncfusion/ej2-angular-buttons       : 25.2.6
@syncfusion/ej2-angular-inputs        : 25.2.6
@syncfusion/ej2-angular-notifications : 25.2.3
ngxtension                            : 3.3.0
```

### Failure Logs

```shell
nx serve web --no-cloud --verbose
[NX CLOUD] Verifying current cloud bundle
[NX CLOUD] Currently installed bundle is invalid, downloading version 2405.23.7.hotfix3  from  https://cloud.nx.app/nx-cloud/static/client-bundle
[NX CLOUD] Done:  /home/alan.rutter/TAGTheAgency/YES/YoungEnterpriseV2/yes-web-portal/.nx/cache/cloud/2405.23.7.hotfix3

 NX   Nx Cloud Manually Disabled

Nx will continue running, but nothing will be written or read from the remote cache.
Run details will also not be available in the Nx Cloud UI.

[...]

PR #26265, opened 2024-05-30
fix(angular):  module federation -  devRemotes is made optional for static remotes

MaxSassen:
The new feature `module-federation: add remote configuration override (https://github.com/nrwl/nx/pull/19694)` causes issues for developers that are using static remotes...

## Current Behavior
After upgrading and migrating Nx to 19.1.0, when I try to start my module federated host application using nx serve, I get a TypeError as shown in the failure logs.

If I use the command nx serve and specify --devRemotes, it works without any problems.

## Expected Behavior
runs without --devRemotes

## Related Issue(s)
https://github.com/nrwl/nx/issues/26169
https://github.com/nrwl/nx/issues/20973

## Fixes 

Not everyone is starting up MFE with --devRemotes but uses static remotes.
this feature causes issues.

### DIFF
--- a/packages/angular/src/executors/module-federation-dev-server/lib/start-dev-remotes.ts
+++ b/packages/angular/src/executors/module-federation-dev-server/lib/start-dev-remotes.ts
@@ -28,7 +28,7 @@ export async function startRemotes(
       'module-federation-dev-server'
     );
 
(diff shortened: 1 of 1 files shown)


MaxSassen:
my bad, clicked to fast for the branch name...

PR #26272, opened 2024-05-30
chore(nx-dev): skip website publish for older minor versions

isaacplmann:
Update the website publish flow to not publish if the current minor is not the latest version.
Moves the logic into a node script

### DIFF
--- a/scripts/release-docs.ts
+++ b/scripts/release-docs.ts
@@ -1,0 +1,34 @@
+import { execSync } from 'child_process';
+
+// The GITHUB_REF_NAME is a full version (i.e. 17.3.2). The branchName will strip the patch version number.
+// We will publish docs to the website branch based on the current tag (i.e. website-17)
+console.log(`Comparing ${process.env.GITHUB_REF_NAME} to npm versions`);
+
+const majorVersion = process.env.GITHUB_REF_NAME?.split('.')[0];
+const minorVersion = process.env.GITHUB_REF_NAME?.split('.')
+  .slice(0, 2)
+  .join('.');
+const releasedVersions: string[] = JSON.parse(
+  execSync(`npm show nx@^${majorVersion} version --json`).toString()
+);
+const versionIsReleased =
+  minorVersion &&
+  releasedVersions.some((version) => version.includes(minorVersion));
+const latestVersion = releasedVersions.slice().sort().pop();
+const versionIsLatest = minorVersion && latestVersion?.includes(minorVersion);
+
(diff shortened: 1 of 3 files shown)


ndcunningham (approved):

scripts/release-docs.ts
```diff
@@ -0,0 +1,34 @@
...
+  .slice(0, 2)
+  .join('.');
+const releasedVersions: string[] = JSON.parse(
+  execSync(`npm show nx@^${majorVersion} version --json`).toString()
+);
+const versionIsReleased =
+  minorVersion &&
+  releasedVersions.some((version) => version.includes(minorVersion));
+const latestVersion = releasedVersions.slice().sort().pop();
+const versionIsLatest = minorVersion && latestVersion?.includes(minorVersion);
```
FrozenPandaz, line 18: I think there's a gap in the logic here.

If the version being published is... `21.1.1`.. `minorVersion` here is `1.1` right? And if the `latestVersion` was `21.11.1`, it would include `1.1`. Both at the end.. and in the middle. 

This would misidentify the `21.1.1` publish as being latest and unintentionally publish the docs as `latest` right?

Now that this is a Node script we can leverage `semver` to do this check. I think it would be easier to read and be more accurate.

Idk if the above scenario will happen anytime soon but could you take another look at this logic so it's more bullet proof?

FrozenPandaz (changes requested):

scripts/release-docs.ts
isaacplmann: `minorVersion` would be `21.1`, but yes the string comparison isn't bullet proof.  I'll use the `semver` library

scripts/release-docs.ts
```diff
@@ -0,0 +1,30 @@
...
+);
+
+const latestVersion = maxSatisfying(releasedVersions, `^${majorVersion}`);
+
+console.log(`Found npm versions:\n${releasedVersions.join('\n')}`);
+
+// Publish if the current version is greater than the latest released version
+
+const branchName = `website-${majorVersion}`;
+if (currentVersion && latestVersion && gt(currentVersion, latestVersion)) {
```
FrozenPandaz, line 21: This happens after the release...

* `currentVersion` will be... the one which was just published
* `latestVersion` will be.... maybe the one which was just published. It won't be higher though. So I don't think this condition will pass...

I think it needs to be equal to the max satisfying...

Does that make sense or am I missing something?

FrozenPandaz (changes requested):
Much more readable after using semver :D

FrozenPandaz (approved):

PR #26277, opened 2024-05-30
fix(core): multi module handling and fall back matching for external nodes

JamesHenry:
## Current Behavior

- Packages such as `minimatch` which block access to their `package.json` AND have multi module format outputs are missed by the target project locator
- There is no fall back matching of external nodes from the graph by name

## Expected Behavior

- The target project locator appropriately finds modules such as `minimatch`
- There is fall back matching of external nodes from the graph by name

## Related Issue(s)

Fixes #


packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -302,6 +316,52 @@ export class TargetProjectLocator {
...
+   * In many cases the package.json will be directly resolvable, so we try that first.
+   * If, however, package exports are used and the package.json is not defined, we will
+   * need to resolve the main entry point of the package and traverse upwards to find the
+   * package.json.
+   *
+   * In some cases, such as when multiple module formats are published, the resolved package.json
+   * might only contain the "type" field - no "name" or "version", so in such cases we keep traversing
+   * until we find a package.json that contains the "name" and "version" fields.
+   */
+  private findExternalPackageJson(
```
FrozenPandaz, line 330: Should we call this.. `readPackageJson`? 

1. It doesn't just find the file, it also reads it.
2. nit: Do we really need to distinguish it's an external package.json? :shrug:

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -302,6 +316,52 @@ export class TargetProjectLocator {
...
+
+    try {
+      // Resolve the main entry point of the package
+      const mainPath = resolveRelativeToDir(packageName, relativeToDir);
+      let dir = dirname(mainPath);
+
+      while (dir !== parse(dir).root) {
+        const packageJsonPath = join(dir, 'package.json');
+        if (existsSync(packageJsonPath)) {
+          const parsedPackageJson = readJsonFile(packageJsonPath);
```
FrozenPandaz, line 351: It's better to try and fail which will only open up the fs once

```
try {
  const parsedPackageJson = readJsonFile(path);
  if (parsedPackageJson.name && parsedPackageJson.version) {
    return parsedPackageJson
  }
} catch {
  // Either it doesn't exist.. or.. it couldn't be parsed.. but the one from package managers should be able to be parsed so it's the same as it doesn't exist.
}
```

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -302,6 +316,52 @@ export class TargetProjectLocator {
...
+          // Ensure the package.json contains the "name" and "version" fields
+          if (parsedPackageJson.name && parsedPackageJson.version) {
+            return parsedPackageJson;
+          }
+        }
+        dir = dirname(dir);
+      }
+
+      throw new Error(`Could not find package.json for ${packageName}`);
```
FrozenPandaz, line 360: This throws an error just to be caught? :thinking: 

might as well return `null` here

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -163,23 +173,27 @@ export class TargetProjectLocator {
...
         packageName,
         fullDirPath
       );
       // The external package.json path might be not be resolvable, e.g. if a reference has been added to a project package.json, but the install command has not been run yet.
-      if (!externalPackageJsonPath) {
+      if (!externalPackageJson) {
+        // Try and fall back to resolving an external node from the graph by name
+        const externalNode = this.npmProjects[`npm:${packageName}`];
+        if (externalNode) {
+          this.npmResolutionCache.set(npmImportForProject, externalNode.name);
```
FrozenPandaz, line 185: I'm not sure, but I think in the original logic, this also set `undefined` if that's the case? so we don't have to continuously resolve to `undefined`?

FrozenPandaz (changes requested):

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
JamesHenry: Hmm, no, we want to keep traversing upwards through directories which don't have package.json files in them and this change would cause it to blow up on the first parent directory that didn't have one?

Real example from today inside `minimatch`:

```
minimatch/dist/cjs/package.json
```

Means we need to go through

```
minimatch/dist/cjs
minimatch/dist
```

before reaching `minimatch/package.json`

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
JamesHenry: Sure, yeah, this was existing

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
JamesHenry: Sure, I just quickly carried it over

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
JamesHenry: It set it to null, yeah, will add

FrozenPandaz (approved):

PR #26280, opened 2024-05-30
feat(bundling): add convert-to-inferred generator for @nx/rollup

jaysoo:
This PR adds `@nx/rollup:convert-to-inferred` generator.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/rollup/src/generators/convert-to-inferred/convert-to-inferred.ts
+++ b/packages/rollup/src/generators/convert-to-inferred/convert-to-inferred.ts
@@ -1,0 +1,129 @@
+import {
+  formatFiles,
+  getProjects,
+  readNxJson,
+  type Tree,
+  updateNxJson,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';
+
+import { extractRollupConfigFromExecutorOptions } from './lib/extract-rollup-config-from-executor-options';
+import { RollupExecutorOptions } from '../../executors/rollup/schema';
+
+interface Schema {
+  project?: string;
+  skipFormat?: boolean;
+}
+
+export async function convertToInferred(tree: Tree, options: Schema) {
+  let migrated = 0;
+
+  const projects = getProjects(tree);
+
+  forEachExecutorOptions<RollupExecutorOptions>(
+    tree,
+    '@nx/rollup:rollup',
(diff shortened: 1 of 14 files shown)


packages/rollup/src/generators/convert-to-inferred/convert-to-inferred.ts
```diff
@@ -0,0 +1,132 @@
...
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';
+
+import { extractRollupConfigFromExecutorOptions } from './lib/extract-rollup-config-from-executor-options';
+import { RollupExecutorOptions } from '../../executors/rollup/schema';
+
+interface Schema {
+  project?: string;
+  all?: boolean;
```
Coly010, line 14: We can remove this, it should be the default behaviour of the generator. Rather than forcing a user to potentially pass it. We can update the conditional below to only check if project specifically is called.

packages/rollup/src/generators/convert-to-inferred/convert-to-inferred.ts
```diff
@@ -0,0 +1,132 @@
...
+  let migrated = 0;
+
+  const projects = getProjects(tree);
+  const originalDefaultOptions = new Map<string, RollupExecutorOptions>();
+
+  forEachExecutorOptions<RollupExecutorOptions>(
+    tree,
+    '@nx/rollup:rollup',
+    (currentTargetOptions, projectName, targetName, configurationName) => {
+      if (!options.all && options.project && projectName !== options.project)
```
Coly010, line 28: As above, we can shorten this to `options.project && projectName !== options.project`.

A user specifying `--project` should be the only flag to determine if all projects are migrated, or if a specific project is migrated

packages/rollup/src/generators/convert-to-inferred/convert-to-inferred.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      const project = projects.get(projectName);
+      const target = project.targets[targetName];
+      console.log(
+        projectName,
+        targetName,
+        configurationName,
+        JSON.stringify(project, null, 2)
+      );
+
```
Coly010, line 39: We should remove this in the final version

Coly010 (approved):
One change, but it's minor

ndcunningham (approved):

PR #26281, opened 2024-05-30
feat(testing): cypress autoCancelAfterFailures

nsmith7989:
Pass autoCancelAfterFailures to cypress executor

Fixes #26264


packages/cypress/src/executors/cypress/schema.json
```diff
@@ -178,6 +178,17 @@
...
+    "autoCancelAfterFailures": {
+      "oneOf": [
+        {
+          "type": "number"
+        },
+        {
+          "type": "boolean"
+        }
+      ],
+      "description": "Specify the number of failures to cancel a run being recorded to the Cloud or `false` to disable auto-cancellation."
```
leosvelperez, line 191: Let's be more explicit in the description about which "Cloud" to avoid potential confusion with Nx Cloud. Also, the boolean value can only be `false`, so let's allow only that value.
```suggestion
        {
          "type": "boolean",
          "const": false
        }
      ],
      "description": "Specify the number of failures to cancel a run being recorded to the Cypress Cloud or `false` to disable auto-cancellation."
```

leosvelperez (changes requested):
Thanks for the contribution!

Please update the schema as suggested, and it should be good to go 💪🏻

leosvelperez (approved):

PR #26298, opened 2024-05-31
fix(bundling): load rollup config using the rollup version installed in the workspace

jaysoo:
This PR updates `@nx/rollup/plugin` so it loads the config file using `loadConfigFile` from the Rollup version installed in the workspace.

This fixes the issue when initializing in the [axios](https://github.com/axios/axios) repo, since the `require` call in their config is no longer valid ESM in Rollup 4.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/rollup/src/plugins/plugin.ts
+++ b/packages/rollup/src/plugins/plugin.ts
@@ -2,23 +2,19 @@ import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
 import { basename, dirname, join } from 'path';
 import { existsSync, readdirSync } from 'fs';
 import {
-  type TargetConfiguration,
   type CreateDependencies,
   type CreateNodes,
-  readJsonFile,
-  writeJsonFile,
-  detectPackageManager,
   CreateNodesContext,
+  detectPackageManager,
   joinPathFragments,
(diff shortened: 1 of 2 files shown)


packages/rollup/src/plugins/plugin.ts
```diff
@@ -91,6 +87,28 @@ async function buildRollupTarget(
...
+    // Using the workspace rollup ensures that the config file is compatible with the `loadConfigFile` function.
+    // e.g. rollup@2 supports having `require` calls in rollup config, but rollup@4 does not.
+    const m = require(require.resolve('rollup/loadConfigFile', {
+      paths: [dirname(configFilePath)],
+    }));
+    // Rollup 2 has this has default export, but it is named in 3 and 4.
+    // See: https://www.unpkg.com/browse/rollup@2.79.1/dist/loadConfigFile.js
+    loadConfigFile = typeof m === 'function' ? m : m.loadConfigFile;
+  } catch {
+    console.log('loadConfigFile 2');
```
FrozenPandaz, line 107: Remove

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #26333, opened 2024-06-03
feat(nx-dev): website-18 noindex

isaacplmann:
Add No_NEXT_PUBLIC_NO_INDEX environment variable to turn off indexing


ndcunningham (approved):

nx-dev/nx-dev/.env.local
```diff
@@ -0,0 +1,4 @@
+NEXT_PUBLIC_SEARCH_INDEX=nx-canary-production
+NEXT_PUBLIC_SEARCH_API_KEY=07340b85db83fd49c1f904f883cc3ef8
+NEXT_PUBLIC_SEARCH_APP_ID=PCTGM1JTQL
+NEXT_PUBLIC_NO_INDEX=false
```
FrozenPandaz, line 4: This should be set to `true` if we do NOT want to index this site right?

FrozenPandaz (changes requested):
Is there a setting for this in `vercel` but any chance?

PR #26356, opened 2024-06-04
fix(core): fall back to pacakgeManager if yarn --version failed

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/utils/package-manager.ts
+++ b/packages/nx/src/utils/package-manager.ts
@@ -7,7 +7,7 @@ import { dirSync } from 'tmp';
 import { promisify } from 'util';
 import { readNxJson } from '../config/configuration';
 import { readPackageJson } from '../project-graph/file-utils';
-import { readFileIfExisting, writeJsonFile } from './fileutils';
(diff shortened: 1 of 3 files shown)


packages/nx/src/utils/package-manager.ts
```diff
@@ -186,10 +186,35 @@ export function getPackageManagerVersion(
...
+          versionFormPackageJson
+        ) {
+          version = versionFormPackageJson;
+        }
+      }
+    }
+  }
+  if (!version) {
+    throw new Error(
+      `Cannot determine the version of the package manager ${packageManager}.`
```
FrozenPandaz, line 214:
```suggestion
      `Cannot determine the version of ${packageManager}.`
```

FrozenPandaz (changes requested):
Can we also fallback in `getPackageManagerCommands` and assume the `latest` version?

We don't have to like hardcode what version it would be.. but we can force the logic down that path.

packages/nx/src/utils/package-manager.ts
xiongemi: changed

FrozenPandaz (approved):

Issue #26023, opened 2024-05-24
NX_TASK_TARGET_PROJECT is not exposed in NX 19


vcardins:
### Current Behavior

Until version 18, I relied on the env var `NX_TASK_TARGET_PROJECT` to set my router basename, however on upgrading to nx 19 it is not available. 

### Expected Behavior

The `NX_TASK_TARGET_PROJECT` var should be available from `process.env`

### Steps to Reproduce

1. Migrate to nx 19
2. Start up a demo app
3. Type somewhere: `console.log(process.env.NX_TASK_TARGET_PROJECT)`
4. Nothing is logged

### Nx Report

```shell
Node   : 20.11.1
OS     : darwin-arm64
yarn   : 1.22.19

nx                 : 19.0.8
@nx/js             : 19.0.8
@nx/jest           : 19.0.8
@nx/linter         : 19.0.8
@nx/eslint         : 19.0.8
@nx/workspace      : 19.0.8
@nx/cypress        : 19.0.8
@nx/devkit         : 19.0.8
@nx/eslint-plugin  : 19.0.8
@nx/playwright     : 19.0.8
@nx/react          : 19.0.8
@nx/storybook      : 19.0.8
@nrwl/tao          : 19.0.8
@nx/vite           : 19.0.8
@nx/web            : 19.0.8
@nx/webpack        : 19.0.8
typescript         : 5.4.5
---------------------------------------
Registered Plugins:
@nx/storybook/plugin
```

### Package Manager Version

1.22.19

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

FrozenPandaz:
This is related to the security fix here: https://nx.dev/blog/2024-05-08-nx-19-release#breaking-change-updating-bundled-environment-variables-to

@jaysoo do you have any guidance for how to pass this information into the build?

vcardins:
@jaysoo Any idea?

myownsumm:
same thing for me. no variables passed into `process.env` from `.env` file:
* NX updated to 19.1.0
* react 18.3.1
* @nx/vite 19.1.0

vcardins:
@myownsumm Are you suggesting the the env var `NX_PUBLIC_TASK_TARGET_PROJECT` is available during build time? Could you confirm that?

NicolasBelliard:
Hello!
Is there any news about this subject?

I made some test on this test project: https://github.com/NicolasBelliard/nx-version-19

[...]

PR #26357, opened 2024-06-04
docs(core): adding warning following the breaking change in the environment variable documentation

NicolasBelliard:
Following the breaking change with the process environment variable are no available.

I see that this [issue](https://github.com/nrwl/nx/issues/26023) has been created.

Therefore, I think a warning in the documentation is needed to inform that this variables are not accessible for Nx version greater or equal to 19.

## Current Behavior
On the current Nx documentation, there is no information regarding this breaking change:

image

Doc link: https://nx.dev/reference/environment-variables#environment-variables

## Expected Behavior
Adding a warning to inform about this breaking change:

image

## Related Issue(s)
[](https://github.com/nrwl/nx/issues/26023)

Let me know if it is ok for you and if you want to rephrase the following warning.

Nicolas Belliard


jaysoo:
Hi @NicolasBelliard , this warning isn't correct as the variables are set and available in Node. They just aren't bundled into browser code anymore.

I would update this page: https://nx.dev/recipes/react/use-environment-variables-in-react#for-nonvite-react-applications

jaysoo (approved):

jaysoo:
Merged, thanks!

Issue #21855, opened 2024-02-16
`nx release` publishes TS files in combination with default (integrated) generators


lorenzodejong:
### Current Behavior

Currently setting up a default (integrated) monorepo preset yields incorrect behavior when using `nx release`. By default, packages generated from `@nx/js` with the `publishable` option will generate a `project.json` which outputs changes to the /dist folder on the root of the monorepo. However the `package.json` in the src folder points to `./src/package.json`.

Running the `nx release` command seems to be directly publishing the package from the source directory, causing the actual TypeScript source files to be published instead of the resolved build output in the `dist` folder.

It seems like the current implementation of `nx release` is only intended to be used in package-based repositories, where you specify your package.json to point to a `dist` folder present in the actual source folder.

The [documentation](https://nx.dev/recipes/nx-release/get-started-with-nx-release#run-without) also seems to be incorrect on this regard. If you look at the example of the published tarball you'll see that actual `.ts` files are getting published:

```
// ...
📦  @myorg/pkg-1@0.0.2
=== Tarball Contents ===

233B README.md
277B package.json
53B  src/index.ts
61B  src/lib/pkg-1.ts
=== Tarball Details ===
// ...
```

### Expected Behavior

I expect `nx release` to be directly compatible with projects generated from the default provided Nx generators, without having to change the behavior of the `package.json` and `project.json`.

### Steps to Reproduce

1. Create a new repository with `npx create-nx-workspace release-example-workspace`, select to create an `Integrated Monorepo` when asked for it.
2. Generate a library with `npx nx g @nx/js:lib mylib`
3. Run `npx nx release --first-release --dry-run`
4. Run `npx nx release publish --dry-run`

Observed behavior: `.ts` files are present in the tarball.

### Nx Report

```shell
Node   : 18.13.0
   OS     : darwin-arm64
   npm    : 8.19.3

   nx                 : 18.0.4
   @nx/js             : 18.0.4
   @nx/linter         : 18.0.4
   @nx/eslint         : 18.0.4
   @nx/workspace      : 18.0.4
   @nx/devkit         : 18.0.4
   @nx/eslint-plugin  : 18.0.4
   @nrwl/tao          : 18.0.4
   typescript         : 5.3.3
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

cpierceworld:
I ran into the same issue with an Angular project with some "publishable" libraries. #21560

[...]

PR #26370, opened 2024-06-04
docs(release): publish from dist directory recipe

fahslaj:
Add a docs recipe for the two supported methods for tracking dependencies - using the source package.json + explicit versions, or using the dist package.json with file or workspace references

#21855


docs/shared/recipes/nx-release/publish-custom-dist-directory.md
```diff
@@ -0,0 +1,79 @@
+# Publish a Custom Dist Directory
+
+This recipe guides you through configuring Nx Release to version and publish packages that are built to a custom `dist` directory instead their source project root. This is often the case when using Nx generators or custom builders that output to a root `dist` directory.
```
JamesHenry, line 3:
```suggestion
This recipe guides you through configuring Nx Release to version and publish packages that are built to a custom `dist` directory instead of their source project root. This is often the case when using Nx generators or custom builders that output to a centralized `dist` directory at the workspace root.
```

docs/shared/recipes/nx-release/publish-custom-dist-directory.md
```diff
@@ -0,0 +1,79 @@
...
+  "dependencies": {
+    "my-other-package": "^0.1.1",
+    "my-other-package-2": "1.2.0"
+  }
+}
+```
+
+Nx Release will read the current version (0.1.1) from the source package.json file, update it to the new version, and update all dependencies in this package.json file. Changes to the package.json file will be staged and committed unless git operations are disabled. In this case, the `packageRoot` is the same as the Nx project root.
+
+### Use File References and Do Not Track Versions in Source Control
```
JamesHenry, line 30: ...or workspace (not just `file:`)

docs/shared/recipes/nx-release/publish-custom-dist-directory.md
```diff
@@ -0,0 +1,79 @@
...
+  "name": "my-package",
+  "version": "0.1.1",
+  "dependencies": {
+    "my-other-package": "0.1.1",
+    "my-other-package-2": "1.2.0"
+  }
+}
+```
+
+Note that the version number is not present in the source package.json file. This is because the source package.json file is never updated in the versioning process; its data is instead written to the dist package.json file. To ensure the correct version number is used in the dist package.json file, `"release.version.generatorOptions.currentVersionResolver": "git-tag"` should be used.
```
JamesHenry, line 57: `registry` could also be used, that's what we do on the nx repo itself

docs/shared/recipes/nx-release/publish-custom-dist-directory.md
```diff
@@ -0,0 +1,79 @@
...
+
+To configure this behavior, add the following configuration to the `nx.json` file, or the `project.json` file of relevant projects:
+
+```json {% fileName="nx.json" %}
```
JamesHenry, line 61:
```suggestion
```jsonc {% fileName="nx.json" %}
```

JamesHenry: ^ that's what I would use if only thinking about the markdown. Worth checking that jsonc is supported/works fine on the website

JamesHenry (changes requested):

docs/shared/recipes/nx-release/publish-custom-dist-directory.md
```diff
@@ -0,0 +1,79 @@
+# Publish a Custom Dist Directory
+
+This recipe guides you through configuring Nx Release to version and publish packages that are built to a custom `dist` directory instead of their source project root. This is often the case when using Nx generators or custom builders that output to a centralized `dist` directory at the workspace root.
+
+## The Package Root
+
+Nx Release has the concept of a "package root", which is different than the project root. The package root is the directory from which the package is versioned and published. By default, the package root is the project root detected by Nx. Due to the modular nature of Nx Release, the package root is configured independently for the version and publish steps.
```
JamesHenry, line 7:
```suggestion
Nx Release has the concept of a "package root", which is different than the project root. The package root is the directory from which the package is versioned and published. By default, the package root is the project root detected by Nx. Due to the modular nature of Nx Release, the package root can be configured independently for the version and publish steps.
```

JamesHenry (approved):

PR #26375, opened 2024-06-05
feat(nx-cloud): automatically open cloud URL in browser

mandarini:
Marking as draft because we're not sure if we want this functionality or not.

Automatically open the browser with the cloud onboarding URL.


mandarini:
Closing since we will not implement

PR #26420, opened 2024-06-06
fix(bundling): set project type correct for buildable vite projects

jaysoo:
This PR:

1. fixes `@nx/vite/plugin` so that it infers the correct `projectType`
2. changes the default `package-json-workspaces` plugin such that `projectType` is not inferred is `appsDir` and `libsDir` are not set
3. Update PDV to look for the normalized `type` property on projects if `projectType` is missing (due to no longer being inferred by any plugin).

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/plugins/package-json-workspaces/create-nodes.spec.ts
+++ b/packages/nx/src/plugins/package-json-workspaces/create-nodes.spec.ts
@@ -59,7 +59,6 @@ describe('nx package.json workspaces plugin', () => {
               },
             },
             "name": "root",
-            "projectType": "library",
             "root": ".",
             "sourceRoot": ".",
             "targets": {
@@ -98,7 +97,6 @@ describe('nx package.json workspaces plugin', () => {
               },
             },
             "name": "lib-a",
-            "projectType": "library",
             "root": "packages/lib-a",
             "sourceRoot": "packages/lib-a",
             "targets": {
@@ -145,7 +143,6 @@ describe('nx package.json workspaces plugin', () => {
               },
             },
             "name": "lib-b",
-            "projectType": "library",
             "root": "packages/lib-b",
             "sourceRoot": "packages/lib-b",
             "targets": {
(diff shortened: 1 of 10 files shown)


Coly010 (approved):

packages/nx/src/plugins/package-json-workspaces/create-nodes.spec.ts
```diff
@@ -175,4 +172,113 @@ describe('nx package.json workspaces plugin', () => {
...
+      '/root'
+    );
+
+    expect(
+      createNodeFromPackageJson('packages/mylib/package.json', '/root')
+        .projects['packages/mylib'].projectType
+    ).toEqual('library');
+  });
+
+  it('should infer library project type if only libsDir is set', () => {
```
MaxKless, line 254: This test is named exactly as the one before, making it a bit confusing

packages/nx/src/plugins/package-json-workspaces/create-nodes.ts
```diff
@@ -88,22 +88,33 @@ export function buildProjectConfigurationFromPackageJson(
...
+      nxJson?.workspaceLayout?.appsDir != nxJson?.workspaceLayout?.libsDir &&
+      directory.startsWith(nxJson.workspaceLayout.appsDir)
+        ? 'application'
+        : 'library';
+  } else if (
+    typeof nxJson?.workspaceLayout?.libsDir !== 'undefined' &&
+    directory.startsWith(nxJson.workspaceLayout.libsDir)
+  ) {
+    projectConfiguration.projectType = 'library';
+  }
```
MaxKless, line 115: How come there's no third case that handles only `appsDir` being set?

MaxKless (approved):
left two small comments but looks good

packages/nx/src/plugins/package-json-workspaces/create-nodes.ts
jaysoo: In that case, we just don't set `projectType`, which was the previous logic. It's either both `appsDir` and `libsDir`, or only `libsDir`. The rest will fall back to `project.type` rather than `project.data.projectType`.

Coly010 (approved):

FrozenPandaz (approved):
Sorry, there might be a big merge conflict. Should be good once you rebase.

PR #26458, opened 2024-06-07
chore(react): Update description for --minimal to be clearer

ndcunningham:


isaacplmann (approved):

jaysoo:
This option also applies to Jest, where the extra spec file should not be present for lib and apps.

jaysoo (changes requested):
I think the generators need to be fixed instead.

PR #26465, opened 2024-06-07
feat(nextjs): add experimental-build-mode option to support compile only

dfr-exnaton:
## Current Behavior

The nextjs build executor does not support all build flags such as `--experimental-build-mode` (see https://nextjs.org/docs/app/api-reference/next-cli#build). 

## Expected Behavior

For certain deployment models (e.g. mult-environment builds), it is helpful to only compile but not generate pages at build time. See https://github.com/vercel/next.js/discussions/46544 for a discussion.

### DIFF
--- a/docs/generated/packages/next/executors/build.json
+++ b/docs/generated/packages/next/executors/build.json
@@ -72,6 +72,11 @@
       "experimentalAppOnly": {
         "type": "boolean",
         "description": "Only build 'app' routes"
+      },
+      "experimentalBuildMode": {
+        "type": "string",
+        "description": "Change the build mode.",
+        "enum": ["compile", "generate"]
       }
     },
     "required": ["outputPath"],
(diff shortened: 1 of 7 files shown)


cyrus-za:
Thanks for opening this PR. This has been high on my wishlist for quite some time

docs/generated/packages/next/executors/build.json
```diff
@@ -72,6 +72,11 @@
       "experimentalAppOnly": {
         "type": "boolean",
         "description": "Only build 'app' routes"
+      },
+      "experimentalBuildMode": {
+        "type": "string",
+        "description": "Change the build mode.",
+        "enum": ["compile", "generate"]
```
dfr-exnaton, line 79: There is some confusion w.r.t. to the "default" option. The documentation shows it as available, if you pass it the CLI will complain though.

dfr-exnaton:
@ndcunningham could you have a quick look? This is a super small and non-intrusive change making lives for users of nextjs  in a 12-factor setting much easier... thanks a lot!

ndcunningham (approved):
LGTM thanks!

PR #26484, opened 2024-06-10
feat(vite): add createNodesV2 function

Coly010:
- chore(vite): move single file processing of plugin to own function
- feat(vite): add CreateNodesV2 function

## Current Behavior

There is no implementation for createNodesV2

## Expected Behavior

There should be an implementation for createNodesV2

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -2,6 +2,8 @@ import {
   CreateDependencies,
   CreateNodes,
   CreateNodesContext,
+  createNodesFromFiles,
+  CreateNodesV2,
   detectPackageManager,
   joinPathFragments,
   logger,
@@ -17,6 +19,7 @@ import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash
 import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
 import { getLockFileName } from '@nx/js';
 import { loadViteDynamicImport } from '../utils/executor-utils';
+import { hashObject } from 'nx/src/hasher/file-hasher';
 
 export interface VitePluginOptions {
   buildTargetName?: string;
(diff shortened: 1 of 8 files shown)


packages/vite/src/plugins/plugin.ts
```diff
@@ -23,76 +28,123 @@ export interface VitePluginOptions {
   previewTargetName?: string;
   serveStaticTargetName?: string;
 }
+type ViteTargets = Pick<ProjectConfiguration, 'targets' | 'metadata'>;
 
 const cachePath = join(workspaceDataDirectory, 'vite.hash');
 const targetsCache = readTargetsCache();
```
FrozenPandaz, line 34: Don't read this in the moduleScope, there's no point.

packages/vite/src/plugins/plugin.ts
```diff
@@ -23,76 +28,123 @@ export interface VitePluginOptions {
...
+    const oldCache = readTargetsCache();
+    writeJsonFile(cachePath, {
+      ...oldCache,
+      ...targetsCache,
+    });
+  }
 }
 
 export const createDependencies: CreateDependencies = () => {
   writeTargetsToCache();
```
FrozenPandaz, line 58: This is not needed anymore.

packages/vite/src/plugins/plugin.ts
```diff
@@ -23,76 +28,123 @@ export interface VitePluginOptions {
...
-    if (
-      !siblingFiles.includes('package.json') &&
-      !siblingFiles.includes('project.json')
-    ) {
-      return {};
-    }
+    logger.warn(
+      '`createNodes` is deprecated. Update your plugin to utilize createNodesV2 instead. In Nx 20, this will change to the createNodesV2 API.'
+    );
+    return createNodesInternal(configFilePath, options, context, targetsCache);
```
FrozenPandaz, line 90: It's okay to use an `{}` here for `targetsCache`.

This function shouldn't really get called unless it's a rare case that someone wrapped our plugin `createNodes` function.

FrozenPandaz (changes requested):

jaysoo (approved):

packages/vite/src/plugins/plugin.ts
```diff
@@ -23,76 +28,109 @@ export interface VitePluginOptions {
...
-    ...targetsCache,
-  });
+function writeTargetsToCache(cachePath, results?: Record<string, ViteTargets>) {
+  writeJsonFile(cachePath, results);
 }
 
 export const createDependencies: CreateDependencies = () => {
-  writeTargetsToCache();
+  logger.warn(
+    "The 'createDependencies' function is now a no-op. It will be replaced via the 'CreateNodesV2' API."
```
FrozenPandaz, line 43:
```suggestion
    "The 'createDependencies' function is now a no-op. This functionality is included in 'createNodesV2'."
```

This has to be moved to a jsDoc.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #26491, opened 2024-06-10
fix(js): filter project dependencies when calculating topological ordering

### LINKED ISSUE #26490: Js release-version: topological ordering iterates all edges per node instead of only its own
### Current Behavior

Function [sortProjectsTopologically](https://github.com/nrwl/nx/blob/c7401b56576d0f09324d0886af23102523fdd64e/packages/js/src/generators/release-version/utils/sort-projects-topologically.ts#L42) processes all edges for each project in the `processQueue`. This means that a project `A` will process many edges that have nothing to do with it, creating a wrong topological ordering.

For instance, in a simple repo with dependencies:
```
C --> B --> A
```

the topological ordering algorithm will:
* pick `A`
* go through the dependency `C --> B` and add `C` to the `processQueue`
* go through the dependency `B --> A` and add `B` to the `processQueue`
* pick `C`
* pick `B`

resulting in order `[A, C, B]` instead of the desired `[A, B, C]`

### Expected Behavior

As the comment says in the code, when processing nodes in the project graph, we should 
```Process each project that depends on the current node```. 

That means taking only the edges whose target is the node which we just picked from the `processQueue`.

### GitHub Repo

https://github.com/mpsanchis/nx-basic-repo

### Steps to Reproduce

1. Clone repo
2. Go to branch `all-independent`
3. Tag all projects in git (you can do it with `node tools/add-tags.js HEAD~1 100.0.0`) for instance
4. run `nx release version --dry-run`
5. Verify that `app1` is not bumped despite of depending on `lib1`, because it has been wrongly sorted in the topological ordering

### Nx Report

```shell
Node   : 18.20.2
OS     : darwin-arm64
pnpm   : 8.8.0

nx                 : 19.2.2
@nx/js             : 19.2.2
@nx/jest           : 19.2.2
@nx/linter         : 19.2.2
@nx/eslint         : 19.2.2
@nx/workspace      : 19.2.2
@nx/devkit         : 19.2.2
@nx/eslint-plugin  : 19.2.2
@nx/playwright     : 19.2.2
@nx/react          : 19.2.2
@nrwl/tao          : 19.2.2
@nx/web            : 19.2.2
@nx/webpack        : 19.2.2
typescript         : 5.4.5
Registered Plugins:
@nx/webpack/plugin
@nx/eslint/plugin
@nx/playwright/plugin
@nx/jest/plugin
```

### Failure Logs

_No response_

### Package Manager Version

pnpm 8.10.0

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Adding this issue because I saw it was necessary when opening a PR (in the contributing guide).
Please let me know if there's something else to do. It's my first PR 😄

mpsanchis:
Topological ordering of projects in the graph, used when calculating versions with `@nx/js:release-version`, has a bug. Projects are not sorted topologically correctly, leading to unexpected versioning behaviour.

Fixes #26490

### DIFF
--- a/packages/js/src/generators/release-version/utils/sort-projects-topologically.spec.ts
+++ b/packages/js/src/generators/release-version/utils/sort-projects-topologically.spec.ts
@@ -1,114 +1,243 @@
 import { sortProjectsTopologically } from './sort-projects-topologically';
+import { DependencyType, StaticDependency } from '@nx/devkit';
 
 describe('sortProjectsTopologically', () => {
-  it('should return empty array if no projects are provided', () => {
-    const projectGraph = {
-      dependencies: {},
-      nodes: {},
-    };
-    const projectNodes = [];
-    const result = sortProjectsTopologically(projectGraph, projectNodes);
-    expect(result).toEqual([]);
+  describe('edge cases', () => {
+    it('should return empty array if no projects are provided', () => {
+      const projectGraph = {
+        dependencies: {},
+        nodes: {},
+      };
+      const projectNodes = [];
+      const result = sortProjectsTopologically(projectGraph, projectNodes);
+      expect(result).toEqual([]);
+    });
+
+    it('should return a single project if only one project is provided', () => {
+      const projectGraph = {
+        dependencies: {},
+        nodes: {
+          project1: {
+            name: 'project1',
+            data: {
+              root: '',
+            },
+            type: 'app' as const,
+          },
+        },
+      };
+      const projectNodes = [projectGraph.nodes.project1];
+      const result = sortProjectsTopologically(projectGraph, projectNodes);
+      expect(result).toEqual([projectGraph.nodes.project1]);
+    });
+
+    it('should return the original list of nodes if a circular dependency is present', () => {
+      const projectGraph = {
+        dependencies: {
+          project1: [
+            {
+              source: 'project1',
+              target: 'project2',
+              type: 'static',
+            },
+          ],
+          project2: [
+            {
+              source: 'project2',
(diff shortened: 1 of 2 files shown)


JamesHenry (changes requested):
Thank you @mpsanchis, please can you add a unit test in `packages/js/src/generators/release-version/utils/sort-projects-topologically.spec.ts` that demonstrates the behavior you are addressing here (i.e. it should fail before your change, and pass afterwards)

mpsanchis:
Hi @JamesHenry 
Thanks for the quick reply. I have added two "interesting" tests. One of which (`should return [C,B,A] if A depends on B and B depends on C`) failed before the PR. I have also grouped the test cases.
Please let me know if you guys have other ways of doing things 😄

JamesHenry:
Thank you, please remove the indirection/abstraction from the tests. It will be a lot more lines of code, but a lot simpler to read which is the priority with tests

JamesHenry:
@mpsanchis Thank you this LGTM, I pushed a commit which greatly reduces the churn in test file and keeps things inline. I always prefer to keep additions of tests covering new behaviour separate from refactoring of existing tests where possible https://github.com/nrwl/nx/pull/26491/files#diff-dd7e8070cc17859170fab8e1ff1d38ce36fa3e1bf5cc0ec7928b7d7d0a117480

JamesHenry (approved):

JamesHenry:
Thanks again @mpsanchis!

PR #26494, opened 2024-06-10
fix(release): false positive npm dist-tag add

### LINKED ISSUE #26240: nx release publish launch 'npm dist-tag add' command on a package version which does not exists on the npm registry
### Current Behavior

 During nx release publish execution and when there is only one version of the specified package on the remote registry, `npm dist-tag add ${packageName}@${packageJson.version} ${tag} --"${registryConfigKey}=${registry}"` is executed against a package version which does not exist in the registry.

In our case, it leads to an "unhandled" 400 error when it's executed with dist-tag "latest" on Sonatype Nexus Repository OSS 3.55.0-01.

### Expected Behavior

When the current package version to publish is not found on the remote registry, nx release publish should not try to execute the command `npm dist-tag add ${packageName}@${packageJson.version} ${tag} --"${registryConfigKey}=${registry}"` and error out.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create a new package in an nx workspace and update package.json version of the package to "0.1.0-beta.0"
2. Execute nx release publish --tag=latest
3. Update the created package package.json version to "0.1.0"
4. Execute nx release publish --tag=latest

### Nx Report

```shell
Node   : 20.9.0
OS     : win32-x64
pnpm   : 8.15.1

nx                 : 18.3.5
@nx/js             : 18.3.5
@nx/jest           : 18.3.5
@nx/linter         : 18.3.5
@nx/eslint         : 18.3.5
@nx/workspace      : 18.3.5
@nx/angular        : 18.3.5
@nx/cypress        : 18.3.5
@nx/devkit         : 18.3.5
@nx/eslint-plugin  : 18.3.5
@nrwl/tao          : 18.3.5
@nx/web            : 18.3.5
@nx/webpack        : 18.3.5
typescript         : 5.4.5
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

The issue comes from the return type of the npm view command which is not consistent for the parsed "versions" key.

"versions" key return an array of string when there is more than one version published to the remote registry and a simple string while there is still only one version.

Hence if the versions key contains for example "0.1.0-beta.0" and i'am trying to publish "0.1.0", this part of the code:
https://github.com/nrwl/nx/blob/7088b09f472bb767d386df73ca0534e09386d7db/packages/js/src/executors/release-publish/release-publish.impl.ts#L113
 return true as it's a substring match.

We fixed it by preemptively adding a type check and reassigning the string value into an array before the check is executed.

I'am willing to submit a PR if necessary.

JamesHenry:
## Current Behavior

`npm dist-tag add` can be run when not desired if exactly one version of a package exists in a registry (because in this case the data type of the `versions` field is a string instead of an array of strings (WAT)).

## Expected Behavior

`npm dist-tag add` is only run when appropriate

## Related Issue(s)

Fixes #26240

### DIFF
--- a/packages/js/src/executors/release-publish/release-publish.impl.ts
+++ b/packages/js/src/executors/release-publish/release-publish.impl.ts
@@ -110,7 +110,12 @@ export default async function runExecutor(
         };
       }
 
-      if (resultJson.versions.includes(currentVersion)) {
+      // If only one version of a package exists in the registry, versions will be a string instead of an array.
+      const versions = Array.isArray(resultJson.versions)
+        ? resultJson.versions
+        : [resultJson.versions];
+
+      if (versions.includes(currentVersion)) {
         try {
           if (!isDryRun) {
             execSync(npmDistTagAddCommandSegments.join(' '), {


jaysoo:
Approved of the PR here: https://github.com/nrwl/nx/pull/26496

PR #26507, opened 2024-06-11
fix(core): convert-to-monorepo support for root project without project.json

jaysoo:
This PR fixes `convert-to-monorepo` generator, so it works after `nx init` on a single project.

## Current Behavior

If you use :

```
npm create vite react-app -- --template=react-ts
cd react-app
npx nx init
```

Then, you'll get an error when running `nx g convert-to-monorepo`.

```
 NX   The following projects are defined in multiple locations:

- react-app:
  - .
  - packages/react-app

To fix this, set a unique name for each project in a project.json inside the project's root. If the project does not currently have a project.json, you can create one that contains only a name.

MultipleProjectsWithSameNameError: The following projects are defined in multiple locations:
- react-app:
  - .
  - packages/react-app
  - 
```

## Expected Behavior
The generator should work, and you can run commands successfully afterwards (e.g. `nx serve react-app`).

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.from-nx-init.spec.ts
+++ b/packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.from-nx-init.spec.ts
@@ -1,0 +1,34 @@
+import 'nx/src/internal-testing-utils/mock-project-graph';
+
+import {
+  readJson,
+  readProjectConfiguration,
+  Tree,
+  updateJson,
+} from '@nx/devkit';
+import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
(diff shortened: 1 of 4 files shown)


packages/workspace/src/generators/convert-to-monorepo/convert-to-monorepo.from-nx-init.spec.ts
```diff
@@ -0,0 +1,34 @@
+import 'nx/src/internal-testing-utils/mock-project-graph';
```
jaysoo, line 1: Had to extract this to another file, because it was causing `preset.spec.ts` to fail on agents for some reason, without any useful failures/errors printed. It could not be reproduced locally.

FrozenPandaz (approved):

PR #26511, opened 2024-06-11
fix(testing): add the preset from the jest config to the inferred task inputs

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/plugins/plugin.ts
+++ b/packages/jest/src/plugins/plugin.ts
@@ -12,14 +12,18 @@ import {
   TargetConfiguration,
   writeJsonFile,
 } from '@nx/devkit';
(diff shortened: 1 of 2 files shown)


packages/jest/src/plugins/plugin.ts
```diff
@@ -157,12 +161,15 @@ async function buildJestTargets(
...
   }
 
+  const rawConfig = await loadConfigFile(absConfigFilePath);
   const config = await readConfig(
     {
       _: [],
       $0: undefined,
     },
-    absConfigFilePath
+    rawConfig,
```
FrozenPandaz, line 170: Does this skip the loading from `jest`? :eyes:

FrozenPandaz (approved):

PR #26516, opened 2024-06-11
feat(webpack): add convertConfigToWebpackPlugin

ndcunningham:
This PR introduces functionality for users who currently use the `withNx` and `withReact` plugins in their webpack configuration to migrate to the `NxAppWebpackPlugin`.

The `nxUseLegacyPlugin` wraps the legacy style function so that it continues to work with the standardized generated webpack config.

By implementing this change, the aim is to provide a consistent method for users opting to transition to inferred targets. This ensures a smoother migration process, offering better integration and reducing potential configuration complexities.

### DIFF
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/normalize-options.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/normalize-options.ts
@@ -76,7 +76,7 @@ export function normalizeOptions(
 
   const sourceRoot = projectNode.data.sourceRoot ?? projectNode.data.root;
 
-  if (!options.main) {
+  if (!combinedPluginAndMaybeExecutorOptions.main) {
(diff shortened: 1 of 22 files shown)


packages/webpack/src/plugins/nx-webpack-plugin/lib/normalize-options.ts
```diff
@@ -76,7 +76,7 @@ export function normalizeOptions(
 
   const sourceRoot = projectNode.data.sourceRoot ?? projectNode.data.root;
 
-  if (!options.main) {
+  if (!combinedPluginAndMaybeExecutorOptions.main) {
```
ndcunningham, line 79: This should be checking the combined options (executor and config) and not only what has been set inside the config

jaysoo (approved):

PR #26523, opened 2024-06-12
feat(devkit): add aggregateLog util for executor to plugin migration

Coly010:
## Current Behavior

There is no way to aggregate logs and output at the end of execution for the executor to plugin migrations

## Expected Behavior

There should be a way to aggregate logs and output them at the end of execution

Usage:
```ts
// Call with the executorName that is being migrated
// Make sure the log itself does not contain dynamic values
// Add the project root that is being migrated

aggregateLog({executorName: '@nx/vite:build', project: projectRoot, log: 'Encountered X in project.json. Do Y to handle this manually.'})

// Then in GeneratorCallback run a task
return () => { flushLogs() };
```

Example Output
image

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/generators/plugin-migrations/aggregate-log-util.ts
+++ b/packages/devkit/src/generators/plugin-migrations/aggregate-log-util.ts
@@ -1,0 +1,66 @@
+import { output, logger } from 'nx/src/devkit-exports';
+
+interface AggregateLogOptions {
+  project: string;
+  log: string;
+  executorName: string;
+}
+
+interface AggregateLogItem {
+  log: string;
+  projects: Set<string>;
+}
+
+/**
+ * @example
+ * // Instantiate a new object
+ * const migrationLogs = new AggregatedLog();
+ *
+ * // Add logs
+ * migrationLogs.addLog({executorName: '@nx/vite:build', project: 'app1', log: 'Migrate X manually'});
+ *
+ * // Flush all logs
+ * migrationLogs.flushLogs()
+ */
+export class AggregatedLog {
+  logs: Map<string, Map<string, AggregateLogItem>> = new Map();
+
+  addLog({ project, log, executorName }: AggregateLogOptions): void {
+    if (!this.logs.has(executorName)) {
+      this.logs.set(executorName, new Map());
+    }
+
+    const executorLogs = this.logs.get(executorName);
+    if (!executorLogs.has(log)) {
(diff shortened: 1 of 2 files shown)


packages/devkit/src/generators/plugin-migrations/aggregate-log-util.ts
```diff
@@ -0,0 +1,60 @@
...
+  executorName: string;
+}
+
+interface AggregateLogItem {
+  log: string;
+  projects: Set<string>;
+}
+
+class AggregatedLog {
+  static logs: Map<string, Map<string, AggregateLogItem>> = new Map();
```
FrozenPandaz, line 16: I'm a bit confused why any of these are static :eyes: 

This makes it hard to separate out different aggregate logs into different instances. Can this all be not static?

Even if that is not what we intend to do right now, I don't really see a point to making that harder. We could instantiate one in the module scope of the migrate utilities which would do the same thing right?

Maybe it would make more sense if I saw a usage of it.

packages/devkit/src/generators/plugin-migrations/aggregate-log-util.ts
```diff
@@ -0,0 +1,60 @@
...
+  log: string;
+  executorName: string;
+}
+
+interface AggregateLogItem {
+  log: string;
+  projects: Set<string>;
+}
+
+class AggregatedLog {
```
FrozenPandaz, line 15: This is going to be used in the migrations? Could you add 1 usage of it?

FrozenPandaz (changes requested):

packages/devkit/src/generators/plugin-migrations/aggregate-log-util.ts
Coly010: It's intentionally static. 

Some `convert-to-inferred` generators will run migration util multiple times for different executors. Static allows us to collect the data and store it in one place. Then we can flush it all at the end of the generator.

If we added a new instance per migration util, we would have collect all that data again for flushing at the end of a single generator.

When we run multiple `convert-to-inferred` generators at once, having multiple instances would compound this. 
Static means that even when running multiple `convert-to-inferred` generators, we have a single state for all the generators, all the executors that are being migrated, and can flush it all, one time, at the end, after all generators have completed.

FrozenPandaz (approved):

PR #26526, opened 2024-06-12
feat(angular): add helper function to devkit to add viewProviders to a component

Olster:
## Current Behavior

Angular Nx utils have `addProviderToComponent` function to add a provider to an Angular component, but are missing a function to add viewProviders.

## Expected Behavior

There should be a function like `addProviderToComponent` to add view providers to a component.


leosvelperez (approved):
Thanks for the contribution!

PR #26580, opened 2024-06-17
fix(graph): fix search params reload when back to graph

xiongemi:
## Current Behavior

- change useIntervalWhen
  - when this function is used, it is mostly calling a fetch request, change it to take an async function
  - wait for the current fetch request to finish before firing the next one
  - change the name to usePoll
- currently, when going back to graph, it needs to remove the search params expand and sometimes need to add ?project= search params
  - delete expand search params is called after adding the project search params
  
```

 setSearchParams(
        (currentSearchParams) => {
          currentSearchParams.delete('expanded');
          return currentSearchParams;
        },
        { replace: true, preventScrollReset: true }
      );

```

this function does not contain the ?project= search param key, caused the graph page to reload twice
- change the function useRouteConstructor to delete the search params `expanded` there

## Related Issue(s)

Fixes #


graph/shared/src/lib/use-interval-when.ts
```diff
@@ -1,11 +1,11 @@
 import { useEffect, useRef } from 'react';
 
 export const useIntervalWhen = (
```
FrozenPandaz, line 3: `usePoll`

FrozenPandaz (changes requested):

jaysoo (approved):

FrozenPandaz (approved):

PR #26601, opened 2024-06-19
feat(remix): add convert-to-inferred generator

Coly010:
- feat(remix): add convert-to-inferred migration
- feat(remix): add serve executor logic

## Related Issue(s)

Fixes #


packages/remix/src/generators/convert-to-inferred/lib/build-post-target-transformer.ts
```diff
@@ -0,0 +1,236 @@
...
+    migrationLogs.addLog({
+      project: projectName,
+      executorName: '@nx/remix:build',
+      log: "Unable to migrate `generatePackageJson` to Remix Config. Use the `@nx/dependency-checks` ESLint rule to update your project's package.json.",
+    });
+
+    delete options.generatePackageJson;
+  }
+
+  if ('generateLockfile' in options) {
```
jaysoo, line 129: The lint rule won't help here since it only updates the source `package.json` file. We should just say this isn't currently supported.

Coly010: Ok no problem

jaysoo (approved):

PR #26608, opened 2024-06-20
fix(nx-dev): allow linking to headers that are `code` wrapped

barbados-clemens:
before if a header that was using `code` in the title (i.e. launch
template) the header should so the link icon but would not link anywhere
because the rendered id tag would be an empty string

empty id tag for headers

after the id tag is correctly linked by checking the rendering children
contains a `code` tag and pulls the code children out.

added benefit includes the code headers being linked in the side nav
correct too

side by side diff with changes

Example showing working from preview: https://nx-dev-git-docs-allow-linking-code-headers-nrwl.vercel.app/ci/reference/launch-templates#launchtemplatestemplatenameinitsteps


isaacplmann (approved):
Thanks!

PR #26615, opened 2024-06-20
docs(core): remove tasksRunnerOptions

isaacplmann:
Remove the tasksRunnerOptions section in the nx.json docs


jaysoo (approved):

dancer1325:
Hi @isaacplmann , is there some reason to remove it? Maybe some idea to remove from the nx-json schema?

isaacplmann:
We'll eventually have a different solution for people and we're removing these docs to try to limit the number of people that will have to migrate

PR #26622, opened 2024-06-21
feat(graph): add atomizer label to target groups 

MaxKless:
## Current Behavior
Atomized Groups are treated just like any other groups in the PDV

## Expected Behavior
We want to let people know that something was created by the Atomizer and also surface more information to users.

### DIFF
--- a/graph/ui-tooltips/src/lib/property-info-tooltip.tsx
+++ b/graph/ui-tooltips/src/lib/property-info-tooltip.tsx
@@ -10,9 +10,7 @@ type PropertyInfoTooltipType =
   | 'dependsOn'
   | 'options'
   | 'configurations'
-  | 'release'
-  | 'atomizer-cloud'
-  | 'atomizer-no-cloud';
+  | 'release';
 
 type PropertyInfoTooltipTypeOptions = {
   docsUrl?: string;
@@ -82,18 +80,6 @@ const PROPERTY_INFO_TOOLTIP_TYPE_OPTIONS: Record<
     description:
       "The nx-release-publish target is used to publish your project with nxrelease. Don't invoke this directly - use nx release publish instead.",
     docsUrl: 'https://nx.dev/nx-api/nx/documents/release',
-  },
-  'atomizer-cloud': {
-    heading: 'Atomizer',
-    description:
-      'This target has been automatically split into multiple targets by file.',
-    docsUrl: 'https://nx.dev/ci/features/split-e2e-tasks',
-  },
-  'atomizer-no-cloud': {
-    heading: 'Atomizer',
-    description:
-      'This target has been automatically split into multiple targets by file. We recommend running this only with distribution enabled via Nx Cloud to avoid potential performance overheads.',
-    docsUrl: 'https://nx.dev/ci/features/split-e2e-tasks',
   },
 };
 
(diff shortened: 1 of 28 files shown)


graph/ui-tooltips/src/lib/property-info-tooltip.tsx
```diff
@@ -81,6 +83,20 @@ const PROPERTY_INFO_TOOLTIP_TYPE_OPTIONS: Record<
...
+    heading: 'Atomizer',
+    description:
+      'This target has been automatically split into multiple targets by file.',
+    docsUrl:
+      'https://nx.dev/ci/features/split-e2e-tasks#automatically-split-e2e-tasks-by-file-atomizer',
+  },
+  'atomizer-no-cloud': {
+    heading: 'Atomizer',
+    description:
+      'This target has been automatically split into multiple targets by file. We recommend running this only with distribution enabled via Nx Cloud to avoid potential performance overheads.',
```
jaysoo, line 96: The text here appear title-cased, I guess there is styling applied to the tooltip that needs to be fixed.

graph/ui-tooltips/src/lib/property-info-tooltip.tsx
```diff
@@ -81,6 +83,20 @@ const PROPERTY_INFO_TOOLTIP_TYPE_OPTIONS: Record<
       "The nx-release-publish target is used to publish your project with nxrelease. Don't invoke this directly - use nx release publish instead.",
     docsUrl: 'https://nx.dev/nx-api/nx/documents/release',
   },
+  'atomizer-cloud': {
+    heading: 'Atomizer',
+    description:
+      'This target has been automatically split into multiple targets by file.',
+    docsUrl:
+      'https://nx.dev/ci/features/split-e2e-tasks#automatically-split-e2e-tasks-by-file-atomizer',
```
jaysoo, line 91: Leave out the `#` part, it's not necessary.

graph/ui-tooltips/src/lib/property-info-tooltip.tsx
jaysoo: I would not mention the negatives (e.g. avoiding performance overheads). Focus on the value so users can try it a decide for themselves whether it makes sense for them or not. Something like:

```
Nx automatically split this potentially slow target into smaller targets by file that are faster to run. We recommend enabling Nx Cloud to benefit from better distribution, and more granular caching and flaky task re-runs.
```

Maybe @isaacplmann can fix the wording. I think we should mention that atomized targets make the other three features better (distribution, caching and re-runs).

jaysoo (approved):

packages/jest/src/plugins/plugin.spec.ts
```diff
@@ -81,6 +81,7 @@ describe('@nx/jest/plugin', () => {
                       },
                     ],
                     "metadata": {
+                      "atomizedTarget": undefined,
```
FrozenPandaz, line 84: Please get rid of the `undefined`. Only set the property when `options.ciTargetName` is defined. I don't really know why but `jest` sometimes isn't happy with this on different systems :shrug:

packages/nx/src/config/workspace-json-project-json.ts
```diff
@@ -128,6 +128,8 @@ export interface TargetMetadata {
   [k: string]: any;
   description?: string;
   technologies?: string[];
+  nonAtomizedTarget?: string;
+  atomizedTarget?: string;
```
FrozenPandaz, line 132: This is not used at the moment as far as I can see? If so, can we remove it?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #26629, opened 2024-06-21
feat(core): add --help content to project details view

jaysoo:
This PR adds help text for each inferred target that provides `metadata.help.command`. To pass options/args to the target, users are directed to open `project.json` and copy the values from `--help` output into `options` property of the target.

To display the options help section, the inferred target must provide metadata as follows:

```json5
 metadata: {
      help: {
        command: `foo --help`
        example: {
          options: {
            bar: true
          },
        },
      },
    },
```

The `help.command` value will be used to retrieve help text for the underlying CLI (e.g. `jest --help`). The `help.example` property contains sample options and args that users can add to their `project.json` file -- currently rendered in the hover tooltip of `project.json` hint text.

---

Example with `vite build --help`:

Screenshot 2024-06-21 at 3 06 21 PM

Result of clicking `Run`:
Screenshot 2024-06-21 at 3 06 24 PM

`project.json` tooltip hint:
Screenshot 2024-06-25 at 12 44 02 PM

### DIFF
--- a/packages/cypress/src/plugins/plugin.ts
+++ b/packages/cypress/src/plugins/plugin.ts
@@ -4,6 +4,7 @@ import {
   createNodesFromFiles,
   CreateNodesV2,
   detectPackageManager,
+  getPackageManagerCommand,
   joinPathFragments,
   logger,
   normalizePath,
@@ -183,6 +184,7 @@ async function buildCypressTargets(
   options: CypressPluginOptions,
   context: CreateNodesContext
 ): Promise<CypressTargets> {
+  const pmc = getPackageManagerCommand();
   const cypressConfig = await loadConfigFile(
     join(context.workspaceRoot, configFilePath)
   );
@@ -211,6 +213,12 @@ async function buildCypressTargets(
       metadata: {
         technologies: ['cypress'],
         description: 'Runs Cypress Tests',
+        example: {
+          command: `${pmc.exec} cypress run --help`,
+          example: {
+            args: ['--dev', '--headed'],
+          },
+        },
       },
     };
 
@@ -271,6 +279,12 @@ async function buildCypressTargets(
             technologies: ['cypress'],
             description: `Runs Cypress Tests in ${relativeSpecFilePath} in CI`,
             nonAtomizedTarget: options.targetName,
+            example: {
+              command: `${pmc.exec} cypress run --help`,
+              example: {
+                args: ['--dev', '--headed'],
+              },
+            },
           },
         };
         dependsOn.push({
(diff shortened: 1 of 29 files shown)


packages/cypress/src/plugins/plugin.ts
```diff
@@ -183,6 +184,7 @@ async function buildCypressTargets(
   options: CypressPluginOptions,
   context: CreateNodesContext
 ): Promise<CypressTargets> {
+  const pmc = getPackageManagerCommand();
```
FrozenPandaz, line 187: We need to be careful with this function. It spawns a function to check the version of some package managers.

Can you please move this higher so that we only call it once per plugin per graph creation?

In it's current spot, it will happen once per config file.

packages/nx/src/command-line/graph/graph.ts
```diff
@@ -1164,3 +1178,27 @@ async function createJsonOutput(
 
   return response;
 }
+
+function getHelpTextFromTarget(
```
FrozenPandaz, line 1182: Are we supporting this for the static web version of the graph app? No right?

FrozenPandaz (changes requested):

isaacplmann (approved):

MaxKless (approved):

packages/nx/src/command-line/graph/graph.ts
MaxKless: No that would have to be persisted somehow in `buildEnvironmentJs`. And we'd have to compute every single one of the help commands in the different projects beforehand. So maybe we should try to hide the whole feature in that case?
We can easily check that via `useEnvironmentConfig` and `environment`/`localMode` just like it's done in `graph/shared/src/lib/project-graph-data-service/get-project-graph-data-service.ts`.

jaysoo: That compute will be very expensive. We can hide it.

Coly010 (approved):

packages/nx/src/command-line/graph/graph.ts
jaysoo: Also hiding it for rendering in docs.

packages/cypress/src/plugins/plugin.ts
```diff
@@ -77,15 +80,17 @@ export const createNodes: CreateNodes<CypressPluginOptions> = [
     logger.warn(
       '`createNodes` is deprecated. Update your plugin to utilize createNodesV2 instead. In Nx 20, this will change to the createNodesV2 API.'
     );
-    return createNodesInternal(configFile, options, context, {});
+    const pmc = getPackageManagerCommand();
```
FrozenPandaz, line 83: This is still once per configFile. Can we please move it to module scope?

FrozenPandaz (changes requested):

xiongemi (approved):

FrozenPandaz (approved):

PR #26642, opened 2024-06-23
feat(core): update `actions/setup-node` to `v4` in `ci-workflow` generator

LayZeeDK:
Currently, the following warning is output when using GitHub Actions.

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/setup-node@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

## Current Behavior
- The `actions/setup-node` GitHub Action is installed in version 3.x which uses the deprecated Node.js 16 runtime for GitHub Actions

## Expected Behavior
- The `actions/setup-node` GitHub Action is installed in version 4.x which uses the supported Node.js 20 runtime for GitHub Actions

## Related Issue(s)

Fixes #


LayZeeDK:
I need a little help to update the Jest snapshots in the `workspace` project.

I don't have Rust on my local Windows machine and pnpm errs in the dev container.

I'll try to solve it but if someone with write permissions could update the relevant Jest snapshots, that would be great.

meeroslav (approved):
Thank you Lars!

I added the missing snapshot changes and refreshed the branch so build should be passing now.

PR #26645, opened 2024-06-23
feat(core): uncomment start-ci-run

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/generators/ci-workflow/__snapshots__/generator.spec.ts.snap
+++ b/packages/gradle/src/generators/ci-workflow/__snapshots__/generator.spec.ts.snap
@@ -17,9 +17,10 @@ jobs:
     steps:
       - checkout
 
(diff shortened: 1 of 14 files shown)


FrozenPandaz (changes requested):
Can you do it for the JS CI pipelines as well please?

FrozenPandaz (changes requested):
Let's have it so this generator checks if Nx Cloud is enabled. If it is, have the step commented in. If not, have the step commented out.

Also, make sure the comments make sense for both cases.

When Nx Cloud is enabled, the comment should not mention connecting your workspace or enabling task distribution.

When Nx Cloud is not enabled, the comment should mention connecting thew orkspace and the to uncomment the line.

Use `isNxCloudUsed`

FrozenPandaz (approved):

PR #26706, opened 2024-06-26
feat(nextjs): Add convert-to-inferred generator

ndcunningham:
This PR enables the ability to migrate project(s) from using nextjs executors to inferred targets.


packages/next/plugins/with-nx.ts
```diff
@@ -210,12 +213,11 @@ function withNx(
...
 
       const { createWebpackConfig } = require('@nx/next/src/utils/config');
+      // If we have file replacements or assets, inside of the next config we pass '.' as the workspaceRoot so that the paths are relative to the project.
       nextConfig.webpack = (a, b) =>
         createWebpackConfig(
-          workspaceRoot,
-          projectDirectory,
-          options.fileReplacements,
-          options.assets
+          _nextConfig.nx?.fileReplacements ? '.' : workspaceRoot,
```
jaysoo, line 219: For `@nx/next:build` executor usages, we still want to ensure paths relative to workspace root continue to work. I don't think we want to change how `createWebpackConfig` functions, but rather we should normalize the paths from `fileReplacements` and `assets` prior to calling `createWebpackConfig`.

packages/next/src/generators/convert-to-inferred/lib/build-post-target-transformer.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    delete options.generateLockfile;
+  }
+
+  if ('watch' in options) {
+    // Watch is default for serve not available while running 'build'
+    delete options.watch;
+  }
+
+  if ('debug' in options) {
```
jaysoo, line 117: We don't need to warn if the CLI flag that remains in `project.json` still work. Only warn if something isn't supported.

packages/next/plugins/with-nx.ts
jaysoo: make `.` an absolute paths

ndcunningham: Shouldn't `path.resolve` which is inside `createWebpackConfig` already ensures that the path we pass is absolute: https://nodejs.org/api/path.html#pathresolvepaths ?

I passed the `projectDirectory`  instead of `.` which is better for readability

jaysoo (approved):

Coly010 (approved):

packages/next/src/generators/convert-to-inferred/lib/update-next-config.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+describe('UpdateNextConfig', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+
+  // TODO: Add tests
+  it('should work', () => {
+    expect(true).toBe(true);
```
ndcunningham, line 14: Missing unit test i'll add these

PR #26709, opened 2024-06-26
fix(linter): dependency checks should respect pnpm workspace versions

### LINKED ISSUE #26707: Dependency-checks rule does not respect the pnpm `workspace: *` version signature
### Current Behavior

I am working on few various types of NX monorepo package based projects and I have faced an issue with nx-dependency-checks eslint rule after upgrading to PNPM version 9.

PNPM enforces to use "workspace:" protocol for workspace packages but ESLint fails when I use my workspace libraries with such depndendencies mark.

I am using ESLint flat config and so far I have been marking my workspace packages devDependencies in package.json as e.g. "@org/my-lib": "*" but after upgrading to PNPM v9, I changed marks to "@org/my-lib": "workspace:*".

ESLint fails with error:
The version specifier does not contain the installed version of "@org/my-lib" package: 0.0.1

### Expected Behavior

ESLint properly maps the version, like it does for `*`

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create several libs in pnpm workspace with pnpm v9
2. Set pnpm workspace
3. Import lib A into lib B
4. Run lint with dependency-checks on lib B

### Nx Report

```shell
19.2.0 (likely on all v19 versions)
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

My pnpm-workspace.yaml looks like this:
packages:
- 'apps/**'
- 'libs/*

meeroslav:
## Related Issue(s)

Fixes #26707

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.spec.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.spec.ts
@@ -1580,68 +1580,125 @@ describe('Dependency checks (eslint)', () => {
           - external1"
     `);
   });
-});
 
-it('should require swc if @nx/js:swc executor', () => {
-  const packageJson = {
-    name: '@mycompany/liba',
-    dependencies: {
-      external1: '^16.0.0',
-    },
-  };
+  it('should not report * and workspace:*', () => {
+    const packageJson = {
+      name: '@mycompany/liba',
+      dependencies: {
+        external1: '*',
+        external2: 'workspace:*',
+      },
+    };
 
-  const swcrc = {
-    jsc: {
-      externalHelpers: true,
-    },
-  };
+    const fileSys = {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/dependency-checks.spec.ts
```diff
@@ -1580,68 +1580,125 @@ describe('Dependency checks (eslint)', () => {
           - external1"
     `);
   });
-});
 
-it('should require swc if @nx/js:swc executor', () => {
```
meeroslav, line 1585: This test has been moved into `describe` block from outside

JamesHenry (approved):

PR #26741, opened 2024-06-27
fix(gradle): fix gradle exclude src/test

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/generators/init/init.ts
+++ b/packages/gradle/src/generators/init/init.ts
@@ -146,7 +146,7 @@ allprojects {
   }
 }
 
-function updateNxJsonConfiguration(tree: Tree) {
+export function updateNxJsonConfiguration(tree: Tree) {
   const nxJson = readNxJson(tree);
 
   if (!nxJson.namedInputs) {
@@ -158,7 +158,7 @@ function updateNxJsonConfiguration(tree: Tree) {
   );
   const productionFileSet = nxJson.namedInputs.production ?? [];
   nxJson.namedInputs.production = Array.from(
-    new Set([...productionFileSet, 'default', '!{projectRoot}/test/**/*'])
+    new Set([...productionFileSet, 'default', '!{projectRoot}/src/test/**/*'])
   );
   updateNxJson(tree, nxJson);
 }
--- a/packages/gradle/src/migrations/19-4-1/change-regex-test-production.spec.ts
+++ b/packages/gradle/src/migrations/19-4-1/change-regex-test-production.spec.ts
@@ -1,0 +1,71 @@
+import { Tree, readNxJson } from '@nx/devkit';
+import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
+import update from './change-regex-test-production';
+
+describe('change-regex-test-production', () => {
+  let tree: Tree;
+
+  beforeAll(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+
+  it('should not add to the namedInputs if it does not exist', async () => {
+    tree.write(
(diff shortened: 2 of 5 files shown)


packages/gradle/src/generators/init/init.ts
```diff
@@ -158,7 +158,7 @@ function updateNxJsonConfiguration(tree: Tree) {
   );
   const productionFileSet = nxJson.namedInputs.production ?? [];
   nxJson.namedInputs.production = Array.from(
-    new Set([...productionFileSet, 'default', '!{projectRoot}/test/**/*'])
+    new Set([...productionFileSet, 'default', '!{projectRoot}/src/test/**/*'])
```
FrozenPandaz, line 161: I think this needs a migration :thinking: 

Workspaces right now do not have the right production fileset.

FrozenPandaz (changes requested):

packages/gradle/src/migrations/19-4-1/change-regex-test-production.ts
```diff
@@ -0,0 +1,19 @@
...
+    return;
+  }
+  if (tree.exists('nx.json')) {
+    const nxJson = readNxJson(tree);
+    const production = nxJson?.namedInputs?.production;
+    if (production?.includes('!{projectRoot}/test/**/*')) {
+      production.splice(production.indexOf('!{projectRoot}/test/**/*'), 1);
+      tree.write('nx.json', JSON.stringify(nxJson, null, 2));
+    }
+    updateNxJsonConfiguration(tree);
```
FrozenPandaz, line 17: Don't reuse this function.

1. Do not update `production` namedInput if it does not exist.
2. Replace the incorrect one with the correct one when it does exist

FrozenPandaz (changes requested):

packages/gradle/src/migrations/19-4-1/change-regex-test-production.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    update(tree);
+    expect(readNxJson(tree)).toMatchInlineSnapshot(`
+      {
+        "namedInputs": {
+          "default": [
+            "{projectRoot}/**/*",
+          ],
+          "production": [
+            "default",
+            "!{projectRoot}/src/test/**/*",
```
FrozenPandaz, line 53: This one should do nothing

packages/gradle/src/migrations/19-4-1/change-regex-test-production.spec.ts
```diff
@@ -0,0 +1,89 @@
...
```
FrozenPandaz, line 26: This one should do nothing

FrozenPandaz (changes requested):

packages/gradle/src/migrations/19-4-1/change-regex-test-production.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (!hasGradlePlugin(tree)) {
+    return;
+  }
+  if (tree.exists('nx.json')) {
+    const nxJson = readNxJson(tree);
+    const production = nxJson?.namedInputs?.production;
+    if (production?.includes('!{projectRoot}/test/**/*')) {
+      production[production.indexOf('!{projectRoot}/test/**/*')] =
+        '!{projectRoot}/src/test/**/*';
+      tree.write('nx.json', JSON.stringify(nxJson, null, 2));
```
FrozenPandaz: use `updateNxJson`

FrozenPandaz (approved):

PR #26750, opened 2024-06-28
fix(vite): infer build and serve targets when rollupOptions.input is present

jaysoo:
This PR adds a check for `build.rollupOptions.input` in addition to `build.rollupOptions.inputs`. Plugins like `sveltekit()` will use the former property, and we should infer build/server targets correctly in those cases.

## Current Behavior
`sveltekit()` is not supported

## Expected Behavior
`sveltekit()` and other setups using `input` are supported

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -365,7 +365,7 @@ function getOutputs(
 
   const isBuildable =
     build?.lib ||
-    build?.rollupOptions?.inputs ||
+    build?.rollupOptions?.input ||
     existsSync(join(workspaceRoot, projectRoot, 'index.html'));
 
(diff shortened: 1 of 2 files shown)


packages/vite/src/plugins/plugin.ts
```diff
@@ -365,6 +365,7 @@ function getOutputs(
 
   const isBuildable =
     build?.lib ||
+    build?.rollupOptions?.input ||
     build?.rollupOptions?.inputs ||
```
Coly010, line 369: Is `inputs` even a valid RollupOption? https://rollupjs.org/configuration-options/#input

Coly010 (approved):
1 comment

PR #26754, opened 2024-06-28
docs(core): link to add plugin instructions

isaacplmann:
Link from the plugin registry to the instructions add a plugin to the registry


juristr (approved):
Works for now, but I wonder whether we should have it somewhere at the top. But I have no good idea rn how to make it not too intrusive as the main use case here is to search for existing one.

PR #26767, opened 2024-06-29
fix(nx-dev): plugin-registry menu

juristr:
## Current Behavior

When on https://nx.dev/plugin-registry the side-bar menu doesn't open on mobile.

## Expected Behavior

Should obviously open. Also adjusted the top-level header menu "Documentation" -> "Docs" (just sounds better)

## Related Issue(s)

Fixes #


isaacplmann (changes requested):
I don't think it looks good to have a sidebar menu on that page.  You're just repeating the extending Nx menu.  If you want the plugins to be combined with extending Nx, then we should actually combine the two sections and come up with a design for that.

isaacplmann:
I realize now that you're trying add the top menu on mobile.  You succeeded at that, but also added a sidebar on desktop - that shouldn't be there.

juristr:
> I realize now that you're trying add the top menu on mobile. You succeeded at that, but also added a sidebar on desktop - that shouldn't be there.

Ohh...yeah now I understand...that component needs refactoring. I'll fix that.

jaysoo (approved):

PR #26777, opened 2024-07-01
feat(module-federation): use @module-federation/enhanced for withModuleFederation

Coly010:
## Current Behavior

We currently use the `ModuleFederationPlugin` from `webpack` itself. 
This doesn't receive the same attention as the new `@module-federation/enhanced` package, which is where all new work by Zack Jackson and Bytedance is being conducted.

## Expected Behavior

Our `withModuleFederation` utils should use the `@module-federation/enhanced` package.

## Related Issue(s)

#26322 
Fixes #


packages/react/migrations.json
```diff
@@ -202,6 +202,15 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "19.4.0-module-federation": {
+      "version": "19.4.0-rc.2",
```
jaysoo, line 207: Let's use 19.5 since 19.4 is released.

jaysoo (approved):

PR #26782, opened 2024-07-01
feat(module-federation): use proxy servers to proxy to single file server for static remotes

### LINKED ISSUE #26318: Angular MFE setup is broken after migration from 17 to 18 or 19
### Current Behavior

We have a setup that uses NX for angular micro frontends. We utilize dynamic module federation for that. We have a shell app and several remote apps. Obviously, we have a manifest file that we use locally to run MFEs. In the manifest, we have the names of remote apps and the corresponding hosts and ports. A very standard setup that is suggested and even generated by NX. Ports in the manifest are different, of course, but after migration to the latest version, NX now runs just one static server for all remote apps. No problem, we can change all ports in manifest to point to the same host and port right? But what do we do when we want to run just one remote app via a dev server? Now I have to go and change the port in the manifest file for that remote app. I cannot commit this, as this is not what we expect from a dev server. So now, at least, the local setup for our MFEs is broken if we migrate to the latest version, and we are stuck on 17 because of this.

[Here is the replication of the project after migration to 19th](https://github.com/VanTigranyan/nx-mfe-repl-19) version. It doesn't even run.

### Expected Behavior

We expect the old setup to still be supported, either by opting out of serving all remote apps by just one static server or with some other solution. The NX documentation and guides do not reflect the new logic of the local development setup, so we don't know what to do.

[Here is the replication of our setup for NX 17](https://github.com/VanTigranyan/nx-mfe-repl-17), which works perfectly well. If we migrate it using NX to latest versions it breaks. 

### GitHub Repo

https://github.com/VanTigranyan/nx-mfe-repl-19

### Steps to Reproduce

We can replicate this with 2 methods.
1. Clone [this repo](https://github.com/VanTigranyan/nx-mfe-repl-17). This is the replication of our current setup using NX 17.
2. Try to migrate it to NX 18 or 19
3. Verify that the setup is broken by running `nx serve shell`
4. If we fix the issue, we still cannot run remote applications on separate ports without using --devRemotes, that serves remote apps via webpack dev server making the testing cycle very long.

Second method.
1. Clone [already migrated repo](https://github.com/VanTigranyan/nx-mfe-repl-19).
2. npm install
3. nx serve shell
4. Verify that the setup is broken
5. Follow the steps for the first variant above.

### Nx Report

```shell
`After migration`:

Node   : 18.19.1
OS     : darwin-arm64
npm    : 9.9.3

nx                 : 19.1.1
@nx/js             : 19.1.1
@nx/jest           : 19.1.1
@nx/linter         : 19.1.1
@nx/eslint         : 19.1.1
@nx/workspace      : 19.1.1
@nx/angular        : 19.1.1
@nx/devkit         : 19.1.1
@nx/eslint-plugin  : 19.1.1
@nrwl/tao          : 19.1.1
@nx/web            : 19.1.1
@nx/webpack        : 19.1.1
typescript         : 5.4.5

`Before migration`:

   Node   : 18.19.1
   OS     : darwin-arm64
   npm    : 9.9.3
   
   nx                 : 17.1.2
   @nx/js             : 17.1.2
   @nx/jest           : 17.1.2
   @nx/linter         : 17.1.2
   @nx/eslint         : 17.1.2
   @nx/workspace      : 17.1.2
   @nx/angular        : 17.1.2
   @nx/cypress        : 17.1.2
   @nx/devkit         : 17.1.2
   @nx/eslint-plugin  : 17.1.2
   @nrwl/tao          : 17.1.2
   @nx/web            : 17.1.2
   @nx/webpack        : 17.1.2
   typescript         : 5.2.2
```

### Failure Logs

```shell
Error logs after migration:

> nx run shell:serve:development


 NX  Starting module federation dev-server for shell with 3 remotes


 NX   Cannot read properties of undefined (reading 'find')


TypeError: Cannot read properties of undefined (reading 'find')
    at startRemotes (/Users/vtigranyan/projects/wf-fusion/nx-mfe-repl-18/node_modules/@nx/angular/src/executors/module-federation-dev-server/lib/start-dev-remotes.js:16:58)
    at moduleFederationDevServerExecutor (/Users/vtigranyan/projects/wf-fusion/nx-mfe-repl-18/node_modules/@nx/angular/src/executors/module-federation-dev-server/module-federation-dev-server.impl.js:81:61)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at getLastValueFromAsyncIterableIterator (/Users/vtigranyan/projects/wf-fusion/nx-mfe-repl-18/node_modules/nx/src/utils/async-iterator.js:15:19)
    at iteratorToProcessStatusCode (/Users/vtigranyan/projects/wf-fusion/nx-mfe-repl-18/node_modules/nx/src/command-line/run/run.js:44:29)
    at handleErrors (/Users/vtigranyan/projects/wf-fusion/nx-mfe-repl-18/node_modules/nx/src/utils/params.js:9:16)
    at process.<anonymous> (/Users/vtigranyan/projects/wf-fusion/nx-mfe-repl-18/node_modules/nx/bin/run-executor.js:59:28)

————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Running target serve for project shell failed

Failed tasks:

- shell:serve:development

Hint: run the command with --verbose for more details.
```

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

To summarize, we have two problems.
1. Migration breaks MFE setup.
2. If we fix this, now we cannot serve remote apps separately on a static server, making dynamic module federation un-runnable locally without using --devRemotes.

Coly010:
## Current Behavior

Remotes that are served for a host are usually served from a single file server running on a single port.
We perform some mapping logic during the build of the host application to update the locations the remotes can be found at to point to the single file server.

This works, but it's also wrong, as it breaks the flow that users expect.
It also breaks dynamic remotes.

## Expected Behavior

Continue to serve the remotes from a single file server, as this helps reduce the amount of resources used on developers machines.
Use express to create proxy servers that will proxy requests from the original remote location to the single file server.

This allows applications to continue to work without us having to interfere and map any remote locations.
It also solves the issue with dynamic remotes.

## Related Issue(s)

Fixes #26318


packages/angular/src/executors/module-federation-dev-server/module-federation-dev-server.impl.ts
```diff
@@ -39,7 +40,6 @@ export async function* moduleFederationDevServerExecutor(
   // Force Node to resolve to look for the nx binary that is inside node_modules
   const nxBin = require.resolve('nx/bin/nx');
   const options = normalizeOptions(schema);
-  options.staticRemotesPort ??= options.port + 1;
```
jaysoo, line 42: Removed this because it needs to be defaulted from the results of `getRemotes` (just like in React). Otherwise there could be a port conflict if one of the dev remotes has a port that's host port + 1.

jaysoo (approved):

VanTigranyan:
Hi @Coly010, thanks a lot for this fix. Is there a documentation that I can look into understand how this works?

Coly010:
@VanTigranyan No documentation as of yet, but it is coming.

Essentially, the process is now

1. Build all static remotes
2. Start a single `http-server` serving all built remotes at an arbitrary port
3. For each static remote, start an express server at the port defined in the remote's `serve` target
  3a. express server will proxy requests to the `http-server` with the correct endpoint for the remote.
  
 This means we continue to get the machine resource savings without doing any under the hood mapping of remote locations etc.

PR #26790, opened 2024-07-02
fix(nx-cloud): onboarding cloud version handling

mandarini:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/nx-cloud/utilities/url-shorten.spec.ts
+++ b/packages/nx/src/nx-cloud/utilities/url-shorten.spec.ts
@@ -1,6 +1,6 @@
 import {
   removeVersionModifier,
-  compareCalver,
+  compareCleanCloudVersions,
   getNxCloudVersion,
 } from './url-shorten';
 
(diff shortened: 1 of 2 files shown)


packages/nx/src/nx-cloud/utilities/url-shorten.ts
```diff
@@ -141,10 +141,38 @@ async function getNxCloudVersion(apiUrl: string): Promise<string | null> {
...
+  }
+  if (v1.month !== v2.month) {
+    return v1.month > v2.month ? 1 : -1;
+  }
+  if (v1.day !== v2.day) {
+    return v1.day > v2.day ? 1 : -1;
+  }
+
+  return 0;
+}
```
jdpearce, line 178: I think this is technically wrong, we don't use `year.month.day`, we use `YYMM.DD.BUILD`

mandarini: I am pretty sure the comparison results will be correct regardless, but I will update the names of the variables and function to avoid confusion.

packages/nx/src/nx-cloud/utilities/url-shorten.ts
```diff
@@ -14,8 +13,10 @@ export async function shortenedCloudUrl(
   );
 
   const version = await getNxCloudVersion(apiUrl);
-
-  if (version && lt(removeVersionModifier(version), '2406.11.5')) {
+  if (
```
jdpearce, line 16: What if this function fails?

Tbh, I'd wrap this entire function in a `try/catch`.

packages/nx/src/nx-cloud/utilities/url-shorten.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  compareCleanCloudVersions,
+  getNxCloudVersion,
+} from './url-shorten';
+
+jest.mock('axios', () => ({
+  get: jest.fn(),
+}));
+
+describe('URL shorten various functions', () => {
+  describe('compareCleanCloudVersions', () => {
```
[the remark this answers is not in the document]
mandarini: Updated

packages/nx/src/nx-cloud/utilities/url-shorten.ts
mandarini: Updated

catfireparty (approved):

Cammisuli (approved):

PR #26793, opened 2024-07-02
feat(misc): add nx syncing mechanism and update the typescript-sync generator

leosvelperez:
- Add the `nx sync` command to run sync generators and apply changes to bring the workspace up to date according to their logic.
- Add the `nx sync:check` command to validate that the workspace is up to date by running the sync generators without applying the changes. It can be used on CI as a validation check.
- Update the task runner to run the sync generators (or obtain their state from the daemon) and prompt the user whether to apply the changes, if any
  - This is only run if the `NX_ENABLE_SYNC_GENERATORS` environment variable is set to `'true'`
  - Allow the user to configure a default value for the prompt
- Update the `@nx/js:typescript-sync` generator (keep tsconfig project references in sync with the project graph) with misc fixes

## Related Issue(s)

Fixes #


packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -0,0 +1,315 @@
...
+        sourceProjectTsconfigPath,
+        dependencies,
+        sourceProjectNode.data.root
+      ) || hasChanges;
+  }
+
+  if (hasChanges) {
+    await formatFiles(tree);
+
+    return 'Based on the workspace project graph, some TypeScript configuration files are missing project references to the projects they depend on.';
```
FrozenPandaz, line 182: Ensure this format supports having a callback task.

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -0,0 +1,315 @@
...
+): boolean {
+  const tsConfig = readJson<Tsconfig>(tree, tsConfigPath);
+  // We have at least one dependency so we can safely set it to an empty array if not already set
+  const references = [];
+  const originalReferencesSet = new Set();
+  const newReferencesSet = new Set();
+  for (const ref of tsConfig.references ?? []) {
+    const normalizedPath = normalizeReferencePath(ref.path);
+    originalReferencesSet.add(normalizedPath);
+    if (!ref.path.startsWith('../')) {
```
FrozenPandaz, line 204: Consider nested projects

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -0,0 +1,315 @@
...
+          if (tree.exists(possibleRuntimeTsConfigPath)) {
+            referencePath = possibleRuntimeTsConfigPath;
+            break;
+          }
+        }
+      }
+    }
+    const relativePathToTargetRoot = relative(projectRoot, referencePath);
+    if (!newReferencesSet.has(relativePathToTargetRoot)) {
+      newReferencesSet.add(relativePathToTargetRoot);
```
FrozenPandaz, line 240: Don't delete if the file exists.

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -0,0 +1,315 @@
...
+  if (hasChanges) {
+    tsConfig.references = references;
+    writeJson(tree, tsConfigPath, tsConfig);
+  }
+
+  return hasChanges;
+}
+
+// Collect the dependencies of a project recursively sorted from root to leaf
+function collectProjectDependencies(
```
FrozenPandaz, line 260: Check if we have this logic elsewhere.

packages/nx/src/command-line/sync/command-object.ts
```diff
@@ -0,0 +1,33 @@
...
+  check?: boolean;
+  verbose?: boolean;
+}
+
+export const yargsSyncCommand: CommandModule<
+  Record<string, unknown>,
+  SyncOptions
+> = {
+  command: 'sync',
+  describe:
```
FrozenPandaz, line 13: Set this to `false` to not generate and show documentation

packages/nx/src/daemon/client/client.ts
```diff
@@ -342,6 +359,56 @@ export class DaemonClient {
     return this.sendMessageToDaemon(message);
   }
 
+  async getSyncGeneratorChanges(
+    generators: string[]
+  ): Promise<SyncGeneratorChangesResult[]> {
```
FrozenPandaz, line 364: Update this to be only enough information to trigger a flush.

packages/nx/src/daemon/client/client.ts
```diff
@@ -342,6 +359,56 @@ export class DaemonClient {
...
+        if (change.content) {
+          // We need to convert the content back to a buffer
+          change.content = Buffer.from(change.content);
+        }
+      }
+    }
+    return results;
+  }
+
+  clearSyncGeneratorChanges(generators: string[]): Promise<void> {
```
FrozenPandaz, line 383: This should change to flush the changes (which will clear the cache)

packages/nx/src/daemon/client/client.ts
```diff
@@ -342,6 +359,56 @@ export class DaemonClient {
...
+  }
+
+  getRegisteredSyncGenerators(): Promise<string[]> {
+    const message: HandleGetRegisteredSyncGeneratorsMessage = {
+      type: GET_REGISTERED_SYNC_GENERATORS,
+    };
+    return this.sendToDaemonViaQueue(message);
+  }
+
+  updateContextFiles(
```
FrozenPandaz, line 398: nit: Rename to `updateWorkspaceContext`?

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -413,3 +432,9 @@ async function resetInternalStateIfNxDepsMissing() {
     await resetInternalState();
   }
 }
+
+function notifyProjectGraphRecomputationListeners(projectGraph: ProjectGraph) {
+  for (let listener of projectGraphRecomputationListeners) {
```
FrozenPandaz, line 437:
```suggestion
  for (const listener of projectGraphRecomputationListeners) {
```

FrozenPandaz (changes requested):

packages/js/src/generators/typescript-sync/typescript-sync.ts
leosvelperez: Not in the way is used here and this one might get simplified anyways if it's confirmed we don't need transitive deps.

packages/js/src/generators/typescript-sync/typescript-sync.ts
leosvelperez: It's a bit more complex than that and I'll address it in a separate PR after this one lands.

packages/nx/src/daemon/server/sync-generators.ts
```diff
@@ -0,0 +1,245 @@
...
+    `running sync generator ${generator}`
+  );
+
+  // run the generator and cache the result
+  syncGeneratorsCacheResultPromises.set(
+    generator,
+    runSyncGenerator(tree, generator, projects).then((result) => {
+      log(generator, 'changes:', result.changes.map((c) => c.path).join(', '));
+      return result;
+    })
```
FrozenPandaz, line 230: Does this handle when the sync generator is not async?

packages/nx/src/daemon/server/sync-generators.ts
```diff
@@ -0,0 +1,245 @@
...
+    const getProjectsConfigurations = async () => {
+      if (projects) {
+        return projects;
+      }
+
+      const { projectGraph } = await getCachedSerializedProjectGraphPromise();
+      projects =
+        readProjectsConfigurationFromProjectGraph(projectGraph).projects;
+
+      return projects;
```
FrozenPandaz, line 62: This could fail.. we could either "noop" when that happens... or we can use the partial graph. Either way, even if we noop, we should say that the workspace is out of sync.

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -198,6 +209,171 @@ export async function runCommand(
...
+        : willErrorOnCiMessage,
+    ],
+  });
+
+  const applyChanges =
+    nxJson.sync?.applyChanges === true ||
+    (await promptForApplyingSyncGeneratorChanges());
+
+  if (applyChanges) {
+    const spinner = ora('Syncing workspace configuration...');
```
FrozenPandaz, line 309: Can we keep it more general as the "workspace" is out of sync rather than the "workspace configuration" is out of sync?

FrozenPandaz (changes requested):

isaacplmann (approved):
Generated documentation looks fine.  We'll need a recipe later describing how plugins can hook into the sync feature.

packages/nx/src/daemon/server/sync-generators.ts
leosvelperez: Good catch. I'll update it so this logic doesn't error.

I don't think we should say the workspace is out of sync when the project graph is broken. `nx sync` does not concern itself with ensuring the project graph is correct. Its concern is to ensure that the workspace is in sync with the project graph (or any other file that's used as the source of truth).

If the project graph is broken, `nx sync` simply can't identify whether things are out of sync or not. In that case, the workspace is not "out of sync" (in terms of sync generators returning that there are files to sync); it is in a broken state (the project graph can't be calculated).

When the project graph is broken, the relevant error from the graph creation will surface and we won't get to the `nx sync` (or the `run-command`) syncing handling. So, we wouldn't even need to report that the workspace is out of sync because the graph is broken. The more relevant error for the broken graph would be surfaced first.

packages/nx/src/daemon/server/sync-generators.ts
leosvelperez: It does. The `runSyncGenerator` is an async function that'll always return a `Promise`. I've also tested it with a mocked sync version of the `typescript-sync` generator.

packages/nx/src/daemon/server/sync-generators.ts
leosvelperez: I updated the daemon processing for sync generators to handle a broken project graph. It will be a noop, and the results won't be cached. That allows for recovery once the project graph is fixed. As mentioned above, we don't need to report the workspace is out of sync because the client would never get to request the syncing status when the project graph is broken.

FrozenPandaz (approved):

PR #26806, opened 2024-07-03
fix(core): ignore errors from cleanupNativeFileCache (revert to previous behaviour)

MaxKless:
## Current Behavior
errors can crop up when running `nx reset` on windows. 

## Expected Behavior
those errors are not critical so they should be ignored.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/reset/reset.ts
+++ b/packages/nx/src/command-line/reset/reset.ts
@@ -54,7 +54,7 @@ export async function resetHandler(args: ResetCommandOptions) {
     try {
       await cleanupNativeFileCache();
     } catch {
-      errors.push('Failed to clean up the native file cache.');
+      // ignore, deleting the native file cache is not critical and can fail if another process is locking the file
     }
     try {
       await cleanupWorkspaceData();


ErickRodrCodes:
Some clues here Max:
1. apparently rimraf might not be the best tool here, as I was able to track where the temp folders.... and it is creating an absurd amount of those folders:

image

my guess is it dies trying to `rmSync` such folder here:

```
function cleanupNativeFileCache() {
    return incrementalBackoff(INCREMENTAL_BACKOFF_FIRST_DELAY, INCREMENTAL_BACKOFF_MAX_DURATION, () => {
        (0, fs_extra_1.rmSync)((0, native_file_cache_location_1.getNativeFileCacheLocation)(), { recursive: true, force: true });
    });
}
```

Might be an issue with fs-extra?

FrozenPandaz (approved):

MaxKless:
Thanks for doing some extra reseach @ErickRodrCodes !
You're right, it's the nxls process which is keeping this file in memory. This shouldn't really cause issues because we can just ignore the error that pops up when deleting the file. We do have to make sure everything is deleted though.
And in your case it looks like way too many folders are being generated. Unless you have 15 workspaces are open at once?

Could you create a new issue with some info about your setup and all the folders that are being generated. I'm also interested in your OS user... is it changing or something?

Thanks, would love to work together on tracking it down :)

ErickRodrCodes:
sure thing. I can tell you that when I have hit `nx reset` due to some change in the workspace (building plugins and testing it in two projects using verdaccio), there are times that the command 'hangs' itself. there are other times that I see spawn terminals showing up (if you have windows 11 and use terminal and not powershell as your default terminal for your OS [not vs code]) at the point that those will fill my taskbar with empty spawn terminals. 

Sometimes multiple spawns are generated at the same time, sometimes 3, some times dozens... so I'm not sure what is the variable here that might cause such increase on the amount of spawn children requesting something.

I will generate an issue later on with more details. I assume it is safe to run an `nx migrate latest` as this will merge the changes for cleanupNativeFileCache

PR #26814, opened 2024-07-03
docs(core): sync top menu features with sidebar

isaacplmann:
- Syncs top menu features drop down names with sidebar
- Moves Setup enterprise feature under Enterprise -> On Premise
- Collapse concepts section in CI tab


jaysoo:
We should also add `Automatically Split E2E Tasks (Atomizer)` and `Identify and Re-run Flaky Tasks` to the Features menu. We don't need the `new` label.

jaysoo (approved):

PR #26824, opened 2024-07-04
fix(nextjs): schema type for unitTestRunner for library

Demianeen:
## Current Behavior
The [docs](https://nx.dev/nx-api/next/generators/library#unittestrunner) say that you can use vitest as a unitTestRunner option. It also worked when I tested with the following command:
```sh
nx g @nx/next:library shared --dir=modules/shared --style none --dry-run --projectNameAndRootFormat as-provided --component false --unitTestRunner vitest
```

But types say otherwise:
CleanShot 2024-07-04 at 16 24 44@2x

## Expected Behavior
Types will allow 'vitest' option for a unitTestRunner

## Related Issue(s)
none

### DIFF
--- a/packages/next/src/generators/library/schema.d.ts
+++ b/packages/next/src/generators/library/schema.d.ts
@@ -13,7 +13,7 @@ export interface Schema {
   pascalCaseFiles?: boolean;
   routing?: boolean;
   appProject?: string;
-  unitTestRunner: 'jest' | 'none';
+  unitTestRunner: 'jest' | 'vitest' | 'none';
   linter: Linter | LinterType;
   component?: boolean;
   publishable?: boolean;


ndcunningham:
Hi, this alone would not be enough to enable `vitest` as your `unitTestRunner`.

If you want to put together an updated PR here is what you would need to do:

Currently, we are only generating `jest` so you would need to create the `vitest` config and also install the necessary packages.

It is done like this in our react generator: https://github.com/nrwl/nx/blob/master/packages/react/src/generators/application/application.ts#L224

So you would need to change: https://github.com/nrwl/nx/blob/master/packages/next/src/generators/application/application.ts#L68 to to check for which `unitTestRunner` was passed in and generate the files accordingly.

Demianeen:
I did a little typo in the issue description. I tried to run generate command with vitest like:
```
nx g @nx/next:library shared --dir=modules/shared --style none --dry-run --projectNameAndRootFormat as-provided --component false --unitTestRunner vitest
```
And it worked for me. The docs also say that you can use vitest value as a unitTestRunner for this next generator

From: Nicholas Cunningham ***@***.***>
Date: Thursday, 4 July 2024 at 19:04
To: nrwl/nx ***@***.***>
Cc: Feliche-Demian Netliukh ***@***.***>, Author ***@***.***>
Subject: Re: [nrwl/nx] fix(next): schema type for unitTestRunner (PR #26824)

Hi, this alone would not be enough to enable vitest as your unitTestRunner.

If you want to put together an updated PR here is what you would need to do:

Currently, we are only generating jest so you would need to create the vitest config and also install the necessary packages.

It is done like this in our react generator: https://github.com/nrwl/nx/blob/master/packages/react/src/generators/application/application.ts#L224

So you would need to change: https://github.com/nrwl/nx/blob/master/packages/next/src/generators/application/application.ts#L68 to to check for which unitTestRunner was passed in and generate the files accordingly.

xiongemi (approved):

PR #26838, opened 2024-07-05
chore(core): update pnpm/action-setup to v4 in ci-workflow generator

niklas-wortmann:
I just created a new nx monorepo with pnpm and in the github.ci file it uses pnpm/action-setup@v2. 
I ran into the following issue https://github.com/pnpm/action-setup/issues/136 and updating to pnpm/action-setup@v4 solved the issue for me


meeroslav (changes requested):
Looks good. Thank you for providing the fix.

Please update also tests/fixtures to match the template change.

meeroslav (approved):

PR #26864, opened 2024-07-08
fix(nx-dev): Footer add spacing to show hidden content

ndcunningham:
Before:
Screenshot 2024-07-08 at 1 03 30 PM

After: 
Screenshot 2024-07-08 at 1 04 29 PM


ndcunningham:
Not needed

PR #26898, opened 2024-07-11
feat(core): error when running atomized tasks outside of DTE

MaxKless:
## Current Behavior
Nx allows running atomized tasks (like `e2e-ci`) anywhere.

## Expected Behavior
Nx allows running atomized tasks (like `e2e-ci`) only in distributed environments like Nx Agents or DTE. We'll leave that actual check to nx cloud but in the default runner, we'll throw an error.
It's possible to escape from this with an env var.

### DIFF
--- a/docs/nx-cloud/features/split-e2e-tasks.md
+++ b/docs/nx-cloud/features/split-e2e-tasks.md
@@ -7,11 +7,11 @@ width="100%" /%}
 
 In almost every codebase, e2e tests are the largest portion of the CI pipeline. Typically, e2e tests are grouped by application so that whenever an application's code changes, all the e2e tests for that application are run. These large groupings of e2e tests make caching and distribution less effective. Also, because e2e tests deal with a lot of integration code, they are at a much higher risk to be flaky.
 
-You could manually address these problems by splitting your e2e tests into smaller tasks, but this requires developer time to maintain and adds additional configuration overhead to your codebase. Or, you could allow Nx to automatically split your Cypress or Playwright e2e tests by file.
+You could manually address these problems by splitting your e2e tests into smaller tasks, but this requires developer time to maintain and adds additional configuration overhead to your codebase. Or, you could allow Nx to automatically split your e2e tests by file. Doing this will split your large e2e tasks into smaller, atomized tasks.
 
 ## Set up
 
-To enable automatically split e2e tasks, you need to turn on [inferred tasks](/concepts/inferred-tasks#existing-nx-workspaces) for the [@nx/cypress](/nx-api/cypress), [@nx/playwright](/nx-api/playwright), or [@nx/jest](/nx-api/jest) plugins. Run this command to set up inferred tasks:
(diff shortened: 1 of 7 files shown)


packages/nx/src/tasks-runner/task-graph-utils.ts
```diff
@@ -66,3 +70,42 @@ export function makeAcyclic(graph: {
...
+  projectGraph: ProjectGraph
+): void {
+  if (process.env['NX_SKIP_ATOMIZER_VALIDATION']) {
+    return;
+  }
+  const tasksWithAtomizer = Object.values(taskGraph.tasks).filter(
+    (task) =>
+      projectGraph.nodes[task.target.project]?.data?.targets?.[
+        task.target.target
+      ]?.metadata?.nonAtomizedTarget !== undefined
```
FrozenPandaz, line 85: Mm, not sure I like this. `metadata` is not supposed to affect running of tasks :thinking:

FrozenPandaz (changes requested):

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -388,7 +388,15 @@ export class TaskOrchestrator {
         task,
         this.projectGraph
       );
-      if (
+
+      if (targetConfiguration.executor === 'nx:noop') {
```
FrozenPandaz, line 392: Was this important?

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -129,6 +133,14 @@ function createTaskGraphAndValidateCycles(
     }
   }
 
+  // validate that no atomized tasks like e2e-ci are used without Nx Cloud
+  if (
+    !isNxCloudUsed(readNxJson()) &&
```
FrozenPandaz, line 138: Mm this makes sense

docs/nx-cloud/features/split-e2e-tasks.md
```diff
@@ -34,6 +34,13 @@ nx add @nx/playwright
 nx add @nx/jest
 ```
 
+{% /tab %}
+{% tab label="Gradle" %}
+
+```shell {% skipRescope=true %}
+nx add @nx/gradle
+```
+
```
FrozenPandaz, line 43: Can we leave out the gradle changes until we've tested it on Ocean and ensure that it's production ready?

FrozenPandaz (changes requested):
Looking good! :+1: Few more changes and we should be good to go.

packages/nx/src/tasks-runner/task-orchestrator.ts
MaxKless: bit of a leftover from when I was working there but it makes more sense to me to handle special cases first and then the general cases... I can remove it though.

MaxKless: I reverted the change anyways

FrozenPandaz (approved):

PR #26913, opened 2024-07-11
feat(gradle): change test glob to include Tests and Test

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/plugin/nodes.spec.ts
+++ b/packages/gradle/src/plugin/nodes.spec.ts
@@ -202,192 +202,233 @@ describe('@nx/gradle/plugin', () => {
     `);
   });
 
-  it('should create nodes with atomized tests targets based on gradle for nested project root', async () => {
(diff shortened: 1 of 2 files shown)


packages/gradle/src/plugin/nodes.ts
```diff
@@ -44,7 +44,8 @@ export interface GradlePluginOptions {
   testTargetName?: string;
   classesTargetName?: string;
   buildTargetName?: string;
-  [taskTargetName: string]: string | undefined;
+  excludeTestFiles?: string[];
```
FrozenPandaz, line 47: Remove this option. Change the glob so that it specifically finds `Test.java`, `Tests.java`, `Test.kt`, and `Tests.kt` files.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #26919, opened 2024-07-12
fix(core): don't add quotes to exec arguments

KaelWD:
## Current Behavior

Only a single executable + arguments can be provided, not a shell script
for example I wanted to run a command in each package based on whether certain files exist:

```sh
for i in yarn.lock pnpm-lock.yaml pnpm-workspace.yaml .npmrc; do test -e $i && git add $i; done
```

There doesn't seem to be any way to run this with nx currently, even `nx exec -- sh -c '...script...'` fails.

## Expected Behavior

you can just pass a script as a single argument and it'll run in each package

```sh
nx exec -- 'for i in yarn.lock pnpm-lock.yaml pnpm-workspace.yaml .npmrc; do test -e $i && git add $i; done'
```

## Related Issue(s)

I found #13457 after making this change, it should fix that use case too.

### DIFF
--- a/e2e/nx/src/run.test.ts
+++ b/e2e/nx/src/run.test.ts
@@ -757,6 +757,10 @@ describe('Nx Running Tests', () => {
 
       updateJson<PackageJson>('package.json', (v) => {
         v.workspaces = ['libs/*'];
+        v.scripts.compound = `nx exec -- "echo 'HELLO' && pwd"`;
+        v.scripts.shellCommand = `nx exec -- 'for i in package.json project.json; do test -e $i && echo $i; done'`;
+        v.scripts.singleQuote = `nx exec -- 'node -e console.log("HELLO")'`;
+        v.scripts.doubleQuote = `nx exec -- "node -e console.log('HELLO')"`;
         return v;
       });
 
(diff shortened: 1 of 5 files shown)


xiongemi:
this pr borke the ci test when script content is something like:
```
nx exec -- node -e \"const fs = require('fs');fs.mkdirSync('../../tmp/exec-outputs-test', {recursive: true});fs.writeFileSync('../../tmp/exec-outputs-test/file.txt', 'Outputs')\"
```

I finally figure out the issue, it is not with execSync vs spawnSync, it is about when to add quotes " around args

xiongemi (approved):

e2e/nx/src/run.test.ts
```diff
@@ -768,6 +768,11 @@ describe('Nx Running Tests', () => {
           scripts: {
             build: 'nx exec -- echo HELLO',
             'build:option': 'nx exec -- echo HELLO WITH OPTION',
+            compound: 'nx exec -- "echo HELLO && echo COMPOUND"',
+            'compound:option':
+              'nx exec -- "echo HELLO && echo COMPOUND" && echo OPTION',
```
KaelWD, line 773: I don't understand what this is supposed to be testing, `&& echo OPTION` will always be run in the root context.

KaelWD:
Ok I finally managed to get tests running locally, I see what you mean. Adding quotes to everything isn't really a solution though

If I have these two scripts: 
```
nx exec -- node -e 'console.log("HELLO")'
nx exec -- node -e "console.log('HELLO')"
```

The shell strips quotes so argv is
```
[`node`, `-e`, `console.log("HELLO")`]
[`node`, `-e`, `console.log('HELLO')`]
```

Adding double quotes breaks the first one, and single quotes breaks the second. Just doing `argv.join(' ')` leaves bare parentheses in the command which the shell will incorrectly try to interpret.  
It seems like the only option is to NOT add quotes, and require the entire script to be passed as a single string if it contains any quotes or other characters the shell might try to interpret.
```
nx exec -- "node -e 'console.log(\"HELLO\")'"
nx exec -- "node -e \"console.log('HELLO')\""
```

e2e/nx/src/run.test.ts
xiongemi: i removed this one, but added one for shell and one for node

xiongemi:
the ci failed due to there are documentation changes not checked in. could you run `pnpm documentation` and check in the changes?

xiongemi:
the changes broke the ci test like:
```
nx exec -- "node -e const fs = require('fs');fs.mkdirSync('../../tmp/exec-outputs-test', {recursive: true});fs.writeFileSync('../../tmp/exec-outputs-test/file.txt', 'Outputs')"
```

xiongemi (commented):
could you make sure it would work with node script?

xiongemi (changes requested):
could you make sure it would work with node script?

xiongemi:
i am going to close this pr for now because I don't think it would work with node script in the ci test like:
```
nx exec -- "node -e const fs = require('fs');fs.mkdirSync('../../tmp/exec-outputs-test', {recursive: true});fs.writeFileSync('../../tmp/exec-outputs-test/file.txt', 'Outputs')"
```

PR #26948, opened 2024-07-16
feat(core): ci-workflow adds workflow file to sharedGlobal inputs

jaysoo:
When we generate the workflow file, users should also add the new workflow file to the `sharedGlobals` named input, otherwise changes may land without verifying that all tasks are successful.

This is already called out in the CI tutorials, and doing it automatically improves the DX.

## Related Issue(s)

Fixes #


isaacplmann (approved):
Yes, this makes the flow nicer.

FrozenPandaz (approved):

PR #26950, opened 2024-07-16
docs(core): add docs for parallelism

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/reference/project-configuration.md
+++ b/docs/shared/reference/project-configuration.md
@@ -200,6 +200,26 @@ If you are using distributed task execution and disable caching for a given targ
 
 {% /callout %}
 
+### Parallelism
+
+In Nx 19.5.0+, tasks can be configured to support parallelism or not. By default, tasks are run in parallel with other tasks on a given machine. However, in some cases, tasks can require a shared resource such as a port or memory. For these cases, setting `"parallelism": false`, will ensure that those tasks will not run in parallel with other tasks on a single machine. For example, if the `e2e` tasks all require port 4200, running them in parallel will conflict so the targets can specify to not support parallelism:
+
+```json {% fileName="project.json" %}
+{
+  "targets": {
+    "e2e": {
+      "parallelism": false
+    }
+  }
+}
+```
+
+{% callout type="warning" title="Note: Parallelism is only per machine" %}
+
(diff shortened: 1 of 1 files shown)


docs/shared/reference/project-configuration.md
```diff
@@ -200,6 +200,27 @@ If you are using distributed task execution and disable caching for a given targ
 
 {% /callout %}
 
+### Parallelism
+
+In Nx 19.5.0 and higher, you can configure parallelism for a target by setting `"parallelism": false` in the target's configuration. This ensures that the target will not run in parallel with other tasks. For example, if you want the `e2e` target to run tests alone, you can specify:
```
FrozenPandaz, line 205:
```suggestion
In Nx 19.5.0+, you can configure parallelism for a target by setting `"parallelism": false` in the target's configuration. This ensures that the target will not run in parallel with other tasks. For example, if you do not want the `e2e` target to alongside any other tasks, you can specify:
```

docs/shared/reference/project-configuration.md
```diff
@@ -200,6 +200,27 @@ If you are using distributed task execution and disable caching for a given targ
...
+{
+  "targets": {
+    "e2e": {
+      "parallelism": false
+    }
+  }
+}
+```
+
+Parallelism is a feature in Nx that allows multiple tasks to run concurrently. By default, Nx tries to execute as many tasks in parallel as possible to optimize build and test times. However, there are scenarios where running tasks in parallel is not desirable. For example, tasks that rely on shared resources or have side effects might fail or produce inconsistent results if run concurrently.
```
FrozenPandaz, line 217:
```suggestion
Parallelism is a feature in Nx that allows multiple tasks to run in parallel. By default, Nx tries to execute tasks in parallel to optimize build and test times. However, there are scenarios where running tasks in parallel is not desirable. For example, tasks that rely on shared resources or have side effects might fail or produce inconsistent results when run alongside other tasks.
```

docs/shared/reference/project-configuration.md
```diff
@@ -200,6 +200,27 @@ If you are using distributed task execution and disable caching for a given targ
...
+
+To control parallelism, you can set the "parallelism" option in a target's configuration. This option is available starting from version 19.5.0 and can be set to either true or false.
+
+- `"parallelism": true` (default): The task can run in parallel with other tasks.
+- `"parallelism": false`: The task will run alone, without any other tasks running concurrently.
```
FrozenPandaz, line 222: Remove

FrozenPandaz (changes requested):

FrozenPandaz:
This is waiting on Nx Cloud changes

AgentEnder (approved):

PR #26971, opened 2024-07-17
Podcast page

ndcunningham:


nx-dev/data-access-documents/src/lib/blog.api.ts
```diff
@@ -20,56 +20,33 @@ export class BlogApi {
...
-    const authors = JSON.parse(
-      readFileSync(join(this.options.blogRoot, 'authors.json'), 'utf8')
+  async getPodcastBlogs(): Promise<BlogPostDataEntry[]> {
+    return sortPodcasts(
+      await this.getBlogs((post) =>
+        post.tags.map((t) => t.toLowerCase()).includes('podcast')
+      )
     );
-    const allPosts: BlogPostDataEntry[] = [];
+  }
```
bcabanes, line 29: Is there a reason Podcasts are using the "blog" api? Can we have the logic for Podcasts in a distinct api?

nx-dev/ui-podcast/src/lib/podcast-list.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { BlogAuthors } from '@nx/nx-dev/ui-blog';
+import Link from 'next/link';
+
+export interface PodcastListProps {
+  podcasts: BlogPostDataEntry[];
+}
+
+export function PodcastList({ podcasts }: PodcastListProps): JSX.Element {
+  return podcasts.length < 1 ? (
+    <></>
```
bcabanes, line 11: Can you test before calling the component? It will simplify template.

bcabanes: Or do we want to say something when there is no podcast? @juristr

bcabanes (changes requested):

nx-dev/ui-podcast/src/lib/listen-on.tsx
```diff
@@ -0,0 +1,53 @@
...
+            key={platform.name}
+            className="inline-block cursor-pointer place-items-center rounded-2xl border border-slate-100 bg-white p-4 text-slate-600 transition-all hover:scale-[1.02] hover:text-slate-950 dark:border-slate-800/60 dark:bg-slate-950  dark:text-slate-400 dark:hover:text-white"
+          >
+            <a
+              href={platform.url}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="flex h-full w-full items-center justify-center"
+            >
+              <platform.icon className="h-10 w-10 sm:h-8 sm:w-8" />
```
bcabanes, line 46:
```suggestion
              <platform.icon className="h-10 w-10 sm:h-8 sm:w-8 shrink-0" />
```

nx-dev/ui-podcast/src/lib/listen-on.tsx
```diff
@@ -0,0 +1,53 @@
...
+    },
+    {
+      name: 'Spotify',
+      url: 'https://open.spotify.com/show/6Axjn4Qh7PUWlGbNqzE7J4',
+      icon: SpotifyIcon,
+    },
+  ];
+
+  return (
+    <ul className="flex gap-6 sm:gap-3">
```
bcabanes, line 33:
```suggestion
    <ul className="flex gap-6 sm:gap-4">
```

nx-dev/ui-podcast/src/lib/podcast-list.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <Link
+              href={`/blog/${post.slug}`}
+              key={post.slug}
+              className="relative flex items-center gap-6 border-b border-slate-200 py-5 text-sm before:absolute before:inset-x-[-16px] before:inset-y-[-2px] before:z-[-1] before:rounded-xl before:bg-slate-200 before:opacity-0 last:border-0 before:hover:opacity-100 dark:border-slate-800 dark:before:bg-slate-800/50"
+              prefetch={false}
+            >
+              <span className="w-1/2 flex-none text-balance text-lg text-slate-500 sm:w-8/12 dark:text-white">
+                Episode {post.episode}: {post.title}
+              </span>
+              <span className="hidden w-2/12 flex-none sm:inline-block">
```
bcabanes, line 37: Can you your the `<time>` tag here?

bcabanes, line 37: Can you put that in a component so you only have to loop an call it? It will make it more readable.

nx-dev/data-access-documents/src/lib/blog.model.ts
```diff
@@ -9,6 +9,7 @@ export type BlogPostDataEntry = {
   reposts: string[];
   updated?: string;
   pinned?: boolean;
+  episode?: number;
```
juristr, line 12: I don't think you need a dedicated frontmatter entry here. A podcast entry should be almost just like a normal blog post and the sorting can happen via the publish date

nx-dev/data-access-documents/src/lib/blog.api.ts
juristr: Maybe we can rename it, but a podcast is basically a blog post with at the top having the voice/video included, some short description and potentially a transcript. That way when created it also appears as a blog post on the /blog (to get some more promotion automatically) and the handling is exactly the same.

Basically to simplify things in terms of data handling and authoring.

nx-dev/ui-podcast/src/lib/podcast-list.tsx
juristr: I would put something there, yes. Even though that's not gonna be the case as we're going to push this only once the first podcast edition is created, which Zack is in the process of doing. He has [a PR open](https://github.com/nrwl/nx/pull/26984), I just told him to fix a couple of things first before publishing

juristr (changes requested):

nx-dev/data-access-documents/src/lib/blog.api.ts
ndcunningham: I added a different API essentially it is the same but if we want to expand or update podcast APIs we can do it inside it's own concern

nx-dev/ui-podcast/src/lib/podcast-list.tsx
ndcunningham: Added an empty state, aka Stay tuned. @juristr lmk if the wording is fine and we can leave it as is.

bcabanes (approved):

jaysoo (approved):

PR #26976, opened 2024-07-18
feat(js): added a verdaccio listen address to configuration

Vegtory:
## Current Behavior
Running Verdaccio through nx always uses localhost by default to listen to.
The Verdaccio config.yaml listen address is always overridden by the executor because it sets the --listen argument.

## Expected Behavior
In our situation, where we are developing in a VM we want Verdaccio to listen to all requests by setting it to 0.0.0.0 instead of localhost.

## Related Issue(s)
No Issue registered

Fixes #
Added a listen address to the Verdaccio configuration and executor.
Decided to make it a simple implementation to prevent breaking changes.

### DIFF
--- a/docs/generated/packages/js/executors/verdaccio.json
+++ b/docs/generated/packages/js/executors/verdaccio.json
@@ -23,6 +23,10 @@
         "type": "number",
         "description": "Port of local registry that Verdaccio should listen to"
       },
+      "listenAddress": {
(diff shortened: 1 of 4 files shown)


Wixewr:
I could not get documentation to work, maybe one of the maintainers would be so kind to run it for this PR?

AgentEnder (approved):

PR #26997, opened 2024-07-18
Add filters to blog page

ndcunningham:


nx-dev/ui-blog/src/lib/more-blogs.tsx
```diff
@@ -1,11 +1,58 @@
...
+  AcademicCapIcon,
+  ChatBubbleOvalLeftEllipsisIcon,
+  XCircleIcon,
+} from '@heroicons/react/24/outline';
 
 export interface MoreBlogsProps {
   blogs: BlogPostDataEntry[];
 }
 
+const filters = [
```
juristr, line 18: should we deduplicate this one with the one in the `blog-container.tsx`? Otherwise if we add support for a new topic we might miss to update either of the two.

nx-dev/ui-blog/src/lib/filters.tsx
```diff
@@ -0,0 +1,140 @@
...
+            </Link>
+          </li>
+        ))}
+      </ul>
+
+      {/* MOBILE */}
+      <div className="relative lg:hidden">
+        <Menu as="div" className="inline-block text-left">
+          <Menu.Button className="inline-flex w-full justify-center rounded-md border border-slate-400 bg-white px-4 py-2 text-sm font-medium text-slate-700 shadow-sm transition hover:bg-slate-50 hover:text-slate-600 dark:border-slate-700 dark:bg-slate-800 dark:text-slate-200 dark:hover:bg-slate-700">
+            Filters
```
juristr, line 101: Minor: but can we call it "Topics"?  sounds better than filters

juristr (changes requested):

juristr (approved):

PR #19611, opened 2023-10-13
feat(core): add support for wildcards in dependsOn

fxposter:
Now it is possible to define targets like this:

```
{
  "targets": {
    "build-css": {},
    "build-js": {},
    "test": {
      "dependsOn": ["build-*"]
    },
  }
}
```

## Current Behavior
No support for wildcard target dependencies.

## Expected Behavior
This PR is an example of what I described here: https://github.com/nrwl/nx/issues/19414.

## Related Issue(s)
Closes #19414

PR #27029, opened 2024-07-21
docs(core): document wildcard support for dependsOn field

elevenpassin:
## Current Behavior

This feature is not yet documented...

## Expected Behavior

Document wildcard support in `dependsOn` field for project configurations.

Note: Reading through the PR below that impls. this wildcard support, I couldn't quite figure out if it was actually able to also work in case of object syntax and whether this will also work for `targetDefaults`... so if you know any better, please lemme know so I can update this PR 🙏 

## Related Issue(s)

documents #19611

### DIFF
--- a/docs/shared/reference/project-configuration.md
+++ b/docs/shared/reference/project-configuration.md
@@ -434,7 +434,7 @@ Starting from v19.5.0, wildcards can be used to define dependencies in the `depe
         "build-*", // support for using wildcards in dependsOn, matches: "build-css", "build-js" targets of dependencies
         "*build-*" // matches tasks: "build-css", "build-js" as well as "task-with-build-in-middle" targets of dependencies.
       ]
-    },
+    }
   }
 }
 ```
--- a/docs/shared/reference/project-configuration.md
+++ b/docs/shared/reference/project-configuration.md
@@ -431,13 +431,10 @@ Starting from v19.5.0, wildcards can be used to define dependencies in the `depe
           "target": "build", // target name
           "params": "ignore" // "forward" or "ignore", defaults to "ignore"
         },
-        "build-*", // support for using wildcards in dependsOn, matches: build-css, build-js and also then-build-something-else
-        "^build-*" // matches only build-css, build-js
+        "build-*", // support for using wildcards in dependsOn, matches: "build-css", "build-js" targets of dependencies
+        "*build-*" // matches tasks: "build-css", "build-js" as well as "task-with-build-in-middle" targets of dependencies.
       ]
     },
-    "build-css": {},
-    "build-js": {},
-    "then-build-something-else": {}
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/shared/reference/project-configuration.md
```diff
@@ -420,6 +420,28 @@ You can also express task dependencies with an object syntax:
...
+```json
+{
+  "targets": {
+    "test": {
+      "dependsOn": [
+        {
+          "target": "build", // target name
+          "params": "ignore" // "forward" or "ignore", defaults to "ignore"
+        },
+        "build-*", // support for using wildcards in dependsOn, matches: build-css, build-js and also then-build-something-else
```
AgentEnder, line 434: This wouldn't match then-build-something-else, `^{pattern}` means "apply this pattern to targets on my dependencies", not "target name starts with pattern". These are glob patterns, not regex.

AgentEnder (changes requested):
Thanks for the contribution, just one small change.

docs/shared/reference/project-configuration.md
elevenpassin: I don't have a ^symbol in front of the `build-*` pattern though 🤔  unless you are saying that `{pattern}` also means "target starts with pattern"

AgentEnder: Yeah, kind of. Wildcards are needed to match anything thats not explicit. See: https://www.glob101.com/view/t2Tzx5SUwo6SZkHO3bTi

isaacplmann: `*build-*` would find all three tasks.

elevenpassin: Ah! Apologies, I got myself confused 😮 I will update the PR now.

elevenpassin: Updated: https://github.com/nrwl/nx/pull/27029/commits/90fc87e7cc02f928db1fec1bb3e5c4c12b7bca38

docs/shared/reference/project-configuration.md
```diff
@@ -420,6 +420,28 @@ You can also express task dependencies with an object syntax:
...
+        "^build-*" // matches only build-css, build-js
```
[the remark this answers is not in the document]
elevenpassin: Updated: https://github.com/nrwl/nx/pull/27029/commits/90fc87e7cc02f928db1fec1bb3e5c4c12b7bca38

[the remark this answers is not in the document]
isaacplmann: I think there's still a misunderstanding in your latest update.

`build-*` will match `build-css`, `build-js` on the same project
`^build-*` will match `build-css`, `build-js` on project dependencies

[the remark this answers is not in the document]
elevenpassin: Updated: d09e3326e (will be pushed as soon as the push commit hook passed).

isaacplmann (approved):

AgentEnder (approved):

PR #27067, opened 2024-07-23
fix(core): fix watch daemon error

FrozenPandaz:
## Current Behavior

`nx watch` throws an error saying the daemon is not running when the daemon is indeed running.

## Expected Behavior

`nx watch` does not throw an error saying the daemon is not running when the daemon is indeed running.

## Related Issue(s)

Fixes #


jaysoo (approved):

Cammisuli (approved):

trhinehart-godaddy:
is this live? running into this issue w/ latest lerna

PR #27083, opened 2024-07-24
fix(linter): ensure too many warnings is only logged if enabled

skrtheboss:
## Current Behavior

ESLint found too many warnings (maximum: -1).

## Expected Behavior

Nothing should be logged if maximum is set to -1
## Related Issue(s)

Fixes #


leosvelperez (approved):
Thanks for the contribution!

PR #27092, opened 2024-07-24
fix(bundling): get workspace package prefix length correctly #20817

### LINKED ISSUE #20817: esbuild - paths from tsconfig.base.json do not work properly in the build output
### Current Behavior

When using imports that have similar paths, esbuild will not output a manifest in main.js that properly allows importing all libs if they are ordered in certain orders in the tsconfig.base.json

For example:

The following paths in tsconfig.base.json will not order correctly in the manifest generated for here: https://github.com/nrwl/nx/blob/32baa4dab1a254cb5da4fcb922dc2ef0fa2bc025/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts#L246

```
"path": {
      "@worklabs/db": ["libs/db/core/src/index.ts"],
      "@worklabs/db/migrations": ["libs/db/migrations/src/index.ts"],
      "@worklabs/db/models": ["libs/db/models/src/index.ts"],
}
```

Because @worklabs/db was first in paths, any imports when going through the hijacked _resolveFilename that were @worklabs/db/migrations or @worklabs/db/models were directed to @worklabs/db files which did not work 

Interestingly, ts in VSCode found no issues with the order of paths. Nor did serve using @nx/js:node 

### Expected Behavior

If I have the above entries, they should be sorted so that the longer ones are first and do not mistakenly get matched to the wrong import 

Locally, I was able to fix this behaviour by editing `getRegisterFileContent` function to sort by whole string length first like so

```
const sortedKeys = Object.keys(paths)
      .sort((a, b) => b.length - a.length)
      .sort((a, b) => getPrefixLength(b) - getPrefixLength(a));
```

### GitHub Repo

_No response_

### Steps to Reproduce

1. Add libs that use subpaths, adding the longer ones secondary 
2. Try build with esbuild 
3. Try use the build with the outputs

### Nx Report

```shell
Convert compiler options from json failed, No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["node_modules/jest-extended/types/index.d.ts"]' and 'exclude' paths were '["node_modules","dist","build","coverage"]'.

   Node   : 20.10.0
   OS     : darwin-arm64
   pnpm   : 8.11.0
   
   nx (global)        : 17.2.0
   nx                 : v17.2.5
   @nx/js             : v17.2.5
   @nx/jest           : 17.2.0
   @nx/linter         : v17.2.5
   @nx/eslint         : v17.2.5
   @nx/workspace      : v17.2.5
   @nx/cypress        : v17.2.5
   @nx/devkit         : v17.2.5
   @nx/esbuild        : 17.2.0
   @nx/eslint-plugin  : v17.2.5
   @nx/node           : v17.2.5
   @nx/react          : v17.2.5
   @nx/storybook      : v17.2.5
   @nrwl/tao          : v17.2.5
   @nx/vite           : v17.2.5
   @nx/web            : v17.2.5
   @nx/webpack        : 17.2.0
   typescript         : 5.3.3
     - @nx/jest@17.2.0
     - @nrwl/jest@17.2.0
     - @nx/esbuild@17.2.0
     - @nrwl/esbuild@17.2.0
     - @nx/webpack@17.2.0
     - @nrwl/webpack@17.2.0
   
   To fix this, run `nx migrate nx@v17.2.5`
```

### Failure Logs

```shell
Depending on the module we get different errors:

If the output is valid, but the contents are different we get something like: 

migrations.runMigrations is not a function

Otherwise we get errors about reading a property of undefined
```

### Package Manager Version

8.11.0

### Operating System

- [X] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

We picked this up while migrating to NX for "almost-monorepo" (code was all in one repo, but structure and tooling was lagging), so this wasn't something that was working before and now doesn't

Coly010:
## Current Behavior

When there are workspace libraries with the following import paths:

```
@org/core
@org/core/db
@org/core/acme
```

They need to be sorted for the manifest such that matching finds the most specific first.
The logic for this is currently based off whether an `*` exists, which doesn't work when the paths are specific.

## Expected Behavior

Try to use `*` first and fallback to `/` to determine package prefix length.

## Related Issue(s)

Fixes #20817

### DIFF
--- a/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
+++ b/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
@@ -291,7 +291,16 @@ require('${mainFile}');
 }
 
 function getPrefixLength(pattern: string): number {
-  return pattern.substring(0, pattern.indexOf('*')).length;
+  const prefixIfWildcard = pattern.substring(0, pattern.indexOf('*')).length;
+  const prefixWithoutWildcard = pattern.substring(
+    0,
+    pattern.lastIndexOf('/')
+  ).length;
+  // if the pattern doesn't contain '*', then the length is always 0
+  // This causes issues when there are sub packages such as
+  // @nx/core
+  // @nx/core/testing
+  return prefixIfWildcard || prefixWithoutWildcard;
 }
 
 function getTsConfigCompilerPaths(context: ExecutorContext): {


packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -291,7 +291,16 @@ require('${mainFile}');
...
+  const prefixIfWildcard = pattern.substring(0, pattern.indexOf('*')).length;
+  const prefixWithoutWildcard = pattern.substring(
+    0,
+    pattern.lastIndexOf('/')
+  ).length;
+  // if the pattern doesn't contain '*', then the length is always 0
+  // This causes issues when there are sub packages such as
+  // @nx/core
+  // @nx/core/testing
+  return prefixIfWildcard || prefixWithoutWildcard;
```
jaysoo, line 303: I would just inline `prefixIfWildcard` and `prefixWithoutWildcard`, and it'll short-circuit the latter if previous is > 0.

jaysoo (approved):

PR #27113, opened 2024-07-25
fix(testing): ensure project build occurs before e2es are run

Coly010:
- feat(devkit): add util for adding target default to nx.json
- feat(testing): setup targetDefault for ciTargetName when @nx/cypress/plugin is added
- feat(testing): add migration to add targetDefault for ciTargetName for @nx/cypress/plugin
- feat(testing): add targetDefault for ciTargetName when setting up @nx/playwright/plugin

## Current Behavior

Cypress and Playwright e2e targets eventually run the build of the application within the serve-static target that they have. However, the e2e tasks themselves do not dependOn build so not only is it possible that the e2e tasks end up building from scratch, they could also run first before build tasks.

## Expected Behavior

Ensure the build is run before the e2e tasks

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/cypress/src/migrations/update-19-5-4/add-ci-target-name-target-default.ts
+++ b/packages/cypress/src/migrations/update-19-5-4/add-ci-target-name-target-default.ts
@@ -1,0 +1,41 @@
+import {
+  type ExpandedPluginConfiguration,
+  type Tree,
+  readNxJson,
+  formatFiles,
+} from '@nx/devkit';
+import { addTargetDefault } from '@nx/devkit/src/generators/target-defaults-utils';
+import type { CypressPluginOptions } from '../../plugins/plugin';
+
+export default async function (tree: Tree) {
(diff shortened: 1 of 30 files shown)


packages/cypress/src/migrations/update-19-5-4/add-ci-target-name-target-default.ts
```diff
@@ -0,0 +1,41 @@
...
+    return;
+  }
+
+  const cypressPlugin = nxJson.plugins.find((p) =>
+    typeof p === 'string'
+      ? p === '@nx/cypress/plugin'
+      : p.plugin === '@nx/cypress/plugin'
+  );
+
+  if (!cypressPlugin) {
```
FrozenPandaz, line 22: Can we also check for the webpack/vite/next/remix and whichever other plugins add the serve-static target? :eyes:

FrozenPandaz (changes requested):

PR #27140, opened 2024-07-26
docs(react): change word "Angular" to "React"

EasonLin0716:
"Hence we can easily import them into other libraries and our Angular application. As an example, let's use the pre-generated ProductsComponent component from our libs/products library."

It's a tutorial providing a React project example, I think "Angular" is typing wrong

## Current Behavior

Hence we can easily import them into other libraries and our **Angular** application. As an example, let's use the pre-generated ProductsComponent component from our libs/products library.

## Expected Behavior

Hence we can easily import them into other libraries and our **React** application. As an example, let's use the pre-generated ProductsComponent component from our libs/products library.

### DIFF
--- a/docs/shared/tutorials/react-monorepo.md
+++ b/docs/shared/tutorials/react-monorepo.md
@@ -430,7 +430,7 @@ All libraries that we generate automatically have aliases created in the root-le
 }
 ```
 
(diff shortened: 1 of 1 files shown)


ndcunningham (approved):
LGTM, thank you!

PR #27157, opened 2024-07-26
fix(nx-dev): add background on pricing tiles & blog link to pricing page #27157

ndcunningham:


bcabanes (approved):

isaacplmann:
The pipeline failure is a problem with the link checker script, not an actual broken link.  I'm working on fixing the script now.

isaacplmann:
I updated the link checker in this PR instead

isaacplmann (approved):

ndcunningham:
> I updated the link checker in this PR instead

@isaacplmann 
It looks like when we define a navigation to any page that is inside `nx-dev/app/` with a  `#` included it will cause the checker to fail. Is it possible to improve this in another PR without ignoring the routes?

isaacplmann:
Yes, definitely possible to make the link checker smarter.  We used to not have very many hand-coded pages, so it wasn't as important.  Now that nx.app pages are merged in, it matters more.

PR #27159, opened 2024-07-26
fix(webpack): include hash in asset filenames so they do not conflict

### LINKED ISSUE #18272: Compilation error (the same filename for multiple chunks) when import an image twice with the same filename
### Current Behavior

When I try to serve my project, I get this error
```
[webpack-dev-middleware] Error: Conflict: Multiple chunks emit assets to the same filename preview.png (chunks main and main)
```

### Expected Behavior

I should get no error; images should have different names.

### GitHub Repo

https://github.com/shlajin/nx-multiple-chunks-bug-demo

### Steps to Reproduce

I started with a brand new repo to illustrate the problem.

```
npx create-nx-workspace --pm yarn
Need to install the following packages:
  create-nx-workspace@16.5.5
Ok to proceed? (y) y

 >  NX   Let's create a new workspace [https://nx.dev/getting-started/intro]

✔ Where would you like to create your workspace? · multiple-chunks-bug-demo
✔ Which stack do you want to use? · react
✔ What framework would you like to use? · none
✔ Standalone project or integrated monorepo? · integrated
✔ Application name · multiple-chunks-bug-demo
✔ Which bundler would you like to use? · webpack
✔ Default stylesheet format · css
✔ Enable distributed caching to make your CI faster · No

 >  NX   Creating your v16.5.5 workspace.
```

I added a library (no tests, not buildable, not publishable).
```
yarn nx g @nx/react:lib sample-library
```

I added 2 different images into the following folders:
```
libs/sample-library/src/lib/apple/preview.png
libs/sample-library/src/lib/orange/preview.png
```

And added 2 files next to each of the images that simply import and re-export the image:

```
// libs/sample-library/src/lib/orange/Orange.ts
import Preview from './preview.png';

export const orangeData = {
  image: Preview,
};


// libs/sample-library/src/lib/apple/Apple.ts
import Preview from './preview.png';

export const appleData = {
  image: Preview,
};
```

Then, in `libs/sample-library/src/index.ts` I re-exported my 2 files:
```
import { appleData } from "./lib/apple/Apple"
import { orangeData } from "./lib/orange/Orange"

export const data = {
  appleData,
  orangeData
}
```

To recap, I have 2 folders named `apple` and `orange`, each folder contains a `preview.png` file and one .ts file each. The *.ts file simply re-exports the image (obtains the image URL as per default behaviour), and the `index.ts` of the whole library re-exports the *.ts files.

Now if I try to use the images in the app, e.g.

```
// apps/multiple-chunks-bug-demo/src/app/app.tsx
import { data } from "@multiple-chunks-bug-demo/sample-library"

export function App() {
  return (
    <div>
      <h1>DEMO</h1>
      <img src={data.appleData.image} />
      <img src={data.orangeData.image} />
    </div>
  );
}

export default App;
```

and run

```
yarn nx serve multiple-chunks-bug-demo
```

I get

```
[webpack-dev-middleware] Error: Conflict: Multiple chunks emit assets to the same filename preview.png (chunks main and main)
```

I attached a demo repo for your convenience.

Thank you!

### Nx Report

```shell
Node   : 18.14.0
   OS     : darwin-arm64
   yarn   : 1.22.19

   nx                 : 16.5.5
   @nx/js             : 16.5.5
   @nx/jest           : 16.5.5
   @nx/linter         : 16.5.5
   @nx/workspace      : 16.5.5
   @nx/cypress        : 16.5.5
   @nx/devkit         : 16.5.5
   @nx/eslint-plugin  : 16.5.5
   @nx/react          : 16.5.5
   @nrwl/tao          : 16.5.5
   @nx/web            : 16.5.5
   @nx/webpack        : 16.5.5
   typescript         : 5.1.6

✨  Done in 0.83s.
```

### Failure Logs

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

jaysoo:
Our assets are generated as flat assets in dist, which allows using assets from workspace libs. This prevents users from having different assets with the same filename (e.g. `foo/image.png` and `bar/image.png`). This will error out in the dev server with conflicting filenames.

We cannot use `[path][name]` because of assets that are outside of the app folder (e.g. `../../libs/ui/src/assets/image.png`). Thus the best option is to include hash.

Note: Also re-enabled the e2e tests for `react.test.ts` file since it is now using Playwright instead of Cypress.

## Current Behavior
Assets with the same filename will error in dev-mode.

## Expected Behavior
Assets with the same filename works.

## Related Issue(s)

Fixes #18272

### DIFF
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-web-config.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-web-config.ts
@@ -332,6 +332,7 @@ export function applyWebConfig(
 
   config.output = {
     ...config.output,
+    assetModuleFilename: '[name].[contenthash:20][ext]',
     crossOriginLoading: options.subresourceIntegrity
       ? ('anonymous' as const)
       : (false as const),
@@ -404,9 +405,6 @@ export function applyWebConfig(
             maxSize: 10_000, // 10 kB
           },
         },
-        generator: {
-          filename: `[name]${hashFormat.file}[ext]`,
-        },
       },
       // SVG: same as image but we need to separate it so it can be swapped for SVGR in the React plugin.
       {
@@ -417,17 +415,11 @@ export function applyWebConfig(
             maxSize: 10_000, // 10 kB
           },
         },
-        generator: {
-          filename: `[name]${hashFormat.file}[ext]`,
-        },
       },
       // Fonts: Emit separate file and export the URL.
       {
         test: /\.(eot|otf|ttf|woff|woff2)$/,
         type: 'asset/resource',
-        generator: {
-          filename: `[name]${hashFormat.file}[ext]`,
-        },
       },
       ...rules,
     ],
--- a/e2e/react/src/react.test.ts
+++ b/e2e/react/src/react.test.ts
@@ -68,6 +68,8 @@ describe('React Applications', () => {
       const libName = uniq('lib');
       const libWithNoComponents = uniq('lib');
       const logoSvg = readFileSync(join(__dirname, 'logo.svg')).toString();
+      const blueSvg = `<?xml version="1.0"?><svg xmlns="http://www.w3.org/2000/svg" version="1.2" baseProfile="tiny" viewBox="0 0 30 30"><rect x="10" y="10" width="10" height="10" fill="blue"/></svg>`;
+      const redSvg = `<?xml version="1.0"?><svg xmlns="http://www.w3.org/2000/svg" version="1.2" baseProfile="tiny" viewBox="0 0 30 30"><rect x="10" y="10" width="10" height="10" fill="red"/></svg>`;
 
       runCLI(
         `generate @nx/react:app ${appName} --style=css --bundler=webpack --no-interactive --skipFormat`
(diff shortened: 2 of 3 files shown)


packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-web-config.ts
```diff
@@ -332,6 +332,7 @@ export function applyWebConfig(
 
   config.output = {
     ...config.output,
+    assetModuleFilename: '[name].[contenthash][ext]',
```
ndcunningham, line 335: I think this needs to be `hash` at least in development mode for HMR to work. Maybe this needs to be a ternary assetModuleFilename: ('mode === 'production' ? '[name].[contenthash][ext]' : '[name].[hash:8][ext]' )
or we update the `getOutputHashFormat` use that.

> Hash: This is a hash generated every time that webpack compiles, in dev mode this is good for cache busting during development but shouldn't be used for long term caching of your files. This will overwrite the Hash on every build of your project.

ndcunningham (approved):

PR #27188, opened 2024-07-29
feat(release): allow {releaseGroupName} to be interpolated in releaseTagPattern

JamesHenry:
## Current Behavior

There is no way to interpolate the release group name in the `releaseTagPattern`, so for workspaces with a lot of groups, a lot of repetition is required.

## Expected Behavior

`{releaseGroupName}` will now get interpolated in the tag pattern for release groups with fixed versioning (independent projects are still based on the `{projectName}` as before)

## Related Issue(s)

Fixes #


JamesHenry:
Accidentally merged before checks and reviews could complete, immediately reverted in #27189

PR #27195, opened 2024-07-29
fix(nx-dev): OG image meta tag

ndcunningham:
This PR defaults the OG image if one of the correct file types is not provided.

### DIFF
--- a/nx-dev/data-access-documents/src/lib/blog.api.ts
+++ b/nx-dev/data-access-documents/src/lib/blog.api.ts
@@ -1,5 +1,5 @@
-import { readFileSync, readdirSync } from 'fs';
-import { join, basename } from 'path';
+import { readFileSync, readdirSync, accessSync, constants } from 'fs';
+import { join, basename, parse, resolve } from 'path';
 import { extractFrontmatter } from '@nx/nx-dev/ui-markdoc';
 import { sortPosts } from './blog.util';
 import { BlogPostDataEntry } from './blog.model';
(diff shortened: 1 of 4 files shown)


nx-dev/data-access-documents/src/lib/blog.api.ts
```diff
@@ -143,4 +150,50 @@ export class BlogApi {
...
+
+      if (!foundExt) {
+        return { image: defaultImage, type: defaultType };
+      }
+
+      return {
+        image: imagePath.replace(ext, foundExt),
+        type: foundExt.replace('.', ''),
+      };
+    }
```
juristr, line 196: nit: for code readability I prefer an `else` here to see the two options

nx-dev/data-access-documents/src/lib/blog.api.ts
```diff
@@ -143,4 +150,50 @@ export class BlogApi {
...
+    const defaultImage = 'https://nx.dev/socials/nx-media.png';
+    const defaultType = 'png';
+
+    if (!imagePath) {
+      return { image: defaultImage, type: defaultType };
+    }
+    const { ext } = parse(imagePath);
+
+    if (!allowedExtensions.includes(ext)) {
+      const foundExt = allowedExtensions.find((allowedExt) => {
```
juristr, line 177: I'd love if this at some point would automatically run a conversion. But this def works for now

juristr (approved):

PR #27205, opened 2024-07-30
fix(nx-dev): clean up heading text

mandarini:
Fixing 3 issues when storing the page sections:

1.  The heading of a node can be like this: `title: 'Angular Monorepo Tutorial - Part 1: Code Generation' description: In this tutorial you'll create a frontend-focused workspace with Nx`. We only want to keep the `title` part.
2. Some headings have `{% highlightColor="green" %}`. We want to remove that from the heading of the page section (and the link text)
3. There's an [error in `tags.json` where the path is incorrect](https://github.com/nrwl/nx/blob/master/docs/generated/manifests/tags.json#L126). We want to fix that on the fly, so that the embeddings script finds the correct pages.

### DIFF
--- a/tools/documentation/create-embeddings/src/main.mts
+++ b/tools/documentation/create-embeddings/src/main.mts
@@ -83,7 +83,9 @@ export function processMdxForSearch(content: string): ProcessedMdx {
     const [firstNode] = tree.children;
 
     const heading =
-      firstNode.type === 'heading' ? toString(firstNode) : undefined;
+      firstNode.type === 'heading'
+        ? removeTitleDescriptionFromHeading(toString(firstNode))
+        : undefined;
     const slug = heading ? slugger.slug(heading) : undefined;
 
     return {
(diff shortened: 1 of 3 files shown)


tools/documentation/create-embeddings/src/main.mts
```diff
@@ -420,6 +423,15 @@ function getAllFilesWithItemList(data): WalkEntry[] {
...
         // the path is the relative path to the file within the nx repo
         // the url_partial is the relative path to the file within the docs site - under nx.dev
+
+        // there's an error in `tags.json` where the path is incorrect
+        if (item.file.includes('generated/packages/generated/packages')) {
+          item.file = item.file.replace(
+            'generated/packages/generated/packages',
+            'generated/packages'
+          );
+        }
```
FrozenPandaz, line 433: Can this be fixed in `tags.json`?

FrozenPandaz (changes requested):

tools/documentation/create-embeddings/src/main.mts
mandarini: Yes, I'll do that!

isaacplmann (approved):

FrozenPandaz (approved):

PR #27210, opened 2024-07-30
fix(core): disable plugin isolation by default on windows for now

MaxKless:
it's causing issues right now so we need time to fix it.


AgentEnder (approved):
😞

PR #27217, opened 2024-07-30
feat(misc): add onboarding a/b testing

FrozenPandaz:
## Current Behavior

The onboarding message during `create-nx-workspace` is not ideal.

## Expected Behavior

We are A/b testing some options for the `create-nx-workspace` onboarding message.

## Related Issue(s)

Fixes #


packages/create-nx-workspace/src/create-workspace.ts
```diff
@@ -51,31 +51,27 @@ export async function createWorkspace<T extends CreateWorkspaceOptions>(
...
-      await setupCI(
-        directory,
-        nxCloud,
-        packageManager,
-        nxCloudInstallRes?.code === 0
-      );
+      await setupCI(directory, nxCloud, packageManager);
     }
+
+    const { connectCloudUrl, output } = await getOnboardingInfo(
```
ndcunningham, line 63: should we short-circuit this if the `token` is invalid?
Like for example not having an internet connection?

FrozenPandaz: If there's an error getting the token because of internet issues, then the process will exit 1 without reaching this part.

ndcunningham (approved):

PR #27221, opened 2024-07-31
chore(repo): improve the `devcontainer` setup

tinesoft:
- respect `pnpm` version declared in root `package.json`
- improve performances when installing node modules
- add troubleshooting instructions in `CONTRIBUTING.md` to help solve common issue related to outdated `GLIBC` version

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,0 @@
-FROM mcr.microsoft.com/devcontainers/typescript-node:20-bullseye
-
-# Update the underlying (Debian) OS, to make sure we have the latest security patches and libraries like 'GLIBC' 
-RUN sudo apt-get update  && sudo apt-get -y upgrade
-
-# Update pnpm
-RUN npm install -g pnpm@8.15.7
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,31 +3,46 @@
 {
   "name": "NxDevContainer",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "build": {
-    // Path is relative to the devcontainer.json file.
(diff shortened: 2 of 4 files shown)


tinesoft:
From what I have understood, `GLIBC` is used to compile native stuff used by Nx ( i.e `rust` related code), so I would say that the version of the compiler library need to evolve along with the version of rust Rust language and  features that are used in the project.

When I initially setup the devcontainer, `v2.31` of `GLIBC` was needed, now it is  `v2.33` at minimum

tinesoft:
also I think the `sudo apt-get update` performed whenever the devcontainer is created (see `postCreateCommand.sh`) will take care  of automatically updating the `glibc` binaries,  but  that would only be for minor/fix versions (this has already worked before when I first created the devcontainer setup)

As for automating the entire thing starting from the base image that is used by the devcontainer, maybe we can create a small script that would extract it  and run the underlying `docker` command to build/mount the container and then `nx build` command as part of an e2e nightly test...

That way we can spot when the image needs to be updated.

I can look into that and provide another PR if the idea is validated?

tinesoft:
Any progress on reviewing this ?  @FrozenPandaz  @vsavkin

AgentEnder (approved):

isaacplmann (approved):

tinesoft:
Thanks for approving/merging this.

Another thing that we can do, to optimize/improve the DX when opening the repo in `Github Codespaces`, is to activate the [pre-built image of the underlying devcontainer](https://github.com/nrwl/nx/settings/codespaces), so that future contributors can get started quickly

image

Then 
image

tinesoft:
> **Isaac**: How often does the version of glibc get updated? Is there a way to automatically use the latest version?

@isaacplmann  do you want me to investigate how to automate the update of the devcontainer image as [described here](https://github.com/nrwl/nx/pull/27221#issuecomment-2263445879) ?

PR #27240, opened 2024-07-31
fix(testing): e2e-ci should use serve-static or vite preview for playwright and cypress

Coly010:
- fix(vite): preview should dependOn build
- fix(react): playwright should use vite preview
- fix(vue): playwright should use vite preview
- fix(web): playwright should use vite preview
- chore(testing): add e2e test

## Current Behavior

Currently, `playwright` uses the `vite serve` command when setting up the web server to run the e2e tests against.

The `vite preview` command/target should also depend on `vite build`.

## Expected Behavior

`playwright` should use the `vite preview` command when setting up the web server

`vite preview` targets add a `dependsOn["build"]`

Ensure `serve-static` has a dependsOn: ['build']

Cypress should use the `ciBaseUrl` if it exists when running the `e2e-ci` targets

Migrations for Playwright and Cypress to use serve-static and preview correctly

## Related Issue(s)

Fixes #


packages/react/src/generators/application/lib/add-e2e.ts
```diff
@@ -83,9 +83,9 @@ export async function addE2e(
         js: false,
         linter: options.linter,
         setParserOptionsProject: options.setParserOptionsProject,
-        webServerCommand: `${getPackageManagerCommand().exec} nx ${
-          options.e2eWebServerTarget
-        } ${options.name}`,
+        webServerCommand: `${getPackageManagerCommand().exec} nx run ${
```
FrozenPandaz, line 86: Also make changes for Cypress

FrozenPandaz (changes requested):

Coly010:
Playwright can also use ‘serve-static’ instead of ‘serve’

FrozenPandaz (approved):

packages/cypress/migrations.json
```diff
@@ -29,6 +29,12 @@
       "version": "18.1.0-beta.3",
       "description": "Update to Cypress ^13.6.6 if the workspace is using Cypress v13 to ensure workspaces don't use v13.6.5 which has an issue when verifying Cypress.",
       "implementation": "./src/migrations/update-18-1-0/update-cypress-version-13-6-6"
+    },
+    "update-19-5-5-update-ci-webserver-for-vite": {
+      "cli": "nx",
+      "version": "19.5.5-beta.0",
```
FrozenPandaz, line 35: Lets' schedule this for 19.6.0

FrozenPandaz (approved):

e2e/webpack/src/webpack.legacy.test.ts
```diff
@@ -2,6 +2,7 @@ import {
   checkFilesExist,
   cleanupProject,
   killProcessAndPorts,
+  listFiles,
```
FrozenPandaz, line 5: This doesn't look used.

e2e/webpack/src/webpack.legacy.test.ts
```diff
@@ -113,17 +114,17 @@ describe('Webpack Plugin (legacy)', () => {
...
     );
     updateFile(
       `${appName}/webpack.config.js`,
       `
       const { join } = require('path');
-        const {NxWebpackPlugin} = require('@nx/webpack');
+        const {NxAppWebpackPlugin} = require('@nx/webpack/app-plugin');
         module.exports = {
           output: {
-            path: join(__dirname, '../dist/app9524918'),
```
FrozenPandaz, line 126: That's weird.. so it always outputted to this random directory?

e2e/webpack/src/webpack.legacy.test.ts
```diff
@@ -113,17 +114,17 @@ describe('Webpack Plugin (legacy)', () => {
     updateFile(
       `${appName}/src/main.ts`,
       `
-      document.querySelector('proj-root').innerHTML = '<h1>Welcome</h1>';
+      document.querySelector('proj-root')!.innerHTML = '<h1>Welcome</h1>';
```
FrozenPandaz, line 117: Why is this `!` necessary?

FrozenPandaz (approved):

e2e/webpack/src/webpack.legacy.test.ts
Coly010: ah debug leftover

e2e/webpack/src/webpack.legacy.test.ts
Coly010: TS type checking was finding that querySelector could possibly be null. Which is valid

e2e/webpack/src/webpack.legacy.test.ts
Coly010: Yep. Which was kinda fine before because we weren't _checking_ the dist folder afterwards, just that the build worked, but with the `e2e` now using static serve, the file-server couldn't find the correct files because the dist folder output didnt match what it was expecting

PR #27266, opened 2024-08-01
fix(core): move token generation into new

FrozenPandaz:
## Current Behavior

The Nx Cloud token is generated after the workspace and preset have been generated. This makes it hard to use the Nx Cloud token for generating the workspace and preset.

## Expected Behavior

The Nx Cloud token is generated in the workspace creation. This makes it easier to use the Nx Cloud token during the workspace creation process.

## Related Issue(s)

Fixes #


ndcunningham (approved):
LGTM

PR #27311, opened 2024-08-06
feat(angular): add vitest option to angular

yjaaidi:
- This adds `vitest` option to `unitTestRunner` for Angular generators.
- This **does not** add vitest option to `create-nx-workspace` but I think we should provide this as an option in the future. Please let me know if you wantme to add this here or as a future feature.


packages/vite/src/utils/generator-utils.ts
```diff
@@ -347,6 +348,8 @@ export interface ViteConfigFileOptions {
   imports?: string[];
   plugins?: string[];
   coverageProvider?: 'v8' | 'istanbul' | 'custom';
+  setupFile?: string;
```
yjaaidi, line 351: I wonder if we really need to pass the `setupFile` here as the path is already hardcoded in some places. e.g. https://github.com/nrwl/nx/blob/6cc833013666bf62afde35871bb6d1871b39b886/packages/vite/src/generators/init/init.ts#L25

packages/vite/src/utils/generator-utils.ts
```diff
@@ -356,11 +359,15 @@ export function createOrEditViteConfig(
...
   vitestFileName?: boolean
 ) {
-  const { root: projectRoot } = readProjectConfiguration(tree, options.project);
+  const { root: projectRoot, projectType } = readProjectConfiguration(
+    tree,
+    options.project
+  );
 
+  const extension = options.useEsmExtension ? 'mts' : 'ts';
   const viteConfigPath = vitestFileName
```
yjaaidi, line 368: Vite commonjs support is deprecated and angular vitest plugin supports ESM-only, that is why we have to use `.mts` if package type is not `module`.

packages/angular/src/generators/utils/add-vitest.ts
```diff
@@ -0,0 +1,88 @@
...
+  skipPackageJson: boolean;
+  strict: boolean;
+};
+
+export async function addVitest(
+  tree: Tree,
+  options: AddVitestOptions
+): Promise<void> {
+  if (!options.skipPackageJson) {
+    process.env.npm_config_legacy_peer_deps ??= 'true';
```
Coly010, line 22: Why is this needed?

packages/angular/src/generators/utils/add-vitest.ts
```diff
@@ -0,0 +1,88 @@
...
+      true
+    );
+  }
+
+  const { createOrEditViteConfig, viteConfigurationGenerator } = ensurePackage<
+    typeof import('@nx/vite')
+  >('@nx/vite', nxVersion);
+
+  const relativeTestSetupPath = joinPathFragments('src', 'test-setup.ts');
+  const isEsmImplicit = readJson(tree, 'package.json').type === 'module';
```
Coly010, line 41: In Nx workspaces, applications and libraries may have their own package.json files which can define `type: module`. 

IIRC, Vite will always look for the closest `package.json` file to determine module format. 

I think this needs to be updated to the package.json path for the application/library, and if it doesn't exist, then fallback to root of workspace.

packages/angular/src/utils/versions.ts
```diff
@@ -28,5 +28,6 @@ export const tsNodeVersion = '10.9.1';
 export const jestPresetAngularVersion = '~14.1.0';
 export const typesNodeVersion = '18.16.9';
 export const jasmineMarblesVersion = '^0.9.2';
+export const analogVitestAngular = '1.7.0';
```
Coly010, line 31: Should this be pinned? 

Could we get away with `~1.7.0`?

packages/vite/src/generators/configuration/schema.d.ts
```diff
@@ -1,5 +1,5 @@
 export interface ViteConfigurationGeneratorSchema {
-  uiFramework: 'react' | 'none';
+  uiFramework: 'angular' | 'react' | 'none';
```
Coly010, line 2: Are we using the `angular` value to determine anything? Could Angular's call to this generator use `none`?

Coly010 (commented):
Looks great! I need to test it out properly, but other than the comments I've added, I don't see anything untoward. 

Could you add some `e2e` tests that cover using Vitest as a selected option so that we can ensure behaviour continues to work going forward?

yjaaidi:
> Looks great! I need to test it out properly, but other than the comments I've added, I don't see anything untoward.
> 
> Could you add some `e2e` tests that cover using Vitest as a selected option so that we can ensure behaviour continues to work going forward?

Of course! By the way, I also wanted to add an e2e to test the watch mode which is generally a common issue. As a matter of fact, the watch was detecting file changes and running the tests without rebuilding the TS. For some reason, I had to upgrade the angular plugin to 1.7.0 to make it work.
🤔 I wonder if it is worth exploring e2e tests for watch mode which can be pretty tricky.

packages/angular/src/generators/utils/add-vitest.ts
yjaaidi: 😅 no clue! I just naively mimicked the jest generator thinking that it was a general trick to prevent the generator from failing in case the vitest plugin is not compatible with the angular version or something.

packages/angular/src/generators/utils/add-vitest.ts
yjaaidi: Wouldn't it be safer to stick to vite's behavior and crawl for `package.json` parent per parent until reaching the root of the workspace?
By the way, this probably deserves a util in `@nx/devkit`. It is common for plugin authors to crawl for a file until the root of the workspace.

packages/angular/src/utils/versions.ts
yjaaidi: `~1.7.0` should definitely work.
That said, this is an interesting topic to discuss beyond this PR. I personally like the predictability of generators and migrations pinning the versions explicitly and totally relying on Nx to update them.
This means that we'd need something like a custom Renovate manager to pick the versions from this file (if it's not already there and I missed it)

packages/vite/src/generators/configuration/schema.d.ts
yjaaidi: Good catch! I'll remove it.
IIRC, I think that at some point, the logic of adding the `src/test-setup.ts` was in this generator then I moved it up.

packages/angular/src/utils/versions.ts
Coly010: I understand that, but using a range allows for multiple things.

1. Our e2es will catch breaking changes if a package we depend on releases a version which breaks anything for us
2. New workspaces aren't locked down to an older version OTTB

Our migrations run based on specific Nx releases. If we pinned every version of every dependency, we'd need a lot more migrations and lot more releases of Nx more frequently to keep users up to date. There's a high maintenance tax to pay there, especially when a range of versions tends to be sufficient.

packages/angular/src/generators/utils/add-vitest.ts
Coly010: Yes to both, feel free to add one :)

packages/angular/src/utils/versions.ts
yjaaidi: Agreed! The plugin already received two patches since I opened this PR 😊

> Our e2es will catch breaking changes if a package we depend on releases a version which breaks anything for us
Not really. Some bugs might go through.
In that case, we would simply update the range.
The only challenge is that users might struggle with the repro as the same nx version might use different plugin versions depending on when the workspace is generated.

Anyway, this also depends on the stability of the package we are depending on. In this case, I am very confident about this one. The analog team is doing a great job 😉

yjaaidi: Done!

packages/vite/src/generators/configuration/schema.d.ts
yjaaidi: Done!

packages/angular/src/generators/utils/add-vitest.ts
yjaaidi: Done!

packages/angular/src/generators/utils/add-vitest.ts
yjaaidi: Done!

packages/vite/src/generators/init/init.ts
```diff
@@ -21,7 +21,8 @@ export function updateNxJsonSettings(tree: Tree) {
   if (productionFileSet) {
     productionFileSet.push(
       '!{projectRoot}/**/?(*.)+(spec|test).[jt]s?(x)?(.snap)',
-      '!{projectRoot}/tsconfig.spec.json'
+      '!{projectRoot}/tsconfig.spec.json',
+      '!{projectRoot}/src/test-setup.[jt]s'
```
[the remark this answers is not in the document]
Coly010: some of our packages supporting an `js` option which transpiles all generated files to js. Some people use it.

packages/vite/src/utils/generator-utils.ts
Coly010: i think this is safer and paves the way for better safety moving forward

Coly010 (approved):

packages/devkit/src/utils/find-file-in-closest-parent-folder.ts
```diff
@@ -0,0 +1,29 @@
...
+import { dirname } from 'path';
+
+/**
+ * Find a file in the closest parent folder.
+ *
+ * @param tree
+ * @param path The path relative to the workspace root to start searching from.
+ * @param fileName The name of the file to search for.
+ */
+export function findFileInClosestParentFolder(
```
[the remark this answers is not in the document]
yjaaidi: Exactly! I thought that using `.ts` when possible would be less confusing to users and machines.
For instance, Playwright VSCode Plugin does not detect `playwright.config.mts` files. So there might be some plugins/tools that do not detect `vite.config.mts`.

Other than that, if you are fine with generating `vite.config.mts` all the time, we can get rid of this function and I'll be happy to fix this.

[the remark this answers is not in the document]
Coly010: Let's generate vite.config.mts all the time and we can remove this function :)

the-ult:
Is there an news on an ETA when this awesome new feature is gonna be available?

yjaaidi:
> Is there any news on an ETA when this awesome new feature is gonna be available?

Hey @the-ult , thanks for your interest to this PR.
I just made the changes requested by @jaysoo so hopefully it could ship soon.

the-ult:
> > Is there any news on an ETA when this awesome new feature is gonna be available?
> 
> Hey @the-ult , thanks for your interest to this PR. I just made the changes requested by @jaysoo so hopefully it could ship soon.

Awesome.
Thanks for all the great work!

packages/angular/src/utils/versions.ts
```diff
@@ -28,5 +28,6 @@ export const tsNodeVersion = '10.9.1';
 export const jestPresetAngularVersion = '~14.1.0';
 export const typesNodeVersion = '18.16.9';
 export const jasmineMarblesVersion = '^0.9.2';
+export const analogVitestAngular = '~1.7.2';
```
brandonroberts, line 31: Bumping this would be recommended, as it contains fixes for coverage reports and sourcemaps for the debugger.

```suggestion
export const analogVitestAngular = '~1.9.1';
```

jaysoo (approved):

brandonroberts (approved):
🫡

yjaaidi:
`@analogjs/vitest-angular` is moving fast. There's a tradeoff there, but shouldn't we use a caret match instead of holding users back?

packages/angular/src/utils/versions.ts
yjaaidi: Thanks buddy!

Coly010:
> @analogjs/vitest-angular is moving fast. There's a tradeoff there, but shouldn't we use a caret match instead of holding users back?

I'd much rather pay the maintenance tax for updating minors more frequently given we support current Angular Major + 2 previous major versions with Nx. We can ensure better stability with all the versions we support this way.

ErickRodrCodes:
This is going to be so 🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥, Now we have to wait for serving/building to be under vite.

PR #27328, opened 2024-08-07
feat(testing): support positional arg for jest executor

MJez29:
## Current Behavior

The syntax to test a single file is different when using the jest executor versus the jest cli

```sh
jest <file>
# vs
nx test project --testFile <file>
```

## Expected Behavior

I would expect the jest executor to support a single positional argument as well

```sh
nx test project <file>
```

This is one of the common questions/complaints that I hear from developers when transitioning from our standalone apps to our monorepo

This change supports both syntaxes so that this is not a breaking change and prints a warning when both are used

### DIFF
--- a/docs/shared/packages/jest/jest-plugin.md
+++ b/docs/shared/packages/jest/jest-plugin.md
@@ -161,6 +161,16 @@ To run Jest tests via nx use
 nx test frontend
 ```
 
+### Testing Specific Files
+
+Using a single positional argument or the `--testFile` flag will run all test files matching the regex. For more info check out the [Jest documentation](https://jestjs.io/docs/cli#jest-regexfortestfiles).
+
+```shell
+nx test frontend HomePage.tsx
+# or
+nx test frontend --testFile HomePage.tsx
+```
+
 ### Watching for Changes
 
 Using the `--watch` flag will run the tests whenever a file changes.
--- a/packages/jest/src/executors/jest/schema.json
+++ b/packages/jest/src/executors/jest/schema.json
@@ -53,7 +53,11 @@
     },
     "testFile": {
       "description": "The name of the file to test.",
-      "type": "string"
+      "type": "string",
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      }
     },
     "tsConfig": {
       "description": "The name of the Typescript configuration file.",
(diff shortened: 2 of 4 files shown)


isaacplmann:
Run `nx documentation` locally and commit the changes to get the CI working

MJez29:
> TypeError: stripJsonComments is not a function
I see this error now 🤔

JamesHenry:
@MJez29 It was an issue on an older version of master, I've just pulled in the latest for you

MJez29:
@isaacplmann @JamesHenry how can I resolve the CI failure now? It looks like a timeout to me

packages/jest/src/executors/jest/schema.json
```diff
@@ -51,6 +51,13 @@
       "x-completion-type": "file",
       "x-completion-glob": "jest.config@(.js|.ts)"
     },
+    "_": {
+      "type": "array",
+      "description": "The name of the file to test as a single positional argument. Equivalent to `--testFile`.",
+      "items": {
+        "type": "string"
+      }
+    },
```
leosvelperez, line 60: We don't need an extra option for this. We can specify a positional argument as the default source for any option (https://nx.dev/extending-nx/recipes/generator-options#default).

So, instead of adding this new option, you can edit the existing `testFile` to:

```diff
    "testFile": {
      "description": "The name of the file to test.",
      "type": "string",
+     "$default": {
+       "$source": "argv",
+       "index": 0
+     }
    },
```

With that in place, you get the desired behavior and don't need to handle any extra `_` option. The `testFile` option will get assigned the value of the first positional argument when the `--testFile` flag is not provided. The `--testFile` flag can still be used, in which case, it will override the positional argument if also provided.

leosvelperez (changes requested):

docs/shared/packages/jest/jest-plugin.md
```diff
@@ -161,6 +161,16 @@ To run Jest tests via nx use
 nx test frontend
 ```
 
+### Testing a Single File
```
leosvelperez, line 164: I'd rephrase it a bit because the regex can match multiple files. So, the section is really showing how to test specific files.
```suggestion
### Testing Specific Files
```

leosvelperez (changes requested):
Sorry, I missed this in my previous review. One last small thing left, and it should be good to go 🚀

leosvelperez (approved):

PR #27334, opened 2024-08-08
feat(testing): allow usage of jest 30 pre-release versions

leosvelperez:
Allow usage of Jest v30 pre-release version in preparation for the upcoming stable release. 

Note that this PR is not intended to change any current generator for creating/setting up projects with Jest 30. That will be done when Jest v30 becomes stable.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/package.json
+++ b/packages/jest/package.json
@@ -47,6 +47,7 @@
     "jest-util": "^29.4.1",
     "minimatch": "9.0.3",
     "resolve.exports": "1.1.0",
+    "semver": "^7.5.3",
     "tslib": "^2.3.0",
     "yargs-parser": "21.1.1"
   },
--- a/packages/jest/src/plugins/plugin.ts
+++ b/packages/jest/src/plugins/plugin.ts
@@ -1,3 +1,4 @@
+import type { Config } from '@jest/types';
 import {
   CreateNodes,
   CreateNodesContext,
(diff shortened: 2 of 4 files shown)


packages/jest/src/plugins/plugin.ts
```diff
@@ -211,23 +213,24 @@ async function buildJestTargets(
...
       // nx-ignore-next-line
-    })) as typeof import('jest-runtime');
+      requireJestUtil<typeof import('jest-runtime')>('jest-runtime');
 
     const jestContext = await Runtime.createContext(config.projectConfig, {
       maxWorkers: 1,
       watchman: false,
     });
 
+    const jest = require('jest') as typeof import('jest');
```
FrozenPandaz, line 225: This variable doesn't seem to get used?

FrozenPandaz, line 225: In a separate PR, can we change it so that when we resolve `jest`, we resolve it from the config file's directory?

packages/jest/package.json
```diff
@@ -35,18 +35,19 @@
     "migrations": "./migrations.json"
   },
   "dependencies": {
-    "@jest/reporters": "^29.4.1",
-    "@jest/test-result": "^29.4.1",
+    "@jest/reporters": "^29.4.1 || ^30.0.0-alpha.6",
```
FrozenPandaz, line 38: These are our `dependencies`... so this would pretty much resolve `30.0.0-alpha6` right? People shouldn't really care.. but should we depend on Jest's alpha?

FrozenPandaz (changes requested):

packages/jest/package.json
leosvelperez: That's a good point. At least `pnpm` resolves one or the other depending on the version that gets resolved from other package dependencies or the repo dependencies. I haven't tested with the other package managers. Still, I agree that's better to be explicit and not depend on the alpha. Given these are dependencies, they shouldn't cause any conflicts with whatever the user has.

FrozenPandaz (approved):

Issue #26900, opened 2024-07-11
Including Assets in Build doesn't work (regression)


ondrej-stanek-ozobot:
### Current Behavior

`nrwl/nx 19.4.2` doesn't copy the assets specified explicitly in the app's `project.json` per documentation [Including Assets in Your Build](https://nx.dev/recipes/tips-n-tricks/include-assets-in-build#including-assets-in-your-build).

### Expected Behavior

The assets listed in the "assets" section in the app's `project.json` should be copied to the `dist` folder on build.

### Steps to Reproduce

1. initialize a new nx workspace:
```bash
npx -y create-nx-workspace@19.4.2 --appName copyTest --preset next --e2eTestRunner none --style css --nextSrcDir test_nx_19.4.2 --nxCloud skip .
```
2. populate some dummy files to be copied:
```bash
test_nx_19.4.2$ mkdir -p path-to-my-project/src
test_nx_19.4.2$ touch path-to-my-project/readme.md
test_nx_19.4.2$ touch path-to-my-project/executors.json
test_nx_19.4.2$ touch path-to-my-project/src/some.js
```
3. modify the `apps/copyTest/project.json` to contain the copy directives (per the [documentation](https://nx.dev/recipes/tips-n-tricks/include-assets-in-build#including-assets-in-your-build)):

```json
{
  "name": "copyTest",
  "$schema": "../../node_modules/nx/schemas/project-schema.json",
  "sourceRoot": "apps/copyTest",
  "projectType": "application",
  "tags": [],
  "// targets": "to see all targets run: nx show project nx_test --web",
  "targets": {
    "build": {
      "executor": "@nx/next:build",
      "outputs": ["{options.outputPath}"],
      "defaultConfiguration": "production",
      "options": {
        "outputPath": "dist/apps/copyTest",
          "assets": [ 
          // Copies all the markdown files at the root of the project to the root of the output bundle
          "path-to-my-project/*.md",
          {
            "input": "./path-to-my-project/src", // look in the src folder
            "glob": "**/!(*.ts)", // for any file (in any folder) that is not a typescript file
            "output": "./src" // put those files in the src folder of the output bundle
          },
          {
             "input": "./path-to-my-project", // look in the project folder
             "glob": "executors.json", // for the executors.json file
             "output": "." // put the file in the root of the output bundle
          }
        ]
      }
    }
  }
}
```

3. build the app:
```bash
npx nx build copyTest
```

4. list the contents of the build folder to verify if assets were copied:
[...]

PR #27337, opened 2024-08-08
fix(nextjs): Fix order, and wrong number call to createWebpackConfig()

tvdoan:
## Current Behavior
Assets are not copied to the dist folder after running @nx/next:build

## Expected Behavior
Assets should be copied to the dist folder after @nx/next:build

## Related Issue(s)
https://github.com/nrwl/nx/issues/26900

Fixes #
Fix the order, and wrong number of parameters call to createWebpackConfig()

### DIFF
--- a/packages/next/plugins/with-nx.ts
+++ b/packages/next/plugins/with-nx.ts
@@ -212,7 +212,12 @@ function withNx(
 
       const userWebpackConfig = nextConfig.webpack;
 
-      const { createWebpackConfig } = require('@nx/next/src/utils/config');
+      const { createWebpackConfig } = require(require.resolve(
+        '@nx/next/src/utils/config',
+        {
+          paths: [workspaceRoot],
+        }
+      )) as typeof import('@nx/next/src/utils/config');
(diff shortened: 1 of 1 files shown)


packages/next/plugins/with-nx.ts
```diff
@@ -221,8 +226,8 @@ function withNx(
             ? joinPathFragments(workspaceRoot, projectDirectory)
             : workspaceRoot,
           projectDirectory,
-          _nextConfig.nx?.fileReplacements || options.fileReplacements,
-          _nextConfig.nx?.assets || options.assets
+          _nextConfig.nx?.assets || options.assets,
+          _nextConfig.nx?.fileReplacements || options.fileReplacements
```
tvdoan, line 230: @ndcunningham The order is incorrect here.  `createWebpackConfig()` expects assets to be the last argument.  Please see here https://github.com/nrwl/nx/blob/6d83ae2fb1806dbc3f6505cdd7d682ae4b662169/packages/next/src/utils/config.ts#L19

ndcunningham (approved):
LGTM

alexandr2110pro:
Hey guys. When can we expect it to be released?

jaysoo:
19.6 should be today

PR #27345, opened 2024-08-08
feat(module-federation): Update SSR enabling static serving for remotes

ndcunningham:
## Current Behavior

Currently, the default for remotes is to server them as development. Which means a separate node process for each remote that is inside the workspace.

This does not scale well and can lead to out of memory exceptions.

## Expected Behavior

Remotes will start as static by default, which allows for better scaling as the remotes increase.

## Related Issue(s)

TODO
- [ ] Migrations


packages/angular/src/executors/module-federation-ssr-dev-server/lib/start-static-remotes.ts
```diff
@@ -0,0 +1,51 @@
...
+  ssrStaticRemotesConfig: StaticRemotesConfig,
+  context: ExecutorContext,
+  options: Schema
+) {
+  if (ssrStaticRemotesConfig.remotes.length === 0) {
+    return;
+  }
+
+  // The directories are usually generated with /browser and /server suffixes so we need to copy them to a common directory
+  const commonOutputDirectory = join(workspaceRoot, 'tmp/static-remotes');
```
Coly010, line 18: Can we have add a listener for when original process exits / sigterms to remove this tmp directory?

This will need done for React too.

packages/react/src/generators/setup-ssr/setup-ssr.ts
```diff
@@ -176,7 +176,7 @@ export async function setupSsrGenerator(tree: Tree, options: Schema) {
       },
     },
     serve: {
-      executor: '@nx/webpack:ssr-dev-server',
+      executor: '@nx/react:module-federation-ssr-dev-server',
```
Coly010, line 179: I don't think we can do this. This `setup-ssr` is a generator, which may be used in non-module federation apps.

Coly010:
Sorry, I know this is adding more work, but we'll also need to add this change to the start ssr proxies: https://github.com/nrwl/nx/pull/27381

packages/angular/src/executors/module-federation-ssr-dev-server/lib/start-static-remotes.ts
ndcunningham: This makes sense, we should do it.

ndcunningham: On second thought let's leave it, the process is being forcibly ended quicker than these have time to run. 
Let's take a look at enhancing them later on.

jaysoo (approved):

Coly010 (approved):

PR #27379, opened 2024-08-12
feat(angular): support angular v18.2.0

leosvelperez:
## Related Issue(s)

Fixes #


FrozenPandaz (approved):
Do we want to merge this as is and then change the versions to the stable version when it's out?

FrozenPandaz (approved):

PR #27381, opened 2024-08-12
fix(module-federation): remote proxies should use https when host is configured with ssl #27360

### LINKED ISSUE #27360: Proxies for dynamic remotes do not support SSL
### Current Behavior

I have an application that uses dynamic federation, and loading remotes does not work with SSL enabled during development.

#26782 introduced this new proxy for remotes to static servers, which works great, but it just doesn't take into account SSL at all.

### Expected Behavior

An application that uses dynamic federation can use SSL during development.

### GitHub Repo

https://github.com/sknoslo/ng-mf-ssl

### Steps to Reproduce

1. Start the app with `nx serve host`
2. Navigate to `https://localhost:4200/todo` to load the todo app, you'll see the connection refused in the network devtools

### Nx Report

```shell
Node           : 20.16.0
OS             : win32-x64
npm            : 10.8.1

nx                 : 19.5.7
@nx/js             : 19.5.7
@nx/jest           : 19.5.7
@nx/linter         : 19.5.7
@nx/eslint         : 19.5.7
@nx/workspace      : 19.5.7
@nx/angular        : 19.5.7
@nx/devkit         : 19.5.7
@nx/eslint-plugin  : 19.5.7
@nrwl/tao          : 19.5.7
@nx/web            : 19.5.7
@nx/webpack        : 19.5.7
typescript         : 5.5.4
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

I was able to patch `startRemoteProxies` to start an https server if SSL is configured for a remote, and this solves the issue. I also attempted to only serve the host with SSL, and not the remotes, but the static server itself is still started with SSL and the proxies fail to verify the secure connection because it is a self signed certificate. In the http-proxy-middleware config, I had to set `secure: false` to get around that.

Coly010:
## Current Behavior

When the host uses SSL, the remote proxies are not created on a secure server.

## Expected Behavior

When the host uses SSL, the remote proxies should be created on a secure server

## Related Issue(s)

Fixes #27360

### DIFF
--- a/packages/webpack/src/utils/module-federation/start-remote-proxies.ts
+++ b/packages/webpack/src/utils/module-federation/start-remote-proxies.ts
@@ -1,13 +1,33 @@
 import type { Express } from 'express';
 import { logger } from '@nx/devkit';
 import { StaticRemotesConfig } from './parse-static-remotes-config';
+import { existsSync, readFileSync } from 'fs';
 
 export function startRemoteProxies(
   staticRemotesConfig: StaticRemotesConfig,
-  mappedLocationsOfRemotes: Record<string, string>
+  mappedLocationsOfRemotes: Record<string, string>,
+  sslOptions?: { pathToCert: string; pathToKey: string }
 ) {
   const { createProxyMiddleware } = require('http-proxy-middleware');
   const express = require('express');
+  let sslCert: Buffer;
+  let sslKey: Buffer;
+  if (sslOptions && sslOptions.pathToCert && sslOptions.pathToKey) {
+    if (existsSync(sslOptions.pathToCert) && existsSync(sslOptions.pathToKey)) {
+      sslCert = readFileSync(sslOptions.pathToCert);
+      sslKey = readFileSync(sslOptions.pathToKey);
+    } else {
+      logger.warn(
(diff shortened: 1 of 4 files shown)


packages/webpack/src/utils/module-federation/start-remote-proxies.ts
```diff
@@ -1,25 +1,46 @@
 import type { Express } from 'express';
-import { logger } from '@nx/devkit';
+import { logger, ProjectGraph } from '@nx/devkit';
```
ndcunningham, line 2: Do you need the `ProjectGraph` ref here?

ndcunningham (changes requested):

packages/webpack/src/utils/module-federation/start-remote-proxies.ts
Coly010: Nope, fixed, good catch

ndcunningham (approved):

jaysoo (approved):

PR #27410, opened 2024-08-13
chore(repo): add comments to make sure code stays manually synced with nx console

MaxKless:
this code is needed in a version-independent way in Nx Console and right now there's no good place to share it. 
So for now we will duplicate it


jaysoo (approved):

jaysoo:
> Is this only duplicated for legacy versions? and imported for later versions?

That's a good point. We could try importing it first, but we still have to update it for legacy versions for consistency/correctness.

MaxKless:
very good point

FrozenPandaz (approved):

PR #27411, opened 2024-08-13
fix(module-federation): Add migration for ssr server file to run on it's own port

ndcunningham:
## Current Behavior

Currently, if you have a SSR Mfe project and you run multiple ssr projects using the 'serve' target it would fail due to conflicting ports unless explicitly set `PORT=4202 node dist/acme/server.js`.

## Expected Behavior

Now the default is to have the server take the same port as it was generated with so that if you run multiple server apps that depends on other apps in dev mode there will be no conflicts.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/migrations.json
+++ b/packages/react/migrations.json
@@ -47,6 +47,12 @@
       "version": "19.6.0-beta.4",
       "description": "Ensure Module Federation DTS is turned off by default.",
       "factory": "./src/migrations/update-19-6-0/turn-off-dts-by-default"
+    },
+    "update-module-federation-ssr-server-file": {
+      "cli": "nx",
+      "version": "19.6.0-beta.4",
+      "description": "Update the server file for Module Federation SSR port value to be the same as the 'serve' target port value.",
+      "factory": "./src/migrations/update-19-6-0/update-ssr-server-port"
     }
   },
   "packageJsonUpdates": {
--- a/packages/react/src/migrations/update-19-6-0/update-ssr-server-port.spec.ts
+++ b/packages/react/src/migrations/update-19-6-0/update-ssr-server-port.spec.ts
@@ -1,0 +1,206 @@
+import { readProjectConfiguration, type Tree } from '@nx/devkit';
+import { createTreeWithEmptyWorkspace } from 'nx/src/devkit-testing-exports';
+import hostGenerator from '../../generators/host/host';
+import { Linter } from '@nx/eslint';
+import updateSsrServerPort from './update-ssr-server-port';
+describe('update-19-6-0 update-ssr-server-port migration', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+  });
+
+  it('should update host and remote port server files', async () => {
+    await hostGenerator(tree, {
+      name: 'shell',
+      e2eTestRunner: 'none',
+      unitTestRunner: 'none',
+      ssr: true,
+      linter: Linter.EsLint,
+      projectNameAndRootFormat: 'as-provided',
+      style: 'css',
+      remotes: ['product'],
+    });
+    const remotePort = readProjectConfiguration(tree, 'product').targets.serve
+      .options.port;
+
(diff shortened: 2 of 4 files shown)


packages/react/src/migrations/update-19-6-0/update-ssr-server-port.ts
```diff
@@ -0,0 +1,132 @@
...
+import { minimatch } from 'minimatch';
+
+export default function update(tree: Tree) {
+  const projects = getProjects(tree);
+  const executors = [
+    '@nx/webpack:ssr-dev-server',
+    '@nx/react:module-federation-ssr-dev-server',
+  ];
+
+  for (const project of projects.values()) {
```
jaysoo, line 19: It'd be easier to use `forEachExecutorOptions` here?

packages/react/src/migrations/update-19-6-0/update-ssr-server-port.ts
```diff
@@ -0,0 +1,132 @@
...
+  tree: Tree,
+  project: ProjectConfiguration
+) {
+  let hasMfeServerConfig = false;
+  let hasServerFile = false;
+  visitNotIgnoredFiles(tree, project.root, (filePath) => {
+    if (minimatch(filePath, '**/module-federation*.server.config.*')) {
+      hasMfeServerConfig = true;
+    }
+    if (minimatch(filePath, '**/server.ts')) {
```
jaysoo, line 127: I don't think we need `minimatch` here since later we're only handling `<project-root>/server.ts`.

packages/react/src/migrations/update-19-6-0/update-ssr-server-port.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      ssr: true,
+      linter: Linter.EsLint,
+      projectNameAndRootFormat: 'as-provided',
+      style: 'css',
+    });
+
+    const hostPort = readProjectConfiguration(tree, 'host').targets.serve
+      .options.port;
+
+    if (tree.exists('host/server.ts')) {
```
jaysoo, line 137: I don't think we need to check this right? It's always there from the host generator?

jaysoo (approved):

packages/react/src/migrations/update-19-6-0/update-ssr-server-port.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      expect(tree.read('host/server.ts', 'utf-8')).toContain(
+        `port = ${hostPort}`
+      );
+      expect(tree.read('host/server.ts', 'utf-8')).toMatchInlineSnapshot(`
+        "import * as path from 'path';
+        import express from 'express';
+        import cors from 'cors';
+        import { handleRequest } from './src/main.server';
+        const port = 4200;
```
jaysoo, line 158: I think we should keep `process.env['PORT']` in here, but just update the default port if the env var is undefined.

packages/react/src/migrations/update-19-6-0/update-ssr-server-port.ts
ndcunningham: yup will use that instead

packages/react/migrations.json
```diff
@@ -41,6 +41,12 @@
       "version": "18.1.1-beta.0",
       "description": "Ensure targetDefaults inputs for task hashing when '@nx/webpack:webpack' is used are correct for Module Federation.",
       "factory": "./src/migrations/update-18-1-1/fix-target-defaults-inputs"
+    },
+    "update-module-federation-ssr-server-file": {
+      "cli": "nx",
+      "version": "19-6-0-beta.4",
```
jaysoo, line 47: `19.6.0-beta.4`

ndcunningham: Should be already there I think.

jaysoo: No, `19-6-0-beta.4` is invalid, it needs to be dots.

ndcunningham: 🤦🏿

PR #27418, opened 2024-08-14
fix(core): improve the dx of the sync commands and gracefully handle exiting the prompt when running tasks

leosvelperez:
- Update the `nx sync` output when there are changes to sync (currently there's no output)
- Update the `nx sync` and `nx sync:check` output when there are no changes to sync (currently there's no output)
- Handle exiting the prompt for applying sync generator changes when running tasks

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/run-command.ts
+++ b/packages/nx/src/tasks-runner/run-command.ts
@@ -369,7 +369,7 @@ async function promptForApplyingSyncGeneratorChanges(): Promise<boolean> {
     return await prompt<{ applyChanges: 'yes' | 'no' }>([promptConfig]).then(
       ({ applyChanges }) => applyChanges === 'yes'
     );
-  } catch (e) {
+  } catch {
     process.exit(1);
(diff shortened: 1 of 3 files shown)


packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -344,30 +344,34 @@ Please make sure to commit the changes to your repository.`);
...
+        ),
+    };
 
-  return await prompt<{ applyChanges: 'yes' | 'no' }>([promptConfig]).then(
-    ({ applyChanges }) => applyChanges === 'yes'
-  );
+    return await prompt<{ applyChanges: 'yes' | 'no' }>([promptConfig]).then(
+      ({ applyChanges }) => applyChanges === 'yes'
+    );
+  } catch (e) {
```
JamesHenry, line 372: Nit: could miss off the `(e)`

JamesHenry (approved):

PR #27461, opened 2024-08-15
docs(core): add scroll_25, scroll_50, scroll_75, and scroll_90 events to track engagement

jaysoo:
This PR adds events to track engagement in our docs. Since we use a scrollable `<div>` in our docs, the normal `scroll` events in GA do not work.

A new `<ScrollableContent>` component is added that will do two things:
- Send `scroll_25`, `scroll_50`, `scroll_75`, and `scroll_90` events whenever the user scrolls to 25%, 50%, 75%, of 90% of the content
- Optionally reset scroll top to zero whenever router changes (existing behavior)

All of the places where we have content in a scrollable `<div>` is replaced with `<ScrollableContent>`.

Note: 90% means user has reached the bottom, since it's not usually possible to get to 100%.


isaacplmann (approved):

ndcunningham (commented):
Would this also track if the user scrolls via the table of contents?

jaysoo:
> Would this also track if the user scrolls via the table of contents?

Yes, it's just any scroll, whether through mouse of through the anchor links.

bcabanes (approved):

ndcunningham (approved):

PR #27474, opened 2024-08-16
feat(release): support groupPreVersionCommand for release groups

jogelin:
## Current Behavior
In Nx release configuration, the property `preVersionCommand` can be set only at the root level in the `nx.json`:

```json
"release": {
    "version": {
        "preVersionCommand": "nx run-many -t build"
    }
},    
```
It is not possible to specify it per group meaning that if we want to release only one group, we have to build all releasable projects.

## Expected Behavior
We should be able to specify `preVersionCommand` per group:

```json
"release": {
    "groups": {
        "toolkits": {
            "projects": ["*-toolkit"],
            "version": {
                "groupPreVersionCommand": "nx run-many -t build -p \"*-toolkit\""
             }
        }
    }
},
```
In that way, we can align the list of projects related to the group and the list of projects to build


JamesHenry (changes requested):
Given some feedback on X

JamesHenry (changes requested):
Still pending the discussed updates

JamesHenry:
Updated the PR title to match what will be coming to the PR

docs/shared/recipes/nx-release/build-before-versioning.md
```diff
@@ -13,3 +13,22 @@ In order to ensure that projects are built before the new version is applied to
...
+        "version": {
+          "groupPreVersionCommand": "npx nx run-many -t build -p my-lib-one,my-lib-two"
+        }
+      }
+    }
+  }
+}
+```
+
+The `groupPreVersionCommand` will run in addition of the global `preVersionCommand`.
```
JamesHenry, line 34:
```suggestion
The `groupPreVersionCommand` will run in addition to the global `preVersionCommand`.
```

packages/nx/src/command-line/release/config/config.ts
```diff
@@ -355,8 +356,16 @@ export async function createNxReleaseConfig(
     );
 
   // these options are not supported at the group level, only the root/command level
-  const rootVersionWithoutGlobalOptions = { ...rootVersionConfig };
+  const rootVersionWithoutGlobalOptions = {
```
JamesHenry, line 359: You left this underneath the comment saying it is not something which is supported at the group level, which doesn't make sense, but more broadly why are you creating a relationship between preVersionCommand and groupPreVersionCommand?

They are separate things, right now the way you have things it looks like if you set a preVersionCommand it will be run twice, once as a preVersionCommand and once as a groupPreVersionCommand?

Please remove this relationship

packages/nx/src/command-line/release/config/config.spec.ts
```diff
@@ -3313,6 +3331,7 @@ describe('createNxReleaseConfig()', () => {
                   "conventionalCommits": false,
                   "generator": "@nx/js:release-version",
                   "generatorOptions": {},
+                  "groupPreVersionCommand": "nx run-many -t build",
```
JamesHenry, line 3334: This confirms what I said in my previous comment - this should not be set in this example, we do not want the command to be duplicated from global pre to group pre

JamesHenry (changes requested):

JamesHenry (approved):

JamesHenry:
Thank you @jogelin!

PR #27476, opened 2024-08-16
fix(linter): update the @nx/dependency-checks rule to read the package.json content from the rule context

### LINKED ISSUE #27412: dependency-check lint rule, --fix needs to be run twice
### Current Behavior

lib-01 `package.json` file has a missing peerDependency (`@angular/core`) and one dependency too many (`@angular/animations`).
In order for both errors to be fixed I need to run `nx lint lib-01 --fix` two consecutive times

### Expected Behavior

Both errors in lib-01 `package.json` file should be fixed by the first `nx lint lib-01 --fix` run

### GitHub Repo

https://github.com/r4b6i0/nx-dependency-checks-bug

### Steps to Reproduce

1. check that lib-01 `package.json` file contains the peerDependencies `@angular/common` and `@angular/animations`, which are not correct based on lib-01 `import` statements
2. run the command `nx lint lib-01 --fix`
3. `package.json` file now contains only the peerDependency `@angular/common`
4. run the command `nx lint lib-01 --fix` again
5. `package.json` file now contains both the peerDependencies `@angular/common` and `@angular/core` which are the correct dependencies based on lib-01 `import` statements

### Nx Report

```shell
Node   : 18.20.4
OS     : win32-x64
npm    : 10.7.0

nx                 : 18.3.5
@nx/js             : 18.3.5
@nx/jest           : 18.3.5
@nx/linter         : 18.3.5
@nx/eslint         : 18.3.5
@nx/workspace      : 18.3.5
@nx/angular        : 18.3.5
@nx/cypress        : 18.3.5
@nx/devkit         : 18.3.5
@nx/eslint-plugin  : 18.3.5
@nrwl/tao          : 18.3.5
@nx/web            : 18.3.5
@nx/webpack        : 18.3.5
typescript         : 5.4.5
Registered Plugins:
@nx/cypress/plugin
@nx/eslint/plugin
```

### Failure Logs

_No response_

### Package Manager Version

npm 10.7.0

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

To create this example I started with a new empty nx@18.3.5 repo using the command `npx create-nx-workspace@18.3.5`, added a new angular library with the command `nx g @nx/angular:library lib-01 --publishable --import-path=@my-org/my-lib-01`, modified the library's `package.json` so that the `peerDependencies` are not correct.
I did nothing else.

leosvelperez:
## Current Behavior

The `@nx/dependency-checks` rule sometimes doesn't apply all identified fixes and wrongly succeeds. ESLint performs multiple passes to try to apply fixes, and in each pass the rule receives the updated source code with the previously applied fixes. This allows merging different fixes, but the `@nx/dependency-checks` rule always reads the `package.json` file from the filesystem, caches it globally, and mutates it. It never uses the updated source code provided in the rule context.

## Expected Behavior

The `@nx/dependency-checks` rule should apply all identified fixes [as long as they don't conflict](https://eslint.org/docs/latest/extend/custom-rules#conflicting-fixes).

## Related Issue(s)

Fixes #27412

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -8,6 +8,7 @@ import { satisfies } from 'semver';
 import {
   getAllDependencies,
   getPackageJson,
+  getProductionDependencies,
 } from '../utils/package-json-utils';
 import { readProjectGraph } from '../utils/project-graph-utils';
 import {
@@ -152,11 +153,7 @@ export default ESLintUtils.RuleCreator(
     const expectedDependencyNames = Object.keys(npmDependencies);
 
     const packageJson = JSON.parse(context.sourceCode.getText());
-    const projPackageJsonDeps = {
-      ...packageJson.dependencies,
-      ...packageJson.peerDependencies,
-      ...packageJson.optionalDependencies,
-    };
+    const projPackageJsonDeps = getProductionDependencies(packageJson);
 
     const rootPackageJsonDeps = getAllDependencies(rootPackageJson);
 
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -8,7 +8,6 @@ import { satisfies } from 'semver';
 import {
   getAllDependencies,
   getPackageJson,
-  getProductionDependencies,
 } from '../utils/package-json-utils';
 import { readProjectGraph } from '../utils/project-graph-utils';
 import {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -8,7 +8,6 @@ import { satisfies } from 'semver';
 import {
   getAllDependencies,
   getPackageJson,
-  getProductionDependencies,
```
JamesHenry, line 11: I believe this is the only place this function is used, so we should just change it to work like `getAllDependencies` and be given package.json contents instead of reading from disk?

Either that or delete the now unused `getProductionDependencies`

JamesHenry (changes requested):

JamesHenry (approved):

PR #27478, opened 2024-08-16
docs(core): plugin atomizer set up

isaacplmann:
Update plugin overview pages to describe atomizer set up

### DIFF
--- a/docs/shared/packages/playwright/playwright-plugin.md
+++ b/docs/shared/packages/playwright/playwright-plugin.md
@@ -65,6 +65,24 @@ The `@nx/playwright/plugin` is configured in the `plugins` array in `nx.json`.
 
 The `targetName` and `ciTargetName` options control the name of the inferred Playwright tasks. The default names are `e2e` and `e2e-ci`.
 
+### Splitting E2E Tests
+
+The `@nx/playwright/plugin` will automatically split your e2e tasks by file if you provide a `ciTargetName`. You can read more about this feature [here](/ci/features/split-e2e-tasks). This will create a target with that name which can be used in CI to run the tests for each file in a distributed fashion.
+
+```json {% fileName="nx.json" %}
+{
+  "plugins": [
+    {
+      "plugin": "@nx/playwright/plugin",
+      "options": {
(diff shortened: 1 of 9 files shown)


docs/shared/packages/playwright/playwright-plugin.md
```diff
@@ -65,6 +65,24 @@ The `@nx/playwright/plugin` is configured in the `plugins` array in `nx.json`.
 
 The `targetName` and `ciTargetName` options control the name of the inferred Playwright tasks. The default names are `e2e` and `e2e-ci`.
 
+### Splitting E2E Tests
+
+The `@nx/playwright/plugin` will automatically split your e2e tasks by file if you provide a `ciTargetName`. You can read more about this feature [here](/ci/features/split-e2e-tasks). This will create a target with that name which can be used in CI to run the tests for each file in a distributed fashion.
```
jaysoo, line 70: I think we should mention "atomizer" here for people searching for it.

Maybe instead of `You can read more about this feature here` we can say `You can read more about the Atomizer feature here`.

docs/shared/packages/playwright/playwright-plugin.md
```diff
@@ -65,6 +65,24 @@ The `@nx/playwright/plugin` is configured in the `plugins` array in `nx.json`.
 
 The `targetName` and `ciTargetName` options control the name of the inferred Playwright tasks. The default names are `e2e` and `e2e-ci`.
 
+### Splitting E2E Tests
```
jaysoo, line 68: Unrelated to this change, but the Playwright overview is formatted wrong. The `Nx 18+` tab contains all the inferred and task splitting content as well -- it should just be the `nx add @nx/playwright` codeblock.

Screenshot 2024-08-16 at 10 33 57 AM

jaysoo (approved):

docs/shared/packages/playwright/playwright-plugin.md
isaacplmann: Inferred and task splitting doesn't work in Nx < 18

PR #27489, opened 2024-08-16
fix(angular): tests on windows

### LINKED ISSUE #22248: cache miss because of npmDependencies on nx reset
### Current Behavior

nx lint has negative cache miss if the npm changed
it seems nx cache add the new dependencies incorrectly in the cache if the Daemon Server is running and you must reset it everytime you change branch or dependency

### Expected Behavior

nx to have deterministic behavior, same source should always give cache hit

### GitHub Repo

_No response_

### Steps to Reproduce

1. in any existing repo run `npm i jsonc-eslint-parser` and use that parser in your eslint configuration
2. run `nx lint`
3. run `nx reset`
4. run `nx lint`

### Nx Report

```shell
Node   : 20.11.0
   OS     : win32-x64
   npm    : 10.4.0
   
   nx                 : 18.0.4
   @nx/js             : 18.0.4
   @nx/jest           : 18.0.4
   @nx/linter         : 18.0.4
   @nx/eslint         : 18.0.4
   @nx/workspace      : 18.0.4
   @nx/angular        : 18.0.4
   @nx/cypress        : 18.0.4
   @nx/devkit         : 18.0.4
   @nx/eslint-plugin  : 18.0.4
   @nx/nest           : 18.0.4
   @nx/node           : 18.0.4
   @nrwl/tao          : 18.0.4
   @nx/web            : 18.0.4
   @nx/webpack        : 18.0.4
   typescript         : 5.3.3
   Community plugins:
   ng-mocks : 14.12.1
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

image

robertIsaac:
## Current Behavior
`nx test angular` fail on windows

## Expected Behavior
`nx test angular` to pass on windows

## notes
in `packages/angular/src/generators/utils/storybook-ast/component-info.ts` file I used `join(moduleFolderPath)` to convert it from always being `/` to be OS separator
there are two other options I can do, but I don't have enough knowledge to take the decision
1. to generate `moduleFolderPath` with OS separator instead of always `/`
2. to make `candidateFile` to always be `/`

## Related Issue(s)
it might Fixes #22248
because now I see that when a new file is added, it's added to the end of the tree, but when you reset it's then added to its place
also it explain why same code sometimes hit cache in windows, but miss cache on linux, as the returned result different between the OSs
this should make sure that the returned result is always the same

### DIFF
--- a/packages/angular/src/generators/utils/storybook-ast/component-info.ts
+++ b/packages/angular/src/generators/utils/storybook-ast/component-info.ts
@@ -6,7 +6,7 @@ import {
   visitNotIgnoredFiles,
 } from '@nx/devkit';
 import { ensureTypescript } from '@nx/js/src/utils/typescript/ensure-typescript';
-import { basename, dirname, extname, join, relative, sep } from 'path';
+import { basename, dirname, extname, relative } from 'path';
 import type { Identifier, SourceFile, Statement } from 'typescript';
 import { getTsSourceFile } from '../../../utils/nx-devkit/ast-utils';
 import { getInstalledAngularVersionInfo } from '../version-utils';
@@ -261,7 +261,7 @@ function getComponentInfoFromDir(
 
   const componentImportPathChildren: string[] = [];
   visitNotIgnoredFiles(tree, dir, (filePath) => {
-    componentImportPathChildren.push(filePath);
+    componentImportPathChildren.push(normalizePath(filePath));
   });
 
   for (const candidateFile of componentImportPathChildren) {
@@ -273,10 +273,10 @@ function getComponentInfoFromDir(
       );
       if (content.match(classAndComponentRegex)) {
         path = candidateFile
-          .slice(0, candidateFile.lastIndexOf(sep))
-          .replace(join(moduleFolderPath), '.');
+          .slice(0, candidateFile.lastIndexOf('/'))
+          .replace(moduleFolderPath, '.');
         componentFileName = candidateFile.slice(
-          candidateFile.lastIndexOf(sep) + 1,
+          candidateFile.lastIndexOf('/') + 1,
           candidateFile.lastIndexOf('.')
         );
         break;
--- a/packages/angular/src/generators/utils/storybook-ast/component-info.ts
+++ b/packages/angular/src/generators/utils/storybook-ast/component-info.ts
@@ -6,7 +6,7 @@ import {
   visitNotIgnoredFiles,
 } from '@nx/devkit';
 import { ensureTypescript } from '@nx/js/src/utils/typescript/ensure-typescript';
-import { basename, dirname, extname, relative } from 'path';
+import { basename, dirname, extname, join, relative, sep } from 'path';
 import type { Identifier, SourceFile, Statement } from 'typescript';
 import { getTsSourceFile } from '../../../utils/nx-devkit/ast-utils';
 import { getInstalledAngularVersionInfo } from '../version-utils';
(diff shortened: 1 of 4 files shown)


packages/angular/src/generators/utils/storybook-ast/component-info.ts
```diff
@@ -256,10 +256,10 @@ function getComponentInfoFromDir(
       );
       if (content.match(classAndComponentRegex)) {
         path = candidateFile
-          .slice(0, candidateFile.lastIndexOf('/'))
-          .replace(moduleFolderPath, '.');
+          .slice(0, candidateFile.lastIndexOf(sep))
+          .replace(join(moduleFolderPath), '.');
         componentFileName = candidateFile.slice(
-          candidateFile.lastIndexOf('/') + 1,
+          candidateFile.lastIndexOf(sep) + 1,
```
leosvelperez, line 262: Let's revert this. Generators should work with Unix-based paths. The issue here is that `visitNotIgnoredFiles` is invoking the callback function with OS-dependent paths. So, instead of this change, we can change [line 247](https://github.com/nrwl/nx/pull/27489/files#diff-00a320109b1ece45fa3cc3440d9b6a23b925aa98480f589c43df47254b88c375R247) with:

```ts
componentImportPathChildren.push(normalizePath(filePath));
```

packages/nx/src/generators/tree.ts
```diff
@@ -293,7 +293,10 @@ export class FsTree implements Tree {
       return !r?.isDeleted;
     });
     // Dedupe
-    return Array.from(new Set(res));
+    res = Array.from(new Set(res));
+    // in windows the result may not be sorted
+    res.sort();
+    return res;
```
leosvelperez, line 299: Let's revert this. Most of the time, the children don't "need" to be sorted and this could be a breaking change in some scenarios. So let's avoid it. This is a specific need for a couple of test assertions, so we can just sort the result before asserting in those particular test cases.

packages/nx/src/generators/tree.spec.ts
```diff
@@ -387,9 +387,9 @@ describe('tree', () => {
 
         expect(tree.children('parent')).toEqual([
```
leosvelperez, line 388:
```suggestion
        expect(tree.children('parent').sort()).toEqual([
```

packages/workspace/src/generators/preset/__snapshots__/preset.spec.ts.snap
leosvelperez, line 1: This file no longer exists. Could you please rebase your changes?

leosvelperez (changes requested):
Thank you for raising this! Hopefully, we can address these issues soon. I left a couple of comments to address.

packages/workspace/src/generators/preset/__snapshots__/preset.spec.ts.snap
robertIsaac: done

packages/nx/src/generators/tree.ts
robertIsaac: done

packages/angular/src/generators/utils/storybook-ast/component-info.ts
robertIsaac: done

robertIsaac:
@leosvelperez all comments are addressed now

leosvelperez (approved):
Thank you!

PR #27491, opened 2024-08-16
fix: missing export for setRemoteDefinition

menosprezzi:
## Current Behavior
It's not possible to import `setRemoteDefinition` from `@nx/angular/mf`.

## Expected Behavior
Export `setRemoteDefinition` in `@nx/angular/mf`


menosprezzi:
Hello @Coly010, could take a quick look in this small PR? It a fix from #27051

Coly010 (approved):

PR #27492, opened 2024-08-16
fix(module-federation): ssr uses async-node with runtime plugin

Coly010:
- fix(module-federation): SSR should work correctly
- fix(module-federation): ssr-dev-servers should wait for remotes before starting final executor

## Related Issue(s)

Fixes #


packages/react/src/executors/module-federation-ssr-dev-server/module-federation-ssr-dev-server.impl.ts
```diff
@@ -76,9 +76,14 @@ function startSsrStaticRemotesFileServer(
   ssrStaticRemotesConfig: StaticRemotesConfig,
   context: ExecutorContext,
   options: ModuleFederationSsrDevServerOptions
-) {
+):
+  | AsyncGenerator<{ success: boolean; baseUrl?: string }>
+  | AsyncIterable<{ success: boolean; baseUrl?: string }> {
   if (ssrStaticRemotesConfig.remotes.length === 0) {
-    return;
+    return createAsyncIterable(({ next, done }) => {
```
jaysoo, line 83: This could be simplified if `startSsrStaticRemotesFileServer` is an async generator.

```
if (ssrStaticRemotesConfig.remotes.length === 0) {
  yield { success: true};
  return;
}
```

But then `return staticRemotesIter` needs to be `yield* staticRemotesIter`.

jaysoo (approved):

FrozenPandaz (approved):

PR #27496, opened 2024-08-17
feat(core): expose graph json type

### LINKED ISSUE #3283: Provide types for JSON output of `--graph`
## Description
The commands which generate a graph do not have a type that can be imported so that consumers can know the type of the data. The data is created in this function: https://github.com/nrwl/nx/blob/8cf69c4b9abb2194ee2bbe374b6db97a8dce857b/packages/nx/src/command-line/graph/graph.ts#L1164

## Motivation

Nx is great and the CLI may be enough in some cases.

But when setting up a monorepo some people/company may need to have custom scripts running and thanks to NX we can get a JSON object which represents what has been affected.

On our side we need this and we ended up typing what we needed from the output.

I think it may be quite nice to have that exported by NX so that we can safely consume the JSON output of nx affected commands **even we we update NX version**. 

## Revisions

This issue was originally created for `print-affected` which has been deprecated. I have revised it for the newer commands.

ghost:
Closes #3283

## Current Behavior
See #3283

## Expected Behavior
See #3283

## Related Issue(s)
#3283

Fixes #
-  expose graph json type

### DIFF
--- a/docs/generated/devkit/GraphJson.md
+++ b/docs/generated/devkit/GraphJson.md
@@ -1,0 +1,27 @@
+# Interface: GraphJson
+
+## Table of contents
+
+### Properties
+
+- [graph](../../devkit/documents/GraphJson#graph): ProjectGraph
+- [taskPlans](../../devkit/documents/GraphJson#taskplans): Record<string, string[]>
+- [tasks](../../devkit/documents/GraphJson#tasks): TaskGraph
+
+## Properties
+
+### graph
+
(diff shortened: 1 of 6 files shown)


paweltwardziak:
@FrozenPandaz and @maxime1992 FYI

FrozenPandaz (approved):

FrozenPandaz:
Thank you for your contribution! I switched it to export from `@nx/devkit`.

PR #27515, opened 2024-08-19
fix(core): not parse hex to number

### LINKED ISSUE #26889: Unexpected number conversion of command line args
### Current Behavior

When passing in something number-like as an arg to a task, it seems that Nx converts it to a number. For example:

```bash
GIT_COMMIT="81919e4"
echo "Building with tag $GIT_COMMIT"
nx run app:docker-build --tag $GIT_COMMIT --repo org/repo
```

I get the following results:

```
Building with tag 81919e4

 NX   Running target docker-build for project app and 20 tasks it depends on:

> nx run dep1:build  [local cache]
...
> nx run app:docker-build --tag 819190000 --repo org/repo
```

Note that Nx interpreted the git commit as numeric and then expanded the `e4` to add `0000` at the end.

### Expected Behavior

Even if the values are numeric, they should not be modified when passing them to the task. So above...

```
Building with tag 81919e4

 NX   Running target docker-build for project app and 20 tasks it depends on:

> nx run dep1:build  [local cache]
...
> nx run app:docker-build --tag 81919e4 --repo org/repo
```

### GitHub Repo

https://github.com/czycha/nx-number-conversion-bug

### Steps to Reproduce

1. Install deps.
2. Run `nx build app --tag 81919e4` in your command line.
    - Bad behavior: echos "819190000"
    - Expected behavior:  echos "81919e4"

### Nx Report

```shell
Node   : 21.7.2
OS     : darwin-arm64
npm    : 10.5.0

nx (global)    : 19.4.2
nx             : 19.4.2
@nx/js         : 19.4.2
@nx/workspace  : 19.4.2
@nx/devkit     : 19.4.2
@nrwl/tao      : 19.4.2
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

xiongemi:
## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/26889

### DIFF
--- a/packages/nx/src/command-line/yargs-utils/shared-options.ts
+++ b/packages/nx/src/command-line/yargs-utils/shared-options.ts
@@ -1,8 +1,17 @@
-import { Argv } from 'yargs';
+import { Argv, ParserConfigurationOptions } from 'yargs';
 
 interface ExcludeOptions {
   exclude: string[];
 }
+
+export const defaultYargsParserConfiguration: Partial<ParserConfigurationOptions> =
+  {
+    'strip-dashed': true,
+    'unknown-options-as-args': true,
+    'populate--': true,
+    'parse-numbers': false,
+    'parse-positional-numbers': false,
+  };
 
 export function withExcludeOption(yargs: Argv): Argv<ExcludeOptions> {
   return yargs.option('exclude', {
@@ -151,11 +160,7 @@ export function withBatch(yargs: Argv) {
 
 export function withAffectedOptions(yargs: Argv) {
   return withExcludeOption(yargs)
-    .parserConfiguration({
-      'strip-dashed': true,
-      'unknown-options-as-args': true,
-      'populate--': true,
-    })
+    .parserConfiguration(defaultYargsParserConfiguration)
     .option('files', {
       describe:
         'Change the way Nx is calculating the affected command by providing directly changed files, list of files delimited by commas or spaces',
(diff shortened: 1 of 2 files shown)


packages/nx/src/command-line/yargs-utils/shared-options.ts
```diff
@@ -214,6 +216,8 @@ export function withRunManyOptions<T>(
       'strip-dashed': true,
       'unknown-options-as-args': true,
       'populate--': true,
+      'parse-numbers': false,
+      'parse-positional-numbers': false,
```
FrozenPandaz, line 220: I feel like other stuff will break with this... what about like `--maxWarnings 5` or `--parallel 8`? :eyes: 

Did you check those?

FrozenPandaz (changes requested):

packages/nx/src/command-line/yargs-utils/shared-options.ts
xiongemi: i test it out:
- for yargs option with type number (e.g. `maxParallel`), it will be converted as number, `'parse-numbers': false,` will be applied to for non-nx override options (e.g. `_: [ '--max-warnings', '4', '--tag', '81919e4' ],`
- for parallel, the yargs option is string, however, there is a function `readParallelFromArgsAndEnv` that will do the parsing (https://github.com/nrwl/nx/blob/add4183d0a9acebd4ba7e28674a4597281c55828/packages/nx/src/utils/command-line-utils.ts#L189)
- for the non nx args that is going to pass to executor, there is additional parsing in function createOverrides (https://github.com/nrwl/nx/blob/add4183d0a9acebd4ba7e28674a4597281c55828/packages/nx/src/utils/command-line-utils.ts#L42) which does not have `'parse-numbers': false,`, so it will parse the args like `maxWarnings` to number

in summary, this `'parse-numbers': false,` will only be applied to non-nx override options, but not when it got passed to executor

FrozenPandaz (approved):

PR #27525, opened 2024-08-19
fix(react): assert test property is defined on webpack rule in nx-react-webpack-plugin

nigelwtf:
The `removeSvgLoaderIfPresent` method in `apply-react-config.ts` iterates through each Webpack rule, calls `toString` on each `test` property, and checks for the presence of the string `"svg"` to see if any existing SVG plugins need to be removed.

Some of the Webpack rules in the config don't have the test property, and calling `toString` without asserting the test property is defined first throws type errors.

This commit introduces a very small change that simply asserts that `rule.test` is not `undefined` before calling `.toString()`.

## Current Behavior
Running a React Webpack library with `svgr` enabled causes compilation errors.

Screenshot 2024-08-20 at 10 36 09 AM

Screenshot 2024-08-20 at 10 39 28 AM

## Expected Behavior
Enabling SVGR in the `NxReactWebpackPlugin` config should compile as normal.

## Related Issue(s)
N/A

### DIFF
--- a/packages/react/plugins/nx-react-webpack-plugin/lib/apply-react-config.ts
+++ b/packages/react/plugins/nx-react-webpack-plugin/lib/apply-react-config.ts
@@ -101,7 +101,10 @@ function removeSvgLoaderIfPresent(
   config: Partial<WebpackOptionsNormalized | Configuration>
 ) {
   const svgLoaderIdx = config.module.rules.findIndex(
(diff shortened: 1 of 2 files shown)


nigelwtf:
Hey @ndcunningham - hope you're doing well.

Not sure how big your backlog is atm—just wondering if you'd had a chance to peek at this yet?

Coly010 (approved):

PR #27530, opened 2024-08-20
fix(js): only sync references when composite is true, preserve comments in other parts of file

JamesHenry:
## Current Behavior

During references syncing the experimental (not yet meant for public consumption) typescript plugin nukes the whole tsconfig, meaning user comments are always lost. Additionally, there is no evaluation of config performed, meaning invalid references are created to projects which do not have `composite: true` applied to them (whether directly or through config inheritance via extends).

## Expected Behavior

We use AST based patching to preserve comments in other parts of the tsconfig files we update, and we only create references to projects with `composite: true`.

## Related Issue(s)

Fixes #


packages/js/package.json
JamesHenry, line 1: The only change here is the addition of `jsonc-parser` (which is already a dependency of the `nx` package), it just bugged me how badly unsorted these deps where so I am ordered them at the same time

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -204,16 +249,57 @@ export async function syncGenerator(tree: Tree): Promise<SyncGeneratorResult> {
...
+ */
+function readRawTsconfigContents(
+  tree: Tree,
+  rawTsconfigContentsCache: Map<string, string>,
+  tsconfigPath: string
+): string {
+  if (!rawTsconfigContentsCache.has(tsconfigPath)) {
+    rawTsconfigContentsCache.set(
+      tsconfigPath,
+      tree.read(tsconfigPath, 'utf-8').toString()
```
leosvelperez, line 264:
```suggestion
      tree.read(tsconfigPath, 'utf-8')
```

packages/js/src/generators/typescript-sync/typescript-sync.spec.ts
```diff
@@ -214,6 +256,57 @@ describe('syncGenerator()', () => {
...
+            "target": "es5"
+          },
+          "references": [{ "path": "./packages/a" }, { "path": "./packages/b" }]
+        }
+        "
+      `);
+    });
+
+    it('should not add a reference if the internally referenced tsconfig.json does not have composite: true', async () => {
+      // Delete composite from a, causing it to now show up in the final tsconfig.json snapshot
```
leosvelperez, line 292:
```suggestion
      // Delete composite from a, causing it to not show up in the final tsconfig.json snapshot
```

packages/js/src/generators/typescript-sync/typescript-sync.spec.ts
```diff
@@ -352,6 +451,78 @@ describe('syncGenerator()', () => {
...
+        }
+        "
+      `);
+    });
+
+    it('should not add a reference if the dependency tsconfig.json does not have composite: true', async () => {
+      addProject('foo', ['bar'], ['tsconfig.build.json']);
+      addProject('bar', [], ['tsconfig.build.json']);
+
+      // Delete composite from bar, causing it to now show up in the final tsconfig.json snapshots below
```
leosvelperez, line 493:
```suggestion
      // Delete composite from bar, causing it to not show up in the final tsconfig.json snapshots below
```

leosvelperez (approved):
Just left some minor nits. Feel free to merge it without them.

PR #27552, opened 2024-08-20
fix(js): allow syncing references under solution configs, regardless of composite

JamesHenry:
## Related Issue(s)

Fixes #


JamesHenry:
Closing for now pending further real world usage

PR #27572, opened 2024-08-21
fix(nx-plugin): allow create-package without e2eProject

isaacplmann:
Current behavior:
`nx g create-package create-my-plugin --project my-plugin` fails because `e2eProject` defaults to the root project.

Expected behavior:
`nx g create-package create-my-plugin --project my-plugin` succeeds.

### DIFF
--- a/packages/plugin/src/generators/create-package/schema.json
+++ b/packages/plugin/src/generators/create-package/schema.json
@@ -67,7 +67,8 @@
     },
     "e2eProject": {
       "type": "string",
-      "description": "The name of the e2e project."
+      "description": "The name of the e2e project.",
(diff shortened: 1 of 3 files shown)


packages/plugin/src/generators/create-package/schema.json
```diff
@@ -67,12 +67,7 @@
     },
     "e2eProject": {
       "type": "string",
-      "description": "The name of the e2e project.",
-      "alias": "p",
-      "$default": {
-        "$source": "projectName"
-      },
-      "x-prompt": "What is the name of the e2e project?"
+      "description": "The name of the e2e project."
```
jaysoo, line 70: Should we keep the prompt and just say `Leave blank to skip e2e tests`?

jaysoo (approved):

PR #27621, opened 2024-08-23
fix(webpack): add migration for proxy config

jaysoo:
Webpack-dev-server v5 changed the proxy config to only accept arrays. This PR adds a migration for JSON proxy config files to automatically do the conversion for users. If it's not a JSON format, then we log a warning and point them to the v5 migration page.

## Related Issue(s)

Fixes #


packages/webpack/src/migrations/update-19-6-3/proxy-config.ts
```diff
@@ -0,0 +1,55 @@
...
+            context: [context],
+            ...json[context],
+          }));
+        }
+        return json;
+      });
+    } else {
+      // For non-JSON files, it's not possible to automatically update the proxy config
+      // since its content can vary greatly.
+      logger.warn(`${options.proxyConfig} needs to be updated manually.
```
leosvelperez, line 22: This could be noisy if logged for multiple proxy config files. It might be better to aggregate the proxy config file paths and log the warning once at the end, listing them.

ndcunningham (approved):

PR #27634, opened 2024-08-26
fix(core): nx migrate should determine angular devkit migration from cli or schematics #20282

### LINKED ISSUE #20282: Cannot run Angular library migration in Nx workspace
### Current Behavior

We have an internal Angular library with migration scripts for use with `ng update`. The scripts work fine in a standard Angular workspace.

When run though `nx migrate` however, the arguments to the script are always undefined. Just like in this StackOverflow post https://stackoverflow.com/questions/76860116/tree-is-undefined-angular-update-schematics-in-a-nx-workspace. Running with FORCE_NG_UPDATE=true results in "Error: This command is not available when running the Angular CLI outside a workspace."

To be clear; the `migrations.json` file is generated and appears to run without errors, it just doesn't actually do anything do to the undefined argument.

As not everyone using this library is using Nx, we can't make an Nx generator instead; and it seems awkward to maintain both an Angular and Nx version of the migration script. The consuming project is currently on Nx 16.3.1.

### Expected Behavior

The library's migration script to run correctly in Nx workspaces.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create Angular library migration with standard structure, ie:
```
export function version4(_options: any): Rule {
  return (host: Tree, context: SchematicContext) => {
    const version = `4.0.0`;
    context.logger.info(`Applying migration for lib to version ${version}`);
    ...
    return host;
  };
```
2. Run in Nx workspace with `nx migrate`. Notice not even the log statement works (though a standard console.log does, its definitely calling the function)

### Nx Report

```shell
Node   : 16.20.1
   OS     : darwin arm64
   yarn   : 1.22.19
   Hasher : Node
   
   nx                 : 16.3.1
   @nx/js             : 16.3.2
   @nrwl/js           : 16.3.1
   @nx/jest           : 16.3.1
   @nrwl/jest         : 16.3.2
   @nx/linter         : 16.3.1
   @nrwl/linter       : 16.3.2
   @nx/workspace      : 16.3.1
   @nx/angular        : 16.3.1
   @nx/cypress        : 16.3.1
   @nx/devkit         : 16.3.2
   @nx/eslint-plugin  : 16.3.1
   @nx/plugin         : 16.3.2
   @nrwl/tao          : 13.1.3
   @nx/web            : 16.3.1
   @nx/webpack        : 16.3.1
   typescript         : 5.0.4
   Community plugins:
   @custom/rollout              : 4.0.0
   @ngrx/component              : 16.0.1
   @ngrx/component-store        : 16.0.1
   @ngrx/eslint-plugin          : 16.0.1
   Local workspace plugins:
         @custom/opex-nx
     - @nx/js@16.3.2
     - @nrwl/jest@16.3.2
     - @nrwl/linter@16.3.2
     - @nx/devkit@16.3.2
     - @nrwl/devkit@16.3.2
     - @nx/plugin@16.3.2
     - @nrwl/nx-plugin@16.3.2
     - @nrwl/tao@13.1.3
   
   To fix this, run `nx migrate nx@16.3.2`
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Coly010:
## Current Behavior

We check the CLI property exists and if it's not `nx` we will use the ng compat layer to run the migration.
If the CLI property does not exist, we check both if the migration is in the `schematics` object on the `migrations.json` and if the contents of the migration implementation contains an import from `@angular-devkit`.

The problem with the fallback is that if only types are imported from `@angular-devkit` the import is stripped from the migration implementation completely.

## Expected Behavior

We had planned already to remove the fallback of reading the file contents.
We had also planned to remove using the `cli` property to determine if the migration needed the `ng compat layer`. 

However, as the `cli` property is still useful for now for package's that needed some manner to circumvent the flawed fallback logic, let's continue to use it until v21.
Log a warning however if the `cli !== 'nx'` and it is placed in the `generators` section of the `migrations.json` to provide ample time for plugin developers to move them to the `schematics` property.

Fallback has been updated to whether or not the migration lives in `schematics` and not flawed read file logic.

## Related Issue(s)

Fixes #20282

### DIFF
--- a/packages/nx/src/command-line/migrate/migrate.ts
+++ b/packages/nx/src/command-line/migrate/migrate.ts
@@ -1699,62 +1699,21 @@ function getImplementationPath(
   return { path: implPath, fnSymbol };
 }
 
-// TODO (v17): This should just become something like:
-// ```
-// return !collection.generators[name] && collection.schematics[name]
-// ```
+// TODO (v21): Remove CLI determination of Angular Migration
 function isAngularMigration(
   collection: MigrationsJson,
   collectionPath: string,
   name: string
 ) {
   const entry = collection.generators?.[name] || collection.schematics?.[name];
-
-  // In the future we will determine this based on the location of the entry in the collection.
-  // If the entry is under `schematics`, it will be assumed to be an angular cli migration.
-  // If the entry is under `generators`, it will be assumed to be an nx migration.
-  // For now, we will continue to obey the cli property, if it exists.
-  // If it doesn't exist, we will check if the implementation references @angular/devkit.
   const shouldBeNx = !!collection.generators?.[name];
   const shouldBeNg = !!collection.schematics?.[name];
-  let useAngularDevkitToRunMigration = false;
-
-  const { path: implementationPath } = getImplementationPath(
-    collection,
-    collectionPath,
-    name
-  );
-  const implStringContents = readFileSync(implementationPath, 'utf-8');
-  // TODO (v17): Remove this check and the cli property access - it is only here for backwards compatibility.
-  if (
-    ['@angular/material', '@angular/cdk'].includes(collection.name) ||
-    [
-      "import('@angular-devkit",
-      'import("@angular-devkit',
-      "require('@angular-devkit",
-      'require("@angular-devkit',
-      "from '@angular-devkit",
-      'from "@angular-devkit',
-    ].some((s) => implStringContents.includes(s))
-  ) {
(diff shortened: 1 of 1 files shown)


FrozenPandaz (approved):
@leosvelperez can you look at this one as well please?

leosvelperez (approved):

PR #27636, opened 2024-08-26
fix(js): keep refs to ignored files and allow opting out of pruning stale refs in typescript sync generator

leosvelperez:
## Current Behavior

When the `@nx/js:typescript-sync` generator updates a tsconfig file of an Nx project, it prunes project references pointing to a path that's not part of the Nx project files (including paths inside a nested project) if they don't belong to a project that is a dependency.

## Expected Behavior

When the `@nx/js:typescript-sync` generator updates a tsconfig file of an Nx project, it should only prune project references pointing to a path that's not part of the Nx project files if it's not ignored and not part of a dependency.

Additionally, a new `nx.sync.ignoredReferences` top-level option in tsconfig files allows specifying custom project references that the `@nx/js:typescript-sync` sync generator should ignore and not prune.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/generators/typescript-sync/typescript-sync.spec.ts
+++ b/packages/js/src/generators/typescript-sync/typescript-sync.spec.ts
@@ -368,7 +368,7 @@ describe('syncGenerator()', () => {
         references: [
           { path: './some/thing' },
           { path: './another/one' },
-          { path: '../packages/c' }, // this is not a dependency, should be pruned
+          { path: '../c' }, // this is not a dependency, should be pruned
         ],
       });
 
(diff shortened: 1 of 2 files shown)


packages/js/src/generators/typescript-sync/typescript-sync.spec.ts
```diff
@@ -391,6 +391,95 @@ describe('syncGenerator()', () => {
...
 
+    it('should not prune existing external project references that are not dependencies but are ignored', async () => {
+      writeJson(tree, 'packages/b/tsconfig.json', {
+        compilerOptions: {
+          composite: true,
+        },
+        references: [
+          { path: './some/thing' },
+          { path: './another/one' },
+          { path: '../../some-path/dir' }, // this is not a dependency but it's ignored, should not be pruned
```
JamesHenry, line 402: Given you need to update this file anyway, please could you make it clear this means git ignored? (I was looking for nx.json updates in the test when I read this)

JamesHenry (changes requested):
Left some comments on the test file

FrozenPandaz (approved):

JamesHenry (approved):

PR #27638, opened 2024-08-26
feat(core): allow disabling registered task sync generators

leosvelperez:
## Current Behavior

There's no way to disable sync generators registered in inferred tasks by a Crystal plugin.

## Expected Behavior

There should be a way to disable sync generators registered in inferred tasks by a Crystal plugin.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/project-graph/utils/project-configuration-utils.ts
+++ b/packages/nx/src/project-graph/utils/project-configuration-utils.ts
@@ -635,6 +635,9 @@ function normalizeTargets(
   sourceMaps: ConfigurationSourceMaps,
   nxJsonConfiguration: NxJsonConfiguration<'*' | string[]>
 ) {
+  const disabledTaskSyncGenerators = new Set(
+    nxJsonConfiguration.sync?.disabledTaskSyncGenerators ?? []
+  );
   for (const targetName in project.targets) {
     project.targets[targetName] = normalizeTarget(
       project.targets[targetName],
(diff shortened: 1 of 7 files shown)


packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -679,6 +682,17 @@ function normalizeTargets(
...
+    if (
+      disabledTaskSyncGenerators.size &&
+      project.targets[targetName].syncGenerators?.length
+    ) {
+      project.targets[targetName].syncGenerators = project.targets[
+        targetName
+      ].syncGenerators.filter(
+        (syncGenerator) => !disabledTaskSyncGenerators.has(syncGenerator)
+      );
+    }
```
FrozenPandaz, line 695: Should we keep it in the graph but not run them? :thinking:

FrozenPandaz (changes requested):

packages/nx/src/project-graph/utils/project-configuration-utils.ts
leosvelperez: I considered that, but I don't see the need or much benefit in doing so. The only thing I can think of is that keeping them in the graph would allow us to display them in the PDV but somehow show that they are disabled. I see little value in that, and keeping them in the graph requires handling the `disabledTaskSyncGenerators` option in multiple places.

Could you explain why you think we should keep them in the graph?

FrozenPandaz (approved):

PR #27646, opened 2024-08-27
fix(expo): add cacheVersion to metro.config

### LINKED ISSUE #27392: metro web builds cache key collision
### Current Behavior

After upgrading to Expo SDK 51, and adpting expo web projects from webpack to metro web, we noticed that sometimes one project's build would have environment variables from a different project in the monorepo.

The issue was intermittent and only happened in pipelines with multiple web builds.

We determined that — if the `metro.config.js` file is **identical** between two or more projects — then **it uses the same cache key**. And when multiple projects are being built — after the first project is completed — subsequent projects will resolve the first project's compiled/cached app config (etc) from metro cache — because they have the same cache-key.

> [!NOTE]
> This is also true for `start` commands, and (presumably) other build related commands. 

_EDIT: updated to improve clarity_

### Expected Behavior

The builds should be isolated and config from one build should not leak into subsequent builds. 

### GitHub Repo

https://github.com/leggomuhgreggo/metro-web-cache-issue-2024-08-09

### Steps to Reproduce

Follow instructions in README.

### Nx Report

```shell
Node           : 18.20.2
OS             : darwin-arm64
yarn           : 1.22.22

nx                 : 19.5.6
@nx/js             : 19.5.6
@nx/jest           : 19.5.6
@nx/linter         : 19.5.6
@nx/eslint         : 19.5.6
@nx/workspace      : 19.5.6
@nx/cypress        : 19.5.6
@nx/detox          : 19.5.6
@nx/devkit         : 19.5.6
@nx/esbuild        : 19.5.6
@nx/eslint-plugin  : 19.5.6
@nx/expo           : 19.5.6
@nx/nest           : 19.5.6
@nx/node           : 19.5.6
@nx/playwright     : 19.5.6
@nx/plugin         : 19.5.6
@nx/react          : 19.5.6
@nx/storybook      : 19.5.6
@nrwl/tao          : 19.5.6
@nx/vite           : 19.5.6
@nx/web            : 19.5.6
@nx/webpack        : 19.5.6
typescript         : 5.5.3
Community plugins:
@nx-go/nx-go           : 2.7.0
@nx-tools/nx-container : 5.1.0
@nxkit/playwright      : 3.0.2
@nxrs/cargo            : 0.3.4
Local workspace plugins:
         @workspace/sg-codepipeline-plugin
         @workspace/sg-codedeploy-plugin
         @workspace/sg-typeorm-plugin
         @workspace/sg-rover-plugin
         @workspace/sg-expo-plugin
         @workspace/repo
         @workspace/sg-vite-plugin
         @sg-monorepo/workspace-plugin
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I've also opened up an issue on the expo repo here:
https://github.com/expo/expo/issues/30930

xiongemi:
## Current Behavior

expo tends to cache the output if there is no change in metro.config.js
we don't need to explicitly set the args --clear when https://github.com/expo/expo/issues/30930 is fixed

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/27392

### DIFF
--- a/packages/expo/plugins/plugin.ts
+++ b/packages/expo/plugins/plugin.ts
@@ -104,7 +104,7 @@ function buildExpoTargets(
     },
     [options.serveTargetName]: {
       command: `expo start --web`,
-      options: { cwd: projectRoot },
+      options: { cwd: projectRoot, args: ['--clear'] },
     },
     [options.runIosTargetName]: {
       command: `expo run:ios`,
@@ -116,7 +116,7 @@ function buildExpoTargets(
     },
     [options.exportTargetName]: {
       command: `expo export`,
-      options: { cwd: projectRoot },
+      options: { cwd: projectRoot, args: ['--clear'] },
       cache: true,
       dependsOn: [`^${options.exportTargetName}`],
       inputs: getInputs(namedInputs),
--- a/packages/expo/src/generators/application/files/base/metro.config.js.template
+++ b/packages/expo/src/generators/application/files/base/metro.config.js.template
@@ -12,6 +12,7 @@ const { assetExts, sourceExts } = defaultConfig.resolver;
  * @type {import('metro-config').MetroConfig}
  */
 const customConfig = {
+  cacheVersion: "<%= projectName %>",
   transformer: {
     babelTransformerPath: require.resolve('react-native-svg-transformer'),
   },
--- a/packages/react-native/src/generators/application/files/app/metro.config.js.template
+++ b/packages/react-native/src/generators/application/files/app/metro.config.js.template
@@ -11,6 +11,7 @@ const { assetExts, sourceExts } = defaultConfig.resolver;
  * @type {import('metro-config').MetroConfig}
  */
 const customConfig = {
+  cacheVersion: "<%= projectName %>",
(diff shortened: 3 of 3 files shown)


packages/expo/plugins/plugin.ts
```diff
@@ -116,7 +116,7 @@ function buildExpoTargets(
     },
     [options.exportTargetName]: {
       command: `expo export`,
-      options: { cwd: projectRoot },
+      options: { cwd: projectRoot, args: ['--clear'] },
```
leggomuhgreggo, line 119: Wanted to note that this issue also occurs with `expo start`, and presumably other build related commands.

I've added a [comment](https://github.com/nrwl/nx/issues/27392#issuecomment-2311522488) to the issue with additional details and a potential solution

xiongemi: i think your solution is better. i will change my pr

jaysoo (approved):

PR #27654, opened 2024-08-27
feat(graph): enable composite graph functionality

nartc:
This PR enables composite graph functionality:
- Experimental feature to enable Composite Graph
- In Composite Graph mode:
  - Nodes are shown by default.
  - Show/Hide All Projects function similarly to regular mode
  - Focus a Composite Node renders the inner nodes with up-to 3 additional containers: Green area contains external nodes that depend on the inner nodes; Orange area contains external nodes that the inner nodes depend depend on; Purple area contains external nodes with circular dependencies with the inner nodes.
    - Focused node can be unfocus/reset.
    - Only one node can be focused at one given time. - Show All projects while having a focused node will unfocus the node.
  - Expand a Composite Node renders the inner nodes of the composite node in-place (i.e: still keep the context of the current graph). Expanded node can be collapsed to go back.
- Styling concerns: please post styling concerns in this ticket https://linear.app/nxdev/issue/APP-2027/composite-graph-styling-feedbacks. We're planning to restyle the graph component itself (nodes, edges, labels) according to Nx Cloud styles then expand to other platform via customizations.

### DIFF
--- a/graph/client/src/app/feature-projects/machines/composite-graph.state.ts
+++ b/graph/client/src/app/feature-projects/machines/composite-graph.state.ts
@@ -34,6 +34,43 @@ export const compositeGraphStateConfig: ProjectGraphStateNodeConfig = {
     }),
   ],
   on: {
+    selectAll: {
+      actions: [
+        assign((ctx, event) => {
+          if (event.type !== 'selectAll') return;
+          ctx.compositeGraph.enabled = true;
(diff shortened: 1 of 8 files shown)


FrozenPandaz (changes requested):
Routes in the projects sidebar for focusing are not working correctly.

PR #27678, opened 2024-08-28
feat(nx-cloud): add 'generate-token' option to connect

bcabanes:
This update introduces a new 'genreate-token' option to force local token creation. It ensures tokens are not created for GitHub-based setups unless explicitly overridden. Adjustments include updated logic in the generator function and schema to accommodate the new option.

### DIFF
--- a/packages/nx/src/nx-cloud/generators/connect-to-nx-cloud/schema.json
+++ b/packages/nx/src/nx-cloud/generators/connect-to-nx-cloud/schema.json
@@ -21,6 +21,10 @@
       "description": "Hide formatting logs",
       "x-priority": "internal"
     },
+    "generateToken": {
+      "type": "boolean",
+      "description": "Explicitly asks for a token to be created, do not override existing tokens from Nx Cloud"
+    },
     "github": {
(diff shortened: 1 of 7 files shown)


bcabanes:
## Behaviour

https://github.com/user-attachments/assets/4bcd14a2-733e-4129-8aad-7cff0e9591fc

packages/nx/src/nx-cloud/generators/connect-to-nx-cloud/schema.json
```diff
@@ -21,6 +21,11 @@
       "description": "Hide formatting logs",
       "x-priority": "internal"
     },
+    "token": {
+      "type": "boolean",
+      "description": "Force local token creation",
+      "x-priority": "user"
```
jaysoo, line 27: We can remove this. We set `important` to surface it higher in Nx Console, or `hidden` to hide it from Nx Console. It's not used for anything else.

jaysoo (approved):

FrozenPandaz (approved):

PR #27685, opened 2024-08-28
fix(core): fixes an issue where using "node:test" results in infinite loop

jaysoo:
This PR fixes an issue with `node:test` not being recognized as a built-in Node module, and since it is not in `node_modules` with a resolvable `package.json`, it hits an infinite loop.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
@@ -34,6 +34,13 @@ const defaultNpmResolutionCache: NpmResolutionCache = new Map();
 const builtInModuleSet = new Set<string>([
   ...builtinModules,
   ...builtinModules.map((x) => `node:${x}`),
+  // These are missing in the builtinModules list
+  // See: https://github.com/nodejs/node/issues/42785
+  // TODO(v20): We should be safe to use `isBuiltin` function instead of keep the set here (https://nodejs.org/api/module.html#moduleisbuiltinmodulename)
+  'test',
+  'node:test',
+  'node:sea',
+  'node:sqlite',
 ]);
 
 export function isBuiltinModuleImport(importExpr: string): boolean {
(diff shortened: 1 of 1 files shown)


packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -364,7 +369,11 @@ export class TargetProjectLocator {
         packageJsonPath ?? resolveRelativeToDir(packageName, relativeToDir);
       let dir = dirname(pathOfFileInPackage);
 
-      while (dir !== parse(dir).root) {
+      while (
+        // If we're at the root, then parse(dir).root is empty (''), so handle it separately.
+        dir !== '.' &&
+        dir !== parse(dir).root
```
FrozenPandaz, line 375: Can we get rid of this `parse` call.. let's use `dirname`.. 

`dir === dirname(dir)` will be true when it's reached the root. That's what we use everywhere else.

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -34,6 +34,11 @@ const defaultNpmResolutionCache: NpmResolutionCache = new Map();
 const builtInModuleSet = new Set<string>([
   ...builtinModules,
   ...builtinModules.map((x) => `node:${x}`),
+  // These are missing in the builtinModules list
+  // See: https://github.com/nodejs/node/issues/42785
+  // TODO(v20): We should be safe to use `isBuiltin` function instead of keep the set here (https://nodejs.org/api/module.html#moduleisbuiltinmodulename)
+  'test',
```
FrozenPandaz, line 40: Let's add a unit test for `test` and `node:test`.

Can you add `sqlite` and `sea` while you're at it?

FrozenPandaz (changes requested):

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
jaysoo: Sure, `node:sea` exists, `sea` does not so I'll add the former.

jaysoo: Same with `node:sqlite`.

FrozenPandaz (approved):

PR #27688, opened 2024-08-29
fix(vite): make typechecking target declare typescript as an externalDependancies for better caching

beaussan:
## Current Behavior

The inputs of the typecheck target of the `@nx/vite/plugin` infered task have `allExternalDependancies` in their input, meaning the whole graph of npm dependancies is their input.

This was picked up by our cache optimiser that detect any cached run commands / run scripts without `externalDependencies` in their input, making the cache way less effective.

## Expected Behavior

The input of the typecheck target behave like the one from the `tsc` plugin and only is affected when `typescript` change and if the code changes.

### DIFF
--- a/packages/vite/src/plugins/__snapshots__/plugin-vitest.spec.ts.snap
+++ b/packages/vite/src/plugins/__snapshots__/plugin-vitest.spec.ts.snap
@@ -53,6 +53,11 @@ exports[`@nx/vite/plugin root project should create nodes 1`] = `
               "inputs": [
                 "production",
                 "^production",
+                {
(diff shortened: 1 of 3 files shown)


JamesHenry (approved):
Thank you @beaussan!

JamesHenry:
As sorry looks like there is a conflict to fix up first

beaussan:
I guess it was also fixed there https://github.com/nrwl/nx/pull/27922, so this pr is not needed anymore

JamesHenry:
Ah ok, thank you for confirming

PR #27705, opened 2024-08-30
feat(core): split runCommand to return status per project

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/init-tasks-runner.ts
+++ b/packages/nx/src/tasks-runner/init-tasks-runner.ts
@@ -24,7 +24,7 @@ export async function initTasksRunner(nxArgs: NxArgs) {
       tasks: Task[];
       parallel: number;
     }): Promise<{
-      status: number;
+      status: NodeJS.Process['exitCode'];
       taskGraph: TaskGraph;
       taskResults: Record<string, TaskResult>;
     }> => {
@@ -51,7 +51,7 @@ export async function initTasksRunner(nxArgs: NxArgs) {
         }, {} as any),
       };
 
-      const status = await invokeTasksRunner({
+      const taskResults = await invokeTasksRunner({
         tasks: opts.tasks,
         projectGraph,
         taskGraph,
@@ -63,9 +63,14 @@ export async function initTasksRunner(nxArgs: NxArgs) {
       });
 
       return {
-        status,
+        status: Object.values(taskResults).some(
+          (taskResult) =>
+            taskResult.status === 'failure' || taskResult.status === 'skipped'
+        )
+          ? 1
+          : 0,
         taskGraph,
-        taskResults: lifeCycle.getTaskResults(),
+        taskResults,
       };
     },
   };
--- a/packages/nx/src/tasks-runner/run-command.ts
+++ b/packages/nx/src/tasks-runner/run-command.ts
@@ -4,7 +4,6 @@ import { join } from 'path';
 import {
   NxJsonConfiguration,
   readNxJson,
-  TargetDefaults,
   TargetDependencies,
 } from '../config/nx-json';
 import { ProjectGraph, ProjectGraphProjectNode } from '../config/project-graph';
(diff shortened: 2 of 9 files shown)


packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -169,49 +169,76 @@ export async function runCommand(
...
+export async function runCommandForTasks(
+  projectsToRun: ProjectGraphProjectNode[],
+  currentProjectGraph: ProjectGraph,
+  { nxJson }: { nxJson: NxJsonConfiguration },
+  nxArgs: NxArgs,
+  overrides: any,
+  initiatingProject: string | null,
+  extraTargetDependencies: Record<string, (TargetDependencyConfig | string)[]>,
+  extraOptions: { excludeTaskDependencies: boolean; loadDotEnvFiles: boolean }
+): Promise<{ [id: string]: TaskStatus }> {
```
FrozenPandaz, line 203: Can we make this `TaskResult`?

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -47,6 +46,7 @@ import { TasksRunner, TaskStatus } from './tasks-runner';
 import { shouldStreamOutput } from './utils';
 import chalk = require('chalk');
 import { IS_WASM } from '../native';
+import { Observable } from 'rxjs';
```
FrozenPandaz, line 49: This should be import type

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -409,7 +436,7 @@ export async function invokeTasksRunner({
   nxArgs: NxArgs;
   loadDotEnvFiles: boolean;
   initiatingProject: string | null;
-}) {
+}): Promise<{ [id: string]: TaskStatus }> {
```
FrozenPandaz, line 439: @xiongemi we can add a lifecycle to get the full TaskResult.

packages/nx/src/tasks-runner/init-tasks-runner.ts
```diff
@@ -51,7 +51,7 @@ export async function initTasksRunner(nxArgs: NxArgs) {
         }, {} as any),
       };
 
-      const status = await invokeTasksRunner({
+      const taskStatus = await invokeTasksRunner({
```
FrozenPandaz, line 54: Get this to return the task results. Then everything will be the same.

FrozenPandaz (changes requested):

packages/nx/src/tasks-runner/life-cycle.ts
```diff
@@ -124,4 +125,14 @@ export class CompositeLifeCycle implements LifeCycle {
...
+
+  getTaskResults(): Record<string, TaskResult> {
+    let results = {};
+    for (let l of this.lifeCycles) {
+      if (l.getTaskResults) {
+        results = { ...results, ...l.getTaskResults() };
+      }
+    }
+    return results;
+  }
```
FrozenPandaz, line 137: I don't think we need to add this here. It doesn't need to be on every life cycle. Just one.

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -571,12 +600,17 @@ export async function invokeTasksRunner({
     taskGraph,
     nxJson
   );
+  const compositedLifeCycle: LifeCycle = new CompositeLifeCycle(
+    constructLifeCycles(lifeCycle)
+  );
```
FrozenPandaz, line 605: This can be:

```suggestion
  const taskResultsLifecycle = new TaskResultsLifeCycle();
  const compositedLifeCycle: LifeCycle = new CompositeLifeCycle([
    ...constructLifeCycles(lifeCycle),
    taskResultsLifecycle
  ]);
```

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #27723, opened 2024-08-31
fix(misc): fix the URL of the @naxodev/nx-cloudflare plugin

vicb:
The former URL returns a 404


NachoVazquez:
Thank you @vicb ! I don't even remember when I made this change.

isaacplmann (approved):

PR #27732, opened 2024-09-03
docs(core): reword import help text

isaacplmann:
Updates the help text for the `nx import` command

### DIFF
--- a/packages/nx/src/command-line/import/command-object.ts
+++ b/packages/nx/src/command-line/import/command-object.ts
@@ -4,7 +4,7 @@ import { withVerbose } from '../yargs-utils/shared-options';
 import { handleErrors } from '../../utils/params';
 
 export const yargsImportCommand: CommandModule = {
-  command: 'import [sourceRepository] [destinationProjectPath]',
+  command: 'import [sourceRepository] [destinationDirectory]',
   describe: false,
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/import/command-object.ts
```diff
@@ -4,22 +4,22 @@ import { withVerbose } from '../yargs-utils/shared-options';
...
+          .positional('destinationProjectPath', {
             type: 'string',
             description:
               'The directory in the current workspace to import into',
           })
-          .option('source', {
+          .option('sourceProjectPath', {
             type: 'string',
             description:
               'The directory in the source repository to import from',
```
FrozenPandaz, line 25: Can we make these `destinationDirectory` and `sourceDirectory` respectively with aliases for their original names?

FrozenPandaz (changes requested):

packages/nx/src/command-line/import/command-object.ts
isaacplmann: All 4 paths are directories, so adding that word doesn't distinguish between the relative project path/directory and the root repository path/directory.

FrozenPandaz (approved):

PR #27761, opened 2024-09-04
feat(core): forward string array arguments

ThePlenkov:
Original discussion:
https://github.com/nrwl/nx/discussions/27760

## Current Behavior
Currently if I create a task like:
```
{
  "targets": {
    "build-cds": {
      "options": {
        "schema": "SCHEMA_NAME ",
        "filter": [
          "TABLE1",
          "TABLE1"
        ]
      }
    } 
  }
}
```

it will still call my command like: `mycommand --schema SCHEMA_NAME`

## Expected Behavior

After my change it would run `mycommand --schema SCHEMA_NAME --filter=TABLE1,TABLE2,TABLE3`

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -54,6 +54,7 @@ const propKeys = [
   'parallel',
   'no-parallel',
   'readyWhen',
+  'readyWhenStatus',
   'cwd',
   'args',
   'envFile',
@@ -238,7 +239,10 @@ function normalizeOptions(
 }
 
 function isArrayOfStrings(arg: unknown): arg is Array<string> {
-  return Array.isArray(arg) && arg.every((item) => typeof item === 'string');
+  if (Array.isArray(arg)) {
+    return arg.every((item) => typeof item === 'string');
+  }
+  return false;
 }
 
 export function interpolateArgsIntoCommand(
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -237,6 +237,10 @@ function normalizeOptions(
   return options as NormalizedRunCommandsOptions;
 }
 
+function isArrayOfStrings(arg: unknown): arg is Array<string> {
+  return Array.isArray(arg) && arg.every((item) => typeof item === 'string');
+}
+
 export function interpolateArgsIntoCommand(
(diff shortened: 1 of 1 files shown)


ThePlenkov:
I have also reached same goal by creating a plugin:
```ts
import {
  CreateNodesContextV2,
  createNodesFromFiles,
  CreateNodesV2,
  ProjectConfiguration,
  readJsonFile,
} from "@nx/devkit";
import { dirname } from "path";

interface Options {
  replace?: {
    regex?: string;
    with?: string;
  };
}

export const createNodesV2: CreateNodesV2<Options> = [
  "**/project.json",
  (projectConfigurationFiles, options, context) => {
    return createNodesFromFiles(
      (configFile, options, context) =>
        createNodesInternal(configFile, options, context),
      projectConfigurationFiles,
      options,
      context,
    );
  },
];

function createNodesInternal(
  configFilePath: string,
  options: Options,
  context: CreateNodesContextV2,
) {
  const root = dirname(configFilePath);
  const project = readJsonFile(configFilePath) as ProjectConfiguration;

  for (const targetName in project.targets) {
    const target = project.targets[targetName];
    for (const optionName in target.options) {
      const option = target.options[optionName];
      if (isArrayOfStrings(option)) {
        const targetOptionName = optionName.replace(/\[\]$/, "");
        target.options[targetOptionName] = option.toString();
      }
    }
  }

  return {
    projects: {
      [root]: project,
    },
  };
}

function isArrayOfStrings(arg: unknown): arg is Array<string> {
  if (Array.isArray(arg)) {
    return arg.every((item) => typeof item === "string");
  }
}
```

And then in configuration something like:
```
  "targets": {
    "build": {
      "options": {
        "schema": "MY_SCHEMA",
        "filter[]": [
          "Foo",
          "Bar"        
        ]
      }
    },
```

unfortunately with my apporach I had to introduce these array brackets in the name because simply replacing the value of the target options will not work. There is another core plugin which is reading project.json in the end and replacing plugin generated values with values from the file.  So I'm generating two options, one array, another string. Array is ignored, string not.

With the current PR this plugin will not be needed because array will be directly parsed to comma-separated string.

AgentEnder:
Hey @ThePlenkov - I'm going to close this one out. In the process of trying to get it to pass CI, I've come to think about it more and I don't think its really a feasible change for a few reasons.

Passing an array to a CLI is handled differently by different tools and I'm not confident this would handle them properly. For args not forwarded, using `args: [....]` should be preferred

PR #27778, opened 2024-09-04
fix(core): collect all logs from forked processes

dgreif:
## Current Behavior

We have a monorepo with ~450 packages and run tasks with 16x parallelism in CI. We've recently noticed that running tasks with `static` logging can lead to random tasks losing logs. Even if the task is a simple `echo SOMETHING`, logs get lost sporadically. 

After digging into the task runner, I found that the forked task runner currently collects logs from `stdout` and `stderr`, then joins those when the process `exit` event is emitted. The problem appears to stem from the fact that the `exit` event is _occasionally_ emitted _before_ the `stdout` and `stderr` streams are closed, leading to lost logs.

## Expected Behavior

_All_ logs should be collected, even if they are emitted after the process `exit` event.

I've updated the task runner to wait for the `stdout` and `stderr` streams to end before collecting logs.

Note: I didn't see any tests specifically for this file, but it should still run all the new code if theres is a happy path case of forked tasks with static logging.

### DIFF
--- a/packages/nx/src/tasks-runner/forked-process-task-runner.ts
+++ b/packages/nx/src/tasks-runner/forked-process-task-runner.ts
@@ -339,29 +339,53 @@ export class ForkedProcessTaskRunner {
         }
 
         let outWithErr = [];
+        let exitCode;
+        let stdoutHasEnded = false;
+        let stderrHasEnded = false;
+        let processHasExited = false;
+
+        const handleProcessEnd = () => {
+          // ensure process has exited and both stdout and stderr have ended before we pass along the logs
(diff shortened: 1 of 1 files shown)


rluvaton:
This seems like fixing the same bug as my pr from 7 months ago
- https://github.com/nrwl/nx/pull/21498

dgreif:
@rluvaton great catch! I'm surprised your PR hasn't gotten any traction yet 😞. Which approach would you lean toward for fixing the issue? I see the merit in your approach, but I'd be worried that some new issue could easily pop up which starts keeping the child process alive again. With the fix I have here, we are less dependent on the child process exiting with specific timing. WDYT?

AgentEnder (approved):

PR #27783, opened 2024-09-05
feat(core): use durations from task history to schedule tasks

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/daemon/server/handle-task-history.ts
+++ b/packages/nx/src/daemon/server/handle-task-history.ts
@@ -1,14 +1,5 @@
-import { TaskRun } from '../../native';
-import { TaskHistory } from '../../utils/task-history';
-
-let taskHistory: TaskHistory;
-
-function getTaskHistory() {
-  if (!taskHistory) {
-    taskHistory = new TaskHistory();
-  }
-  return taskHistory;
-}
+import { getTaskHistory } from '../../utils/task-history';
+import type { TaskRun, TaskTarget } from '../../native';
 
 export async function handleRecordTaskRuns(taskRuns: TaskRun[]) {
   const taskHistory = getTaskHistory();
@@ -25,5 +16,14 @@ export async function handleGetFlakyTasks(hashes: string[]) {
   return {
     response: JSON.stringify(history),
     description: 'handleGetFlakyTasks',
+  };
+}
+
+export async function handleGetEstimatedTaskTimings(targets: TaskTarget[]) {
+  const taskHistory = getTaskHistory();
+  const history = await taskHistory.getEstimatedTaskTimings(targets);
+  return {
+    response: JSON.stringify(history),
+    description: 'handleGetEstimatedTaskTimings',
   };
 }
--- a/packages/nx/src/tasks-runner/task-orchestrator.ts
+++ b/packages/nx/src/tasks-runner/task-orchestrator.ts
@@ -76,7 +76,10 @@ export class TaskOrchestrator {
 
   async run() {
     // Init the ForkedProcessTaskRunner
-    await this.forkedProcessTaskRunner.init();
+    await Promise.all([
+      this.forkedProcessTaskRunner.init(),
+      this.tasksSchedule.init(),
+    ]);
 
     // initial scheduling
     await this.scheduleNextTasks();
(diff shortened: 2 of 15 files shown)


packages/nx/src/tasks-runner/default-tasks-runner.ts
```diff
@@ -137,7 +138,8 @@ async function runAllTasks(
     options,
     context.nxArgs?.nxBail,
     context.daemon,
-    context.nxArgs?.outputStyle
+    context.nxArgs?.outputStyle,
+    getTaskHistory()
```
FrozenPandaz, line 142: We don't have to pass it into the task orchestrator

packages/nx/src/tasks-runner/default-tasks-runner.ts
```diff
@@ -11,6 +11,7 @@ import { cacheDir } from '../utils/cache-directory';
 import { readFile, writeFile, mkdir } from 'fs/promises';
 import { join } from 'path';
 import { CachedResult } from '../native';
+import { getTaskHistory } from '../daemon/server/handle-task-history';
```
FrozenPandaz, line 14: This should come from somewhere that is not the daemon server

packages/nx/src/tasks-runner/tasks-schedule.ts
```diff
@@ -24,13 +25,21 @@ export class TasksSchedule {
   private runningTasks = new Set<string>();
   private completedTasks = new Set<string>();
   private scheduleRequestsExecutionChain = Promise.resolve();
+  private estimatedTaskTimings: Record<string, number>;
 
   constructor(
     private readonly projectGraph: ProjectGraph,
     private readonly taskGraph: TaskGraph,
+    private readonly taskHistory: TaskHistory,
```
FrozenPandaz, line 33: We can get this taskHistory from within here.

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -71,12 +73,16 @@ export class TaskOrchestrator {
     private readonly options: DefaultTasksRunnerOptions,
     private readonly bail: boolean,
     private readonly daemon: DaemonClient,
-    private readonly outputStyle: string
+    private readonly outputStyle: string,
+    private readonly taskHistory: TaskHistory
```
FrozenPandaz, line 77: We don't need to pass this in or create it here in the orchestrator.

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -39,6 +40,7 @@ export class TaskOrchestrator {
   private tasksSchedule = new TasksSchedule(
     this.projectGraph,
     this.taskGraph,
+    this.taskHistory,
```
FrozenPandaz, line 43: We don't need to pass this in or create it here in the orchestrator.

packages/nx/src/tasks-runner/tasks-schedule.ts
```diff
@@ -112,12 +121,26 @@ export class TasksSchedule {
...
+          project2NodeDependencies - project1NodeDependencies;
+
+        if (dependenciesDiff !== 0) {
+          return dependenciesDiff;
+        }
+
+        const task1Timing = this.estimatedTaskTimings[taskId1] ?? 0;
+        const task2Timing = this.estimatedTaskTimings[taskId2] ?? 0;
+
+        return task2Timing - task1Timing;
```
FrozenPandaz, line 143: Change this so that if a task does not have an estimated timing, we assume it is a new task, which should yield more interesting results... thus, we should run those first.

packages/nx/src/daemon/server/handle-task-history.ts
```diff
@@ -1,9 +1,9 @@
-import { TaskRun } from '../../native';
+import { TaskRun, TaskTarget } from '../../native';
 import { TaskHistory } from '../../utils/task-history';
 
 let taskHistory: TaskHistory;
 
-function getTaskHistory() {
+export function getTaskHistory() {
```
FrozenPandaz, line 6: This should move to the utils/task-history

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #27787, opened 2024-09-05
feat(release): allow local dependency version protocols to be preserved, pnpm publish support

JamesHenry:
## Current Behavior

We always update relevant dependency versions to their new semantic version as part of versioning and we only ever publish using `npm publish`.

## Expected Behavior

We allow for the user to configure a preference for preserving local dependency protocols (such as `file:` and `workspace:`) when versioning, whilst maintaining all other aspects of the versioning logic.

This would allow them to combine it with `pnpm publish`, which would handle the dynamic replacement of those values during the packing step. Therefore, when the user is using `pnpm`, we now preferentially publish using `pnpm publish` instead of `npm publish`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/generators/release-version/release-version.ts
+++ b/packages/js/src/generators/release-version/release-version.ts
@@ -1,10 +1,7 @@
 import {
-  PackageManager,
   ProjectGraphProjectNode,
   Tree,
-  detectPackageManager,
   formatFiles,
-  getPackageManagerVersion,
   joinPathFragments,
   output,
   readJson,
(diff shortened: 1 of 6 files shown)


packages/js/src/generators/release-version/release-version.ts
```diff
@@ -993,7 +999,8 @@ let pmVersion: string | undefined;
 const localPackageProtocols = [
   'file:', // all package managers
   'workspace:', // not npm
-  'portal:', // modern yarn only
```
JamesHenry, line 996: I had jumped the gun on this being there in a previous PR, because we don't support it at the project graph level right now, so it's already not an applicable setup

fahslaj (approved):

PR #27797, opened 2024-09-05
fix(release): allow string array for commitArgs and tagArgs

fahslaj:
Allow an array of strings for the "release.git.gitCommitArgs" and "release.git.gitTagArgs" properties, which allow for arguments with spaces to be passed. 

Previously, they each needed to be a string and disallowed spaces.


JamesHenry (changes requested):
Please can we update tagArgs for consistency?

JamesHenry (approved):

PR #27804, opened 2024-09-06
feat(angular): add plugin for inferring nodes from angular.json files

leosvelperez:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/plugins/plugin.ts
+++ b/packages/angular/src/plugins/plugin.ts
@@ -1,0 +1,733 @@
+import {
+  createNodesFromFiles,
+  detectPackageManager,
+  getPackageManagerCommand,
+  readJsonFile,
+  writeJsonFile,
+  type CreateNodesContextV2,
+  type CreateNodesResult,
+  type CreateNodesV2,
+  type ProjectConfiguration,
+  type Target,
+  type TargetConfiguration,
+} from '@nx/devkit';
+import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
+import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
+import { getLockFileName } from '@nx/js';
+import { existsSync, readdirSync, statSync } from 'node:fs';
+import { dirname, join, relative } from 'node:path';
+import * as posix from 'node:path/posix';
+import { hashObject } from 'nx/src/devkit-internals';
+import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
+
+export interface AngularPluginOptions {}
+
+type AngularProjects = Record<
+  string,
+  Pick<ProjectConfiguration, 'projectType' | 'sourceRoot' | 'targets'>
+>;
+
+type AngularTargetConfiguration = {
+  builder: string;
+  options?: Record<string, any>;
+  configurations?: Record<string, any>;
+  defaultConfiguration?: string;
+};
+export type AngularProjectConfiguration = {
+  projectType: 'application' | 'library';
+  root: string;
+  sourceRoot?: string;
+  architect?: Record<string, AngularTargetConfiguration>;
+  targets?: Record<string, AngularTargetConfiguration>;
+};
+type AngularJson = { projects?: Record<string, AngularProjectConfiguration> };
+
+const knownExecutors = {
+  appShell: new Set(['@angular-devkit/build-angular:app-shell']),
+  build: new Set([
+    '@angular-devkit/build-angular:application',
+    '@angular/build:application',
+    '@angular-devkit/build-angular:browser-esbuild',
+    '@angular-devkit/build-angular:browser',
+    '@angular-devkit/build-angular:ng-packagr',
+  ]),
+  devServer: new Set(['@angular-devkit/build-angular:dev-server']),
+  extractI18n: new Set(['@angular-devkit/build-angular:extract-i18n']),
+  prerender: new Set(['@angular-devkit/build-angular:prerender']),
+  server: new Set(['@angular-devkit/build-angular:server']),
+  serveSsr: new Set(['@angular-devkit/build-angular:ssr-dev-server']),
+  test: new Set(['@angular-devkit/build-angular:karma']),
+};
+
+const pmc = getPackageManagerCommand();
+
+function readProjectsCache(cachePath: string): Record<string, AngularProjects> {
+  return existsSync(cachePath) ? readJsonFile(cachePath) : {};
+}
+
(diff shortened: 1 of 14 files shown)


packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+    const namedInputs = getNamedInputs(project.root, context);
+
+    for (const [targetName, angularTarget] of Object.entries(projectTargets)) {
+      const externalDependencies = ['@angular/cli'];
+
+      targets[targetName] = {
+        command: `ng run ${projectName}:${targetName}`,
+        options: { cwd: angularWorkspaceRoot },
+        metadata: {
+          technologies: ['angular'],
```
leosvelperez, line 164: @FrozenPandaz should I also add `webpack` and `esbuild` to the build targets depending on the executor used?

packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
+import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
+import { getLockFileName } from '@nx/js';
+import { existsSync, readdirSync, statSync } from 'node:fs';
+import { dirname, join, relative } from 'node:path';
+import * as posix from 'node:path/posix';
+import { hashObject } from 'nx/src/devkit-internals';
+import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
+
+export interface AngularPluginOptions {}
```
Coly010, line 23: Is this intentionally empty? Is it still needed if so?

packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+type AngularJson = { projects?: Record<string, AngularProjectConfiguration> };
+
+const knownExecutors = {
+  appShell: new Set(['@angular-devkit/build-angular:app-shell']),
+  build: new Set([
+    '@angular-devkit/build-angular:application',
+    '@angular/build:application',
+    '@angular-devkit/build-angular:browser-esbuild',
+    '@angular-devkit/build-angular:browser',
+    '@angular-devkit/build-angular:ng-packagr',
```
Coly010, line 52: Do we need to include third parties such as `ngx-build-plus` and `angular-builders:custom-webpack` etc?

packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+  ]),
+  devServer: new Set(['@angular-devkit/build-angular:dev-server']),
+  extractI18n: new Set(['@angular-devkit/build-angular:extract-i18n']),
+  prerender: new Set(['@angular-devkit/build-angular:prerender']),
+  server: new Set(['@angular-devkit/build-angular:server']),
+  serveSsr: new Set(['@angular-devkit/build-angular:ssr-dev-server']),
```
Coly010, line 58: Do we need to include the `ng-univsersal` builders?

packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+  test: new Set(['@angular-devkit/build-angular:karma']),
```
Coly010, line 59: Do we need to include the experimental jest test runner and the web test runner?

packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+  cachePath: string,
+  results: Record<string, AngularProjects>
+) {
+  writeJsonFile(cachePath, results);
+}
+
+export const createNodesV2: CreateNodesV2<AngularPluginOptions> = [
+  '**/angular.json',
+  async (configFiles, options, context) => {
+    const optionsHash = hashObject(options);
```
Coly010, line 78: How safe is this hash if the options is empty/undefined? (because of empty interface)

packages/angular/src/plugins/plugin.ts
```diff
@@ -0,0 +1,733 @@
...
+    context
+  );
+
+  return { projects: projectsCache[hash] };
+}
+
+async function buildAngularProjects(
+  configFilePath: string,
+  angularWorkspaceRoot: string,
+  options: AngularPluginOptions,
```
Coly010, line 134: I don't see this being used

Coly010 (approved):
some comments, but overall looks good

packages/angular/src/plugins/plugin.ts
leosvelperez: The `CreateNodesFunctionV2` expects a type for the options.  We don't have any options right now, but I added the empty interface as a placeholder in case it's needed later. I can remove it.

packages/angular/src/plugins/plugin.ts
leosvelperez: We could. I guess we could include any widely used executor for which we know how to set up the relevant props (e.g. `cache`, `inputs`, `outputs`, etc.).
 
To be clear, not having a builder here doesn't mean the target won't be inferred. It just means we won't have any special handling for it (like setting `cache`, `inputs`, `outputs`, etc.).

packages/angular/src/plugins/plugin.ts
leosvelperez: Yep, good catch.

packages/angular/src/plugins/plugin.ts
leosvelperez: I decided not to. They are experimental and I doubt they are widely used in any meaningful project. We can add them later.

packages/angular/src/plugins/plugin.ts
leosvelperez: That's handled by the function. Regardless of an empty interface the options can always be undefined since plugins are designed to not require them.

packages/angular/src/plugins/plugin.ts
leosvelperez: Yeah, it's related to having everything wired up for/if options are needed at some point, but we could remove it for now.

packages/angular/src/plugins/plugin.ts
Coly010: Perhaps stick to defaults from Angular and can expand later

packages/angular/src/plugins/plugin.ts
leosvelperez: We discussed this offline and decided not to.

jaysoo (approved):

PR #27820, opened 2024-09-08
feat(core): add support for bun pack

Jordan-Hall:
## Current Behavior
Before release last night/today of bun 1.1.27 bun pm pack didn't exists so we fell upon npm pack.

## Expected Behavior
If bun 1.1.27 or greater we use bun pm pack otherwise we revert to npm pack

https://x.com/bunjavascript/status/1832386689182122092

cc: @jaysoo

### DIFF
--- a/packages/nx/src/utils/package-manager.ts
+++ b/packages/nx/src/utils/package-manager.ts
@@ -26,6 +26,7 @@ export interface PackageManagerCommands {
   exec: string;
   dlx: string;
   list: string;
+  pack: string;
   run: (script: string, args?: string) => string;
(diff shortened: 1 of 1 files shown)


packages/nx/src/utils/package-manager.ts
```diff
@@ -26,6 +26,7 @@ export interface PackageManagerCommands {
   exec: string;
   dlx: string;
   list: string;
+  pack: string;
```
Jordan-Hall, line 29: This has been added but i'm not sure if a better name can be used. bun uses bun install at top level and bunx but rest of its package manager commands is at bun pm level

PR #27825, opened 2024-09-09
docs(core): nx sync

isaacplmann:
- Adds nx sync to command reference
- Updates nx.json reference
- Update project configuration reference
- Adds a sync generators concept page
- Adds a Register a Sync Generator recipe under Extending Nx Recipes


packages/nx/src/command-line/sync/command-object.ts
```diff
@@ -22,7 +22,7 @@ export const yargsSyncCheckCommand: CommandModule<
   SyncArgs
 > = {
   command: 'sync:check',
-  describe: false,
+  describe: 'Check that all sync generators have been run',
```
JamesHenry, line 25:
```suggestion
  describe: 'Check that no changes are required after running all sync generators',
```

docs/shared/reference/project-configuration.md
```diff
@@ -616,6 +616,20 @@ Additionally, when using the expanded object syntax, you can specify individual
 This configuration is usually not needed. Nx comes with reasonable defaults (imported in `nx.json`) which implement the
 configuration above.
 
+### Sync Generators
+
+In the same way that `dependsOn` tells Nx to run a task before this task, the `syncGenerator` property tells Nx to run a generator the task is run. [Sync generators](/concepts/sync-generators) are available in Nx 19.7+.
```
JamesHenry, line 621:
```suggestion
In the same way that `dependsOn` tells Nx to run a task before this task, the `syncGenerator` property tells Nx to run a specific type of generator (aimed at keeping applicable configuration files in sync with your project graph) before the task is run. [Sync generators](/concepts/sync-generators) are available in Nx 19.7+.
```

docs/shared/reference/nx-json.md
```diff
@@ -507,6 +510,32 @@ The `git` property configures the automated git operations that take place as pa
 }
 ```
 
+## Sync
+
+These are global configuration options for the [`nx sync`](/reference/nx-commands#sync) command. The `nx sync` command runs all global and task-specific [sync generators](/concepts/sync-generators) to keep your codebase consistent with itself.
```
JamesHenry, line 515:
```suggestion
These are global configuration options for the [`nx sync`](/reference/nx-commands#sync) command. The `nx sync` command runs all global and task-specific [sync generators](/concepts/sync-generators) to keep applicable files in sync with your project graph.
```

docs/shared/recipes/generators/register-sync-generator.md
```diff
@@ -0,0 +1,110 @@
+# Register a Sync Generator
+
+Sync generators allow you to update code before a task is run or before code is sent to CI. From a technical perspective, a sync generator is no different from any other generator, but it has some additional performance considerations and needs to be registered in a particular way.
```
JamesHenry, line 3:
```suggestion
Sync generators are intended to keep applicable files in sync with your project graph before a task is run. From a technical perspective, a sync generator is no different from any other generator, but it has some additional performance considerations and needs to be registered in a particular way.
```

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
+# Sync Generators
+
+In Nx 19.7, you can use sync generators to update code before a task is run or before code is sent to CI. Sync generators are designed to perform maintenance work that is required for a particular task or for CI.
```
JamesHenry, line 3:
```suggestion
In Nx 19.7, you can use sync generators to keep applicable files in sync with your project graph. These can be for global configuration files or scripts, or at the task level to ensure to that project level configuration files are in sync before a task is run.
```

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
...
+
+Sync Generator Examples:
+
+- Ensure code is formatted in a specific way before CI is run
```
JamesHenry, line 7: I don't think this is an applicable example, we have `nx format` and making that pluggable would be the preferred way to achieve this IMO (because we have internal utils like `formatFiles` which are coupled to format and won't take sync generators into account in any way)

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
...
+- Update a custom CI script with binning strategies based on the current project graph
+- Update TypeScript config files with project references based on the current project graph
+
+## Task Sync Generators
+
+Sync generators can be associated with a particular task. Nx will ensure that the sync generator is run before the task is executed. This is similar to the `dependsOn` property, but for generators instead of task dependencies.
```
JamesHenry, line 13:
```suggestion
Sync generators can be associated with a particular task. Nx will ensure that the sync generator is run before the task is executed (unless `sync.applyChanges` is set to `false` in `nx.json`). This is similar to the `dependsOn` property, but for generators instead of task dependencies.
```

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
...
+
+{% graph title="Project Graph" height="200px" type="project" jsonFile="shared/mental-model/three-projects.json" %}
+{% /graph %}
+{% /side-by-side %}
+
+## Example Usage
+
+When Nx detects that the repository is out of sync, the user will be asked if sync generators should be run. To skip this prompt, set the `sync.applyChanges` property in `nx.json` to either `true` or `false`.
+
+Task sync generators are executed whenever their task is run, but global sync generators need to be triggered manually with `nx sync`. We recommend running `nx sync` in a pre-commit or pre-push Git hook to ensure that all committed code is in sync. Also, add `nx sync:check` to the beginning of your CI scripts so that CI can fail quickly if the code is out of sync.
```
JamesHenry, line 46: I would switch up the emphasis here and say we _strongly_ recommend `nx sync:check` in CI. The git hooks are optional and not as important because folks can easily skip them

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
...
+
+Sync generators can be associated with a particular task. Nx will ensure that the sync generator is run before the task is executed. This is similar to the `dependsOn` property, but for generators instead of task dependencies.
+
+To [register a generator](/extending-nx/recipes/register-sync-generator) as a sync generator for a particular task, add the generator to the `syncGenerators` property of the task configuration.
+
+## Global Sync Generators
+
+Global sync generators are not associated with a particular task and are executed only when the `nx sync` or `nx sync:check` command is explicitly run. They are [registered](/extending-nx/recipes/register-sync-generator) in the `nx.json` file with the `sync.globalGenerators` property.
+
+## Use the Project Graph in a Sync Generator
```
JamesHenry, line 21: This section doesn't show you how to use the project graph in a sync generator?

JamesHenry (changes requested):

packages/nx/src/command-line/sync/command-object.ts
```diff
@@ -10,7 +10,7 @@ export const yargsSyncCommand: CommandModule<
   SyncArgs
 > = {
   command: 'sync',
-  describe: false,
+  describe: 'Run all sync generators',
```
leosvelperez, line 13:
```suggestion
  describe: 'Sync the workspace files by running all the sync generators.',
```

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
+# Sync Generators
+
+In Nx 19.7, you can use sync generators to keep applicable files in sync with your project graph. These can be for global configuration files or scripts, or at the task level to ensure to that project level configuration files are in sync before a task is run.
```
leosvelperez, line 3: I'm unsure if we're targeting 19.7 as the version for the stable release. There are still some PRs that need to land before that. @FrozenPandaz can confirm here.

leosvelperez, line 3: A couple of clarifications here:

- Sync generators "can" keep files in sync with the project graph, but they are not limited to that. They can do anything and are not restricted to using the project graph as the source of truth.
- Similarly, sync generators "can" update configuration files but are not restricted to that type of files. They could also add/update/delete any file.

The example below about formatting files is a clear example of the above, where the project graph is not the source of truth and any file could be formatted.

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
...
+
+Sync Generator Examples:
+
+- Ensure code is formatted in a specific way before CI is run
+- Update a custom CI script with binning strategies based on the current project graph
+- Update TypeScript config files with project references based on the current project graph
+
+## Task Sync Generators
+
+Sync generators can be associated with a particular task. Nx will ensure that the sync generator is run before the task is executed (unless `sync.applyChanges` is set to `false` in `nx.json`). This is similar to the `dependsOn` property, but for generators instead of task dependencies.
```
leosvelperez, line 13: When `sync.applyChanges` is set to `false`, sync generators are still run before running tasks, but the user isn't prompted, and the changes aren't applied.

Note that this is only true when running in interactive environments. When running in CI or in a non-interactive environment, if the workspace is out of sync, an error will be logged, and the process will exit with code 1.

The `sync.applyChanges` allows you to set the default answer to the prompt that asks you whether you want to apply the changes. It doesn't allow you to opt-out of running the sync generators. It only allows you to opt-out of applying the identified changes needed to be made to sync the workspace.

To skip running sync generators when running tasks, you can run tasks with the `--skip-sync` flag or add specific sync generators to the `sync.disabledTaskSyncGenerators` array in `nx.json`.

docs/shared/concepts/sync-generators.md
```diff
@@ -0,0 +1,46 @@
...
+   └─ package.json
+```
+
+{% graph title="Project Graph" height="200px" type="project" jsonFile="shared/mental-model/three-projects.json" %}
+{% /graph %}
+{% /side-by-side %}
+
+## Example Usage
+
+When Nx detects that the repository is out of sync, the user will be asked if sync generators should be run. To skip this prompt, set the `sync.applyChanges` property in `nx.json` to either `true` or `false`.
```
leosvelperez, line 44:
```suggestion
When Nx detects that the repository is out of sync, the user will be asked if the identified changes by the sync generators registered in the tasks should be applied. To skip this prompt, set the `sync.applyChanges` property in `nx.json` to either `true` or `false`.
```

docs/shared/recipes/generators/register-sync-generator.md
```diff
@@ -0,0 +1,110 @@
+# Register a Sync Generator
+
+Sync generators are intended to keep applicable files in sync with your project graph before a task is run. From a technical perspective, a sync generator is no different from any other generator, but it has some additional performance considerations and needs to be registered in a particular way.
```
leosvelperez, line 3: I'm unsure what would be the best way to phrase it, but sync generators are not limited to keeping files in sync with the project graph.

docs/shared/reference/project-configuration.md
```diff
@@ -616,6 +616,20 @@ Additionally, when using the expanded object syntax, you can specify individual
 This configuration is usually not needed. Nx comes with reasonable defaults (imported in `nx.json`) which implement the
 configuration above.
 
+### Sync Generators
+
+In the same way that `dependsOn` tells Nx to run a task before this task, the `syncGenerator` property tells Nx to run a specific type of generator (aimed at keeping applicable configuration files in sync with your project graph) before the task is run. [Sync generators](/concepts/sync-generators) are available in Nx 19.7+.
```
leosvelperez, line 621: Same here about the sync generators not being limited to:

- update configuration files
- use the project graph as the source of truth

docs/shared/reference/nx-json.md
```diff
@@ -507,6 +510,32 @@ The `git` property configures the automated git operations that take place as pa
 }
 ```
 
+## Sync
+
+These are global configuration options for the [`nx sync`](/reference/nx-commands#sync) command. The `nx sync` command runs all global and task-specific [sync generators](/concepts/sync-generators) to keep applicable files in sync with your project graph.
```
leosvelperez, line 515: Same here about the sync generators not being limited to using the project graph as the source of truth.

docs/shared/recipes/generators/register-sync-generator.md
leosvelperez, line 1: I feel this recipe is either titled incorrectly or is for more things than it should be. The title is `Register a Sync Generator`, but we have two sections referring to how to create a sync generator.

docs/shared/recipes/generators/register-sync-generator.md
```diff
@@ -0,0 +1,110 @@
...
+  }
+  return {
+    outOfSyncMessage: 'The legal-message.txt file needs to be created',
+  };
+}
+
+export default mySyncGenerator;
+```
+
+## Performance Considerations
```
leosvelperez, line 50: I think that a recipe about creating sync generators should also suggest that sync generator authors be informative when throwing an error or returning an `outOfSyncMessage.` They should provide guidance to the user on how to address any error.

This is not a perf-related suggestion but an important DX suggestion. Sync generators can block tasks, so providing good feedback is important.

docs/shared/recipes/generators/register-sync-generator.md
```diff
@@ -0,0 +1,110 @@
...
+
+When writing a sync generator, it is important to remember that this generator is going to block the execution of its associated task and any tasks that depend on that task. Because of this fact, make sure to follow the following performance tips:
+
+- Make the generator idempotent. Running the generator multiple times in a row should have the same impact as running the generator a single time.
+- Only write to the file system when a file is actually changed. Avoid reformatting files that have not been actually modified.
```
leosvelperez, line 55:
```suggestion
- Only write to the file system when a file is actually changed. Avoid reformatting files that have not been actually modified. Nx will identify the workspace as out of sync if there's any file change after the sync generator is run.
```

leosvelperez (changes requested):

docs/shared/concepts/sync-generators.md
leosvelperez, line 1: I think we should also mention somewhere that when running tasks in CI (or non-interactive environments) with the workspace being out of sync, the tasks won't run. In that case, the user won't be prompted, an error will be thrown, and the process will exit with code 1.

That's the default behavior, but users can opt out by running tasks with the `--skip-sync` flag.

docs/shared/concepts/sync-generators.md
isaacplmann: Correct.  I updated the title to better describe what this section is about.  How to use the project graph should go in the Create a Sync Generator recipe, not in the concept page.

docs/shared/recipes/generators/create-sync-generator.md
```diff
@@ -0,0 +1,136 @@
...
+  tree: Tree
+): Promise<SyncGeneratorResult> {
+  const projectGraph = await createProjectGraphAsync();
+  const numberOfProjects = Object.values(projectGraph.nodes).length;
+  tree.write(
+    '/project-count.txt',
+    `There are ${numberOfProjects} projects in the repository.`
+  );
+  return {
+    outOfSyncMessage: 'The project-count.txt file needs to be created',
```
JamesHenry, line 68: Nit: I know it's just an example but this would be a more appropriate message given it will also be out of sync when the project count changes.

```suggestion
    outOfSyncMessage: 'The project-count.txt file needs to be update',
```

docs/shared/recipes/generators/create-sync-generator.md
```diff
@@ -0,0 +1,136 @@
...
+  };
+}
+
+export default mySyncGenerator;
+```
+
+## Performance and DX Considerations
+
+When writing a sync generator, it is important to remember that this generator is going to block the execution of its associated task and any tasks that depend on that task. Because of this fact, make sure to follow the following performance tips:
```
JamesHenry, line 77: This is specific to task generators, maybe make that clear?

docs/shared/recipes/generators/create-sync-generator.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+Sync generators can optionally return an `outOfSyncMessage` to display to users when the sync generator needs to be run.
+
+```ts {% fileName="tools/my-plugin/src/generators/my-sync-generator/my-sync-generator.ts" %}
+import { Tree } from '@nx/devkit';
+import { SyncGeneratorResult } from 'nx/src/utils/sync-generators';
+
+export async function mySyncGenerator(
```
JamesHenry, line 32: This one and the one below seem like examples of global sync generators. Maybe worth calling them `myGlobalSyncGenerator` and adding a more relevant example of a task one?

docs/shared/recipes/generators/create-sync-generator.md
JamesHenry, line 1: Now that I've been through the whole thing:

Maybe split the file more 50:50 into task vs global, giving one example of each

JamesHenry (approved):

leosvelperez (approved):

PR #27826, opened 2024-09-09
docs(core): remove William Ghelfi from champions

isaacplmann:
Removes William Ghelfi from the champions list


leosvelperez (approved):

juristr (commented):
lgtm, thx

PR #27831, opened 2024-09-09
fix(vite/test): partial support for vitest@2

rluvaton:
When having no tests the process will hang, this is because `onFinished` event is never emitted when no test exists and before version 2, the process was terminated, but not it was not

this is the vitest PR that changed it:
https://github.com/vitest-dev/vitest/pull/5926


Coly010:
Hi @rluvaton Thanks for creating this PR. 

We will be investigating Vitest 3 support very shortly and will encompass this change in that. I'm going to close this PR in favour of a more complete solution that we will be implementating

PR #27833, opened 2024-09-09
feat(graph): expose functions to render pdv & error page

MaxKless:
we have fully isolated graph & error components with a good api but no way to access them directly from the outside (in console). 

This PR adds two functions to the window object so that we can render the PDV directly instead of needing the entire app with routing and everything.

### DIFF
--- a/graph/client/src/main.tsx
+++ b/graph/client/src/main.tsx
@@ -9,30 +9,58 @@ import { inspect } from '@xstate/inspect';
 import { App } from './app/app';
 import { ExternalApiImpl } from './app/external-api-impl';
 import { render } from 'preact';
+import { ErrorToastUI, ExpandedTargetsProvider } from '@nx/graph/shared';
+import { ProjectDetails } from '@nx/graph-internal/ui-project-details';
+import { ErrorPage } from './app/ui-components/error-page';
 
-if (window.useXstateInspect === true) {
-  inspect({
-    url: 'https://stately.ai/viz?inspect',
-    iframe: false, // open in new window
-  });
-}
+if (window.__NX_RENDER_GRAPH__ === false) {
+  window.renderPDV = (data: any) => {
+    const container = document.getElementById('app');
+    render(
+      <StrictMode>
+        <ExpandedTargetsProvider>
+          <ProjectDetails {...data} />
+        </ExpandedTargetsProvider>
+        <ErrorToastUI errors={data.errors} />
+      </StrictMode>,
+      container
+    );
+  };
 
-window.externalApi = new ExternalApiImpl();
-const container = document.getElementById('app');
-
-if (!window.appConfig) {
-  render(
-    <p>
-      No environment could be found. Please run{' '}
-      <pre>npx nx run graph-client:generate-dev-environment-js</pre>.
-    </p>,
-    container
-  );
+  window.renderError = (data: any) => {
+    const container = document.getElementById('app');
+    render(
+      <StrictMode>
+        <ErrorPage {...data} />
+      </StrictMode>,
+      container
+    );
+  };
 } else {
-  render(
-    <StrictMode>
-      <App />
-    </StrictMode>,
-    container
-  );
+  if (window.useXstateInspect === true) {
+    inspect({
+      url: 'https://stately.ai/viz?inspect',
+      iframe: false, // open in new window
+    });
+  }
+
+  window.externalApi = new ExternalApiImpl();
+  const container = document.getElementById('app');
+
+  if (!window.appConfig) {
+    render(
+      <p>
+        No environment could be found. Please run{' '}
+        <pre>npx nx run graph-client:generate-dev-environment-js</pre>.
+      </p>,
+      container
+    );
+  } else {
+    render(
+      <StrictMode>
+        <App />
+      </StrictMode>,
+      container
+    );
+  }
 }
--- a/graph/client/src/main.tsx
+++ b/graph/client/src/main.tsx
@@ -4,36 +4,55 @@ if (process.env.NODE_ENV === 'development') {
   require('preact/debug');
 }
 
+import { projectDetailsMachine } from './app/console-project-details/project-details.machine';
+/* eslint-disable @nx/enforce-module-boundaries */
+// nx-ignore-next-line
+import type { ProjectGraphProjectNode } from '@nx/devkit';
+// nx-ignore-next-line
+import type { GraphError } from 'nx/src/command-line/graph/graph';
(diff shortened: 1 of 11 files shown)


graph/client/src/main.tsx
```diff
@@ -9,30 +9,58 @@ import { inspect } from '@xstate/inspect';
...
-  inspect({
-    url: 'https://stately.ai/viz?inspect',
-    iframe: false, // open in new window
-  });
-}
-
-window.externalApi = new ExternalApiImpl();
-const container = document.getElementById('app');
+if (window.__NX_RENDER_GRAPH__ === false) {
+  window.renderPDV = (data: any) => {
```
jaysoo, line 17: Can we get more specific on the type for `data`? It wouldn't affect type-safety from Console side, but at least this file would be correct.

graph/client/src/main.tsx
```diff
@@ -9,30 +9,58 @@ import { inspect } from '@xstate/inspect';
...
-if (window.useXstateInspect === true) {
-  inspect({
-    url: 'https://stately.ai/viz?inspect',
-    iframe: false, // open in new window
-  });
-}
-
-window.externalApi = new ExternalApiImpl();
-const container = document.getElementById('app');
+if (window.__NX_RENDER_GRAPH__ === false) {
```
jaysoo, line 16: Ideally, we'd have `renderPDV`, `renderError` and `renderGraph` functions exposed. But for older Nx versions, obviously we need to support the old way.

Can we check on the Console side, whether we're on 19.8, and then call `renderGraph` rather than assuming the main file will call `render` automatically?

graph/ui-project-details/src/lib/project-details/project-details.tsx
```diff
@@ -160,16 +186,69 @@ export const ProjectDetails = ({
   );
 };
 
-export default ProjectDetails;
+// const ProjectsDropdown = ({
```
jaysoo, line 189: Remove? This is in a follow-up?

graph/ui-project-details/src/lib/project-details/project-details.tsx
```diff
@@ -4,17 +4,20 @@ import type { ProjectGraphProjectNode } from '@nx/devkit';
...
-import { TagList } from '../tag-list/tag-list';
+import {
+  ProjectInfoHeader,
+  ViewInProjectGraphButton,
+} from '../project-info-header/project-info-header';
 import { TargetConfigurationGroupList } from '../target-configuration-details-group-list/target-configuration-details-group-list';
 import { TooltipTriggerText } from '../target-configuration-details/tooltip-trigger-text';
 import { TargetTechnologies } from '../target-technologies/target-technologies';
+import { ReactNode, useEffect, useState } from 'react';
+import { Dropdown } from '@nx/graph/ui-components';
```
jaysoo, line 17: Icons are not used?

graph/ui-project-details/src/lib/project-details/project-details.tsx
```diff
@@ -28,25 +31,44 @@ export interface ProjectDetailsProps {
   viewInProjectGraphPosition?: 'top' | 'bottom';
 }
 
-const typeToProjectType = {
-  app: 'Application',
-  lib: 'Library',
-  e2e: 'E2E',
-};
+export interface SingleProjectDetailsProps extends ProjectDetailsPropsBase {
```
jaysoo, line 34: I think a discriminated union type would be cleaner?

```ts
export type SingleProjectDetailsProps = ProjectDetailsPropsBase & {
  type: 'single';
  project: ProjectGraphProjectNode;
  sourceMap: Record<string, string[]>;
};

export type MultiProjectDetailsProps = ProjectDetailsPropsBase & {
  type: 'multi';
  projects: ProjectGraphProjectNode[];
  sourceMaps: Record<string, Record<string, string[]>>;
};

function getSingleProjectProps(
  props: SingleProjectDetailsProps | MultiProjectDetailsProps,
  project: ProjectGraphProjectNode
): SingleProjectDetailsProps {
  return props.type === 'single'
    ? props
    : {
      ...props,
      type: 'single',
      project,
      sourceMap: props.sourceMaps[project.data.root],
    };
}

```

graph/ui-project-details/src/lib/project-info-header/project-info-header.tsx
```diff
@@ -0,0 +1,89 @@
...
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export function ViewInProjectGraphButton({
+  callback,
+}: {
+  callback: () => void;
```
jaysoo, line 78: This should be `onClick` as it is a click event handler. Also, I don't  think it's large/complex enough to warrant its own component. Just inline in the main component function.

jaysoo (changes requested):

graph/client/src/main.tsx
jaysoo, line 16: Instead of adding to the `main.tsx` file, would it be better to create a separate entry for Console? If we add it to `webpack.config.js`.


```js

module.exports = {
  ...
  plugins: [
    new NxAppWebpackPlugin({
      ...
      additionalEntryPoints: [
        {
          entryName: 'main.console',
          entryPath: 'graph/client/src/main.console.tsx',
        },
      ],
    }),
    ...
  ],
};
```

Note: I found a bug with entryPath not supporting relative paths. In the meantime the above will work. And in Console, you can check if `main.console.js` exists or not.

graph/ui-project-details/src/lib/project-details/project-details.tsx
MaxKless: this is still WIP, will remove once it's finalized

graph/ui-project-details/src/lib/project-details/project-details.tsx
MaxKless: it looks nicer and I updated it to be types but I don't like having to pass a `type` property since it affects all the places where the component is used. I'd just stick with inferring the type from `project` vs `projects`

graph/client/src/main.tsx
MaxKless: Yes, this was the very first approach I had. What I don't like about this is that it bundles everything in that seperate entry file, so the graph bundle will go from 3mb -> 5mb. 
It's not the end of the world but it didn't feel great so I went with the current solution instead. Lmk what you think about that :)

jaysoo (approved):

jaysoo:
I tested this on latest and canary Nx releases. Can verify that everything is working with the old `window.externalApi.openProjectDetails(...)` flow and the new `renderPDV(...)` flow.

PR #27843, opened 2024-09-09
chore(repo): enable db cache

FrozenPandaz:
## Current Behavior

The new db cache is not being used.

## Expected Behavior

The new sqlite db cache is being used.

## Related Issue(s)

Fixes #

### DIFF
--- a/.local.env
+++ b/.local.env
@@ -1,0 +1,1 @@
+NX_DB_CACHE=true


AgentEnder (approved):

eliellis:
Hi @FrozenPandaz, watching this work here, and specifically the changes being made to the `TaskOrchestrator` around the internal cache implementation. Presumably everything will move to the `DbCache` approach in Nx 21. I am wondering how this will work for those of us currently using our own `RemoteCache` implementations. Are there any plans to support something similar to the current approach where the `defaultTaskRunner` accepts a `RemoteCache` implementation and that propagates into the appropriate slots within the `TaskOrchestrator`? 

Just looking at things right now (which may be in a transitional state, so I'm not sure), it seems like the `getCache` function, and `DbCache` implementation (as written) would prevent these custom `RemoteCache` implementations from functioning given that the [check](https://github.com/nrwl/nx/blob/7228609d8c8889c672e967770372d28fa1820418/packages/nx/src/tasks-runner/cache.ts#L33) being run is only accounting for circumstance where Nx Cloud is being used, and further on, the [_getRemoteCache](https://github.com/nrwl/nx/blob/7228609d8c8889c672e967770372d28fa1820418/packages/nx/src/tasks-runner/cache.ts#L132) returns `null` for circumstances where Nx Cloud is not being used as well. Would it be possible to continue to propagate these remote cache implementations in `DbCache`? Just thinking out loud, but could `DbCache` be adjusted to take an optional `RemoteCacheV2` implementation in its constructor options, and `_getRemoteCache` could short-circuit if this value has already been set on the `DbCache` instance?

PR #27853, opened 2024-09-10
fix(angular): generate serve-static only if e2eTestRunner is passed

DariusPa:
Newly created app should not generate serve-static target under project.json if e2eTestRunner flag is not passed


Coly010:
Closed in favour of: https://github.com/nrwl/nx/pull/27924

PR #27858, opened 2024-09-10
feat(misc): remove @nrwl/ scoped packages

jaysoo:
This PR removes the `@nrwl/*` packages. Also cleans up references to the legacy packages.

NOTE: The rescope page is now under `/deprecated/rescope`. There is still >1000K monthly traffic to it, which might be due to older Nx users.

## Related Issue(s)

Fixes #


isaacplmann (changes requested):
Need to update the paragraph in rescope deprecated page to indicate that the `@nrwl/` packages are no longer being published to npm

isaacplmann (approved):

FrozenPandaz (approved):

PR #27861, opened 2024-09-10
Update affected.md

nategreen-dh:
Remove extra whitespace and pipe character that was breaking the table of options

## Current Behavior

image

## Expected Behavior

Table fixed in docs for `affected` command.


nategreen-dh:
Closing since it looks like it's already fixed in master.

PR #27882, opened 2024-09-11
fix(nx-cloud): include nxCloudId when generating connect urls

lourw:
## Expected Behavior

NxCloudId should be part of the URL that is created when connecting a new workspace.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/nx-cloud/utilities/is-workspace-claimed.ts
+++ b/packages/nx/src/nx-cloud/utilities/is-workspace-claimed.ts
@@ -1,15 +1,16 @@
-import { getCloudUrl } from './get-cloud-options';
+import { getCloudUrl, isNxCloudId } from './get-cloud-options';
 
-export async function isWorkspaceClaimed(nxCloudAccessToken) {
-  if (!nxCloudAccessToken) return false;
+export async function isWorkspaceClaimed(accessToken: string) {
+  if (!accessToken) return false;
 
   const apiUrl = getCloudUrl();
   try {
+    const requestData = isNxCloudId(accessToken)
+      ? { nxCloudId: accessToken }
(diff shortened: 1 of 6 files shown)


packages/nx/src/nx-cloud/utilities/is-workspace-claimed.ts
```diff
@@ -1,15 +1,16 @@
...
+  if (!accessToken) return false;
 
   const apiUrl = getCloudUrl();
   try {
+    const requestData = isNxCloudId(accessToken)
+      ? { nxCloudId: accessToken }
+      : { nxCloudAccessToken: accessToken };
     const response = await require('axios').post(
       `${apiUrl}/nx-cloud/is-workspace-claimed`,
-      {
```
lourw, line 10: API for this endpoint had changed since it was written. You need to specify the type of token being passed to the API

packages/nx/src/nx-cloud/utilities/is-workspace-claimed.ts
```diff
@@ -1,15 +1,16 @@
...
+import { getCloudUrl, isNxCloudId } from './get-cloud-options';
 
-export async function isWorkspaceClaimed(nxCloudAccessToken) {
-  if (!nxCloudAccessToken) return false;
+export async function isWorkspaceClaimed(accessToken: string) {
+  if (!accessToken) return false;
 
   const apiUrl = getCloudUrl();
   try {
+    const requestData = isNxCloudId(accessToken)
```
FrozenPandaz, line 8: We should be able to tell when invoking this function whether or not it is an ID or an access token based on where we read it/created it from.

Let's address this separately.

FrozenPandaz (approved):

ndcunningham (approved):

PR #27885, opened 2024-09-11
fix(webpack): handle relative paths for additionalEntryPath

jaysoo:
The `NxAppWebpackPlugin` does not support relative paths in `additionalEntryPoints`.

So this will fail:

```js
new NxAppWebpackPlugin({
  ...
  additionalEntryPoints: ['.src/foo.ts']
```

The resolved path is relative to workspace root when it should be project root.

## Current Behavior
Build will fail.
## Expected Behavior
Build should work.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/webpack/src/webpack.test.ts
+++ b/e2e/webpack/src/webpack.test.ts
@@ -132,8 +132,12 @@ describe('Webpack Plugin', () => {
 
   it('should be able to build with NxWebpackPlugin and a standard webpack config file', () => {
     const appName = uniq('app');
-    runCLI(`generate @nx/web:app ${appName} --bundler webpack`);
+    runCLI(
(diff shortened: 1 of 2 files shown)


e2e/webpack/src/webpack.test.ts
```diff
@@ -160,8 +171,9 @@ describe('Webpack Plugin', () => {
 
     runCLI(`build ${appName}`);
 
-    let output = runCommand(`node dist/${appName}/main.js`);
-    expect(output).toMatch(/Hello/);
+    expect(runCommand(`node dist/${appName}/main.js`)).toMatch(/Hello/);
```
ndcunningham, line 174: Should this be `expect(runCommand(`node dist/apps/${appName}/main.js`)).toMatch(/Hello/);`

ndcunningham (approved):

PR #27904, opened 2024-09-13
docs(core): powerpack docs

isaacplmann:
- ✅ Activate powerpack recipe
- ✅ Powerpack owners documentation
- [x] Powerpack custom remote cache documentation
- [x] Powerpack conformance documentation

Infrastructure for powerpack docs

- Adds the ability to generate API docs from ocean packages

To generate API documentation for plugins in the ocean repository, run the `nx documentation` command with the `NX_OCEAN_RELATIVE_PATH` environment variable set to the relative path to your checked out copy of the ocean repo.

```
NX_OCEAN_RELATIVE_PATH=../ocean nx documentation
```

This will create generated API documentation in the `docs/external-generated` folder. This API will be merged into the normal `docs/generated` documentation when the docs site is built.

Because there are two separate output folders, if someone runs `nx documentation` without the `NX_OCEAN_RELATIVE_PATH` environment variable, the ocean documentation will not be overwritten. The ocean documentation will only be updated or deleted when someone explicitly chooses to do so.


docs/external-generated/packages/powerpack-s3-cache/documents/overview.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Overview of the Nx powerpack-s3-cache Plugin
+description: The Nx Plugin for esbuild contains executors and generators that support building applications using esbuild. This page also explains how to configure esbuild on your Nx workspace.
```
JamesHenry, line 3: Why do they all have this esbuild description?

isaacplmann: Just placeholders

FrozenPandaz (approved):

PR #27909, opened 2024-09-13
docs(nx-cloud): update reference to start-ci-run and complete-ci-run

lourw:
## Expected Behavior

Updated references about start-ci-run to include mentions of warnings that you get if you don't include the `--distribute-on` flag. 

Also removed references to the legacy `--stop-all-agents` command in favor of `--complete-ci-run`.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/nx-cloud/reference/nx-cloud-cli.md
+++ b/docs/nx-cloud/reference/nx-cloud-cli.md
@@ -80,14 +80,13 @@ You can configure your CI run by passing the following flags:
 
 ### --distribute-on
 
-Tells Nx Cloud how many agents to use (and what launch templates to use) to distribute tasks. E.g.,
+By default, `npx nx-cloud start-ci-run` is intended for use with [Nx Agents](/ci/features/distribute-task-execution) and expects `--distribute-on` to be configured. It will output a warning if this flag is not set. If you are running a distributed execution with a legacy setup without Nx Agents, you can pass `--distribute-on=manual` to disable this warning.
+
+This command tells Nx Cloud how many agents to use (and what launch templates to use) to distribute tasks. E.g.,
 `npx nx-cloud start-ci-run --distribute-on="8 linux-medium-js"` will distribute CI using 8 agents that are initialized
 using the `linux-medium-js` launch template.
 
-You can use different types of launch templates as follows:
-`npx nx-cloud start-ci-run --distribute-on="3 linux-medium-js, 3 linux-large-js"`.
-
-You can also define the configuration in a file and reference it as follows:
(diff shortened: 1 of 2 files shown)


docs/nx-cloud/reference/nx-cloud-cli.md
```diff
@@ -182,10 +177,10 @@ individual commands as follows:
 {% /tab %}
 {% /tabs %}
 
-## npx nx-cloud stop-all-agents
-
-Same as `npx nx-cloud complete-ci-run`.
```
StalkAltan, line 187: This should stay documented in some way -- tons of users do not touch their pipelines very often and we generated with this command for years.

I would either not touch this, or if we really want to change the text, add something about how `complete-ci-run` should be the preferred command.

docs/nx-cloud/troubleshooting/ci-execution-failed.md
```diff
@@ -16,9 +16,9 @@ This behavior can be disabled by setting the [--require-explicit-completion](/ci
 
 ### A command was issued to stop all Nx Cloud agents
 
-Nx Cloud provides two commands to forcibly stop agents, [stop-all-agents and complete-ci-run](/ci/reference/nx-cloud-cli#npx-nxcloud-stopallagents).
+Nx Cloud provides a command to stop all agents: [npx nx-cloud complete-ci-run](/ci/reference/nx-cloud-cli#npx-nxcloud-completecirun).
```
StalkAltan, line 19: Since `stop-all-agents` does result in this behavior still, I think it should be mentioned here. We can make add something to point out that it's no longer the recommended command, but it will be confusing for someone that doesn't use `complete-ci-run` to come view the paragraph and see a totally separate command listed as the cause

docs/nx-cloud/reference/nx-cloud-cli.md
```diff
@@ -80,14 +80,13 @@ You can configure your CI run by passing the following flags:
 
 ### --distribute-on
 
-Tells Nx Cloud how many agents to use (and what launch templates to use) to distribute tasks. E.g.,
+By default, `npx nx-cloud start-ci-run` is intended for use with [Nx Agents](/ci/features/distribute-task-execution) and expects `--distribute-on` to be configured. It will output a warning if this flag is not set. If you are running a distributed execution with a legacy setup without Nx Agents, you can pass `--distribute-on=manual` to disable this warning.
```
isaacplmann, line 83: Move this paragraph to be the second paragraph in this section.  The definition of the flag should be the first paragraph and then this caveat paragraph

isaacplmann (approved):
Approved with a suggestion to reorder some text.  You'll be able to make the change and immediately merge.

PR #27930, opened 2024-09-16
fix(core): add flag to disable the db and logs when there are errors …

FrozenPandaz:
…connecting to the db

## Current Behavior

There is no flag to disable the DB and little logs for when Nx fails to connect to the DB.

## Expected Behavior

There is a `NX_DISABLE_DB=true` flag which can be used to disable usage of the DB. There are more logs which should help us identify what exactly is failing when connecting to the DB.

## Related Issue(s)

Fixes #


packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -663,7 +663,9 @@ function constructLifeCycles(lifeCycle: LifeCycle) {
   }
   if (!isNxCloudUsed(readNxJson())) {
     lifeCycles.push(
-      !IS_WASM ? new TaskHistoryLifeCycle() : new LegacyTaskHistoryLifeCycle()
+      process.env.NX_DISABLE_DB === 'true' || !IS_WASM
```
JamesHenry, line 666:
```suggestion
      process.env.NX_DISABLE_DB !== 'true' && !IS_WASM
```

JamesHenry (approved):

AgentEnder (approved):

PR #27932, opened 2024-09-16
cleanup(js): replace `fs-extra` with `node:fs`

pralkarz:
This PR is a suggestion to replace `fs-extra` with the native `node:fs` to reduce the dependency count and the bundle size.

## Current Behavior
`@nx/js` uses `fs-extra` which, at present, mostly wraps the native `node:fs` API.

The biggest change is `fs-extra.copySync` -> `fs.cpSync`. The latter has been added in Node 16.7.0, and has been marked as "experimental" since, dropping that status in Node 22. The API seems stable across the releases though, and the affected tests pass, so this change might be acceptable. Other possible solutions:
1. Keep `fs-extra` just for this function, and eventually remove it in the distant future where Node version older than 22 are deemed obsolete.
2. Mirror the `cpSync` behavior by writing a utility that uses stable methods `fs.writeFile`, `fs.stat`, etc.

## Expected Behavior
Reduced dependency count and bundle size by replacing `fs-extra` with `node:fs`.

## Related Issue(s)
N/A


JamesHenry (changes requested):
Thank you for this @ziebam.

It could possibly cause issues for any in flight branches/PRs. To help make things easier to both figure out and enforce, please add something like the following ESLint rule to the js package .eslintrc.json

```json
"no-restricted-imports": ["error", {
    "name": "fs-extra",
    "message": "Please use equivalent utilties from `node:fs` instead."
}]
```

ziebam:
> Thank you for this @ziebam.
> 
> It could possibly cause issues for any in flight branches/PRs. To help make things easier to both figure out and enforce, please add something like the following ESLint rule to the js package .eslintrc.json
> 
> ```json
> "no-restricted-imports": ["error", {
>     "name": "fs-extra",
>     "message": "Please use equivalent utilties from `node:fs` instead."
> }]
> ```

Done. I also merged with `master`, although that might have erroneously triggered GitHub to ask another person for a review, apologies.

JamesHenry:
@ziebam So you tested this catches cases of importing from fs-extra, yeah?

ziebam:
> @ziebam So you tested this catches cases of importing from fs-extra, yeah?

Yup, here's a screenshot for the record.
image

Not sure about the CI failure, investigating right now.

JamesHenry (approved):

JamesHenry:
Thanks @ziebam

PR #27940, opened 2024-09-16
fix(core): import should be rebasable

jaysoo:
When users run `nx import` and open a PR with the imported package, they should be able to use `git rebase --rebase-merges` to update their branch with latest main.

## Why this matters

If the repo requires the PR to be updated with latest main before merge, the user must have a way to do this and avoid having to do the import process all over again.

## Current Behavior
Rebase results in git error

## Expected Behavior
Rebase works

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/import/import.ts
+++ b/packages/nx/src/command-line/import/import.ts
@@ -348,8 +348,9 @@ export async function importHandler(options: ImportOptions) {
     title: `Merging these changes into ${getBaseRef(nxJson)}`,
     bodyLines: [
       `MERGE these changes when merging these changes.`,
-      `Do NOT squash and do NOT rebase these changes when merging these changes.`,
-      `If you would like to UNDO these changes, run "git reset HEAD~1 --hard"`,
+      `Do NOT squash these commits when merging these changes.`,
+      `If you rebase, make sure to use "--rebase-merges" to preserve merge commits.`,
+      `To UNDO these changes, run "git reset HEAD~1 --hard"`,
     ],
   });
 }
--- a/packages/nx/src/command-line/import/utils/prepare-source-repo.ts
+++ b/packages/nx/src/command-line/import/utils/prepare-source-repo.ts
@@ -19,55 +19,24 @@ export async function prepareSourceRepo(
     join(gitClient.root, source)
   );
 
-  if (relativeSourceDir !== '') {
-    if (await gitClient.hasFilterRepoInstalled()) {
-      spinner.start(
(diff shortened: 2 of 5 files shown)


packages/nx/src/command-line/import/utils/prepare-source-repo.ts
```diff
@@ -19,56 +19,25 @@ export async function prepareSourceRepo(
...
+    spinner.start(
+      `Filtering git history to only include files in ${relativeSourceDir}`
     );
-  }
-
-  const destinationInSource = join(gitClient.root, relativeDestination);
-  spinner.start(`Moving files and git history to ${destinationInSource}`);
-
-  // The result of filter-branch will contain only the files in the subdirectory at its root.
-  const files = await gitClient.getGitFiles('.');
```
jaysoo, line 43: NOTE: Moving files is no longer needed since both `git filter-repo` and `git filter-branch` can handle the move automatically, in a way that allows for rebase.

packages/nx/src/utils/git-utils.tree-filter.ts
```diff
@@ -0,0 +1,38 @@
...
+    .split('\x00')
+    .map((s) => s.trim())
+    .filter(Boolean);
+  const srcRegex = new RegExp(`^${src}`);
+  for (const file of files) {
+    if (src === '' || srcRegex.test(file)) {
+      // If source and destination are the same, then keep the file as is.
+      if (src === dest) continue;
+      const destFile = posix.join(dest, file.replace(srcRegex, ''));
+      const dir = posix.dirname(destFile);
```
jaysoo, line 27: NOTE: We have to use POSIX here, even on Windows machine.

Coly010 (approved):

packages/nx/src/command-line/import/import.ts
```diff
@@ -348,8 +348,9 @@ export async function importHandler(options: ImportOptions) {
     title: `Merging these changes into ${getBaseRef(nxJson)}`,
     bodyLines: [
       `MERGE these changes when merging these changes.`,
-      `Do NOT squash and do NOT rebase these changes when merging these changes.`,
-      `If you would like to UNDO these changes, run "git reset HEAD~1 --hard"`,
+      `Do NOT squash these commits when merging these changes.`,
+      `If you rebase, make sure to use "--rebase-merges" to preserve merge commits.`,
```
FrozenPandaz, line 352: I'm guessing github does this under the hood? Hopefully it's the same for other VCS :thinking:

FrozenPandaz (approved):

PR #27958, opened 2024-09-17
fix(core): nx import detects plugins synchronously

jaysoo:
This PR switches `globWithWorkspaceContext` with `globWithWorkspaceContextSync` for `nx init` and `nx import`, since the latter isn't guaranteed to have updated file map during the init/import process.

Note: Also updates Gradle import test to ensure the destination repo does not have uncommitted changes.

## Current Behavior
Sometimes plugins are not detected.

## Expected Behavior
Always detect plugins.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/init/init-v2.ts
+++ b/packages/nx/src/command-line/init/init-v2.ts
@@ -22,7 +22,7 @@ import {
 import { prompt } from 'enquirer';
 import { execSync } from 'child_process';
 import { addNxToAngularCliRepo } from './implementation/angular';
-import { globWithWorkspaceContext } from '../../utils/workspace-context';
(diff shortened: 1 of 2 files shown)


xiongemi (approved):

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -22,7 +22,10 @@ import {
 import { prompt } from 'enquirer';
 import { execSync } from 'child_process';
 import { addNxToAngularCliRepo } from './implementation/angular';
-import { globWithWorkspaceContext } from '../../utils/workspace-context';
+import {
+  globWithWorkspaceContext,
```
FrozenPandaz, line 26: Is this unused now?

FrozenPandaz (approved):

PR #27967, opened 2024-09-18
cleanup(misc): migrate from chalk to picocolors

43081j:
Migrates from `chalk` to the much smaller and faster `picocolors`.

Part of a wider ecosystem cleanup, we are moving many projects over to picocolors or node's native styleText. 

If you let me know you're happy with this change, I'll push a branch for each of the other packages to do the same. 

I will catch up from master when I'm back on my laptop 👍


packages/create-nx-plugin/package.json
```diff
@@ -30,7 +30,7 @@
   "homepage": "https://nx.dev",
   "dependencies": {
     "create-nx-workspace": "file:../create-nx-workspace",
-    "chalk": "^4.1.0",
+    "picocolors": "^1.1.0",
```
JamesHenry, line 33: The repo doesn't use workspaces, so please add this to the root package.json too

JamesHenry:
Thanks @43081j this is very similar in nature to the recently merged: https://github.com/nrwl/nx/pull/27932

Please can you implement the same feedback I gave on there - add the ESLint rule to this project to catch imports of chalk and give the feedback that picocolors should be used instead

packages/create-nx-plugin/package.json
JamesHenry: This is at least in part why your CI run failed

JamesHenry (changes requested):
See comments

JamesHenry:
This change should really be `cleanup`, it should be imperceptible to the end user and not fix a bug nor add a feature.

packages/create-nx-plugin/package.json
43081j: Ah I somehow totally missed that. Had been wondering what kind of setup it was meant to be

Thanks for the explanation

43081j:
Makes sense to me 👍 

Wasn't sure what the correct scope and type of change was. Thanks for the tip

Will resolve the comments soon

43081j:
I'm struggling to understand the current failure

any chance you could help me out? 👀  I think its a failing test but I cant seem to find where in the circleci output we can get hold of which test

JamesHenry:
@43081j I believe it was a memory issue, I have fixed up the lock file conflict and pushed

JamesHenry (approved):

JamesHenry:
Thanks again!

PR #27970, opened 2024-09-18
feat(core): handle prettier not installed when running nx format

leosvelperez:
## Current Behavior

When `prettier` is not installed, running `nx format` results in an uncaught error.

## Expected Behavior

When `prettier` is not installed, running `nx format` should handle the error and log a more appropriate message.

## Related Issue(s)

Fixes #


JamesHenry (changes requested):
Pending discussion

JamesHenry (approved):

PR #27999, opened 2024-09-19
fix(module-federation): collect secondary entry points from exports #26878

### LINKED ISSUE #26878: Nx + Module federation + Angular gives error NG0912 when using ng-zorro-antd
### Current Behavior

Switching between different pages is given us the following warnings:

NG0912: Component ID generation collision detected. Components 'NzSelectClearComponent' and 'NzSelectClearComponent' with selector 'nz-select-clear' generated the same component ID. To fix this, you can change the selector of one of those components or add an extra host attribute to force a different ID. Find more at https://angular.dev/errors/NG0912
NG0912: Component ID generation collision detected. Components 'NzTextCopyComponent' and 'NzTextCopyComponent' with selector 'nz-text-copy' generated the same component ID. To fix this, you can change the selector of one of those components or add an extra host attribute to force a different ID. Find more at https://angular.dev/errors/NG0912

### Expected Behavior

No warning (for us this warning cause errors in production)

### GitHub Repo
[nx-mfe-ng-zorro](https://github.com/Sir-J/nx-mfe-ng-zorro)

### Steps to Reproduce

1. create angular workspace
2. add to package.json `ng-zorro-antd`: `18.0.1`
3. create host application
4. crate remote application
5. start

### Nx Report

```shell

Node   : 20.13.1
OS     : darwin-arm64
npm    : 10.5.2

nx (global)        : 19.0.4
nx                 : 19.4.2
@nx/js             : 19.4.2
@nx/jest           : 19.4.2
@nx/linter         : 19.4.2
@nx/eslint         : 19.4.2
@nx/workspace      : 19.4.2
@nx/angular        : 19.4.2
@nx/cypress        : 19.4.2
@nx/devkit         : 19.4.2
@nx/eslint-plugin  : 19.4.2
@nx/node           : 19.4.2
@nrwl/tao          : 19.4.2
@nx/web            : 19.4.2
@nx/webpack        : 19.4.2
nx-cloud           : 19.0.0
typescript         : 5.4.5
Community plugins:
ng-zorro-antd : 18.0.1
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information
This issue is reproduced when upgrading Angular and `ng-zorro-antd` to version 17 and more. In version 16 this problem not reproduced

Coly010:
## Current Behavior

Some packages that use `exports` in the package.json may point to a directory such as `esm/<entrypoint>`. 
They would still be imported such as `mypackage/secondary-point`.
This needs to be shared still for Module Federation.

However, our current logic for finding secondary entry points for packages looks at the sub dirs of the package and tries to match it with the exports fiels.

Therefore if `mypackage/secondary-point` was set up such that

```
"./secondary-point": {
  default: "esm/secondary-point/index.mjs"
}
```

then it would never be found correctly, causing share problems.

## Expected Behavior

exports of secondary entrypoints should be shared correctly

## Related Issue(s)

Fixes #26878

### DIFF
--- a/packages/webpack/src/utils/module-federation/secondary-entry-points.ts
+++ b/packages/webpack/src/utils/module-federation/secondary-entry-points.ts
@@ -105,6 +105,33 @@ export function recursivelyCollectSecondaryEntryPointsFromDirectory(
   }
 }
 
+function collectPackagesFromExports(
+  pkgName: string,
+  pkgVersion: string,
+  exports: any | undefined,
+  collectedPackages: {
+    name: string;
+    version: string;
+  }[]
+): void {
+  for (const [relativeEntryPoint, exportOptions] of Object.entries(exports)) {
+    if (exportOptions?.['default']?.search(/\.(js|mjs|cjs)$/)) {
+      let entryPointName = joinPathFragments(pkgName, relativeEntryPoint);
+      if (entryPointName.endsWith('.json')) {
+        entryPointName = dirname(entryPointName);
+      }
+      if (entryPointName === '.') {
+        continue;
+      }
+      if (collectedPackages.find((p) => p.name === entryPointName)) {
+        continue;
+      }
+
+      collectedPackages.push({ name: entryPointName, version: pkgVersion });
+    }
+  }
+}
+
 export function collectPackageSecondaryEntryPoints(
   pkgName: string,
   pkgVersion: string,
@@ -130,6 +157,9 @@ export function collectPackageSecondaryEntryPoints(
   }
 
   const { exports } = packageJson;
+  if (exports) {
+    collectPackagesFromExports(pkgName, pkgVersion, exports, collectedPackages);
+  }
   const subDirs = getNonNodeModulesSubDirs(pathToPackage);
   recursivelyCollectSecondaryEntryPointsFromDirectory(
     pkgName,
--- a/packages/webpack/src/utils/module-federation/share.spec.ts
+++ b/packages/webpack/src/utils/module-federation/share.spec.ts
@@ -163,20 +163,110 @@ describe('MF Share Utils', () => {
       ]);
       // ASSERT
       expect(packages).toEqual({
-        '@angular/core': {
-          singleton: true,
-          strictVersion: true,
-          requiredVersion: '~13.2.0',
-        },
         '@angular/common': {
+          requiredVersion: '~13.2.0',
           singleton: true,
           strictVersion: true,
+        },
+        '@angular/common/http': {
           requiredVersion: '~13.2.0',
+          singleton: true,
+          strictVersion: true,
+        },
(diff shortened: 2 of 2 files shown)


packages/webpack/src/utils/module-federation/secondary-entry-points.ts
```diff
@@ -105,6 +106,36 @@ export function recursivelyCollectSecondaryEntryPointsFromDirectory(
...
+  exports: any | undefined,
+  collectedPackages: {
+    name: string;
+    version: string;
+  }[]
+): void {
+  for (const [relativeEntryPoint, exportOptions] of Object.entries(exports)) {
+    if (
+      exportOptions['default'] &&
+      exportOptions['default'].search(/\.(js|mjs|cjs)$/)
```
jaysoo, line 121: Nit: `if (exportOptions['default']?.search(/\.(js|mjs|cjs)$/)`)

jaysoo (approved):

goldsam:
@Coly010 Thank you for addressing this! I was able to drastically simplify my config.

goldsam:
@Coly010 This actually still isn't quite right. The `ng-zorro-antd` package uses a structure where the root package.json specifies exports as well as a version. The exported directories then contain a `package.json` which omits a version. For example, this is what the file `ng-zorro-antd/alert/package.json` looks like:
```json
{
  "module": "../fesm2022/ng-zorro-antd-alert.mjs"
}
```
The code should respect a valid "version" in the `package.json` closest to the import.

packages/webpack/src/utils/module-federation/secondary-entry-points.ts
```diff
@@ -105,6 +105,33 @@ export function recursivelyCollectSecondaryEntryPointsFromDirectory(
...
+  pkgName: string,
+  pkgVersion: string,
+  exports: any | undefined,
+  collectedPackages: {
+    name: string;
+    version: string;
+  }[]
+): void {
+  for (const [relativeEntryPoint, exportOptions] of Object.entries(exports)) {
+    if (exportOptions?.['default']?.search(/\.(js|mjs|cjs)$/)) {
```
rathpc, line 118: @Coly010 This causes the following error in some cases because for whatever reason search is not a function and it does not handle that gracefully.

```
 NX   exportOptions?.default?.search is not a function
```

Can we update this to the following to have a bit more safety around this and prevent that error?

```suggestion
    if (exportOptions?.['default']?.search?.(/\.(js|mjs|cjs)$/)) {
```

rathpc:
@Coly010 I created an issue for the problem mentioned above https://github.com/nrwl/nx/issues/28139

PR #28009, opened 2024-09-20
fix(linter): do not generate docs.recommended property

JamesHenry:
## Current Behavior

We still generate `docs.recommended` which is not on the types.

## Expected Behavior

We do not generate `docs.recommended`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint/src/migrations/update-17-1-0/update-typescript-eslint.spec.ts
+++ b/packages/eslint/src/migrations/update-17-1-0/update-typescript-eslint.spec.ts
@@ -84,7 +84,6 @@ export const rule = ESLintUtils.RuleCreator(() => __filename)({
     type: 'problem',
     docs: {
       description: \`\`,
-      recommended: 'error',
     },
     schema: [],
     messages: {},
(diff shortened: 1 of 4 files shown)


jaysoo (approved):

packages/eslint/src/migrations/update-17-1-0/update-typescript-eslint.spec.ts
```diff
@@ -84,7 +84,6 @@ export const rule = ESLintUtils.RuleCreator(() => __filename)({
     type: 'problem',
     docs: {
       description: \`\`,
-      recommended: 'error',
```
jaysoo, line 87: This was an old migration, I don't think it technically matters if it's updated or not?

FrozenPandaz (approved):

packages/eslint/src/migrations/update-17-1-0/update-typescript-eslint.spec.ts
JamesHenry: It didn't ever serve any purpose so it's just setting people up for an issue doing the road if they run the migration right now

PR #28010, opened 2024-09-20
chore(misc): add custom lint rule for ensuring pnpm-lock.yaml version

JamesHenry:
## Current Behavior

It is possible for all contributors to (accidentally or otherwise) quietly change the version of our pnpm lockfile as part of unrelated PRs

## Expected Behavior

A custom lint rule ensures our lock file version doesn't change on unrelated PRs.

Updating the pnpm lock file version is a very explicit and clear action that requires both updating the lock file and updating the lint rule config.

---

image


.eslintrc.json
```diff
@@ -64,6 +64,16 @@
...
         "@nx/workspace/valid-command-object": "error"
       }
+    },
+    {
+      "files": ["pnpm-lock.yaml"],
+      "parser": "./tools/eslint-rules/raw-file-parser.js",
+      "rules": {
+        "@nx/workspace/ensure-pnpm-lock-version": [
+          "error",
+          { "version": "9.0" }
```
jaysoo, line 74: Should we read this from `packageManager` in `package.json`, so we don't have two places to specify version?

JamesHenry: Discussed on slack

jaysoo (approved):

PR #28019, opened 2024-09-20
cleanup(core): replace `fs-extra` with `node:fs`

pralkarz:

## Current Behavior
`fs-extra` is used.

## Expected Behavior
`node:fs` is used.

## Related Issue(s)
N/A


packages/nx/src/daemon/cache.ts
```diff
@@ -35,7 +30,9 @@ export function deleteDaemonJsonProcessCache(): void {
 export async function writeDaemonJsonProcessCache(
   daemonJson: DaemonProcessJson
 ): Promise<void> {
-  await writeJson(serverProcessJsonPath, daemonJson);
+  // TODO: Ask about it in the PR.
+  // `fs-extra` appends an extra newline at the end when writing JSON. Do we want it here?
+  await writeFile(serverProcessJsonPath, JSON.stringify(daemonJson));
```
ziebam, line 35: As per the comment. What do you think?

packages/nx/src/command-line/init/implementation/add-nx-to-nest.ts
```diff
@@ -22,7 +22,6 @@ import {
 } from './utils';
 import { nxVersion } from '../../../utils/versions';
 import { connectExistingRepoToNxCloudPrompt } from '../../connect/connect-to-nx-cloud';
-
```
JamesHenry, line 25: Please revert removing this newline

packages/nx/src/command-line/release/config/version-plans.ts
```diff
@@ -54,7 +54,9 @@ const versionPlansDirectory = join('.nx', 'version-plans');
 
 export async function readRawVersionPlans(): Promise<RawVersionPlan[]> {
   const versionPlansPath = getVersionPlansAbsolutePath();
-  const versionPlansPathExists = await pathExists(versionPlansPath);
+  const versionPlansPathExists = await access(versionPlansPath)
+    .then(() => true)
+    .catch(() => false);
   if (!versionPlansPathExists) {
```
JamesHenry, line 60: Honestly can we just simplify this such that the if just checks `existsSync()`? I'm not sure why pathExists was used for this one in the first place, and we don't need to reuse `versionPlansPathExists`

packages/nx/src/command-line/release/plan.ts
```diff
@@ -355,7 +356,7 @@ async function _promptForMessage(versionPlanName: string): Promise<string> {
       );
       writeFileSync(messageFilePath, '');
       await launchEditor(messageFilePath);
-      message = readFileSync(messageFilePath, 'utf-8');
+      message = readFileSync(messageFilePath, 'utf8');
```
JamesHenry, line 359: `utf-8` is already a valid encoding value and it's more readable, please revert

```
type BufferEncoding =
            | "ascii"
            | "utf8"
            | "utf-8"
            | "utf16le"
            | "ucs2"
            | "ucs-2"
            | "base64"
            | "base64url"
            | "latin1"
            | "binary"
            | "hex";
```

packages/nx/src/daemon/cache.ts
JamesHenry: We have our own `writeJsonFile` and `readJsonFile` utils that are already built upon `node:fs` in `packages/nx/src/utils/fileutils.ts`, please could you use those instead? You can set the appendNewLine option to make it match the previous behavior

packages/nx/src/daemon/client/client.ts
```diff
@@ -574,8 +573,8 @@ export class DaemonClient {
   }
 
   async startInBackground(): Promise<ChildProcess['pid']> {
-    ensureDirSync(DAEMON_DIR_FOR_CURRENT_WORKSPACE);
-    ensureFileSync(DAEMON_OUTPUT_LOG_FILE);
+    mkdirSync(DAEMON_DIR_FOR_CURRENT_WORKSPACE, { recursive: true });
+    writeFileSync(DAEMON_OUTPUT_LOG_FILE, '');
```
JamesHenry, line 577: This doesn't seem to be equivalent to ensureFileSync: https://github.com/jprichardson/node-fs-extra/blob/d96f2655f181f1c6b1641711619400140a89d027/lib/ensure/file.js#L40

It won't write to the file if it already exists

packages/nx/src/internal-testing-utils/temp-fs.ts
```diff
@@ -47,13 +48,21 @@ export class TempFs {
   }
 
   async createFile(filePath: string, content: string) {
-    await outputFile(joinPathFragments(this.tempDir, filePath), content);
+    const dir = joinPathFragments(this.tempDir, dirname(filePath));
+    if (
+      !(await access(dir)
```
JamesHenry, line 53: existsSync would be much simpler and easier to read

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -300,7 +301,11 @@ export class Cache {
       await Promise.all(
         expandedOutputs.map(async (f) => {
           const src = join(this.root, f);
-          if (await pathExists(src)) {
+          if (
```
JamesHenry, line 304: existsSync again

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -338,7 +343,11 @@ export class Cache {
       await Promise.all(
         expandedOutputs.map(async (f) => {
           const cached = join(cachedResult.outputsPath, f);
-          if (await pathExists(cached)) {
+          if (
+            await access(cached)
```
JamesHenry, line 347: existsSync again

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -413,14 +422,15 @@ export class Cache {
     const tdCommit = join(this.cachePath, `${task.hash}.commit`);
     const td = join(this.cachePath, task.hash);
 
-    if (await pathExists(tdCommit)) {
-      const terminalOutput = await readFile(
-        join(td, 'terminalOutput'),
-        'utf-8'
-      );
+    if (
+      await access(tdCommit)
```
JamesHenry, line 426: existsSync again

packages/nx/src/utils/ignore.ts
```diff
@@ -55,7 +55,7 @@ export function getIgnoreObject(
 function getIgnoredGlobsFromFile(file: string, root?: string): string[] {
   try {
     const results = [];
-    const contents = readFileSync(file, 'utf-8');
+    const contents = readFileSync(file, 'utf8');
```
JamesHenry, line 58: Same as previous comment about utf-8

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -413,14 +422,15 @@ export class Cache {
...
+        .then(() => true)
+        .catch(() => false)
+    ) {
+      const terminalOutput = await readFile(join(td, 'terminalOutput'), 'utf8');
       let code = 0;
       try {
-        code = Number(await readFile(join(td, 'code'), 'utf-8'));
+        code = Number(await readFile(join(td, 'code'), 'utf8'));
```
JamesHenry, line 433: Same as previous comment about utf-8

JamesHenry (changes requested):

JamesHenry:
Thanks again @ziebam!

packages/nx/src/command-line/release/plan.ts
ziebam: I've changed this to match the Node docs, but I'm fine with reverting. 😄

packages/nx/src/daemon/cache.ts
ziebam: Changed per the request, but those utils are synchronous. Is this fine for you or should I add similar ones, but asynchronous?

packages/nx/src/command-line/init/implementation/add-nx-to-nest.ts
ziebam: Oops, it's snuck in by mistake.

packages/nx/src/daemon/client/client.ts
ziebam: Good catch! Should be good now. I think my simple check is sufficient in this case because we're calling `mkdirSync` before, therefore the parent directory should always exist.

ziebam:
https://staging.nx.app/runs/p5g9EF6lsY – not sure about the `nx:test` failure (passes locally). The failure in `cypress:test` is reproducible locally when I run the whole test suite, but it passes when I run just that failing file. My only guess so far is that we're bombarding the filesystem with too many syscalls and it refuses to create the directory before we try to write a file. Gonna investigate further tomorrow.

JamesHenry:
I’m sorry in advance if my insistence for preferring existsSync over await access ends up being significant somehow and the async code is important to be non blocking in some instances (or possibility the JSON utils).

ziebam:
> I’m sorry in advance if my insistence for preferring existsSync over await access ends up being significant somehow and the async code is important to be non blocking in some instances (or possibility the JSON utils).

No worries at all, especially since it seems that this wasn't an issue with `existsSync`. Please check my latest commit message for a guess as to what was the problem, although now I'm not sure why it'd work/not be flaky with the fully asynchronous solution. 🤔

JamesHenry:
Thanks @ziebam, I've requested review from @FrozenPandaz and @jaysoo as well

JamesHenry:
@ziebam Out of an abundance of caution, let's update to please keep the daemon cache interactions async, then this should be good to go

JamesHenry:
Please could you also remove the unused `safelyCleanUpExistingProcess()` function in the daemon cache?

ziebam:
> @ziebam Out of an abundance of caution, let's update to please keep the daemon cache interactions async, then this should be good to go

Sure, I added two new functions which are almost the same as the synchronous ones. I was considering adding an `async` flag to the original function instead, but then it would return `Promise<T> | T` which would require awaiting it in *many* places, so I decided to go with separate helpers in the end.

> Please could you also remove the unused `safelyCleanUpExistingProcess()` function in the daemon cache?

Done!

ziebam:
Can we retry the CI somehow? The failure doesn't seem related to my changes.

packages/nx/src/devkit-exports.ts
```diff
@@ -196,7 +196,12 @@ export type { JsonParseOptions, JsonSerializeOptions } from './utils/json';
 /**
  * @category Utils
  */
-export { readJsonFile, writeJsonFile } from './utils/fileutils';
+export {
```
JamesHenry, line 199: Please revert this bit, this is adding to our public API I believe, let's keep the changes internal

ziebam:
Hmm, no clue why it keeps failing. The only change that could impact that is switching the import from `'fs'` to `'node:fs'`. As far as I know they're exactly the same though. 🤔 Gonna investigate later tonight or tomorrow.

JamesHenry:
Sorry to add even more confusion @ziebam but I found this one today when merging some PRs: https://github.com/nrwl/nx/pull/27626

The PR author on our team is on paternity leave right now, but I will discuss internally and get back to you, we may want sync in the daemon cache read after all

ziebam:
> Sorry to add even more confusion @ziebam but I found this one today when merging some PRs: #27626
> 
> The PR author on our team is on paternity leave right now, but I will discuss internally and get back to you, we may want sync in the daemon cache read after all

No problem, I know very well how it goes in large projects like this one, and I appreciate your thoroughness!

As for the CI failure. I've got the same test to fail locally too, but at `fs.statSync` instead, not when creating the directory. It seems like removing the `node:` directive fixes that though, but I've no idea why that would be the case with fairly modern Node and Jest versions. Passing `NODE_OPTIONS=--experimental-vm-modules` didn't change anything in that regard.

My question is: are you fine with solving the failure by simply removing the `node:` directives in that file or should I try to dig further and really get down to why it would only fail in that single case, seeing as I used that directive in many other files and they work fine?

JamesHenry:
Yeah I don't love the idea of leaving that behind as a footgun when someone comes along later and makes things consistent but without this context. So please let's figure it out if we can!

ziebam:
> Yeah I don't love the idea of leaving that behind as a footgun when someone comes along later and makes things consistent but without this context. So please let's figure it out if we can!

It seems obvious in hindsight, but I forgot to change what's mocked from `'fs'` to `'node:fs'`, should be good now. 😅

JamesHenry:
@ziebam sorry for the conflict now but we decided to go ahead and merge that fix, so after all that we are sticking with sync there because we are hoping it fixes another bug.

Sorry again for the churn, we should be able to land this really soon once we tidy up this conflict.

JamesHenry (approved):
🎉

JamesHenry:
Thanks a lot again!

PR #28033, opened 2024-09-21
cleanup(core): migrate esbuild to use picocolors

43081j:

I used `core` here in the commit since we don't have a scope for the esbuild project


JamesHenry:
Thank you!

PR #28036, opened 2024-09-22
fix(core): use COPYFILE_EXCL when installing native binary

eblocha:
This will prevent a race condition when multiple nx packages of the same version are installed in parallel. For example, if a workspace has nx v19.0.0 installed in package.json, but is using ^19.0.0 for plugin versions, the plugins will install separate copies of nx in parallel. This could cause the native temp file to get clobbered and corrupted.

This change adds the COPYFILE_EXCL mode flag to the copy command, which will open the destination file with the O_EXCL flag. This ensures the destination file is not open by any other process at the same time.

## Current Behavior
If multiple instances of the same nx version is installed, the native bindings file can get clobbered by the post-install script.

## Expected Behavior
The post-install script should not write to the native bindings file in a way that suffers from the race condition.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/index.js
+++ b/packages/nx/src/native/index.js
@@ -1,5 +1,5 @@
 const { join, basename } = require('path');
-const { copyFileSync, existsSync, mkdirSync } = require('fs');
+const { copyFileSync, existsSync, mkdirSync, constants } = require('fs');
 const Module = require('module');
 const { nxVersion } = require('../utils/versions');
 const { getNativeFileCacheLocation } = require('./native-file-cache-location');
@@ -75,10 +75,21 @@ Module._load = function (request, parent, isMain) {
     if (existsSync(tmpFile)) {
       return originalLoad.apply(this, [tmpFile, parent, isMain]);
     }
-    if (!existsSync(nativeFileCacheLocation)) {
-      mkdirSync(nativeFileCacheLocation, { recursive: true });
+
+    mkdirSync(nativeFileCacheLocation, { recursive: true });
+
+    try {
+      copyFileSync(nativeLocation, tmpFile, constants.COPYFILE_EXCL);
+    } catch (e) {
+      if (e.code === 'EEXIST') {
+        // Another process wrote to the file at the same time.
+        // This can happen when multiple instances of the same nx versions are being installed in parallel.
+        // We can ignore this error because the other process is writing the same file content, but won over us.
+      } else {
+        throw e;
+      }
     }
-    copyFileSync(nativeLocation, tmpFile);
+    
     return originalLoad.apply(this, [tmpFile, parent, isMain]);
   } else {
     // call the original _load function for everything else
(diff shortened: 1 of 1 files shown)


packages/nx/src/native/index.js
```diff
@@ -75,10 +75,21 @@ Module._load = function (request, parent, isMain) {
...
     }
-    if (!existsSync(nativeFileCacheLocation)) {
-      mkdirSync(nativeFileCacheLocation, { recursive: true });
+
+    mkdirSync(nativeFileCacheLocation, { recursive: true });
+
+    try {
+      copyFileSync(nativeLocation, tmpFile, constants.COPYFILE_EXCL);
+    } catch (e) {
+      if (e.code === 'EEXIST') {
```
eblocha, line 84: I tested this logic locally:

image

packages/nx/src/native/index.js
```diff
@@ -75,10 +75,21 @@ Module._load = function (request, parent, isMain) {
     if (existsSync(tmpFile)) {
       return originalLoad.apply(this, [tmpFile, parent, isMain]);
     }
-    if (!existsSync(nativeFileCacheLocation)) {
-      mkdirSync(nativeFileCacheLocation, { recursive: true });
+
+    mkdirSync(nativeFileCacheLocation, { recursive: true });
+
+    try {
+      copyFileSync(nativeLocation, tmpFile, constants.COPYFILE_EXCL);
```
eblocha, line 82: This can still be a race condition, because the other process may not have finished writing the file by the time node tries to read it. I could add a loop to append an incrementing number to the file name until success or max retries to mitigate this, but that's not ideal since it creates extra garbage.

packages/nx/src/native/index.js
```diff
@@ -75,10 +75,21 @@ Module._load = function (request, parent, isMain) {
...
+    } catch (e) {
+      if (e.code === 'EEXIST') {
+        // Another process wrote to the file at the same time.
+        // This can happen when multiple instances of the same nx versions are being installed in parallel.
+        // We can ignore this error because the other process is writing the same file content, but won over us.
```
eblocha, line 87: Should there be some kind of warning message logged to warn about using pinned dependency versions?

eblocha:
I'm seeing this file get overwritten when running a build locally. Is this something that's auto-generated? I'm having trouble finding what is causing that to happen.

eblocha:
Hey @MaxKless sorry for the delay in replying here. I'll work on getting an example sometime this weekend.

The original issue was happening when I had two nx plugins installed with a non-exact version specifier, like `"@nx/some-plugin": "^19.0.0"`, however the nx dependency was an exact version, say `19.0.0`.

The plugins would install their _own_ copies of nx under `node_modules/@nx/some-plugin/node_modules/nx`.

If the two plugins happened to use the same version of `nx`, but _different_ from the top-level `nx`, then npm would install these two identical versions in parallel. Since the post-install script writes to a temp file using the nx version, these two parallel installs would try to write to the same file at the same time.

MaxKless:
Hey @eblocha! We've been making various changes to the native binary installation  recently also to prevent race conditions and such. 
I'll close this for now - let me know if you're still running into this though and we can reevaluate.

PR #28037, opened 2024-09-22
cleanup(core): move esbuild to use fdir/picomatch

43081j:
Migrates away from fast-glob to `fdir` and `picomatch`.

this'll improve performance a fair amount (fdir is much faster than fast-glob) and should reduce the install size (picomatch and fdir are both very small).

fast-glob uses micromatch, while this uses picomatch. the difference seems to be that more advanced bash brace expansion isn't supported in picomatch. if that's important to you, its possible we could use micromatch or [zeptomatch](https://github.com/fabiospampinato/zeptomatch)


JamesHenry (changes requested):
There's a conflict from one of your other PRs, and possibly some legit e2e failures

43081j:
No worries, I'll have a look soon 👍

Will catch the branch up too

JamesHenry:
Please ignore the react rspack module federation error, it will be resolved when you sync with latest master. The others might be legit though

JamesHenry:
@43081j I pulled in latest again after all the recent merges just in case

43081j:
That same test seems to fail locally for me too 

Haven't managed to figure out why yet though. It seems to expect a file to exist which doesn't. That suggests the way we compute the input glob is broken or changed but I'm not sure why/how yet

43081j:
Oh wow, whatever got merged in from master seems to have fixed the tests! 

Any chance I can get a re-review? 🙏

JamesHenry:
@43081j Yeah there was a peer dep issue with the Angular stack that was fixed so I was hoping that would be it. Thanks a lot for your patience in getting this green!

With that said, this is a more significant change than previous ones under this umbrella, so will need some more eyes.

I am also curious about seeing a resolution to this issue I saw when investigating picomatch: https://github.com/micromatch/picomatch/issues/136

43081j:
I agree

Other projects we have migrated to picomatch didn't make use of the extra edge cases minimatch supported, so they were fairly easy to move. We also moved a few to zeptomatch which is even smaller and often faster than picomatch, but possibly missing some rarely used pattern types. 

Anyhow, let me know when we have more reviews, if we want to change something 👍

benmccann:
You might consider `tinyglobby`, which uses `fdir`/`picomatch`, but could be better performing as it does some optimizations to avoid having `fdir` crawl folders that are under the `ignore` pattern. It would also make it so that there are minimal code changes here needing review as `tinyglobby` and `fast-glob` are nearly API compatible. Plus `tinyglobby` is already a dependency via `@nx/js`.

> I am also curious about seeing a resolution to this issue I saw when investigating picomatch: https://github.com/micromatch/picomatch/issues/136

I'm not sure that issue is going to have an effect. `fast-glob` uses `picomatch` and doesn't use `minimatch`, so I don't think anything will change there

43081j:
If you want to open a branch which uses tinyglobby instead, I'd be happy to close this

Especially since we already have it in another package of the monorepo. Probably makes sense to use the same one

PR #28059, opened 2024-09-23
fix(core): add flag when db is disabled for task history

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/utils/task-history.ts
+++ b/packages/nx/src/utils/task-history.ts
@@ -1,6 +1,6 @@
 import { daemonClient } from '../daemon/client/client';
 import { isOnDaemon } from '../daemon/is-on-daemon';
-import { NxTaskHistory, TaskRun, TaskTarget } from '../native';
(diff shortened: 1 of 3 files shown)


packages/nx/src/utils/task-history.ts
```diff
@@ -39,9 +39,13 @@ let taskHistory: TaskHistory;
 
 /**
  * This function returns the singleton instance of TaskHistory
- * @returns singleton instance of TaskHistory
+ * @returns singleton instance of TaskHistory, undefined if database is disabled or WASM is not enabled
  */
-export function getTaskHistory(): TaskHistory {
+export function getTaskHistory(): TaskHistory | undefined {
+  if (process.env.NX_DISABLE_DB === 'true' || !IS_WASM) {
+    return;
```
FrozenPandaz, line 46: Can we return `null`? I think it's weird to return `undefined` :shrug:

FrozenPandaz (approved):
Looks fine otherwise.

PR #28065, opened 2024-09-24
cleanup(nextjs): migrate to node FS

43081j:
Switches from `fs-extra` to using node built in functions instead.

Behaviour should all be the same, though this may bump node requirements to 16.x.

We can probably make this change in a few of the packages if this one ends up passing review. if one of you can let me know what the situation is with node versions we support, that'll be super helpful 🙏

### DIFF
--- a/packages/next/src/executors/build/build.impl.ts
+++ b/packages/next/src/executors/build/build.impl.ts
@@ -71,7 +71,7 @@ export default async function buildExecutor(
   }
 
   if (!directoryExists(options.outputPath)) {
-    mkdirSync(options.outputPath);
+    await mkdir(options.outputPath, { recursive; true });
   }
 
   const builtPackageJson = createPackageJson(
--- a/packages/next/src/executors/build/build.impl.ts
+++ b/packages/next/src/executors/build/build.impl.ts
@@ -8,7 +8,7 @@ import {
 } from '@nx/devkit';
 import { createLockFile, createPackageJson, getLockFileName } from '@nx/js';
 import { join, resolve as pathResolve } from 'path';
-import { copySync, existsSync, mkdir, writeFileSync } from 'fs-extra';
+import { cpSync, existsSync, mkdirSync, writeFileSync } from 'node:fs';
 import { gte } from 'semver';
 import { directoryExists } from '@nx/workspace/src/utilities/fileutils';
 import { checkAndCleanWithSemver } from '@nx/devkit/src/utils/semver';
@@ -71,7 +71,7 @@ export default async function buildExecutor(
   }
 
   if (!directoryExists(options.outputPath)) {
-    mkdir(options.outputPath);
+    mkdirSync(options.outputPath);
   }
 
   const builtPackageJson = createPackageJson(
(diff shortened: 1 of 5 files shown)


JamesHenry (changes requested):
FYI @ziebam has been doing good work migrating the packages from fs-extra to node fs.

The core package has proven to be a bit tricky but we already landed a few of the others.

Please can you add the usual lint rule check to this package to align it with the others.

Otherwise LGTM, but please let's align with @ziebam on this particular dependency because they may have other packages in flight

packages/next/src/executors/build/build.impl.ts
```diff
@@ -71,7 +71,7 @@ export default async function buildExecutor(
   }
 
   if (!directoryExists(options.outputPath)) {
-    mkdir(options.outputPath);
+    mkdirSync(options.outputPath);
```
ziebam, line 74:
```suggestion
    await mkdir(options.outputPath, { recursive; true });
```

This would best mimic `mkdir` from `fs-extra` as it always passes `recursive: true` to the underlying `node:fs/promises` API.

packages/next/src/executors/build/lib/create-next-config-file.ts
```diff
@@ -76,7 +75,14 @@ export function createNextConfigFile(
     projectRoot
   );
   for (const moduleFile of moduleFilesToCopy) {
-    ensureDirSync(dirname(join(context.root, options.outputPath, moduleFile)));
+    const moduleFileDir = dirname(
+      join(context.root, options.outputPath, moduleFile)
+    );
+
+    if (!existsSync(moduleFileDir)) {
+      mkdirSync(moduleFileDir);
```
ziebam, line 83:
```suggestion
      mkdirSync(moduleFileDir, { recursive: true });
```

Same here.

ziebam:
> FYI @ziebam has been doing good work migrating the packages from fs-extra to node fs.
> 
> The core package has proven to be a bit tricky but we already landed a few of the others.
> 
> Please can you add the usual lint rule check to this package to align it with the others.
> 
> Otherwise LGTM, but please let's align with @ziebam on this particular dependency because they may have other packages in flight

Thanks for the ping! I left two comments regarding matching the previous behavior as closely as possible. I currently have no other in flight branches except for the `core` package, so we're good on that front.

packages/next/src/executors/build/build.impl.ts
43081j: Makes sense! 👍

packages/next/src/executors/build/lib/create-next-config-file.ts
```diff
@@ -76,7 +75,14 @@ export function createNextConfigFile(
...
   );
   for (const moduleFile of moduleFilesToCopy) {
-    ensureDirSync(dirname(join(context.root, options.outputPath, moduleFile)));
+    const moduleFileDir = dirname(
+      join(context.root, options.outputPath, moduleFile)
+    );
+
+    if (!existsSync(moduleFileDir)) {
+      mkdirSync(moduleFileDir, { recursive: true });
+    }
```
ziebam, line 84: We don't need the `!existsSync(moduleFileDir)` check here, `mkdirSync` won't overwrite an existing directory.

pralkarz (approved):

JamesHenry:
@43081j looks like a legit failure, the `@nx/workspace` package is no longer used as of this PR and should be removed from the package.json

43081j:
I'm unsure what the current failure is as the logs don't seem to tell me much unless I'm being blind

I've caught up from master again so will see how this CI build goes

JamesHenry (approved):

JamesHenry:
Thank you!

PR #28110, opened 2024-09-25
fix(core): Remove nameAndDirectoryFormat option from generators

ndcunningham:
## Current Behavior

Currently, when we use generator's via the CLI passing the `name` arg has more than one responsibility:
1.  it can be used as a path
2. it can be used as the generator's name (e.g. Component name)

This is not intuitive as if we use via Nx Console because it creates some ambiguity around the `--name` property which is normally described as only 2.
image

## Expected Behavior

What is expected is that the `name` property should only have the responsibility of determining the name of the generated module (Component, Directive etc...). 

The new property `path` replaces the `directory` property but fulfils the same purpose as determining where the module should be created. 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/devkit/src/generators/artifact-name-and-directory-utils.ts
+++ b/packages/devkit/src/generators/artifact-name-and-directory-utils.ts
@@ -17,10 +17,19 @@ export type NameAndDirectoryFormat = 'as-provided';
 
 export type ArtifactGenerationOptions = {
   name: string;
-  directory?: string;
+  directory?: string; // Deprecated in favor of path
+  path?: string;
   fileExtension?: 'js' | 'jsx' | 'ts' | 'tsx' | 'vue';
   fileName?: string;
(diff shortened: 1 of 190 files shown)


packages/devkit/src/generators/artifact-name-and-directory-utils.ts
```diff
@@ -53,7 +60,7 @@ export async function determineArtifactNameAndDirectoryOptions(
 ): Promise<
   NameAndDirectoryOptions & {
     // TODO(leo): remove in a follow up
-    nameAndDirectoryFormat: NameAndDirectoryFormat;
+    nameAndDirectoryFormat: 'as-provided';
```
ndcunningham, line 63: Will be removed after all other generators have been updated.

jaysoo (approved):

jaysoo (approved):

PR #28113, opened 2024-09-25
cleanup(bundling): replace `fs-extra` with `node:fs`

pralkarz:

## Current Behavior
`fs-extra` is used.

## Expected Behavior
`node:fs` is used.

## Related Issue(s)
N/A


JamesHenry (approved):

JamesHenry:
Thank you!

PR #28117, opened 2024-09-25
cleanup(expo): replace `fs-extra` with `node:fs`

pralkarz:

## Current Behavior
`fs-extra` is used.

## Expected Behavior
`node:fs` is used.

## Related Issue(s)
N/A


ziebam:
@JamesHenry I can't seem to be able to ask people for review manually, so I'm pinging you in since you helped land the other similar PRs. 😄

JamesHenry (approved):

PR #28153, opened 2024-09-26
feat(react): make vitest the default test runner since it supports ESM and different moduleResolution better

jaysoo:
This PR updates the default test runner for React/Web apps to be `vitest`. It aligned better with our emphasis on modern tooling, and the lack of ESM and proper TS support (using `module` other than `commonjs`) in Jest makes it hard to use in some workspaces.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/generated/packages/next/generators/library.json
+++ b/docs/generated/packages/next/generators/library.json
@@ -70,8 +70,9 @@
       },
       "unitTestRunner": {
         "type": "string",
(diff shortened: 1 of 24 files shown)


ndcunningham (approved):

xiongemi (approved):

Rucellmai61 (changes requested):
El papichulo

PR #28156, opened 2024-09-27
fix(core): fix env override run command

### LINKED ISSUE #27821: .env overrides already-set environment variables when nx:run-command executor function is called from another executor
### Current Behavior

When the ["nx:run-command" executor function](https://github.com/nrwl/nx/blob/0b99d1deff5148075bac31cbd8def7f9d12a17c5/packages/nx/src/executors/run-commands/run-commands.impl.ts#L113-L151) is called from a custom executor, environment variables already set in the executor environment are overridden by environment variables in `.env` in the command's environment.

### Expected Behavior

The following doc indicates that `.env` files do not override existing environment variables

https://github.com/nrwl/nx/blob/0b99d1deff5148075bac31cbd8def7f9d12a17c5/docs/shared/guides/define-environment-variables.md#L35-L54

Also, existing environment variables are not overridden by `.env` if the "nx:run-command" executor is used directly by a target.

### GitHub Repo

https://github.com/ethantkoenig/nx-env-var-bug

### Steps to Reproduce

Run `ENV_VAR=already-set nx run foo:echo-env-var`, and note that the output is `from-file` (set in [`.env`](https://github.com/ethantkoenig/nx-env-var-bug/blob/main/.env)) rather than `already-set`.

### Nx Report

```shell

Node           : 21.7.1
OS             : darwin-arm64
npm            : 10.5.0

nx (global)        : 19.6.5
nx                 : 19.6.5
@nx/js             : 19.6.5
@nx/linter         : 19.6.5
@nx/eslint         : 19.6.5
@nx/workspace      : 19.6.5
@nx/devkit         : 19.6.5
@nx/eslint-plugin  : 19.6.5
@nrwl/tao          : 19.6.5
typescript         : 5.5.4
Registered Plugins:
@nx/eslint/plugin
Community plugins:
@nxlv/python : 19.0.1
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Note that the `foo:echo-env-var` target uses the `@nxlv/python:run-commands` executor, which is [a very thin wrapper around `nx:run-command`](https://github.com/lucasvieirasilva/nx-plugins/blob/07ec9170cfa4b32af8a6a3f67b1a3e154db623f0/packages/nx-python/src/executors/run-commands/executor.ts#L7-L13).

I believe this is happening because the [`options.env`](https://github.com/nrwl/nx/blob/0b99d1deff5148075bac31cbd8def7f9d12a17c5/packages/nx/src/executors/run-commands/schema.json#L145-L151) property is unset when the `@nxlv/python:run-commands` executor function is called, and therefore is unset in the subsequent call of the `nx:run-command` executor function. In contrast, when the `nx:run-command` executor is used directly by a target, Nx appears to prepopulate `options.env` with the process's current environment variables when calling the executor function.

`options.env` is eventually passed as the `env` argument in `processEnv` (shown below). Immediately after line 493, `res` will contain both process environment variables and variables from `.env`, with the latter taking precedence. If `env` contains the process environment variables (as is the case when `nx:run-command` is used directly by a target), lines 495-498 will counter-override the environment variables from `.env` with the process environment variables. But this does not happen if `env` is empty.

https://github.com/nrwl/nx/blob/0b99d1deff5148075bac31cbd8def7f9d12a17c5/packages/nx/src/executors/run-commands/run-commands.impl.ts#L480-L508

I think there are potentially two scenarios in which this behavior is WAI:

1. The `nx:run-command` executor function is not intended to be called from other executors in the first place, though https://nx.dev/extending-nx/recipes/compose-executors#compose-executors seems to indicate that executor functions are in general intended to be called from other executors.
2. I suppose one could stipulate that it is the responsibility of callers of the `nx:run-command` executor function to set the `env` option to include the currently-set  environment variables, but that would seem to violate the principle of least astonishment, and I would argue should at the very least be clearly documented.

xiongemi:
## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/27821

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -21,19 +21,9 @@ const childProcesses = new Set<ChildProcess | PseudoTtyProcess>();
 
 function loadEnvVarsFile(path: string, env: Record<string, string> = {}) {
   unloadDotEnvFile(path, env);
-  const result = loadAndExpandDotEnvFile(path, env, true);
+  const result = loadAndExpandDotEnvFile(path, env);
   if (result.error) {
     throw result.error;
-  }
-}
-
-function loadEnvVars(path?: string, env: Record<string, string> = {}) {
-  if (path) {
-    loadEnvVarsFile(path, env);
-  } else {
-    try {
-      loadEnvVarsFile('.env', env);
-    } catch {}
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -478,25 +468,32 @@ function calculateCwd(
...
     ...process.env,
     ...localEnv,
   };
-  // env file from envFile option takes priority over process env
-  if (process.env.NX_LOAD_DOT_ENV_FILES !== 'false') {
-    loadEnvVars(envFile, res);
+
+  let envFromEnvFile: Record<string, string> = {};
+  if (process.env.NX_LOAD_DOT_ENV_FILES !== 'false' && envFile) {
+    loadEnvVarsFile(envFile, envFromEnvFile);
```
FrozenPandaz, line 491: This should continue to have `localEnv` passed into here.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #28167, opened 2024-09-27
feat(rspack): add convert-webpack generator

Coly010:
## Current Behavior

We do not have an automated method for people to switch to rspack from webpack applications

## Expected Behavior

We should have a generator that will convert webpack application projects to use rspack

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/rspack/src/generators/convert-webpack/convert-webpack.ts
+++ b/packages/rspack/src/generators/convert-webpack/convert-webpack.ts
@@ -1,0 +1,123 @@
+import {
+  addDependenciesToPackageJson,
+  formatFiles,
+  getProjects,
+  type Tree,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import { Schema } from './schema';
+import {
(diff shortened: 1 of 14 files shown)


isaacplmann (approved):

packages/rspack/src/generators/convert-webpack/convert-webpack.ts
```diff
@@ -0,0 +1,137 @@
...
+    );
+  }
+  const project = projects.get(options.project);
+
+  const webpackConfigsToConvert: [string, string][] = [];
+
+  for (const [targetName, target] of Object.entries(project.targets)) {
+    if (
+      target.executor === '@nx/webpack:webpack' ||
+      target.executor === '@nrwl/webpack:webpack'
```
jaysoo, line 30: We can remove `@nrwl/` checks since we stopped publishing it in v20.

jaysoo (approved):

PR #28168, opened 2024-09-27
fix(release): Support path offset from git root to Nx workspace

### LINKED ISSUE #27995: Nx release fails to detect changes if the workspace directory is not the monorepo root directory
### Current Behavior

Given a monorepo with a set of packages, each with its own `project.json` file.

The `nx.json` file resides in a repo subdirectory called `typescript`, along with `package.json`

The path to each package is `typescript/packages/<project_name>`

`nx.json` is set up to use Conventional commits and independent packages:

```
    "release": {
        "version": { "conventionalCommits": true },
        "projectsRelationship": "independent"
    }
```

I have created a commit that changes some file belonging to a package, with a "fix:" prefix in the commit header.

When I run `npx nx release version --dry-run` it fails to detect any changes, ie. `🚫 No changes were detected using git history and the conventional commits standard.`

### Expected Behavior

I expect NX release to be able to detect changes if the NX workspace directory is not the root directory of the monorepo

### GitHub Repo

https://github.com/hkfb/webviz-subsurface-components/tree/nx_release

### Steps to Reproduce

1. Clone the repo and check out the `nx_release` tag
2. In the `typescript` directory run `npx nx release version --dry-run  -p subsurface-viewer`

### Nx Report

```shell
Node   : 20.10.0
OS     : linux-x64
npm    : 8.19.4

nx             : 19.2.3
@nx/js         : 19.2.3
@nx/jest       : 19.2.3
@nx/workspace  : 19.2.3
@nx/devkit     : 19.2.3
@nrwl/tao      : 19.2.3
typescript     : 5.4.3
Community plugins:
@theunderscorer/nx-semantic-release : 2.11.0
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

gionkunz:
@JamesHenry @AgentEnder 

## Current Behavior

Currently Nx Release does not support the use-case where the Nx Workspace is within a nested folder under the Git root directory. As it's quite common to have a backend,infra,frontend separation on the repo top directory and then only initialize the Nx Workspace in the frontend directory, this is probably needed in many projects.

Also, see issue #27995

## Expected Behavior

The version and changelog commands both should support workspaces that are nested within sub-folders of a Git repository.

## Related Issue(s)

Fixes #27995


gionkunz:
Hi @xiongemi @leosvelperez. Do you need anything else for the review of this PR? Please let me know if I need to add anything. We are currently relying on a custom monkey patch using patch-package to make it work in our environment.

m3Lith:
This is a bit of a blocker for us too, would love to see it released..

JamesHenry (changes requested):
Thanks a lot for this @gionkunz and I'm sorry for the delay in getting back to you. I think we can avoid a lot of the changes here by acknowledging that the git absolute root and its offset are basically as constant as the workspaceRoot itself.

Therefore, in `git.ts` I would simply do that following:

```ts
// ...
// ... existing parseGitCommit() function here (with the one tweak to leverage replaceGitRootPathDiff() in the reduce as you have already)
// ...

/**
 * Lazily compute the offset from the workspace root to the git root, if any, one time,
 * it is not something that changes during the execution of a command.
 */
let workspaceGitRootOffset: string;

function replaceGitRootPathDiff(file: string) {
  if (!workspaceGitRootOffset) {
    let absoluteGitRoot: string;
    // This check is probably debatable, we could probably just always run the getAbsoluteGitRoot tbf...
    if (existsSync(join(workspaceRoot, '.git'))) {
      absoluteGitRoot = workspaceRoot;
    } else {
      absoluteGitRoot = getAbsoluteGitRoot();
    }
    workspaceGitRootOffset = getWorkspaceGitRootOffset(
      absoluteGitRoot,
      workspaceRoot
    );
  }

  if (
    !workspaceGitRootOffset ||
    !file.startsWith(workspaceGitRootOffset)
  ) {
    return file;
  }

  return file.replace(workspaceGitRootOffset, '');
}

// (no need to export this util any more)
function getAbsoluteGitRoot() {
  try {
    return execSync('git', ['rev-parse', '--show-toplevel']).trim();
  } catch (e) {
    throw new Error(`Could not determine git root`);
  }
}
```

With your `getWorkspaceGitRootOffset()` function and unit tests remaining the same.

I think that gets us everything we need without changing any other files or function signatures right?

JamesHenry:
Hi again @gionkunz I am going to close this one in favour of https://github.com/nrwl/nx/pull/28650 which solves the problem via a more direct mechanism - namely, the `--relative` flag on the git command.

gionkunz:
@JamesHenry Even better! Thanks!

PR #28186, opened 2024-09-28
cleanup(misc): migrate rollup to picocolors

43081j:

there wasnt a scope for `rollup` so I used `core` here


43081j:
I'm a little stuck with the failures

a re-run produced a different failure, but usually along the lines of an import not existing

locally, this builds just fine and I cant see anything obvious in the diff that'd cause this

JamesHenry (approved):

PR #28202, opened 2024-09-30
fix(core): fix package manager determination logic

### LINKED ISSUE #28201: /bin/sh: 1: bun: not found
### Current Behavior

I'm just trying to create a new workspace using `npx create-nx-workspace` command. I have the following error:

```
Error: Command failed: bun --version
/bin/sh: 1: bun: not found

    at checkExecSyncError (node:child_process:890:11)
    at execSync (node:child_process:962:15)
    at getPackageManagerVersion (/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/src/utils/package-manager.js:101:50)
    at getPackageManagerCommand (/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/src/utils/package-manager.js:39:23)
    at createSandbox (/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/src/create-sandbox.js:20:84)
    at createWorkspace (/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/src/create-workspace.js:19:61)
    at main (/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/bin/create-nx-workspace.js:118:72)
    at Object.handler (/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/bin/create-nx-workspace.js:103:11)
    at /mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/yargs/build/index.cjs:1:8993 {
  status: 127,
  signal: null,
  output: [ null, '', '/bin/sh: 1: bun: not found\n' ],
  pid: 188003,
  stdout: '',
  stderr: '/bin/sh: 1: bun: not found\n'
}
```

### Expected Behavior

It just should work. I don't have bun anywhere.  I debugged the code and found following code:
```ts
/**
 * Detects which package manager was used to invoke create-nx-{plugin|workspace} command
 * based on the main Module process that invokes the command
 * - npx returns 'npm'
 * - pnpx returns 'pnpm'
 * - yarn create returns 'yarn'
 *
 * Default to 'npm'
 */
function detectInvokedPackageManager() {
    let detectedPackageManager = 'npm';
    // mainModule is deprecated since Node 14, fallback for older versions
    const invoker = require.main || process['mainModule'];
    // default to `npm`
    if (!invoker) {
        return detectedPackageManager;
    }
    for (const pkgManager of exports.packageManagerList) {
        if (invoker.path.includes(pkgManager)) {
            detectedPackageManager = pkgManager;
            break;
        }
    }
    return detectedPackageManager;
}

```

So the point is - my cwd is:  `/mnt/wsl/workspace/ubuntu/nx-cap-gen/node_modules/create-nx-workspace/bin` and it does include bun because of uBUNtu.. But it doesn't mean at all it has to use bun manager.

This code needs to be reworked, sorry!

### GitHub Repo

_No response_

### Steps to Reproduce

1.
Just create a folder containing bun in the path like i use:
'/mnt/wsl/workspace/ubuntu/nx-cap-gen'

2. Create a new workspace inside using command:
`npx create-nx-workspace test `

Make sure you have no bun available otherwise it will execute using bun. 

### Nx Report

```shell
Node           : 18.20.4
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.7.0

nx                 : 19.8.2
@nx/js             : 19.8.2
@nx/jest           : 19.8.2
@nx/linter         : 19.8.2
@nx/eslint         : 19.8.2
@nx/workspace      : 19.8.2
@nx/devkit         : 19.8.2
@nx/eslint-plugin  : 19.8.2
@nx/plugin         : 19.8.2
@nrwl/tao          : 19.8.2
typescript         : 5.5.4
Local workspace plugins:
         nx-cap
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

ThePlenkov:
## Current Behavior
Currently create-nx-workspace determines package manager based on the invoker's path and it might be totally misleading if some names are used in the path such as /ubuntu/my_project which is identified as bun command by mistake

## Expected Behavior
It should correctly identify package manager

## Related Issue(s)

Fixes #28201

### DIFF
--- a/nx-dev/ui-common/src/lib/champion-perks.tsx
+++ b/nx-dev/ui-common/src/lib/champion-perks.tsx
@@ -27,7 +27,7 @@ import { SectionHeading } from './typography';
 export function ChampionPerks(): JSX.Element {
   return (
     <article
-      id="nx-is-fast"
+      id="making-of-champion"
       className="relative bg-slate-50 py-28 dark:bg-slate-800/40"
     >
       <div className="mx-auto max-w-7xl px-4 sm:grid sm:grid-cols-2 sm:gap-8 sm:px-6 lg:px-8">
--- a/nx-dev/ui-common/src/lib/headers/menu-items.ts
+++ b/nx-dev/ui-common/src/lib/headers/menu-items.ts
@@ -149,40 +149,6 @@ export const solutions: MenuItem[] = [
     isHighlight: false,
   },
 ];
-const useCaseItems: MenuItem[] = [
-  {
-    name: 'Get actionable feedback',
-    description: 'Enhanced analysis & analytics of your workflows.',
-    href: '',
-    icon: null,
-    isNew: false,
-    isHighlight: false,
-  },
-  {
-    name: 'Reduce CI timings with remote caching',
-    description: 'Share task results & artifacts between CI & teams.',
-    href: '',
-    icon: null,
-    isNew: false,
-    isHighlight: false,
-  },
-  {
-    name: 'Performant task distribution at scale',
-    description: 'Faster & cheaper CI workflows.',
-    href: '',
-    icon: null,
-    isNew: false,
-    isHighlight: false,
-  },
-  {
-    name: 'Improve E2E time execution on CI',
-    description: 'Automatic task splitting.',
-    href: '',
-    icon: null,
-    isNew: false,
-    isHighlight: false,
-  },
-];
 export const learnItems: MenuItem[] = [
   {
     name: 'Step by step tutorials',
--- a/nx-dev/ui-enterprise/src/lib/security.tsx
+++ b/nx-dev/ui-enterprise/src/lib/security.tsx
@@ -10,15 +10,15 @@ export function Security(): JSX.Element {
               Security
             </SectionHeading>
             <p className="mt-6 text-lg">
-              Nx Cloud is certified to the industry standards, is constantly
-              monitored and{' '}
+              Nx Cloud is certified to industry standards, is constantly
+              monitored, and{' '}
               <a
                 href="https://security.nx.app/"
                 target="_blank"
                 rel="nofollow noreferrer"
                 className="underline"
               >
-                issue security Trust Reports powered by Vanta
+                issues security trust reports powered by Vanta
               </a>
               .
             </p>


JamesHenry (changes requested):
I don't think this is working as intended based on the e2e results.

Based on my own experimentation I believe what we want is to replace the existing logic (and your change) with this:

```ts
  if (process.env.npm_config_user_agent) {
    for (const pm of packageManagerList) {
      if (process.env.npm_config_user_agent.startsWith(pm)) {
        return pm;
      }
    }
  }

  if (process.env.npm_execpath) {
    for (const pm of packageManagerList) {
      if (process.env.npm_execpath.includes(pm)) {
        return pm;
      }
    }
  }
  return 'npm';
```

(Obviously make sure the formatting is fixed before pushing)

Results on my machine:

image

ThePlenkov:
@JamesHenry indeed - I didn't have intention to say it's a final code. So far I've been trying still to understand what is the functional meaning of this function. Like what is more important - to understand what is running the code or which npm was used during installation because is not the same.

Please note that if the code is run via node or deno directly - npm_config_user_agent is empty.
I've created some command today to prove that: https://www.npmjs.com/package/which-engine which also supports which-pm.

ThePlenkov:
@JamesHenry is there any way to automate formatting before pushing? like husky + format:write may be?

ThePlenkov:
OK I see now - seems like we are supposed to use pm used to trigger the call. In case if there is no package manager ( let's say we call globally installed create-nx-workspace or via node programmatically ) - then we just use npm as fallback.

@JamesHenry 
How about this?

I intentionally use / in the startsWith to make sure it's yarn and node something like yarn_alternative.

In the includes script I break path into segments and we search for a segment = npm | yarn | bun. And segments like ubuntu will not be taken into account

```
export function detectInvokedPackageManager(): PackageManager {

  if (env.npm_config_user_agent) {
    for (const pm of packageManagerList) {
      if (env.npm_config_user_agent.startsWith(`${pm}/`)) {
        return pm;
      }
    }
  }

  if (env.npm_execpath) {
    for (const pm of packageManagerList) {
      if (env.npm_execpath.split(sep).includes(pm)) {
        return pm;
      }
    }
  }
  return 'npm';

}
```

JamesHenry:
@ThePlenkov thanks looks reasonable

ThePlenkov:
I'm struggling to understand why test is failing. It is failing even locally when I run it on the upstream branch ( without my actual changes ). Can it be because of remote cache?

JamesHenry:
@ThePlenkov I have brought things up to date with the latest, are you able to reproduce locally?

ThePlenkov:
I did, but couldn't reproduce failing tests((

JamesHenry:
@ThePlenkov I think I have traced the issue, however I cannot push directly to the PR because you opened it from your `master` branch. Please open PRs from feature branches in future to allow collaborators to push directly.

Please can you try removing this hardcoded npm reference in the e2e utils, I am not sure why it was done but I believe it could be the cause of the error because based on the logs, the test workspace is getting created with pnpm but then is attempting an `npm install` later:

image

cc @FrozenPandaz

ThePlenkov:
@JamesHenry I'll create another PR better then

PR #28209, opened 2024-10-01
fix(expo): pnpm+workspace build

### LINKED ISSUE #28208: Expo: copyPackageJsonAndLock returns undefined instead of cleanup function in corner cases
### Current Behavior

copyPackageJsonAndLock returns undefined https://github.com/nrwl/nx/blob/8fa7065cf14df6a90896442f90659b00baa1b5b9/packages/expo/src/executors/build/build.impl.ts#L126 when it's ran with pnpm + workspace

### Expected Behavior

it should always return a cleanup function, according to its "docs" (code comment) 

if there's nothing to cleanup, then a noop function should be returned

this is the behaviour that another part of your code https://github.com/nrwl/nx/blob/8fa7065cf14df6a90896442f90659b00baa1b5b9/packages/expo/src/executors/build/build.impl.ts#L43 expects, and fails with a non-descriptive error "resetLocalFunction is not a function"

### GitHub Repo

_No response_

### Steps to Reproduce

run a build e.g. `npx nx build-eas` with pnpm + workspaces set up

### Nx Report

```shell
no need
```

### Failure Logs

```shell
no need
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

dearlordylord:
## Current Behavior

build executors for expo crash when package manager pnpm and workspaces are used

## Expected Behavior

no crashies

## Related Issue(s)

https://github.com/nrwl/nx/issues/28208

Fixes #28208

## FAQ

- why not handle the `undefined` by the calling code instead

The less the calling code knows about inner workings, the better: in this case, doing this would add the implicit dependency on "pnpm and workspaces" to the calling code

- but it's an empty function, my performance

We call it once per user interaction.

- but still, it's an empty function

We return an `empty array` when there are no elements; we don't return `undefined` when the array is empty and `array` when array is non-empty. Hopefully so.

- but,

The doc also says that this function returns a function.

### DIFF
--- a/packages/expo/src/executors/build/build.impl.ts
+++ b/packages/expo/src/executors/build/build.impl.ts
@@ -114,7 +114,7 @@ function copyPackageJsonAndLock(
   packageManager: PackageManager,
   workspaceRoot: string,
   projectRoot: string
-) {
+): () => void {
   const packageJson = pathResolve(workspaceRoot, 'package.json');
   const rootPackageJson = readJsonFile<PackageJson>(packageJson);
   // do not copy package.json and lock file if workspaces are enabled
@@ -123,7 +123,8 @@ function copyPackageJsonAndLock(
       existsSync(pathResolve(workspaceRoot, 'pnpm-workspace.yaml'))) ||
     rootPackageJson.workspaces
   ) {
-    return;
+    // no resource taken, no resource cleaned up
+    return () => {};
   }
 
   const packageJsonProject = pathResolve(projectRoot, 'package.json');


packages/expo/src/executors/build/build.impl.ts
```diff
@@ -114,7 +114,7 @@ function copyPackageJsonAndLock(
   packageManager: PackageManager,
   workspaceRoot: string,
   projectRoot: string
-) {
+): () => void {
```
dearlordylord, line 117: to make not possible to forget again

Balance8 (approved):

xiongemi (approved):

PR #28222, opened 2024-10-01
feat(misc)!: handle v20 deprecations in plugins

jaysoo:
This PR removes these from v20 since they were deprecated and slated for removal:

- `executeWebpackDevServerBuilder` export from `@nx/angular/executors`, users should use `executeDevServerBuilder`
- `withStylus` util from `@nx/next/plugins/with-stylus` since it was deprecated in v17 and has just throw an error that users need to use SASS with Next.js

The `getRollupOptions` function from `@nx/react/plugins/bundle-rollup` has been deprecated as mention previously and slated for removal in v22. New users are using inferred targets from Rollup, and existing projects using this module should run `nx g @nx/rollup:convert-to-inferred` or manually update rollup config to use `withNx` function.

Also, bumped some deprecation for later in v21:

- Remove inline builds from tsc/swc 
- Changes to SVGR to align with Webpack v5 (e.g. `import ReactComponent from './img.svg?svgr'`)
- Remove `isolatedConfig` from Webpack executor -- requires a migration that extracts to a standard webpack config just in case (different from the original one that extracts to `withNx`)

The ESLint TODOs were rescoped to `TODO(eslint)` and we'll look at it in further flat config work rather than tying it to an Nx release.

## Related Issue(s)

Fixes #


ndcunningham (approved):

packages/react/plugins/bundle-rollup.ts
```diff
@@ -1,6 +1,9 @@
 import * as rollup from 'rollup';
 
-// TODO(v20): This should be deprecated and removed in v22.
+// TODO(v22): Remove this in Nx 22 and migrate to explicit rollup.config.js files.
```
FrozenPandaz, line 3: Why v22? Not v21?

FrozenPandaz (approved):

PR #28224, opened 2024-10-01
feat(release)!: releasePublish always returns status code per project

JamesHenry:
BREAKING CHANGE

In v19 of nx, the programmatic usage of `await releasePublish({ ... })` would throw in the case that _any_ project failed to publish. It would return the number `0` (to represent the overall exit code) in the case that all projects were successfully published.

Now, in v20, the programmatic usage of `await releasePublish({ ... })` will always return a `PublishProjectsResult` object, which is a mapping of project names to an object of `{ code: number }` for each project. This allows consumers to make informed decisions about which projects failed to publish and decide how to handle it in their release scripts.

For example, where there are projects called "pkg-a" and "pkg-b" and "pkg-a" is successfully published and "pkg-b" fails:

```ts
const result = await releasePublish({
  dryRun: false
});
// result is set to { "pkg-a": { code: 0 }, "pkg-b": { code: 1 } )
```

### DIFF
--- a/packages/nx/src/command-line/release/publish.ts
+++ b/packages/nx/src/command-line/release/publish.ts
@@ -9,13 +9,13 @@ import {
 } from '../../config/project-graph';
 import { createProjectFileMapUsingProjectGraph } from '../../project-graph/file-map-utils';
 import { createProjectGraphAsync } from '../../project-graph/project-graph';
-import { runCommand } from '../../tasks-runner/run-command';
+import { runCommandForTasks } from '../../tasks-runner/run-command';
 import {
   createOverrides,
   readGraphFileFromGraphArg,
 } from '../../utils/command-line-utils';
-import { output } from '../../utils/output';
 import { handleErrors } from '../../utils/handle-errors';
+import { output } from '../../utils/output';
 import { projectHasTarget } from '../../utils/project-graph-utils';
 import { generateGraph } from '../graph/graph';
 import { PublishOptions } from './command-object';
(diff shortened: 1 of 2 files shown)


packages/nx/src/command-line/release/publish.ts
```diff
@@ -98,53 +113,53 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
...
         projectGraph,
         nxJson,
         releaseGroup.projects,
-        isCLI,
         {
           excludeTaskDependencies: shouldExcludeTaskDependencies,
           loadDotEnvFiles: process.env.NX_LOAD_DOT_ENV_FILES !== 'false',
         }
       );
-      if (status !== 0) {
```
JamesHenry, line 142: This never actually did anything after the throw was added at some point

packages/nx/src/command-line/release/publish.ts
```diff
@@ -98,53 +113,53 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
...
           projectGraph,
           nxJson,
           Array.from(releaseGroupToFilteredProjects.get(releaseGroup)),
-          isCLI,
           {
             excludeTaskDependencies: shouldExcludeTaskDependencies,
             loadDotEnvFiles: process.env.NX_LOAD_DOT_ENV_FILES !== 'false',
           }
         );
-        if (status !== 0) {
```
JamesHenry, line 119: This never actually did anything after the throw was added at some point

jaysoo (approved):

PR #28227, opened 2024-10-01
feat(core): allow circular project dependencies to execute tasks

Cammisuli:
## Current Behavior

If there are project dependencies and not all projects contain the same task target, a circular dependency error is shown. 

## Expected Behavior

If not all circular dependent projects contain the same task target, allow execution of the target.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/create-task-graph.ts
+++ b/packages/nx/src/tasks-runner/create-task-graph.ts
@@ -9,6 +9,8 @@ import { TargetDefaults, TargetDependencies } from '../config/nx-json';
 import { TargetDependencyConfig } from '../devkit-exports';
 import { output } from '../utils/output';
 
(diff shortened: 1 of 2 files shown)


FrozenPandaz (approved):

packages/nx/src/tasks-runner/create-task-graph.ts
```diff
@@ -81,6 +83,8 @@ export class ProcessTasks {
       }
     }
 
+    this.filterNoopTasks();
```
FrozenPandaz, line 86: Can we refer to these as dummy tasks?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #28231, opened 2024-10-01
feat(release)!: version.generatorOptions.updateDependents is "auto" by default

JamesHenry:
BREAKING CHANGE

In Nx v19, `release.version.generatorOptions.updateDependents` is "never" by default, meaning dependents are not automatically patched when their dependencies change (applicable to independent projects).

In Nx v20, `release.version.generatorOptions.updateDependents` is "auto" by default, meaning dependents are automatically patched when their dependencies change (applicable to independent projects).

### DIFF
--- a/packages/js/src/generators/release-version/release-version.spec.ts
+++ b/packages/js/src/generators/release-version/release-version.spec.ts
@@ -454,7 +454,7 @@ To fix this you will either need to add a package.json file at that location, or
       });
 
       describe('updateDependentsOptions', () => {
-        it(`should not update dependents when filtering to a subset of projects by default`, async () => {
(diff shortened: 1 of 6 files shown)


packages/js/src/generators/release-version/release-version.spec.ts
```diff
@@ -494,6 +494,7 @@ To fix this you will either need to add a package.json file at that location, or
             currentVersionResolver: 'disk',
             specifierSource: 'prompt',
             releaseGroup: createReleaseGroup('independent'),
+            updateDependents: 'auto',
```
jaysoo, line 497: Remove this

jaysoo (approved):

PR #28240, opened 2024-10-02
feat(core): support rollup.config.ts

ThePlenkov:
I would like to be able to use rollup.config.ts. Same way it's used by rollup and vite teams too.

## Current Behavior

Currently nx plugin does not reconize project containing rollup.config.ts. It expects only {c|m}js files

## Expected Behavior
When project has rollup.config.ts rollup plugin infers build target automatically. This build target should be able to find the rollup.config.ts. For that purpose a special [configPlugin ](https://rollupjs.org/command-line-interface/#configplugin-plugin)CLI parameter should be used 

## Related Issue(s)

Feature request [#28226](https://github.com/nrwl/nx/discussions/28226)

### DIFF
--- a/packages/rollup/.eslintrc.json
+++ b/packages/rollup/.eslintrc.json
@@ -41,7 +41,8 @@
               "nx",
               "typescript",
               "@swc/core", // Installed to workspace and only used in swc() plugin
-              "postcss"
+              "postcss",
+              "@rollup/plugin-typescript"
             ]
           }
         ]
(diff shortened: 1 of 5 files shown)


ThePlenkov:
I had to revert forced `--configPlugin @rollup/plugin-typescript` because it breaks tests for rollup.config.js. Instead I made it optional. I analyze detected config path extension. If it's ts - then we load typescript config plugin, otherwise we just load the config as is.

ThePlenkov:
it says pipeline is cancelled. I do not see any error logs..

Coly010 (approved):

leosvelperez (approved):

PR #28243, opened 2024-10-02
feat(devkit): bump compatibility to Nx 19 - 21.x

FrozenPandaz:
BREAKING CHANGE

## Current Behavior

* `@nx/devkit` supports Nx 17 - 20.
* Node 18 - 22 is supported 
* `ExecutorContext.projectGraph`, `ExecutorContext.nxJsonConfiguration`, and `ExecutorContext.projectsConfigurations` is marked as optional because `ExecutorContext` in some versions of Nx did not have them.
* `ExecutorContext.workspace` is marked as optional because `ExecutorContext` in some versions of Nx did not have the above properties which contain the same information.
* `ProjectGraphNode` is deprecated.
* `NxPluginV1.processProjectGraph` was deprecated long ago and there has been a warning since.
* `appRootPath` has been deprecated for a long time.
* `parseTargetString` had a variant that did not take either the project graph or the executor context.
* `readNxJson` has a variant which does not take a tree. This was not clearly deprecated.
* There are handlers to require from `@nx/` instead of `@nrwl`
* Nx tries to get a root install from `@nrwl/cli`

## Expected Behavior

* `@nx/devkit` supports Nx 19 - 21.
* Node 20 - 22 is supported 
* `ExecutorContext.projectGraph`, `ExecutorContext.nxJsonConfiguration`, and `ExecutorContext.projectsConfigurations` is marked as required because `ExecutorContext` in Nx 19+ is guaranteed to have them.
* `ExecutorContext.workspace` is removed because the same information is available in the above properties
* `ProjectGraphNode` is removed.
* `NxPluginV1` is no more. All plugins should be `NxPluginV2`.
* `workspaceRoot` is the replacement for `appRootPath`. `appRootPath` is removed.
* `parseTargetString` no longer has a variant that did not take either the project graph or the executor context.
* `readNxJson` still has a variant which does not take a tree but it's clearly deprecated to be removed in Nx 21.
* `@nrwl` packages are no more so we don't have to redirect requires anymore.
* `@nrwl/cli` is no more so Nx shouldn't try to get a root install there
* 

## Related Issue(s)

Fixes #


packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -31,21 +31,11 @@ type NpmResolutionCache = Map<string, string | null>;
...
-  ...builtinModules.map((x) => `node:${x}`),
-  // These are missing in the builtinModules list
-  // See: https://github.com/nodejs/node/issues/42785
-  // TODO(v20): We should be safe to use `isBuiltin` function instead of keep the set here (https://nodejs.org/api/module.html#moduleisbuiltinmodulename)
-  'test',
-  'node:test',
-  'node:sea',
-  'node:sqlite',
-]);
+const experimentalNodeModules = new Set(['node:sqlite']);
```
jaysoo, line 34: Is this necessary? If you run with `node --experimental-sqlite` then `isBuiltin` from `node:module` already includes `node:sqlite`.

packages/nx/src/project-graph/plugins/utils.ts
```diff
@@ -1,62 +1,9 @@
-import { dirname } from 'node:path';
-
-import { toProjectName } from '../../config/to-project-name';
-import { combineGlobPatterns } from '../../utils/globs';
-
-import type { NxPluginV1 } from '../../utils/nx-plugin.deprecated';
-import type { LoadedNxPlugin, NormalizedPlugin } from './internal-api';
 import {
```
jaysoo, line 1: Are there changes to the V2 public APIs? I see that `CreateNodesV2` is slated for removal in v21, and users should be using `CreateNodes` instead. Is that part of v20 changes?

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
FrozenPandaz: Sometimes Nx Console starts the daemon and probably won't run with `--experimental-sqlite` so this ensure that it's handled regardless of it's passed or not.

jaysoo (approved):

Cammisuli (approved):

PR #28247, opened 2024-10-02
fix(core): allow prompts during nx add

jaysoo:
This PR allows `nx add` to initialize plugins that have prompts.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/add/add.ts
+++ b/packages/nx/src/command-line/add/add.ts
@@ -4,7 +4,7 @@ import * as ora from 'ora';
 import { isAngularPluginInstalled } from '../../adapter/angular-json';
 import type { GeneratorsJsonEntry } from '../../config/misc-interfaces';
 import { readNxJson, type NxJsonConfiguration } from '../../config/nx-json';
-import { runNxAsync } from '../../utils/child-process';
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/add/add.ts
```diff
@@ -135,8 +135,12 @@ async function initializePlugin(
       args.push(...options.__overrides_unparsed__);
     }
 
-    await runNxAsync(`g ${pkgName}:${initGenerator} ${args.join(' ')}`, {
-      silent: !options.verbose,
```
FrozenPandaz, line 139: Why this switch? `runNxAsync` handles when the `.nx` Nx installation method is used (non-JS workspaces.

FrozenPandaz (changes requested):

packages/nx/src/command-line/add/add.ts
jaysoo: It wasn't easy to get prompts working using `runNxAsync`. I'll need to look at it closer.

jaysoo: Could switch to `runNxSync` and pass the `stdio` option.

FrozenPandaz (approved):

PR #28259, opened 2024-10-02
chore(nx-dev): Update LTS version for Nx 20

ndcunningham:


gperdomor:
Why not jump to 22 directly which will be LTS this month?

jaysoo (approved):

PR #28265, opened 2024-10-03
docs(linter): mention that eslint.config.cjs is supported

ThePlenkov:
## Current Behavior
Currently it is already possible to use `eslint.config.cjs` because of this [line](https://github.com/nrwl/nx/blob/c655b6cf4f8c65a03edde1ff3c79a7ba093b5c48/packages/eslint/src/utils/flat-config.ts#L7):

However docs section doesn't mention it

## Expected Behavior
This commit is supposed to mention this feature in the plugin documentation


leosvelperez (approved):
Thank you!

PR #28277, opened 2024-10-03
docs(js): add a note to JS plugin overview that --preset=ts changed in Nx 20, and that --preset=app can still be used

jaysoo:
Since `--preset=ts` has changed, we want to provide users with a way to generate the older, integrated setup if they need to.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/packages/js/js-plugin.md
+++ b/docs/shared/packages/js/js-plugin.md
@@ -57,6 +57,12 @@ yarn create nx-workspace my-org --preset=ts
 {% /tab %}
 {% /tabs %}
 
+{% callout type="note" title="Modernized monorepo setup" %}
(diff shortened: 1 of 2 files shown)


ndcunningham (approved):

docs/shared/packages/js/js-plugin.md
```diff
@@ -57,6 +57,12 @@ yarn create nx-workspace my-org --preset=ts
 {% /tab %}
 {% /tabs %}
 
+{% callout type="note" title="Modernized monorepo setup" %}
+Nx 20 updates the TS monorepo setup when using `--preset=ts`. The workspace is set up with [TypeScript Project References] along with the Workspaces feature from [npm](https://docs.npmjs.com/cli/using-npm/workspaces), [yarn](https://yarnpkg.com/features/workspaces), [pnpm](https://pnpm.io/workspaces), and [bun](https://bun.sh/docs/install/workspaces).
```
isaacplmann, line 61: `TypeScript Project References` doesn't link anywhere

PR #28281, opened 2024-10-03
docs(core): ts project references guide

isaacplmann:
Adds a TS project references guide


docs/shared/packages/js/typescript-project-references.md
```diff
@@ -0,0 +1,141 @@
+# Link Projects with TypeScript Project References
+
+In Nx 20, the `@nx/js` plugin provides the ability to link projects in a monorepo using [TypeScript Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) instead of [TypeScript compilerOptions Paths](https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths). The TypeScript team recommends using project references when working in a monorepo, but until now the configuration settings were difficult to maintain. Each project is required to list its own project dependencies in the `references` property of the `tsconfig.json` file so that TypeScript can incrementally compile projects in the correct order. Nx users know how complex the project graph can become, so maintaining those settings manually is daunting.
```
jaysoo, line 3: The references aren't used for linking, but to break the monorepo into smaller units that TypeScript can understand. We can mention it is for performance and better developer experience.

The linking is now done via Workspaces, but users _could_ technically still use `compilerOptions.paths`.

Here's the blurb I sent to Mike for the blog post:

```
## TypeScript Project References For Monorepos

[**Project References**](https://www.typescriptlang.org/docs/handbook/project-references.html) allows TypeScript to process individual projects rather than the entire monorepo as whole. This is a huge improvement in terms of build times and developer experience. However, project references comes with a maintenance cost, as it is up to you to update these references. In a large monorepo, this cost is prohibitive.

With Nx 20, when using `create-nx-workspace --preset=ts`, workspaces will use Project References out of the box. Nx removes the maintenance cost of this feature by automatically updating the references for you when you run `build` or `typecheck` targets. You can also run [`nx sync`](https://nx.dev/reference/nx-commands#sync) to update references, and `nx sync:check` in CI to validate the workspace.

In addition to adopting Project References, we're also changing the way we link projects in a TypeScript monorepos. **Workspaces** is the standard way to link packages in monorepos. This is a feature that all modern package managers support, such as [npm](https://docs.npmjs.com/cli/using-npm/workspaces), [yarn](https://yarnpkg.com/features/workspaces), [pnpm](https://pnpm.io/workspaces), and [bun](https://bun.sh/docs/install/workspaces), so taking advantage of it is a no-brainer. Now we have a proper way to provide type information and have your packages available in a way developers are used to.

Note that we've enabled Project References and Workspaces for the TS preset (i.e. `--preset=ts`), and we're working on extending this support for all other presets soon (Angular, React, Vue, Node). If you are an existing Nx user and want to create an empty workspace in the previous "integrated" style, you can use `create-nx-workspace --preset=apps`.
```

jaysoo (approved):

PR #28290, opened 2024-10-04
fix(vite): Support coverageProvider from nx.json for various generators

mbarzda:
When my project switched default coverage provider to **istanbul**, React generators didn't provide **istanbul** as coverage provider to vitest configuration. 
After investigation, I noticed that **v8** is hardcoded to every generator which uses `@nx/vite`. 
This change removes hardcoded values from generators and provides `coverageProvider` value from nx.json or fallbacks to **v8**.


Coly010:
Hi @mbarzda !

Can you create an issue for this so that we can discuss and ideal fix that works for all use cases and to explore if there is a better design we could implement? 

Thanks for creating this PR, but I'm going to close it until we can discuss this further in an issue

PR #28317, opened 2024-10-06
fix(js): generated tsconfig extends path

### LINKED ISSUE #28316: @nx/js:tsc - wrong extends path used when generated temporary tsconfig
#### Background

When building with the `tsc` executor, a temporary tsconfig file gets generated for you in `{workspaceRoot}/tmp/{relativeProjectRoot}`

This temporary tsconfig file is meant to use the tsconfig `extends` syntax to extend the tsconfig in your project root with some custom behavior based on your build settings

ex:
```json
{
  "extends": "../../../../../packages/backend/db/tsconfig.lib.json"
}
```

#### The code

Internally, this executor is called [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/executors/tsc/tsc.impl.ts#L62)

The first thing it does is take the tsconfig path in your executor's configuration and [normalizes it](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/executors/tsc/tsc.impl.ts#L68). For example

```json
"build": {
  "executor": "@nx/js:tsc",
  "outputs": [
	"{options.outputPath}"
  ],
  "options": {
	"outputPath": "dist/packages/shared/data-types",
	"main": "packages/shared/data-types/src/index.ts",
	"tsConfig": "packages/shared/data-types/tsconfig.lib.json"
  }
}
```

will turn the relative `tsConfig` path into `'/home/user/Documents/git/my-project/packages/backend/db/tsconfig.lib.json'`

This new absolute path is then passed down to create the temporary tsconfig file in the following steps
1. This line to call into `checkDependencies` [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/executors/tsc/tsc.impl.ts#L70-L73)
2. `checkDependencies` calls `createTmpTsConfig` [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/utils/check-dependencies.ts#L29-L34)
3. The exact value for the `extends` field is calculated in `readTsConfigWithRemappedPaths` [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/utils/buildable-libs-utils.ts#L443-L447)
4. And ``readTsConfigWithRemappedPaths` then calculated the exact path [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/utils/buildable-libs-utils.ts#L203-L206)

#### The bug

Note that for the value of `extends` to be calculated correctly in [readTsConfigWithRemappedPaths](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/utils/buildable-libs-utils.ts#L203-L206), you have two options:
1. Make both paths (`dirnameTsConfig`, `tsConfig`) absolute
2. Make both paths relative to `workspaceRoot`

However, instead, right now `dirnameTsConfig` is turned into a relative path while `tsConfig` is kept as an absolute path (remember we normalized it [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/js/src/executors/tsc/tsc.impl.ts#L68))

This results in the value of `generatedTsConfig.extends` to be incorrect, causing builds to fail

### Expected Behavior

The paths should either both be absolute, or both be relative.

The easiest way to do this was to pass in the relative path into the `checkDependencies` function (since the `tsConfig` variable is not accessed anywhere except to calculate `generatedTsConfig.extends`), but changing `dirnameTsConfig` to always be absolute would also be okay

### GitHub Repo

https://github.com/SebastienGllmt/nx-tsc-repro

### Steps to Reproduce

See repro README

#### Builds seems to work fine for me on my repo.... how come you can reproduce it?

The reason this error wasn't caught before is presumably because if the `extends` path points to an invalid path, builds will still work if they were going to work with the default tsc configuration.

Notably, if you add `console.log(tsConfig)`  [here](https://github.com/nrwl/nx/blob/c2af06e0a12529a5c3d1a09024ccc1fabf4e5604/packages/workspace/src/utilities/typescript/compilation.ts#L145), you can see that these builds actually result in the following error that gets supressed

```json5
errors: [
    {
      file: undefined,
      start: undefined,
      length: undefined,
      messageText: "Cannot read file '/home/user/Documents/git/my-project/tsconfig.lib.json'.",
      category: 1,
      code: 5083,
      reportsUnnecessary: undefined,
      reportsDeprecated: undefined
    },
    {
      file: undefined,
      start: undefined,
      length: undefined,
      messageText: `No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.`,
      category: 1,
      code: 18003,
      reportsUnnecessary: undefined,
      reportsDeprecated: undefined
    }
  ],
```

#### Reliably causing an actual build failure

Although these builds always have a configuration error, as I mentioned that doesn't always actually lead to a build error
 
2. `nx build`

### Nx Report

```
Node           : 20.11.1
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.2.4

nx                 : 19.8.2
@nx/js             : 19.8.2
@nx/linter         : 19.8.2
@nx/eslint         : 19.8.2
@nx/workspace      : 19.8.2
@nx/devkit         : 19.8.2
@nx/eslint-plugin  : 19.8.2
@nrwl/tao          : 19.8.2
@nx/vite           : 19.8.2
@nx/web            : 19.8.2
typescript         : 5.6.2
Registered Plugins:
@nx/eslint/plugin
```

### Failure Logs

```
error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?

Cannot read file '/home/user/Documents/git/my-project/tsconfig.lib.json'
```

### Package Manager Version

pnpm 9.6.0

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

SebastienGllmt:
## Current Behavior

The generated `extends` path in tsconfig is incorrectly generated, causing the build to succeed even if it does nothing

## Expected Behavior

The paths should either both be absolute, or both be relative.

The easiest way to do this was to pass in the relative path into the `checkDependencies` function (since the `tsConfig` variable is not accessed anywhere except to calculate `generatedTsConfig.extends`), but changing `dirnameTsConfig` to always be absolute would also be okay

## Related Issue(s)

Fixes #28316

### DIFF
--- a/packages/js/src/executors/tsc/tsc.impl.ts
+++ b/packages/js/src/executors/tsc/tsc.impl.ts
@@ -69,7 +69,7 @@ export async function* tscExecutor(
 
   const { projectRoot, tmpTsConfig, target, dependencies } = checkDependencies(
     context,
-    options.tsConfig
+    _options.tsConfig
   );
 
   if (tmpTsConfig) {


leosvelperez (changes requested):
hey @SebastienGllmt! Thanks for the detailed investigation!

Let's revert the current change and normalize the file in the helper correctly. That way, we'll guarantee that every consumer of the helper is correct regardless of the path they provide.

SebastienGllmt:
@leosvelperez would you be able to take that on? I'm not sure exactly where you'd like the normalization to take place. You can create your own PR if you want

leosvelperez:
Superseded by https://github.com/nrwl/nx/pull/29121.

Issue #28349, opened 2024-10-07
Issues with `@nx/workspace` for Polyglot Projects


KerickHowlett:
### Current Behavior

I have a project that is utilizing the `@nx-go/nx-go` extension that's used for supporting Go applications.

When I attempt to use the `@nx/workspace:move` generator to move a Go library, it'll produce an error. (Please refer to the _Failure Logs_ section for said error.)

Sadly, there isn't a consistent means to get the bun to start showing up, but once it does, the issue will persist throughout the workspace's life cycle.

I was able to trace this to the compiled Javascript files within Nx's installed project files under the `node_modules` directory, which is where I found a workaround. That being to have the `secondaryEntryPointImportPaths` variable (located within the `updateImports` function) set with an empty array on the same line it's declared.

However, this will produce a `tsconfig.base.json` file at the root of the workspace with the text `undefined` contained within it and with no actual JSON to speak of. 

### Expected Behavior

I have a project that is utilizing the `@nx-go/nx-go` extension that's used for supporting Go applications.

The `@nx/workspace:move` schematic generator should be able to move libraries based in Go and other programming languages without any errors popping up.

It should also do it without generating an empty `tsconfig.base.json` file if one didn't exist previously. This tends to be the case when there is no Typescript to speak of within the Nx workspace. For example, an Nx workspace initiated with the `@nx-go/nx-go` preset.

### GitHub Repo

https://github.com/KerickHowlett/lazynx

### Steps to Reproduce

As stated in the _Current Behavior_ section, it isn't consistent to get this bug to start showing up; however, once it does start popping up, the bug will persist throughout the workspace's lifespan.

Also, keep in mind that these commands  will require the installation of Bun,, as that is what I used for my project that's referenced under _GitHub Repo_.

The best way I can think to recreate it would be with the following:

1. Use the following command to create a workspace: `bun x create-nx-workspace lazynx --preset=@nx-go/nx-go`
2. Use the following command to create a library:`bun run nx g @nx-go/nx-go:library --name=foobar --directory=libs/foobar --projectNameAndRootFormat=as-provided`
3. Repeatedly move and/or remove library or libraries repeatedly until the error appears.

### Nx Report

Node           : 22.9.0
OS             : linux-x64
Native Target  : x86_64-linux
bun            : 1.1.25

nx             : 19.6.2
@nx/workspace  : 19.6.2
@nx/devkit     : 19.6.2
@nrwl/tao      : 19.6.2
typescript     : 5.5.4
[...]

PR #28350, opened 2024-10-07
fix(core): allow for non-js libs to be moved without errors or the creation of any unneeded tsconfig files

KerickHowlett:
## Current Behavior

I have a project that is utilizing the `@nx-go/nx-go` extension that's used for supporting Go applications.

When I attempt to use the `@nx/workspace:move` generator to move a Go library, it'll produce an error. (Please refer to the _Failure Logs_ section for said error.)

Sadly, there isn't a consistent means to get the bun to start showing up, but once it does, the issue will persist throughout the workspace's life cycle.

I was able to trace this to the compiled Javascript files within Nx's installed project files under the `node_modules` directory, which is where I found a workaround. That being to have the `secondaryEntryPointImportPaths` variable (located within the `updateImports` function) set with an empty array on the same line it's declared.

However, this will produce a `tsconfig.base.json` file at the root of the workspace with the text `undefined` contained within it and with no actual JSON to speak of. 

## Expected Behavior

I have a project that is utilizing the `@nx-go/nx-go` extension that's used for supporting Go applications.

The `@nx/workspace:move` schematic generator should be able to move libraries based in Go and other programming languages without any errors popping up.

It should also do it without generating an empty `tsconfig.base.json` file if one didn't exist previously. This tends to be the case when there is no Typescript to speak of within the Nx workspace. For example, an Nx workspace initiated with the `@nx-go/nx-go` preset.

## Related Issue(s)

- [#28349](https://github.com/nrwl/nx/issues/28349)

### DIFF
--- a/packages/workspace/src/generators/move/lib/update-imports.ts
+++ b/packages/workspace/src/generators/move/lib/update-imports.ts
@@ -102,16 +102,18 @@ export function updateImports(
       from: mainEntryPointImportPath,
       to: schema.importPath,
     },
-    ...secondaryEntryPointImportPaths.map((p) => ({
-      from: p,
-      // if the import path doesn't start with the main entry point import path,
-      // it's a custom import path we don't know how to update the name, we keep
-      // it as-is, but we'll update the path it points to
-      to:
-        schema.importPath && p.startsWith(mainEntryPointImportPath)
-          ? p.replace(mainEntryPointImportPath, schema.importPath)
-          : null,
-    })),
+    ...(secondaryEntryPointImportPaths
+      ? secondaryEntryPointImportPaths.map((p) => ({
+          from: p,
+          // if the import path doesn't start with the main entry point import path,
+          // it's a custom import path we don't know how to update the name, we keep
+          // it as-is, but we'll update the path it points to
+          to:
+            schema.importPath && p.startsWith(mainEntryPointImportPath)
+              ? p.replace(mainEntryPointImportPath, schema.importPath)
+              : null,
+        }))
+      : []),
   ];
 
   if (
--- a/packages/workspace/src/generators/move/move.spec.ts
+++ b/packages/workspace/src/generators/move/move.spec.ts
@@ -129,4 +129,20 @@ describe('move', () => {
       '../../node_modules/nx/schemas/project-schema.json'
     );
   });
+
+  it('should work without tsconfig.base.json (https://github.com/nrwl/nx/issues/28349)', async () => {
+    await libraryGenerator(tree, {
+      directory: 'my-lib',
+    });
+    tree.delete('tsconfig.base.json');
+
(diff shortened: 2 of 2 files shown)


isaacplmann:
Don't create a new scope.  Use `core` for changes to workspace.  Someone else will review the actual code changes

KerickHowlett:
> Don't create a new scope. Use `core` for changes to workspace. Someone else will review the actual code changes

Oh! That's my bad. I'll undo the commitizen changes and amend my commit messages.

isaacplmann:
@jaysoo can you take a look at this PR?

packages/workspace/src/generators/move/lib/update-imports.ts
```diff
@@ -148,8 +149,8 @@ export function updateImports(
       to: schema.relativeToRootDestination,
     };
 
-    if (tsConfig) {
-      const path = tsConfig.compilerOptions.paths[projectRef.from] as string[];
+    if (tsConfig && paths.length > 0) {
```
jaysoo, line 152: I think moving `writeJson(...)` into the `if (tsConfig)` is enough to fix the issue. 

Updating `getRootTsConfigPathInTree` is fine since it's only used here and in the remove generator. But we should also change the check to `tree.exists(tsConfigPath)` to account for `null`.

KerickHowlett: That thought did cross my mind, but I was worried about false positives. I'll remove the second clause all the same, though. I'll also go ahead and take a look at `tree.exists(tsConfigPath)` so that it'll account for `null`.

I also believe I can greatly clean up the code quite a bit, too, so that it scales far better. I just wanted to hold off on it in favor of getting these essential changes merged quickly first before with as few changes as possible.

KerickHowlett: I looked into updating `tree.exists(filePath)` to change its argument typing to `string | null` directly without just adding on another if-operator clause.

From what I can tell, it would require touching 3 additional files _(with starting line numbers)_:
1. `packages/devkit/src/utils/invoke-nx-generator.ts:L131`
2. `packages/nx/src/generators/tree.ts:L62`
3. `e2e/utils/file-utils.ts:L111`

The changes seem fairly straight forward, but since this effects a critical Nx component, I figured we could save it for a second PR that I can submit right after this one gets approved. If anything else, just to keep the PRs better focused.

With that said, I applied the necessary changes locally on my machine, and all of the tests appear to be passing on my end.

KerickHowlett: Hey @jaysoo! I merged your changes in with mine, so it should be good to go now.

jaysoo (approved):

jaysoo:
There were fixes from the recent TS solutions changes, just applying small fixes on top of them.

PR #28372, opened 2024-10-09
cleanup(core): migrate nx to picocolors

43081j:

I've left some comments to point out notable changes you might be interested in before merging

### DIFF
--- a/packages/nx/src/utils/code-frames.ts
+++ b/packages/nx/src/utils/code-frames.ts
@@ -1,5 +1,5 @@
 // Adapted from https://raw.githubusercontent.com/babel/babel/4108524/packages/babel-code-frame/src/index.js
-import * as chalk from 'chalk';
+import * as pc from 'picocolors';
 
 type Location = {
   column: number;
@@ -11,14 +11,17 @@ type NodeLocation = {
   start?: Location;
 };
 
+const markerString = (str: string) => pc.red(pc.bold(str));
+const messageString = (str: string) => pc.red(pc.bold(str));
+
 /**
  * Chalk styles for code frame token types.
  */
-function getDefs(chalk) {
+function getDefs() {
   return {
-    gutter: chalk.grey,
-    marker: chalk.red.bold,
-    message: chalk.red.bold,
+    gutter: pc.gray,
+    marker: markerString,
+    message: messageString,
   };
 }
 
@@ -107,7 +110,7 @@ export function codeFrameColumns(
     highlight?: (rawLines: string) => string;
   } = {}
 ): string {
-  const defs = getDefs(chalk);
+  const defs = getDefs();
   const lines = rawLines.split(NEWLINE);
   const { start, end, markerLines } = getMarkerLines(loc, lines, opts);
 
@@ -146,5 +149,5 @@ export function codeFrameColumns(
     })
     .join('\n');
 
-  return chalk.reset(frame);
+  return pc.reset(frame);
 }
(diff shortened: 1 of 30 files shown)


packages/nx/src/utils/code-frames.ts
```diff
@@ -11,14 +11,17 @@ type NodeLocation = {
   start?: Location;
 };
 
+const markerString = (str: string) => pc.red(pc.bold(str));
+const messageString = (str: string) => pc.red(pc.bold(str));
+
 /**
  * Chalk styles for code frame token types.
  */
-function getDefs(chalk) {
```
43081j, line 17: I wasn't sure why we passed `chalk` in since we already unconditionally import it here anyway, so I changed it to just reference `pc` directly

although we probably don't even need a `getDefs` and could just have a constant definition

packages/nx/src/utils/output.ts
```diff
@@ -1,12 +1,13 @@
...
+import * as pc from 'picocolors';
 import { EOL } from 'os';
 import * as readline from 'readline';
-import { isCI } from './is-ci';
 import type { TaskStatus } from '../tasks-runner/tasks-runner';
 
 const GH_GROUP_PREFIX = '::group::';
 const GH_GROUP_SUFFIX = '::endgroup::';
 
+const orangeText = (str: string) => `\x1B[38;5;214m${str}\x1B[39m`;
```
43081j, line 9: we only have one occurrence of a non-standard colour (i.e. an rgb term colour). so I went ahead and just wrote this function for it (we did similar in storybook as it was the same situation, one colour)

let me know if this is ok. we could also just change to a basic colour instead of orange

packages/nx/src/utils/output.ts
```diff
@@ -29,18 +30,9 @@ export interface CLISuccessMessageConfig {
   bodyLines?: string[];
 }
 
-/**
- * Automatically disable styling applied by chalk if CI=true
- */
-const forceColor =
-  process.env.FORCE_COLOR === '' || process.env.FORCE_COLOR === 'true';
-if (isCI() && !forceColor) {
```
43081j, line 37: picocolors checks for NO_COLOR. otherwise there's no way to force the colours off as far as I know

that feels like the right thing to do, though.

maybe we should just hope/assume CI sets NO_COLOR? or document that people should if they want to

packages/nx/src/utils/output.ts
```diff
@@ -103,10 +96,10 @@ class CLIOutput {
 
   applyNxPrefix(color = 'cyan', text: string): string {
     let nxPrefix = '';
-    if (chalk[color]) {
-      nxPrefix = chalk.reset.inverse.bold[color](` ${this.cliName} `);
+    if (pc[color]) {
+      nxPrefix = pc.reset(pc.inverse(pc.bold(pc[color](` ${this.cliName} `))));
     } else {
-      nxPrefix = chalk.reset.inverse.bold.keyword(color)(` ${this.cliName} `);
```
43081j, line 109: this used to allow any RGB colour, and now will only allow basic colours (and will otherwise just show it as bold/inverse)

I haven't been far enough down the rabbit hole to know if we ever passed a non-basic colour into this

thoughts?

packages/nx/src/utils/output.ts
xiongemi: i did a search of `chalk.keyword('orange')`, i think there are a few places outside nx packages also use color orange.

JamesHenry:
Sorry for the long delay here folks, I've opened a fresh PR that's up to date with the latest https://github.com/nrwl/nx/pull/34305

PR #28348, opened 2024-10-07
feat: allow optional configuration in infix run command

akwodkiewicz:
As explained in the linked issue, I cannot leverage `yargs` positional arguments syntax in the command template, because the configuration is an _optional_ argument in the middle of the invocation. I must use custom parsing in the handler. Thankfully it's just about splitting the argument on the first `:`.

## Current Behavior

Infix notation does not allow passing target configuration.

## Expected Behavior

Infix notation allows passing target configuration.

## Related Issue(s)

Fixes #28335

PR #28375, opened 2024-10-09
feat(core): allow optional configuration in infix run command

### LINKED ISSUE #28335: `nx run` infix syntax not working with configurations
### Current Behavior

Given a target with configurations defined in `project.json` it is not possible to run a task with a specific configuration when using the infix syntax.

See "Steps to Reproduce" for the minimal reproduction.

- `nx run foobar:migration:run` works fine ✔ 

  ``` 
  > nx run foobar:migration:run 
  
  > echo running migrations
  
  running migrations
  
   NX   Successfully ran target migration for project foobar (38ms)
  ```
  
- `nx migration foobar` works fine ✔ 

  ```
  > nx run foobar:migration:run
  
  > echo running migrations
  
  running migrations
  
   NX   Successfully ran target migration for project foobar (40ms)
  ```
  
- `nx migration:run foobar` does not work 🐛 :x:

  ```
  NX   Cannot find configuration for task foobar:migration:run
  
  ```
  

### Expected Behavior

`nx migration:run foobar` as well as `nx migration:generate foobar` work in the same way as `nx run foobar:migration:run` and `nx run foobar:migration:generate`.

### GitHub Repo

_No response_

### Steps to Reproduce

Use this `project.json`:

```json
{
  "name": "foobar",
  "targets": {
    "migration": {
      "defaultConfiguration": "run",
      "configurations": {
        "run": {
          "commands": ["echo running migrations"]
        },
        "generate": {
          "commands": ["echo generating migrations"]
        }
      }
    }
  }
}
```

### Nx Report

```
Node           : 18.20.4
OS             : darwin-arm64
yarn           : 1.22.22

nx (global)  : 19.6.4
nx           : 20.0.0
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

The same issue exists on both `19.6.4` and `20.0.0`.

akwodkiewicz:
It looks as if the current shape of `args.target` that is parsed by the `run` infix version and passed to the command handler **makes the `configuration` unrecognizable**. It's probably because the "regular" `run` invocation parses `target` and `configuration` separately, and passes them as 2 distinct properties, while the infix version of the command contains the concatenated `target:configuration` in the `target` argument.

As explained in the linked issue, I couldn't leverage `yargs` positional arguments syntax in the command template, because the configuration is an _optional_ argument _in the middle_ of the invocation. Custom parsing in the handler must be used. Thankfully it's just about splitting the argument on the first `:`.

I do not know if it won't break any existing behaviours, for example the ability to run NPM scripts that contain colons in their names.

## Current Behavior

Infix notation does not allow passing target configuration.

## Expected Behavior

Infix notation allows passing target configuration.

## Related Issue(s)

Duplicate of #28348 that got closed by mistake.
Fixes #28335.

### DIFF
--- a/packages/nx/src/command-line/run/command-object.ts
+++ b/packages/nx/src/command-line/run/command-object.ts
@@ -37,6 +37,12 @@ export const yargsNxInfixCommand: CommandModule = {
   command: '$0 <target> [project] [_..]',
   describe: 'Run a target for a project.',
   handler: async (args) => {
+    // divide <target> into target:configuration if possible
+    const firstColonIdx = (args.target as string).indexOf(':');
+    if (firstColonIdx !== -1) {
+      args.target = (args.target as string).substring(0, firstColonIdx);
+      args.configuration = (args.target as string).substring(firstColonIdx + 1);
+    }
     const exitCode = await handleErrors(
       (args.verbose as boolean) ?? process.env.NX_VERBOSE_LOGGING === 'true',
       async () => {


akwodkiewicz:
Oh, right, [this one does not pass](https://staging.nx.app/runs/CAPsRktYse/task/e2e-nx%3Ae2e-ci--src%2Frun.test.ts)

```
 FAIL   e2e-nx  src/run.test.ts (647.795 s)
  Nx Running Tests
    running targets
      ✓ should execute long running tasks (62317 ms)
      ✕ should run npm scripts (9439 ms)
 ```
 
 ```
 Original command: pnpm exec nx echo:dev mylib4312143 -- positional --a=123 --no-b 

 NX   Cannot find configuration for task mylib4312143:echo
```

So with my change `nx echo:dev mylib` is not able to run `echo:dev` NPM script.

I'm afraid if we wanted to solve #28335, we'd need to either:

  - A) rewrite how we're handling NPM scripts (append value from `configuration` to `target`, before looking at the manifest)
  
or

  - B) rewrite how we're handling `run` itself
  
Both of which increase the complexity of the solution.


A) would mean that `nx run lib:target --configuration=foobar` is going to run `target:foobar` NPM script. It might be an issue when somebody actually tries to pass a `--configuration` flag to the `target` NPM script.

B) would mean we should come up with a precedence of targets/scripts. For example, given `nx foo:bar lib`, we could:

  1. First search for `foo` targets with `bar` configuration. If there is a match for both, run it.
  2. If there is a `foo` task defined, but no `bar` configuration for this task, _don't run it_.
  3. Look for `foo:bar` NPM script.
  
Or in point 2. we could always try to run `foo` with ignored `--configuration bar`, as this is probably more in line with regular `run` invocation.

xiongemi:
yes, you are right, it broke the npm script. maybe we should use b option, before splitting the target by ':', but should check whether the target is already valid.
this should not be that complex. let me look into this option.

xiongemi:
i submit another pr to fix this: https://github.com/nrwl/nx/pull/28402

jaysoo:
As discussed, we won't be extending the format of running targets when using `nx <target>:<config> <project>`. You can use `nx <target> <project> --configuration=<config>` instead.

PR #28376, opened 2024-10-09
test(linter): check if package version is bumped correctly

croesnick:
## Current Behavior

Scenario / setup:

- Using `@nx/dependency-check` eslint rule
- Having a library `libs/lib-a` with its own `package.json` containing `"dependencies": { "external-dep": "^5.1.0" }`
- Having a root `package.json` containing `"dependencies": { "external-dep": "^5.2.0" }`
- Running the dependency-check rule results in dependency `external-dep` in `libs/lib-a/package.json` **not being updated**
- It **is updated**, however if `libs/lib-a/package.json` would contain a lower major version, e.g., `"dependencies": { "external-dep": "^4.5.0" }`

## Expected Behavior

dependency-check rule would update `libs/lib-a/package.json` to contain `"dependencies": { "external-dep": "^5.2.0" }`, i.e., `external-dep` would be identical to the version of the root `package.json`.

Addendum:

If the case were reversed, i.e., the lib's `package.json` mentions a **newer** version than the root `package.json`, I would expect the dependency-check to **not bump the version** in the lib. Rationale: The author of the lib intentionally mentioned a newer version as the lib does not work with an older one.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.spec.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.spec.ts
@@ -1234,7 +1234,7 @@ describe('Dependency checks (eslint)', () => {
       name: '@mycompany/liba',
       dependencies: {
         external1: '~16.0.0',
-        external2: '^1.0.0',
+        external2: '^5.1.0',
       },
     };
 


croesnick:
As expected:

```
 FAIL   packages-eslint-plugin  src/rules/dependency-checks.spec.ts
  ● Dependency checks (eslint) › should error if package version is mismatch and fix it

    expect(received).toEqual(expected) // deep equality

    Expected: 2
    Received: 1

      1284 |       }
      1285 |     );
    > 1286 |     expect(failures.length).toEqual(2);
           |                             ^
      1287 |     expect(failures[0].message).toMatchInlineSnapshot(
      1288 |       `"The version specifier does not contain the installed version of "external1" package: 16.1.8."`
      1289 |     );

      at Object.<anonymous> (src/rules/dependency-checks.spec.ts:1286:29)
```

jaysoo:
This is working as expected, since `^5.1.0` and `^5.2.0` both satisfy the same range.

PR #28402, opened 2024-10-10
fix(core): allow configuration in inflix

xiongemi:
## Current Behavior

this pr is meant to replace https://github.com/nrwl/nx/pull/28375

## Expected Behavior

it should work when doing `nx target:configuration project`.

## Related Issue(s)

Fixes #


jaysoo:
As discussed, we won't be extending the format of running targets when using `nx <target>:<config> <project>`. You can use `nx <target> <project> --configuration=<config>` instead.

PR #28413, opened 2024-10-11
fix(rspack): bugfixes

vadimka123:
After migration from webpack to rspack I noticed some issues 

This PR fixing:
1. typeCheck by providing absolute path of tsConfigPath
2. fileReplacements option by providing absolute path
3. using deploy url in index.html file
4. conflicting values for 'process.env.NODE_ENV' ("'development'" !== "\"development\"")


Coly010:
We've addressed these in our recent work to improve rspack and get feature parity with Webpack

PR #28454, opened 2024-10-15
fix(core): add migration to set `useLegacyCache` by default

Cammisuli:
## Current Behavior

Existing workspaces that migrate to v20 use the db cache by default. 

## Expected Behavior

Existing workspaces that migrate to v20 have `useLegacyCache` set to true. Unless they specifically had `enableDbCache` from previous versions of Nx. All new workspaces will not have `useLegacyCache` set by default. 

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/migrations.json
+++ b/packages/nx/migrations.json
@@ -61,7 +61,8 @@
     "use-legacy-cache": {
       "version": "20.0.1",
       "description": "Set `useLegacyCache` to true for migrating workspaces",
-      "implementation": "./src/migrations/update-20-0-1/use-legacy-cache"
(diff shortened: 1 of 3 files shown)


jaysoo (approved):

JamesHenry (approved):

sleeyax:
Is it possible to migrate an existing workspace to the db cache instead? Can't seem to find this documented anywhere.

JamesHenry:
@sleeyax Just migrate to Nx 20 as normal, and remove this `useLegacyCache` flag from your nx.json if it is present. I would advise running `nx reset` any time after your make changes related to the cache (such as adding or removing this flag).

sleeyax:
Gotcha, thanks for the info!

PR #28503, opened 2024-10-17
fix(linter): move eslint-config-prettier to the end for proper override

ghost:
There is a conflict between ESLint and Prettier rules, which causes incorrect formatting behavior.

Reordering the config so that `eslint-config-prettier` is applied last ensures proper rule overrides and resolves conflicts with Prettier.


packages/eslint-plugin/src/flat-configs/javascript.ts
```diff
@@ -80,5 +79,10 @@ export default tseslint.config(
        */
       '@typescript-eslint/no-require-imports': 'off',
     },
-  }
+  },
+  /**
+   * With the new ESLint “flat config” format, you can control what things override what yourself.
+   * One way of solving the above conflict is to reorder the config objects so that eslint-config-prettier is last:
+   */
```
leosvelperez, line 86:
```suggestion
  /**
   * We include it last so it overrides the conflicting rules from the configuration blocks above.
   */
```

packages/eslint-plugin/src/flat-configs/typescript.ts
```diff
@@ -64,5 +63,10 @@ export default tseslint.config(
        */
       '@typescript-eslint/no-require-imports': 'off',
     },
-  }
+  },
+  /**
+   * With the new ESLint “flat config” format, you can control what things override what yourself.
+   * One way of solving the above conflict is to reorder the config objects so that eslint-config-prettier is last:
+   */
```
leosvelperez, line 70:
```suggestion
  /**
   * We include it last so it overrides the conflicting rules from the configuration blocks above.
   */
```

leosvelperez (changes requested):
Thank you for taking the time to contribute!

The change looks good to me. There's only a minor change suggested to the source code comments.

packages/eslint-plugin/src/flat-configs/typescript.ts
leosvelperez: It seems you missed the suggestion above. It has not been applied yet.

leosvelperez (approved):
Thank you!

PR #28522, opened 2024-10-18
feat(testing): add disableJestRuntime option to @nx/jest/plugin to speed up target inference

jaysoo:
This PR adds a new `disableJestRuntime` option to `@nx/jest/plugin`. By setting this to `true`, the inference plugin will not use `jest-config` and `jest-runtime` to retrieve `outputs`, nor matching spec files when `ciTargetName` is used. Instead, we'll use our own config loader and glob/matcher to get both values for the inferred targets.

This speeds up computation quite a bit. With a large monorepo (800 projects, 75K files), the computation goes from 2m 41s to 35s, or ~78% time reduction.

NOTE: This has no effect on existing projects and needs to be opted into.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/plugins/plugin.ts
+++ b/packages/jest/src/plugins/plugin.ts
@@ -1,4 +1,3 @@
-import type { Config } from '@jest/types';
 import {
   CreateNodes,
   CreateNodesContext,
@@ -28,12 +27,23 @@ import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
 import { combineGlobPatterns } from 'nx/src/utils/globs';
 import { dirname, isAbsolute, join, relative, resolve } from 'path';
 import { getInstalledJestMajorVersion } from '../utils/version-utils';
+import {
+  getFilesInDirectoryUsingContext,
+  globWithWorkspaceContext,
+} from 'nx/src/utils/workspace-context';
+import { normalize } from 'node:path';
 
 const pmc = getPackageManagerCommand();
 
 export interface JestPluginOptions {
   targetName?: string;
   ciTargetName?: string;
+  /**
+   *  Whether to use jest-config and jest-runtime to load Jest configuration and context.
+   *  Disabling this is much faster but could be less correct since we are using our own config loader
+   *  and test matcher instead of Jest's.
+   */
+  disableJestRuntime?: boolean;
 }
 
 type JestTargets = Awaited<ReturnType<typeof buildJestTargets>>;
@@ -57,11 +67,19 @@ export const createNodesV2: CreateNodesV2<JestPluginOptions> = [
     const optionsHash = hashObject(options);
     const cachePath = join(workspaceDataDirectory, `jest-${optionsHash}.hash`);
     const targetsCache = readTargetsCache(cachePath);
+    // Cache jest preset(s) to avoid penalties of module load times. Most of jest configs will use the same preset.
+    const presetCache: Record<string, unknown> = {};
 
     try {
       return await createNodesFromFiles(
         (configFile, options, context) =>
-          createNodesInternal(configFile, options, context, targetsCache),
+          createNodesInternal(
+            configFile,
+            options,
+            context,
+            targetsCache,
+            presetCache
+          ),
         configFiles,
         options,
         context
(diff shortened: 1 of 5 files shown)


isaacplmann (approved):

packages/jest/src/plugins/plugin.ts
```diff
@@ -459,3 +581,75 @@ function requireJestUtil<T>(
...
+  if (rawConfig[optionName]) return rawConfig[optionName];
+
+  if (rawConfig.preset) {
+    const dir = dirname(absConfigFilePath);
+    const presetPath = resolve(dir, rawConfig.preset);
+    try {
+      const preset = await loadConfigFile(presetPath);
+      if (preset[optionName]) return preset[optionName];
+    } catch {
+      throw new Error(`Could not load preset file: ${presetPath}`);
```
leosvelperez, line 599: I think we should safely handle this and ignore this error. We can still construct the inferred project/tasks without any given option. We could warn instead.

IMHO, only fatal errors that prevent Nx from building a project should result in a graph error. We should let the tool itself handle the errors when running the tasks. It's also a matter of perspective: is Nx erroring (the graph is an Nx thing) or is the underlying tool erroring (when running the task)? Also, when the graph is in an errored state, no task, even if unrelated, can be run.

So, safely handling this would allow the task to be inferred and run. We'd be letting Jest handle the wrong preset file and error accordingly instead of us failing the graph and preventing the running of any tasks.

packages/jest/src/plugins/plugin.ts
```diff
@@ -459,3 +581,75 @@ function requireJestUtil<T>(
...
+  absConfigFilePath: string,
+  optionName: string
+): Promise<T> {
+  if (rawConfig[optionName]) return rawConfig[optionName];
+
+  if (rawConfig.preset) {
+    const dir = dirname(absConfigFilePath);
+    const presetPath = resolve(dir, rawConfig.preset);
+    try {
+      const preset = await loadConfigFile(presetPath);
```
leosvelperez, line 596: We should cache the preset. This will be called multiple times, and most of the time, it will load the same preset path.

packages/jest/src/plugins/plugin.ts
```diff
@@ -208,90 +199,172 @@ async function buildJestTargets(
...
-              options: {
-                coverage: true,
+      rawConfig.coverageDirectory
+        ? join(context.workspaceRoot, projectRoot, rawConfig.coverageDirectory)
+        : undefined,
+      undefined,
+      context
+    ));
+
+    if (options?.ciTargetName) {
```
leosvelperez, line 222: nit: we can probably avoid a lot of duplication building the target by moving the handling of the `disableJestRuntime` option into the different methods, but that can be a follow-up cleanup.

leosvelperez (changes requested):

packages/jest/src/plugins/plugin.ts
jaysoo: I think it's a rule-of-three situation. Duplicating once is fine, a third time is when you extract.

packages/jest/src/plugins/plugin.ts
jaysoo: I feel like we should do this at the `loadConfigFile` level. It can lead to unexpected caching issues if we're not careful. For example, we need to bust the cache if either the preset changes, or anything it import changes.

packages/jest/src/plugins/plugin.ts
```diff
@@ -208,90 +210,173 @@ async function buildJestTargets(
...
+        const specIgnoreRegexes: undefined | RegExp[] =
+          rawConfig.testPathIgnorePatterns?.map(
+            (p: string) => new RegExp(replaceRootDirInPath(projectRoot, p))
+          );
+
+        for (const testPath of testPaths) {
+          const relativePath = normalizePath(
+            relative(join(context.workspaceRoot, projectRoot), testPath)
+          );
+          if (specIgnoreRegexes?.some((regex) => regex.test(relativePath)))
```
leosvelperez, line 282: It would be more correct to move the check for ignored specs into the `getTestPaths` function or at least to filter the ignored specs before the [line 242](https://github.com/nrwl/nx/pull/28522/files#diff-a1531bb2dca28c044b3dc539f643224c0f946c60208830a3bc91402eb50abeaaR242). It's definitely an edge case and wouldn't break anything, but the way it is right now, it _**could**_ happen that we create the `e2e-ci` target, but then, we don't create any target for any specs because they are all ignored.

As mentioned, it's not exactly an issue, as in nothing would break, but it wouldn't match the plugin's outcome when using the `jest-runtime` package.

jaysoo: Changed the check to ensure that `targetGroup` is not empty, so that we do not create `e2e-ci` target unless there is at least one spec that that in the `targetGroup`.

leosvelperez (approved):

Issue #7047, opened 2021-09-18
recommended vscode extensions contains tslint while selected eslint when create new workspace from scratch


pumano:
`npx create-nx-workspace --preset=angular` add recommended vscode extension tslint while I create clean workspace with angular project using eslint.
## Current Behavior

.vscode extensions.json contains `"ms-vscode.vscode-typescript-tslint-plugin"` but I select eslint
## Expected Behavior

Should add eslint plugin to recommended vscode plugin
## Steps to Reproduce

`npx create-nx-workspace --preset=angular`
select css, eslint
check .vscode/extensions.json

PR #28573, opened 2024-10-23
feat(core): add eslint vscode extension on eslint init

pumano:
## Current Behavior

when create fresh angular or react integrated project no eslint extension out of the box. Few years ago I fill the issue https://github.com/nrwl/nx/issues/7047 about deprecated tslint extension looks like that has been deleted but without change to eslint extension
## Expected Behavior

when I create angular / react  project eslint extension should be added to recommended vscode extensions
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/eslint/src/generators/init/init.spec.ts
+++ b/packages/eslint/src/generators/init/init.spec.ts
@@ -1,6 +1,12 @@
 import 'nx/src/internal-testing-utils/mock-project-graph';
 
-import { NxJsonConfiguration, readJson, Tree, updateJson } from '@nx/devkit';
+import {
+  NxJsonConfiguration,
+  readJson,
+  Tree,
+  updateJson,
+  writeJson,
+} from '@nx/devkit';
 import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
 import { LinterInitOptions, lintInitGenerator } from './init';
 import { setWorkspaceRoot } from 'nx/src/utils/workspace-root';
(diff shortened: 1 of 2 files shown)


pumano:
locally build and tests is OK

MrG9090 (approved):

leosvelperez (changes requested):
Thanks for the contribution!

This makes sense, but let's revert these changes in the `@nx/workspace:new` generator and instead, add the extension in the `@nx/eslint:init` generator. You can check how we do this in other generators (e.g. https://github.com/nrwl/nx/blob/master/packages/jest/src/generators/configuration/lib/update-vscode-recommended-extensions.ts).

pumano:
@leosvelperez OK thank you, I will change it

packages/eslint/src/generators/init/init.spec.ts
```diff
@@ -73,6 +83,30 @@ describe('@nx/eslint:init', () => {
...
+  it('should add the eslint extension to the recommended property', () => {
+    writeJson(tree, '.vscode/extensions.json', {
+      recommendations: [
+        'nrwl.angular-console',
+        'angular.ng-template',
+        'esbenp.prettier-vscode',
+      ],
+    });
+
+    updateVsCodeRecommendedExtensions(tree);
```
leosvelperez, line 95: Let's test the extension is added when invoking the `lintInitGenerator`. That would guard against any potential refactor that, by mistake, could remove the invocation of the `updateVsCodeRecommendedExtensions` helper.
```suggestion
  it('should add the eslint extension to the recommended property', async () => {
    writeJson(tree, '.vscode/extensions.json', {
      recommendations: [
        'nrwl.angular-console',
        'angular.ng-template',
        'esbenp.prettier-vscode',
      ],
    });

    await lintInitGenerator(tree, options);
  });
```

packages/eslint/src/generators/init/init.ts
```diff
@@ -49,6 +50,21 @@ function addTargetDefaults(tree: Tree) {
   updateNxJson(tree, nxJson);
 }
 
+export function updateVsCodeRecommendedExtensions(host: Tree) {
```
leosvelperez, line 53: With the suggested change in the test case, we don't need to export this function anymore.
```suggestion
function updateVsCodeRecommendedExtensions(host: Tree) {
```

leosvelperez (changes requested):
Almost there! Just a couple more comments.

packages/eslint/src/generators/init/init.spec.ts
pumano: Ok I would love to save code that check result after lintInitGenerator executed

leosvelperez (approved):
Thank you!

PR #28587, opened 2024-10-23
chore(core): task distribution defaults to off

isaacplmann:
- Updates workspace:ci-workflow generator to disable task distribution by default
- Updates gradle:ci-workflow generator to disable task distribution by default
- Updates all tutorials to reflect the new generator and adds a line explaining how to enable task distribution

### DIFF
--- a/packages/workspace/src/generators/ci-workflow/files/azure/azure-pipelines.yml__tmpl__
+++ b/packages/workspace/src/generators/ci-workflow/files/azure/azure-pipelines.yml__tmpl__
@@ -55,12 +55,12 @@ jobs:
       # This enables task distribution via Nx Cloud
       # Run this command as early as possible, before dependencies are installed
       # Learn more at https://nx.dev/ci/reference/nx-cloud-cli#npx-nxcloud-startcirun
-      <% if (connectedToCloud) { %># Uncomment this line to enable task distribution
+      <% if (connectedToCloud) { %># Uncomment this line to enable task distribution<% } else { %># Connect your workspace by running "nx connect" and uncomment this line to enable task distribution<% } %>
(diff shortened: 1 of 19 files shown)


packages/workspace/src/generators/ci-workflow/files/azure/azure-pipelines.yml__tmpl__
```diff
@@ -55,8 +55,9 @@ jobs:
       # This enables task distribution via Nx Cloud
       # Run this command as early as possible, before dependencies are installed
       # Learn more at https://nx.dev/ci/reference/nx-cloud-cli#npx-nxcloud-startcirun
-      <% if (connectedToCloud) { %>- script: <%= packageManagerPreInstallPrefix %> nx-cloud start-ci-run --distribute-on="3 linux-medium-js" --stop-agents-after="<% if(hasE2E){ %>e2e-ci<% } else { %>build<% } %>"
-      <% } else { %># Connect your workspace by running "nx connect" and uncomment this
+      <% if (connectedToCloud) { %># Uncomment this line to enable task distribution
```
jaysoo, line 58: We can remove the if-else since it's always commented out now. The instruction to uncomment to enable distribution should be the same.

Same thing for the other CI templates.

jaysoo:
We also need to add a new check to see if either `cypress` or `@playwright/test` and add a step to install browsers after packages are installed.

e.g. If user is using pnpm and Playwright then they should this in their GitHub pipeline file.

```
- run: pnpm install --frozen-lockfile
- run: pnpm exec playwright install
```

packages/workspace/src/generators/ci-workflow/files/azure/azure-pipelines.yml__tmpl__
isaacplmann: The if else is checking if the workspace is connected to Nx cloud already.  Not whether they want task distribution

MrG9090 (approved):

jaysoo (approved):

PR #28672, opened 2024-10-29
fix(linter): ensure .cjs config file is handled correctly for generators #28214

Coly010:
## Current Behavior

When the workspace has `eslint.config.cjs` at the root, generation of new lint projects currently does not check for it correctly 

## Expected Behavior

Ensure utils continue to work as expected when `eslint.config.cjs` is detected at the workspaceRoot

## Related Issue(s)

Fixes #28214

### DIFF
--- a/packages/eslint/src/generators/lint-project/lint-project.ts
+++ b/packages/eslint/src/generators/lint-project/lint-project.ts
@@ -273,7 +273,7 @@ function createEsLintConfiguration(
     });
     const nodeList = createNodeList(importMap, nodes);
     const content = stringifyNodeList(nodeList);
-    const ext = extendedRootConfig.endsWith('.cjs') ? '.cjs' : '.js';
+    const ext = extendedRootConfig?.endsWith('.cjs') ? '.cjs' : '.js';
     tree.write(join(projectConfig.root, `eslint.config${ext}`), content);
   } else {
     writeJson(tree, join(projectConfig.root, `.eslintrc.json`), {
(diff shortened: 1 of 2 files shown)


packages/eslint/src/generators/utils/eslint-file.ts
```diff
@@ -64,7 +64,11 @@ export function isEslintConfigSupported(tree: Tree, projectRoot = ''): boolean {
   if (!eslintFile) {
     return false;
   }
-  return eslintFile.endsWith('.json') || eslintFile.endsWith('.config.js');
+  return (
```
meeroslav, line 67: Should we also include `mjs` here? Use regex instead to match all three.

packages/eslint/src/generators/lint-project/lint-project.ts
```diff
@@ -273,7 +273,8 @@ function createEsLintConfiguration(
     });
     const nodeList = createNodeList(importMap, nodes);
     const content = stringifyNodeList(nodeList);
-    tree.write(join(projectConfig.root, 'eslint.config.js'), content);
+    const ext = extendedRootConfig?.endsWith('.cjs') ? '.cjs' : '.js';
+    tree.write(join(projectConfig.root, `eslint.config${ext}`), content);
```
meeroslav, line 277:
```suggestion
    const ext = extendedRootConfig?.split('.').pop() || 'js';
    tree.write(join(projectConfig.root, `eslint.config.${ext}`), content);
```

meeroslav (approved):

packages/eslint/src/generators/lint-project/lint-project.ts
Coly010: Let's support MJS in a separate PR

packages/eslint/src/generators/utils/eslint-file.ts
Coly010: Let's support MJS in a separate PR

PR #28684, opened 2024-10-29
fix(core): not exit when one plugin installation failed

xiongemi:
## Current Behavior

it tries to install all plugins at once. if failed, it will stop the remaining plugins from being installed.

## Expected Behavior

it will continue if one plugin failed.

success message:
Screenshot 2024-12-11 at 11 36 14 AM
failed message:
Screenshot 2024-12-12 at 2 58 17 PM

## Related Issue(s)

Fixes #


packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -51,18 +54,47 @@ export function installPlugins(
...
+    });
+  }
+  if (failedPlugins.length > 0) {
+    output.error({
+      title: 'Failed to install Plugins',
+      bodyLines: [
+        ...failedPlugins.map((p) => `- ${p}`),
+        'You may need to install them manually.',
+      ],
+    });
```
FrozenPandaz, line 95: So what is the state after this?

The failed plugins will have been `npm` installed.. but they won't be added to `nx.json` right?

Can we show the error details here?

Installing the plugin manually would result in the same thing right? Are there cases where it will suddenly succeed after it just failed? We should instruct people to do something which will fix the issue...

FrozenPandaz (changes requested):

packages/nx/src/command-line/init/init-v2.ts
xiongemi: i changed something like:
Screenshot 2024-11-08 at 5 37 16 PM

packages/js/src/utils/typescript/ts-solution-setup.ts
```diff
@@ -94,5 +94,7 @@ export function assertNotUsingTsSolutionSetup(
     ],
   });
 
-  process.exit(1);
+  throw new Error(
+    `The ${artifactString} doesn't yet support the existing TypeScript setup`
```
leosvelperez, line 98:
```suggestion
    `The ${artifactString} doesn't yet support the existing TypeScript setup. See the error above.`
```

packages/nx/src/command-line/init/configure-plugins.ts
```diff
@@ -0,0 +1,159 @@
...
+      succeededPlugins.push(plugin);
+      spinner.succeed('Installed plugin ' + plugin);
+    } catch (e) {
+      failedPlugins[plugin] = e;
+      spinner.fail('Failed to install plugin ' + plugin);
+    }
+  }
+  host.lock();
+  const changes = host.listChanges();
+  flushChanges(repoRoot, changes);
```
FrozenPandaz, line 88: Ideally, this would be able to work like this but the host will have partial changes when failures are encountered.

A new tree should be created and flushed before the next plugin starts and creates a new tree much like it does currently.

FrozenPandaz (changes requested):
Please create a Loom for this one as well.

packages/nx/src/command-line/init/configure-plugins.ts
```diff
@@ -0,0 +1,210 @@
...
+    repoRoot,
+    {}
+  );
+
+  const implementation: NxGenerator = implementationFactory();
+  const task: GeneratorCallback | void = await implementation(host, options);
+  if (task) {
+    await task();
+  }
+  flushChanges(repoRoot, host.listChanges());
```
FrozenPandaz, line 74: This should be the other way around.

Flush then run tasks.

packages/nx/src/command-line/init/configure-plugins.ts
```diff
@@ -0,0 +1,210 @@
...
+    plugins,
+    updatePackageScripts,
+    repoRoot,
+    verbose
+  );
+
+  if (succeededPlugins.length > 0) {
+    if (destinationGitClient) {
+      destinationGitClient.amendCommit();
+    }
```
FrozenPandaz, line 160: Handle this in `import`

packages/nx/src/command-line/import/import.ts
```diff
@@ -287,42 +282,28 @@ export async function importHandler(options: ImportOptions) {
...
-    }
-  } else if (await needsInstall(packageManager, originalPackageWorkspaces)) {
-    try {
-      output.log({
-        title: 'Installing dependencies for imported code',
-      });
-
-      runInstall(workspaceRoot, getPackageManagerCommand(packageManager));
-
-      await destinationGitClient.amendCommit();
```
FrozenPandaz, line 313: Keep this call in here. You'll need to check if there are successful plugins.

FrozenPandaz (changes requested):

packages/nx/src/command-line/import/import.ts
xiongemi: changed

packages/nx/src/command-line/init/configure-plugins.ts
xiongemi: changed

packages/nx/src/command-line/init/configure-plugins.ts
xiongemi: changed

FrozenPandaz (approved):

PR #28693, opened 2024-10-30
fix(vite): add correct gitignore pattern for vite timestamp files #28685

### LINKED ISSUE #28685: ignoring vite.config temporary files - gitignore doesn't support brace expansion
### Current Behavior

In recent update added `**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*` to the `.gitignore` file but the ignore file doesn't support brace expansion.

also it doesn't need the `**/` at the begining of the string which would better follow how previous file ignores have been added.

https://git-scm.com/docs/gitignore 

### Expected Behavior

file like vite.config.mts.timestamp-1728501368696-41bfd7a885ddf.mjs should be ignored

### GitHub Repo

_No response_

### Steps to Reproduce

Run the migrations for 20.0.6 that add the line to the gitignore then create a file and see if it is actually ignored.

### Nx Report
```

Node           : 20.14.0
OS             : darwin-arm64
npm            : 10.7.0

nx (global)        : 20.0.0
nx                 : 20.0.6
@nx/js             : 20.0.6
@nx/jest           : 20.0.6
@nx/eslint         : 20.0.6
@nx/workspace      : 20.0.6
@nx/devkit         : 20.0.6
@nx/eslint-plugin  : 20.0.6
@nx/express        : 20.0.6
@nx/node           : 20.0.6
@nx/vite           : 20.0.6
@nx/vue            : 20.0.6
@nx/web            : 20.0.6
@nx/webpack        : 20.0.6
typescript         : 5.5.4
Registered Plugins:
@nx/eslint/plugin
@nx/vite/plugin
@nx/jest/plugin
@nx/webpack/plugin
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I am not sure the gitignore solution actually solves the core issue.

Coly010:
## Current Behavior

The pattern added for .gitignore is incorrect

## Expected Behavior

Add the correct pattern

## Related Issue(s)

Fixes #28685

### DIFF
--- a/packages/vite/src/generators/init/init.spec.ts
+++ b/packages/vite/src/generators/init/init.spec.ts
@@ -131,18 +131,14 @@ describe('@nx/vite:init', () => {
 
   it(`should not add multiple instances of the same vite temp file glob to gitignore`, async () => {
     // ARRANGE
-    tree.write(
-      '.gitignore',
-      '**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*'
-    );
+    tree.write('.gitignore', 'vite.config.*.timestamp*');
 
     // ACT
     await initGenerator(tree, {});
 
     // ASSERT
-    expect(tree.read('.gitignore', 'utf-8')).toMatchInlineSnapshot(`
-      "**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*
-      **/vite.config.*.timestamp*"
-    `);
+    expect(tree.read('.gitignore', 'utf-8')).toMatchInlineSnapshot(
+      `"vite.config.*.timestamp*"`
+    );
   });
 });
--- a/packages/vite/src/generators/init/init.spec.ts
+++ b/packages/vite/src/generators/init/init.spec.ts
@@ -140,8 +140,9 @@ describe('@nx/vite:init', () => {
     await initGenerator(tree, {});
 
     // ASSERT
-    expect(tree.read('.gitignore', 'utf-8')).toMatchInlineSnapshot(
-      `"**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*"`
-    );
+    expect(tree.read('.gitignore', 'utf-8')).toMatchInlineSnapshot(`
+      "**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*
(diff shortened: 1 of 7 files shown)


packages/vite/src/generators/init/init.spec.ts
```diff
@@ -140,8 +140,9 @@ describe('@nx/vite:init', () => {
     await initGenerator(tree, {});
 
     // ASSERT
-    expect(tree.read('.gitignore', 'utf-8')).toMatchInlineSnapshot(
-      `"**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*"`
-    );
+    expect(tree.read('.gitignore', 'utf-8')).toMatchInlineSnapshot(`
+      "**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*
+      **/vite.config.*.timestamp*"
+    `);
```
leosvelperez, line 146: We need to adjust this test case, which is supposed to validate that the glob won't be duplicated.

packages/vite/src/utils/add-vite-temp-files-to-gitignore.ts
```diff
@@ -1,7 +1,7 @@
 import { stripIndents, Tree } from '@nx/devkit';
 
 export function addViteTempFilesToGitIgnore(tree: Tree) {
-  let newGitIgnoreContents = `**/vite.config.{js,ts,mjs,mts,cjs,cts}.timestamp*`;
+  let newGitIgnoreContents = `**/vite.config.*.timestamp*`;
```
leosvelperez, line 4: We can skip the leading `**` and the result will be the same. As per [Git docs](https://git-scm.com/docs/gitignore#_pattern_format):

> A leading "\*\*" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo".

```suggestion
  let newGitIgnoreContents = `vite.config.*.timestamp*`;
```

leosvelperez (changes requested):

leosvelperez (approved):

jaysoo (approved):

PR #28694, opened 2024-10-30
chore(angular): remove superfluous test

yjaaidi:
We always use .mts extension. It is not conditional anymore.


leosvelperez (changes requested):
Hi @yjaaidi! Could you please rebase and re-run the tests with `-u` so the obsolete snapshots are also removed?

yjaaidi:
> Hi @yjaaidi! Could you please rebase and re-run the tests with `-u` so the obsolete snapshots are also removed?

✅ done.

leosvelperez (approved):
Thank you!

PR #28696, opened 2024-10-30
fix(angular): use vitest generator instead of vite config generator

Coly010:
## Current Behavior

Using the viteConfigGenerator leads to build and serve targets being overriden

## Expected Behavior

Use the `vitestGenerator` to only add the test target

## Related Issue(s)

Fixes #


yjaaidi:
🙏 Thakns Colum!

leosvelperez (approved):

PR #28724, opened 2024-10-31
feat(nx-plugin): add support for the ts solution config setup to the `@nx/plugin` plugin

leosvelperez:
## Related Issue(s)

Fixes #


packages/js/src/executors/tsc/lib/normalize-options.ts
```diff
@@ -57,5 +57,6 @@ export function normalizeOptions(
       outputPath,
       options.main.replace(`${projectRoot}/`, '').replace('.ts', '.js')
     ),
+    generatePackageJson: options.generatePackageJson ?? true,
```
jaysoo, line 60: Do we need this if the schema defaults to true? This would affect programmatic usage only.

packages/plugin/src/generators/create-package/create-package.ts
```diff
@@ -56,6 +70,20 @@ export async function createPackageGenerator(
     await formatFiles(host);
   }
 
+  if (options.isTsSolutionSetup) {
+    const projectPackageManagerWorkspaceState =
+      getProjectPackageManagerWorkspaceState(host, options.projectRoot);
+
+    if (projectPackageManagerWorkspaceState !== 'included') {
+      tasks.push(
+        getProjectPackageManagerWorkspaceStateWarningTask(
```
jaysoo, line 79: Maybe a later enhancement, but if we know that the project is missing from workspaces then we could add the entry for them automatically. We had the same idea for `nx import` when the entry is missing.

packages/plugin/src/generators/create-package/create-package.ts
```diff
@@ -97,18 +126,29 @@ async function createCliPackage(
...
+  updateJson<PackageJson>(
     host,
     joinPathFragments(options.projectRoot, 'package.json'),
     (packageJson) => {
       packageJson.bin = {
         [options.name]: './bin/index.js',
       };
+      if (isTsSolutionSetup) {
+        packageJson.bin[options.name] = './dist/bin/index.js';
+        delete packageJson.main;
```
jaysoo, line 147: Can we add a comment to describe why these deletions are necessary?

packages/plugin/src/generators/e2e-project/e2e.ts
```diff
@@ -1,50 +1,85 @@
...
+  isTsSolutionSetup: boolean;
 }
 
 async function normalizeOptions(
   host: Tree,
   options: Schema
 ): Promise<NormalizedSchema> {
+  const isTsSolutionSetup = isUsingTsSolutionSetup(host);
+
+  let linter = options.linter;
```
jaysoo, line 55: We do these prompts a lot, maybe it's a clean up item later to extract a shared util to `@nx/js`.

gperdomor:
This will be included in 20.1 release? 🙏🏻

packages/js/src/executors/tsc/lib/normalize-options.ts
leosvelperez: Yeah, we need it precisely to ensure that the programmatic usage gets the correct default value and doesn't get broken with this new option.

packages/plugin/src/generators/create-package/create-package.ts
leosvelperez: Yeah, we discussed this at the beginning and agreed on logging a warning for now. I'll create a separate task to add the entry automatically.

jaysoo (approved):

PR #28735, opened 2024-10-31
feat(nx-cloud): configure import paths for light client when running …

lourw:
…login/logout

## Current Behavior

When you call `nx login` and `nx logout`, we did not pass in import paths to node modules so the displayed messages for each command were just raw text. 

## Expected Behavior

When you call `login` or `logout`, you get formatting on the command output. 

## Related Issue(s)

Fixes #


lourw:
Screenshot 2024-10-31 at 12 32 24 PM
 vs 
 
Screenshot 2024-10-31 at 12 32 11 PM

FrozenPandaz (approved):

PR #28747, opened 2024-11-01
fix(core): disable the db by default

Cammisuli:
## Related Issue(s)

Fixes #


vsavkin (approved):

FrozenPandaz:
I'm going to close this. We migrated existing workspaces to not use the db cache for now instead. It can remain on for new workspaces.

PR #28748, opened 2024-11-01
docs(bundling): correct esbuild options nesting

vmasek:
Fixes:

- [x] correct nesting of options in ESBuild docs


pumano:
@vmasek looks like build failed due to documentation, try to build docs and update snapshots (try to check package.json for commands)

vmasek:
@pumano thanks, I did the PR quickly in GitHub online IDE but I see I need to have Rust & Cargo installed for `nx run @nx/nx-source:documentation` and updating it, so it calls for proper computer experience 😅

pumano:
@vmasek I highly recommend install rust via https://www.rust-lang.org/tools/install because installing from brew can cause problems later

Coly010 (approved):

vmasek:
@Coly010 Thanks for picking it up, I had some difficulties setting up rust and then forgot about this PR 😅

PR #28770, opened 2024-11-03
feat(core): add `--stdin` to affected options

aaronccasanova:
## Current Behavior

Executing `nx affected` with a large list of files using the `--files` option frequently results in `E2BIG` errors during CI runs. This is compounded by our project's deeply nested structure, which returns lengthy file paths even in smaller pull requests.

## Expected Behavior

This PR [adds a `--stdin` flag to all affected commands](https://github.com/nrwl/nx/commit/819d3491b6523eaa2aed9c8fe4ddac4047a9122f), enabling the list of changed files to be provided via standard input. This approach helps mitigate `E2BIG` errors by avoiding long command-line arguments. Usage example:

```bash
shopify-build changed-files | nx affected -t test --stdin
```

The use of a `--stdin` flag for inputing files was inspired by [`jscodeshift`](https://github.com/facebook/jscodeshift?tab=readme-ov-file#usage-cli) and [`polaris-migrator`](https://github.com/Shopify/polaris/blob/c2583f31a8b3eb07970b676664c46365dfb7af59/polaris-migrator/src/cli.ts#L97-L112). This implementation extends the existing `--files` option to minimize changes and maintain backward compatibility. Open to alternative suggestions and improvements.


aaronccasanova:
cc'ing other contributors to these shared options @AgentEnder @FrozenPandaz

FrozenPandaz:
Thank you for your PR. I think passing files in via `stdin` makes sense.

Can we pause to consider if CSV is the best format or if for example, splitting by newlines is better.

The canonical example of why one would use `--files` is if they used other version management tools such as Mercurial instead of `git`. For that reason, I think accepting a newline delimited list, as jscodeshift and polaris-migrator do, works better for Mercurial as the command used would be `hg status --template "{path}\n"`. It's a little harder to provide CSV because mercurial would end up with a trailing `,`. 

It seems like you're using a custom command to list the files. Are there any underlying tools under the hood or was it originally written to be inserted in as `nx affected --files "${shopify changed-files}"`?

aaronccasanova:
> Can we pause to consider if CSV is the best format or if for example, splitting by newlines is better.

I'm impartial and happy to defer to your judgment here. While I omitted the details of our custom command for brevity, we currently transform the results of `shopify-build changed-files` into a comma-separated list, but adapting it to use newlines is straightforward.

AgentEnder:
Closing this out as I rebased it and fixed it up, will merge asap.

PR #28778, opened 2024-11-04
cleanup(core): use built in function to strip control chars

43081j:
Uses `node:util` instead of pulling in `strip-ansi` for stripping control characters.

we already seem to have dropped `strip-ansi` so I'm not sure what's pulling it in for this script to be currently working. presumably is just a deep dev dependency somewhere


JamesHenry (approved):
Thanks so much again!

PR #28803, opened 2024-11-05
feat(rspack): move logic for withWeb to applyWebConfig and bring in line with webpack

Coly010:
## Current Behavior

`withWeb` from `@nx/rspack` is not reflective of what `@nx/webpack` does.

## Expected Behavior

Bring `withWeb` in line with `@nx/webpack`

## Related Issue(s)

Fixes #


ndcunningham (approved):

packages/rspack/src/executors/rspack/schema.d.ts
```diff
@@ -16,13 +18,20 @@ export interface RspackExecutorSchema {
 
   rspackConfig: string;
   optimization?: boolean | OptimizationOptions;
-  sourceMap?: boolean | string;
+  sourceMap?: boolean | 'hidden';
```
ndcunningham, line 21: hmmm I think this should be a string to not restrict a type from being passed in.
For example `inline-source-map` cannot be added which is useful in dev environments. 
The webpack one needs to be updated to match this.

Coly010: Lets create a task to update the webpack one that we can circle back to

packages/rspack/src/plugins/utils/apply-web-config.ts
```diff
@@ -0,0 +1,459 @@
...
+  {
+    useNormalizedEntry,
+  }: {
+    // rspack.Configuration allows arrays to be set on a single entry
+    // rspack then normalizes them to { import: "..." } objects
+    // This option allows use to preserve existing composePlugins behavior where entry.main is an array.
+    useNormalizedEntry?: boolean;
+  } = {}
+): void {
+  if (!process.env['NX_TASK_TARGET_PROJECT']) return;
```
jaysoo, line 35: I know this was copied from the webpack code, but I think it should be checking `global.NX_GRAPH_CREATION` instead.

Coly010: I can update webpack too in that case

Coly010: updated

jaysoo (approved):

PR #28812, opened 2024-11-05
feat(rspack): update withReact plugin for parity with webpack plugin

ndcunningham:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/rspack/src/utils/lib/apply-react-config.ts
+++ b/packages/rspack/src/utils/lib/apply-react-config.ts
@@ -1,0 +1,139 @@
+import { Configuration, RspackOptionsNormalized } from '@rspack/core';
+import { SvgrOptions } from '../with-react';
+
+export function applyReactConfig(
+  options: { svgr?: boolean | SvgrOptions },
+  config: Partial<RspackOptionsNormalized | Configuration> = {}
+): void {
+  if (global.NX_GRAPH_CREATION) return;
+
+  const isDev =
+    process.env.NODE_ENV === 'development' || config.mode === 'development';
+  addReactBaseConfig(config, isDev);
+  addHotReload(config, isDev);
+
+  if (options.svgr !== false || typeof options.svgr === 'object') {
+    removeSvgLoaderIfPresent(config);
+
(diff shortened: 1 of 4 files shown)


packages/rspack/src/utils/lib/apply-react-config.ts
```diff
@@ -0,0 +1,87 @@
...
+        ...(babelLoader.options['plugins'] || []),
+        isDev && [
+          require.resolve('react-refresh/babel'),
+          { skipEnvCheck: true },
+        ],
+      ].filter(Boolean);
+    }
+  }
+
+  config.plugins = [
```
jaysoo, line 83: Can't we just do this?

```
if (isDev) {
  config.plugins.push(new ReactRefreshPlugin({ overlay: false }));
}
```

jaysoo (approved):

packages/rspack/src/utils/with-react.ts
```diff
@@ -1,77 +1,29 @@
 import { Configuration } from '@rspack/core';
-import { withWeb } from './with-web';
 import { NxRspackExecutionContext } from './config';
+import { withWeb, WithWebOptions } from './with-web';
+import { applyReactConfig } from './lib/apply-react-config';
 
-export function withReact(opts = {}) {
+export interface SvgrOptions {
```
Coly010, line 6: You'll need to move this out of here and likely into `src/plugins/utils/models.ts`

There's a circular dep now between `with-react` and `apply-react-config`

packages/rspack/src/utils/lib/apply-react-config.ts
```diff
@@ -0,0 +1,87 @@
+import { Configuration, RspackOptionsNormalized } from '@rspack/core';
+import { SvgrOptions } from '../with-react';
```
Coly010, line 2: This is part 2 of the circular dep

Coly010 (changes requested):
Just a circular dep removal and this should be fine.

Has it been tested with Standard React app and with Module Federation?

Coly010 (approved):

PR #28820, opened 2024-11-06
fix(core): jest plugin should work with nodenext set in tsconfig

ndcunningham:
## Current Behavior

When you create a new project using the new ts setup to support ts project references jest does not work out of the box due to incompatibility with the module being set inside tsconfig to `nodenext`.

## Expected Behavior

Jest should work out of the box with our current setup

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/generators/configuration/configuration.ts
+++ b/packages/jest/src/generators/configuration/configuration.ts
@@ -123,39 +123,7 @@ export async function configurationGeneratorInternal(
     await formatFiles(tree);
   }
 
-  tasks.push(getUnsupportedModuleResolutionWarningTask(tree));
-
   return runTasksInSerial(...tasks);
-}
-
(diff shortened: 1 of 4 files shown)


jaysoo:
LGTM, but we need to remove this warning as well:

```
 NX   Compiler option 'moduleResolution' in tsconfig.base.json must be 'node' or 'node10'

Jest requires 'moduleResolution' to be set to 'node' or 'node10' to work properly. It would need to be changed in the "tsconfig.base.json" file. It's not enough to override the compiler option in the project's tsconfig file.
Alternatively, you can use the environment variable `TS_NODE_COMPILER_OPTIONS='{"moduleResolution": "node10"}'` to override Jest's usage of ts-node.
```

jaysoo (approved):

PR #28824, opened 2024-11-06
fix(core): add workspaces path if package path is not included

xiongemi:
## Current Behavior

- when import to a path where is not in the workspaces, it currently just shows a warning. however, it will cause an error like "module not found" because there are packages not installed.

## Expected Behavior

- automatically add to the workspaces
Screenshot 2024-11-08 at 12 58 42 AM

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/import/import.ts
+++ b/packages/nx/src/command-line/import/import.ts
@@ -1,8 +1,6 @@
-import { dirname, isAbsolute, join, relative, resolve } from 'path';
-import { minimatch } from 'minimatch';
+import { isAbsolute, join, relative, resolve } from 'path';
 import { existsSync, promises as fsp } from 'node:fs';
 import * as chalk from 'chalk';
-import { load as yamlLoad } from '@zkochan/js-yaml';
 import { cloneFromUpstream, GitRepository } from '../../utils/git-utils';
 import { stat, mkdir, rm } from 'node:fs/promises';
 import { tmpdir } from 'tmp';
@@ -13,8 +11,10 @@ import { detectPlugins, installPlugins } from '../init/init-v2';
 import { readNxJson } from '../../config/nx-json';
 import { workspaceRoot } from '../../utils/workspace-root';
 import {
+  addPackagePathToWorkspaces,
   detectPackageManager,
   getPackageManagerCommand,
+  getPackageWorkspaces,
   isWorkspacesEnabled,
   PackageManager,
   PackageManagerCommands,
@@ -28,7 +28,7 @@ import {
   getPackagesInPackageManagerWorkspace,
   needsInstall,
 } from './utils/needs-install';
-import { readPackageJson } from '../../project-graph/file-utils';
+import { minimatch } from 'minimatch';
 
 const importRemoteName = '__tmp_nx_import__';
 
@@ -280,6 +280,13 @@ export async function importHandler(options: ImportOptions) {
     });
   }
 
+  await handleMissingWorkspacesEntry(
+    packageManager,
+    pmc,
+    relativeDestination,
+    destinationGitClient
+  );
+
   // If install fails, we should continue since the errors could be resolved later.
   let installFailed = false;
   if (plugins.length > 0) {
@@ -324,8 +331,6 @@ export async function importHandler(options: ImportOptions) {
       ],
     });
   }
-
-  await warnOnMissingWorkspacesEntry(packageManager, pmc, relativeDestination);
 
   if (source != destination) {
     output.warn({
@@ -391,13 +396,15 @@ async function createTemporaryRemote(
   await destinationGitClient.fetch(remoteName);
 }
 
-// If the user imports a project that isn't in NPM/Yarn/PNPM workspaces, then its dependencies
-// will not be installed. We should warn users and provide instructions on how to fix this.
-async function warnOnMissingWorkspacesEntry(
+/**
+ * If the user imports a project that isn't in the workspaces entry, we should add that path to the workspaces entry.
+ */
+async function handleMissingWorkspacesEntry(
   pm: PackageManager,
   pmc: PackageManagerCommands,
-  pkgPath: string
-) {
+  pkgPath: string,
+  destinationGitClient: GitRepository
+): Promise<void> {
   if (!isWorkspacesEnabled(pm, workspaceRoot)) {
     output.warn({
       title: `Missing workspaces in package.json`,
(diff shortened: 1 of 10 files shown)


packages/nx/src/utils/package-manager.ts
```diff
@@ -478,3 +481,76 @@ export async function packageRegistryPack(
...
+    const updatedPackageJson = {
+      ...packageJson,
+      workspaces,
+    };
+    const packageJsonPath = join(root, 'package.json');
+    writeJsonFile(packageJsonPath, updatedPackageJson);
+  } else if (packageManager === 'pnpm') {
+    const yamlPath = join(root, 'pnpm-workspace.yaml');
+    let yaml: any = {};
+    if (existsSync(yamlPath)) {
```
jaysoo, line 549: Would this preserve YAML comments? I haven't tried it yet.

packages/nx/src/command-line/import/import.ts
```diff
@@ -391,76 +392,66 @@ async function createTemporaryRemote(
...
+  let added = false;
+  try {
+    added = addPackagePathToWorkspaces(pm, workspaceRoot, pkgPath);
+  } catch {
+    return;
+  }
+  if (!added) {
+    return;
+  }
+  await destinationGitClient.amendCommit();
```
FrozenPandaz, line 412: I think this should all be under the `isWorkspacesEnabled` section right?

packages/nx/src/command-line/import/import.ts
```diff
@@ -391,76 +392,66 @@ async function createTemporaryRemote(
...
           ? [
               `We recommend enabling Bun workspaces to install dependencies for the imported project.`,
-              `Add \`"workspaces": ["${pkgPath}"]\` to package.json and run "${pmc.install}".`,
+              `Added \`"workspaces": ["${pkgPath}"]\` to package.json.`,
               `See: https://bun.sh/docs/install/workspaces`,
             ]
           : [
               `We recommend enabling PNPM workspaces to install dependencies for the imported project.`,
-              `Add the following entry to to pnpm-workspace.yaml and run "${pmc.install}":`,
+              `Added the following entry to to pnpm-workspace.yaml:`,
```
FrozenPandaz, line 437: I don't think this should change. If someone doesn't have `npm workspaces` setup, we shouldn't enable it for them. We should keep it as is and tell them to enable it altogether. If someone does have `npm workspaces` enabled, then that's when we should add it for them.

packages/nx/src/command-line/import/import.ts
```diff
@@ -391,76 +392,66 @@ async function createTemporaryRemote(
...
+      bodyLines:
+        pm === 'npm' || pm === 'yarn' || pm === 'bun'
+          ? [
+              `The imported project (${pkgPath}) is missing the "workspaces" field in package.json.`,
+              `Added "${pkgPath}" to workspaces.`,
+            ]
+          : [
+              `The imported project (${pkgPath}) is missing the "packages" field in pnpm-workspaces.yaml.`,
+              `Added "${pkgPath}" to packages.`,
+            ],
```
FrozenPandaz, line 454: Can we make this a success message instead of a warning? It should tell people that nx has successfully added the path to the workspaces/packages field.

FrozenPandaz (changes requested):

packages/nx/src/command-line/import/import.ts
xiongemi: changed

packages/nx/src/command-line/import/import.ts
xiongemi: changed back

packages/nx/src/command-line/import/import.ts
xiongemi: changed

packages/nx/src/utils/package-manager.ts
xiongemi: changed the code, it now preserves comments

packages/nx/src/command-line/import/import.ts
```diff
@@ -280,21 +284,44 @@ export async function importHandler(options: ImportOptions) {
...
   // If install fails, we should continue since the errors could be resolved later.
   let installFailed = false;
   if (plugins.length > 0) {
+    output.log({ title: 'Installing Plugins' });
     try {
-      output.log({ title: 'Installing Plugins' });
-      installPlugins(workspaceRoot, plugins, pmc, updatePackageScripts);
-
-      await destinationGitClient.amendCommit();
+      runPackageManagerInstallPlugins(workspaceRoot, pmc, plugins);
```
FrozenPandaz, line 299: This is unrelated.

packages/nx/src/command-line/import/import.ts
```diff
@@ -428,39 +455,47 @@ async function warnOnMissingWorkspacesEntry(
...
-      const packageJson = readPackageJson();
-      workspaces = packageJson.workspaces;
-    } else if (pm === 'pnpm') {
-      const yamlPath = join(workspaceRoot, 'pnpm-workspace.yaml');
-      if (existsSync(yamlPath)) {
-        const yamlContent = await fsp.readFile(yamlPath, 'utf-8');
-        const yaml = yamlLoad(yamlContent);
-        workspaces = yaml.packages;
-      }
+    let workspaces: string[] = getPackageWorkspaces(pm, workspaceRoot) ?? [];
```
FrozenPandaz, line 458: getPackageWorkspaces should at least return an empty array if workspaces is indeed enabled.

packages/nx/src/command-line/import/import.ts
```diff
@@ -428,39 +455,47 @@ async function warnOnMissingWorkspacesEntry(
...
-                  `Add "${pkgsDir}/*" to packages run "${pmc.install}".`,
-                ],
-        });
-      }
+    let added = false;
+    try {
+      added = addPackagePathToWorkspaces(pm, workspaceRoot, pkgPath);
+    } catch {
+      // ignore error
+    }
```
FrozenPandaz, line 469: Why would this error?

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -36,33 +37,73 @@ export interface InitArgs {
   integrated?: boolean; // For Angular projects only
 }
 
-export function installPlugins(
+export function runPackageManagerInstallPlugins(
```
FrozenPandaz, line 40: This is unrelated

packages/nx/src/utils/package-manager.ts
```diff
@@ -478,3 +494,120 @@ export async function packageRegistryPack(
...
+    packageManager === 'bun'
+  ) {
+    const packageJson = readPackageJson(root);
+    workspaces = packageJson.workspaces;
+  } else if (packageManager === 'pnpm') {
+    const pnpmWorkspacePath = join(root, 'pnpm-workspace.yaml');
+    if (existsSync(pnpmWorkspacePath)) {
+      const { packages } =
+        readYamlFile<{ packages: string[] }>(pnpmWorkspacePath) ?? {};
+      workspaces = packages;
```
FrozenPandaz, line 520:
```suggestion
      workspaces = packages ?? [];
```

packages/nx/src/utils/package-manager.ts
```diff
@@ -478,3 +494,120 @@ export async function packageRegistryPack(
...
+export function addPackagePathToWorkspaces(
+  packageManager: PackageManager = detectPackageManager(),
+  root: string = workspaceRoot,
+  packagePath: string
+): boolean {
+  let workspaces: string[] = getPackageWorkspaces(packageManager, root) ?? [];
+  const isPkgIncluded = workspaces.some((w) => minimatch(packagePath, w));
+  if (isPkgIncluded) {
+    return false;
+  }
```
FrozenPandaz, line 544: Can we not check this.. and always add it? It's up to the user of this function to check if it already exists.And then this function does not need to return a boolean.

packages/nx/src/utils/package-manager.ts
```diff
@@ -478,3 +494,120 @@ export async function packageRegistryPack(
...
+      writeFileSync(
+        pnpmWorkspacePath,
+        YAMLStringify({
+          packages: [packagePath],
+        })
+      );
+    }
+    return true;
+  }
+  return false;
```
FrozenPandaz, line 612: All the package managers are handled right?

FrozenPandaz (changes requested):

packages/nx/src/command-line/init/init-v2.ts
xiongemi: removed

packages/nx/src/utils/package-manager.ts
xiongemi: changed

FrozenPandaz (approved):

PR #28825, opened 2024-11-06
feat(rspack): move logic for withNx to applyBaseConfig and bring in line with webpack

Coly010:
## Current Behavior

`withNx` from `@nx/rspack` is not reflective of what `@nx/webpack` does.

## Expected Behavior

Bring `withNx` in line with `@nx/webpack`


packages/rspack/src/plugins/utils/apply-base-config.ts
```diff
@@ -0,0 +1,418 @@
...
+              syntax: 'typescript',
+              decorators: true,
+              tsx: true,
+            },
+            transform: {
+              react: {
+                pragma: 'React.createElement',
+                pragmaFrag: 'React.Fragment',
+                throwIfNamespace: true,
+                development: false,
```
ndcunningham, line 404: Before, I think this was determined based on `NODE_ENV` or `options.mode` should this still be the case?

Coly010: Hmm 🤔 good spot, yes likely

ndcunningham (approved):

packages/rspack/src/utils/config.ts
```diff
@@ -90,13 +90,15 @@ function ensureNxRspackExecutionContext(ctx: NxRspackExecutionContext): void {
   const configurationName = process.env.NX_TASK_TARGET_CONFIGURATION;
   const projectGraph = readCachedProjectGraph();
   const projectNode = projectGraph.nodes[projectName];
+  console.log('>>>DBG>>>', projectNode, projectName, projectNode.data.root);
```
jaysoo, line 93: Remove

jaysoo (approved):

PR #28828, opened 2024-11-07
feat(nx-dev): version picker

isaacplmann:
Adds a version picker to the footer of nx.dev


nx-dev/ui-common/src/lib/selector.tsx
```diff
@@ -0,0 +1,77 @@
...
+                {props.items.map((item, personIdx) => (
+                  <Listbox.Option
+                    key={personIdx}
+                    className={({ active }) =>
+                      `relative cursor-pointer select-none list-none px-3 py-2 hover:bg-slate-50 dark:hover:bg-slate-800`
+                    }
+                    value={item}
+                  >
+                    {({ selected, active }) => (
+                      <span className={'block truncate font-medium'}>
```
ndcunningham, line 64: Could this be an anchor tag instead?

nx-dev/ui-common/src/lib/selector.tsx
```diff
@@ -0,0 +1,77 @@
...
+              enterTo="transform scale-100 opacity-100"
+              leave="transition duration-75 ease-out"
+              leaveFrom="transform scale-100 opacity-100"
+              leaveTo="transform scale-95 opacity-0"
+            >
+              <Listbox.Options
+                static
+                className="absolute z-10 mt-1 max-h-60 w-full overflow-auto rounded-sm bg-white py-1 pl-0 text-base shadow-md focus:outline-none sm:text-sm dark:bg-slate-800/60 dark:focus-within:ring-sky-500"
+              >
+                {props.items.map((item, personIdx) => (
```
ndcunningham, line 55: could this be renamed to `idx` instead of `personIdx`?

juristr (changes requested):
Couple of comments

- can we stay on the same page when switching version? If the page is not available, ideally we'd just go to /docs if possible (it just happens that we version the entire nx.dev website, but it's the docs which is what the user is interested in
- can we backport this to `website-19` etc? otherwise you switch to v19 and there's no version switcher
- side-note: having the version switcher in the footer is not really super visible. maybe that was intentional

isaacplmann:
Cleaned up the code.

- If we stay on the same page when you switch versions, you could get a 404.  I updated the link so that it goes to `/docs` instead.
- Yes, I can back port it to the older versions after this merges.
- Yes, the version switcher is intentionally in the footer.  It isn't a high frequency action. Once someone knows that its there, they'll just always open the version of the docs that they're on.  For most people, it just gets in the way if it is at the top.

juristr (approved):

PR #28854, opened 2024-11-08
fix(core): set default branch to main if no git

xiongemi:
## Related Issue(s)

Fixes #


packages/nx/src/command-line/init/implementation/utils.ts
```diff
@@ -65,6 +65,15 @@ export function createNxJsonFile(
 }
 
 function deduceDefaultBase() {
+  try {
```
FrozenPandaz, line 68: We shouldn't have to run a different command to make sure git doesn't exist...

Can we check the error some how?

We're running git in the line below.. that will error if there's no git... can we somehow check the first error to see if it says that `git` is not installed?

Also it seems like the final `catch` here says to return `master`... so maybe this should return `master` too? :eyes: I don't know if now is the time to change that return to `main`

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #28859, opened 2024-11-08
fix(core): do not use unix dotfile locks

FrozenPandaz:
## Current Behavior

The unix dotfiles are incompatible with posix dotfiles. Having the 2 modes being switched between yields weird errors. 

## Expected Behavior

Posix dotfiles seem fine for what we're doing so Nx will stick with that mode for all cases.

## Related Issue(s)

Fixes #


Cammisuli (approved):

Cammisuli:
Tested on angular-eslint and a simplified ngrx build, and things are working there. 👍🏼

PR #28863, opened 2024-11-09
fix(graph): enable action on the ViewOnGraph button regardless of the…

Ahmed-Hakeem:
## Current Behavior
ViewOnGraph button doesn't work until you open the accordion

## Expected Behavior
ViewOnGraph button should direct to graph without the need to open the accordion first.

I believe it's better from the UX prespective that user shouldn't click twice on the same button to fire its action.


MaxKless (approved):
works well, thanks :)

PR #28913, opened 2024-11-13
feat(rspack): update executor to be in line with webpack

ndcunningham:
This PR brings the rspack executor `@nx/rspack:rspack` inline with webpack.

It also prepares the executor to be used with the soon to be implemented `NxRspackAppPlugin` so that we can support executor and inferred targets.


Coly010 (approved):
Approved, but there's a HMR issue when `styles` are set. I don't know if related to this PR or not, but it might be worth solving it now as this is being worked on

PR #28925, opened 2024-11-13
docs(core): add information for s3 compatible providers

Cammisuli:
## Related Issue(s)

Fixes #

### DIFF
--- a/docs/external-generated/packages/powerpack-s3-cache/documents/overview.md
+++ b/docs/external-generated/packages/powerpack-s3-cache/documents/overview.md
@@ -107,6 +107,10 @@ Storing your credentials in the `nx.json` file is the least secure of the 4 auth
 
 ### 3. Configure S3 Cache
 
+{% callout title="S3 compatible providers" %}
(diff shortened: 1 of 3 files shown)


docs/external-generated/packages/powerpack-s3-cache/documents/overview.md
```diff
@@ -107,6 +107,10 @@ Storing your credentials in the `nx.json` file is the least secure of the 4 auth
 
 ### 3. Configure S3 Cache
 
+{% callout title="S3 compatible providers" %}
+To use `@nx/powerpack-s3-cache` with S3 compatible providers, `endpoint` will need to be provided. Some providers also need to have `forcePathStyle` set to true in the configuration. Refer to your provider's documentation to see if this setting is needed for your instance.
+{% /callout %}
```
FrozenPandaz, line 112: This seems weird as the first thing. Should we move it down in this section?

I think we also add an explicit example of how to set it up with MinIO.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #28998, opened 2024-11-19
fix(js): only add typescript project references for explicit dependencies in sync generator

### LINKED ISSUE #28997: [sync] typescript-sync generator adds project references for implicit dependencies
### Current Behavior

The typescript sync generator adds references for all projects in the dependency graph if they have a tsconfig (with `composite: true`).

This currently includes implicit dependencies, which are not explicitly referenced in code and therefore don't require project references.

See #28839 - where batch builds also skip implicit dependencies.

For a tangible example, if I have a monorepo with some infrastructure project (eg CDK), and a website project (eg a React app) - I might have my infrastructure project deploy my website but not directly reference any of its code, and so use an implicit dependency to ensure build order. The sync generator will still add a reference as both of these are TypeScript projects, but I can get an error such as this when `moduleResolution` is is different for the two projects:

```
      ../website/tsconfig.app.json:3:3 - error TS5095: Option 'bundler' can only be used when 'module' is set to 'preserve' or to 'es2015' or later.
      
      3   "compilerOptions": {
          ~~~~~~~~~~~~~~~~~
      
      ../website/tsconfig.app.json:3:3 - error TS5109: Option 'moduleResolution' must be set to 'NodeNext' (or left unspecified) when option 'module' is set to 'NodeNext'.
      
      3   "compilerOptions": {
          ~~~~~~~~~~~~~~~~~
```

### Expected Behavior

Implicit dependency tsconfigs should not be referenced by the projects which depend on them. Only explicit dependencies should be referenced in project tsconfigs.

### GitHub Repo

https://github.com/cogwirrel/nx-sync-generator-implicit-deps-example

### Steps to Reproduce

1. Clone the [example repo](https://github.com/cogwirrel/nx-sync-generator-implicit-deps-example)
2. `pnpm i`
3. `pnpm nx run-many --target build --all` and observe error

If you then delete the references to the website in `packages/infra/tsconfig.json` and `packages/infra/tsconfig.lib.json` and rebuild, Nx prompts to run the sync generator. If we skip running the sync generator, the build succeeds, otherwise if we run the sync generator the build fails.

### Nx Report

```shell

Node           : 20.18.0
OS             : darwin-arm64
pnpm           : 9.13.0

nx             : 20.1.2
@nx/js         : 20.1.2
@nx/eslint     : 20.1.1
@nx/workspace  : 20.1.2
@nx/devkit     : 20.1.2
@nx/react      : 20.1.1
@nx/vite       : 20.1.2
@nx/web        : 20.1.1
typescript     : 5.5.4
Registered Plugins:
@nx/js/typescript
@nx/vite/plugin
  - @nx/eslint@20.1.1
  - @nx/react@20.1.1
  - @nx/web@20.1.1

To fix this, run `nx migrate nx@20.1.2`
```

### Failure Logs

```shell
Failure logs will vary depending on the use case, but in the example code they are as follows:


      ../website/tsconfig.app.json:3:3 - error TS5095: Option 'bundler' can only be used when 'module' is set to 'preserve' or to 'es2015' or later.
      
      3   "compilerOptions": {
          ~~~~~~~~~~~~~~~~~
      
      ../website/tsconfig.app.json:3:3 - error TS5109: Option 'moduleResolution' must be set to 'NodeNext' (or left unspecified) when option 'module' is set to 'NodeNext'.
      
      3   "compilerOptions": {
          ~~~~~~~~~~~~~~~~~
```

### Package Manager Version

pnpm 9.13.0

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

cogwirrel:
This change omits references to implicit dependency tsconfigs for typescript projects in the sync generator, since given that they are not referenced directly in code there is no need for project references.

## Current Behavior
TypeScript sync generator adds references to any dependency project which has a tsconfig (where composite is true), including implicit dependencies where these references are unnecessary and can potentially cause build failures.

See [example repo](https://github.com/cogwirrel/nx-sync-generator-implicit-deps-example).

## Expected Behavior
Only explicit dependencies should be referenced in tsconfigs.

## Related Issue(s)
Fixes #28997


adrian-operatorai:
@Coly010  @leosvelperez  - Any updates on this one? I am currently blocked on upgrading NX due to this :(

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -465,13 +465,22 @@ function collectProjectDependencies(
...
     const targetProjectNode = projectGraph.nodes[dep.target];
     if (!targetProjectNode) {
       // It's an npm dependency
       continue;
     }
 
+    if (implicitDependencies.has(targetProjectNode.name)) {
+      // It's an implicit dependency and therefore not explicitly referenced in code
+      continue;
+    }
```
leosvelperez, line 482: A project could potentially have a static/dynamic dependency on another project and also have it as an implicit dependency. While we could argue that's a bad user configuration, it's currently a possible/valid configuration. In that case, we still want to process the target project if it's also defined as a static/dynamic dependency.
```suggestion
  for (const dep of projectGraph.dependencies[projectName]) {
    const targetProjectNode = projectGraph.nodes[dep.target];
    if (!targetProjectNode || dep.type === 'implicit') {
      // It's an npm or an implicit dependency
      continue;
    }
```

leosvelperez (changes requested):

packages/js/src/generators/typescript-sync/typescript-sync.ts
cogwirrel: @leosvelperez thanks for the review! Great callout - have committed your suggestion :)

leosvelperez (approved):
Thank you!

jaysoo (approved):

PR #29014, opened 2024-11-20
fix(angular): make scam-to-standalone replace correct module

Olster:
## Current Behavior
Without this fix, the regular expression would replace other modules that have old module name in their name.  
E.g. `MapIconModule` contains `IconModule` and when migrating `IconModule`, the migration would wrongfully update `MapIconModule` to `MapIconComponent`.

## Expected Behavior
Migration changes only the Module/Component that's being migrated.


leosvelperez (approved):
Thank you!

PR #29017, opened 2024-11-21
fix(core): exponential retry on cache put fail

mike-fam:
Fix exponential retries so that it behaves according to the comments.

The current comments and implementation of this exponential backoff are not behaving correctly, causing issues when many (~100) parallel tasks access the same local cache on a Network Drive. I've updated it to have the correct functionality while maintaining the maximum allowable time.

## Current Behavior
The current exponential backoff retry logic only allows up to 1364 ms (4 + 4² + 4³ + 4⁴ + 4⁵), while the comment allows up to 20480 ms, albeit with confusing maths.

## Expected Behavior
Exponential backoff retry is run correctly with the maximum allowable time of 20480 ms

## Related Issue(s)
#21926

Fixes #21926


mike-fam:
I'm not sure if the failed command (`nx documentation --no-dte`) is due to my change. It works fine locally for me.

mike-fam:
@AgentEnder Please let me know if there's something I can do to improve this PR or make the builds pass.

Thanks

mike-fam:
@AgentEnder @Cammisuli Please take a look and give feedback if need be. Thank you 🙏 🙏

Cammisuli (approved):

PR #29021, opened 2024-11-21
feat(module-federation): move withModuleFederation for angular to new package

Coly010:
## Current Behavior

The `withModuleFederation` helper currently lives in the `@nx/angular` package.
With the goal of consolidating the module federation support into a single package, this introduces a divergence in where module-federation support lies

## Expected Behavior

Move `withModuleFederation` helper for angular into the `@nx/module-federation` package, exposed via `@nx/module-federation/angular`.
Adds a migration to migrate existing projects to use the new package

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/module-federation/index.ts
+++ b/packages/angular/module-federation/index.ts
@@ -1,2 +1,9 @@
-export { withModuleFederation } from '../src/utils/mf/with-module-federation';
-export { withModuleFederationForSSR } from '../src/utils/mf/with-module-federation-ssr';
+import {
+  withModuleFederation,
+  withModuleFederationForSSR,
+} from '@nx/module-federation/angular';
+
+/**
+ * @deprecated Update the import path to `@nx/module-federation/angular` instead. This import path will be removed in Nx v22.
+ */
+export { withModuleFederation, withModuleFederationForSSR };
--- a/packages/angular/src/migrations/update-20-2-0/migrate-with-mf-import-to-new-package.spec.ts
+++ b/packages/angular/src/migrations/update-20-2-0/migrate-with-mf-import-to-new-package.spec.ts
@@ -1,0 +1,138 @@
+import { type Tree } from '@nx/devkit';
+import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
+import migrateWithMfImport from './migrate-with-mf-import-to-new-package';
+
+describe('migrate-with-mf-import-to-new-package', () => {
+  let tree: Tree;
+
+  beforeEach(() => {
+    tree = createTreeWithEmptyWorkspace();
+    tree.write(
+      'apps/shell/project.json',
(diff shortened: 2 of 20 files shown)


packages/angular/src/migrations/update-20-2-0/migrate-with-mf-import-to-new-package.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+      // ASSERT
+      expect(tree.read('apps/shell/webpack.config.ts', 'utf-8'))
+        .toMatchInlineSnapshot(`
+        "import { withModuleFederation } from '@nx/module-federation/angular';
+        "
+      `);
+    });
+
+    it('should migrate the require correctly for withMfSSR even run twice', async () => {
```
jaysoo, line 119: Nit: `even when run twice`, or the migration should be idempotent.

packages/angular/module-federation/index.ts
```diff
@@ -1,2 +0,0 @@
-export { withModuleFederation } from '../src/utils/mf/with-module-federation';
```
jaysoo, line 1: Should do still support the exports, and add deprecation notice for v22?

It's possible that the migration doesn't work 100%.

packages/angular/src/migrations/update-20-2-0/migrate-with-mf-import-to-new-package.ts
```diff
@@ -0,0 +1,57 @@
...
+      projects.add(projectConfig.root);
+    }
+  );
+
+  for (const projectRoot of projects) {
+    visitNotIgnoredFiles(tree, projectRoot, (filePath) => {
+      if (!filePath.endsWith('.ts') && !filePath.endsWith('.js')) {
+        return;
+      }
+      let contents = tree.read(filePath, 'utf-8');
```
jaysoo, line 31: Should we also guard the file name, so we don't read in all TS/JS files in the project?

Coly010: I considered it. But it could be an arbitrary string. 

We could pull the webpack config path that's passed to the executor, but we'd also need to check every possibly configuration they have for that option too.

it's a complexity vs perf benefit.

We're already scoping the files read to TS/JS under the project root for projects using custom webpack configs in angular. I'm not sure if we need another level of optimization

packages/angular/src/migrations/update-20-2-0/migrate-with-mf-import-to-new-package.spec.ts
Coly010: yeah there's a describe block around those called `idempotent`
I'll update the grammar

jaysoo (approved):

PR #29026, opened 2024-11-21
fix(core): replace red in console with blue

### LINKED ISSUE #19901: Ability to configure the colours used by nx run-many (to avoid red)

## Description

When calling `nx run-many` it helpfully shows each project script in a different colour.  It would be good if we could configure the colours this uses - or at the very least disable red from being used.

## Motivation

Red is the colour associated with errors - whenever I see this in my console I immediately think there is a problem.  This unnecessarily grabs my attention when nx has chosen to use red to colour-code one of my projects:

image

## Suggested Implementation

Give the option to define a list of colours, or even colours for each specific project so it's consistent on different machines.

## Alternate Implementations

Option to disable red colours from being used.

hisuwh:
## Current Behavior

Red is used as a colour when running multiple processes in the command line.  This causes confusion as red is universally the colour for danger/errors.

## Expected Behavior

Red should not be a colour in the console and blue should be used instead.

## Related Issue(s)

Fixes #19901

### DIFF
--- a/packages/nx/src/tasks-runner/running-tasks/node-child-process.ts
+++ b/packages/nx/src/tasks-runner/running-tasks/node-child-process.ts
@@ -113,8 +113,8 @@ function addPrefixTransformer(prefix?: string) {
 const colors = [
   chalk.green,
   chalk.greenBright,
-  chalk.red,
-  chalk.redBright,
+  chalk.blue,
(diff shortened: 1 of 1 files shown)


FrozenPandaz (approved):

PR #29061, opened 2024-11-25
fix(testing): add exports config for cypress plugin

barbados-clemens:
## Current Behavior

unable to import the cypress preset with mjs cypress configs and yarn PnP

## Expected Behavior

able to import the cypress preset when using mjs cypress configs and yarn PnP

## Related Issue(s)

BREAKING CHANGE: export configuration change for @nx/cypress plugin
technically shouldn't be breaking since it re-exports all existing imports, but honestly you never know with exports 😅

### DIFF
--- a/packages/cypress/package.json
+++ b/packages/cypress/package.json
@@ -23,6 +23,18 @@
   ],
   "main": "./index",
   "typings": "./index.d.ts",
+  "exports": {
+    ".": "./index.js",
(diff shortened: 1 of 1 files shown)


barbados-clemens:
oops. just needed to add extension 🤦

```js
import { nxE2EPreset } from '@nx/cypress/plugins/cypress-preset.js';
```

PR #29062, opened 2024-11-25
chore(repo): use nx-cloud conformance

JamesHenry:


isaacplmann (approved):

.circleci/config.yml
```diff
@@ -114,6 +114,12 @@ jobs:
             pnpm nx-cloud record -- nx format:check --base=$NX_BASE --head=$NX_HEAD &
             pids+=($!)
 
+            pnpm nx-cloud record -- nx sync:check
+            pids+=($!)
+
```
FrozenPandaz, line 119: I don't think we need this until we get owners right?

FrozenPandaz (changes requested):
Shouldn't we add some rules as well?

FrozenPandaz (approved):

PR #29075, opened 2024-11-26
fix(module-federation): migration does not handle external nodes and errors

Coly010:
## Current Behavior
The React + Angular migrations intended to update the path for the `ModuleFederationConfig` imports in webpack and rspack config files will fail on externalNodes in the project graph that have `@nx/webpack` or `@nx/rspack` listed as a dependency.

## Expected Behavior
If the dependency is discovered in an `externalNode` we should skip that node, instead of continuing with the migration.

### DIFF
--- a/packages/angular/src/migrations/update-20-2-0/migrate-mf-imports-to-new-package.ts
+++ b/packages/angular/src/migrations/update-20-2-0/migrate-mf-imports-to-new-package.ts
@@ -17,6 +17,9 @@ export default async function migrateMfImportsToNewPackage(tree: Tree) {
       (dep) => dep.target === 'npm:@nx/webpack'
     );
     if (usesNxWebpack) {
+      if (graph.externalNodes[project]) {
+        continue;
+      }
       const root = graph.nodes[project].data.root;
       rootsToCheck.add(root);
     }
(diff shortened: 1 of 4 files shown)


packages/angular/src/migrations/update-20-2-0/migrate-mf-imports-to-new-package.ts
```diff
@@ -17,6 +17,9 @@ export default async function migrateMfImportsToNewPackage(tree: Tree) {
       (dep) => dep.target === 'npm:@nx/webpack'
     );
     if (usesNxWebpack) {
+      if (!graph.nodes[project]) {
+        continue;
+      }
```
leosvelperez, line 22: nit: we could do this check at the start of the loop to avoid searching for deps

leosvelperez (approved):

packages/angular/src/migrations/update-20-2-0/migrate-mf-imports-to-new-package.ts
Coly010: v.good point

ndcunningham (approved):

PR #29077, opened 2024-11-26
fix(misc): update artifact generator option descriptions and cleanup leftovers

leosvelperez:
- Update artifact generator schemas:
  - Clarify `path` is the artifact file path relative to the current working directory
  - Clarify `name` is the artifact symbol name
  - Remove prompt for `name` and remove it from the important options (won't be displayed by default in Nx Console generation UI, it will be part of the collapsed options) given that most of the time, it's meant to match the filename (last segment of the `path`)
  - Remove some leftover options related to the name and path formats that were previously missed
- Fix an issue with NestJS generators
- Fix an issue with Next `page` generator

## Related Issue(s)

Fixes #


packages/angular/src/generators/component/schema.json
```diff
@@ -9,17 +9,16 @@
   "properties": {
     "path": {
       "type": "string",
-      "description": "The path at which to create the component file, relative to the workspace root. By default, it is set to the root of the project.",
+      "description": "The file path to the component without the file extension and suffix (e.g. `my-app/src/app/my-component/my-component` generates the file `my-app/src/app/my-component/my-component.component.ts`). Relative to the current working directory.",
```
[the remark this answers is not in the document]
jaysoo: The description should be shorter, and the behavior should follow the principle of least surprise.

packages/angular/src/generators/component/schema.json
```diff
@@ -9,17 +9,16 @@
...
       "$default": {
         "$source": "argv",
         "index": 0
       },
-      "x-prompt": "Where to create the component?"
+      "x-prompt": "What is the component file path?"
```
jaysoo, line 17: `Where to generate the component?` ?

packages/angular/src/generators/component/schema.json
```diff
@@ -9,17 +9,16 @@
...
     },
     "name": {
       "type": "string",
-      "description": "The name of the component.",
-      "x-prompt": "What name would you like to use for the component?"
+      "description": "The component symbol name. It not provided, it defaults to the last segment of the provided `path`."
```
jaysoo, line 21: `The component symbol name.Defaults to the last segment of the file path.

packages/angular/src/generators/component/schema.json
leosvelperez: I considered that one, but it is less explicit about the fact that the option is meant to be a file path. I find it more ambiguous. While the current one explicitly asks for a file path.

Coly010 (approved):

jaysoo (approved):

PR #29078, opened 2024-11-26
chore(repo): add conformance rule for our package.json files

JamesHenry:


package.json
```diff
@@ -50,9 +50,9 @@
     "@eslint/eslintrc": "^2.1.1",
     "@eslint/js": "^8.48.0",
     "@floating-ui/react": "0.26.6",
-    "@jest/reporters": "^29.4.1",
-    "@jest/test-result": "^29.4.1",
-    "@jest/types": "^29.4.1",
+    "@jest/reporters": "29.7.0",
```
JamesHenry, line 53: Our jest packages weren't aligned for some reason

tools/workspace-plugin/jest.config.ts
```diff
@@ -1,7 +1,8 @@
 /* eslint-disable */
 export default {
   displayName: 'workspace-plugin',
-  preset: '../../jest.preset.js',
+  // TODO: For some reason our patched jest resolve cannot work with @nx/powerpack-conformance
```
JamesHenry, line 4: Needs some time spent on it, there's a lot of custom complexity there, the default resolver works fine...

packages/rspack/package.json
```diff
@@ -60,5 +60,8 @@
   },
   "nx-migrations": {
     "migrations": "./migrations.json"
+  },
+  "publishConfig": {
```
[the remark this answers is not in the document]
FrozenPandaz: I thought we needed this for provenance... how does rspack have provenance without this?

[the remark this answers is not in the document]
JamesHenry: It's not for provenance, it's needed because the default access for new scoped package is restricted and so the first time we publish a new package we need this explicit config. Technically after that point we no longer need it anymore, but it doesn't do any harm, just makes its state explicit, and so having this be a rule for all package.json files means we won't run into any issues when new packages are added in future.

tools/workspace-plugin/jest.config.ts
JamesHenry: Looks like this is blocking the PR after all because the existing code in this library needs the custom resolver, damn

JamesHenry: Found it, the custom resolver special cased all `@nx/` packages, but we don't want it to do that with `@nx/powerpack-` ones, so I adjusted that in https://github.com/nrwl/nx/pull/29078/commits/6bf90638fc40823c39be8c3e3055d2d63e62c4bd

jaysoo (approved):

PR #29092, opened 2024-11-27
fix(core): improve local plugin loading by reading only tsconfig comp

jogelin:
## Current Behavior

When loading local plugins, Nx compiles TypeScript code on the fly. To achieve this, Nx uses the TypeScript function parseJsonConfigFileContent, which processes the entire configuration.

In large repositories, this process can take more than 10 seconds per plugin. Even though the process is parallelized, it significantly impacts performance when generating the graph.

## Expected Behavior

Loading local plugins should be fast to allow immediate graph generation. Since only the compiler options are needed, we can read the root tsconfig file directly.

With this approach, loading a plugin now takes just 200ms, even for large monorepos.

### DIFF
--- a/packages/nx/src/plugins/js/utils/register.ts
+++ b/packages/nx/src/plugins/js/utils/register.ts
@@ -293,8 +293,7 @@ export function getTranspiler(
  * @returns cleanup method
  */
 export function registerTranspiler(
-  compilerOptions: CompilerOptions,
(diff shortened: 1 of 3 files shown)


leosvelperez:
Thanks, @jogelin! Sorry for not getting to this earlier. This has changed a bit since then, and the issue was fixed by https://github.com/nrwl/nx/pull/29774.

PR #29130, opened 2024-11-30
test(expo): add breaking E2E test for an Expo module that includes post-installation output

hassankhan:
## Current Behavior

After running `nx run app:install expo-router`, nx updates the root `package.json` but fails before updating `app/package.json`.

## Expected Behavior

After running `nx run app:install expo-router`, nx should update both the root `package.json` AND `app/package.json`.

## Related Issue(s)

Relates to #29131

### DIFF
--- a/e2e/expo/src/expo.test.ts
+++ b/e2e/expo/src/expo.test.ts
@@ -137,6 +137,29 @@ describe('@nx/expo', () => {
       },
     });
   });
+  
+  it('should install packages with post-installation output', async () => {
+    // run install command
+    let installResults = await runCLIAsync(
+      `install ${appName} --force --no-interactive`
+    );
+    expect(installResults.combinedOutput).toContain(
+      'Successfully ran target install'
+    );
+
+    installResults = await runCLIAsync(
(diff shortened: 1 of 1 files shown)


xiongemi:
i rerun your e2e test, it passed locally.
there is already an e2e test to expo install: https://github.com/nrwl/nx/blob/cbfc6fe97f32cacb5fe380a9cc423a5d73948531/e2e/expo/src/expo.test.ts#L117

hassankhan:
Hi @xiongemi, in previous versions of Nx, installing any Expo module with a config plugin would print a warning message and exit with a nonzero code:

```
No Expo config was found. Please create an Expo config (app.json or app.config.js) in your project root.
Please add the following to your Expo config

{
  "plugins": [
    "expo-font"
  ]
}
```

It seems like this message (and the associated return value) no longer show up in the latest version of Nx. 

I do think this is a slight regression, since previously at least users were informed that a plugin had to be added. Ideally, `@nx/expo` would be able to pass the correct `app.json` along to `expo install`.

PR #29138, opened 2024-12-02
cleanup(core): migrate to nanospinner

43081j:
Migrates from `ora` to `nanospinner`, a much lighter package.

this should roughly be a drop in replacement and will be faster/lighter


JamesHenry:
@43081j I was thinking we would go for `nanospinner` for this, but found an issue that I linked to from the umbrella cleanup issue: https://github.com/es-tooling/ecosystem-cleanup/issues/117#issuecomment-2456690935

Please could you confirm this library does not have this issue? (Should be able to modify the linked stackblitz) and how the two libraries compare?

43081j:
I'll take a look when I can 👍

The two libraries are pretty well aligned, so it comes mostly down to personal preference. picospinner is a bit more customisable

Both maintainers are active in our community too, so you really could just flip a coin here 😅

JamesHenry (changes requested):
@43081j I added picospinner to my comparison with ora and nanospinner:
https://stackblitz.com/edit/stackblitz-starters-dmnx9n?file=package.json&view=editor

It seems like it doesn't actually allow for the error handling at all. Maybe it is calling process.exit itself?

- ora: works as expected but bloated
- nanospinner: has double printing issue on exit
- picospinner: does not allow for exceptions to be handled

We cannot proceed until one of `nanospinner` or `picospinner` matches the functionality/robustness of ora I'm afraid

43081j:
All good, I'll look into this today or tomorrow 👍

I'm mobile at the minute so can't dig into it but I'll figure it out once I'm at a laptop

JamesHenry:
Hi @43081j, 

Following up on this one:

With the latest versions of `picospinner` and `nanospinner`, it's `nanospinner` that seems to best match `ora` for our usage:

image

It's because https://github.com/usmanyunusov/nanospinner/issues/32 was explicitly addressed.

https://github.com/tinylibs/picospinner/pull/5 remains in limbo somewhat. I think we could probably press ahead with `nanospinner` for now,  but remain open to `picospinner` in future?

43081j:
its been a while so i forget what the disagreement was, but as far as i remember, it was because none of these spinners should really allow output while they're enabled

when the spinner is enabled, you wouldn't expect output to occur (because we're outputting the spinner frames).

errors are no different and shouldn't output either.

picospinner seems to be doing the right thing here and we're wrong for not stopping the spinner before outputting errors.

however, that does make your try/catch a little more verbose, so i understand why ora and nanospinner would add special behaviour for it.

we should go with nanospinner for that reason

43081j:
i've rebased and updated to use nanospinner 👍

JamesHenry:
Awesome, thanks! Please can we add the usual lint import check to prevent regressions

jaysoo:
Hey, thanks again for taking the initiative on this and on https://github.com/nrwl/nx/pull/30078.

Unfortunately, both PRs now have many conflicts, and we've been focused on other areas of the tool. There are other places like in `packages/create-nx-workspace` that also use `ora`, so I would suggest on PR to update everything.

If you'd like, opening a discussion on why this is important can also help the team prioritize frequently requested features.

I'm going to close this as it has gone stale.

PR #29141, opened 2024-12-02
fix(js): switch from fast-glob to tinyglobby

benmccann:
## Current Behavior
17 dependencies: https://npmgraph.js.org/?q=fast-glob

## Expected Behavior
2 dependencies: https://npmgraph.js.org/?q=tinyglobby

## Related Issue(s)

https://github.com/es-tooling/ecosystem-cleanup/issues/117

### DIFF
--- a/package.json
+++ b/package.json
@@ -289,6 +289,7 @@
     "tar-stream": "~2.2.0",
     "tcp-port-used": "^1.0.2",
     "terser-webpack-plugin": "^5.3.3",
+    "tinyglobby": "^0.2.10",
     "tmp": "~0.2.1",
     "tree-kill": "1.2.2",
     "ts-jest": "29.1.0",
--- a/packages/js/.eslintrc.json
+++ b/packages/js/.eslintrc.json
@@ -8,6 +8,10 @@
         "no-restricted-imports": [
           "error",
           {
+            "name": "fast-glob",
+            "message": "Please use `tinyglobby` instead."
+          },
+          {
             "name": "fs-extra",
             "message": "Please use equivalent utilities from `node:fs` instead."
           }
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -49,7 +49,6 @@
     "columnify": "^1.6.0",
     "detect-port": "^1.5.1",
     "enquirer": "~2.3.6",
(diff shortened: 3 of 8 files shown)


JamesHenry:
Thanks @benmccann got some task failures to take a look

benmccann:
Thanks @JamesHenry. When I run `pnpm install` it's not updating the `package-lock.json` file. Is there some special setup this repository has around that? It seems the be the cause of the CI errors. I'm not sure if this is just some issue specific to my machine. If you're not familiar with this issue, any chance you could push a commit updating the lockfile for me?

> /home/workflows/workspace/packages/js/package.json
  61:5  error  The version specifier does not contain the installed version of "tinyglobby" package: 0.2.6  @nx/dependency-checks

JamesHenry:
@benmccann The issue is that you haven't added `tinyglobby` to the root `package.json`

The repo doesn't (yet) use package manager workspaces, so your reference in the `@nx/js` package.json doesn't influence the installed packages in any way.

You can verify this with `pnpm why`:

```sh
❯ pnpm why tinyglobby                                                                                                                х 1 24s
Legend: production dependency, optional only, dev only

@nx/nx-source /Users/james/Repos/open-source/nrwl/nx-alt

devDependencies:
nuxt 3.13.2
├─┬ @nuxt/devtools 1.4.2
│ └── tinyglobby 0.2.6
└── tinyglobby 0.2.6
```

Right now the only way tinyglobby gets into the workspace is via a direct and transitive dependency from the `nuxt` package.

Therefore the fix is to run `pnpm add tinyglobby@^0.2.10`

JamesHenry (changes requested):
Thanks @benmccann, sorry I should have remembered this earlier:

Every time we make one of these kinds of PRs we should add a lint entry to prevent regressions: https://github.com/nrwl/nx/blob/5d6119199937b8552655745d39b075cbe4a7df6c/packages/js/.eslintrc.json#L8

(It would be easy for a member of the team not familiar with this effort to add fast-glob back in again)

JamesHenry (approved):

JamesHenry:
FYI I made it a fix pre merge so it gets called out in the changelog

PR #29147, opened 2024-12-02
feat(rsbuild): add intial package

Coly010:
## Current Behavior
We do not currently have a package or plugin to officially support Rsbuild (http://rsbuild.dev)

## Expected Behavior
We should have a package to manage Rsbuild support


jaysoo (approved):

packages/rsbuild/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "main": "index.js",
+  "executors": "./executors.json",
+  "dependencies": {
+    "tslib": "^2.3.0"
+  },
+  "peerDependencies": {},
+  "nx-migrations": {
+    "migrations": "./migrations.json"
+  }
+}
```
JamesHenry, line 34: Please add `"publishConfig": { "access": "public" }` to match other packages.

#29062 would catch this but I haven't had time to revisit the jest resolver issues yet

JamesHenry (changes requested):

packages/rsbuild/package.json
Coly010: done

JamesHenry (approved):

FrozenPandaz (approved):

PR #29150, opened 2024-12-02
docs(core): update e2e-ci glob example

barbados-clemens:
## Current Behavior

the glob pattern in the e2e-ci example is incorrect for nested directories example mentioned.
## Expected Behavior

when using nested directories the **/** pattern is needed, so the
example was incorrect
it's also confusing why something wouldn't work, so added more info as
to the different common glob patterns (`**/*`, `**/**`) to be more clear

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/reference/nx-json.md
+++ b/docs/shared/reference/nx-json.md
@@ -329,8 +329,7 @@ Task Atomizer plugins create several targets with a similar pattern. For example
 ```
 
 {% callout type="info" title="Pattern Matching" %}
-Nx uses glob patterns for matching against the target name. This means that the `**/*` pattern above is required because the target name contains a `/`. If your target name does not contain a `/`, you can use a simpler pattern like `e2e-ci-*`.
-If your tests are nested within a directory, you can use the `**/**` pattern to match all directories.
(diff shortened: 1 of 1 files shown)


Coly010 (approved):

docs/shared/reference/nx-json.md
```diff
@@ -330,6 +330,7 @@ Task Atomizer plugins create several targets with a similar pattern. For example
 
 {% callout type="info" title="Pattern Matching" %}
 Nx uses glob patterns for matching against the target name. This means that the `**/*` pattern above is required because the target name contains a `/`. If your target name does not contain a `/`, you can use a simpler pattern like `e2e-ci-*`.
+If your tests are nested within a directory, you can use the `**/**` pattern to match all directories.
```
isaacplmann, line 333:
```suggestion
Nx uses glob patterns to match against the target name. This means that the `**/**` pattern above is required because the target name contains a `/`. If your target name does not contain a `/`, you can use a simpler pattern like `e2e-ci-*`.
```

isaacplmann (approved):

PR #29151, opened 2024-12-02
feat(module-federation): move common executor logic to module-federation package

Coly010:
## Current Behavior
The logic for the `module-federation-dev-server` and `module-federation-ssr-dev-server` is duplicated across Angular, React and Rspack.

The majority of this logic is the same, and the duplication causes an increased maintenance tax.

## Expected Behavior
Move the logic into a utility that is exposed from `@nx/module-federation`.


packages/module-federation/src/executors/utils/start-remote-iterators.ts
```diff
@@ -0,0 +1,123 @@
...
+import {
+  startSsrStaticRemotesFileServer,
+  startStaticRemotesFileServer,
+} from './start-static-remotes-file-server';
+import {
+  type ExecutorContext,
+  readProjectsConfigurationFromProjectGraph,
+} from '@nx/devkit';
+
+export async function startRemoteIterators(
```
leosvelperez, line 20: nit: What about just `startRemotes`, `serveRemotes` or `runRemotes`? That's what the function does and the iterators are just an implementation detail.

leosvelperez (approved):

packages/module-federation/src/executors/utils/start-remote-iterators.ts
Coly010: It is an impl detail, but I think it makes sense in this case as it indicates that you're getting iterators for both static and dev remotes, which are then combined thereafter.

If we wanted to rename, `runRemotes` might be the best option, but that can be a cleanup for a different time if we decide on it.

jaysoo (approved):

PR #29153, opened 2024-12-02
docs(core): add state management blog post

mhartington:
## Related Issue(s)

Fixes #


docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+title: Angular State Management in 2024
+slug: angular-state-management-2024
+authors: ['Mike Hartington']
+tags: [angular]
+cover_image: /blog/images/2024-12-04/thumbnail.png
+---
+
+## Revisiting State Management
+
+The topic of state management in apps is something developers can spend countless hours discussing and never agree on what the "right" solution is. Truth is, there are so many solutions out there these days and they all share similar concepts that you can't really pick wrong. This is true is most frontend frameworks, but especially true in Angular, where the framework has introduced several new features in the past year that make managing state simpler than ever. Let's take a look at some approaches to state management in Angular that take advantage of modern features of the framework and can set us up for success in 2024.
```
isaacplmann, line 11:
```suggestion
The topic of state management in apps is something developers can spend countless hours discussing and never agree on what the "right" solution is. Truth is, there are so many solutions out there these days and they all share similar concepts that you can't really pick wrong. This is true in most frontend frameworks, but especially true in Angular, where the framework has introduced several new features in the past year that make managing state simpler than ever. Let's take a look at some approaches to state management in Angular that take advantage of modern features of the framework and can set us up for success in 2024.
```

docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+  }
+  update(item) {
+    this.state.update((oldState) =>
+      oldState.map((e) => (e.id === item.id ? item : e))
+    );
+  }
+}
+```
+
+Now this is just a very basic approach to managing state, but if you do not want to reach for an additional libray, you could get pretty far with this basic solution. Throw in a few `effect`s in there if need to performs some side effects, and you're golden.
```
isaacplmann, line 57:
```suggestion
Now this is just a very basic approach to managing state, but if you do not want to reach for an additional libray, you could get pretty far with this basic solution. Throw a few `effect`s in there if needed to perform some side effects, and you're golden.
```

docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Signal State
+
+Now if you've been around Angular long enough, you've probably reached for NgRx before, and with good reason. NgRx provides a standard way of managing state in your app that is scalable and testable. In the past, NgRx has provided a store solution based on RxJS, but in more recent release, NgRx has provided a new API based Signals, Signal State and Signal Store.
```
isaacplmann, line 61:
```suggestion
If you've been around Angular long enough, you've probably reached for NgRx before, and with good reason. NgRx provides a standard way of managing state in your app that is scalable and testable. In the past, NgRx has provided a store solution based on RxJS, but in more recent releases, NgRx provides a new API based on Signals, Signal State and Signal Store.
```

docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+      ...oldState,
+      items: oldState.items.map((e) =>
+        e.id === item.id ? { ...item, name: 'bar' } : e
+      ),
+    }));
+  }
+}
+```
+
+Instantly, some things should stand out to you. First, we use the new `signalState` function, instead of the raw `signal` API. Now we can have a more type safe mechanism for interacting with our state. Second, we're no long passing an array. `signalState` only accepts and object/record like value. If you need arrays, you put that on a property of the state.
```
isaacplmann, line 95:
```suggestion
Instantly, some things should stand out to you. First, we use the new `signalState` function, instead of the raw `signal` API. Now we can have a more type safe mechanism for interacting with our state. Second, we're no long passing an array. `signalState` only accepts an object/record like value. If you need an array, you put that on a property of the state.
```

docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+
+Finally, the way we interact with our state is different. Instead of manipulating the state directly, we use the `patchState` function instead. `patchState` takes the state we want to manipulate, and uses a function to return a new version of that state. To add a new item to our `items` object, we can simply use the spread operator. Removing an item means we use `filter`, and updating an item means we use `map`. What's great about this is not only are we doing things in an immutable way, we're also getting all the types from our state. If we pass along a type that our state doesn't recognize, we'll get a type error before we even save.
+
+## Signal Store
+
+So Signal State is a more prescriptive way of handling smaller state, be it in a component or service. What is Signal Store all about? Signal Store is the more robust solution that you would expect for NgRx. It still is based on Signals, keeps the structure that most larger teams would want for this state solutions. Again, let's rework our previous example and update it for Signal Store.
```
isaacplmann, line 101:
```suggestion
So Signal State is a more prescriptive way of handling smaller state, be it in a component or service. What is Signal Store all about? Signal Store is the more robust solution that you would expect for NgRx. It still is based on Signals, keeping the structure that most larger teams would want for this state solutions. Again, let's rework our previous example and update it for Signal Store.
```

docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+        items: oldState.items.map((e) =>
+          e.id === item.id ? { ...item, name: 'bar' } : e
+        ),
+      }));
+    },
+  }))
+);
+```
+
+Here, we're staring to see a more structure approach to managing state that isolates our state interactions from the rest of our app. Instead of creating a service, we use the `signalStore` function instead. `signalStore` will return an injectable service instead that we provide to a component, or our root app instance. From here, we pass a `withState` function to provide any actual state value to the store. Like `signalState`, this is an object/record only.
```
isaacplmann, line 133:
```suggestion
Here, we're starting to see a more structured approach to managing state that isolates our state interactions from the rest of our app. Instead of creating a service, we use the `signalStore` function instead. `signalStore` will return an injectable service instead that we provide to a component, or our root app instance. From here, we pass a `withState` function to provide any actual state value to the store. Like `signalState`, this is an object/record only.
```

docs/blog/2024-12-04-angular-state-management-2024.md
```diff
@@ -0,0 +1,143 @@
...
+
+For modifying our store, we can use the `withMethods` function and pass any methods we want to expose to our app. What stands out here is that our stores value is accessible without having to inject it. Similar to `signalState`, we use the `patchState` to make any changes we need. Since the mechanism to modify the store in `signalStore` is very close to what we had in `signalState`, it's very approachable to migrating from your simple local store, to something more full featured. So if you're app and team grow significantly, there is a great path forward.
```
isaacplmann, line 135:
```suggestion
For modifying our store, we can use the `withMethods` function and pass any methods we want to expose to our app. What stands out here is that our store's value is accessible without having to inject it. Similar to `signalState`, we use the `patchState` to make any changes we need. Since the mechanism to modify the store in `signalStore` is very close to what we had in `signalState`, it's very approachable when migrating from your simple local store to something more full featured. So if your app and team grow significantly, this is a great path forward.
```

isaacplmann (changes requested):
Minor typo/grammar changes

isaacplmann (approved):

PR #28847, opened 2024-11-08
feat(angular): add support for angular v19

leosvelperez:
## Third-party deps support for Angular v19

- [x] `jest-preset-angular`
  - [x] PRs:
    - [x] https://github.com/thymikee/jest-preset-angular/pull/2835
  - [x] Released:
    - [x] RC: https://github.com/thymikee/jest-preset-angular/releases/tag/v14.4.0-rc.0
    - [x] Stable: https://github.com/thymikee/jest-preset-angular/releases/tag/v14.4.0
- [x] Angular ESLint
  - [x] PRs:
    - [x] https://github.com/angular-eslint/angular-eslint/pull/2109
  - [x] Released:
    - [x] https://github.com/angular-eslint/angular-eslint/releases/tag/v19.0.0
- [x] Storybook
  - [x] PRs:
    - [x] https://github.com/storybookjs/storybook/pull/29659
    - [x] https://github.com/storybookjs/storybook/pull/29677
  - [x] Released:
    - [x] https://github.com/storybookjs/storybook/pull/29679
- [ ] NgRx
  - [x] PRs:
    - [x] https://github.com/ngrx/platform/pull/4602
  - [ ] Released:
    - [x] Beta: https://github.com/ngrx/platform/blob/main/CHANGELOG.md#1900-beta0-2024-11-20
    - [ ] Stable:
- [ ] Analog
  - [x] PRs:
    - [x] https://github.com/analogjs/analog/pull/1447
    - [x] https://github.com/analogjs/analog/pull/1451
  - [ ] Released:
    - [x] Beta: https://github.com/analogjs/analog/releases/tag/v1.10.0-beta.6
    - [ ] Stable:

## Current Behavior

Angular v19 is not supported.

## Expected Behavior

Angular v19 should be supported.

## Related Issue(s)

 
Fixes #29028

PR #29161, opened 2024-12-03
fix(angular): peer dependency issues v19

JosefBredereck:
## Current Behavior
image

## Expected Behavior
Peer dependencies should not be an issue with the version update.

## Related Issue(s)
Introduced with #28847


leosvelperez:
This was already fixed by https://github.com/nrwl/nx/pull/29164

PR #29170, opened 2024-12-03
fix(misc): use the ts sync generator with other bundler tasks

leosvelperez:
## Current Behavior

Only targets using `tsc` trigger the `@nx/js:typescript-sync` generator to run.

## Expected Behavior

Generating projects with other bundlers should also infer the `@nx/js:typescript-sync` generator in their relevant targets.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -38,7 +38,9 @@ export interface VitePluginOptions {
 type ViteTargets = Pick<ProjectConfiguration, 'targets' | 'metadata'>;
 
 function readTargetsCache(cachePath: string): Record<string, ViteTargets> {
-  return existsSync(cachePath) ? readJsonFile(cachePath) : {};
+  return process.env.NX_CACHE_PROJECT_GRAPH !== 'false' && existsSync(cachePath)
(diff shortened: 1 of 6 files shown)


packages/vite/src/plugins/plugin.ts
```diff
@@ -38,7 +38,9 @@ export interface VitePluginOptions {
 type ViteTargets = Pick<ProjectConfiguration, 'targets' | 'metadata'>;
 
 function readTargetsCache(cachePath: string): Record<string, ViteTargets> {
-  return existsSync(cachePath) ? readJsonFile(cachePath) : {};
+  return process.env.NX_CACHE_PROJECT_GRAPH !== 'false' && existsSync(cachePath)
```
ndcunningham, line 41: I guess we should update the other plugins to follow this as well.

ndcunningham (approved):

packages/vite/src/plugins/plugin.ts
leosvelperez: Yeah, some plugins have this. Others don't, but they should all have it.

jaysoo (approved):

PR #29202, opened 2024-12-04
fix(core): Adding react apps/libs to workspaces so they can be referenced.

ndcunningham:
## Current Behavior

This PR address a few things:

1. When we generate an app or a library using the new TS Solution they are not added to the workspaces/packages option.
2. Currently, when we use `pnpm` to generate a workspace the `pnpm-workspace.yaml` looks like below which is not correct
```
packages: 
  - packages/**
```

## Expected Behavior

1. Libraries and apps should be referenced correctly out of the box so that they can be symlinked after installation.
2. The intended `pnpm-workspace.yaml` should look similar to:
```
packages: 
  - 'packages/**'
  - 'demo'
```

### DIFF
--- a/e2e/utils/create-project-utils.ts
+++ b/e2e/utils/create-project-utils.ts
@@ -258,6 +258,7 @@ export function runCreateWorkspace(
   const pm = getPackageManagerCommand({ packageManager });
 
   let command = `${pm.createWorkspace} ${name} --preset=${preset} --nxCloud=skip --no-interactive`;
+
   if (appName) {
     command += ` --appName=${appName}`;
   }
--- a/e2e/expo/src/expo-legacy.test.ts
+++ b/e2e/expo/src/expo-legacy.test.ts
@@ -79,7 +79,7 @@ describe('@nx/expo (legacy)', () => {
 
   it('should serve with metro', async () => {
     let process: ChildProcess;
-    const port = 8081;
+    const port = 8051;
 
     try {
       process = await runCommandUntil(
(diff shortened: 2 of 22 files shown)


e2e/utils/create-project-utils.ts
```diff
@@ -76,10 +76,12 @@ export function newProject({
   name = uniq('proj'),
   packageManager = getSelectedPackageManager(),
   packages,
+  workspaces = false,
 }: {
   name?: string;
   packageManager?: 'npm' | 'yarn' | 'pnpm' | 'bun';
   readonly packages?: Array<NxPackage>;
+  workspaces?: boolean;
```
leosvelperez, line 84: I think we should just allow setting the preset. So, the usage would be `newProject({ preset: 'ts' })`.

The `--workspaces` flag is not needed for the `ts` preset. It's actually ignored with that preset. It would only be needed when using another preset that supports the new TS setup behind the flag (right now, only React). We don't generate any of the React presets in the e2e tests, so it doesn't seem we would need to use it. We can still add it to support using other presets that do need it.

e2e/react/src/react-ts-solution.test.ts
leosvelperez, line 1: IMHO, the current expectations in this file should be tested in unit tests. These are cheap expectations of generated code with a quite expensive test setup. We're not running tasks on the generated code.

Ideally, we would assert that running tasks work as expected with the new TS setup. We can check apps/libs are added to the workspace by adding deps between apps/libs and running tasks.

packages/react/src/utils/add-app-to-pnpm-workspace.ts
leosvelperez, line 1: It would be better to have this helper in the `js` package or as a devkit internal util. Other stacks will also need to use it.

leosvelperez (changes requested):

packages/react/src/utils/add-app-to-pnpm-workspace.ts
ndcunningham: I can move it there I don't see a problem with that.

e2e/react/src/react-ts-solution.test.ts
ndcunningham: I think we should still check that the app/lib is being added to the workspace. This does not really have anything to do with dependencies.

jaysoo (approved):

leosvelperez (approved):

PR #29222, opened 2024-12-05
feat(core): use custom resolution to resolve from source local plugins with artifacts pointing to the outputs

leosvelperez:
## Current Behavior

Local Nx plugins in the new TS setup can't be resolved properly if they aren't built first. Graph plugins can't be built either because the graph is needed to run a task, but the plugin must be built to construct the graph.

## Expected Behavior

Local Nx plugins should work in the new TS setup. A custom resolution is added to resolve the local plugin artifacts from the source.

It will try to use a `development` condition from the `exports` entry in `package.json` if it exists. If it doesn't, it will fall back to guess the source based on the artifact path and some commonly known/used source dirs: `.`, `./src`, `./src/lib`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/plugins/package-json/create-nodes.ts
+++ b/packages/nx/src/plugins/package-json/create-nodes.ts
@@ -138,6 +138,11 @@ export function createNodeFromPackageJson(
   const hash = hashObject({
     ...json,
     root: projectRoot,
+    /**
+     * Increment this number to force processing the package.json again. Do it
+     * when the implementation of this plugin is changed and results in different
+     * results for the same package.json contents.
+     */
     bust: 1,
   });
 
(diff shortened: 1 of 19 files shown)


jaysoo (approved):

packages/nx/src/plugins/package-json/create-nodes.ts
```diff
@@ -138,6 +138,7 @@ export function createNodeFromPackageJson(
   const hash = hashObject({
     ...json,
     root: projectRoot,
+    bust: 1,
```
leosvelperez, line 141: After this PR, with the same file content, a different result is obtained, so we need to bust the cache to ensure projects are processed.

jaysoo (approved):

packages/nx/src/plugins/package-json/create-nodes.ts
jaysoo: There should be a comment to explain why this is here and when to update it.

Issue #28123, opened 2024-09-26
Cannot find project 'web'


dungtm-mc:
### Current Behavior

Anytime I run `npm start web`, i always got: `Cannot find project 'web'`

### Expected Behavior

I just want to run the project

### Steps to Reproduce

Firstly I ran project successfully, but accidentally, the project is crashed like now.

### Nx Report

```shell
Node           : 20.15.1
OS             : win32-x64
Native Target  : wasm32
npm            : 10.7.0

nx (global)        : 19.8.2
nx                 : 19.7.3
@nx/js             : 19.7.3
@nx/jest           : 19.7.3
@nx/linter         : 19.7.3
@nx/eslint         : 19.7.3
@nx/workspace      : 19.7.3
@nx/devkit         : 19.7.3
@nx/eslint-plugin  : 19.7.3
@nx/nest           : 19.7.3
@nx/node           : 19.7.3
@nx/react          : 19.7.3
@nrwl/tao          : 19.7.3
@nx/vite           : 19.7.3
@nx/web            : 19.7.3
@nx/webpack        : 19.7.3
typescript         : 5.5.4
---------------------------------------
Registered Plugins:
@nx/vite/plugin
@nx/eslint/plugin
@nx/jest/plugin
@nx/webpack/plugin
nx/plugins/package-json
```

### Failure Logs

```shell
Admin@Maduro MINGW64 /d/Workspace/caai (main)
$ npm start web

> caia@0.0.0 start
> nx serve web


 NX   The Nx Daemon is unsupported in WebAssembly environments. Some things may be slower than or not function as expected.


 NX   Cannot find project 'web'
```

### Package Manager Version

npm 20.15.1

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

binarybro:
Same here. Since yesterday nx does not see any projects in nx workspaces anymore on my system. That also happens when I create an entirely new workspace using` npx create-nx-workspace` to create a fresh workspace. It seems the graph does not yield any projects in the workspace.

Scaffolding a new project using `npx nx g @nrwl/angular:application some-project` will add a project to the graph, which then shows up. However, if I delete the .nx folder in the workspace, that project is also never reflected on the graph again. It seems graph creation is broken.

Node           : 20.9.0
OS             : darwin-arm64
Native Target  : aarch64-macos
yarn           : 4.3.1

nx                 : 19.7.2
@nx/js             : 19.7.2
[...]

Issue #28000, opened 2024-09-19
fileMap is missing files after upgrade from v19.4.4 to v19.5.0


cainlevy:
### Current Behavior

After upgrading to ~v19.7.4~ v19.5.0, the `@nx/dependency-checks` rule believes that some of my packages have obsolete dependencies, while in fact the dependencies are referenced by existing code and are properly included with ~v18.3.1~ v19.4.4.

So far, the problem appears to be that the files importing the dependencies are no longer included in the file-map.json and thus are not parsed for dependencies. The pattern I am observing is that file-map.json does not include the contents of `src/lib` for any of my packages, and this will miss any dependencies referenced by those files.

NX ~v19~ v19.5.0 shows the following in `fileMap.projectFileMap.validators`:

```json
        {
          "file": "packages/validators/src/index.ts",
          "hash": "3051616070451820306"
        },
```

NX ~v18~ v19.4.4 includes more:

```json
        {
          "file": "packages/validators/src/index.ts",
          "hash": "3051616070451820306"
        },
        {
          "file": "packages/validators/src/lib/validators.spec.ts",
          "hash": "11102733480037190733"
        },
        {
          "file": "packages/validators/src/lib/validators.ts",
          "hash": "1428833592629210108",
          "deps": [
            "npm:date-fns"
          ]
        },
```

### Expected Behavior

I expect the project-graph.json will reference all files in my packages and be aware of all dependencies used by those projects.

### Steps to Reproduce

1. Install NX v19.5.0
2. Create a TypeScript `test` package with `packages/test/src/lib/file.ts`
3. Add a dependency to `packages/test/package.json` and import it from `file.ts`
4. Run `@nx/dependency-checks` lint
5. Observe that it believes the dependency to be obsolete
6. Observe that `packages/test/src/lib/file.ts` is missing from `.nx/workspace-data/project-graph.json` and from `.nx/workspace-data/file-map.json`

### Nx Report

```shell
Node           : 20.11.1
OS             : darwin-arm64
Native Target  : aarch64-macos
npm            : 10.2.4

nx                 : 19.7.4
@nx/js             : 19.7.4
@nx/jest           : 19.7.4
@nx/linter         : 19.7.4
@nx/eslint         : 19.7.4
@nx/workspace      : 19.7.4
@nx/cypress        : 19.7.4
@nx/devkit         : 19.7.4
@nx/esbuild        : 19.7.4
@nx/eslint-plugin  : 19.7.4
[...]

Issue #27295, opened 2024-08-05
Nx should clear the cache after a migration


PowerSupply:
### Current Behavior

Here is a description of the problem that pointed me to the conclusion in the title:

After upgrading to 19.5 from 19.0.4 the dependency-check lint rule started to fail with false negatives.

It said:
error  The "@nx/devkit" package is not used by "xxx" project

But it is used! And the lint worked before the upgrade to 19.5. Running the command with `--skip-nx-cache` still fails. But what is weird is that going to the file where @nx/devkit is imported and just making a small inconsequential change to it (like moving the line with the import to being second instead of first) makes the error go away. And then it doesn't appear again.

Running`nx reset` and then trying the command again seems to fix it which points to some sort of local cache error. I think Nx should reset itself when migrating to a new version since it seems like the cache becomes dirty. But I do not understand why the error still appeared when i used `--skip-nx-cache`.

### Expected Behavior

The problem I had should not have appeared. Proposed solution is to run nx reset as a part of migration to a new version of nx.

### Steps to Reproduce

1. Use Nx 19.0.4. Have a project that passes lint.
2. Upgrade to Nx 19.5.
3. Try to run the lint for the same project. It may fail even if it should succeed and making a small inconsequential change to the file where the import is fixes the problem. Nx reset also fixes the problem.

### Nx Report

```shell
Node           : 18.15.0
OS             : darwin-arm64
Native Target  : aarch64-macos
npm            : 9.5.0

nx                 : 19.5.6
@nx/js             : 19.5.6
@nx/jest           : 19.5.6
@nx/linter         : 19.5.6
@nx/eslint         : 19.5.6
@nx/workspace      : 19.5.6
@nx/angular        : 19.5.6
@nx/cypress        : 19.5.6
@nx/devkit         : 19.5.6
@nx/eslint-plugin  : 19.5.6
@nx/playwright     : 19.5.6
@nx/plugin         : 19.5.6
@nrwl/tao          : 19.5.6
@nx/vite           : 19.5.6
@nx/web            : 19.5.6
@nx/webpack        : 19.5.6
typescript         : 5.4.5
```

### Failure Logs

```shell
error  The "@nx/devkit" package is not used by "xxx" project
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

pascalmann:
Same problem for me.
Some libs disappear from the project graph, so the dependency-check can't find them (and other tasks too).
[...]

PR #29245, opened 2024-12-06
fix(core): prevent parent gitignore files from affecting workspace traversal

### LINKED ISSUE #27368: Nx 19.5 cannot find projects after initial creation
### Current Behavior

I created a new project with `nx g @nx/nest:app backend`. Process went through without any errors. However, If I now try to compile, nx tells me it cannot find the projects. Calling `nx show projects` gives me an empty list.

This happens from version 19.5.0 on. I downgraded to 19.4.4 and there it does work. `nx show projects` gives me all existing projects. Upgrading back to 19.5.0 lets the projects vanish again

### Expected Behavior

I expect a freshly generated project to appear and be found when using `nx show projects`.

### GitHub Repo

_No response_

### Steps to Reproduce

1. `bunx create-nx-workspace my-workspace`
2. `cd my-workspace`
3. `bun nx add @nx/nest`
4. `bun nx g @nx/nest:app my-nest-app`
5. `bun nx show projects`

### Nx Report

```shell
Node           : 22.5.1
OS             : darwin-arm64
bun            : 1.1.22

nx                 : 19.5.7
@nx/js             : 19.5.7
@nx/jest           : 19.5.7
@nx/linter         : 19.5.7
@nx/eslint         : 19.5.7
@nx/workspace      : 19.5.7
@nx/devkit         : 19.5.7
@nx/eslint-plugin  : 19.5.7
@nx/nest           : 19.5.7
@nx/node           : 19.5.7
@nrwl/tao          : 19.5.7
@nx/web            : 19.5.7
@nx/webpack        : 19.5.7
typescript         : 5.4.5
Registered Plugins:
@nx/webpack/plugin
@nx/eslint/plugin
@nx/jest/plugin
```

### Failure Logs

_No response_

### Package Manager Version

bun 1.1.22

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

adambiggs:
## Current Behavior
When the parent directory of a Nx workspace contains a `.gitignore` file, those ignores can affect file traversal within the workspace. This makes project resolution non-deterministic since it depends on gitignore patterns that exist outside of the workspace's control.

For example, if a workspace is inside a directory with `*` in its .gitignore, Nx fails to properly traverse workspace files.

The root cause is that the `ignore` crate's [`WalkBuilder::parents()`](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#method.parents) method defaults to `true`, causing it to respect parent directory gitignore files.

## Expected Behavior
Nx should only respect .gitignore files that exist within the workspace directory. Parent .gitignore files should have no effect on workspace traversal, ensuring consistent and deterministic project resolution regardless of where the workspace is located.

## Related Issue(s)
Fixes #27368, #28000, #27295, #28123, #29413

@nrwl/nx-pipelines-reviewers Given that this fix affects core project resolution behavior, it would be helpful to have a prerelease version so other users can confirm weather it resolves other issues that may or may not be related.

### DIFF
--- a/packages/nx/src/native/walker.rs
+++ b/packages/nx/src/native/walker.rs
@@ -176,6 +176,7 @@ where
     let mut walker = WalkBuilder::new(&directory);
     walker.require_git(false);
     walker.hidden(false);
+    walker.parents(false);
     walker.git_ignore(use_ignores);
     if use_ignores {
         walker.add_custom_ignore_filename(".nxignore");
@@ -314,5 +315,21 @@ nested/child-two/
                 "v1/packages/pkg-b/pkg-b.txt"
             )
         );
+    }
+
+    #[test]
+    fn is_unaffected_by_parent_gitignore() {
+        let parent_temp = assert_fs::TempDir::new().unwrap();
+        parent_temp.child(".gitignore").write_str("*").unwrap();
+        parent_temp.child("workspace/file1.txt").write_str("test").unwrap();
+        parent_temp.child("workspace/project.json").write_str("test").unwrap();
+
+        let workspace_path = parent_temp.path().join("workspace");
+        let mut files: Vec<_> = nx_walker(&workspace_path, true)
+            .map(|f| f.normalized_path)
+            .collect();
+        files.sort();
+
+        assert_eq!(files, vec!["file1.txt".to_string(), "project.json".to_string()]);
     }
 }


AgentEnder (approved):

AgentEnder:
Need to double check some behavior around nested gitignore with this change,  but excited that this may work

Cammisuli:
This will work with simple workspaces that are just their own repo. But if someone has Nx inside a folder within a repo like so:
```
org/
│
├── .git/
│
├── nx-workspace/
│   └── nx.json
│
└── java-project/
    ├── src/
    ├── pom.xml
    └── README.md
```

Then we want to have parent folders activated. We need to think of a proper solution that will support both cases.

adambiggs:
@Cammisuli do folks really have multiple Nx workspaces in a single repo? That seems like a pretty niche edge case unless I'm missing something.

If anyone does run into that issue, a quick & easy workaround would be to simply duplicate the root `.gitignore` into the `nx-workspace` subdirectory. To me that seems like a small price to pay for deterministic project graph resolution.

Do you think it would be worth merging this to unblock everyone from upgrading to versions newer than v19.4.4, and then think of a solution for the multi-workspace monorepo scenario if/when it becomes a problem?

adambiggs:
Was this new behaviour intentional based on a feature request or something? Or was it only introduced because it's the default behaviour of `WalkBuilder`?

I believe this PR restores the same behaviour as before the issue was introduced in 19.5.0. Would your example above have the same problem with 19.4.4?

packages/nx/src/native/walker.rs
```diff
@@ -176,6 +176,7 @@ where
     let mut walker = WalkBuilder::new(&directory);
     walker.require_git(false);
     walker.hidden(false);
+    walker.parents(false);
     walker.git_ignore(use_ignores);
     if use_ignores {
         walker.add_custom_ignore_filename(".nxignore");
```
Cammisuli, line 182:
```suggestion
    walker.git_ignore(use_ignores);
    if use_ignores {
        // disable the parent gitignore if the directory is a git repo, otherwise use the parent gitignore
        walker.parents(!directory.join(".git").exists());
        walker.add_custom_ignore_filename(".nxignore");
```

packages/nx/src/native/walker.rs
```diff
@@ -315,4 +316,20 @@ nested/child-two/
...
+        parent_temp.child("workspace/project.json").write_str("test").unwrap();
+
+        let workspace_path = parent_temp.path().join("workspace");
+        let mut files: Vec<_> = nx_walker(&workspace_path, true)
+            .map(|f| f.normalized_path)
+            .collect();
+        files.sort();
+
+        assert_eq!(files, vec!["file1.txt".to_string(), "project.json".to_string()]);
+    }
```
Cammisuli, line 334:
```suggestion
     #[test]
    fn is_unaffected_by_parent_gitignore_if_traversal_root_is_a_git_root() {
        let parent_temp = assert_fs::TempDir::new().unwrap();
        parent_temp.child(".gitignore").write_str("*").unwrap();
        parent_temp.child("workspace/.git").touch().unwrap();
        parent_temp
            .child("workspace/file1.txt")
            .write_str("test")
            .unwrap();
        parent_temp
            .child("workspace/project.json")
            .write_str("test")
            .unwrap();

        let workspace_path = parent_temp.path().join("workspace");
        let mut files: Vec<_> = nx_walker(&workspace_path, true)
            .map(|f| f.normalized_path)
            .collect();
        files.sort();

        assert_eq!(
            files,
            vec!["file1.txt".to_string(), "project.json".to_string()]
        );
    }

    #[test]
    fn is_affected_by_parent_gitignore_if_traversal_root_is_not_a_git_root() {
        let parent_temp = assert_fs::TempDir::new().unwrap();
        parent_temp.child(".gitignore").write_str("*").unwrap();
        parent_temp
            .child("workspace/file1.txt")
            .write_str("test")
            .unwrap();
        parent_temp
            .child("workspace/project.json")
            .write_str("test")
            .unwrap();

        let workspace_path = parent_temp.path().join("workspace");
        let mut files: Vec<_> = nx_walker(&workspace_path, true)
            .map(|f| f.normalized_path)
            .collect();
        files.sort();

        assert!(files.is_empty());
    }
}
```

Cammisuli (changes requested):
I added some suggestions to satisfy both cases. This will essentially enable parents if the traversal root (aka the nx workspace root) is not the `.git` root. Otherwise it will disable parents if the traversal root is a git root.

adamalton:
> I added some suggestions to satisfy both cases. This will essentially enable parents if the traversal root (aka the nx workspace root) is not the .git root. Otherwise it will disable parents if the traversal root is a git root.

If I'm reading this correctly, your suggested changes will make it so that if the nx workspace root contains a `.git` folder then `.gitignore` files in parent directories will be ignored, but if the nx workspace root does *not* contain a `.git` folder then `.gitignore` files in parent directories will be read.

Assuming that's an accurate description of the behaviour, then it will not correctly handle the case where the nx workspace is inside a subdirectory in a repo _and_ the user has also got a `.gitignore` file higher up the tree somewhere, e.g. in their home directory.

I would suggest that a more desirable behaviour would be that `.gitignore` files in parent directories are taken into account _until_ a `.git` folder is encountered. That `.git` folder might be in the nx workspace root, in which case no parent `.gitignore` files would be read at all. But in the case where the nx workspace is nested, parent `.gitignore` files will be read until we reach the "top" of the git repo under which the nx workspace is controlled, and then no more parent `.gitignore` files will be read.

That would seem like a more desirable behaviour to me.

AgentEnder:
I'm going to close this PR out in favor of https://github.com/nrwl/nx/pull/31911 which incorporates the various suggested feedbacks

Issue #29244, opened 2024-12-06
Nx unnecessarily invokes package manager


JacobLey:
### Current Behavior

When creating the project graph, NX will [invoke the package manager](https://github.com/nrwl/nx/blob/master/packages/nx/src/utils/package-manager.ts#L218) (with `--version` option), despite not _necessarily_ requiring that context.

It is loaded because [Nx _may_ need to provide the context to generate scripts](https://github.com/nrwl/nx/blob/master/packages/nx/src/utils/package-json.ts#L180), but it is possible there are no scripts defined.

Even if they were defined, it is already possible to [parse the version from the package.json itself](https://github.com/nrwl/nx/blob/master/packages/nx/src/utils/package-manager.ts#L225-L237).

A quick JSON load from a file is almost certainly [more performant that invoking the package manager directory](https://github.com/pnpm/pnpm/issues/6447). 

In my particular case, I am actually invoking `@nx/devkit`'s [createProjectGraphAsync](https://github.com/nrwl/nx/blob/master/packages/nx/src/project-graph/project-graph.ts#L230) directly because I am exposing my plugin's functionality as CLI (so need to recreate the context that is usually passed to a plugin).
When _testing_ this code, I use [c8](https://www.npmjs.com/package/c8) which relies on node's [inspector](https://nodejs.org/api/process.html#processfeaturesinspector) but that isn't actually available when [pnpm is installed directly](https://pnpm.io/installation#in-a-docker-container) (not via corepack). When Nx calls pnpm internally, I get [lots of noisy errors about the inspector not being available](https://github.com/nodejs/node/blob/0547dcfc005ae7d9b60d31a7edc90f5a180f907a/lib/internal/util.js#L829) (I wouldn't expect it to be), and would prefer to avoid having these invocations exposed externally.

### Expected Behavior

Nx does not invoke the package manager directly until it absolutely needs to (e.g. when invoking a script).

Possible routes:
1. Defer loading package manager until loading scripts (would still trigger, but not _every_ time)
2. Attempt to load version from package.json _before_ invoking package manager
3. Attach version as an optional field in `nx.json`, similar to [packageManager](https://github.com/nrwl/nx/blob/master/packages/nx/schemas/nx-schema.json#L406). This may be preferable to own the config, rather than relying on corepack's [experimental field](https://nodejs.org/api/packages.html#packagemanager).
[...]

PR #29248, opened 2024-12-07
fix(core): defer loading package manager until necessary

JacobLey:
Only load package manager context when actually necessary for scripts

## Related Issue(s)
https://github.com/nrwl/nx/issues/29244


JamesHenry (commented):
Thanks @JacobLey the optimization makes sense to me

JamesHenry (approved):

PR #29249, opened 2024-12-07
fix(core)!: respect packageManager field in package.json when detecting version

JacobLey:
Attept to read package manager version from config before invoking package manager CLI

BREAK CHANGE: If you have a mismatch between the `packageManager` field in `package.json` and the actual version installed in the environment, it may lead to unexpected behavior when installing. This should not be a problem if you are using corepack already.

## Related Issue(s)
https://github.com/nrwl/nx/issues/29244

### DIFF
--- a/packages/nx/src/utils/package-manager.ts
+++ b/packages/nx/src/utils/package-manager.ts
@@ -213,29 +213,34 @@ export function getPackageManagerVersion(
   packageManager: PackageManager = detectPackageManager(),
   cwd = process.cwd()
 ): string {
+  const nxJson = readNxJson();
+  if (nxJson.cli.packageManagerVersion) {
+    return nxJson.cli.packageManagerVersion;
+  }
   let version;
-  try {
-    version = execSync(`${packageManager} --version`, {
-      cwd,
-      encoding: 'utf-8',
-      windowsHide: true,
-    }).trim();
-  } catch {
-    if (existsSync(join(cwd, 'package.json'))) {
-      const packageVersion = readJsonFile<PackageJson>(
-        join(cwd, 'package.json')
-      )?.packageManager;
-      if (packageVersion) {
-        const [packageManagerFromPackageJson, versionFromPackageJson] =
-          packageVersion.split('@');
-        if (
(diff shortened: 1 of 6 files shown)


packages/nx/src/utils/package-manager.ts
```diff
@@ -213,30 +213,35 @@ export function getPackageManagerVersion(
...
-        ) {
-          version = versionFromPackageJson;
-        }
+  if (existsSync(join(cwd, 'package.json'))) {
+    const packageVersion = readJsonFile<PackageJson>(
+      join(cwd, 'package.json')
+    )?.packageManager;
+    if (packageVersion) {
+      const [packageManagerFromPackageJson, versionFromPackageJson] =
+        packageVersion.split('@');
```
leosvelperez, line 227: I know this is existing logic, but the parsing must be a bit more complex than that. It's less of an issue in the current approach because it's only used as a fallback that, most of the time, shouldn't be hit. With the new changes, it will be the first thing we check, so we need to ensure we correctly parse it. The `packageManager` specifier can contain an SHA-224 hash or point to a URL to a `.js` file or a `.tgz` file. For example:

```json
{
  "packageManager": "yarn@3.2.3+sha224.953c8233f7a92884eee2de69a1b92d1f2ec1655e66d08071ba9a02fa"
}
```

```json
{
  "packageManager": "yarn@https://registry.npmjs.org/@yarnpkg/cli-dist/-/cli-dist-3.2.3.tgz#sha224.16a0797d1710d1fb7ec40ab5c3801b68370a612a9b66ba117ad9924b"
}
```

We need to handle that accordingly. Please also add some unit tests to cover the different scenarios.

You can read more at https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages.

leosvelperez (dismissed):
Thanks for the contribution!

I think it makes sense to invert the operations to gain some performance. I left some comments to address. Could you please also benchmark this and post the results in the PR description?

jaysoo (approved):

FrozenPandaz (approved):

PR #29262, opened 2024-12-09
fix(angular): handle removed angular-eslint rules in root eslint config files and update package

leosvelperez:
## Current Behavior

The migration to remove Angular ESLint rules that were removed in v19 does not handle these rules in the root eslint config files.

## Expected Behavior

The migration to remove Angular ESLint rules that were removed in v19 should handle these rules in the root eslint config files.

## Related Issue(s)

Fixes #


JamesHenry (approved):

JamesHenry (changes requested):
Let's go to 19.0.2

JamesHenry (approved):

Coly010 (approved):

Issue #27849, opened 2024-09-10
@nx/eslint:lint-project create a fresh graph at every run


beaussan:
### Current Behavior

`@nx/eslint:lint-project` create a fresh graph every time, making custom plugins wrapping other generators (such as `@nx/react:library`) test creating a lot of graph, running the tests extremely slow

```
 PASS   my-plugin  my-plugin/src/generator.spec.ts (26.856 s)
  lib generator
    ✓ should generate 5 libs with skipEslint=false (25324 ms)
    ✓ should generate 5 libs with skipEslint=true (167 ms)
```

### Expected Behavior

`@nx/eslint:lint-project` read the cached graph instead of creating a new graph every time

### GitHub Repo

https://github.com/beaussan/repro-nx-plugin-test-daemon

### Steps to Reproduce

Here is a test file that trigger the issue:

```
import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
import { Tree, readProjectConfiguration } from '@nx/devkit';

import { libraryGenerator as nxReactLibGenerator } from '@nx/react/src/generators/library/library';
import { Linter } from '@nx/eslint';

const numberOfLibs = 5;

describe('lib generator', () => {
  let tree: Tree;

  beforeEach(() => {
    tree = createTreeWithEmptyWorkspace();
  });

  it(`should generate ${numberOfLibs} libs with skipEslint=false`, async () => {
    for (let i = 0; i < numberOfLibs; i++) {
      await nxReactLibGenerator(tree, {
        bundler: 'vite',
        // waiting on an answer for a better way from the nx team, since their getRelativeCwd is not a public api _yet_.
        compiler: 'babel',
        directory: `my-lib-${i}`,
        linter: Linter.EsLint,
        name: `my-lib-${i}`,
        projectNameAndRootFormat: 'as-provided',
        setParserOptionsProject: true,
        skipPackageJson: true,
        skipTsConfig: true,
        style: 'styled-components',
        unitTestRunner: 'vitest',
      });
      const config = readProjectConfiguration(tree, `my-lib-${i}`);
      expect(config).toBeDefined();
    }
  }, 90000);

  it(`should generate ${numberOfLibs} libs with skipEslint=true`, async () => {
    for (let i = 0; i < numberOfLibs; i++) {
      await nxReactLibGenerator(tree, {
        bundler: 'vite',
        // waiting on an answer for a better way from the nx team, since their getRelativeCwd is not a public api _yet_.
        compiler: 'babel',
[...]

PR #29286, opened 2024-12-10
feat(react): add support for React 19 for new Workspaces

Coly010:
## Current Behavior
We currently have no support for React 19, generating only React 18 applications.

## Expected Behavior
Add utils to determine what version of React is installed in the workspace.
If React 18 is the main version of react installed, continue to generate React 18 projects.
If React 19 is the main version of react installed, generate React 19 projects.
If no React version is installed or can be determined, generate React 19 projects.

### DIFF
--- a/packages/react/src/generators/application/application.spec.ts
+++ b/packages/react/src/generators/application/application.spec.ts
@@ -1,9 +1,8 @@
-import 'nx/src/internal-testing-utils/mock-project-graph';
-
 import { installedCypressVersion } from '@nx/cypress/src/utils/cypress-version';
 import {
   getPackageManagerCommand,
   getProjects,
+  ProjectGraph,
   readJson,
   readNxJson,
   Tree,
@@ -20,6 +19,17 @@ const { load } = require('@zkochan/js-yaml');
 // need to mock cypress otherwise it'll use the nx installed version from package.json
 //  which is v9 while we are testing for the new v10 version
 jest.mock('@nx/cypress/src/utils/cypress-version');
+
+let projectGraph: ProjectGraph;
+jest.mock('@nx/devkit', () => {
+  const original = jest.requireActual('@nx/devkit');
+  return {
+    ...original,
+    createProjectGraphAsync: jest
+      .fn()
+      .mockImplementation(() => Promise.resolve(projectGraph)),
+  };
+});
 
 const packageCmd = getPackageManagerCommand().exec;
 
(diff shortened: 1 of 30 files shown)


packages/react/src/generators/application/application.spec.ts
```diff
@@ -19,6 +18,17 @@ import { Schema } from './schema';
 //  which is v9 while we are testing for the new v10 version
 jest.mock('@nx/cypress/src/utils/cypress-version');
 
+let projectGraph: ProjectGraph;
+jest.mock('@nx/devkit', () => {
```
jaysoo, line 22: Rather than read the package from graph, use `readModulePackageJson` from `nx/src/devkit-internals` to find the version of React.

jaysoo: @Coly010 brought up a point of the function not supporting Yarn PnP, and also that the root package.json not always containing `react`. The graph contains the deduped version so it will always exist.

Coly010: The graph provides a more accurate source of truth for package versions across different package managers and package manager resolution types (bun and yarn pnp where no node_modules folder exists).

Let's stick to reading graph.

gperdomor:
@Coly010 how this affects Next apps?...

Coly010:
Blocked until Unit Testing solution is ready

beaussan:
I see that this pr introduce a create graph async during a generator, this is already causing issues regarding eslint (see https://github.com/nrwl/nx/issues/27849), can we envision a generator parameter to explicitly say the version of react to expect?

Otherwise, every generator that are built on top of this will have their test and run go way slower due to creating a fresh graph

gperdomor:
@Coly010 can you add vitest support to next app generator too? 🙏🏻

ndcunningham (approved):

ndcunningham:
@Coly010 you may also want to remove `NX_SERVE_STATIC_BUILD_RUNNING` from https://github.com/nrwl/nx/blob/master/packages/next/plugins/with-nx.ts#L217-L221

jaysoo (approved):

Coly010:
> @Coly010 you may also want to remove `NX_SERVE_STATIC_BUILD_RUNNING` from https://github.com/nrwl/nx/blob/master/packages/next/plugins/with-nx.ts#L217-L221

I think this is fine to stay as it will handle people that _are_ currently using serve-static (via a target in project.json)

p4cino:
gentlemen when merge 😀

FrozenPandaz (approved):

PR #29293, opened 2024-12-10
docs(nx-dev): add tailoring nx for your org post

philipjfulcher:


docs/blog/2024-12-10-tailoring-nx-for-your-organization.md
```diff
@@ -0,0 +1,182 @@
...
+
+- the project naming convention based on the type of library and its scope
+- the directory structure based on nested folders for scope
+- the import naming convention
+- tag name convention based on type and scope
+
+This is a lot to remember! Even the most diligent engineer will forget one of these or miss a typo. **A custom generator** can solve this problem. Consider a custom generator for your organization that accepts options like this:
+
+```bash
+nx generate @org/plugin:library --name util --scope shared --type util
```
juristr, line 45: We should re-emphasize at the end of the section that this is the command to run the custom generator you just built. I feel it might get lost here as we say "Consider a custom generator for ..." which sounds "hypothetical" etc.. :)

docs/blog/2024-12-10-tailoring-nx-for-your-organization.md
```diff
@@ -0,0 +1,182 @@
...
+    };
+  },
+];
+```
+
+This is a very naive implementation, but it gives you an idea of what implementing these task inference plugins looks like. [Check out this much more developed example.](/extending-nx/tutorials/tooling-plugin)
+
+## Publish to your org with `nx release`
+
+Plugins can also be bundled and published for your organization. Since we’re the monorepo people, it might seem wild to suggest that you would have multiple monorepos in your organization; but this is a common scenario, and we whole-heartedly support it.
```
juristr, line 130: again emphasize that you can just have plugins in your monorepo and run them right away without any further compilatio or anything. Just so we repeat this and ppl don't miss it.

docs/blog/2024-12-10-tailoring-nx-for-your-organization.md
```diff
@@ -0,0 +1,182 @@
...
+- Add the proper tags to the project
+- Select the correct set of tools
+
+All of that without having to refer to your documentation on how to run the generator the right way. And most of the functionality is provided by Nx! You just need to provide some logic specific to your organization.
+
+These generators will also show up in [Nx Console](/getting-started/editor-setup) alongside the other generators from Nx plugins, so engineers will have quick access.
+
+{% youtube src="https://www.youtube.com/watch?v=myqfGDWC2go" /%}
+
+## Configure your tools in one place with custom inferred tasks
```
juristr, line 97: I think we need to do some more explanation here. Inferred tasks sounds like super complicated. I'd mention things like
- project crystal
- inference plugins can be configured once at the root level and then automatically whenever that specific "config" file it monitors is found, you can perform some action (whatever that might be)

> Writing your own task inference plugin allows you handle these situations. Let’s say the hottest, coolest new tool hits the ecosystem, and you need to use it today.

I'd not mention it in this way. Even today if a new tool hits the market and you add it to an Nx workspace you can run it right away by defining a package.json scripts and go. But then you can create a Crystal plugin to using task inference to automate how the new tool is used within your monorepo, s.t. not every single new project needs to redefine it etc

juristr (changes requested):

juristr (approved):

PR #29299, opened 2024-12-10
feat(js): update vue/node app and lib generators to support TS solutions

jaysoo:
## Related Issue(s)

Fixes #


packages/playwright/src/generators/configuration/configuration.ts
```diff
@@ -97,7 +97,12 @@ export async function configurationGeneratorInternal(
     if (isTsSolutionSetup) {
       // skip eslint from typechecking since it extends from root file that is outside rootDir
       if (options.linter === 'eslint') {
-        tsconfig.exclude = ['dist', 'eslint.config.js'];
+        tsconfig.exclude = [
```
jaysoo, line 100: I realized that PW didn't exclude the same files as our other generators, so this updates them. It'll also help once we rename `eslint.config.js` to `eslint.config.cjs` to resolve CJS/ESM issues.

Coly010 (approved):

packages/vue/src/generators/application/files/common/src/vue-shims.d.ts.template
```diff
@@ -0,0 +1,5 @@
+declare module '*.vue' {
```
jaysoo, line 1: This is needed because `typecheck` never worked without Volar in VSCode. The shims here at least makes it pass even though nothing meaningful is checked. We already do the same for Vue libs.

packages/vue/src/generators/library/library.ts
```diff
@@ -86,14 +120,23 @@ export async function libraryGeneratorInternal(tree: Tree, schema: Schema) {
...
+    (options.publishable || options.bundler !== 'none')
+  ) {
     updateJson(tree, `${options.projectRoot}/package.json`, (json) => {
       json.name = options.importPath;
       return json;
     });
   }
 
-  if (!options.skipTsConfig) {
+  if (options.bundler === 'none' && options.addPlugin) {
```
leosvelperez, line 133: It would be weird to happen, but `addPlugin` can be `false` and the workspace could still have the plugin registered from a previous generation. So, we should always ensure the project is excluded. The helper will ignore if the plugin is not registered.
```suggestion
  if (options.bundler === 'none') {
```

packages/vue/src/utils/create-ts-config.ts
```diff
@@ -1,7 +1,92 @@
...
+
+      const types = new Set(json.compilerOptions.types ?? []);
+      types.add('vite/client');
+
+      json.compilerOptions.types = Array.from(types);
+
+      return json;
+    });
+  } else {
+  }
```
leosvelperez, line 86:
```suggestion
  }
```

packages/vue/src/generators/library/lib/create-library-files.ts
```diff
@@ -31,8 +31,22 @@ export function createLibraryFiles(host: Tree, options: NormalizedSchema) {
...
+  ) {
+    writeJson(host, joinPathFragments(options.projectRoot, 'package.json'), {
+      name: options.name,
+      version: '0.0.1',
+      main: './index.js',
+      types: './index.d.ts',
+      exports: {
+        '.': {
+          import: './index.mjs',
+          require: './index.js',
```
leosvelperez, line 46: We should add the `types` condition.

packages/vue/src/utils/create-ts-config.ts
```diff
@@ -1,7 +1,92 @@
...
+  } as any;
+
+  // inline tsconfig.base.json into the project
+  if (options.rootProject) {
+    json.compileOnSave = false;
+    json.compilerOptions = {
+      ...shared.tsConfigBaseOptions,
+      ...json.compilerOptions,
+    };
+    json.exclude = ['node_modules', 'tmp'];
```
leosvelperez, line 66: We shouldn't inline the compiler options. Regardless of being a root project, the `tsconfig.json` file still extends from `tsconfig.base.json` and includes no files.

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -135,6 +135,7 @@ async function createPreset(tree: Tree, options: Schema) {
       e2eTestRunner: options.e2eTestRunner ?? 'playwright',
       addPlugin,
       nxCloudToken: options.nxCloudToken,
+      useTsSolution: options.workspaces,
```
leosvelperez, line 138: Don't we need to forward `formatter` here?

packages/workspace/src/generators/preset/preset.ts
```diff
@@ -163,6 +164,7 @@ async function createPreset(tree: Tree, options: Schema) {
...
```
leosvelperez, line 167: Same here.

packages/express/src/generators/application/schema.json
```diff
@@ -33,14 +33,17 @@
     "linter": {
       "description": "The tool to use for running lint checks.",
       "type": "string",
-      "enum": ["eslint"],
-      "default": "eslint"
+      "enum": ["eslint", "none"],
+      "default": "none",
```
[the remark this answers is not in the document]
jaysoo: No, we'll just default them to `none` for all. It's a generator, and users can always choose the right one from the prompt. The programmatic usage doesn't change.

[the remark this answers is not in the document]
leosvelperez: Then, we need to adjust the js library generator to align that.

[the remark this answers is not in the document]
jaysoo: Yes, we can do a follow-up PR to address that.

packages/node/src/generators/application/application.ts
```diff
@@ -198,16 +204,30 @@ function addProject(tree: Tree, options: NormalizedSchema) {
...
+    writeJson(tree, joinPathFragments(options.appProjectRoot, 'package.json'), {
+      name: getImportPath(tree, options.name),
+      version: '0.0.1',
+      private: true,
+      nx: {
+        name: options.name,
+        projectType: 'application',
+        sourceRoot: project.sourceRoot,
+        targets: project.targets,
+        tags: project.tags,
```
leosvelperez, line 217: Tags could still be empty
```suggestion
        tags: project.tags?.length ? project.tags : undefined,
```

packages/nuxt/src/generators/application/application.spec.ts
```diff
@@ -194,4 +200,177 @@ describe('app', () => {
...
+        ]
+      `);
+      expect(readJson(tree, 'myapp/tsconfig.json')).toMatchInlineSnapshot(`
+        {
+          "compilerOptions": {},
+          "extends": "../tsconfig.base.json",
+          "files": [],
+          "include": [
+            ".nuxt/nuxt.d.ts",
+          ],
```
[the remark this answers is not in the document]
leosvelperez: This is still relevant. We should not generate this `include` here.

packages/vue/src/generators/library/library.spec.ts
```diff
@@ -438,4 +439,209 @@ module.exports = [
...
+          // worker: {
+          //  plugins: [ nxViteTsPaths() ],
+          // },
+          test: {
+            watch: false,
+            globals: true,
+            environment: 'jsdom',
+            include: ['src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
+            reporters: ['default'],
+            coverage: { reportsDirectory: '../coverage/my-lib', provider: 'v8' },
```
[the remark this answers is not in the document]
leosvelperez: This still needs fixing.

packages/vue/src/generators/library/library.ts
leosvelperez: This still needs fixing.

packages/vue/src/utils/create-ts-config.ts
```diff
@@ -1,7 +1,72 @@
...
+) {
+  const json = {
+    files: [],
+    include: [],
+    references: [
+      {
+        path: type === 'app' ? './tsconfig.app.json' : './tsconfig.lib.json',
+      },
+    ],
+  } as any;
```
leosvelperez, line 50: We need to extend from the `tsconfig.base.json`, or is that done somewhere else?

packages/nuxt/src/generators/application/schema.d.ts
```diff
@@ -14,6 +14,7 @@ export interface Schema {
   setParserOptionsProject?: boolean;
   style?: 'css' | 'scss' | 'less' | 'none';
   nxCloudToken?: string;
+  useTsSolution?: boolean;
```
leosvelperez, line 17: We need to define `formatter` here to forward it to the js `init` generator.

packages/node/src/generators/application/schema.d.ts
```diff
@@ -23,6 +23,7 @@ export interface Schema {
   docker?: boolean;
   isNest?: boolean;
   addPlugin?: boolean;
+  useTsSolution?: boolean;
```
leosvelperez, line 26: We need to define `formatter` here to forward it to the js `init` generator.

packages/nest/src/generators/application/schema.d.ts
```diff
@@ -15,6 +15,7 @@ export interface ApplicationGeneratorOptions {
   rootProject?: boolean;
   strict?: boolean;
   addPlugin?: boolean;
+  useTsSolution?: boolean;
```
leosvelperez, line 18: We need to define `formatter` here to forward it to the js `init` generator.

PR #29322, opened 2024-12-12
fix(angular): correctly normalize dev-server options

SebastianPodgajny:
For Angular > 18 schema hmr option was ignored, it was impossible to disable hmr and default value changed in ng19 https://github.com/angular/angular/issues/59058

## Related Issue(s)

Fixes #


leosvelperez (approved):
Thank you!

PR #29324, opened 2024-12-12
fix(angular): add-localize-polyfill-to-targets fails if polyfills is a string

stianmorsund:
## Current Behavior
Migration fails if polyfills is a string

## Expected Behavior
Migration should work even if polyfills is a string


leosvelperez (approved):
Thank you!

PR #29331, opened 2024-12-12
fix(core): Update move/remove workspace generators to work with ts project references

ndcunningham:
## Current Behavior

When a workspace is setup to use ts project references the move/remove generators from `@nx/workspace` do not work correctly or result in an incorrect state for the workspace.

## Expected Behavior

This should work OOTB.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/workspace/src/generators/remove/lib/update-tsconfig.spec.ts
+++ b/packages/workspace/src/generators/remove/lib/update-tsconfig.spec.ts
@@ -7,6 +7,7 @@ import {
 import { createTreeWithEmptyWorkspace } from '@nx/devkit/testing';
 import { Schema } from '../schema';
 import { updateTsconfig } from './update-tsconfig';
+import * as tsSolution from '../../../utilities/typescript/ts-solution-setup';
 
 // nx-ignore-next-line
 const { libraryGenerator } = require('@nx/js');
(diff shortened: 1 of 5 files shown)


packages/workspace/src/utilities/typescript/ts-solution-setup.ts
```diff
@@ -0,0 +1,81 @@
...
+import { FsTree } from 'nx/src/generators/tree';
+import { type PackageJson } from 'nx/src/utils/package-json';
+
+function isUsingPackageManagerWorkspaces(tree: Tree): boolean {
+  return isWorkspacesEnabled(tree);
+}
+
+function isWorkspacesEnabled(
+  tree: Tree
+  // packageManager: PackageManager = detectPackageManager(),
```
jaysoo, line 16: Remove unused code

packages/workspace/src/generators/remove/lib/update-tsconfig.ts
```diff
@@ -11,32 +12,52 @@ import {
...
-      for (const importPath in json.compilerOptions.paths) {
-        for (const path of json.compilerOptions.paths[importPath]) {
-          const project = findProjectForPath(
-            normalizePath(path),
-            projectMapping
-          );
-          if (project === schema.projectName) {
-            delete json.compilerOptions.paths[importPath];
-            break;
+      if (!isUsingTsSolution) {
```
jaysoo, line 33: Nit: the non-negative check should for readability

packages/workspace/src/generators/remove/lib/update-tsconfig.spec.ts
```diff
@@ -212,4 +213,42 @@ describe('updateTsconfig', () => {
...
+
+  it('should work with tsSolution setup', async () => {
+    jest.spyOn(tsSolution, 'isUsingTsSolutionSetup').mockReturnValue(true);
+
+    await libraryGenerator(tree, {
+      directory: 'my-lib',
+    });
+
+    const tsconfigContent = {
+      ...readJson(tree, 'tsconfig.base.json'),
```
jaysoo, line 225: I don't think we need the entire content of `tsconfig.base.json`.  The base file needs to exist and only set `compilerOptions`, and the `tsconfig.json` file should have empty `include` and `files` to be considered TS solution

jaysoo (approved):

jaysoo:
The flaky test for release e2e will be fixed separately, it isn't affected by this PR. Merging.

PR #29334, opened 2024-12-12
feat(eslint): rename eslint.config.js to eslint.config.cjs to resolve them as CommonJS

jaysoo:
This PR updates our generators to use `eslint.config.cjs` instead of `eslint.config.js` so that Node resolution will treat it as CommonJS. This solves an issue where having `"type": "module"` in `package.json` will result in an error when Node tries to resolve the config file as ESM.

Also allows us to clean up out Remix generators to not have to rename to `eslint.config.cjs` to solve the same issue.

## Related Issue(s)

Fixes #


AgentEnder (approved):

packages/workspace/src/generators/move/lib/update-project-root-files.ts
```diff
@@ -51,7 +51,7 @@ export function updateFilesForRootProjects(
     if (!allowedExt.includes(ext)) {
       continue;
     }
-    if (file === '.eslintrc.json' || file === 'eslint.config.js') {
+    if (file === '.eslintrc.json' || file === 'eslint.config.cjs') {
```
JamesHenry, line 54: Why was this one a replacement when others were applied in addition to the existing?

jaysoo: Fixed

packages/eslint/src/generators/utils/flat-config/ast-utils.spec.ts
```diff
@@ -124,6 +124,10 @@ describe('ast-utils', () => {
             ]
         }).map(config => ({
             ...config,
+            files: [
```
JamesHenry, line 127: This doesn't seem right? This will be setting files to _only_ resolve `cts` and `mts`? Unless eslint is doing some magical merging behind the scenes?

JamesHenry (changes requested):
Mostly LGTM!

packages/eslint/src/generators/utils/flat-config/ast-utils.spec.ts
jaysoo: This one specifically is when converting eslintrc to flat config when called from `convertEslintJsonToFlatConfig`. In those cases the `.ts` and `.tsx` extensions are already specified. I could add it to the ast utils as well just in case.

jaysoo: I updated it.

JamesHenry (approved):

PR #29348, opened 2024-12-13
feat(core): add `prefixColor` property to `@nx:run-commands`

nc1z:
## Current Behavior

Running tasks in parallel, we can set prefixes. However, there is no way to label our prefixes with colors. Currently the only way is to set the entire line to the color we want. This is not ideal as users may want to keep the default colors for the logs and only want to use the prefix colors as identifiers.

## Expected Behavior

PR to add this option

## Screenshots (Examples)

Tested with example screenshots below

### Before (v20.2.2 latest release)
description

image

### This PR (local release)
description

image

## Related Issue(s)

Fixes #


nc1z:
Update PR description with images tested with local release vs latest release.

Build/CI should pass once `nx run e2e-release:e2e-ci--src/circular-dependencies.test.ts` is fixed

AgentEnder (approved):

PR #29349, opened 2024-12-13
feat(rsbuild): add react and vue support for app generation

Coly010:
## Current Behavior
We do not have a generator that can scaffold a react or vue app using rsbuild.

## Expected Behavior
Update the react application generator to support a bundler option of rsbuild
Update the vue application generator to support a bundler option of rsbuild


packages/rsbuild/package.json
```diff
@@ -33,10 +33,32 @@
...
+    },
+    "./generators": {
+      "types": "./generators.d.ts",
+      "default": "./generators.js"
+    },
+    "./config-utils": {
+      "types": "./config-utils.d.ts",
+      "default": "./config-utils.js"
+    },
+    "./plugin": {
```
jaysoo, line 59: We can use `@nx/rsbuild` for the plugin rather than `@nx/rsbuild/plugin`. Thee other plugins couldn't do this because the `.` export was already used for other things.

packages/rsbuild/src/utils/ast-utils.spec.ts
```diff
@@ -0,0 +1,400 @@
...
+        'apps/my-app/rsbuild.config.ts',
+        './src/index.html'
+      );
+
+      // ASSERT
+      expect(tree.read('apps/my-app/rsbuild.config.ts', 'utf-8'))
+        .toMatchInlineSnapshot(`
+        "import {defineConfig} from '@rsbuild/core';
+              export default defineConfig({
+                html: {template: './src/index.html',
```
jaysoo, line 64: Can we fix the formatting here? Moving forward we should assume Prettier may not be used.

packages/rsbuild/src/utils/ast-utils.spec.ts
```diff
@@ -0,0 +1,400 @@
...
+
+      // ACT
+      addCopyAssets(tree, 'apps/my-app/rsbuild.config.ts', './src/assets');
+
+      // ASSERT
+      expect(tree.read('apps/my-app/rsbuild.config.ts', 'utf-8'))
+        .toMatchInlineSnapshot(`
+        "import {defineConfig} from '@rsbuild/core';
+              export default defineConfig({
+            output: {
```
jaysoo, line 122: Same formatting issue here. Check other places the AST modifies as well.

packages/react/src/utils/has-rsbuild-plugin.ts
```diff
@@ -0,0 +1,10 @@
+import { readNxJson, Tree } from '@nx/devkit';
+
+export function hasRsbuildPlugin(tree: Tree) {
+  const nxJson = readNxJson(tree);
+  return !!nxJson.plugins?.some((p) =>
+    typeof p === 'string'
+      ? p === '@nx/rsbuild/plugin'
```
jaysoo, line 7: Rename to `@nx/rsbuild` if we change the inference plugin entry. If not, it should be a follow-up PR.

packages/rsbuild/src/utils/add-build-plugin.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    );
+
+    // ASSERT
+    expect(tree.read('apps/my-app/rsbuild.config.ts', 'utf-8'))
+      .toMatchInlineSnapshot(`
+      "import { pluginReact } from '@rsbuild/plugin-react';
+        import { defineConfig } from '@rsbuild/core';
+            import { less } from '@rsbuild/plugin-less';
+            export default defineConfig({
+              plugins: [less(),pluginReact()],
```
jaysoo, line 76: Formatting: missing a space between plugins.

packages/rsbuild/src/utils/add-build-plugin.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+      'less'
+    );
+
+    // ASSERT
+    expect(tree.read('apps/my-app/rsbuild.config.ts', 'utf-8'))
+      .toMatchInlineSnapshot(`
+      "import { less } from '@rsbuild/plugin-less';
+        import { defineConfig } from '@rsbuild/core';
+            export default defineConfig({
+          plugins: [less()],
```
jaysoo, line 34: Formatting: indentation is off.

e2e/react/src/react-rsbuild.test.ts
```diff
@@ -0,0 +1,102 @@
+import {
+  checkFilesExist,
+  cleanupProject,
+  newProject,
+  runCLI,
+  runCLIAsync,
+  uniq,
+} from '@nx/e2e/utils';
+
+describe('Build React applications and libraries with Rsbuild', () => {
```
jaysoo, line 10: Can we add a test that runs Playwight e2e? To be sure that the server actually works and e2e is configured correctly out of the box.

Should we also make the port 4200 to match the other setups?

packages/rsbuild/package.json
Coly010: Let's leave this as is in this PR and follow up the change in a separate PR as there will be more things touched by the change from `@nx/rsbuild/plugin` to `@nx/rsbuild` 
(checking if build plugin exists, adding the plugin to nx.json, init, etc)

packages/react/src/utils/has-rsbuild-plugin.ts
Coly010: Will be in a follow up

e2e/react/src/react-rsbuild.test.ts
Coly010: Yep, can update the port and make sure playwright tests are running.

e2e/vue/src/vue.test.ts
```diff
@@ -33,6 +40,29 @@ describe('Vue Plugin', () => {
...
+    let result = runCLI(`test ${app}`);
+    expect(result).toContain(`Successfully ran target test for project ${app}`);
+
+    result = runCLI(`build ${app}`);
+    expect(result).toContain(
+      `Successfully ran target build for project ${app}`
+    );
+
+    // TODO: enable this when tests are passing again.
+    // Colum confirmed locally that the generated config and the playwright tests are working.
```
Coly010, line 58: image
image

Coly010: The other vue e2e test also has the runE2e block commented out. 
We should investigate why it doesn't like CI later

jaysoo (approved):

PR #29376, opened 2024-12-16
docs(core): common tasks page

isaacplmann:
Create a common task names concept page.

The main purpose of this page is to have a good search result for people searching for "build", "test" or "lint" in our docs search.  These terms are regularly in the top 10 most searched terms.


docs/shared/concepts/common-tasks.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+You can also [override the inferred task configuration](/concepts/inferred-tasks#overriding-inferred-task-configuration) as needed.
+
+{% /tab %}
+{% tab label="Custom Script" %}
+
+You can define your own `build` task in your project configuration. Here is an example that uses `ts-node` to run a node script.
+
+```json {% fileName="packages/my-project/project.json" %}
```
jaysoo, line 86: Can we use `package.json` here instead of `project.json`. It'll align better with our new TS setup. We can also offer both example in tabs and show `package.json` in the first tab.

jaysoo (approved):

PR #29382, opened 2024-12-16
feat(web): support custom cors headers

aidant:
## Current Behavior

It is not possible to pass custom cors headers to the `@nx/web:file-server` executor despite `http-server` supporting it.

## Expected Behavior

`@nx/web:file-server` should pass-through a string of custom cors headers to `http-server`.


leosvelperez (approved):
Thank you!

PR #29393, opened 2024-12-17
fix(js): Configure typescript plugin to handle non-buildable libs

ndcunningham:
## Current Behavior

Currently, we are excluding non-buildable libs from the `@nx/js/typescript` plugin. Although that allows non-buildable projects from have the build target being inferred it also as a side-effect removes the `typecheck` target which is unintended. Additionally, to breaks the pattern of being self containment that TS project solutions brings as we were modifying the root `nx.json`

## Expected Behavior

1. Non-buildable libs should not have a build target.
2. Non-buildable libs should have the typecheck target.
3. Buildable libs remain unchanged and should have both a build and typecheck target.
4. Remove the `exclude` from `nx.json` for non-buildable libs.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/plugins/typescript/plugin.ts
+++ b/packages/js/src/plugins/typescript/plugin.ts
@@ -19,7 +19,15 @@ import {
 import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
 import { minimatch } from 'minimatch';
 import { existsSync, readdirSync, statSync } from 'node:fs';
-import { basename, dirname, join, normalize, relative, sep } from 'node:path';
+import {
+  basename,
+  dirname,
+  join,
+  normalize,
+  relative,
+  resolve,
+  sep,
+} from 'node:path';
 import { hashArray, hashFile, hashObject } from 'nx/src/hasher/file-hasher';
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import { getLockFileName } from 'nx/src/plugins/js/lock-file/lock-file';
@@ -201,6 +209,7 @@ async function createNodesInternal(
     context.workspaceRoot,
     projectRoot
   );
+
   const nodeHash = hashArray([
     ...[
       fullConfigPath,
@@ -313,7 +322,16 @@ function buildTscTargets(
   }
 
   // Build target
-  if (options.build && basename(configFilePath) === options.build.configName) {
+  if (
+    options.build &&
+    basename(configFilePath) === options.build.configName &&
+    isValidPackageJsonBuildConfig(
+      tsConfig,
+      context.workspaceRoot,
+      projectRoot,
+      configFilePath
+    )
+  ) {
     internalProjectReferences ??= resolveInternalProjectReferences(
       tsConfig,
       context.workspaceRoot,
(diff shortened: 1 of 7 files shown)


packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -578,6 +596,95 @@ function getOutputs(
...
+ * that the `main`, `module`, or `exports` do not include paths from the `rootDir`.
+ * Or if `outFile` is defined, it should not be within the `rootDir`.
+ *
+ * @param tsConfig The TypeScript configuration object.
+ * @param workspaceRoot The workspace root path.
+ * @param projectRoot The project root path.
+ * @param tsConfigPath The path to the TypeScript configuration file.
+ * @returns `true` if the package has a valid build configuration; otherwise, `false`.
+ */
+function isValidPackageJsonBuildConfig(
```
jaysoo, line 610: Nit: `projectExportsFromBuildOutput` or `projectExportsFromSource` would be better names since it's not taking in the `package.json` path as an arg, and it's not technically determining whether the file is valid for a builds or not.

packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -578,6 +596,95 @@ function getOutputs(
...
+  }
+  const packageJson = readJsonFile(
+    joinPathFragments(projectRoot, 'package.json')
+  );
+  const { rootDir, outFile } = tsConfig.options;
+  if (!rootDir) {
+    console.error(
+      `The 'rootDir' option is not defined in the ${tsConfigPath}.`
+    );
+    return false;
```
jaysoo, line 629: We should default it to `src/` since that's what we use across the board. The default `rootDir` for TS is the common folder for source files, excluding `d.ts` files. So in that case it would be aligned with `src` in our setup.

packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -578,6 +596,95 @@ function getOutputs(
...
+  const exports = packageJson?.exports;
+
+  // Checks if the value is a path within the `src` directory.
+  const containsInvalidPath = (
+    value: string | Record<string, string>
+  ): boolean => {
+    if (typeof value === 'string') {
+      return isPathWithinSrc(value);
+    } else if (typeof value === 'object') {
+      return Object.values(value).some((subValue) => {
```
jaysoo, line 662: Let's exclude `types` exports field, as someone could have a non-publishable but buildable library like this:

```
"exports": {
  ".": {
      "types": "./src/index.ts", 
      "default": "./dist/index.js"
  }
}
```

In this case, it's probably for incremental builds, where the build target _has_ to be there, but they can still point types to source.

packages/js/src/plugins/typescript/plugin.ts
ndcunningham: initially, I passed `package.json` in as an arg but refactored and removed it. I should've updated this name as well, will update 👍🏿 .

packages/js/src/plugins/typescript/plugin.ts
ndcunningham: Ahh yea we can default it to be `src` it would remove this error state. I could do a `warn` instead to show we are defaulting. All these would be inside `.nx/workspace-data/` dir.

jaysoo (approved):

PR #29412, opened 2024-12-19
fix(js): handle resolution from within node_module in the sync generator

### LINKED ISSUE #29411: `@nx/js:typescript-sync` doesn't follow the resolution from within `node_modules`
### Current Behavior

When you have tsconfig coming from within `node_modules`, either because they are from a local workspace, or because the tsconfg are in a local package, then the resolution is not following the path, leading to false negative to "is this tsconfig using composite".

This is due to the resolve function from `ts.System` returning the full path, and so including the `workspaceRoot` that is also append by the `FsTree`

### Expected Behavior

`@nx/js:typescript-sync` should follow paths in `node_modules`

### GitHub Repo

https://github.com/beaussan/github-repros/tree/repro-typescript-local-tsconfig-package

### Steps to Reproduce

- Create a typescript based monorepo (`npx create-nx-workspace --preset=ts`)
- Create two libs (`nx g @nx/js:library --directory=example-lib --no-interactive`)
- Run Nx sync => should work and referenced are populated
- Create a lib of ts config, past the content of tsconfig.base in it
- Use it in both libs
- Run Nx Sync => no references are generated

### Nx Report

```shell

Node           : 22.12.0
OS             : darwin-arm64
yarn           : 4.5.3

nx             : 20.2.2
@nx/js         : 20.2.2
@nx/workspace  : 20.2.2
@nx/devkit     : 20.2.2
typescript     : 5.6.3
Registered Plugins:
@nx/js/typescript
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

beaussan:
closed #29411

## Current Behavior

The resolve function from ts.System returning the full path, and so including the workspaceRoot that is also append by the FsTree

## Expected Behavior

Resolution within node_modules are correctly followed

## Related Issue(s)

Fixes #29411

### DIFF
--- a/packages/js/src/generators/typescript-sync/typescript-sync.ts
+++ b/packages/js/src/generators/typescript-sync/typescript-sync.ts
@@ -114,10 +114,20 @@ export async function syncGenerator(tree: Tree): Promise<SyncGeneratorResult> {
   const tsSysFromTree: ts.System = {
     ...ts.sys,
     fileExists(path) {
-      return tsconfigExists(tree, tsconfigInfoCaches, path);
+      // Given ts.System.resolve resolve full path for tsconfig within node_modules
+      // We need to remove the workspace root to ensure we don't have double workspace root within the Tree
+      const correctPath = path.startsWith(tree.root)
+        ? relative(tree.root, path)
+        : path;
+      return tsconfigExists(tree, tsconfigInfoCaches, correctPath);
     },
     readFile(path) {
-      return readRawTsconfigContents(tree, tsconfigInfoCaches, path);
+      // Given ts.System.resolve resolve full path for tsconfig within node_modules
+      // We need to remove the workspace root to ensure we don't have double workspace root within the Tree
+      const correctPath = path.startsWith(tree.root)
+        ? relative(tree.root, path)
+        : path;
+      return readRawTsconfigContents(tree, tsconfigInfoCaches, correctPath);
     },
   };
 


packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -114,10 +115,20 @@ export async function syncGenerator(tree: Tree): Promise<SyncGeneratorResult> {
   const tsSysFromTree: ts.System = {
     ...ts.sys,
     fileExists(path) {
-      return tsconfigExists(tree, tsconfigInfoCaches, path);
+      // Given ts.System.resolve resolve full path for tsconfig within node_modules
+      // We need to remove the workspace root to ensure we don't have double workspace root within the Tree
+      const correctPath = path.startsWith(workspaceRoot)
+        ? path.replace(workspaceRoot, '')
```
leosvelperez, line 121: Let's use `tree.root`:
```suggestion
      const correctPath = path.startsWith(tree.root)
        ? relative(tree.root, path)
```

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -114,10 +115,20 @@ export async function syncGenerator(tree: Tree): Promise<SyncGeneratorResult> {
...
+        : path;
+      return tsconfigExists(tree, tsconfigInfoCaches, correctPath);
     },
     readFile(path) {
-      return readRawTsconfigContents(tree, tsconfigInfoCaches, path);
+      // Given ts.System.resolve resolve full path for tsconfig within node_modules
+      // We need to remove the workspace root to ensure we don't have double workspace root within the Tree
+      const correctPath = path.startsWith(workspaceRoot)
+        ? path.replace(workspaceRoot, '')
```
leosvelperez, line 129: Let's use `tree.root`:
```suggestion
      const correctPath = path.startsWith(tree.root)
        ? relative(tree.root, path)
```

packages/js/src/generators/typescript-sync/typescript-sync.ts
```diff
@@ -5,6 +5,7 @@ import {
   logger,
   parseJson,
   readNxJson,
+  workspaceRoot,
```
leosvelperez, line 8: Let's use `tree.root`:
```suggestion
```

leosvelperez (changes requested):
Thank you! The changes look good to me. Let's just change it to use `tree.root` instead of importing `workspaceRoot`.

leosvelperez (approved):
LGTM!

PR #28119, opened 2024-09-26
feat(core): add support for bun.lock and generating bun lock file

Jordan-Hall:
Fixes: #26640 

This branch made from https://github.com/nrwl/nx/pull/27820 to avoid merge conflicts

This should only be merged once bun added the feature https://x.com/jarredsumner/status/1839056611765268651 tacking https://github.com/oven-sh/bun/issues/11863. - It maybe jsonc so will have to update the generating lock file once it's confirmed

Issue: 
Before `bun.lock` bun only support binary lock files which makes alot of NX flexibility around hashing and creating the lock file impossible  without using bun install.

Solution:
When generating with package lock. We automatically generate `bun.lock` files forcing users to `bun.lock` with a me logged message saying migrated to lock

notes:
This could be deemed as a breaking changes to 3rd party plugins if their relay on any of the package manger detection and lock files

PR #29423, opened 2024-12-19
fix(core): allow text lockfile for bun

ludicast:
## Current Behavior
Nx does not respect a `bun.lock`, which is [the new lockfile](https://bun.sh/blog/bun-lock-text-lockfile) for bun (currently available and to be the default in bun `1.2`)

The text-lockfile will allow dependabot to run, as well as prevent the merge conflicts caused with a binary lockfile.

## Expected Behavior
Nx will respect a `bun.lock`

Note that we are still defaulting to the `bun.lockb` style here for generators, this is just to allow repos containing `bun.lock`.  After `bun.lock` becomes the default there should be a second PR to make it the default for generators.

## Related Issue(s)

Fixes #29422


ludicast:
cc @Jordan-Hall

e2e/utils/get-env-info.ts
```diff
@@ -23,10 +23,11 @@ export function getPublishedVersion(): string {
 }
 
 export function detectPackageManager(dir: string = ''): PackageManager {
-  return existsSync(join(dir, 'bun.lockb'))
-    ? 'bun'
-    : existsSync(join(dir, 'yarn.lock'))
+  return existsSync(join(dir, 'yarn.lock'))
     ? 'yarn'
```
ludicast, line 27: made yarn the first lockfile just to make the bun/pnpm indentation identical for multiple lockfiles

Jordan-Hall: > e bun/pnpm indentation identical for multiple lockfiles

Doesnt bun current support to yarn lock still, This is why bun came frist

Jordan-Hall:
@ludicast  This is a good start but personally I would change the interface. We need to keep bun lock first because bun still goes to yarn with additional parm.

I think its best to include a parser and writer as part of this PR.

I did some work on this in preparation a while ago. Feel free to take a look at interface changes

https://github.com/nrwl/nx/pull/28119

e2e/utils/get-env-info.ts
ludicast: right, great point, will reverse

ludicast:
@Jordan-Hall - if by parser you mean writing something in the vein of https://github.com/nrwl/nx/blob/main/packages/nx/src/plugins/js/lock-file/yarn-parser.ts it def would need to be a follow-on (it's enough code to get caught up in PR purgatory)

Jordan-Hall:
Yes i was because as you can see from the PR i did their loads of work needs to be done. I dont think just adding new vars  is the right idea. If not can you at least do the grunt work without that section

matthewrj:
nx now doesn't work with the default bun settings. Has this PR gone stale or are we working towards getting it merged?

Jordan-Hall:
> nx now doesn't work with the default bun settings. Has this PR gone stale or are we working towards getting it merged?

Whats the current issue. this PR directly wont resolve anything if im not mistaken

JamesHenry:
@Jordan-Hall I believe they are referring to Bun 1.2 and onwards defaulting to the new text based lockfile

JamesHenry:
@ludicast Thanks a lot for this!

packages/nx/src/plugins/js/lock-file/lock-file.ts
```diff
@@ -182,7 +185,7 @@ function getLockFilePath(packageManager: PackageManager): string {
     return NPM_LOCK_PATH;
   }
   if (packageManager === 'bun') {
-    return BUN_LOCK_PATH;
+    return BUN_LOCK_PATH; // defaulting to binary lockfile
```
JamesHenry, line 188: I feel like we should be checking the Bun version and if `>= 1.2` this should be defaulting to the text based one?

JamesHenry (changes requested):

JamesHenry:
@ludicast You will please need to ensure CI is green, right now you have invalid syntax in this PR and the nx package cannot be built

Jordan-Hall:
> @ludicast You will please need to ensure CI is green, right now you have invalid syntax in this PR and the nx package cannot be built

Even so their supporting  both and I think the interface is no longer fit for purpose having single lock files.

JamesHenry:
@Jordan-Hall I understand where you are coming from, but (A) other package managers have had stable lock file names for many years (B) it seems unlikely would go through this churn again any time soon, so implementing some more complex handling just to make it feel like a cleaner implementation is probably overkill at this point. If something similar happened again we could definitely consider a broader refactor.

@ludicast Thanks again for this, to ensure it gets merged as soon as possible I am going to fold the changes into the linked PR

matthewrj:
@Jordan-Hall the issue with the current defaults is I run into this bug: https://github.com/nrwl/nx/issues/29429. I have no binary lockfile, so I get the error 
> Error: Failed to convert JavaScript value `Null` into rust type `String`

PR #29426, opened 2024-12-19
Feat support bun pack

Jordan-Hall:
Current Behavior
Before release last night/today of bun 1.1.27 bun pm pack didn't exists so we fell upon npm pack.

Expected Behavior
If bun 1.1.27 or greater we use bun pm pack otherwise we revert to npm pack

https://x.com/bunjavascript/status/1832386689182122092

cc: @jaysoo


leosvelperez:
Thanks for the contribution!

Unfortunately, `bun pm pack` [creates a tarball of the current workspace](https://bun.sh/docs/cli/pm#pack) (similar to `yarn`), which is not what the current `packageRegistryPack` helper function is intended to do. That helper function is intended to produce a tarball of packages retrieved from the registry.

I'm closing this since the changes wouldn't produce the desired result.

PR #29431, opened 2024-12-19
fix(js): Update typescript plugin check for buildable projects

ndcunningham:
## Current Behavior

Currently, we use `rootDir` to check if the project is buildable. This might not be correct in the case where the transpiled files are inside source.

## Expected Behavior

It should work the for projects as long as the main / exports files are not source files.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/plugins/typescript/plugin.spec.ts
+++ b/packages/js/src/plugins/typescript/plugin.spec.ts
@@ -1696,7 +1696,7 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
       // Sibling package.json
       await applyFilesToTempFsAndContext(tempFs, context, {
         'libs/my-lib/tsconfig.json': `{}`,
-        'libs/my-lib/tsconfig.lib.json': `{}`,
+        'libs/my-lib/tsconfig.lib.json': `{"compilerOptions": {"outDir": "dist"}}`,
         'libs/my-lib/tsconfig.build.json': `{}`,
         'libs/my-lib/package.json': `{}`,
       });
@@ -1744,7 +1744,7 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
       // Sibling package.json
       await applyFilesToTempFsAndContext(tempFs, context, {
         'libs/my-lib/tsconfig.json': `{}`,
-        'libs/my-lib/tsconfig.lib.json': `{}`,
+        'libs/my-lib/tsconfig.lib.json': `{"compilerOptions": {"outDir": "dist"}}`,
         'libs/my-lib/tsconfig.build.json': `{}`,
         'libs/my-lib/package.json': `{}`,
       });
@@ -1794,7 +1794,7 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
       // Sibling package.json
       await applyFilesToTempFsAndContext(tempFs, context, {
         'libs/my-lib/tsconfig.json': `{}`,
-        'libs/my-lib/tsconfig.lib.json': `{"compilerOptions": {"rootDir": "src"}}`,
+        'libs/my-lib/tsconfig.lib.json': `{"compilerOptions": {"outDir": "dist"}}`,
         'libs/my-lib/tsconfig.build.json': `{}`,
         'libs/my-lib/package.json': `{"main": "dist/index.js"}`,
       });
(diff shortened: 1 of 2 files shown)


packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -622,30 +628,48 @@ function isValidPackageJsonBuildConfig(
...
-    );
+  const outDir = tsConfig.options.outDir
+    ? tsConfig.options.outDir
+    : tsConfig.options.outFile
+    ? dirname(tsConfig.options.outFile)
+    : false;
+
+  if (!outDir) {
+    // If we have neither `outDir` nor `outFile` set in the tsconfig file, we assume the project is not configured to be built.
+    return false;
```
jaysoo, line 639: Technically you can build a project, but the output will be siblings of source. Like:

```
src/
  index.ts
  index.js <-- output
```

Although we don't generate our projects this way, someone could have theirs configured as such.

If we don't address this, can we add a comment here to do a follow-up?

packages/js/src/plugins/typescript/plugin.spec.ts
```diff
@@ -2783,10 +2785,9 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
         `);
       });
 
-      it('should add the inline output files when `outDir` is not defined', async () => {
+      xit('should add the inline output files when `outDir` is not defined', async () => {
```
jaysoo, line 2788: If we don't address this, we should add a TODO to do a follow-up.

jaysoo (approved):

packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -622,30 +618,22 @@ function isValidPackageJsonBuildConfig(
...
-    );
-  }
-
-  const isPathWithinSrc = (path: string): boolean => {
-    const resolvedRootDir = resolve(workspaceRoot, projectRoot, rootDir);
-    const pathToCheck = resolve(workspaceRoot, projectRoot, path);
-
-    return pathToCheck.startsWith(resolvedRootDir);
+  const isPathSourceFile = (path: string): boolean => {
+    const ext = extname(path);
```
FrozenPandaz, line 622: Can we not go off of whether or not the path is in the outDIr of the `tsconfig.json` or in the outputs of the build? Rather than the extension?

FrozenPandaz (changes requested):

packages/js/src/plugins/typescript/plugin.ts
ndcunningham: `outDir` is not always available and if it is not the `outDir` would be the same as the source in this scenario we would need to ensure that the file type is not the same as what is inside the source.

packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -622,30 +622,39 @@ function isValidPackageJsonBuildConfig(
...
 
   // If `outFile` is defined, check the validity of the path.
   if (tsConfig.options.outFile) {
-    if (isPathWithinSrc(tsConfig.options.outFile)) {
+    if (isPathSourceFile(tsConfig.options.outFile)) {
       return false;
     }
   }
 
   const buildPaths = ['main', 'module'];
```
leosvelperez, line 655: We should check `exports` before `main` and `module`. From Node docs:

> When a package has an ["exports"](https://nodejs.org/docs/latest-v22.x/api/packages.html#exports) field, this will take precedence over the "main" field when importing the package by name.

See: https://nodejs.org/docs/latest-v22.x/api/packages.html#main

leosvelperez (changes requested):

packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -622,68 +625,80 @@ function isValidPackageJsonBuildConfig(
...
-    }
-  }
-
-  const buildPaths = ['main', 'module'];
-  for (const field of buildPaths) {
-    if (packageJson[field] && isPathWithinSrc(packageJson[field])) {
+    if (isPathSourceFile(tsConfig.options.outFile)) {
       return false;
     }
   }
```
leosvelperez, line 651: I missed this before, but we don't need to check where `outFile` points to. We can remove this block entirely. We only need to check where the `package.json` entry points point in regards to the output directories/files as defined by the tsconfig.

This code block will always return `false`. If `outFile` is defined, then the `outDir` will be `dirname(outFile)` so, calling `isPathSourceFile(outFile)` will check if `outFile` is inside its own directory, which will always be.

leosvelperez (changes requested):

leosvelperez (approved):

PR #29453, opened 2024-12-21
cleanup(core): move esbuild to use tinyglobby

43081j:
An alternative to #28037, switches the esbuild package to use `tinyglobby` instead of `fast-glob`.

This improves performance a fair chunk and reduces install complexity/size.

cc @benmccann @JamesHenry


benmccann (approved):

JamesHenry (changes requested):
Please add the eslint entry as usual to avoid regressions

43081j:
@JamesHenry this should be all good now

JamesHenry (approved):
Thanks a lot @43081j!

PR #29472, opened 2024-12-26
fix(vite): fall back to file matching when resolved file does not exist in `nx-vite-ts-paths` plugin

estebanfelipep:
The fallback path resolution was only happening if the resolvedFile was a falsy value, so it was returning inexistent filepaths.

This pull request includes a small but important change to the `packages/vite/plugins/nx-tsconfig-paths.plugin.ts` file. The change improves the reliability of resolving file paths by checking if the resolved path exists before proceeding with fallback file matching.

* [`packages/vite/plugins/nx-tsconfig-paths.plugin.ts`](diffhunk://#diff-ad026b24ed45d9df484cabc7e277fc9b4d7759560af36bf357cbc4186725ae0bL174-R176): Added a check to verify if the resolved file path exists before using fallback file matching.
* 

## Current Behavior

The resolver doesn't resolve the full file path, it ignores the file extension. For example, for a path like this:
`import mergeClassNames from '@projects/global/utils/mergeClassNames'`

It resolves to 
`absolutePath.../projects/libs/global/src/utils/mergeClassNames`

When vite is building a project, it doesn't find the file and it errors.

This happens because the resolved file path is not a falsy value, thus, it doesn't run the fallback file path matching.

## Expected Behavior

What is expected is that it resolves to (notice the file extension)
`absolutePath.../projects/libs/global/src/utils/mergeClassNames.ts`


leosvelperez (approved):
Thanks for the contribution!

PR #29490, opened 2024-12-30
docs(nx-cloud): use assignment rules with self provisioned agents

lourw:
## Current Behavior

The initial version of the docs only allow the use of assignment rules with Nx Agents.

## Expected Behavior

The updated docs include an overview and example of how to configure assignment rules to work with self provisioned agents (and DTE in 'manual' mode).

## Related Issue(s)

Fixes #


isaacplmann (approved):

lourw:
@isaacplmann I added some more details and edited some of the grammar. Can I just get one more once-over from you?

docs/nx-cloud/reference/assignment-rules.md
```diff
@@ -1,83 +1,35 @@
...
-    runs-on:
-      - linux-large-js
-
-  - project: lib1
-    target: test
-    runs-on:
-      - linux-medium-js
-```
+{% callout type="note" title="If you are using self-hosted agents, you must define your own agent types" %}
+You can define your own agent types and attach them to your self-hosted agents using the `NX_AGENT_LAUNCH_TEMPLATE` environment variable. Ensure that for each `runs-on` field in your assignment rules, you have corresponding agents in your agent pool that have the same agent type.
```
isaacplmann, line 30: Is it "you must" or "you can"?  One of these needs to be changed

docs/nx-cloud/reference/assignment-rules.md
```diff
@@ -122,7 +72,110 @@ assignment-rules:
...
+    runs-on:
+      - linux-medium
+
+  - configuration: development
+    runs-on:
+      - linux-medium
+      - linux-large
+```
+
+Note that the agent types supplied in the `runs-ons` property will be used to determine which agents will have rules applied to them.
```
isaacplmann, line 98:
```suggestion
Note that the agent types supplied in the `runs-on` property will be used to determine which agents will have rules applied to them.
```

isaacplmann (approved):

PR #29531, opened 2025-01-06
fix(js): infer outputs correctly when both rootDir and outDir are set for tsconfig

jaysoo:
When the tsconfig has `rootDir` and `outDir` both defined, then the `*.tsbuildinfo` file is not cached. This makes incremental tsc not work through distribution (e.g. agents).

For example, given this `tsconfig.lib.json` file:

```json
{
  "compilerOptions": {
    "outDir": "out-tsc/lib-1",
    "rootDir": "src"
  }
}
```

The outputs (e.g. `*.d.ts` files) are under `{projectRoot}/out-tsc/lib-1`, but the tsbuild info file is under `{projectRoot}/out-tsc/tsconfig.lib.tsbuildinfo`.

## Current Behavior
tsbuildinfo file is not cached

## Expected Behavior
tsbuildinfo file is cached

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/plugins/typescript/plugin.ts
+++ b/packages/js/src/plugins/typescript/plugin.ts
@@ -571,14 +571,14 @@ function getOutputs(
           join(workspaceRoot, projectRoot)
         );
         outputs.add(
-          joinPathFragments(
-            pathToInputOrOutput(
+          pathToInputOrOutput(
+            joinPathFragments(
               config.options.outDir,
-              workspaceRoot,
-              projectRoot
+              relativeRootDir,
+              `*.tsbuildinfo`
             ),
-            relativeRootDir,
-            `*.tsbuildinfo`
+            workspaceRoot,
+            projectRoot
           )
         );
       }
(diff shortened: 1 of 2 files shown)


packages/js/src/plugins/typescript/plugin.ts
```diff
@@ -549,18 +549,36 @@ function getOutputs(
...
-            workspaceRoot,
-            projectRoot
+          joinPathFragments(
+            pathToInputOrOutput(
+              config.options.outDir,
+              workspaceRoot,
+              projectRoot
+            ),
+            relativeRootDir,
+            `*.tsbuildinfo`
```
leosvelperez, line 581: Rather than joining the result of `pathToInputOrOutput` with the relative path to the tsbuildinfo file, we should instead join first to get the full path and provide that to `pathToInputOrOutput`.
```suggestion
          pathToInputOrOutput(
            joinPathFragments(
              config.options.outDir,
              relativeRootDir,
              `*.tsbuildinfo`
            ),
            workspaceRoot,
            projectRoot
```

leosvelperez (approved):

PR #29534, opened 2025-01-07
fix(linter): add `cjs`, `cts`, `mjs`, and `mts` to the default extensions for the inference plugin

beaussan:
## Current Behavior

On the default generated eslint on a ts reference workspace, the config allows for linting of cjs and mjs files

```js
const nx = require('@nx/eslint-plugin');

module.exports = [
  ...nx.configs['flat/base'],
  ...nx.configs['flat/typescript'],
  ...nx.configs['flat/javascript'],
  {
    ignores: ['**/dist'],
  },
  {
    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx'],
    rules: {
      '@nx/enforce-module-boundaries': [
        'error',
        {
          enforceBuildableLibDependency: true,
          allow: ['^.*/eslint(\\.base)?\\.config\\.[cm]?js$'],
          depConstraints: [
            {
              sourceTag: '*',
              onlyDependOnLibsWithTags: ['*'],
            },
          ],
        },
      ],
    },
  },
  {
    files: [
      '**/*.ts',
      '**/*.tsx',
      '**/*.js',
      '**/*.jsx',
      '**/*.cjs',
      '**/*.mjs',
    ],
    // Override or add rules here
    rules: {},
  },
];

```

However, the default glob is not matching them

## Expected Behavior
All cjs and mjs files are included in the lint target detection


leosvelperez (approved):
Thank you!

I updated it to also include `cts` and `mts`.

PR #29535, opened 2025-01-07
feat(module-federation): upgrade to latest version of @module-federation/enhanced

Coly010:
## Current Behavior
We currently use version `0.7.6` of `@module-federation/enhanced`.
This version does not contain the `externalRuntime` experiment to reduce the remoteEntry bundle size.

## Expected Behavior
Upgrade Nx dependency of `@module-federation/enhanced` to version `0.8.8`.


packages/module-federation/migrations.json
```diff
@@ -1 +1,25 @@
...
+          "alwaysAddToPackageJson": false
+        },
+        "@module-federation/sdk": {
+          "version": "0.8.8",
+          "alwaysAddToPackageJson": false
+        },
+        "@module-federation/node": {
+          "version": "2.6.21",
+          "alwaysAddToPackageJson": false
+        }
```
leosvelperez, line 21: Do they need to be pinned versions?

The migration version for `@module-federation/node` would pin it, while new projects would get the range `^2.6.21` based on the constant in the `versions.ts` file.

The other three packages match their constant but don't match the package.json dependencies where they are set to the range `^0.8.8`.

Coly010: No, they can be a range, I'll update

packages/angular/src/utils/versions.ts
```diff
@@ -14,8 +14,8 @@ export const typesCorsVersion = '~2.8.5';
 export const expressVersion = '~4.18.2';
 export const typesExpressVersion = '4.17.14';
 export const browserSyncVersion = '^3.0.0';
-export const moduleFederationNodeVersion = '~2.6.11';
-export const moduleFederationEnhancedVersion = '0.7.6';
+export const moduleFederationNodeVersion = '^2.6.21';
+export const moduleFederationEnhancedVersion = '0.8.8';
```
leosvelperez, line 18:
```suggestion
export const moduleFederationEnhancedVersion = '^0.8.8';
```

leosvelperez (approved):

jaysoo (approved):

JamesHenry (approved):

PR #29546, opened 2025-01-08
fix(js): ensure js libraries' build produce esm output in ts solution setup

leosvelperez:
- Update js libraries for bundlers `esbuild`, `swc`, and `tsc` to produce ESM output when using the TS solution setup.
- Fix `esbuild` and `swc` executors so they generate declaration files even when skipping type-checking.
- Add `cjs` and `cts` to the `ignoredFiles` pattern for the rollup config file in the eslint config.
- Ensure running an install after a js library is generated when using the TS solution setup.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/utils/typescript/run-type-check.ts
+++ b/packages/js/src/utils/typescript/run-type-check.ts
@@ -22,6 +22,7 @@ interface BaseTypeCheckOptions {
   incremental?: boolean;
   rootDir?: string;
   projectRoot?: string;
+  ignoreDiagnostics?: boolean;
 }
 
 type Mode = NoEmitMode | EmitDeclarationOnlyMode;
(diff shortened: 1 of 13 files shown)


packages/js/src/utils/typescript/run-type-check.ts
```diff
@@ -103,9 +106,9 @@ export async function runTypeCheck(
 
   const result = program.emit();
 
-  const allDiagnostics = ts
-    .getPreEmitDiagnostics(program as Program)
-    .concat(result.diagnostics);
+  const allDiagnostics = options.ignoreDiagnostics
```
jaysoo, line 109: Makes sense for correctness. We should look at it again in the future to see if want to allow builds without typechecks, maybe a separate `typedefs` task for swc and esbuild so users that do not need `d.ts` files don't need to pay the penalty.

jaysoo (approved):

PR #29552, opened 2025-01-08
feat(core): allow tasks to run with a substring of project name

jaysoo:
When projects use only `package.json` and not `project.json`, we need to set the simple name in the `nx` property of `package.json`. This isn't ideal because it's yet another Nx-specific thing that needs to be configured when our goal is to reduce boilerplate.

This PR allows users to pass a substring that matches exactly one project when running a task.

For example, if `@acme/foo` is the name in `package.json`, then running `nx build foo` will match it.

If more than one projects match, then an error is thrown showing all the matched projects, and the user has to be more specific or type in the fully qualified name.

## Current Behavior
Users need to pass exact matches for project names when running tasks.

## Expected Behavior
User can pass a substring matching the project name.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/run/run-one.ts
+++ b/packages/nx/src/command-line/run/run-one.ts
@@ -21,6 +21,7 @@ import { TargetDependencyConfig } from '../../config/workspace-json-project-json
 import { readNxJson } from '../../config/configuration';
 import { calculateDefaultProjectName } from '../../config/calculate-default-project-name';
 import { generateGraph } from '../graph/graph';
+import { findMatchingProjects } from '../../utils/find-matching-projects';
 
 export async function runOne(
   cwd: string,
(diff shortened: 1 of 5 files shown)


AgentEnder (approved):

packages/nx/src/command-line/run/run-one.ts
```diff
@@ -84,27 +87,62 @@ export async function runOne(
...
   }
 }
 
-function getProjects(projectGraph: ProjectGraph, project: string): any {
-  if (!projectGraph.nodes[project]) {
-    output.error({
-      title: `Cannot find project '${project}'`,
-    });
-    process.exit(1);
+function getProjects(
```
FrozenPandaz, line 98: What if this function used `findMatchingProjects` and then checks if the resulting matching projects is a set of a single project? :eyes:

FrozenPandaz (changes requested):

packages/nx/src/command-line/run/run-one.ts
jaysoo: Done in both `run-one.ts` and `project.ts`

packages/nx/src/utils/find-matching-projects.ts
```diff
@@ -167,6 +167,17 @@ function addMatchingProjectsByName(
   }
 
   if (!isGlobPattern(pattern.value)) {
+    const regex = new RegExp(`\\b${pattern.value}\\b`);
```
FrozenPandaz, line 170: Apparently `\b` does not include underscores... should we make it include `_` as a boundary?

Rust projects in particular have a lot of `_`

FrozenPandaz (changes requested):

packages/nx/src/utils/find-matching-projects.ts
jaysoo: Yes, we should keep `_`. I'll udpate it.

FrozenPandaz (approved):

PR #29554, opened 2025-01-08
feat(nx-dev): add trial pages for Powerpack and Enterprise, update contact sales team page

ndcunningham:


yuzupartycat:
@ndcunningham Taking another look, I think we should have the name of the products on one line instead of broken up. Can we have Nx Powerpack together on the second line. Same for the Nx Enterprise trial page. 
image
image

ndcunningham:
> @ndcunningham can we re-upload the PDF, but change the file name before to remove the "Enterprise - CTA" section. I didn't realize it would show here, but that should just be for internal use. image



@yuzupartycat I think we need a new PDF with the **title** changed. The text you referred to is not the file change as that has been changed to `fast-ci-whitepaper.pdf`

vsavkin (approved):

PR #29561, opened 2025-01-09
cleanup(testing): migrate jest to `picocolors`

Phillip9587:
migrates `@nx/jest` from `chalk` to `picocolors`

Part of https://github.com/es-tooling/ecosystem-cleanup/issues/117

## Related Issue(s)

Fixes #


packages/jest/.eslintrc.json
```diff
@@ -4,13 +4,28 @@
...
       "rules": {
-        "no-restricted-imports": ["error", "@nx/workspace"]
+        "no-restricted-imports": [
+          "error",
+          "@nx/workspace",
+          {
+            "name": "chalk",
+            "message": "Please use `picocolors` in place of `chalk` for rendering terminal colors"
+          }
+        ]
```
leosvelperez, line 28: Let's move `excludedFiles` and the rule definition here to the block for `"files": ["*.ts", "*.tsx", "*.js", "*.jsx"]`. So, we have the rule in a single block that includes all js/ts files except migrations.

After that, this block should have empty rules and no `excludedFiles`. We'll later clean up some boilerplate in these config files, but for now, just ensuring we don't duplicate that rule needlessly is enough.

leosvelperez (changes requested):
Thank you!

Left just one comment to simplify the eslint config.

JamesHenry (approved):

packages/jest/.eslintrc.json
Phillip9587: I looked into it and we don't need the `excludedFiles` anymore because the old migrations using `@nx/workspace` don't exist anymore.

leosvelperez (approved):

PR #29565, opened 2025-01-09
fix(js): generate js libs with exports in package.json and ensure esm output when using rollup bundler

leosvelperez:
- Ensure libs are generated with `exports` in `package.json`
- Generate `types` instead of `typings` in package.json
- Update js lib with rollup to only output esm
- Update `tsconfig.spec.json` for js libraries with rollup to set `module: esnext` and `moduleResolution: bundler` (they use `@swc/jest`)
- Fix `@nx/js/typescript` issue with absolute paths when normalizing inputs/outputs
- Fix `@nx/js/typescript` issue identifying buildable libs
- Fix express app generator not installing `@types/express`

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/generators/library/library.ts
+++ b/packages/js/src/generators/library/library.ts
@@ -1,12 +1,12 @@
 import {
   addDependenciesToPackageJson,
-  installPackagesTask,
   addProjectConfiguration,
   ensurePackage,
   formatFiles,
   generateFiles,
   GeneratorCallback,
   getPackageManagerCommand,
+  installPackagesTask,
   joinPathFragments,
   names,
   offsetFromRoot,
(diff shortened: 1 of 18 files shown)


packages/js/src/generators/library/library.ts
```diff
@@ -1094,56 +1134,59 @@ function determineEntryFields(
...
-          : './index.cjs',
-        typings: options.isUsingTsSolutionConfig
-          ? './dist/index.d.ts'
-          : './index.d.ts',
-      };
+      if (options.isUsingTsSolutionConfig) {
+        return {
+          type: 'module',
+          main: './dist/index.js',
+          module: './dist/index.js',
```
jaysoo, line 1180: I think we can skip `module` field since `type: module` implies that `main` is ESM. The `module` field is only useful in older bundler setups that use `main` for CJS and `module` for ESM. However, modern bundlers shouldn't need it anymore.

jaysoo (approved):

packages/js/src/generators/library/library.ts
jaysoo, line 1180:
```suggestion
```

PR #29568, opened 2025-01-09
cleanup(react-native): use native fetch for isPackagerRunning check and remove `node-fetch` dependency

Phillip9587:
Removes the `node-fetch` dependency and uses the native fetch global for the `isPackagerRunning()` check

## Current Behavior

- Dependency `node-fetch` is used.

## Expected Behavior

- Removed the `node-fetch` dependency.
- Used the native `fetch` global.

## Related Issue(s)

Fixes #

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,7 +9,21 @@
   "extends": ["plugin:storybook/recommended"],
   "rules": {
     "@typescript-eslint/explicit-module-boundary-types": "off",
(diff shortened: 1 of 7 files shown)


Coly010 (approved):

JamesHenry (changes requested):
Let's add the usual lint check to ensure other folks who have not seen this do not reintroduce node-fetch.

In this case, we can add it to the root eslint config I believe

JamesHenry (approved):

PR #29572, opened 2025-01-09
fix(gradle): check if java is installed

xiongemi:
## Expected Behavior

throw an error when `./gradlew wrapper` or `java -version` failed to run:
Screenshot 2025-01-09 at 5 33 38 PM
Screenshot 2025-01-09 at 5 32 20 PM

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/utils/exec-gradle.ts
+++ b/packages/gradle/src/utils/exec-gradle.ts
@@ -23,7 +23,7 @@ export function execGradleAsync(
   args: ReadonlyArray<string>,
   execOptions: ExecFileOptions = {}
 ): Promise<Buffer> {
-  return new Promise<Buffer>((res, rej) => {
+  return new Promise<Buffer>((res, rej: (stdout: Buffer) => void) => {
     const cp = execFile(gradleBinaryPath, args, {
       cwd: dirname(gradleBinaryPath),
       shell: true,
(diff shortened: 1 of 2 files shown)


packages/gradle/src/utils/get-project-report-lines.ts
```diff
@@ -32,6 +33,38 @@ export async function getProjectReportLines(
...
+        [
+          gradlewFile,
+          new Error(
+            `Could not find Gradle. Please install Gradle and try again: https://gradle.org/install/. ${e.message}`
+          ),
+        ],
+      ],
+      []
+    );
+  }
```
FrozenPandaz, line 66: Can we achieve this without running extra commands?

Can we check the error that comes back from the `exec` within `execGradleAsync`?

FrozenPandaz (changes requested):

packages/gradle/src/utils/exec-gradle.ts
```diff
@@ -23,7 +23,7 @@ export function execGradleAsync(
   args: ReadonlyArray<string>,
   execOptions: ExecFileOptions = {}
 ): Promise<Buffer> {
-  return new Promise<Buffer>((res, rej) => {
+  return new Promise<Buffer>((res, rej: (stdkout: Buffer) => void) => {
```
FrozenPandaz, line 26:
```suggestion
  return new Promise<Buffer>((res, rej: (stdout: Buffer) => void) => {
```

FrozenPandaz (approved):

PR #29577, opened 2025-01-09
fix(core): support subpath exports when constructing the project graph

### LINKED ISSUE #29486: NX Packages without default export are excluded from dependencies
### Current Behavior

When a project declares an [exports](https://nodejs.org/api/packages.html#exports) field and does _not_ include a `"."` export, it will be ignored as a dependency.

e.g.
```json
// foo/project.json
{
    "name": "foo",
    "exports": {
        "./foo": "./foo.js"
    }
}
```
and 
```json
// bar/project.json
{
    "name": "bar",
    "dependencies": {
       "bar": "workspace:^"
    }
}
```

Nx will not respect the dependency that `bar` has on `foo`, despite the explicit declaration in package.json.

This is reflected in both the Web UI and graph output, as well as he `dependencies` field for project graph from `@nx/devkit`.

```ts
import { createProjectGraphAsync } from '@nx/devkit';

const projectGraph = await createProjectGraphAsync();

projectGraph.dependencies['bar']
// [{ source: 'bar', target: 'npm:foo', type: 'static' }]
// The `npm:` prefix indicates an external dependency, despite it being a local dependency
```

### Expected Behavior

When projects depend on each otheir via package.json (`pnpm`'s `"workspace:^"` specifier is used as an example, but is hardly the limit), Nx should respect that dependency when building the project graph.

That means that the project graph should explicitly call out the dependency as seen the web UI (or a JSON output), and the `dependencies` field for project graph from `@nx/devkit` should also declare the project as a local dependency.

e.g. for example `bar`'s dependency on `foo` above
```ts
import { createProjectGraphAsync } from '@nx/devkit';

const projectGraph = await createProjectGraphAsync();

projectGraph.dependencies['bar']
// [{ source: 'bar', target: 'foo', type: 'static' }]
```

### GitHub Repo

https://github.com/JacobLey/issue-recreator/tree/nx-exports-dependency

### Steps to Reproduce

1. Create a project ("foo") that includes an [exports](https://nodejs.org/api/packages.html#exports) field in the package.json. Export any path other than `"."` (I suspect this issue would also occur if the exports object is empty)
2. Create a second project ("bar") that depends on the first. This dependency can be via traditional `dependencies` field
3. Run `nx graph` and inspect the apparent lack of dependency of foo from bar.

### Nx Report

```shell
Node           : 22.12.0
OS             : linux-arm64
Native Target  : aarch64-linux
pnpm           : 9.15.2
```

### Failure Logs

```shell
No explicit failure logs.

Just omission of project as dependency.

Any "failures" would manifest because of tasks being executed out of order and dependencies not being available
```

### Package Manager Version

9.15.2

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Context is also shared in the issue recreation, which helps demonstrate the fact that this dependency is only omitted when using `exports` that don't include the `"."` field.

Repeated here for simplicity:

The logic to declare a dependency as "local" as opposed to "external" is implemented in [this line](https://www.npmjs.com/package/nx-update-ts-references).

```ts
const localProject = targetProjectLocator.findDependencyInWorkspaceProjects(d);
```

The implementation logic for this method is [fairly straightforward](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts#L257).
If the name of the project exists in the map, return in it

```ts
return this.packageEntryPointsToProjectMap[dep]?.name ?? null;
```

However it will also calculate this map (memoized), which [exposes the bug](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/utils/packages.ts#L18-L24):
```ts
if (!packageExports || typeof packageExports === 'string') {
    // no `exports` or it points to a file, which would be the equivalent of
    // an '.' export, in which case the package name is the entry point
    result[packageName] = project;
} else {
    for (const entryPoint of Object.keys(packageExports)) {
        result[join(packageName, entryPoint)] = project;
    }
}
```

`result['foo']` will get written because `join('foo', '.')` returns `foo`.
`result['foobar']` would also work because it lacks an exports field.

Howecer `result['bar/bar']` is the result of bar because it only has an export for that path.
But the dependency is checked against `'bar'`, and therefore appears to be an external dependency.

robbtraister:
Sibling dependencies that rely exclusively on subpath exports are excluded from the dependency graph because there is no exact match.

This adds a fallback to look for subpath exports if the exact match is not found.

This also adds logic to respect conditional exports independent from subpath exports.

## Current Behavior

Importing a workspace dependency via subpath export fails to match the package name, and so is not included in the dependency graph.

### Example

```apps/api/package.json```
```json
  "name": "@my-org/api",
  "dependencies": {
    "@my-org/services": "workspace:*"
  }
```

```libs/services/package.json```
```json
  "name": "@my-org/services",
  "exports": {
    "./email": "./dist/email.js"
  }
```

The `@my-org/api` app should be able to import the email service with `import { EmailService } from "@my-org/services/email"`.

However, the `getPackageEntryPointsToProjectMap` implementation results in an object with a key of `@my-org/services/email`, but not `@my-org/services`. This is not specifically a problem, except that `findDependencyInWorkspaceProjects` only considers exact matches within those object keys.

## Expected Behavior

Importing a workspace dependency via subpath export should be included in the dependency graph.

I also addressed a related issue where the following resulted in keys of `@my-org/services/default` and `@my-org/services/types`, which is incorrect according to the subpath/conditional export rules.
```json
  "exports": {
    "default": "./dist/index.js",
    "types": "./dist/index.d.ts"
  }
```

## Related Issue(s)

Fixes #29486

### DIFF
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
@@ -263,8 +263,8 @@ export class TargetProjectLocator {
       this.packageEntryPointsToProjectMap[dep]?.name ??
       // if the package exports do not include ".", look for subpath exports
       Object.entries(this.packageEntryPointsToProjectMap).find(([entryPoint]) =>
-        entryPoint.startsWith(`${dep}/`)
-      )?.[1].name ??
+        dep.startsWith(`${entryPoint}/`)
+      )?.[1]?.name ??
       null
     );
   }
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
@@ -259,7 +259,14 @@ export class TargetProjectLocator {
       this.nodes
     );
 
-    return this.packageEntryPointsToProjectMap[dep]?.name ?? null;
+    return (
+      this.packageEntryPointsToProjectMap[dep]?.name ??
+      // if the package exports do not include ".", look for subpath exports
+      Object.entries(this.packageEntryPointsToProjectMap).find(([entryPoint]) =>
+        entryPoint.startsWith(`${dep}/`)
+      )?.[1].name ??
+      null
+    );
   }
 
   private resolveImportWithTypescript(
--- a/packages/nx/src/plugins/js/utils/packages.ts
+++ b/packages/nx/src/plugins/js/utils/packages.ts
@@ -21,7 +21,14 @@ export function getPackageEntryPointsToProjectMap<
       result[packageName] = project;
     } else {
       for (const entryPoint of Object.keys(packageExports)) {
-        result[join(packageName, entryPoint)] = project;
+        // if entrypoint begins with '.', it is a relative subpath export
+        // otherwise, it is a conditional export
+        // https://nodejs.org/api/packages.html#conditional-exports
+        if (entryPoint.startsWith('.')) {
+          result[join(packageName, entryPoint)] = project;
+        } else {
+          result[packageName] = project;
+        }
       }
     }
   }
--- a/packages/nx/src/plugins/js/utils/packages.ts
+++ b/packages/nx/src/plugins/js/utils/packages.ts
@@ -15,7 +15,11 @@ export function getPackageEntryPointsToProjectMap<
     }
 
     const { packageName, packageExports } = metadata.js;
-    if (!packageExports || typeof packageExports === 'string') {
+    if (
+      !packageExports ||
+      typeof packageExports === 'string' ||
+      !Object.keys(packageExports).length
+    ) {
       // no `exports` or it points to a file, which would be the equivalent of
       // an '.' export, in which case the package name is the entry point
       result[packageName] = project;
@@ -29,6 +33,10 @@ export function getPackageEntryPointsToProjectMap<
         } else {
           result[packageName] = project;
         }
+      }
+      // if there was no '.' entrypoint, ensure the package name is matched with the project
+      if (!result[packageName]) {
+        result[packageName] = project;
       }
     }
   }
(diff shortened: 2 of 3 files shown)


leosvelperez (approved):
Thanks for the contribution!

I updated it to include some extra changes to cover more scenarios and added test coverage to ensure we don't regress.

jaysoo (approved):

packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -259,7 +259,14 @@ export class TargetProjectLocator {
       this.nodes
     );
 
-    return this.packageEntryPointsToProjectMap[dep]?.name ?? null;
+    return (
+      this.packageEntryPointsToProjectMap[dep]?.name ??
+      // if the package exports do not include ".", look for subpath exports
+      Object.entries(this.packageEntryPointsToProjectMap).find(([entryPoint]) =>
+        dep.startsWith(`${entryPoint}/`)
+      )?.[1]?.name ??
```
robbtraister, line 267: @leosvelperez 
I don't think this part is necessary, now that you've added the logic to ensure the `.` entrypoint is defined in the map.

packages/nx/src/plugins/js/utils/packages.ts
```diff
@@ -15,13 +15,28 @@ export function getPackageEntryPointsToProjectMap<
...
+        // https://nodejs.org/api/packages.html#conditional-exports
+        if (entryPoint.startsWith('.')) {
+          result[join(packageName, entryPoint)] = project;
+        } else {
+          result[packageName] = project;
+        }
+      }
+      // if there was no '.' entrypoint, ensure the package name is matched with the project
+      if (!result[packageName]) {
+        result[packageName] = project;
```
robbtraister, line 39: I considered an approach like this, but I wasn't sure if this map was used elsewhere and this could be misleading.

leosvelperez: Yeah, this is only used to identify dependencies when building the project graph, so it doesn't need to be that strict. If the user has a wrong import path that is not allowed by the definition of exports, the build tool will complain, and they'll have to fix it anyway.

PR #29580, opened 2025-01-10
feat(nx-dev): generate migration detail pages

isaacplmann:
Generates list of migrations on the plugin overview page and a standalone `/migrations` page.

To add sample code changes for a migration that has an implementation file, create a `.md` file with the same name as the implementation file in the same folder as the implementation file.  i.e. `move-cache-directory.md` for `move-cache-directory.ts`.

Migrations that have `packages` defined will have a table generated with the package updates listed.

Separate PRs will be created to add sample code changes for each migration with an implementation.

The migration list on the plugin overview page: [Angular migrations](https://nx-dev-git-docs-migration-details-nrwl.vercel.app/nx-api/angular#migrations)
Standalone migration list page: [Angular migrations](https://nx-dev-git-docs-migration-details-nrwl.vercel.app/nx-api/angular/migrations)
Sample migration with added markdown file details: [17.0.0-move-cache-directory](https://nx-dev-git-docs-migration-details-nrwl.vercel.app/nx-api/nx#1700movecachedirectory)
Sample migration with only package updates: [Angular 20.4.0](https://nx-dev-git-docs-migration-details-nrwl.vercel.app/nx-api/angular#2040packageupdates)
Sample migration without any markdown file details: [update-angular-cli-version-19-1-0](https://nx-dev-git-docs-migration-details-nrwl.vercel.app/nx-api/angular#updateangularcliversion1910)
- This last sample is very bare-bones and the reason why we need these pages in the first place.  People don't know what migrations are actually doing.  Follow up PRs will address pages like this.


docs/generated/manifests/nx-api.json
```diff
@@ -434,6 +434,701 @@
         "type": "generator"
       }
     },
+    "migrations": {
+      "/nx-api/angular/migrations/remove-library-generator-simple-module-name-option": {
```
jaysoo, line 438: I think these should be sorted from latest to earliest. We might also want to make the `package.json` updates more clear, like `20.2.0 (update packages)` rather than just `20.2.0`.

leosvelperez:
@isaacplmann do we plan to render the requirements some migrations have? I think it would make sense to do so as it adds clarity about why some migrations might run or not. Examples of those requirements:

- package update `requires`: https://github.com/nrwl/nx/blob/master/packages/angular/migrations.json#L1322-L1328
- migration generator `requires`: https://github.com/nrwl/nx/blob/master/packages/angular/migrations.json#L330-L332

Coly010 (approved):

jaysoo (approved):

FrozenPandaz (approved):

PR #29584, opened 2025-01-10
fix(react): setup mf env var as input for rspack

Coly010:
## Current Behavior
The Module Federation `NxRuntimeLibraryControlPlugin` relies on an environment variable that is set during the build/serve process.
This could potentially lead to issues with cache restoration.

The impact should be minimal as the runtime control plugin should only be added to the module federation config when the env var is set.
It should only be set by the `module-federation-dev-server` executor which is invoked during serve - a non-cacheable task already.

## Expected Behavior
Ensure the env var is set as an input to ensure maximum accuracy with module federation builds via rspack executor

### DIFF
--- a/packages/react/src/migrations/update-18-0-0/add-mf-env-var-to-target-defaults.ts
+++ b/packages/react/src/migrations/update-18-0-0/add-mf-env-var-to-target-defaults.ts
@@ -8,10 +8,11 @@ import {
 import { addMfEnvToTargetDefaultInputs } from '../../utils/add-mf-env-to-inputs';
 
 export default async function (tree: Tree) {
-  if (!hasModuleFederationProject(tree)) {
+  const bundler = hasModuleFederationProject(tree);
+  if (!bundler) {
     return;
   }
-  addMfEnvToTargetDefaultInputs(tree);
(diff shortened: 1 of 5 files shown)


packages/react/src/migrations/update-18-0-0/add-mf-env-var-to-target-defaults.ts
```diff
@@ -8,10 +8,11 @@ import {
 import { addMfEnvToTargetDefaultInputs } from '../../utils/add-mf-env-to-inputs';
 
 export default async function (tree: Tree) {
-  if (!hasModuleFederationProject(tree)) {
+  const bundler = hasModuleFederationProject(tree);
```
jaysoo, line 11: Should there be another migration for 20.4 for projects created after v18?

Coly010: There is :) 
https://github.com/nrwl/nx/pull/29584/files#diff-7752a3351af4cb540782fcaf8e3f320256efd9abc08ff467b466f9ec6b3b7402R50-R55

jaysoo (approved):

PR #29585, opened 2025-01-10
fix(vite): plugin should infer 'vite' for dev/serve command

Coly010:
## Current Behavior
We currently have a `serveTargetName` that defaults to `serve` in the `@nx/vite/plugin`.
This infers a `vite serve` command for the `serve` task to vite projects.

While not incorrect, it could be abrasive for users coming from the `vite` ecosystem to Nx.
The command to start the Vite Dev Server is `vite`, so we should infer this.

`create-vite` also creates a package.json script with `"dev": "vite"` meaning users are likely running `npm run dev`.

This creates two points of differences for vite ecosystem users.

## Expected Behavior
Deprecate `serveTargetName` in favour of `devTargetName` to more closely align with lanugage from the Vite ecosystem.

Infer the command `vite` instead of `vite serve` for the `serve` and `dev` tasks.

### DIFF
--- a/packages/vite/src/plugins/plugin.spec.ts
+++ b/packages/vite/src/plugins/plugin.spec.ts
@@ -321,6 +321,26 @@ describe('@nx/vite/plugin', () => {
                         "^build",
                       ],
                     },
+                    "dev": {
+                      "command": "vite",
+                      "metadata": {
+                        "description": "Starts Vite dev server",
+                        "help": {
+                          "command": "npx vite --help",
+                          "example": {
+                            "options": {
+                              "port": 3000,
+                            },
+                          },
+                        },
+                        "technologies": [
+                          "vite",
+                        ],
+                      },
+                      "options": {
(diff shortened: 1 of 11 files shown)


packages/vite/src/plugins/plugin.spec.ts
```diff
@@ -316,6 +316,26 @@ describe('@nx/vite/plugin', () => {
                         "{workspaceRoot}/dist/{projectRoot}",
                       ],
                     },
+                    "dev": {
+                      "command": "vite",
```
jaysoo, line 320: Maybe we should keep `vite dev`? What happens during `nx init` or `nx import` when we're replacing package scripts?

Coly010: This is going to require some changes to `updatePackageScripts` logic

jaysoo (approved):

PR #29609, opened 2025-01-13
feat(js): infer build-deps and watch-deps targets for incremental builds

jaysoo:
This PR adds `build-deps` and `watch-deps` targets to buildable JS projects to help with incremental builds.

A use-case for this is if an app (e.g. Vite React app) has buildable dependencies that need to be rebuilt when they change.

Say, you create a React app and lib as follows:

```
nx g @nx/react:app apps/react-app --bundler vite 
nx g @nx/react:lib packages/react-lib --bundler vite
```

And import `react-lib` inside the app.

```jsx
import { ReactLib } from '@acme/react-lib';
//...
return <ReactLib />
```

The user can then run:

```
nx watch-deps react-app
```

And then serve the app in another terminal:
```
nx serve react-app
```

Then whenever code is updated for a buildable dependency, it'll be rebuilt and then reloaded in the app.

### DIFF
--- a/packages/js/src/plugins/typescript/util.ts
+++ b/packages/js/src/plugins/typescript/util.ts
@@ -1,0 +1,41 @@
+import { readJsonFile, type TargetConfiguration } from '@nx/devkit';
+import { existsSync } from 'node:fs';
+import { type PackageManagerCommands } from 'nx/src/utils/package-manager';
+import { join } from 'path';
+
+/**
+ * Allow uses that use incremental builds to run `nx watch-deps` to continuously build all dependencies.
+ */
+export function addBuildAndWatchDepsTargets(
+  workspaceRoot: string,
+  projectRoot: string,
+  targets: Record<string, TargetConfiguration>,
+  options: { buildDepsTargetName?: string; watchDepsTargetName?: string },
+  pmc: PackageManagerCommands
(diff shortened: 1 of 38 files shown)


packages/js/src/plugins/typescript/util.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (!projectName) return;
+
+  if (projectName) {
+    const buildDepsTargetName = options.buildDepsTargetName ?? 'build-deps';
+    targets[buildDepsTargetName] = {
+      dependsOn: ['^build'],
+    };
+    targets[options.watchDepsTargetName ?? 'watch-deps'] = {
+      dependsOn: [buildDepsTargetName],
+      command: `${pmc.exec} nx watch --projects ${projectName} --includeDependentProjects -- ${pmc.exec} nx build-deps ${projectName}`,
```
Coly010, line 38:
```suggestion
      command: `${pmc.exec} nx watch --projects ${projectName} --includeDependentProjects -- ${pmc.exec} nx ${buildDepsTargetName} ${projectName}`,
```
This should use the inferred task name right?

Coly010 (changes requested):

Coly010 (approved):

PR #29613, opened 2025-01-14
feat(core): Add ESM support for Eslint config file

ndcunningham:
This pull request includes changes to migrate ESLint configuration files from CommonJS (`.cjs`) to ECMAScript modules (`.mjs`) as the default.

### ESLint Configuration Generation Changes

The changes also ensure consistent generated eslint configs based on the base eslint config.
- If the workspace root has an `eslint.config.cjs` or `eslint.config.js` with `module.exports`. When you create a library or application it will generate an accompanying config at path `{projectRoot}/eslint.config.cjs` of the same format.
- If the workspace root has an `eslint.config.mjs` or `eslint.config.mjs` with `export default`. When you create a library or application it will generate an accompanying config at path `{projectRoot}/eslint.config.mjs`.
- If no eslint config is found at the workspace root one will be created `eslint.config.mjs`

### DIFF
--- a/packages/eslint/src/generators/lint-project/lint-project.ts
+++ b/packages/eslint/src/generators/lint-project/lint-project.ts
@@ -17,8 +17,11 @@ import {
 } from '@nx/devkit';
 
 import { Linter as LinterEnum, LinterType } from '../utils/linter';
-import { findEslintFile } from '../utils/eslint-file';
-import { join } from 'path';
+import {
+  determineEslintConfigFormat,
+  findEslintFile,
+} from '../utils/eslint-file';
+import { extname, join } from 'path';
 import { lintInitGenerator } from '../init/init';
 import type { Linter } from 'eslint';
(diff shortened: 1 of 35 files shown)


Coly010 (approved):

packages/eslint/src/generators/lint-project/lint-project.ts
```diff
@@ -199,6 +204,15 @@ function createEsLintConfiguration(
   const pathToRootConfig = extendedRootConfig
     ? `${offsetFromRoot(projectConfig.root)}${extendedRootConfig}`
     : undefined;
+
+  if (extendedRootConfig) {
+    // We do not want to mix the formats
+    options.eslintConfigFormat = tree
+      .read(extendedRootConfig, 'utf-8')
+      .includes('export default')
```
jaysoo, line 212: The file may not be formatted correctly, especially now that we're making Prettier optional. We should parse out the AST and check that default export exists.

jaysoo (approved):

PR #29617, opened 2025-01-14
fix(core): update package script logic to handle cli tool name as command

Coly010:
## Current Behavior
When we have an inferred target command that matches the entry point of the cli tool, there is a chance we do not replace package scripts correctly

e.g.

```
{
  "dev": "vite",
  "build": "tsc -b && vite build",
  "preview": "vite preview"
}
```

this could result in package scripts being updated to

```
{
  "dev": "nx dev",
  "build": "tsc -b && nx vite:build",
  "preview": "nx dev preview"
}
```

## Expected Behavior
We should update the package scripts correctly to match the desired inferred target

```
{
  "preview": "nx preview"
}
```

### DIFF
--- a/packages/devkit/src/utils/add-plugin.ts
+++ b/packages/devkit/src/utils/add-plugin.ts
@@ -232,6 +232,7 @@ function processProject(
   if (!tree.exists(packageJsonPath)) {
     return;
   }
+
   const packageJson = readJson<PackageJson>(tree, packageJsonPath);
   if (!packageJson.scripts || !Object.keys(packageJson.scripts).length) {
     return;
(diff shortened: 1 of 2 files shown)


leosvelperez (approved):

packages/devkit/src/utils/add-plugin.ts
```diff
@@ -232,17 +232,20 @@ function processProject(
   if (!tree.exists(packageJsonPath)) {
     return;
   }
-  const packageJson = readJson<PackageJson>(tree, packageJsonPath);
+  let packageJson = readJson<PackageJson>(tree, packageJsonPath);
```
JamesHenry, line 235: Nit: Is this change left behind from some older iteration of this code? It doesn't seem to serve a purpose?

Coly010: It is

JamesHenry (approved):

PR #29630, opened 2025-01-15
docs(core): add local cache information for powerpack

Cammisuli:
## Expected Behavior

Also includes update ENV variables

## Related Issue(s)

Fixes #


FrozenPandaz (approved):

docs/shared/packages/powerpack-azure-cache/powerpack-azure-cache-plugin.md
```diff
@@ -81,3 +81,29 @@ Finally, you need to configure your Nx cache in the `nx.json` file. The `contain
 | --------------- | -------------------------------- |
 | **container**   | The name of the container to use |
 | **accountName** | The name of blob storage account |
+
+# Cache Modes
+
+By default, the remote cache will try to write and read from the remote cache while running locally. This means that permissions must be set for users who are expected to access the remote cache.
```
isaacplmann, line 87:
```suggestion
By default, Nx will try to write and read from the remote cache while running locally. This means that permissions must be set for users who are expected to access the remote cache.
```

isaacplmann: Not sure what the subject of this sentence should be, but I don't the remote cache is reading from itself.  This comment applies to all the cache docs.

isaacplmann: Maybe the subject should be "the Powerpack Azure Cache Plugin"

isaacplmann (approved):
Approving so you can merge once things are updated

PR #29638, opened 2025-01-15
fix(misc): update e2e config generators to align with new TS solution setup

jaysoo:
This PR updates the `@nx/detox:app` generator to match the new TS solution setup. The `@nx/cypress:configuration` and `@nx/cypress:configuration` generators are also updated so that they can be run on existing projects and generator the correct tsconfig files.

The Playwright/Cypress example can be seen as follows:

```shell
# Skip e2e
nx g @nx/react:app apps/demo --bundler vite --e2eTestRunner none

# now configure e2e
nx g @nx/playwright --project demo
```

Now if you add this line to `apps/demo/e2e/example.spec.ts`:
```
const x: number = 'a';
```

And run `nx typecheck demo`, it will pass. This happens because the  `e2e/**/*.ts` pattern is missing. Thus, we need to ensure that a `tsconfig.e2e.json` project is added for the Playwright spec files. Same thing with Cypress.

The Detox generator does not support adding configuration to existing project, so we don't quite get the same problem. The fix for Detox is just to make sure the tsconfig content is not following the old (integrated) version, but the updated TS solution version.

## Current Behavior
Detox TS setup is incorrect. Running Cypress and Playwright configuration generator on existing projects generate invalid setup, such that spec files are not typechecked.

## Expected Behavior
E2E generators should all generate correct TS setup.

## Related Issue(s)

Fixes #


packages/playwright/src/generators/configuration/files/playwright.config.ts.template
```diff
@@ -1,7 +1,6 @@
 import { defineConfig, devices } from '@playwright/test';
 import { nxE2EPreset } from '@nx/playwright/preset';
-<% if(!webServerCommand || !webServerAddress) { %>// eslint-disable-next-line @typescript-eslint/no-unused-vars <% } %>
-import { workspaceRoot } from '@nx/devkit';
+<% if (webServerCommand && webServerAddress) { %>import { workspaceRoot } from '@nx/devkit';<% } %>
```
jaysoo, line 3: This import is unused unless both options are set, which leads to typecheck errors in the new setup. We should just not import it if we're not going to use it.

Coly010 (approved):

packages/detox/src/generators/application/application.ts
```diff
@@ -38,6 +43,26 @@ export async function detoxApplicationGeneratorInternal(
...
   const depsTask = ensureDependencies(host, options);
 
+  updateTsconfigFiles(
+    host,
+    options.e2eProjectRoot,
+    'tsconfig.json',
+    {
+      outDir: 'out-tsc/detox',
+      allowJs: true,
+      types: ['node', 'jest', 'detox'],
```
leosvelperez, line 53: We need to add the `module`/`moduleResolution` expected options for Detox.

jaysoo: Done

leosvelperez (approved):

xiongemi (approved):

PR #29649, opened 2025-01-16
fix(js): ignore bundler config files correctly in eslint config when using vitest in js library generator

leosvelperez:
## Current Behavior

Generating a js library with esbuild or rollup as bundlers and vitest as the unit test runner, results in an eslint configuration where the bundler config file is not ignored from the `@nx/dependency-checks` rule.

## Expected Behavior

Generating a js library with esbuild or rollup as bundlers and vitest as the unit test runner, should result in an eslint configuration where the bundler config file is ignored from the `@nx/dependency-checks` rule.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/generators/library/library.ts
+++ b/packages/js/src/generators/library/library.ts
@@ -471,25 +471,25 @@ export async function addLint(
           ruleSeverity = value;
           ruleOptions = {};
         }
-        if (options.bundler === 'vite' || options.unitTestRunner === 'vitest') {
-          ruleOptions.ignoredFiles ??= [];
-          ruleOptions.ignoredFiles.push(
-            '{projectRoot}/vite.config.{js,ts,mjs,mts}'
-          );
-          o.rules['@nx/dependency-checks'] = [ruleSeverity, ruleOptions];
+        const ignoredFiles = new Set(ruleOptions.ignoredFiles ?? []);
+        if (options.bundler === 'vite') {
+          ignoredFiles.add('{projectRoot}/vite.config.{js,ts,mjs,mts}');
         } else if (options.bundler === 'rollup') {
-          ruleOptions.ignoredFiles ??= [];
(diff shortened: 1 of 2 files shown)


Coly010 (approved):

packages/js/src/generators/library/library.ts
```diff
@@ -471,25 +471,25 @@ export async function addLint(
           ruleSeverity = value;
           ruleOptions = {};
         }
-        if (options.bundler === 'vite' || options.unitTestRunner === 'vitest') {
-          ruleOptions.ignoredFiles ??= [];
-          ruleOptions.ignoredFiles.push(
-            '{projectRoot}/vite.config.{js,ts,mjs,mts}'
-          );
-          o.rules['@nx/dependency-checks'] = [ruleSeverity, ruleOptions];
+        const ignoredFiles = new Set(ruleOptions.ignoredFiles ?? []);
```
jaysoo, line 474: Technically you could remove the `else if` from 477 and achieve the same result I think.

```js
if (options.bundler === "vite" || options.unitTestRunner === 'vitest') {
  ignoredFiles.add("{projectRoot}/vite.config.{js,ts,mjs,mts}");
}

if (options.bundler === "rollup") {
  ignoredFiles.add("{projectRoot}/rollup.config.{js,ts,mjs,mts,cjs,cts}");
} else if (options.bundler === "esbuild") {
  ignoredFiles.add("{projectRoot}/esbuild.config.{js,ts,mjs,mts}");
}

```

jaysoo (approved):

packages/js/src/generators/library/library.ts
leosvelperez: It works, but I don't think it reads as well because you already check for `bundler === vite` first, so, there's no way it can be `bundler === rollup` if that was the case. Checking for bundlers and unit test runners separately feels better to me.

PR #29652, opened 2025-01-16
docs(core): project linking concept and recipe

isaacplmann:
Adds two pages:
- Concept page describing two options for [TypeScript project linking](https://nx-dev-git-docs-workspaces-project-linking-nrwl.vercel.app/concepts/typescript-project-linking) (workspaces and path aliases).
    - Also describes performance benefits of TS Project References that are possible with workspaces
- Recipe describing how to manually [switch from path aliases to workspaces project linking](https://nx-dev-git-docs-workspaces-project-linking-nrwl.vercel.app/recipes/tips-n-tricks/switch-to-workspaces-project-references)


docs/shared/concepts/typescript-project-linking.md
```diff
@@ -0,0 +1,277 @@
...
+
+To create a new Nx workspace that links projects with package manager workspaces, use the `--workspaces` flag.
+
+```shell
+npx create-nx-workspace --workspaces
+```
+
+### Set Up Package Manager Workspaces
+
+The configuration for package manager workspaces varies based on which package manager you're using.
```
juristr, line 33: I know you mention the package.json linking in the various tabs since PNPM has a different protocol. Can we bring it up more clearly though that for the "workspaces" based pacakge linking you need to have the dependency to the packages defined in the project's `package.json`?

like if `A -> B` then `A`'s `package.json` needs to have a dependency to `B`. 

I know in theory NPM (and maybe yarn and bun) don't strictly require this, while PNPM (and yarn pnp?) do. But I'd almost lean towards defining it so things don't go wrong.

docs/shared/concepts/typescript-project-linking.md
```diff
@@ -0,0 +1,277 @@
...
+With workspaces enabled, you can also configure TypeScript project references to speed up your build and typecheck tasks.
+
+The root `tsconfig.base.json` should contain a `compilerOptions` property and no other properties. `compilerOptions.composite` and `compilerOptions.declaration` should be set to `true`. `compilerOptions.paths` should not be set.
+
+```jsonc {% fileName="tsconfig.base.json" %}
+{
+  "compilerOptions": {
+    // Required compiler options
+    "composite": true,
+    "declaration": true
```
juristr, line 109: I'd also add `declarationMaps: true`.

docs/shared/concepts/typescript-project-linking.md
```diff
@@ -0,0 +1,277 @@
...
+      ]
+    }
+  }
+}
+```
+
+## Project Linking with TypeScript Path Aliases
+
+{% callout type="warning" title="Incompatible with TypeScript Project References" %}
+You can not use TypeScript project references with this style of project linking.
```
juristr, line 255: this is technically not correct.

You can use TS path aliases with project references. The former is for the name resolution while the latter defines how TS creates "programs" for type checking/compilation.

Now, while possible, it's not what we recommend for sure. We should somewhere mention though that the recommendation goes:

- Workspaces + proj references (preferred for new workspaces)
- TS path aliases (how Nx does it now and will still be valid for a long time)

juristr (changes requested):

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -0,0 +1,230 @@
...
+packages:
+  # specify a package in a direct subdir of the root
+  - 'my-app'
+  # all packages in direct subdirs of packages/
+  - 'packages/*'
+  # all packages in subdirs of libs/
+  - 'libs/**'
+  # exclude packages that are inside test directories
+  - '!**/test/**'
+```
```
juristr, line 55: should this be aligned with the ones for npm, bun etc..?

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -0,0 +1,230 @@
...
+
+{% /tab %}
+{% /tabs %}
+
+## Update Root TypeScript Configuration
+
+The root `tsconfig.base.json` should contain a `compilerOptions` property and no other properties. `compilerOptions.composite` and `compilerOptions.declaration` should be set to `true`. `compilerOptions.paths` should not be set.
+
+Note: Before you delete the `paths` property, copy the project paths for use in the `tsconfig.json` file.
```
juristr, line 76: 🤔 can we copy them as is to the individual projects? Does that works? Also do we intend 

a) a complete switch with this recipe in which case the path aliases are not required any more at all, 
b) a path for an incremental migration, in which case I'm wondering whether it'd be possible to keep some packages still resolved over `tsconfig.base.json` path aliases 🤔

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -0,0 +1,230 @@
...
+  }
+}
+```
+
+{% /tab %}
+{% /tabs %}
+
+## Update Root TypeScript Configuration
+
+The root `tsconfig.base.json` should contain a `compilerOptions` property and no other properties. `compilerOptions.composite` and `compilerOptions.declaration` should be set to `true`. `compilerOptions.paths` should not be set.
```
juristr, line 74: I believe the `rootDir` also needs to be removed as that now resides within each project-level `tsconfig.json` when switching to TS proj refs.

juristr (changes requested):

docs/shared/concepts/typescript-project-linking.md
```diff
@@ -0,0 +1,308 @@
...
+With workspaces enabled, you can also configure TypeScript project references to speed up your build and typecheck tasks.
+
+The root `tsconfig.base.json` should contain a `compilerOptions` property and no other properties. `compilerOptions.composite` and `compilerOptions.declaration` should be set to `true`. `compilerOptions.paths` should not be set.
+
+```jsonc {% fileName="tsconfig.base.json" %}
+{
+  "compilerOptions": {
+    // Required compiler options
+    "composite": true,
+    "declaration": true,
```
leosvelperez, line 139: Setting `composite: true` enables `declaration: true` by default. It's not wrong to set it and it's required to have `declaration` enabled. So, the wrong setup is if the user already had `declaration: false`, in which case, they need to set it to `true` or remove the option so the default applies.

FWIW, we generate the `tsconfig.base.json` without setting `declaration`.

docs/shared/concepts/typescript-project-linking.md
```diff
@@ -0,0 +1,308 @@
...
+Typecheck (1 nested leaf pkg updated): 26.66 seconds, max memory 407.54 MB
+Typecheck (2 nested leaf pkg updated): 31.17 seconds, max memory 889.86 MB
+Typecheck (1 nested root pkg updated): 26.67 seconds, max memory 393.78 MB
+```
+
+These performance benefits will be more noticeable for larger repositories, but even small code bases will see some benefits.
+
+### Local TypeScript Path Aliases
+
+When using project references, you can not define path aliases in the root `tsconfig.base.json` file because TypeScript does not merge the path aliases when doing the typecheck calculation, but you can set path aliases in an individual project's `tsconfig.app.json` file. For instance, you could define paths like this in an application's tsconfig file.
```
leosvelperez, line 266: This reads as if this is a consequence of using TS project references, but this is the case with or without using TS project references.

Also, I think it should be clearer that it's not that you "can not" define, but rather that they would be overridden by the TS path aliases defined in tsconfig files that extend from the `tsconfig.base.json`, so they would be useless. I know that's what you're saying and maybe it's me, but I think the message is better conveyed with something along these lines:

```suggestion
If you define TS path aliases in individual project's tsconfig files, you should not define them also in the root `tsconfig.base.json` file because TypeScript does not merge them and the ones defined in the extended file would be completely overwritten by the ones defined in the tsconfig extending from the base config. For instance, you could define paths like this in an application's tsconfig file.
```

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -0,0 +1,230 @@
+# Switch to Workspaces and Project References
+
+In order to take advantage of the [performance benefits](/concepts/typescript-project-linking#typescript-project-references-performance-benefits) of TypeScript project references, you need to use package manager workspaces for local [project linking](/concepts/typescript-project-linking). If you are currently using TypeScript path aliases for project linking, follow the steps in this guide to switch to workspaces project linking and enable TypeScript project references.
```
leosvelperez, line 3: As mentioned in a previous comment, package manager workspaces are not a requirement for using TS project references. However, they are recommended regardless of whether you use TS project references. They enable native Node.js module resolution rather than having to resolve through a custom TypeScript resolution for TS path aliases.

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
leosvelperez: Yeah, we should probably advocate for an incremental migration.

In any case, this note is unclear to me, or at least it feels like it's missing a "follow-up". The note mentions copying the project paths, but we don't say or show an explicit snippet/example of what to do with them.

Ideally, TS path aliases should be converted to entry points defined in the `package.json` of each project:

- if you have a TS path alias matching the package name, it would be the `exports['.']` or `main`/`module` field defined in the `package.json`
- if you have a TS path alias that starts with `<package name>/` (a secondary entry point), it would be a corresponding `exports['<relative path after package name>'] (e.g. for `@foo/bar/baz` it would be `exports[./baz]`)
- if you have a TS path aliases that doesn't match any of the above, you can either:
  - keep the TS path alias (not recommended)
  - define an entry point for it and update the code currently use the TS path alias key and replace it with the new entry point

juristr (approved):

PR #29658, opened 2025-01-17
feat(release): allow invalid conventional commits to be considered via custom config

tandrup:
This allows users with repos that are a mix of conventional commits and other commits to configure that non-conventional commits will result in for example a patch version bump.

## Current Behavior
Currently any commit that does match the conventional commit regex is dropped, and does not cause a version bump.

## Expected Behavior
Due the content of our monorepo we have a mix of conventional commits and other commits. With this change we will be able to configure that a non-conventional commit should cause for example a patch semver version bump.
The default behaviour remains the same and will not bump versions for non-conventional commits.

## Related Issue(s)
Haven't been able to find any.

### DIFF
--- a/packages/nx/src/command-line/release/config/conventional-commits.ts
+++ b/packages/nx/src/command-line/release/config/conventional-commits.ts
@@ -94,5 +94,12 @@ export const DEFAULT_CONVENTIONAL_COMMITS_CONFIG: NxReleaseConfig['conventionalC
           hidden: true,
         },
       },
+      __INVALID__: {
+        semverBump: 'none',
+        changelog: {
+          title: 'Invalid based on conventional commits specification',
+          hidden: true,
+        },
+      },
     },
   };
--- a/packages/nx/src/command-line/release/utils/git.ts
+++ b/packages/nx/src/command-line/release/utils/git.ts
@@ -412,7 +412,12 @@ export function parseConventionalCommitsMessage(message: string): {
 } | null {
   const match = message.match(ConventionalCommitRegex);
   if (!match) {
-    return null;
+    return {
+      type: '__INVALID__',
+      scope: '',
+      description: message,
+      breaking: false,
+    };
   }
 
   return {
--- a/e2e/release/src/conventional-commits-config.test.ts
+++ b/e2e/release/src/conventional-commits-config.test.ts
@@ -48,7 +48,7 @@ describe('nx release conventional commits config', () => {
   let pkg5: string;
   let pkg6: string;
 
-  beforeAll(async () => {
+  beforeEach(async () => {
     newProject({
       packages: ['@nx/js'],
     });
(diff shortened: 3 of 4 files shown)


mta-trackunit:
@JamesHenry please let me know if there is anything I can do to assist in releasing this non breaking change.

packages/nx/src/command-line/release/config/conventional-commits.ts
```diff
@@ -94,5 +94,12 @@ export const DEFAULT_CONVENTIONAL_COMMITS_CONFIG: NxReleaseConfig['conventionalC
           hidden: true,
         },
       },
+      unknown: {
+        semverBump: 'none',
+        changelog: {
+          title: '🤷 Unknown',
```
JamesHenry, line 100:
```suggestion
          title: 'Invalid based on conventional commits specification',
```

packages/nx/src/command-line/release/utils/git.ts
```diff
@@ -412,7 +412,12 @@ export function parseConventionalCommitsMessage(message: string): {
 } | null {
   const match = message.match(ConventionalCommitRegex);
   if (!match) {
-    return null;
+    return {
+      type: 'unknown',
```
JamesHenry, line 416:
```suggestion
      type: '__INVALID__',
```

JamesHenry (changes requested):
I'm not a huge fan of this, because the whole point of adopting the conventional commits standard is to have consistent constraints.

If you are not able to enforce this consistently, you might consider using version plans instead, where the source of truth for bumps is moved outside of your commits and into dedicated files on disk instead:

https://nx.dev/recipes/nx-release/file-based-versioning-version-plans

Having said that, I am ok to support this as long as we are explicit about the fact that this commit is not valid when the project has been set up to use conventional commits, and I also want to use a value that is extremely likely to not have been used by users before as a scope (`__INVALID__`) so I have added some suggested changes for the values to use here (naturally then for the title you can use whatever you actually want in your nx.json)

JamesHenry:
Please also add a test in `e2e/release/src/conventional-commits-config.test.ts` which uses an invalid commit structure (1) with the default config (2) with custom config like you would want in your repo

tandrup:
> Please also add a test in `e2e/release/src/conventional-commits-config.test.ts` which uses an invalid commit structure (1) with the default config (to prove no breaking change/regression) (2) with custom config like you would want in your repo

Good idea. I've added the test now.

tandrup:
@JamesHenry Is there any other changes needed?

JamesHenry (approved):

JamesHenry:
Thanks again @tandrup!

nkalinov:
I must say thank you for this as I needed the same and was doing a rather hacky workaround in my custom release executor.

PR #29691, opened 2025-01-20
feat(webpack, rspack): support multiple configurations

ndcunningham:
This pull request includes changes to support multi-configuration mode for both Rspack and Webpack. 

## Currently
Currently our plugin only supports single configurations
```js
module.exports =  { 
  ...config
}
```
Which works in most cases but some applications can have mutliple configs that serve different platforms.

## Changes
With these changes, the Webpack and Rspack plugins will also support multi-configuration.
```js
module.exports = [ 
   { ...clientConfig },
   { ...serverConfig }
 ]

### DIFF
--- a/packages/rspack/src/utils/read-rspack-options.ts
+++ b/packages/rspack/src/utils/read-rspack-options.ts
@@ -5,53 +5,88 @@ import { readNxJsonFromDisk } from 'nx/src/devkit-internals';
 
 /**
  * Reads the Rspack options from a give Rspack configuration. The configuration can be:
- * 1. A standard config object
- * 2. A standard function that returns a config object
- * 3. A Nx-specific composable function that takes Nx context, rspack config, and returns the config object.
+ * 1. A single standard config object
+ * 2. A standard function that returns a config object (standard Rspack)
+ * 3. An array of standard config objects (multi-configuration mode)
+ * 4. A Nx-specific composable function that takes Nx context, rspack config, and returns the config object.
  *
  * @param rspackConfig
  */
 export async function readRspackOptions(
   rspackConfig: unknown
-): Promise<Configuration> {
-  let config: Configuration;
-  if (isNxRspackComposablePlugin(rspackConfig)) {
-    config = await rspackConfig(
-      {},
-      {
-        // These values are only used during build-time, so passing stubs here just to read out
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/rspack/src/utils/read-rspack-options.ts [hunk 1]
```diff
@@ -5,53 +5,66 @@ import { readNxJsonFromDisk } from 'nx/src/devkit-internals';
...
+            root: workspaceRoot,
+            cwd: undefined,
+            isVerbose: false,
+            nxJsonConfiguration: readNxJsonFromDisk(workspaceRoot),
+            projectGraph: null,
+            projectsConfigurations: null,
+          },
+        }
+      );
+    } else if (typeof rspackConfig === 'function') {
```
jaysoo, line 49: Is it valid to export a function that returns multiple config objects? Or can you export an array that contains functions? Or both?

It's not clear from the webpack/rspack docs which is a valid mix of array and function, can you double-check this case?

ndcunningham: Each element in the array can be either a function or a config object.
Each function returns a config object.

jaysoo: So webpack/rspack do not support a function that returns an array?

ndcunningham: It does, I have updated the PR to handle these cases as well.

Coly010: > So webpack/rspack do not support a function that returns an array?
You can: https://github.com/Coly010/ng-rspack-build/blob/main/packages/build/src/lib/config/create-config.ts#L12
Usage: https://github.com/Coly010/ng-rspack-build/blob/main/apps/rspack/ssr/css/rspack.config.js#L3

jaysoo (approved):

jaysoo:
LGTM, we should add a couple of e2e test cases to cover multi-builds.

packages/rspack/src/utils/read-rspack-options.ts
ndcunningham: This is supported with the PR.

The error you will get is if your `module.exports` is a function that returns an array which has 1 or more functions.

PR #29698, opened 2025-01-21
Typescript project references article

ZackDeRose:
https://nx-dev-git-fork-zackderose-typescript-project-refer-8c9adb-nrwl.vercel.app/blog/typescript-project-references


docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
+---
+title: What Are TypeScript Project References?
+slug: typescript-project-references
+authors: [Zack DeRose]
+tags: [typescript, monorepo, nx]
+cover_image: /blog/images/2025-01-21/ts-islands.png
+---
+
+## Connecting Projects At The TypeScript Level
```
juristr, line 9: Having a title right in the beginning always feels off to me as there's already the main article title immediately followed by another one :) I think we can just drop it here and this first paragraphs serve as the intro

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+is-odd
+  index.ts
+  tsconfig.json
+tsconfig.base.json
+```
+
+If we try to run our build script on the `is-odd` project, we see that TypeScript is unable to run the `tsc` command because at the TypeScript level, `is-odd` is not aware of the `is-even` module:
+
+```shell
+ . > cd is-odd
```
juristr, line 26: Instead of "cd" into the directory which I think is not relevant here, you should be able to pass a "path" property:
https://github.com/nrwl/nx/tree/master/docs#terminal-output

Like:

‎``` {% command="tsc" path="~/is-odd" %}
index.ts:1:24 - error TS2792: Cannot find module 'is-even'. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?

1 import { isEven } from 'is-even';
                         ~~~~~~~~~

Found 1 error in index.ts:1
‎```

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+      "is-even": ["./is-even/index.ts"],
+      "is-odd": ["./is-odd/index.ts"]
+    }
+  }
+}
+```
+
+By having the individual `tsconfig.json` files `extend` this base config, they will all get these paths, and now our build command will work:
+
+```shell
```
juristr, line 60: I think we can leave this one away

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
+---
+title: What Are TypeScript Project References?
```
mhartington, line 2: I wonder if a title like "A Beginner's Guide to TypeScript Project References" or "Everything You Need to Know About TypeScript Project References".

Something that has a bit more SEO juice

docs/blog/2025-01-21-project-references.md
mhartington, line 9: I think we can cut this paragraph title, feels redundant

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+
+```ts
+import { isEven } from 'is-even';
+```
+
+There are alternatives to path aliases to allow for this name to be resolved. The most recent enhancements in Nx use the `workspaces` functionality of your package manager of choice (npm/pnpm/yarn/bun) as the way of resolving these names.
+
+With this set up, we can now use the `-b` or `--build` option when building `is-odd` - let's run it now with the `--verbose` flag on:
+
+```shell
```
juristr, line 102: use the terminal-output fence for this one instead of shell.

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+
+## So What Does This Mean For You
+
+As a practical/pragmatic developer - the TLDR of all of this information is project references allow for more performant builds.
+
+We've put together [a repo to demonstrate the perfomance gains](https://github.com/jaysoo/typecheck-timings), summarized by this graphic:
+
+![results](/blog/images/2025-01-21/results.png)
+
+In addition to the time savings we saw reduced memory usage (~< 1GB vs 3 GB). This makes sense given what we saw about how project references work. This is actually a very good thing for CI pipelines, as exceeding memory usuage is a common issue we see with our clients for their TypeScript builds. Less memory usuage means we can use smaller machines, which saves on the CI costs.
```
juristr, line 145:
```suggestion
In addition to the time savings we saw reduced memory usage (~< 1GB vs 3 GB). This makes sense given what we saw about how project references work. This is actually a very good thing for CI pipelines, as exceeding memory usage is a common issue we see with our clients for their TypeScript builds. Less memory usage means we can use smaller machines, which saves on the CI costs.
```

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+tsconfig.base.json
+```
+
+Notice how both `is-even` AND `is-odd` now have a compiled `index.d.ts` declaration file and `index.js`. They also both have a `tsconfig.base.json` file now (this holds the additional data TypeScript needs to determine which builds are needed). With the `--build` option, TypeScript is now operating as a build orchestrator - by finding all referenced projects, determining if they are out-of-date, and then building them in the correct order.
+
+## So What Does This Mean For You
+
+As a practical/pragmatic developer - the TLDR of all of this information is project references allow for more performant builds.
+
+We've put together [a repo to demonstrate the perfomance gains](https://github.com/jaysoo/typecheck-timings), summarized by this graphic:
```
mhartington, line 141: Let's update the repo link to https://github.com/nrwl/typecheck-timings

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+
+? Would you like to sync the identified changes to get your workspace up to date? …
+Yes, sync the changes and run the tasks
+No, run the tasks without syncing the changes
+
+You can skip this prompt by setting the `sync.applyChanges` option to `true` in your `nx.json`.
+For more information, refer to the docs: https://nx.dev/concepts/sync-generators.
+```
+
+So, offload the mental load to your tools, focus on building your solution, and you should have the best of all worlds.
```
mhartington, line 177: Let's throw our social links down at the bottom

docs/blog/images/2025-01-21/results.png
juristr, line 1: this one has really low contrast. is there any way to improve it?

mhartington (changes requested):
Just a few comments

mhartington (dismissed):
Just a few comments

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+
+```
+
+As we can see, there's an error associated with this line:
+
+```typescript
+import { isEven } from 'is-even';
+```
+
+TypeScript needs to be informed as to how to find the module named `is-even`. The error message here actually suggests that we may have forgotten to add aliases to the 'paths' option. To do this we can adjust our `tsconfig.json` file at the root of the monorepo:
```
juristr, line 45:
```suggestion
TypeScript needs to be informed as to how to find the module named `is-even`. The error message here actually suggests that we may have forgotten to add aliases to the `paths` option. To do this we can adjust our `tsconfig.json` file at the root of the monorepo:
```

juristr (changes requested):

juristr (changes requested):

docs/blog/images/2025-01-21/results.png
mhartington: There is this version, with dark mode off 

Screenshot 2025-01-21 at 10 50 12 AM

juristr: yeah maybe that's better 🤔. For me it is mostly the text on the side is really small plus also contrast wise not great. So it's hard to read.

docs/blog/2025-01-21-project-references.md
juristr, line 141:
```suggestion
We've put together [a repo to demonstrate the performance gains](https://github.com/jaysoo/typecheck-timings), summarized by this graphic:
```

docs/blog/2025-01-21-project-references.md
jaysoo: I agree with Juri that the title underneath the graphic feels off. Maybe one intro sentence makes more sense. Like: `Project References is a feature to help scale your TypeScript monorepo. In this post you will learn all about how it works and how it helps with the performance of your workspaces.

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+
+Found 1 error in index.ts:1
+
+```
+
+As we can see, there's an error associated with this line:
+
+```typescript
+import { isEven } from 'is-even';
+```
```
leosvelperez, line 43: I feel like this should be shown as part of the setup before the paragraph:

> If we try to run our build script on the `is-odd` project...

So, we give all the relevant context before building and seeing the error.

docs/blog/2025-01-21-project-references.md
```diff
@@ -0,0 +1,177 @@
...
+
+By adding boundaries at the TypeScript level, we can significantly cut down on the "surface area" that TypeScript has to contend with when doing its job. This way, rather than TypeScript seeing our entire monorepo as one unit, it can now understand our workspace as a series of connected "islands" or nodes.
+
+![Islands of TypeScript](/blog/images/2025-01-21/ts-islands.png)
+
+To add this to our previous example, we'll adjust the `tsconfig.json` file for `is-odd` since it depends on `is-even`:
+
+```json
+{
+  "extends": "../tsconfig.base.json",
```
leosvelperez, line 80: We don't show anywhere that we need to set `composite: true`.

juristr:
> Will there be a migration for this?

We're looking into that. For now it's behind a flag `--workspaces` as we want more feedback. But yes, there will be guides and potentially some automated migration.

There's a series of articles coming out, this one specifically just talks about project references 

>  Also I saw somewhere that it won't work for Angular for now? Is that the case? If so maybe also add note about it here.

It doesn't work for Angular. That's not on the Nx side, but on the Angular side. We're looking into it though.

juristr (approved):

leosvelperez (approved):

PR #29705, opened 2025-01-21
feat(js): remove nx property from generated package.json files

jaysoo:
This PR updates our generators to no longer generate with `nx` in `package.json` by default. The only times it is needed is if you pass add `tags` or `implicitDependencies` to the project config.

This PR replaces our `projectType` checks to use the `getProjectType` util from `@nx/js` to prefer the project config, but otherwise will check for our conventions (e.g. using `exports` for libs, `tsconfig.lib.json` vs `tsconfig.app.json`).

## Impact
- There shouldn't be any behavioral changes to existing projects that have explicit `projectType`, `name`, etc. in with `project.json` or `package.json` (via `nx` property).
- For new projects created under the new TS setup, the `nx` property will no longer be there. Generators with logic that depend on `projectType` will now check for `tsconfig.lib.json` and `tsconfig.app.json` (so all of our generators are covered). If none of those tsconfig files are found, then we check `package.json`, since libraries are required to have `exports` to be consumed.

### DIFF
--- a/packages/vue/src/generators/library/library.ts
+++ b/packages/vue/src/generators/library/library.ts
@@ -47,6 +47,12 @@ export async function libraryGeneratorInternal(tree: Tree, schema: Schema) {
     );
   }
 
+  // If we are using the new TS solution
+  // We need to update the workspace file (package.json or pnpm-workspaces.yaml) to include the new project
+  if (options.isUsingTsSolutionConfig) {
+    addProjectToTsSolutionWorkspace(tree, options.projectRoot);
+  }
+
   if (options.isUsingTsSolutionConfig) {
     writeJson(tree, joinPathFragments(options.projectRoot, 'package.json'), {
       name: getImportPath(tree, options.name),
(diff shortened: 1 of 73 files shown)


packages/vue/src/generators/library/library.ts
```diff
@@ -47,19 +47,27 @@ export async function libraryGeneratorInternal(tree: Tree, schema: Schema) {
...
-        sourceRoot: `${options.projectRoot}/src`,
-        tags: options.parsedTags?.length ? options.parsedTags : undefined,
-      },
+      nx: options.parsedTags?.length
+        ? {
+            name: options.name,
+            projectType: 'library',
+            sourceRoot: `${options.projectRoot}/src`,
+            tags: options.parsedTags,
+          }
```
leosvelperez, line 69:
```suggestion
        ? {
            tags: options.parsedTags,
          }
```

leosvelperez (approved):

PR #29713, opened 2025-01-22
fix(nuxt): correct webServerAddress and webServerCommand options in add-e2e.ts generator. 

makp0:
## Current Behavior
Currently, the add-e2e.ts generator has redundant lines for the webServerAddress and webServerCommand configuration when generating Playwright configuration. 
Produces `TypeError: Invalid URL` when running playwright tests on a clean create-nx-workspace project. 

## Expected Behavior
The generator should configure webServerAddress and webServerCommand correctly. The changes fix the configuration generation for Playwright tests.


Coly010 (approved):
Great catch, thank you

PR #29726, opened 2025-01-23
fix(core): don't set legacy-peer-deps by default

### LINKED ISSUE #22066: `nx release` updating lock file un-syncs lock file with package.json
### Current Behavior

When executing the `nx release` command the package-lock.json file is updated (which is expected). But it also removes peer-dependencies which are actually in use. Therefore subsequent `npm ci` calls fail, because the package.json and the package-lock.json file are not in sync anymore (executing `npm i` adds the dependencies back to the lock file).

In my case this refers to the following transitive (peer-)dependencies of `@nx/js`:
- `@types/node`
- `typescript`
- `undici-types`

### Expected Behavior

The `nx release` command updates the package-lock file, but only updates the versions of the newly release packages and does not remove any **used** (peer-)dependencies

### GitHub Repo

https://github.com/m-thalmann/codestyle-packages/

### Steps to Reproduce

1. Create a Nx Package-Based project (pure JS, no Typescript)
2. Create a package and commit the changes
3. Setup Nx Release
   - I used the following config:
      ```json
      {
        "release": {
          "projects": ["*"],
          "version": {
            "conventionalCommits": true
          },
          "git": {
            "commitMessage": "build(release): create new versions of packages"
          },
          "changelog": {
            "projectChangelogs": {
              "renderOptions": {
                "authors": false
               }
            }
          }
        }
      }
      ```
4. Run `nx release --skip-publish --first-release` (or omit the skip-publish, does not matter) (its also not dependent on the first release flag)
5. Inspect the create commit: The package-lock should be updated with the above mentioned dependencies removed
6. Run `npm ci` -> this will fail

### Nx Report

```shell
Node   : 20.11.0
OS     : win32-x64
npm    : 10.2.4

nx (global)    : 18.0.3
nx             : 18.0.5
@nx/js         : 18.0.5
@nx/workspace  : 18.0.5
@nx/devkit     : 18.0.5
@nrwl/tao      : 18.0.5
typescript     : 5.3.3
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [X] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

This issue occurs on both our local windows developer machines and the linux build servers.

I tried to find the root cause but I was not able to reproduce this issue when running the `npm install --package-lock-only` command (which is used by the update lock file function in our case).
I've also tried executing it from a custom JS script but this did also not reproduce the above mentioned issue.

Maybe it is something in the [packages/js/src/generators/release-version/utils/update-lock-file.ts](https://github.com/nrwl/nx/blob/master/packages/js/src/generators/release-version/utils/update-lock-file.ts) file but I have no idea how to start.

### LINKED ISSUE #29537: [Nx Release] Out of sync package-lock.json after `nx release version` execution
### Current Behavior

Running `nx release version` leave us with an out of sync package-lock.json file, so running `npm ci` is not possible

### Expected Behavior

`nx release version` should update files without issues and `npm ci` should works...

### GitHub Repo

https://github.com/gperdomor/cvb

### Steps to Reproduce

1. Clone this repo https://github.com/gperdomor/cvb and install reps
2. Execute `nx release version` and pick an option, like minor, pre-release, etc
3. Execute `npm ci`

### Nx Report

```shell

Node           : 22.12.0
OS             : darwin-x64
Native Target  : x86_64-macos
npm            : 11.0.0

nx (global)        : 20.3.0
nx                 : 20.3.0
@nx/js             : 20.3.0
@nx/eslint         : 20.3.0
@nx/workspace      : 20.3.0
@nx/devkit         : 20.3.0
@nx/eslint-plugin  : 20.3.0
@nx/vite           : 20.3.0
@nx/web            : 20.3.0
typescript         : 5.6.3
Registered Plugins:
@nx/js/typescript
@nx/vite/plugin
@nx/eslint/plugin
```

### Failure Logs

```shell
npm ci            
npm error code EUSAGE
npm error
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm error
npm error Missing: tailwindcss@3.4.17 from lock file
npm error Missing: @alloc/quick-lru@5.2.0 from lock file
npm error Missing: arg@5.0.2 from lock file
npm error Missing: chokidar@3.6.0 from lock file
npm error Missing: didyoumean@1.2.2 from lock file
npm error Missing: dlv@1.1.3 from lock file
npm error Missing: jiti@1.21.7 from lock file
npm error Missing: lilconfig@3.1.3 from lock file
npm error Missing: normalize-path@3.0.0 from lock file
npm error Missing: object-hash@3.0.0 from lock file
npm error Missing: postcss-import@15.1.0 from lock file
npm error Missing: postcss-js@4.0.1 from lock file
npm error Missing: postcss-load-config@4.0.2 from lock file
npm error Missing: postcss-nested@6.2.0 from lock file
npm error Missing: postcss-selector-parser@6.1.2 from lock file
npm error Missing: sucrase@3.35.0 from lock file
npm error Missing: anymatch@3.1.3 from lock file
npm error Missing: glob-parent@5.1.2 from lock file
npm error Missing: is-binary-path@2.1.0 from lock file
npm error Missing: readdirp@3.6.0 from lock file
npm error Missing: picomatch@2.3.1 from lock file
npm error Missing: binary-extensions@2.3.0 from lock file
npm error Missing: postcss-value-parser@4.2.0 from lock file
npm error Missing: read-cache@1.0.0 from lock file
npm error Missing: camelcase-css@2.0.1 from lock file
npm error Missing: yaml@2.7.0 from lock file
npm error Missing: cssesc@3.0.0 from lock file
npm error Missing: pify@2.3.0 from lock file
npm error Missing: picomatch@2.3.1 from lock file
npm error Missing: commander@4.1.1 from lock file
npm error Missing: glob@10.4.5 from lock file
npm error Missing: lines-and-columns@1.2.4 from lock file
npm error Missing: mz@2.7.0 from lock file
npm error Missing: ts-interface-checker@0.1.13 from lock file
npm error Missing: any-promise@1.3.0 from lock file
npm error Missing: thenify-all@1.6.0 from lock file
npm error Missing: thenify@3.3.1 from lock file
npm error Missing: foreground-child@3.3.0 from lock file
npm error Missing: jackspeak@3.4.3 from lock file
npm error Missing: minimatch@9.0.5 from lock file
npm error Missing: minipass@7.1.2 from lock file
npm error Missing: package-json-from-dist@1.0.1 from lock file
npm error Missing: path-scurry@1.11.1 from lock file
npm error Missing: signal-exit@4.1.0 from lock file
npm error Missing: @isaacs/cliui@8.0.2 from lock file
npm error Missing: @pkgjs/parseargs@0.11.0 from lock file
npm error Missing: string-width@5.1.2 from lock file
npm error Missing: string-width-cjs@4.2.3 from lock file
npm error Missing: strip-ansi@7.1.0 from lock file
npm error Missing: strip-ansi-cjs@6.0.1 from lock file
npm error Missing: wrap-ansi@8.1.0 from lock file
npm error Missing: wrap-ansi-cjs@7.0.0 from lock file
npm error Missing: lru-cache@10.4.3 from lock file
npm error Missing: eastasianwidth@0.2.0 from lock file
npm error Missing: emoji-regex@9.2.2 from lock file
npm error Missing: ansi-regex@6.1.0 from lock file
npm error Missing: ansi-styles@6.2.1 from lock file
npm error
npm error Clean install a project
npm error
npm error Usage:
npm error npm ci
npm error
npm error Options:
npm error [--install-strategy <hoisted|nested|shallow|linked>] [--legacy-bundling]
npm error [--global-style] [--omit <dev|optional|peer> [--omit <dev|optional|peer> ...]]
npm error [--include <prod|dev|optional|peer> [--include <prod|dev|optional|peer> ...]]
npm error [--strict-peer-deps] [--foreground-scripts] [--ignore-scripts] [--no-audit]
npm error [--no-bin-links] [--no-fund] [--dry-run]
npm error [-w|--workspace <workspace-name> [-w|--workspace <workspace-name> ...]]
npm error [-ws|--workspaces] [--include-workspace-root] [--install-links]
npm error
npm error aliases: clean-install, ic, install-clean, isntall-clean
npm error
npm error Run "npm help ci" for more info
```

### Package Manager Version

npm 11.0.0

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

simon-abbott:
## Current Behavior

Currently certain actions, notably `nx release`, run `npm` with `legacy-peer-deps` enabled. This is an issue because running with `legacy-peer-deps` turned when it is normally off (which is the default) causes the installed dependency graph to be different. This then can break other scripts that expect the structure to be the same, including `npm ci`.

Changing the value of `legacy-peer-deps` causes npm to install a different graph, which can break builds and the `npm ci` command. Any users who rely on `legacy-peer-deps` should already have it set in an `.npmrc` file, so this should not affect them. It does, however, fix bugs with using `nx` in repos that _don't_ have `legacy-peer-deps` enabled, which is the vast majority of them.

## Expected Behavior

Running `nx` commands, especially `nx release`, shouldn't change the structure of the installed packages at all.

## Related Issue(s)

Fixes #22066
Fixes #29537

---

Note: I was unable to fully test this locally because the `e2e` target the contribution guide says to run no longer exists, and trying to run `e2e-local` fails with `Error: Jest: Got error running globalSetup - /workspaces/nx/e2e/utils/global-setup.ts, reason: 1` ([example run](https://staging.nx.app/runs/opdQWr4Vay)). I tried asking about this error in the Discord, but didn't get a reply, so I'm just posting this PR anyway in the hopes that CI passes.


FrozenPandaz:
For context, we've tried removing this in previous major versions. It ended up breaking people so we should proceed with caution when doing this change.

simon-abbott:
Do you happen to remember any issues that were filed around that? I'd like to take a look and make sure we're not going to repeat history.

That being said, I really can't see any potential issues unless people are relying on `legacy-peer-deps` being set without mandating it in their `.npmrc` file. And if that's the case then it's a really easy fix on their end that we can add to migration notes.

FrozenPandaz:
Yes, I got more context around the issues.

The story here is mostly around Angular where some packages like `@ngrx/store` keep a strict peer dependency on the current major version of `@angular/core`. The Nx team keeps a close eye on the Angular framework releases and starts working on supporting the new version ahead of time so we're ready to release support for new major versions of Angular the same week that it is released. The Ngrx team does not have this capacity so their support for the new major version of Angular usually lags for weeks to a month after.

Now this doesn't mean that older versions of `@ngrx/store` are not compatible with `@angular/core`, we test that the version of `@ngrx/store` is functionally compatible with the new major version of `@angular/core` hence we ignore these peer dependency warnings as we've heeded them and ensured that despite the warning, things do work properly.

Ok, so with this background, this does not have to be the way that it has to be. I'd personally also like to have it removed. We're open to other methods of ensuring that those users have an alternative way to define this npm configuration without relying on Nx's default behavior.

Generating a `.npmrc` with this configuration into the workspace for Ngrx users may work. @leosvelperez what do you think about this approach?

Does anyone have any other suggestions?

As for timing, I imagine that not all issues are known when changing this default behavior so I think it should be in a major version update. Unfortunately, we just released Nx 21, so the next major is in October 2025. This gives us more time to figure this out but means it won't be released for quite some time as well.

To workaround the issue where `npm ci` fails due to the `package-lock.json`, you can pass `--legacy-peer-deps` there or configure it in your `.npmrc` to make sure that `npm ci` runs via the same mode. I'm also open to other solutions that we can implement in a non-breaking fashion which does not alter this default behavior for the interim.

Sorry that despite this PR being opened for some time, we still do not have clear direction. Thank you for your work and also for understand that we wanted to be cautious about merging this given the history. We will continue to work towards a resolution.

simon-abbott:
> The story here is mostly around Angular where some packages like `@ngrx/store` keep a strict peer dependency on the current major version of `@angular/core`. The Nx team keeps a close eye on the Angular framework releases and starts working on supporting the new version ahead of time so we're ready to release support for new major versions of Angular the same week that it is released. The Ngrx team does not have this capacity so their support for the new major version of Angular usually lags for weeks to a month after.
> 
> Now this doesn't mean that older versions of `@ngrx/store` are not compatible with `@angular/core`, we test that the version of `@ngrx/store` is functionally compatible with the new major version of `@angular/core` hence we ignore these peer dependency warnings as we've heeded them and ensured that despite the warning, things do work properly.

Ok, I that makes sense as a background, but I feel like the real issue in that case is that users sometimes want to install a pair of packages that don't list explicit support for each other. I still think the right solution there is to tell those people to either wait for explicit support or to enable `legacy-peer-deps` themselves if they want to intentionally break the versioning requirements.

It's also worth considering the fact that `@ngrx/store` isn't the only package affected. What if a change updates both `@ngrx/store` and some other package that also has a peer dependency at the same time? `@ngrx/store` is safe to update, but the other package might not be. Since `legacy-peer-deps` can't be applied per-package, applying it to make `@ngrx/store` update when it claims it shouldn't has a very real risk of breaking other packages at the same time.

Finally, what if some update to Angular comes out that _isn't_ compatible with the latest `@ngrx/store`? If `legacy-peer-deps` is enabled then it will happily accept that update even though it is incompatible.

> Ok, so with this background, this does not have to be the way that it has to be. I'd personally also like to have it removed. We're open to other methods of ensuring that those users have an alternative way to define this npm configuration without relying on Nx's default behavior.
> 
> Generating a `.npmrc` with this configuration into the workspace for Ngrx users may work. @leosvelperez what do you think about this approach?

That would be a step forwards (at least then `npm` and `nx` would behave the same), and it is one that would solve my specific problem (since we don't use NgRx), but it still feels like fixing the symptom, not the issue.

> Does anyone have any other suggestions?

I would suggest having `nx` have no say in the matter at all. If users want to break modern peer dep requirements they can enable `legacy-peer-deps` on their own. If this is a common issue for `@ngrx/store` they should probably add a note to their `README`, since this mismatch issue affects everyone equally, not just those who use `nx`.

Also, if you have already tested to make sure that the `@ngrx/*` packages work with the latest Angular, why not work with the NgRx team to put up PRs updating the `peerDeps` range to include both versions? If they're truly compatible then surely that would be a better approach then overriding npm's safeguards, and if they're not then they shouldn't be installed at the same time regardless.

> As for timing, I imagine that not all issues are known when changing this default behavior so I think it should be in a major version update. Unfortunately, we just released Nx 21, so the next major is in October 2025. This gives us more time to figure this out but means it won't be released for quite some time as well.

I 100% agree. This is definitely a (potentially) breaking change, and should be versioned accordingly.

> To workaround the issue where `npm ci` fails due to the `package-lock.json`, you can pass `--legacy-peer-deps` there or configure it in your `.npmrc` to make sure that `npm ci` runs via the same mode. I'm also open to other solutions that we can implement in a non-breaking fashion which does not alter this default behavior for the interim.

We're setting the environment variable `npm_config_legacy_peer_deps=false` when calling affected `nx` commands, which has been working well enough for now.

> Sorry that despite this PR being opened for some time, we still do not have clear direction. Thank you for your work and also for understand that we wanted to be cautious about merging this given the history. We will continue to work towards a resolution.

Thanks for the information and update! I hope my feedback helps guide a final decision about how to deal with this issue. Let me know when a final direction has been decided on and this PR is ready to be merged and I will happily rebase it. I'm holding off on doing that now since I will most likely have to do it again before it gets merged, and I'd rather not waste time doing a rebase that will just have to be redone again down the line if I have the choice.

leosvelperez:
> Ok, I that makes sense as a background, but I feel like the real issue in that case is that users sometimes want to install a pair of packages that don't list explicit support for each other. I still think the right solution there is to tell those people to either wait for explicit support or to enable `legacy-peer-deps` themselves if they want to intentionally break the versioning requirements.
> 
> It's also worth considering the fact that `@ngrx/store` isn't the only package affected. What if a change updates both `@ngrx/store` and some other package that also has a peer dependency at the same time? `@ngrx/store` is safe to update, but the other package might not be. Since `legacy-peer-deps` can't be applied per-package, applying it to make `@ngrx/store` update when it claims it shouldn't has a very real risk of breaking other packages at the same time.
> 
> Finally, what if some update to Angular comes out that _isn't_ compatible with the latest `@ngrx/store`? If `legacy-peer-deps` is enabled then it will happily accept that update even though it is incompatible.

100%, we have nothing to dispute there. We agree that Nx always forcing `legacy-peer-deps` is incorrect and that we should remove it. It's a workaround we added a while ago to bypass peer deps issues caused by packages that were not as frequently updated but were working well from our testing.

> That would be a step forwards (at least then `npm` and `nx` would behave the same), and it is one that would solve my specific problem (since we don't use NgRx), but it still feels like fixing the symptom, not the issue.
>
> I would suggest having nx have no say in the matter at all. If users want to break modern peer dep requirements they can enable legacy-peer-deps on their own.

We'll have more time to discuss the final solution given we can only do it in the next major. I personally agree with you. I think Nx shouldn't decide for the user and shouldn't generate the `.npmrc` file, the same way it shouldn't force `legacy-peer-deps` on users. We'll discuss this more internally and update here.

> If this is a common issue for @ngrx/store they should probably add a note to their README, since this mismatch issue affects everyone equally, not just those who use nx.
>
> Also, if you have already tested to make sure that the `@ngrx/*` packages work with the latest Angular, why not work with the NgRx team to put up PRs updating the `peerDeps` range to include both versions? If they're truly compatible then surely that would be a better approach then overriding npm's safeguards, and if they're not then they shouldn't be installed at the same time regardless.

This is a bit more nuanced. NgRx doesn't necessarily see it as an issue. From their perspective, users should wait for them to support any given Angular major version. This would be the same with any package that is dependent on others. They need to explicitly support new versions before they can be used with those versions. Even though the NgRx team know and [even state that new major versions are supported by a lower version, their stance is to keep each NgRx major aligned solely with a single major version of Angular](https://github.com/ngrx/platform/issues/4603).

---

To reiterate, we acknowledge the current state is not great, and we should stop forcing the `legacy-peer-deps` behavior. Whether we generate an `.npmrc` or not is to be decided. We'll talk about it more internally. As mentioned, we can't do this change right now because it would be a breaking change. So, we need to wait until the next major version. We'll keep this updated.

simon-abbott:
> 100%, we have nothing to dispute there. We agree that Nx always forcing `legacy-peer-deps` is incorrect and that we should remove it. It's a workaround we added a while ago to bypass peer deps issues caused by packages that were not as frequently updated but were working well from our testing.

> We'll have more time to discuss the final solution given we can only do it in the next major. I personally agree with you. I think Nx shouldn't decide for the user and shouldn't generate the `.npmrc` file, the same way it shouldn't force `legacy-peer-deps` on users. We'll discuss this more internally and update here.

That all makes sense! I just wanted to make sure my stance was spelled out so it can be taken into consideration during the decision process.

> This is a bit more nuanced. NgRx doesn't necessarily see it as an issue. From their perspective, users should wait for them to support any given Angular major version. This would be the same with any package that is dependent on others. They need to explicitly support new versions before they can be used with those versions. Even though the NgRx team know and [even state that new major versions are supported by a lower version, their stance is to keep each NgRx major aligned solely with a single major version of Angular](https://github.com/ngrx/platform/issues/4603).

Huh. What an odd policy, and not one that I agree with. But if that's their stance then what can you do.

> To reiterate, we acknowledge the current state is not great, and we should stop forcing the `legacy-peer-deps` behavior. Whether we generate an `.npmrc` or not is to be decided. We'll talk about it more internally. As mentioned, we can't do this change right now because it would be a breaking change. So, we need to wait until the next major version. We'll keep this updated.

No worries! There are multiple workarounds (setting `npm_config_legacy_peer_deps=false` or enabling `legacy-peer-deps` for all installs), so this isn't a critical issue, it would just be nice to get that override out of our scripts. I can definitely wait until the next major cycle.

leosvelperez:
I'm closing this in favor of #33014. Thanks for bringing this up and taking the time to contribute!

simon-abbott:
> I'm closing this in favor of #33014. Thanks for bringing this up and taking the time to contribute!

@leosvelperez Fair enough, but I did notice that #33014 doesn't have an analogy for the regression test I added [here](https://github.com/nrwl/nx/pull/29726/files#diff-f43b2a8a746c164b1e904b39c696ae4c6991092fd28e7681eb08820b0fcd1208). You might want to consider copying that over too.

PR #29727, opened 2025-01-23
test(release): test that peer deps don't get broken

simon-abbott:
> [!WARNING]
> This is a test PR making sure that the new test in #29726 actually does anything. It SHOULD NOT BE MERGED. I am doing this here because I cannot for the life of me get the e2e tests to run locally.

## Related Issue(s)

Fixes #


simon-abbott:
Ok perfect, glad I tested it!

PR #29736, opened 2025-01-23
fix(rspack): should be inferred by default

ndcunningham:
This PR updates the Rspack Plugin to be inferred by default.

### Currently
When you generate a project using rspack it would not be inferred and would add the executor to `project.json`.

### After
Generating a project using rspack will add it to inferred plugins inside `nx.json` and update the `rspack.config.js` to be a standard config and use Nx plugins.

### DIFF
--- a/e2e/rspack/tests/rspack.spec.ts
+++ b/e2e/rspack/tests/rspack.spec.ts
@@ -9,8 +9,10 @@ import {
   updateFile,
   runCLI,
   runCommand,
+  createFile,
+  readJson,
 } from '@nx/e2e/utils';
-import { execSync } from 'child_process';
+
 import { writeFileSync } from 'fs';
 import { join } from 'path';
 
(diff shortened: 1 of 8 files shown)


packages/react/src/generators/application/lib/bundlers/add-rspack.ts
```diff
@@ -19,38 +18,11 @@ export async function initRspack(
...
-      )
-    ),
-    tsConfig: joinPathFragments(options.appProjectRoot, 'tsconfig.app.json'),
-    target: 'web',
-    newProject: true,
-    framework: 'react',
-  });
-  tasks.push(rspackTask);
-}
-
```
Coly010, line 53: If someone decides they dont want crystal - or their workspace already has `useInferencePlugins: false` what would happen with this missing?

ndcunningham: It will behave the same as Webpack. I tested this everything should be working.

e2e/rspack/tests/rspack.spec.ts
```diff
@@ -24,12 +26,28 @@ describe('rspack e2e', () => {
...
+    const nxJSON = readJson('nx.json');
+    const rspackPlugin = nxJSON.plugins.find(
+      (plugin) => plugin.plugin === '@nx/rspack/plugin'
+    );
+
+    expect(rspackPlugin).toBeDefined();
+  });
+
+  // This is disabled as the generator is no longer relevant
+  xit('should create rspack root project and additional apps', async () => {
```
Coly010, line 50: Instead of just skipping this test, can we actually remove it?

Coly010 (approved):
I've approved it to unblock merging, but could you also add an e2e that 100% confirms non-crystal usage

PR #29762, opened 2025-01-27
Revert "fix(core): support subpath exports when constructing the project graph"

jaysoo:
Reverts nrwl/nx#29577

There is an issue with workspaces using tsconfig path alias, and when an invalid import is found, it'll match on the wrong package due to the name matching the prefix.


leosvelperez (approved):

FrozenPandaz (approved):

robbtraister:
@jaysoo can you elaborate on the issue that caused this to be reverted? I'm happy to take another shot at supporting subpath imports, but your description is not sufficient for me to understand the problem.

leosvelperez:
@robbtraister we're restoring the functionality in https://github.com/nrwl/nx/pull/29795. It's a more comprehensive fix that covers a few more scenarios and prevents the issue we saw while testing the pre-release versions.

PR #29776, opened 2025-01-28
fix(react): add release option for @nx/react:lib --publishable

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react/src/generators/library/library.ts
+++ b/packages/react/src/generators/library/library.ts
@@ -6,11 +6,10 @@ import {
   GeneratorCallback,
   installPackagesTask,
   joinPathFragments,
-  readNxJson,
+  readProjectConfiguration,
   runTasksInSerial,
   Tree,
-  updateJson,
-  updateNxJson,
+  updateProjectConfiguration,
   writeJson,
 } from '@nx/devkit';
 import { getRelativeCwd } from '@nx/devkit/src/generators/artifact-name-and-directory-utils';
@@ -37,6 +36,11 @@ import {
 } from '@nx/js/src/utils/typescript/ts-solution-setup';
 import { determineEntryFields } from './lib/determine-entry-fields';
 import { sortPackageJsonFields } from '@nx/js/src/utils/package-json/sort-fields';
+import {
+  addReleaseConfigForNonTsSolution,
+  addReleaseConfigForTsSolution,
+  releaseTasks,
+} from '@nx/js/src/generators/library/utils/add-release-config';
 
 export async function libraryGenerator(host: Tree, schema: Schema) {
   return await libraryGeneratorInternal(host, {
@@ -77,7 +81,7 @@ export async function libraryGeneratorInternal(host: Tree, schema: Schema) {
 
   if (options.isUsingTsSolutionConfig) {
     writeJson(host, `${options.projectRoot}/package.json`, {
-      name: options.importPath,
+      name: options.importPath ?? options.name,
       version: '0.0.1',
       ...determineEntryFields(options),
       nx: options.parsedTags?.length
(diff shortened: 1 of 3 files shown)


packages/react/src/generators/library/library.ts
```diff
@@ -236,11 +240,19 @@ export async function libraryGeneratorInternal(host: Tree, schema: Schema) {
...
 
-  if (options.publishable || options.buildable) {
-    updateJson(host, `${options.projectRoot}/package.json`, (json) => {
-      json.name = options.importPath;
-      return json;
-    });
+  if (options.publishable) {
+    const projectConfiguration = readProjectConfiguration(host, options.name);
+    if (options.isUsingTsSolutionConfig) {
+      addReleaseConfigForTsSolution(host, options.name, projectConfiguration);
```
JamesHenry, line 246: Missing an await here?

JamesHenry (changes requested):

packages/react/src/generators/library/library.ts
```diff
@@ -103,7 +107,7 @@ export async function libraryGeneratorInternal(host: Tree, schema: Schema) {
   tasks.push(lintTask);
 
   // Set up build target
-  if (options.buildable && options.bundler === 'vite') {
```
JamesHenry, line 106: I see that you marked buildable as deprecated, and at the same time have removed these references to it.

Is bundler already always being set reliably? Is removing buildable a safe thing to do? (e.g. it's deprecated but not removed)

packages/react/src/generators/library/schema.d.ts
```diff
@@ -3,6 +3,7 @@ import type { SupportedStyles } from '../../../typings/style';
 
 export interface Schema {
   appProject?: string;
+  /** @deprecated Use bundler instead. */
```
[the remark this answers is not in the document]
xiongemi: it is marked deprecated in schema.json. for the bundler option, the default is none. teven if buildable is true, without setting bundler value, it will not add the build option

packages/react/src/generators/library/library.ts
xiongemi: bundler has a default value none. if the bundler is none, it will not add a build target regardless buildable value.

packages/react/src/generators/library/lib/create-files.ts
```diff
@@ -68,10 +68,7 @@ export function createFiles(host: Tree, options: NormalizedSchema) {
     });
   }
 
-  if (
-    (options.publishable || options.buildable) &&
-    !options.isUsingTsSolutionConfig
-  ) {
+  if (options.publishable && !options.isUsingTsSolutionConfig) {
```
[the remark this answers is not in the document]
xiongemi: currently logic is that even if buildable is true, without a bundler set, it will still create a package.json, but there is no use of the this package.json

packages/react/src/generators/library/library.ts
xiongemi: added

packages/react/src/generators/library/library.ts
xiongemi: changed it back

packages/react/src/generators/library/lib/create-files.ts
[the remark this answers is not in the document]
xiongemi: change it back

packages/react/src/generators/library/schema.d.ts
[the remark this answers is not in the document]
xiongemi: changed it back

JamesHenry (approved):

PR #29804, opened 2025-01-29
fix(js): do not infer `tsc` tasks with verbose output by default

### LINKED ISSUE #28677: Can't use `--clean` flag with `@nx/js/typescript` plugin
### Current Behavior

Currently, it looks like the [--clean](https://nx.dev/nx-api/js/executors/tsc#clean) flag leads to an error when used with the `@nx/js/typescript` plugin:

```
> tsc --build --emitDeclarationOnly --pretty --verbose --clean

error TS6370: Options 'clean' and 'verbose' cannot be combined.

Warning: command "tsc --build --emitDeclarationOnly --pretty --verbose --clean" exited with non-zero status code
```

I see this issue is known per [these code comments](https://github.com/nrwl/nx/blob/master/packages/js/src/plugins/typescript/plugin.ts#L228-L232), but thought I'd open an issue to track it since there doesn't appear to be a way to work around this without using a different executor. Thanks!

### Expected Behavior

Should be able to use the `--clean` flag with the `@nx/js/typescript` plugin without an error, at least in some configuration.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Configure a target with the `@nx/js/typescript` plugin
2. Provide the `--clean` option with a boolean value
3. Run the target
4. Observe an error

### Nx Report

Node           : 20.11.1
OS             : darwin-arm64
yarn           : 4.5.1

nx             : 20.0.6
@nx/js         : 20.0.6
@nx/workspace  : 20.0.6
@nx/devkit     : 20.0.6
typescript     : 5.4.5
Registered Plugins:
@nx/js/typescript

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

leosvelperez:
## Current Behavior

The `@nx/js/typescript` plugin infers tasks with `--verbose`. This can prevent users from running the same task with `--clean.` It can also produce a lot of logs that might not be too relevant.

## Expected Behavior

The `@nx/js/typescript` plugin should not infer tasks with `--verbose`. This is more aligned with other tools.

A new plugin option `verboseOutput` is added to allow inferring all tasks with `--verbose` if desired.

Note: This revealed that some things were working (e.g., `dependsOn`) because all the `typecheck` commands inferred by the different plugins matched. As soon as the command is different, the different inferred tasks are not merged, which is expected. We shouldn't rely on that, and each plugin inferring the task should set the right options/metadata. The different plugins were updated in this PR accordingly (they don't have the verbose option).

We'll follow up on this later, so only the `@nx/js/typescript` plugin infers the `typecheck` task. This is a breaking change so it will be for Nx v21.

## Related Issue(s)

Fixes #28677

### DIFF
--- a/packages/js/src/plugins/typescript/plugin.spec.ts
+++ b/packages/js/src/plugins/typescript/plugin.spec.ts
@@ -71,7 +71,7 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
               "targets": {
                 "typecheck": {
                   "cache": true,
-                  "command": "tsc --build --emitDeclarationOnly --pretty --verbose",
+                  "command": "tsc --build --emitDeclarationOnly",
                   "dependsOn": [
                     "^typecheck",
                   ],
@@ -126,7 +126,7 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
               "targets": {
                 "typecheck": {
                   "cache": true,
-                  "command": "tsc --build --emitDeclarationOnly --pretty --verbose",
+                  "command": "tsc --build --emitDeclarationOnly",
                   "dependsOn": [
                     "^typecheck",
                   ],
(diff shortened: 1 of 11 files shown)


Coly010 (approved):

jaysoo (approved):
LGTM

PR #29815, opened 2025-01-30
docs(core): update ts migration recipe

isaacplmann:
Updates migration to TS project references recipe


docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -172,26 +172,79 @@ The root `tsconfig.json` file should extend `tsconfig.base.json` and not include
...
+{% tabs %}
+{% tab label="Non-buildable library" %}
+
+```json {% fileName="libs/ui/package.json" %}
+{
+  "name": "@myorg/ui",
+  "exports": {
+    ".": {
+      "types": "./src/index.d.ts",
+      "import": "./src/index.js",
```
electrocnic, line 215: I would suggest to add a short sentence explaining the difference of when to use *.js and when to use *.ts here.
(As far as I understood: Use *.ts if the lib is not buildable, use *.js if it is buildable?)

electrocnic: actually just saw right now that these new sections are non-buildable vs. buildable, so please double-check if these settings are correct.

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -211,12 +264,14 @@ Each project's `tsconfig.json` file should extend the `tsconfig.base.json` file
 }
 ```
 
-Each project's `tsconfig.lib.json` file extends the project's `tsconfig.json` file and adds `references` to the `tsconfig.lib.json` files of project dependencies.
+Each project's `tsconfig.lib.json` file extends the root `tsconfig.base.json` file and adds `references` to the `tsconfig.lib.json` files of project dependencies.
```
electrocnic, line 267: A short explanation **why** we suddenly need to extend the .lib.jsons from the base.jsons instead of the lib-local tsconfig.jsons like before, would be much appreciated, because this change is confusing.

as I understand this documentation, instead of having:

```jsonc
// tsconfig.json:
{
  "extends": "../../tsconfig.base.json",
```
```jsonc
// tsconfig.lib.json:
{
  "extends": "./tsconfig.json",
```

we are now supposed to have:

```jsonc
// tsconfig.json:
{
  "extends": "../../tsconfig.base.json",
```
```jsonc
// tsconfig.lib.json:
{
  "extends": "../../tsconfig.base.json",
```

which does not really make sense to me and adds the need to over-duplicate lines which could be derived from the common tsconfig.json (even tho I know that extends is overwriting instead of deepmerging, but it does not overwrite everything, I think one layer is merged).

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -172,26 +172,79 @@ The root `tsconfig.json` file should extend `tsconfig.base.json` and not include
...
+{% /tabs %}
+
 {% callout type="warning" title="Package Names with Multiple Slashes" %}
 The `package.json` name can only have one `/` character in it. This is more restrictive than the TypeScript path aliases. So if you have a project that you have been referencing with `@myorg/shared/ui`, you'll need to make the `package.json` name be something like `@myorg/shared-ui` and update all the import statements in your codebase to reference the new name.
 {% /callout %}
 
 ## Update Individual Project TypeScript Configuration
 
-Each project's `tsconfig.json` file should extend the `tsconfig.base.json` file and list `references` to the project's dependencies.
+Each project's `tsconfig.json` file should extend the `tsconfig.base.json` file and list `references` to the project's dependencies. Remove any `compilerOptions` listed and combine them with the options listed in the `tsconfig.lib.json` and `tsconfig.spec.json` files.
```
electrocnic, line 247: as explained below, I am questioning this, as I think, the compileroptions are indeed merged with the extends option in a tsconfig, but not deepmerged. meaning: primitive entries will either be overwritten or merged with other entries, but nested records will be overwritten completely. (not sure what the intention is behind this instruction otherwise).

I did not verify this behavior, but it is my observation and I think my projects should not be able to compile if that was not the case, as I rely on this assumption for many files. Also I found a little backing for my claim in the second comment below this post: https://stackoverflow.com/a/71355500/3433137
"@cbdeveloper it merges compilerOptions -- you can see by running tsc --showconfig"

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -234,8 +289,10 @@ The project's `tsconfig.spec.json` does not need to reference project dependenci
 
 ```jsonc {% fileName="libs/ui/tsconfig.spec.json" %}
 {
-  "extends": "./tsconfig.json",
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    // outDir should be local to the project and not in the same folder as any other tsconfig.*.json
+    "outDir": "./out-tsc/spec"
```
electrocnic, line 295: why? Is it safe to have everything (build outputs) in the workspace root outdir? Any reasons to not do it that way?

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
```diff
@@ -252,6 +309,38 @@ The project's `tsconfig.spec.json` does not need to reference project dependenci
...
+3. Make sure the `build.lib.name` matches the full name of the project, including the organization.
+
+```ts {% fileName="libs/ui/vite.config.ts" %}
+export default defineConfig({
+  // ...
+  plugins: [
+    // any needed plugins, but remove nxViteTsPaths
+    react(),
+    nxCopyAssetsPlugin(['*.md', 'package.json']),
+    dts({
```
electrocnic, line 327: what is dts, a short link or description would be helpful

electrocnic (changes requested):

docs/shared/recipes/tips-n-tricks/switch-to-workspaces-project-references.md
isaacplmann: Good catch.  I had the wrong settings for non-buildable

jaysoo (approved):

PR #29817, opened 2025-01-30
fix(react-native): add release option to publisable react-native library

xiongemi:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/react-native/src/generators/library/library.ts
+++ b/packages/react-native/src/generators/library/library.ts
@@ -8,10 +8,12 @@ import {
   names,
   offsetFromRoot,
   ProjectConfiguration,
+  readJson,
   runTasksInSerial,
   toJS,
   Tree,
   updateJson,
+  updateProjectConfiguration,
   writeJson,
 } from '@nx/devkit';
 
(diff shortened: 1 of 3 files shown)


packages/react-native/src/generators/library/library.ts
```diff
@@ -188,13 +192,21 @@ async function addProject(
       nx: {
         tags: options.parsedTags?.length ? options.parsedTags : undefined,
       },
-      files: options.publishable ? ['dist', '!**/*.tsbuildinfo'] : undefined,
+      files: options.publishable
+        ? ['dist', 'src/index.ts', '!**/*.tsbuildinfo']
```
JamesHenry, line 196: The TypeScript file gets published? That doesn't seem right?

JamesHenry (changes requested):

ndcunningham (approved):

JamesHenry (changes requested):
@xiongemi Whilst this is green and approved, if it were to be merged it would immediately start failing because of the changes to versioning which have landed. The branch needs to be rebased and the snapshots of the command output in the new e2e tests will be need to be updated please

JamesHenry (approved):

PR #29820, opened 2025-01-30
fix(core): allow postTaskExecution hook without preTaskExecution hook

HaasStefan:
## Current Behavior

`postTaskExecution` hook is only called if `preTaskExecution` hook is exported in the plugin.

## Expected Behavior

`postTaskExecution` hook should be independent from `preTaskExecution` hook. I.e., if `preTaskExecution` hook is not exported, `postTaskExecution` hook should still be called after a task execution.

## Related Issue(s)

Fixes #

Fixed `postTaskExecution` hook


Cammisuli (approved):

FrozenPandaz (approved):

PR #29846, opened 2025-02-03
fix(misc) update typescript-project-linking.md typo

kategable:
## Related Issue(s)

Fixes #
Changing `naive` to `native`.


lppedd:
Maybe in the case _naive_ was used to convey something that looks easy/simple but that will bring up issues at some point.

isaacplmann:
I did mean "naive" as in simple or your first instinct. What would be a better way to phrase that so it is more clear?

kategable:
> I did mean "naive" as in simple or your first instinct. What would be a better way to phrase that so it is more clear?

Ow got it. Maybe a "basic way"...

PR #29860, opened 2025-02-03
feat(angular): add convert-to-rspack generator

Coly010:
## Current Behavior
Nx currently does not offer an automated method for switching from an Angular Webpack build to an Angular Rspack build.

## Expected Behavior
Provide a generator `convert-to-rspack` in `@nx/angular` that will allow conversion from an Angular Webpack build to an Angular Rspack build.

Usage: `nx g convert-to-rspack --project=myapp`

## TODO
- [x] handle more builder options
- [x] take existing custom webpack configs and migrate into the rspack config that is created


packages/angular/src/generators/convert-to-rspack/convert-to-rspack.ts
```diff
@@ -0,0 +1,96 @@
...
+  const { project: projectName } = schema;
+  const project = readProjectConfiguration(tree, projectName);
+  const tasks: GeneratorCallback[] = [];
+
+  const createConfigOptions: Partial<CreateConfigOptions> = {
+    root: project.root,
+  };
+  const buildAndServeTargetNames: string[] = [];
+
+  for (const [targetName, target] of Object.entries(project.targets)) {
```
HaasStefan, line 28: Maybe, add a validation check before to print a warning in case none of the executors are present. In case a user is trying to convert a project with custom executors.

Coly010: Yes there is a lot more that needs to go into this generator.

it’s still in draft for now to reflect this status

LayZeeDK:
question: Any way to make this work with Nx 18.3 and Angular 17.3?

LayZeeDK:
Guide at https://angular-rspack.dev/guide/migration/from-webpack

Coly010:
> question: Any way to make this work with Nx 18.3 and Angular 17.3?

Currently Angular Rspack is only supporting Angular 19

leosvelperez (approved):

FrozenPandaz (approved):

PR #29871, opened 2025-02-04
feat(vite): add support for vite 6

Coly010:
## Current Behavior
Nx currently supports Vite at version 5. Nx does not generate vite configurations using Vite 6 which has been released.
 

## Expected Behavior
Nx should use Vite 6 for vite projects.

## TODO
- [x] Add Package Update Migrations for Existing Projects
- [x] Add AST migrations to handle breaking change in resolve.extensions
- [x] Add migration to install `jsx` or `jiti` to handle processing TS postcss config files


packages/vite/src/migrations/update-20-5-0/update-resolve-conditions.md
```diff
@@ -0,0 +1,51 @@
+#### Update `resolve.conditions` to include defaults
+
+The `resolve.conditions` option had previously had defaults provided by Vite. These defaults were `['module', 'browser', 'development|production']`.
```
jaysoo, line 3: Wording seems off. I'd say something like:

In previous Vite versions, the `resolve.condition` option had a default value of `[...]`. This default was removed in Vite 6, so this migration adds it to your existing configuration to ensure that the behavior remains intact.

packages/vite/src/migrations/update-20-5-0/update-resolve-conditions.md
```diff
@@ -0,0 +1,51 @@
...
+
+{% tabs %}
+{% tab label="Before" %}
+
+```{% fileName="vite.config.ts" %}
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+  resolve: {
+    conditions: ['require'],
```
jaysoo, line 26: Would this be merged with the default in previous Vite versions? Or should the migration keep the conditions exactly how they are specified in the config?

packages/vite/src/migrations/update-20-5-0/update-resolve-conditions.ts
```diff
@@ -0,0 +1,86 @@
...
+        resolvePropertyNode,
+        CONDITIONS_PROPERTY_SELECTOR
+      );
+      if (conditionsPropertyNodes.length === 0) {
+        // do not add conditions property if it does not already exist
+        continue;
+      } else {
+        const conditionPropertyNode = conditionsPropertyNodes[0];
+        const conditionsArrayValues = JSON.parse(
+          conditionPropertyNode.getText().replace(/'/g, '"')
```
jaysoo, line 70: A backtick (`) could be used too, should we replace this, or maybe we don't need `JSON.parse`?

packages/vite/src/utils/generator-utils.ts
```diff
@@ -547,7 +549,7 @@ ${
 import { defineConfig } from 'vite';
 ${imports.join(';\n')}${imports.length ? ';' : ''}
 
-export default defineConfig({
+export default defineConfig(() => ({
```
jaysoo, line 552: Is this `defineConfig` change necessary?

Coly010: it is to support test config now unfortuantely

hugonasciutti: Hi,

If you encountered this error or it keeps happening:

image

I managed to fix it with: `as UserConfig`
```tsx
import { defineConfig, UserConfig } from 'vite'
export default defineConfig({} as UserConfig)
```

Maybe won't even need the `'v8' as const`

hugonasciutti: I believe it doesn't change the Vite behavior, it only helps TS with the typing. This `defineConfig` has a lot of overloading too.

image

packages/vite/src/migrations/update-20-5-0/update-resolve-conditions.ts
Coly010: Parsing as JSON helps us get an actual array, which we can convert to a set to prevent duplicate entries. I'll add back tick to this.

jaysoo (approved):

PR #29895, opened 2025-02-05
fix(nextjs): enhance support for custom server with SWC configuration

ndcunningham:
This pull request contains a few changes to enhance our swc support for Next.js with a custom server.

### Issues
Currently, we have a few issues with our configuration when using executors for Next.js with a custom server:

1. The custom server does not have an independent build configuration.
2. The custom server does not have an independent output directory.
3. Serving via `@nx/next-server` or via `@nx/js:node` with configurations `production` and `development` does not always work. (These are contained inside `project.json`).

### Changes
All the above issues have been addressed

1. We now have an independent swc build configuration called`.server.swrc` (_follows the same format as `.eslintrc`, `.babelrc`_) etc...
2. Now each custom server output will be named `{app}-server` such that if you have multiple custom servers for multiple apps the names will not clash.
3. Serving now works out of the box but can be adjusted to suit your needs via updating the custom server entry file `main.ts`

### DIFF
--- a/e2e/next/src/next-legacy.test.ts
+++ b/e2e/next/src/next-legacy.test.ts
@@ -279,4 +279,44 @@ describe('@nx/next (legacy)', () => {
     await killPort(prodServePort);
     await killPort(selfContainedPort);
   }, 600_000);
+
+  it('should support --custom-server flag (swc)', async () => {
+    const appName = uniq('app');
+
+    runCLI(
+      `generate @nx/next:app ${appName} --no-interactive --custom-server --linter=eslint --unitTestRunner=jest`,
+      { env: { NX_ADD_PLUGINS: 'false' } }
+    );
+
+    // Check for custom server files added to source
+    checkFilesExist(`${appName}/server/main.ts`);
+    checkFilesExist(`${appName}/.server.swcrc`);
+
+    const result = runCLI(`build ${appName}`);
+
+    checkFilesExist(`dist/${appName}-server/server/main.js`);
+
+    expect(result).toContain(
+      `Successfully ran target build for project ${appName}`
+    );
+  }, 300_000);
+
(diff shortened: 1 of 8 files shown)


e2e/next/src/next-legacy.test.ts
```diff
@@ -279,4 +279,44 @@ describe('@nx/next (legacy)', () => {
...
+    const appName = uniq('app');
+
+    runCLI(
+      `generate @nx/next:app ${appName} --no-interactive --custom-server --linter=eslint --unitTestRunner=jest`,
+      { env: { NX_ADD_PLUGINS: 'false' } }
+    );
+
+    // Check for custom server files added to source
+    checkFilesExist(`${appName}/server/main.ts`);
+    checkFilesExist(`${appName}/.swcrc.server`);
```
jaysoo, line 293: Let's use `.server.swcrc` to keep consistent with jest tests using `.spec.swcrc`. For RC files like `.eslintrc` and `.babelrc`, the optional extension usually goes to the end, so `.server` would imply it is the extension.

packages/next/src/utils/add-swc-to-custom-server.ts
```diff
@@ -14,8 +14,13 @@ import {
 } from '@nx/js/src/utils/versions';
 import { addSwcConfig } from '@nx/js/src/utils/swc/add-swc-config';
 
-export function configureForSwc(tree: Tree, projectRoot: string) {
-  const swcConfigPath = joinPathFragments(projectRoot, '.swcrc');
+export function configureForSwc(
+  tree: Tree,
+  projectRoot: string,
+  swcConfigName = '.swcrc',
+  additonalExludes = []
```
jaysoo, line 21: Minor: Needs `[] as string[]` or else it is inferred as `any[]`.

jaysoo (approved):

PR #29905, opened 2025-02-06
feat(remix): add continuous task support

Coly010:
## Current Behavior
When `serve, dev, start` targets are inferred by the `@nx/remix` plugin, they are not inferring `continuous`.

## Expected Behavior
When `NX_CONTINUOUS_TASKS_PREVIEW` env var exists, infer the `continuous` property.

### DIFF
--- a/packages/remix/src/plugins/plugin.ts
+++ b/packages/remix/src/plugins/plugin.ts
@@ -305,6 +305,7 @@ function devTarget(
   isUsingTsSolutionSetup: boolean
 ): TargetConfiguration {
   const devTarget: TargetConfiguration = {
+    continuous: true,
     command:
       remixCompiler === RemixCompiler.IsVte
         ? 'remix vite:dev'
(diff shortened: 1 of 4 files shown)


packages/remix/src/plugins/plugin.ts
```diff
@@ -138,13 +137,25 @@ async function createNodesInternal(
...
+      join(context.workspaceRoot, configFilePath),
+      join(
+        context.workspaceRoot,
+        getLockFileName(detectPackageManager(context.workspaceRoot))
+      ),
+    ].map(hashFile),
+    hashObject(options),
+    hashObject(packageJson),
+  ]);
+  const hash = `${nodeHash}_${configFilePath}`;
```
FrozenPandaz, line 158: You should do these changes in a separate PR to `master`.

packages/remix/src/plugins/plugin.ts
```diff
@@ -305,6 +316,7 @@ function devTarget(
   isUsingTsSolutionSetup: boolean
 ): TargetConfiguration {
   const devTarget: TargetConfiguration = {
+    continuous: true,
```
FrozenPandaz, line 319: I think this is the only change we need here.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #29908, opened 2025-02-06
feat(testing): support vitest 3

Coly010:
## Current Behavior
We currently still generate Vitest configuration using Vitest v1.

Vitest v3 has released so Nx is pretty far behind now.

## Expected Behavior
Update vitest version to v3 and use it to configure new projects.
Existing workspaces using v1 will continue to use v1.


packages/react/src/generators/application/__snapshots__/application.spec.ts.snap
```diff
@@ -222,8 +222,8 @@ if (import.meta.vitest) {
   });
 
   it('should have a greeting as the title', () => {
-    const { getByText } = render(<App />);
-    expect(getByText(new RegExp('Welcome plain', 'gi'))).toBeTruthy();
+    const { getAllByText } = render(<App />);
```
jaysoo, line 225: How come this needed to change?

packages/vite/src/utils/version-utils.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export async function isVitestV1(tree: Tree) {
+  let installedVitestVersion = await getInstalledVitestVersionFromGraph();
+  if (!installedVitestVersion) {
+    installedVitestVersion = getInstalledReactVersion(tree);
+  }
+  return major(installedVitestVersion) === 1;
+}
+
+export function getInstalledReactVersion(tree: Tree): string {
```
jaysoo, line 45: Why does the function contain "React"?

packages/react/src/generators/application/__snapshots__/application.spec.ts.snap
Coly010: React Testing Library is returning multiple elements containing the text

Coly010: only one exists, but the parents seem to be included rather than the last child

packages/vite/src/utils/version-utils.ts
Coly010: Leftover - should be changed to vitest

packages/vite/src/utils/version-utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  const pkgJson = readJson(tree, 'package.json');
+  const installedReactVersion =
+    pkgJson.dependencies && pkgJson.dependencies['vitest'];
+
+  if (
+    !installedReactVersion ||
+    installedReactVersion === 'latest' ||
+    installedReactVersion === 'next'
```
jaysoo, line 53: Looks like `next` is not a valid tag, Vitest uses `latest` and `beta`.

jaysoo (approved):

PR #29909, opened 2025-02-06
fix(vite): ignore vite temp files in eslint config

leosvelperez:
## Current Behavior

Vite config temp files can sometimes cause errors to be thrown by ESLint.

## Expected Behavior

Vite config temp files should be ignored by ESLint.

## Related Issue(s)

Fixes #


Coly010 (approved):

Coly010:
I think Remix also has it's own add vite to git ignore -does this need updated to also include eslint?

jaysoo (approved):

packages/remix/src/generators/application/lib/ignore-vite-temp-files.ts
```diff
@@ -0,0 +1,75 @@
+import { ensurePackage, readJson, stripIndents, type Tree } from '@nx/devkit';
+import { getPackageVersion } from '../../../utils/versions';
+
+export async function ignoreViteTempFiles(
```
leosvelperez, line 4: I decided to duplicate this helper for now. It should not change much in the future, and I'd like to avoid exposing an internal util from the `@nx/vite` package, which currently doesn't expose any utils. The `@nx/remix:application` generator has a custom vite setup, so it doesn't use the `@nx/vite:init` generator.

AgentEnder (approved):

PR #29913, opened 2025-02-06
feat(nx-dev): webinar page

isaacplmann:
Adds a webinar page and a script to pull the webinar content from Notion


juristr (changes requested):
When opening a past webinar it doesn't link to anything

Screenshot 2025-02-07 at 15 13 58

Also the image on the right looks weird. Are we going to change it? Also the combination between the right-side image and the large preview of the upcoming webinar is strange. Maybe we don't even need an image on the right 🤔 

Screenshot 2025-02-07 at 15 12 56

We should also link the page in the resources menu

image

juristr (approved):

PR #29917, opened 2025-02-06
feat(testing): disable jest runtime by default when inferring targets

jaysoo:
This PR updates `@nx/jest/plugin` such that `disableJestRuntime` option is true by default. Users will need to set it to false to bring in `jest-config` and `jest-runtime` to compute atomized targets. We're leaving it as an option if anyone runs into discrepancies between our calculation and what jest-runtime calculates for test files within a project.

## Current Behavior
Jest runtime is used by default and is potentially slow if you use many transforms, presets, etc. in the jest config.

## Expected Behavior
Jest runtime is not used by default, and users have to option of enabling it.

## Related Issue(s)

Fixes #


packages/jest/src/plugins/plugin.spec.ts
```diff
@@ -352,354 +354,14 @@ describe('@nx/jest/plugin', () => {
 
       const results = await createNodesFunction(
         ['proj/jest.config.js'],
-        { targetName: 'test' },
+        { targetName: 'test', disableJestRuntime },
         context
       );
 
       expect(results).toMatchSnapshot();
```
FrozenPandaz, line 361: Can you change this to `matchInlineSnapshot`? I want them both to actually assert against the same snapshot... not their own respective snapshots which happen to be the same at the moment.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #29919, opened 2025-02-06
feat(react): Update react-router-dom to latest

ndcunningham:
Update `react-router-dom` and `remix` package versions to latest.
Currently, there should be no breaking changes from our generators.


Coly010:
I think this just needs a migration for the package versions and it should be good.

Coly010 (approved):

PR #29965, opened 2025-02-11
feat(react): add react-router plugin

ndcunningham:
This PR introduces the React Router plugin in Nx. 
The new functionality adds a react-router plugin entry into `nx.json`, projects that are React-Router V7 via `react-router.config.(m|c)?[jt]s` will have their targets inferred.

### Changes
Update the React plugin to have a react-router (RR V7) plugin export.
The RR V7 will only infer targets if we have a `react-router.config.(m|c)?[jt]s` and also a `vite.config.(m|c)?[jt]s`. 

Under the hood the RR V7 CLI uses vite for compilation. 
That being said, apps are not limited to only use vite for RR V7. Should you choose to use it the compilation will not be done via RR V7 CLI.


packages/react/react-router-plugin.ts
```diff
@@ -0,0 +1,4 @@
+export {
+  createNodesV2,
+  ReactRouterPluginOptions,
+} from './src/plugins/react-router-plugin';
```
jaysoo, line 4: Let's make the path `@nx/react/router-plugin` so we don't repeat "react" twice.

packages/vite/src/plugins/plugin.ts
```diff
@@ -123,6 +123,16 @@ async function createNodesInternal(
...
+  const hasReactRouterConfig = siblingFiles.some((configFile) => {
+    const parts = configFile.split('.');
+    return (
+      parts[0] === 'react-router' && parts[1] === 'config' && parts.length > 2
+    );
+  });
+  if (hasReactRouterConfig) {
+    return {};
+  }
+
```
Coly010, line 135: This is too eager.

Users might still be using vitest to test their project which is inferred by the Vite plugin. 

This file shows how we avoid remix for the build and serve targets in the `buildViteTargets` function. 

We should replicate that to ensure vitest targets are still inferred correctly.

Coly010 (changes requested):

packages/vite/src/plugins/plugin.ts
ndcunningham: Yeah, good point updated.

Coly010 (approved):

maik-bol:
We're currently blocked since were missing this plugin. I'd rather not have to depend on a specific PR version so any ETA on when this is going to be merged?

jonsoku-dev:
I really want this pr..! Thank you for your service!

FrozenPandaz (approved):

PR #29999, opened 2025-02-12
feat(webpack): use sass-embedded and modern-compiler for sass

Coly010:
## Current Behavior
Webpack and Rspack currently use `sass` and its Legacy API with `sass-loader`. 
There is also no method to pass stylePreprocessorOptions other than `includePaths` to the loaders.

## Expected Behavior
Switch to using `modern-compiler` api to remove deprecation warnings and improve build performance.
Allow users to choose between `sass` and `sass-embedded` for sass compiler implementation.

Expand the `stylePreprocesserOptions` interface to accept `includePaths`, `sassOptions` and `lessOptions` that will be passed to the appropriate loader.


packages/webpack/src/plugins/nx-webpack-plugin/nx-app-webpack-plugin-options.ts
```diff
@@ -215,6 +219,10 @@ export interface NxAppWebpackPluginOptions {
    * Use tsconfig-paths-webpack-plugin to resolve modules using paths in the tsconfig file.
    */
   useTsconfigPaths?: boolean;
+  /**
+   * The implementation of the SASS compiler to use. Can be either `sass` or `sass-embedded`. Defaults to `sass-embedded`.
+   */
+  sassImplementation?: 'sass' | 'sass-embedded';
```
jaysoo, line 225: Add a `TODO(v22)` so we can remove this an use `sass-embedded` by default. And also migrate existing users.

jaysoo (approved):

FrozenPandaz (approved):

PR #30003, opened 2025-02-12
feat(module-federation): add NxModuleFederationPlugin for inferred usage

Coly010:
## Current Behaviour
Currently, Module Federation with Nx is forced to use executors to provide the best DX.

## Expected Behaviour
As part of the transition to inferred targets, we will need Rspack plugins that replicates the DX provided by our executors.
Add `NxModuleFederationPlugin` and `NxModuleFederationDevServerPlugin` to handle this.

### DIFF
--- a/packages/module-federation/src/plugins/utils/build-static-remotes.ts
+++ b/packages/module-federation/src/plugins/utils/build-static-remotes.ts
@@ -1,0 +1,84 @@
+import { fork } from 'node:child_process';
+import { join } from 'path';
+import { createWriteStream } from 'node:fs';
+import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
+import { workspaceRoot } from '@nx/devkit';
+import { StaticRemoteConfig } from '../../utils';
+import { NxModuleFederationDevServerConfig } from '../models';
+
+export async function buildStaticRemotes(
+  staticRemotesConfig: Record<string, StaticRemoteConfig>,
+  options: NxModuleFederationDevServerConfig,
+  nxBin: string
+) {
+  const remotes = Object.keys(staticRemotesConfig);
+  if (!remotes.length) {
+    return;
+  }
+  const mappedLocationOfRemotes: Record<string, string> = {};
+  for (const app of remotes) {
+    mappedLocationOfRemotes[app] = `http${options.ssl ? 's' : ''}://${
+      options.host
+    }:${options.staticRemotesPort}/${staticRemotesConfig[app].urlSegment}`;
+  }
+
+  await new Promise<void>((res, rej) => {
+    console.log(`NX Building ${remotes.length} static remotes...`);
+    const staticProcess = fork(
+      nxBin,
+      [
+        'run-many',
+        `--target=build`,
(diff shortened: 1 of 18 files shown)


packages/module-federation/src/plugins/utils/build-static-remotes.ts
```diff
@@ -0,0 +1,84 @@
...
+      const ANSII_CODE_REGEX =
+        // eslint-disable-next-line no-control-regex
+        /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g;
+      const stdoutString = data.toString().replace(ANSII_CODE_REGEX, '');
+      stdoutStream.write(stdoutString);
+
+      // in addition to writing into the stdout stream, also show error directly in console
+      // so the error is easily discoverable. 'ERROR in' is the key word to search in webpack output.
+      if (stdoutString.includes('ERROR in')) {
+        console.log(stdoutString);
```
ndcunningham, line 57: should this be logged via `console.error`?

packages/module-federation/src/plugins/utils/build-static-remotes.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+
+      if (stdoutString.includes('Successfully ran target build')) {
+        staticProcess.stdout?.removeAllListeners('data');
+        console.info(`NX Built ${remotes.length} static remotes`);
+        res();
+      }
+    });
+    staticProcess.stderr?.on('data', (data) => console.log(data.toString()));
```
ndcunningham, line 66: Similar here, should this be logged via `console.error` it would show up as red colour if it is an error right?
Maybe easier to detect.

ndcunningham (commented):
It would be great to add some unit / e2e tests to be sure of its core functionality and prevent regressions down the line.

packages/module-federation/src/plugins/utils/build-static-remotes.ts
Coly010: This was copied from the existing `build-static-remotes` function that we have for executor usage. I think this is fine for now.

packages/module-federation/src/plugins/utils/build-static-remotes.ts
Coly010: Same answer as: https://github.com/nrwl/nx/pull/30003#discussion_r1956358498

packages/module-federation/src/plugins/utils/build-static-remotes.ts
ndcunningham: Do you not think it is a better DX for errors to be logged as `console.error`?

ndcunningham (approved):
Approved as to not block it from being merged

packages/module-federation/src/plugins/utils/build-static-remotes.ts
Coly010: Could be a good opportunity to change it 🤔 @jaysoo you originally added this logic for handling errors - is there a reason you used console.log and not console.error?

PR #30029, opened 2025-02-13
fix(core): change to use init generator during import

xiongemi:
## Current Behavior

- call init generator using implementationFactory, causing issue because schema.json file isnt respected, and then NX_INTERACTIVE is never set to true

## Expected Behavior

- change back to run init command

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/init/configure-plugins.ts
+++ b/packages/nx/src/command-line/init/configure-plugins.ts
@@ -1,19 +1,13 @@
 import * as createSpinner from 'ora';
 import { bold } from 'chalk';
+import { execSync } from 'child_process';
 
 import {
   getPackageManagerCommand,
   PackageManagerCommands,
 } from '../../utils/package-manager';
-import { GitRepository } from '../../utils/git-utils';
 import { output } from '../../utils/output';
-import { flushChanges, FsTree } from '../../generators/tree';
-import {
-  Generator as NxGenerator,
-  GeneratorCallback,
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/init/configure-plugins.ts
```diff
@@ -64,19 +59,16 @@ export async function installPlugin(
...
-    {}
+  execSync(
+    `${pmc.exec} nx g ${plugin}:init --keepExistingVersions ${
+      updatePackageScripts ? '--updatePackageScripts' : ''
+    } ${verbose ? '--verbose' : ''}`,
+    {
+      stdio: [0, 1, 2],
+      cwd: repoRoot,
+      windowsHide: false,
+    }
```
Cammisuli, line 70: This should match closer to the older implementation here:
https://github.com/nrwl/nx/blob/68db3e032cfe24c66b44a230033fe3032536ba53/packages/nx/src/command-line/add/add.ts#L121-L162

* use the init generator that was found
* try/catch the exec

xiongemi: actually, i added it back it. i did not remove above code

Cammisuli: Oh I see, sorry I missed it because it was collapsed 😅

Cammisuli (approved):

ndcunningham (approved):

PR #30036, opened 2025-02-14
fix(devkit): support pnpm catalogs

### LINKED ISSUE #30035: PNPM Catalogs are not supported
### Current Behavior

When I run `nx generate plugin` it throws the following error:

>  NX   The package.json lists a version of typescript that Nx is unable to validate - (catalog:my-catalog)

With this typescript dependency in my package.json:

```json
    "typescript": "catalog:my-catalog",
```

### Expected Behavior

In a PNPM workspace utilizing the catalogs feature, Nx should not validate for a valid semver.

### GitHub Repo

_No response_

### Steps to Reproduce

1. pnpm workspace
2. use catalogs
3. run nx g plugin

### Nx Report

```shell
failed
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

HaasStefan:
## Current Behavior

`nx generate plugin` throws the following error if a package like typescript is using the PNPM catalog feature in the package.json:

>  NX   The package.json lists a version of typescript that Nx is unable to validate - (catalog:my-catalog)

```json
"typescript": "catalog:my-catalog",
```

This is due too the semver validation that fails with the catalog.

## Expected Behavior

If I am using a pnpm catalog, the package dependency should not be checked for a valid semver.

## Related Issue(s)

Fixes #30035

### DIFF
--- a/packages/devkit/src/utils/semver.ts
+++ b/packages/devkit/src/utils/semver.ts
@@ -11,6 +11,10 @@ export function checkAndCleanWithSemver(
     return newVersion;
   }
 
+  if (version.startsWith('~') || version.startsWith('^')) {
+    newVersion = version.substring(1);
+  }
+
   if (!valid(newVersion)) {
     throw new Error(
       `The package.json lists a version of ${pkgName} that Nx is unable to validate - (${version})`
--- a/packages/devkit/src/utils/semver.ts
+++ b/packages/devkit/src/utils/semver.ts
@@ -6,12 +6,9 @@ export function checkAndCleanWithSemver(
 ): string {
   let newVersion = version;
 
-  if (valid(newVersion)) {
+  const isPnpmCatalog = newVersion.startsWith('catalog:');
+  if (valid(newVersion) || isPnpmCatalog) {
     return newVersion;
-  }
-
-  if (version.startsWith('~') || version.startsWith('^')) {
-    newVersion = version.substring(1);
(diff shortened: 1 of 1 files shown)


AgentEnder (changes requested):
This util is used to get version numbers which are later used with semver helper methods, see:  https://github.com/nrwl/nx/blob/2d195007b0558c5d157e2705c3d5fb916cefa3ee/packages/angular/src/generators/setup-tailwind/lib/detect-tailwind-installed-version.ts#L16-L21

As such, we can't really just return the version with the `catalog:` protocol still there as that would result in downstream code failing as its expecting a semver.

I think to truly fix the issue you are seeing we'd need to be able to resolve the version, but I also don't love putting custom code in here for parsing pnpm-workspace.yaml and such.

HaasStefan:
@AgentEnder, what are you suggesting? Would using a parser to retrieve the pnpm version number be alright if it is not colocated in this file?

leosvelperez:
Hey @HaasStefan! Thanks for taking the time to contribute this!

I'm going to close it for the time being. We will spend some time over the next few weeks to provide comprehensive support for PNPM Catalogs that would cover Craigory's concerns and will properly fix the attached issue.

PR #29774, opened 2025-01-28
fix(core): Optimize loading local plugin TS options

ethantkoenig:
## Current Behavior

All of the TS source trees are unnecessarily traversed when a local plugin is being registered.

## Expected Behavior

TS source trees are not unnecessarily traversed.

## Related Issue(s)

Fixes #29102 (see issue for more context)

PR #30062, opened 2025-02-16
fix(core): Respect `extends` in local plugin TS options

### LINKED ISSUE #30007: Default imports broken in custom executor in 20.4.1
### Current Behavior

I have a custom executor (in a workspace plugin) which has this import:
```ts
import glob from 'fast-glob';
```
Running the executor will output:
```
 NX   (0 , _fastglob.default) is not a function
```

Even if I fix this particular case (don't use the default import), there will be many others in libraries that I cannot control, e.g. when `jest-config` is reading the `jest.config.ts` that has a default export, it will import the config file as an object with the default key (this is the other direction reading ts code from js).

I assume there's something in the way SWC is being used to run custom executors.

### Expected Behavior

I would expect default exports to be imported correctly:
1. importing a JS library with a default export from a TS script should not error.
2. importing a TS file with a default export from another library should import the default export and not an object with the default property.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create a custom executor with the nx plugin
2. Use a default export

or (reverse)

1. Create a custom executor with the nx plugin
2. use `{ parseJestConfig } from '@nx/jest/src/executors/jest/jest.impl'` to read a `jest.config.ts` with a default export.

### Nx Report

```shell
Node           : 20.18.1
OS             : darwin-arm64
pnpm           : 9.15.4

nx                 : 20.4.1
@nx/js             : 20.4.1
@nx/jest           : 20.4.1
@nx/eslint         : 20.4.1
@nx/workspace      : 20.4.1
@nx/devkit         : 20.4.1
@nx/eslint-plugin  : 20.4.1
@nx/nest           : 20.4.1
@nx/node           : 20.4.1
@nx/plugin         : 20.4.1
@nx/webpack        : 20.4.1
typescript         : 5.7.3
Local workspace plugins:
         workspace-plugin
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Works in 20.4.0, seems it was introduced in 20.4.1

ethantkoenig:
Fix a bug introduced in https://github.com/nrwl/nx/pull/29774.

## Current Behavior

If a local plugin's `tsconfig.json` uses `extends`, the referenced base config is not loaded.

## Expected Behavior

If a local plugin's `tsconfig.json` uses `extends`, the referenced base config is loaded.

## Related Issue(s)

Fixes #30007


alumni:
```
TypeError: s.codePointAt is not a function
    at codePointAt (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:14500:12)
    at codePointUnchecked (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:12073:12)
    at Object.scan (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:12684:18)
    at nextTokenWithoutCheck (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:33362:36)
    at nextToken (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:33372:12)
    at Object.parseJsonText2 [as parseJsonText] (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:32931:5)
    at parseJsonText (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:32759:17)
    at parseConfigFileTextToJson (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:42125:26)
    at readJsonOrUndefined (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:20956:18)
    at readJson (<root>/node_modules/.pnpm/typescript@5.7.3/node_modules/typescript/lib/typescript.js:20960:10)
```

packages/nx/src/plugins/js/utils/typescript.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { readTsConfigOptions } from './typescript';
+import { mkdtemp, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
+
+describe.only('readTsConfigOptions', () => {
```
leosvelperez, line 6:
```suggestion
describe('readTsConfigOptions', () => {
```

packages/nx/src/plugins/js/utils/typescript.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('should handle extends', async () => {
+    const dir = await mkdtemp(join(tmpdir(), 'typescript.spec'));
+    await writeFile(
+      join(dir, 'a.json'),
+      JSON.stringify({ extends: './b.json' })
+    );
+    await writeFile(
+      join(dir, 'b.json'),
+      JSON.stringify({ compilerOptions: { strict: true } })
+    );
```
leosvelperez, line 16: Let's use the `TempFs` class exported by `packages/nx/src/internal-testing-utils.ts`. You can see an example of how to use it at `packages/nx/src/config/to-project-name.spec.ts`.

leosvelperez (changes requested):
Thank you for contributing!

The change makes sense. I just requested a couple of changes to the test file but nothing major. If you don't have time to address them, please let me know and I'll update the PR.

alumni:
> Thank you for contributing!
> 
> The change makes sense. I just requested a couple of changes to the test file but nothing major. If you don't have time to address them, please let me know and I'll update the PR.

@leosvelperez it still fails for workspace plugins (see the stack trace above)

leosvelperez:
@alumni thanks for the extra feedback and testing!

@ethantkoenig we should update the host to:

```ts
const host: Partial<ts.ParseConfigHost> = {
  ...tsModule.sys,
  readDirectory: () => [],
};
```

That way, we have everything that's needed to read the config files (including extended config files from third-party packages) while still avoiding scanning directories, which are only used to identify the `fileNames` property in the parsed configuration.

alumni:
> ```ts
> const host: Partial<ts.ParseConfigHost> = {
>   ...tsModule.sys,
>   readDirectory: () => [],
> };
> ```

Seems to works for project config that I have.

`Partial` _might_ not be necessary, `ParseConfigHost` seems to be a subset of `System`.

Thanks both :)

packages/nx/src/plugins/js/utils/typescript.spec.ts
ethantkoenig: Whoops, fixed.

packages/nx/src/plugins/js/utils/typescript.spec.ts
ethantkoenig: Done

ethantkoenig:
> @ethantkoenig we should update the host to:
> 
> ```ts
> const host: Partial<ts.ParseConfigHost> = {
>   ...tsModule.sys,
>   readDirectory: () => [],
> };
> ```
>
> That way, we have everything that's needed to read the config files (including extended config files from third-party packages) while still avoiding scanning directories, which are only used to identify the `fileNames` property in the parsed configuration.

Done
 
> Please add extra test cases that cover extending from third-party packages.

Added a test, though I was unable to come up with a third-party test case that fails with the old code, so the test might not be super useful. @alumni if you have a minimal reproducible example you are able to easily share, that would be helpful.

leosvelperez (approved):

PR #30078, opened 2025-02-17
cleanup(js): migrate to nanospinner

43081j:
Migrates the `js` package to use `nanospinner` for CLI spinners.

Same as we have already done for the core package.


JamesHenry (approved):

JamesHenry (changes requested):
I just remembered we didn't actually land any spinner changes yet. I need to revisit the threads we had about exception recover etc

43081j:
from what i remember, it doesn't make much difference which we use of nanospinner or picospinner

picospinner introduced a `disableNewLineEnding` option for the behaviour you'd been asking for (ability to somehow output stuff while the spinner is still running). and nanospinner does the same by default

though the preference is you should stop the spinner when you want to output something...

i went with nanospinner here since it tolerates that stuff out of the box

43081j:
@JamesHenry if i catch this up, can you help me get it over the line?

this will drop chalk from the tree finally i think (its a deep dependency thanks to ora)

jaysoo:
Hey, thanks again for the putting the effort on this PR. Just as with https://github.com/nrwl/nx/pull/29138#issuecomment-3676110795, I'm going to close this as it has gone stale.

PR #30079, opened 2025-02-17
cleanup(js): switch to picomatch

43081j:
Switches from `minimatch` to `picomatch`.

This should support almost every glob syntax that was previously supported, with the exception of some far edge case range patterns.


MaxKless (approved):

JamesHenry:
Thanks as always @43081j - sorry that there appears to be a conflict now, this should be good to be merged once that is fixed up

JamesHenry (approved):

PR #30082, opened 2025-02-18
fix(core): re-enable CRA migration to Vite

jaysoo:
This PR brings back the CRA migration that was missing since Nx 18.

## Current Behavior

`nx init` does not migrate CRA apps 

## Expected Behavior
`nx init` migrates CRA apps to Vite since CRA is deprecated

## Related Issue(s)

Fixes #


packages/nx/src/command-line/init/implementation/react/index.ts
```diff
@@ -32,41 +21,50 @@ type NormalizedOptions = Options & {
...
+export async function addNxToCraRepo(_options: Options) {
+  if (!_options.force) {
     checkForCustomWebpackSetup();
   }
 
-  output.log({ title: '🐳 Nx initialization' });
+  const options = await normalizeOptions(_options);
+  await addBundler(options);
+
+  execSync(`echo "node_modules" >> .gitignore`, {
```
JamesHenry, line 35: We should be using platform agnostic `fs` utilities, not shell commands, for these file updates

JamesHenry (dismissed):

juristr (approved):
works. thank you.

Issue #29429, opened 2024-12-19
Failed to convert JavaScript value `Null` into rust type `String`


adgoudz:
### Current Behavior

`nx graph` outputs an error for each tsconfig file in [agdz/sample-workspace](https://github.com/adgoudz/sample-workspace):

```
 NX   Failed to create nodes

An error occurred while processing files for the @nx/js/typescript plugin.
  - packages/foo/tsconfig.json: Failed to convert JavaScript value `Null` into rust type `String`
    Error: Failed to convert JavaScript value `Null` into rust type `String`
        at hashArray (/path/to/sample-workspace/node_modules/nx/src/hasher/file-hasher.js:9:12)
        at createNodesInternal (/path/to/sample-workspace/node_modules/@nx/js/src/plugins/typescript/plugin.js:97:50)
        at /path/to/sample-workspace/node_modules/@nx/js/src/plugins/typescript/plugin.js:41:95
        at /path/to/sample-workspace/node_modules/nx/src/project-graph/plugins/utils.js:10:33
        at Array.map (<anonymous>)
        at createNodesFromFiles (/path/to/sample-workspace/node_modules/nx/src/project-graph/plugins/utils.js:8:35)
        at exports.createNodesV2 (/path/to/sample-workspace/node_modules/@nx/js/src/plugins/typescript/plugin.js:41:60)
        at LoadedNxPlugin.createNodes (/path/to/sample-workspace/node_modules/nx/src/project-graph/plugins/internal-api.js:34:65)
        at LoadedNxPlugin.createNodes.<computed> (/path/to/sample-workspace/node_modules/nx/src/project-graph/plugins/internal-api.js:44:34)
  - ...
```

Other commands fail with:

```
 NX   Failed to process project graph. Run "nx reset" to fix this. Please report the issue
```

### Expected Behavior

Nx should ignore optional files when hashing nodes. In my case, Nx is failing to process the project graph when a project doesn't create a lockfile (i.e., package-lock.json).

### GitHub Repo

https://github.com/adgoudz/sample-workspace

### Steps to Reproduce

1. `npx nx graph`

### Nx Report

```shell
NX   Report complete - copy this into the issue template

Node           : 20.5.1
OS             : darwin-arm64
Native Target  : aarch64-macos
npm            : 10.9.2

nx (global)    : 20.2.2
nx             : 20.2.2
@nx/js         : 20.2.2
@nx/workspace  : 20.2.2
@nx/devkit     : 20.2.2
[...]

PR #30090, opened 2025-02-18
fix(core): make hash_array resilient to None values

JamesHenry:
## Current Behavior

If `null`/`undefined` is given to `hashArray` we end up with an unclear napi-rs error.

## Expected Behavior

The underlying `hash_array` on the rust side is resilient to `None` values, and can therefore handle when `null`/`undefined` are passed through from the JS side.

## Related Issue(s)

Related to #29429

### DIFF
--- a/packages/nx/src/native/hasher.rs
+++ b/packages/nx/src/native/hasher.rs
@@ -13,8 +13,8 @@ pub fn hash_array(input: Vec<Option<String>>) -> String {
         if s.is_none() {
             trace!("Encountered None value in hash_array input: {:?}", input);
         }
-        s.as_ref()
-    }).map(|s| s.as_str()).collect::<Vec<_>>().join(",");
+        s.as_deref()
+    }).collect::<Vec<_>>().join(",");
     let content = joined.as_bytes();
     hash(content)
 }
--- a/packages/nx/src/native/hasher.rs
+++ b/packages/nx/src/native/hasher.rs
@@ -8,8 +8,13 @@ pub fn hash(content: &[u8]) -> String {
 }
 
 #[napi]
-pub fn hash_array(input: Vec<String>) -> String {
-    let joined = input.join(",");
+pub fn hash_array(input: Vec<Option<String>>) -> String {
+    let joined = input.iter().filter_map(|s| {
+        if s.is_none() {
+            trace!("Encountered None value in hash_array input: {:?}", input);
+        }
+        s.as_ref()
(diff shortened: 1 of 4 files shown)


packages/nx/src/native/hasher.rs
```diff
@@ -8,8 +8,13 @@ pub fn hash(content: &[u8]) -> String {
...
 #[napi]
-pub fn hash_array(input: Vec<String>) -> String {
-    let joined = input.join(",");
+pub fn hash_array(input: Vec<Option<String>>) -> String {
+    let joined = input.iter().filter_map(|s| {
+        if s.is_none() {
+            trace!("Encountered None value in hash_array input: {:?}", input);
+        }
+        s.as_ref()
+    }).map(|s| s.as_str()).collect::<Vec<_>>().join(",");
```
Cammisuli, line 17: Use deref here so that we don't have to iterate through the vec again with the second map. 
```suggestion
    let joined = input.iter().filter_map(|s| {
        if s.is_none() {
            trace!("Encountered None value in hash_array input: {:?}", input);
        }
        s.as_deref()
    }).collect::<Vec<_>>().join(",");
```

If we didnt want to add the trace when with `is_none`, then we can do this, which is quicker:
```
input.into_iter()
        .flatten() // removes None and unwraps Some(String)
        .collect::<Vec<_>>()
        .join(",")
```

The time difference between the filter_map and flatten is miniscule, but still fun to know :D 
```
iter_filter_map_deref   time:   [68.009 µs 68.214 µs 68.425 µs]
                        change: [-1.2825% -0.5797% +0.2147%] (p = 0.15 > 0.05)
                        No change in performance detected.

into_iter_flatten       time:   [63.492 µs 63.643 µs 63.815 µs]
                        change: [-1.0299% -0.3958% +0.3385%] (p = 0.26 > 0.05)
                        No change in performance detected.
```

Cammisuli (changes requested):

packages/nx/src/native/hasher.rs
Cammisuli: for reference, heres the bench for the initial implementation with as_ref + as_str
```
iter_filter_map_ref     time:   [75.861 µs 76.254 µs 76.705 µs]
                        change: [-0.9861% -0.0645% +0.8159%] (p = 0.89 > 0.05)
                        No change in performance detected.
```

Cammisuli (approved):

PR #30111, opened 2025-02-19
fix(misc): fix misc issues in project generators for the ts solution setup

leosvelperez:
The following are the main changes in the context of the TS solution setup:

- Ensure `name` in `package.json` files is set to the import path for all projects
- Set `nx.name` in `package.json` files when the user provides a name different than the package name (import path)
- Clean up project generators so they don't set the `nx` property in `package.json` files unless strictly needed
- Fix `@nx/vue:application` generator so it creates the Nx config in a `package.json` file for e2e projects
- Ensure `@types/node` is installed in `vitest` generator
- Fix generated Vite config typing error (surfaced with Vite 6)
- Ensure `jsonc-eslint-parser` is installed when the `@nx/dependency-checks` rule is added to the ESLint config
- Misc minor alignment changes

## Related Issue(s)

Fixes #


jaysoo (commented):
The changes LGTM. I'll spend time testing it today.

jaysoo (approved):

PR #30114, opened 2025-02-19
fix(release): properly terminate case in switch statement

tristanzander:
This corrects an issue when the changelog renderer is incorrect and shows `Cannot read properties of undefined (reading 'map')`, but it hides the real error message.

## Related Issue(s)

Fixes #


JamesHenry (approved):
Thank you!

PR #30139, opened 2025-02-22
chore(core): nx plugin submission @unpowerpacking/codeowners

Jordan-Hall:
I'll totally understand if this is rejected. Its aiming just of OOS and small projects

### DIFF
--- a/community/approved-plugins.json
+++ b/community/approved-plugins.json
@@ -513,5 +513,10 @@
     "name": "nx-foundry",
     "description": "Foundry generators and inferred tasks for Nx",
     "url": "https://github.com/juliangsibecas/nx-foundry"
+  },
+  {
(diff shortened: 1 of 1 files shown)


isaacplmann:
I understand why you would want to write this plugin and fully support your ability to write whatever you want with open source code.  But we're also not obligated to use our platform to promote software that directly undermines one of our paid offerings.

OSS and small projects can use the official plugin for free already.

Jordan-Hall:
> I understand why you would want to write this plugin and fully support your ability to write whatever you want with open source code. But we're also not obligated to use our platform to promote software that directly undermines one of our paid offerings.
> 
> OSS and small projects can use the official plugin for free already.

Wait small projects can use the plugin for free? From my understanding the free was just for remote caching

PR #30151, opened 2025-02-24
fix(react-native): change build target to be crystalized

xiongemi:
## Related Issue(s)

Fixes #


jaysoo (commented):
Seems good. I need to test it out as well.

jaysoo (approved):
Tested locally. LGTM

PR #30187, opened 2025-02-26
fix(eslint): ensure that packages in the same group as updated together

jaysoo:
This PR updates `migrations.json` for `@nx/eslint` and `@nx/angular` so that `@typescript-eslint/*` and `@angular-eslint/*` packages that are installed in the workspaces are updated to the same version.

I've updated it for v20 and v19, but if we need to go back further we could. This will also require patch versions for each major that we want to fix.

**Note:** Previously there were two entries for `@nx/esilnt@20.4.0` that handles cases where one of `typescript-eslint` or `@typescript-eslint/eslint-plugin` are > 8. The packages in both entries overlap, so I collapsed them down to one entry that checks `typescript-eslint > 8`.

## Current Behavior
If user has `@typescript-eslint/rule-tester` or `@angular-eslint/test-utils` installed, our migrations do not update these versions. This makes them out of sync with `@typescript-eslint/utils`, etc. which can lead to problems.

## Expected Behavior
Packages in the same `@typescript-eslint/*` and `@angular-eslint/*` are updated together. If they are not installed, we don't add them to the workspace.

## Related Issue(s)

Fixes #


Coly010 (approved):

packages/eslint/migrations.json
```diff
@@ -104,6 +104,22 @@
         },
         "@typescript-eslint/utils": {
           "version": "^7.16.0"
+        },
+        "@typescript-eslint/eslint-plugin-tslint": {
```
JamesHenry, line 108: This package isn't published any more, it's last version was `7.0.2`

JamesHenry (changes requested):
tslint plugin is no longer published and doesn't have these versions

JamesHenry (approved):

PR #30191, opened 2025-02-26
fix(storybook): fix package.json updates so @storybook packages are in sync during migration

jaysoo:
This PR is the same as https://github.com/nrwl/nx/pull/30187 but for `@storybook` packages. We want to make sure that workspaces that have other `@storybook/*` packages installed have their versions updated along with the packages we use. Otherwise version mismatches can lead to errors due to changing APIs.

This PR also adds a conformance rule that prevents mistakes from going out in future migrations.


ndcunningham (approved):

Coly010 (approved):

tools/workspace-plugin/src/conformance-rules/migration-groups/index.ts
```diff
@@ -0,0 +1,123 @@
...
+        },
+        { missing: [] as string[], present: [] as string[] }
+      );
+      if (result.missing.length > 0) {
+        violations.push({
+          message: `Package.json updates for "${key}" is missing packages in a group: ${result.missing.join(
+            ', '
+          )}. Versions of packages in a group must be in sync. Other packages in the group: ${result.present.join(
+            ', '
+          )}`,
```
leosvelperez, line 97: I'm not sure mentioning other packages in the group is relevant for this specific violation. Maybe the violation can be reworded so it lists the missing packages (as is doing right now) and the version that they should have (if there aren't more than 1).

tools/workspace-plugin/src/conformance-rules/migration-groups/index.ts
```diff
@@ -0,0 +1,123 @@
...
+          sourceProject,
+          file: migrationsPath,
+        });
+      } else {
```
leosvelperez, line 101: I would remove the `else` and always check that the present packages don't have any mismatch. Right now, we only check it when all packages are present.

leosvelperez (approved):

PR #30209, opened 2025-02-27
fix(js): handle plugin correctly for non-buildable libs and don't add duplicated pattern to workspaces config

leosvelperez:
## Current Behavior

- Generating a non-buildable library in a workspace using the TS solution setup where a plugin registration for `@nx/js/typescript` already exists and doesn't configure a build target, results in that plugin registration being updated excluding the new project and a new registration being added including the project but inferring the build target.
- Generating a library in a sub-directory that matches a pattern in the package manager workspaces configuration, results in a more specific pattern being added to the workspace configuration.

## Expected Behavior

- Generating a non-buildable library in a workspace using the TS solution setup where a plugin registration for `@nx/js/typescript` already exists and doesn't configure a build target, should not modify that plugin registration and it should not add an extra one.
- Generating a library in a sub-directory that matches a pattern in the package manager workspaces configuration, should not add a more specific pattern.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/js/src/utils/package-manager-workspaces.ts
+++ b/packages/js/src/utils/package-manager-workspaces.ts
@@ -1,9 +1,10 @@
 import {
-  detectPackageManager,
   getPackageManagerVersion,
   output,
   readJson,
+  readNxJson,
   type GeneratorCallback,
+  type PackageManager,
   type Tree,
 } from '@nx/devkit';
 import picomatch = require('picomatch');
@@ -27,7 +28,13 @@ export function getProjectPackageManagerWorkspaceState(
 
   const patterns = getGlobPatternsFromPackageManagerWorkspaces(
     tree.root,
-    (path) => readJson(tree, path, { expectComments: true })
+    (path) => readJson(tree, path, { expectComments: true }),
+    (path) => {
+      const content = tree.read(path, 'utf-8');
+      const { load } = require('@zkochan/js-yaml');
+      return load(content, { filename: path });
+    },
+    (path) => tree.exists(path)
   );
   const isIncluded = patterns.some((p) =>
     picomatch(p)(join(projectRoot, 'package.json'))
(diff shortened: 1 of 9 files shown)


packages/js/src/utils/package-manager-workspaces.ts
```diff
@@ -58,7 +61,22 @@ export function isWorkspacesEnabled(
   return false;
 }
 
+function detectPackageManager(tree: Tree, dir: string = ''): PackageManager {
```
JamesHenry, line 64: Why was this inlined?

leosvelperez: The one from devkit is not tree-based. It works with the filesystem directly, but these are used in a generator context, which is tree-based.

JamesHenry: I appreciate what you are saying, but I imagine the reason this hasn't come up before is that the distinction isn't important, because the package manager (proxied via the lock file name) simply is not likely at all to change during the execution of a generator (and it certainly wouldn't during the ones in question here)

JamesHenry: I just don't love that this would be the 3rd time we have a detectPackageManager function and none of them share any logic...

leosvelperez: Yeah, that makes sense. I'll revert it.

Coly010: In CNW with TS + PM Workspaces - will this ever be executed when there are _no_ files persisted to disk already?

I don’t think so because we always create the sandbox first before running generators. 

However, having one that works with a tree might be useful for anyone Extending Nx with custom plugins, generators, create-package etc. 

Just not sure if it would be in the scope of this PR

leosvelperez: When the lock file doesn't exist on disk, the install package tasks would use the existing `detectPackageManager` function, so using it inside the generators before the file exists yields the same result. That's why, as James mentioned, it should be fine in most cases, if not all.

leosvelperez: @JamesHenry PR is updated now.

JamesHenry (approved):

jaysoo (approved):

PR #30211, opened 2025-02-27
fix(js): do not add typecheck target if tsc is used for build

jaysoo:
This PR adds support for skipping `typecheck` targets when using `@nx/js/typescript`. Inside `tsconfig.json` for each project, you can set `nx.addTypecheckTarget` to `false` to not infer `typecheck`.

This allows use to skip `typecheck` for JS projects using `tsc` to build. Since `tsc` is already used during build, we don't need to run `typecheck` that is just duplicated work.

## Current Behavior
JS libs using `tsc` to build will do typechecking twice. Once during `build` and once during `typecheck`.

## Expected Behavior
JS libs using `tsc` do not infer `typecheck` by default. And users can change this behavior by setting `nx.addTypecheckTarget` in `tsconfig.json`.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/packages/js/js-plugin.md
+++ b/docs/shared/packages/js/js-plugin.md
@@ -113,7 +113,7 @@ You can also set `typecheck` and `build` options to `false` to not infer the cor
 
 To disable `typecheck` task for a specific project, set the `nx.addTypecheckTarget` property to `false` in `tsconfig.json`.
 
-```json {% fileName="tsconfig.json" highlightLines=["10-12"] %}
+```json {% fileName="packages/pkg1/tsconfig.json" highlightLines=["10-12"] %}
 {
   "extends": "../../tsconfig.base.json",
   "files": [],
--- a/docs/shared/packages/js/js-plugin.md
+++ b/docs/shared/packages/js/js-plugin.md
@@ -63,6 +63,72 @@ Nx 20 updates the TS monorepo setup when using `--preset=ts`. The workspace is s
 To create with the older setup for TS monorepo with `compilerOptions.paths`, use `create-nx-workspace --preset=apps`.
 {% /callout %}
 
+### How @nx/js Infers Tasks
+
+The `@nx/js` plugin will add a `typecheck` task to projects that have a `tsconfig.json`. It adds a `build` task for projects that have a runtime tsconfig file (defaults to `tsconfig.lib.json`).
+
+### View Inferred Tasks
+
+To view inferred tasks for a project, open the [project details view](/concepts/inferred-tasks) in Nx Console or run `nx show project my-project` in the command line.
+
+### @nx/js Configuration
+
+The `@nx/js/typescript` plugin is configured in the `plugins` array in `nx.json`.
+
+```json {% fileName="nx.json" %}
+{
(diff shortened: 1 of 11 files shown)


Coly010 (approved):

e2e/vite/src/vite-ts-solution.test.ts
```diff
@@ -103,10 +103,5 @@ ${content}`
     expect(runCLI(`build ${reactApp}`)).toContain(
       `Successfully ran target build for project @proj/${reactApp} and 5 tasks it depends on`
     );
-
-    // check typecheck
-    expect(runCLI(`typecheck ${reactApp}`)).toContain(
-      `Successfully ran target typecheck for project @proj/${reactApp} and 6 tasks it depends on`
-    );
```
leosvelperez, line 110: Why is this removed? The app uses the `vite` bundler, so it should have a typecheck task. Or is it failing due to a different issue?

e2e/js/src/js-ts-solution.test.ts
```diff
@@ -129,23 +129,6 @@ ${content}`
       `Successfully ran target build for project @proj/${viteParentLib} and 5 tasks it depends on`
     );
 
-    // check typecheck
-    expect(runCLI(`typecheck ${esbuildParentLib}`)).toContain(
```
leosvelperez, line 133: Why are we removing all typecheck tasks? Should we only remove the one from `tscParentLib`?

docs/shared/packages/js/js-plugin.md
```diff
@@ -63,6 +63,72 @@ Nx 20 updates the TS monorepo setup when using `--preset=ts`. The workspace is s
 To create with the older setup for TS monorepo with `compilerOptions.paths`, use `create-nx-workspace --preset=apps`.
 {% /callout %}
 
+### How @nx/js Infers Tasks
+
+The `@nx/js` plugin will add a `typecheck` task to projects that have a `tsconfig.json`. It adds a `build` task for projects that have a runtime tsconfig file (defaults to `tsconfig.lib.json`).
```
leosvelperez, line 68:
```suggestion
The `@nx/js/typescript` plugin will add a `typecheck` task to projects that have a `tsconfig.json`. It adds a `build` task for projects that have a runtime tsconfig file (defaults to `tsconfig.lib.json`).
```

docs/shared/packages/js/js-plugin.md
```diff
@@ -63,6 +63,72 @@ Nx 20 updates the TS monorepo setup when using `--preset=ts`. The workspace is s
...
+    }
+  ]
+}
+```
+
+### Disable Typechecking
+
+To disable `typecheck` task for a specific project, set the `nx.addTypecheckTarget` property to `false` in `tsconfig.json`.
+
+```json {% fileName="tsconfig.json" highlightLines=["10-12"] %}
```
leosvelperez, line 116:
```suggestion
```json {% fileName="packages/pkg1/tsconfig.json" highlightLines=["10-12"] %}
```

leosvelperez (approved):

isaacplmann (approved):

PR #30224, opened 2025-02-28
fix: check if dependencies & devDependencies exist before deleting dep

### LINKED ISSUE #30222: removeDependenciesFromPackageJson: Ensure dependencies field exist
### Current Behavior

If there is no field `dependencies` or `devDependencies` migration that removes deps fails

### Expected Behavior

It should work. Add a `if` statement here :
`packages/devkit/src/utils/package-json.ts`
```js
      for (const dep of dependencies) {
        if (json.dependencies) delete json.dependencies[dep];
      }
      for (const devDep of devDependencies) {
        if (json.devDependencies) delete json.devDependencies[devDep];
      }
```

### GitHub Repo

_No response_

### Steps to Reproduce

1. Remove `dependencies` from your package.json
2. Run a migration that remove a dependency

### Nx Report

```shell
...
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

### LINKED ISSUE #30223: `@nx/vite:test` doesn't respect `reportsDirectory` when set in `project.json`
### Current Behavior

Using the `@nx/vite:test` executor I set the `reportsDirectory` in the option in `project.json`
The `reportsDirectory` from `project.json` is not taken into consideration, only when it's set in vite config.

### Expected Behavior

Using the `@nx/vite:test` executor I set the `reportsDirectory` in the option in `project.json`
The `reportsDirectory` from `project.json` is taken into consideration and is used correctly (like it does with `Jest`)

### GitHub Repo

_No response_

### Steps to Reproduce

1. Take any project or create a new one
2. Create a target that uses the `@nx/vite:test` executor
3. Set `reportsDirectory` in the `options` section

### Nx Report

```shell
No report.
Nx version: v19.8.13
```

### Failure Logs

```shell

```

### Package Manager Version

pnpm: v9.9.0

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

GrandSchtroumpf:
## Current Behavior
`removeDependenciesFromPackageJson` tries to delete dep from `undefined` if there is no `dependencies` or `devDependencies` in package.json

## Expected Behavior
First check if `dependencies` or `devDependencies` exists

## Related Issue(s)
Fixes #30222


samir-ajdarpasic:
@GrandSchtroumpf I think the linked issue is not the correct one.

packages/devkit/src/utils/package-json.ts
```diff
@@ -287,15 +287,18 @@ export function removeDependenciesFromPackageJson(
     )
   ) {
     updateJson(tree, packageJsonPath, (json) => {
-      for (const dep of dependencies) {
-        delete json.dependencies[dep];
+      if (json.dependencies) {
+        for (const dep of dependencies) {
+          delete json.dependencies[dep];
+        }
+        json.dependencies = sortObjectByKeys(json.dependencies);
```
ndcunningham, line 294: Since we are improving this code if there are no dependencies we could also remove the dependencies instead of attempting to sort it.

packages/devkit/src/utils/package-json.ts
```diff
@@ -287,15 +287,18 @@ export function removeDependenciesFromPackageJson(
...
       }
-      for (const devDep of devDependencies) {
-        delete json.devDependencies[devDep];
+      if (json.devDependencies) {
+        for (const devDep of devDependencies) {
+          delete json.devDependencies[devDep];
+        }
+        json.devDependencies = sortObjectByKeys(json.devDependencies);
```
ndcunningham, line 300: Same as above

packages/devkit/src/utils/package-json.ts
FrozenPandaz: I think we can do this separately.

FrozenPandaz (approved):

PR #30230, opened 2025-02-28
fix(bundling): fix esbuild to work with ts project references

ndcunningham:
## Current Behavior

If we are using `esbuild` as our bundler and ts project references (`--workspaces`) local libraries are not building are not resolved in the build artifacts.

## Expected Behavior

When using ts project references with esbuild all types libraries (buildable / non-buildable) should work out of the box.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
+++ b/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
@@ -141,7 +141,11 @@ export function createPathsFromTsConfigReferences(
   const {
     isValidPackageJsonBuildConfig,
   } = require('@nx/js/src/plugins/typescript/util');
-  const { readTsConfig } = require('@nx/devkit');
+  const { readTsConfig } = require('@nx/js');
+  const {
+    findRuntimeTsConfigName,
+  } = require('@nx/js/src/utils/typescript/ts-solution-setup');
+
   const deps = findAllProjectNodeDependencies(
     context.projectName,
     context.projectGraph
@@ -167,35 +171,21 @@ export function createPathsFromTsConfigReferences(
     }, [])
   );
 
-  const getTsConfigPath = (
-    projectType: string,
-    projectPath: string
-  ): string => {
-    if (projectType === 'app') {
-      return existsSync(join(projectPath, 'tsconfig.app.json'))
-        ? 'tsconfig.app.json'
-        : 'tsconfig.json';
-    }
-
-    return existsSync(join(projectPath, 'tsconfig.lib.json'))
-      ? 'tsconfig.lib.json'
-      : 'tsconfig.json';
-  };
-
   // for each dep we check if it contains a build target
   // we only want to add the paths for projects that do not have a build target
   return deps.reduce((acc, dep) => {
     const projectNode = context.projectGraph.nodes[dep];
     const projectPath = joinPathFragments(workspaceRoot, projectNode.data.root);
-    const projTsConfig = readTsConfig(
-      getTsConfigPath(projectNode.type, projectPath)
-    ) as any;
+    const resolvedTsConfigPath =
+      findRuntimeTsConfigName(projectPath) ?? 'tsconfig.json';
+    const projTsConfig = readTsConfig(resolvedTsConfigPath) as any;
 
     const projectPkgJson = readJsonFile(
       joinPathFragments(projectPath, 'package.json')
     );
 
     if (
+      projTsConfig &&
       !isValidPackageJsonBuildConfig(
         projTsConfig,
         workspaceRoot,
--- a/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
+++ b/packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
@@ -6,6 +6,7 @@ import {
   joinPathFragments,
   normalizePath,
   ProjectGraphProjectNode,
+  readJsonFile,
   workspaceRoot,
 } from '@nx/devkit';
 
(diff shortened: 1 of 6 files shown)


packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -123,6 +127,50 @@ export function buildEsbuildOptions(
...
+  );
+  const referencesAsPaths = new Set(
+    tsConfig.references.map((ref) => joinPathFragments(workspaceRoot, ref.path))
+  );
+
+  // for each dep we check if it contains a build target
+  // we only want to add the paths for projects that do not have a build target
+  return deps.reduce((acc, dep) => {
+    const projectNode = context.projectGraph.nodes[dep];
+    if (!projectNode.data.targets?.build) {
```
leosvelperez, line 154: Can we do anything about the hardcoded `build` task name? Folks might have different build task names.

packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -123,6 +127,50 @@ export function buildEsbuildOptions(
...
+      const projectPath = joinPathFragments(
+        workspaceRoot,
+        projectNode.data.root
+      );
+      const projectPkgJson = readJsonFile(
+        joinPathFragments(projectPath, 'package.json')
+      );
+      // if the project has a name and a main entry point, we add it to the paths
+      if (projectPkgJson?.name && projectPkgJson?.main) {
```
leosvelperez, line 163: We must consider `exports` first, and only if `exports` is not defined can we fall back to `main` and/or `module.`

packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -123,6 +127,50 @@ export function buildEsbuildOptions(
...
+  const {
+    findAllProjectNodeDependencies,
+  } = require('nx/src/utils/project-graph-utils');
+
+  const deps = findAllProjectNodeDependencies(context.projectName);
+  const tsConfig = readJsonFile(
+    joinPathFragments(context.root, 'tsconfig.json')
+  );
+  const referencesAsPaths = new Set(
+    tsConfig.references.map((ref) => joinPathFragments(workspaceRoot, ref.path))
```
leosvelperez, line 147: `ref.path` could be a path pointing to a tsconfig file. We're using this `Set` as if it's always a directory, but even though that's what we generate, it could be modified by the user and point to a file. We should normalize it to make sure it's a directory.

packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -123,6 +127,50 @@ export function buildEsbuildOptions(
...
+ * @param context ExecutorContext
+ */
+export function createPathsFromTsConfigReferences(
+  context: ExecutorContext
+): Record<string, string[]> {
+  const {
+    findAllProjectNodeDependencies,
+  } = require('nx/src/utils/project-graph-utils');
+
+  const deps = findAllProjectNodeDependencies(context.projectName);
```
leosvelperez, line 142: nit: we can provide the project graph we already have available to avoid the operation of retrieving it from the cache unnecessarily.
```suggestion
  const deps = findAllProjectNodeDependencies(
    context.projectName,
    context.projectGraph
  );
```

leosvelperez (changes requested):

packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -123,6 +127,142 @@ export function buildEsbuildOptions(
...
+
+      return acc;
+    }, [])
+  );
+
+  const getTsConfigPath = (
+    projectType: string,
+    projectPath: string
+  ): string => {
+    if (projectType === 'app') {
```
jaysoo, line 174: Not that `projectType` may not be set, especially in the new TS setup. I think we can just check for existence of `tsconfig.lib.json`, `tsconfig.app.json`, then `tsconfig.json` as fallback.

There is a `findRuntimeTsConfigName` util you can use, but if you need `tsconfig.json` as fallback then you need `findRuntimeTsConfigName(tree, projectRoot) ?? 'tsconfig.json')`.

packages/esbuild/src/executors/esbuild/lib/build-esbuild-options.ts
```diff
@@ -123,6 +127,142 @@ export function buildEsbuildOptions(
...
+      ? 'tsconfig.lib.json'
+      : 'tsconfig.json';
+  };
+
+  // for each dep we check if it contains a build target
+  // we only want to add the paths for projects that do not have a build target
+  return deps.reduce((acc, dep) => {
+    const projectNode = context.projectGraph.nodes[dep];
+    const projectPath = joinPathFragments(workspaceRoot, projectNode.data.root);
+    const projTsConfig = readTsConfig(
```
jaysoo, line 190: I think we can use `findRuntimeTsConfigName`, and if it returns `null` then assume that the project is non-buildable.

ndcunningham: `findRuntimeTsConfigName` requires `tree` but this is the executor that has context.

ndcunningham: `tree` which is required for `findRuntimeTsConfigName`  doesn't typical exists inside an executor and it doesn't seem like a common pattern to create it using `FsTree`.

jaysoo: It could be done like `isUsingTsSolutionSetup`, where `FsTree` is used if not passed.

```
export function isUsingTsSolutionSetup(tree?: Tree): boolean {
  tree ??= new FsTree(workspaceRoot, false);

  return (
    isUsingPackageManagerWorkspaces(tree) &&
    isWorkspaceSetupWithTsSolution(tree)
  );
}
```

jaysoo (approved):

leosvelperez (approved):

PR #30251, opened 2025-03-03
fix(nx): update using-tailwind-css-with-angular-projects.md

artkwi:


leosvelperez:

Please note that both the current content and your change are correct configurations. You would use one or the other depending on whether you want to extend or override the Tailwind CSS configuration. So, while correct, the current change in the PR does not fix an issue because the guide is not about deciding whether to extend or override the Tailwind CSS default theme, nor is it advising you to do one over the other. It just shows an example of a preset, and it has to pick one way.

I'm closing this for the time being. If you think there's still an issue and the guide needs to be updated, please explain your reasons to make this change.

PR #30268, opened 2025-03-05
fix(core): handle package manager workspaces configuration in move generator

leosvelperez:
## Current Behavior

Moving a project included in the package manager workspaces setup to a new destination that's not matched by that configuration results in the project not included in the package manager workspaces setup.

## Expected Behavior

Moving a project included in the package manager workspaces setup to a new destination that's not matched by that configuration should result in the new destination included in the workspaces setup.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/workspace/package.json
+++ b/packages/workspace/package.json
@@ -42,7 +42,9 @@
     "chalk": "^4.1.0",
     "enquirer": "~2.3.6",
     "tslib": "^2.3.0",
-    "yargs-parser": "21.1.1"
+    "yargs-parser": "21.1.1",
+    "picomatch": "4.0.2",
+    "@zkochan/js-yaml": "0.0.7"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/workspace/package.json
+++ b/packages/workspace/package.json
@@ -39,12 +39,12 @@
   },
   "dependencies": {
     "@nx/devkit": "file:../devkit",
+    "@zkochan/js-yaml": "0.0.7",
     "chalk": "^4.1.0",
     "enquirer": "~2.3.6",
-    "tslib": "^2.3.0",
-    "yargs-parser": "21.1.1",
     "picomatch": "4.0.2",
-    "@zkochan/js-yaml": "0.0.7"
+    "tslib": "^2.3.0",
+    "yargs-parser": "21.1.1"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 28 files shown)


packages/workspace/package.json
```diff
@@ -42,7 +42,9 @@
     "chalk": "^4.1.0",
     "enquirer": "~2.3.6",
     "tslib": "^2.3.0",
-    "yargs-parser": "21.1.1"
+    "yargs-parser": "21.1.1",
+    "picomatch": "4.0.2",
```
JamesHenry, line 46: Please can we put these in the correct order to avoid an unnecessary diff later?

JamesHenry: We should probably add a rule to enforce this programmatically

packages/workspace/src/utilities/typescript/ts-solution-setup.ts
```diff
@@ -75,3 +77,47 @@ export function isUsingTsSolutionSetup(tree?: Tree): boolean {
...
+) {
+  const isIncluded = isProjectIncludedInPackageManagerWorkspaces(
+    tree,
+    projectDir
+  );
+  if (isIncluded) {
+    return;
+  }
+
+  // If dir is "libs/foo" then use "libs/*" so we don't need so many entries in the workspace file.
```
JamesHenry, line 93: Do we really need to run the risk of side-effects at all?

Why not just go for the project's specific location and leave it at that? Folks can always adjust and generalize if they want to, but this way Nx will always be correct and the move will be side-effect free for other packages?

JamesHenry (changes requested):
See comments

packages/workspace/package.json
leosvelperez: Sure. I added them by using the `lint --fix`, so we should also look into having the rule add packages in the right order.

packages/workspace/src/utilities/typescript/ts-solution-setup.ts
leosvelperez: This makes sense. This file is actually a mirror of another helper in the JS package (to prevent circular deps). I will fix this in both places and will explore adding the pattern only if it doesn't pick up any extra projects. Otherwise, we explicitly add the project root. That way, we avoid including already existing projects that were not previously matched, while still being able to have more generic patterns to not pollute this file with a lot of entries.

Coly010 (approved):

FrozenPandaz (approved):

JamesHenry (approved):

PR #30306, opened 2025-03-07
chore(core): include bun mention in bug.yml

ludicast:
Adds mention of bun to packagemanager section.  Since the list of packagemanagers is growing and they all have the same `--version` flag (even deno does this), I removed the repeated flags.

## Related Issue(s)

Fixes #


JamesHenry:
Thanks!

PR #30316, opened 2025-03-09
feat(react): Add react-router to create-nx-workspace and react app generator

ndcunningham:
This pull request introduces improvements to React Router integration and removes the Remix preset.

## Key Changes:
- Updated `create-nx-workspace` to support React Router.
- Removed the Remix option from `create-nx-workspace`, but the package remains to support existing users.

## SSR & React Router Support
- New users who want SSR in their React apps can enable it via the React option and select React Router for SSR support.
- The ecosystem has shifted to migrating from Remix to React Router for SSR needs.
- This option is only available for plain React apps and uses Vite. Other types of React apps (Micro Frontends, Webpack, Rspack, etc.) remain unaffected.

## Default Routing Behavior
`--routing` is now enabled by default when creating a React app using `create-nx-workspace`, aligning with Angular’s default behaviour.


xiongemi:
just got a question: what is the diff between --routing and --react-router? if i pass in --routing, would be there other routing library to use?

packages/react/src/generators/application/application.ts
```diff
@@ -73,6 +75,35 @@ export async function applicationGeneratorInternal(
 
   const options = await normalizeOptions(tree, schema);
 
+  if (options.routing) {
```
xiongemi, line 78: is it possible to simply the logics here? combine these 2 options somehow?

packages/react/src/utils/versions.ts
```diff
@@ -17,6 +17,7 @@ export const typesReactIsVersion = '19.0.0';
 export const reactViteVersion = '^4.2.0';
 
 export const typesNodeVersion = '18.16.9';
+export const typesReactRouterNodeVersion = '^20.0.0';
```
xiongemi, line 20: so now we will generate apps with 2 node version, without react router, it will be at version 18, with reach router, it will be at version 20. i am just wondering, whether it will easier just to upgrade node to 20 for all.

ndcunningham: It is possible that we could just upgrade to node 20 the LTS is currently node v22. 
This is types as well so might be a good option to update it.

ndcunningham:
> just got a question: what is the diff between --routing and --useReactRouter? if i pass in --routing, would be there other routing library to use?

You get `react-router-dom` similar to what is on master. The `--useReactRouter` is to future-proof so we can also support **library** mode for react router without impacting users currently using `react-router-dom`

FrozenPandaz (approved):

xiongemi (approved):

isaacplmann (approved):

PR #30325, opened 2025-03-10
fix(release): properly handle deeply-transitive updates

### LINKED ISSUE #29375: Nx release command is not bumping all the dependents chain
### Current Behavior

I set up a new NX repository containing four libraries, each depending on the one above it in a cascading (waterfall) structure, as illustrated in the attached image. 
Image

When I make a change to `mylib4`, the `mylib` library version and the indirect dependent does not have their versions bumped during the release process. I expected `mylib` to reflect a version change, just like `mylib2`.

But you will see something like the following:
Image

### Expected Behavior

I expect `@my-workspace/mylib` is changed to also include a version bump and an update to the dependency `@my-workspace/mylib2` version. Even if there are no direct changes to `@my-workspace/mylib`, the update should still bubble up from `@my-workspace/mylib4`.
```
    "name": "@my-workspace/mylib",
-  "version": "0.0.1",
+  "version": "0.1.0",
    "dependencies": {
-     "@my-workspace/mylib2": "0.0.1"
+     "@my-workspace/mylib2": "0.1.0"
    }
```

### GitHub Repo

https://github.com/moroale93/nx-issue-test

### Steps to Reproduce

1. Install dependencies:
```bash
npm install
```
2. modify something on `mylib4` (like the exported variable value).
3. commit those changes (locally and the right [conventional commit standard](https://www.conventionalcommits.org/en/v1.0.0/)) 
4. run the release command (dry run for example)
```bash
nx release --skip-publish --dry-run --verbose
```

### Nx Report

```shell
➜  nx-issue-test git:(main) ✗ nx report

Node           : 18.20.4
OS             : darwin-arm64
npm            : 10.7.0

nx (global)            : 20.2.2
nx                     : 20.2.2
@nx/js                 : 20.2.2
@nx/jest               : 20.2.2
@nx/eslint             : 20.2.2
@nx/workspace          : 20.2.2
@nx/devkit             : 20.2.2
@nx/eslint-plugin      : 20.2.2
@nx/module-federation  : 20.2.2
@nx/react              : 20.2.2
@nx/web                : 20.2.2
@nx/webpack            : 20.2.2
typescript             : 5.6.3
Registered Plugins:
@nx/webpack/plugin
@nx/eslint/plugin
@nx/jest/plugin

➜  nx-issue-test git:(main) ✗
```

### Failure Logs

```shell
no failures
```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

This is my current nx release config:
```
"release": {
    "projects": ["libs/*"],
    "version": {
      "generatorOptions": {
        "fallbackCurrentVersionResolver": "disk",
        "updateDependents": "auto",
        "skipLockFileUpdate": true,
        "currentVersionResolver": "git-tag",
        "specifierSource": "conventional-commits"
      }
    },
    "changelog": {
      "projectChangelogs": true,
      "automaticFromRef": true
    },
    "projectsRelationship": "independent"
  }
```

### Observation

It looks like the `targetMatchesRequirement` variable on the `resolve-local-package-dependencies.js` file, is set to false, so `mylib` is not added to the `localPackageDependencies`. That it's due to the fact that on the previous execution `mylib2` got bumped, so the 2 versions does not match.
Indeed, if I change that variable to true, the release command works as expected.

simon-abbott:
## Current Behavior
Using `nx release` to update a package that has recursively-transitive dependencies (i.e. `pkg-a` -> `pkg-b` -> `pkg-c` -> `pkg-d`) fails to properly update the full version tree, resulting in failures.

## Expected Behavior
The versions should all be updated properly, even if they are far away in the dependency graph.

## Related Issue(s)

Fixes #29375

### DIFF
--- a/packages/js/src/generators/release-version/release-version.spec.ts
+++ b/packages/js/src/generators/release-version/release-version.spec.ts
@@ -1221,6 +1221,22 @@ Valid values are: "auto", "", "~", "^", "="`,
             },
           ],
         },
+        'project-with-double-transitive-dependency-on-my-lib': {
+          projectRoot:
+            'libs/project-with-double-transitive-dependency-on-my-lib',
+          packageName: 'project-with-double-transitive-dependency-on-my-lib',
+          version: '0.0.1',
+          packageJsonPath:
+            'libs/project-with-double-transitive-dependency-on-my-lib/package.json',
+          localDependencies: [
+            {
+              // Depends on my-lib via project-with-transitive-dependency-on-my-lib
+              projectName: 'project-with-transitive-dependency-on-my-lib',
+              dependencyCollection: 'devDependencies',
+              version: '^0.0.1',
+            },
+          ],
+        },
       });
     });
     afterEach(() => {
@@ -1253,6 +1269,20 @@ Valid values are: "auto", "", "~", "^", "="`,
             "project-with-dependency-on-my-lib": "^0.0.1",
           },
           "name": "project-with-transitive-dependency-on-my-lib",
+          "version": "0.0.1",
+        }
+      `);
+      expect(
+        readJson(
+          tree,
+          'libs/project-with-double-transitive-dependency-on-my-lib/package.json'
+        )
+      ).toMatchInlineSnapshot(`
+        {
+          "devDependencies": {
+            "project-with-transitive-dependency-on-my-lib": "^0.0.1",
+          },
+          "name": "project-with-double-transitive-dependency-on-my-lib",
           "version": "0.0.1",
         }
       `);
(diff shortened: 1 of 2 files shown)


JamesHenry:
Hi @simon-abbott, I'm sorry for the delay on this one, as you might have seen it's down to the fact that the implementation of versioning has been totally rewritten to be more powerful and flexible. The legacy implementation can be considered frozen, and this should not be an issue on the new implementation.

The new one landed as opt-in in 20.8, you can set `release.version.useLegacyVersionsing: false` in your `nx.json` in that version and later, or in 21 it becomes the default, and there is an automated migration for existing workspaces, so you can also just wait until tomorrow when nx 21.0.0 should hopefully drop.

Cheers!

PR #30375, opened 2025-03-14
docs(angular): add modern angular testing with nx blog

Coly010:
## Related Issue(s)

Fixes #


jaysoo:
Looks like the internal link checker failed.

docs/blog/2025-03-17-modern-angular-testing-with-nx.md
```diff
@@ -0,0 +1,384 @@
+---
+title: 'Modern Angular Testing with Nx'
+slug: nx-made-cursor-smarter
+authors: ['Jack Hsu']
+tags: ['angular', 'nx']
+cover_image: /blog/images/articles/cursor-nx-mcp-heroimg.jpg
+description: 'Learn how Nx enhances Angular testing by integrating modern tools like Playwright and Vitest, improving test speed, reliability, and CI scalability.'
+---
+
+Testing is a crucial part of any application to ensure correctness and guard against regression, and Angular is no exception. Here at Nx, we're big fans of Angular, and we think that we can drastically improve the unit testing and end-to-end (E2E) experience for Angular developers.
```
juristr, line 10: Let's add a callout component at the beginning for the "Angular Week series" that links all blog posts so people can follow up easily and it also creates some anticipation/excitement about what's coming. 

For an example you can look here:

https://github.com/nrwl/nx/blob/398ab354e52e1d411954f2a36f346330d099df52/docs/blog/2025-01-29-new-nx-experience.md?plain=1#L11-L19

Obviously for the upcoming ones there will just be the blog post title (which we need to define 😬) for now and we'll link them as the articles get pushed up

Coly010: I'm going to need the actual names of the other blog posts

mhartington: We'll be updating this throughout the week, so no action needed right now

mhartington (approved):

PR #30064, opened 2025-02-16
feat(core): support bun's new text-based lockfile and use bun publish within nx release

JamesHenry:
## Current Behavior

- Bun's new default (as of `1.2`) text-based lockfile is not picked up by Nx
- We currently only support `pnpm` for dynamically replacing local dependency protocols at publish time.

## Expected Behavior

- Bun's text-based `bun.lock` is picked up by Nx

- We support both `pnpm` and `bun` for dynamically replacing local dependency protocols at publish time, and include more specific messaging for why modern `yarn` is not (hopefully "yet") supported.

Also resolves the package manager detection logic by folding in https://github.com/nrwl/nx/pull/28546

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/28201

FIxes https://github.com/nrwl/nx/discussions/29422

Fixes #29494

PR #30389, opened 2025-03-15
fix(nx/js): support's bun text lock file in nx release

patlux:
This PR fixes an error when running `nx release`:

## Current Behavior

```sh
 NX   Updating bun lock file


 NX   Staging changed files with git


 NX   fatal: pathspec 'bun.lockb' did not match any files
```

## Expected Behavior

```sh
 NX   Updating bun lock file


 NX   Staging changed files with git
```

## Related Issue(s)

https://github.com/nrwl/nx/pull/30064

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/lock-file.ts
+++ b/packages/nx/src/plugins/js/lock-file/lock-file.ts
@@ -171,7 +171,7 @@ export function getLockFileName(packageManager: PackageManager): string {
     return NPM_LOCK_FILE;
   }
   if (packageManager === 'bun') {
-    return BUN_LOCK_FILE;
+    return isBunTextLockFile() ? BUN_TEXT_LOCK_FILE : BUN_LOCK_FILE;
   }
   throw new Error(`Unknown package manager: ${packageManager}`);
 }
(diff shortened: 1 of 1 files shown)


packages/nx/src/plugins/js/lock-file/lock-file.ts
```diff
@@ -187,20 +187,21 @@ function getLockFilePath(packageManager: PackageManager): string {
...
   throw new Error(`Unknown package manager: ${packageManager}`);
 }
 
+function isBunTextLockFile() {
+  try {
+    const bunVersion = execSync('bun --version').toString().trim();
+    // In version 1.2.0, bun switched to a text based lockfile format by default
+    return gte(bunVersion, '1.2.0');
+  } catch {
+    return false;
```
patlux, line 201: I'm not quite sure which is the better default setting.

ezraripps: In versions above 1.2.0 you can still tell bun to use old format. Probably would make more sense to just look for either file

JamesHenry:
Sorry for the delay on this one @patlux, we have a more holistic fix here #31973 so closing in favour of that one

I have personally started to use bun more and really liking it, so we are working hard to make sure the support in Nx is as good as it can be

Issue #30403, opened 2025-03-17
Nx Release forces Prettier and has no way to disable, breaking biome builds


zackarychapple:
### Current Behavior

run `nx release`
tsconfig file gets modified. There is no way to make prettier not run as part of the nx release process. 
## before
```ts
"include": ["**/*.tsx", "**/*.ts"]
```
## after
```ts
  "include": [
    "**/*.tsx",
    "**/*.ts"
  ]
```

Line 1001 of the `release-version.ts` file has the following:
```ts
   /**
     * Ensure that formatting is applied so that version bump diffs are as minimal as possible
     * within the context of the user's workspace.
     */
    await formatFiles(tree);
```

### Expected Behavior

Prettier should be at a minimum able to skipped, and ideally not called at all since we are using biome.

### GitHub Repo

https://github.com/ZephyrCloudIO/zephyr-documentation/pull/98

### Steps to Reproduce

1. `pnpm run create-release` see tsconfig.json formatted and committed 
2. `pnpm run lint:fix` format fixed

### Nx Report

```shell
NX   Report complete - copy this into the issue template

Node           : 21.7.1
OS             : darwin-arm64
Native Target  : aarch64-macos
pnpm           : 10.6.3

nx             : 20.6.0
@nx/js         : 20.6.0
@nx/workspace  : 20.6.0
@nx/devkit     : 20.6.0
typescript     : 5.7.3
---------------------------------------
Registered Plugins:
@nx/js/typescript
---------------------------------------
Cache Usage: 0.00 B / 372.19 GB
```

### Failure Logs

```shell

```

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

pnpm v10.6.3

zackarychapple:
thanks!

PR #30404, opened 2025-03-17
fix(release): adding option to skip formatting in release

zackarychapple:
## Current Behavior
Cannot skip prettier on `nx release`

## Expected Behavior
Should be able to skip prettier for those of us using biome

## Related Issue(s)

Fixes #
https://github.com/nrwl/nx/issues/30403


JamesHenry:
Thanks for raising this @zackarychapple. This one isn't quite functionally correct, but in general I think the real issue here is that `formatFiles` is too naive - it is only checking if `prettier` is resolvable in your `node_modules`.

Therefore I've tested an alternate solution which requires no changes to release code: https://github.com/nrwl/nx/pull/30426

In the future we'll make `formatFiles` pluggable to more than just `prettier` as well, but this should fully unblock you for now

PR #30420, opened 2025-03-18
feat(nx-dev): LLM optimized rendering

juristr:
Implements the [llms.txt](https://llmstxt.org/) "standard" for the nx.dev docs


juristr:
will pick up later

PR #30459, opened 2025-03-22
fix(core): fix local registry not being considered when using `bun` p…

tinesoft:
…ackage manager

## Current Behavior

When using `bun` package manager,  the local registry url set when starting `verdaccio` is ignored by bun, which continues to point to default NPM registry, which in return causes an error when trying to publish package
```
> nx run common:nx-release-publish

bun publish error:
error: missing authentication (run `bunx npm login`)
```
## Expected Behavior

Verdaccio should work when using `bun` pm

## Related Issue(s)

Fixes #


JamesHenry (approved):
We obviously haven't been able to test this working on our side, but I trust you 😄

Issue #22945, opened 2024-04-22
NestJS reload/refresh not working in v18


poeck:
### Current Behavior

When I serve my NestJS project in development and change a file, it reloads perfectly the first time. When I change the file again, it stop refreshing and does nothing. This also happens in a fresh project with `npx create-nx-workspace my-workspace --preset=nest`

### Expected Behavior

It should refesh every time I change a file.

### Steps to Reproduce

1. Create a new nest project `npx create-nx-workspace my-workspace --preset=nest`
2. Run `nx serve api`
3. Change a file
4. See how it reloads
5. Change the file again
6. Nothing happens

### Nx Report

```shell
NX   Report complete - copy this into the issue template

Node   : 18.18.2
OS     : linux-x64
pnpm   : 8.15.2

nx (global)        : 18.3.3
nx                 : 18.3.3
@nx/js             : 18.3.3
@nx/eslint         : 18.3.3
@nx/workspace      : 18.3.3
@nx/eslint-plugin  : 18.3.3
@nx/express        : 18.3.3
@nx/nest           : 18.3.3
@nx/next           : 18.3.3
@nx/node           : 18.3.3
@nx/web            : 18.3.3
@nx/webpack        : 18.3.3
typescript         : 5.4.5
---------------------------------------
Registered Plugins:
@nx/webpack/plugin
@nx/eslint/plugin
@nx/next/plugin
```

### Failure Logs

```shell
Daemon Log:

[NX Daemon Server] - 2024-04-22T19:14:18.253Z - Started listening on: /tmp/9e9a81d0946c1ff77af8/d.sock
[NX Daemon Server] - 2024-04-22T19:14:18.254Z - [WATCHER]: Subscribed to changes within: /home/paul/dev/jokedial (native)
[NX Daemon Server] - 2024-04-22T19:14:18.255Z - Established a connection. Number of open connections: 1
[NX Daemon Server] - 2024-04-22T19:14:18.255Z - Established a connection. Number of open connections: 2
[NX Daemon Server] - 2024-04-22T19:14:18.256Z - [REQUEST]: Client Request for Project Graph Received
[NX Daemon Server] - 2024-04-22T19:14:18.313Z - [REQUEST]: Updated workspace context based on watched changes, recomputing project graph...
[NX Daemon Server] - 2024-04-22T19:14:18.313Z - [REQUEST]: 
[NX Daemon Server] - 2024-04-22T19:14:18.313Z - [REQUEST]: 
[...]

Issue #21734, opened 2024-02-08
Project disappears from the graph after editing files with Vim or Neovim


JakeDern:
### Current Behavior

Every time I edit a `project.json` file in my repo it disappears from the project graph and I cannot execute any more targets against it. Running `nx reset` restores the project.

Here's an example:

<img width="517" alt="image" src="https://github.com/nrwl/nx/assets/33842784/0adcb2c1-df91-43c8-943f-f2ec7140108a">

### Expected Behavior

I would expect the daemon to process the update.

### GitHub Repo

https://github.com/JakeDern/nx-test

### Steps to Reproduce

1. Run an nx target for any project in the repo
2. Edit the project.json for the target. Even if it's just removing and adding the same text.
3. Try to run the target again.

### Nx Report

```shell
>  NX   Report complete - copy this into the issue template

   Node   : 20.11.0
   OS     : linux-x64
   npm    : 10.2.4

   nx (global)    : 18.0.2
   nx             : 18.0.3
   @nx/js         : 18.0.3
   @nx/workspace  : 18.0.3
   @nx/devkit     : 18.0.3
   @nrwl/tao      : 18.0.3
```

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I suspect this is environment specific as it's such an obvious bug. I can reproduce this in a fresh repo on my machine, which is what I've linked and is what the screenshots are from. Any tips to debug would be greatly appreciated.

JakeDern:
Took the time this morning to try and reproduce this on my personal desktop and the same thing happens. 

I'm using wsl2 environment with ubuntu 20.04 in both cases.  I hadn't used nx before on this machine at all, so I simply did
`npm install -g nx@latest` , `npm install`, and then did the same test as above.

JakeDern:
Here's some basic version information from the second environment: 
![image](https://github.com/nrwl/nx/assets/33842784/d08cd77d-d5da-4f22-baaf-7b0a8b3a3c53)

JakeDern:
As another data point, I cloned/installed on windows and cannot reproduce the issue. Perhaps is something with wsl2 environment specifically:

In between each of these commands I made miscellaneous project.json edits:

![image](https://github.com/nrwl/nx/assets/33842784/d33c261b-77bf-4295-a271-1e5c5768f8e8)

JakeDern:
[...]

PR #30472, opened 2025-03-23
docs(misc): document vim/nvim nx watch issue

Cryt1c:
If you are using Vim/Nvim and you are saving a file this file is not edited directly. The content is saved to a backup file and after this step the original file gets deleted and the backup file gets renamed. The nx watch command can't handle this at the moment. To help users of this command find this information I propose this documentation change.

## Related Issue(s)

General information and the workaround:
https://github.com/nrwl/nx/issues/22945
Investigation:
https://github.com/nrwl/nx/issues/21734#issuecomment-1962274644

Fixes #

### DIFF
--- a/docs/generated/packages/nx/documents/watch.md
+++ b/docs/generated/packages/nx/documents/watch.md
@@ -5,7 +5,7 @@ description: 'Watch for changes within projects, and execute commands.'
 
 # watch
 
-Watch for changes within projects, and execute commands.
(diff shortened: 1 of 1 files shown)


FrozenPandaz:
Instead of documenting this I'm hoping we can fix this. @AgentEnder can you take a look please?

jaysoo:
This should be addressed in the file watcher already, I don't think we need to call it out in docs.

PR #30500, opened 2025-03-25
fix(nextjs): update documentation for buildable libraries and bundler configuration

ndcunningham:
## Current Behavior

Currently, the docs does not differentiate between inferred target outputs and executor outputs.

## Expected Behavior

Since the outputs differ we should highlight the difference in our docs.
Also, if we pass a bundler to the Next.js lib app generator it should respect that option.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/packages/next/plugin-overview.md
+++ b/docs/shared/packages/next/plugin-overview.md
@@ -231,13 +231,53 @@ Next.js applications can be build with:
 nx build my-new-app
 ```
 
-And if you generated a library with --buildable, then you can build a library as well:
+And if you generated a library with `--bundler`, then you can build a library as well:
 
 ```shell
 nx build my-new-lib
 ```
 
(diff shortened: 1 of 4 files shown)


docs/shared/packages/next/plugin-overview.md
```diff
@@ -231,13 +231,53 @@ Next.js applications can be build with:
...
 ```
 
-After running a build, the output will be in the `dist` folder. You can customize the output folder by setting `outputPath` in the project's `project.json` file.
+After running a build, the output will be in the `{workspaceRoot}/dist/{projectRoot}` folder.
+
+{% tabs %}
+
+{% tab label="Using inferred tasks" %}
+
+You can customize the output folder path by update the bundler's config. For example vite's config can be updated in `vite.config.ts`:
```
jaysoo, line 246: I would also mention using `@nx/next/plugin` in `nx.json` and link to https://nx.dev/concepts/inferred-tasks#inferred-tasks-project-crystal page.

jaysoo (approved):

PR #30527, opened 2025-03-27
docs(react): update react lib generator

MaxKless:
the examples are outdated, the new ones make more sense

### DIFF
--- a/packages/react/src/generators/library/schema.json
+++ b/packages/react/src/generators/library/schema.json
@@ -7,12 +7,12 @@
   "type": "object",
   "examples": [
     {
-      "command": "nx g lib mylib --directory=myapp",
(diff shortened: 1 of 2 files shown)


packages/react/src/generators/library/schema.json
```diff
@@ -7,12 +7,12 @@
   "type": "object",
   "examples": [
     {
-      "command": "nx g lib mylib --directory=myapp",
-      "description": "Generate `libs/myapp/mylib`"
+      "command": "nx g @nx/react:lib mylib",
```
Coly010, line 10:
```suggestion
      "command": "nx g @nx/react:lib libs/mylib",
```

Coly010 (changes requested):

jaysoo (approved):

Issue #30410, opened 2025-03-18
@nx/nest; Error: Cannot find module when importing a library generated with @nx/nest:library


WoodenMaiden:
### Current Behavior

Whenever I import a nest library created with ``@nx/nest:library`` in an app, it cannot resolve modules

### Expected Behavior

A successful module resolution 

### Steps to Reproduce

1. Create a nest app
2. Create a nest library: ``nx generate @nx/nest:library --name=workspace-models --directory=libs/workspace-models --no-interactive``
3. run ``nx serve myapp``, it works
4. Import the module into the app```ts
import { Module } from '@nestjs/common';
//...
import { WorkspaceModelsModule } from '@fair-space-services/workspace-models'; 

@Module({
  imports: [
    WorkspaceModelsModule
  ],
})
export class AppModule {}

```
4. ``nx sync && nx serve myapp``, the error occurs



### Nx Report

```shell
NX   Report complete - copy this into the issue template

Node           : 23.5.0
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.9.2

nx (global)        : 20.5.0
nx                 : 20.6.0
@nx/js             : 20.6.0
@nx/jest           : 20.6.0
@nx/eslint         : 20.6.0
@nx/workspace      : 20.6.0
@nx/devkit         : 20.6.0
@nx/eslint-plugin  : 20.6.0
@nx/nest           : 20.6.0
@nx/node           : 20.6.0
@nx/web            : 20.6.0
@nx/webpack        : 20.6.0
typescript         : 5.7.3
---------------------------------------
Registered Plugins:
@nx/js/typescript
@nx-go/nx-go
@nx/webpack/plugin
@nx/eslint/plugin
@nx/jest/plugin
@nx/js/typescript
---------------------------------------
Community plugins:
@nx-go/nx-go : 3.3.1
---------------------------------------
Cache Usage: 1.51 MB / 39.06 GB
```

### Failure Logs

```shell
Nx read the output from the cache instead of running the command for 1 out of 1 tasks.

Debugger listening on ws://localhost:9229/fa1b2107-52df-40cb-9588-d31de60b3db3
Debugger listening on ws://localhost:9229/fa1b2107-52df-40cb-9588-d31de60b3db3
For help, see: https://nodejs.org/en/docs/inspector

[...]

PR #30538, opened 2025-03-28
fix(webpack): nestjs workspace libs referencing when using ts solution

### LINKED ISSUE #30492: Webpack build does not add local libraries to bundled js file
### Current Behavior

When using  the workspace mode and typscript project references which is the current default, webpack does not understand local libraries and includes them as external dependency, this is not ideal cos said local libraries are not going to be published anywhere, it is simple a piece of logic shared across apps that I want to be included in the bundled js file. I canont seem to find a way to alter this behaviour as adding the path to the tsconfig of the webpack app does not seem to do anything.

### Expected Behavior

When using nx in integrated mode, a webpack app is able to understand that the local libraries should not be included as external dependencies and is able to bundle any local dependency imported probably due to being specified in the tsconfig paths options.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create an nx repo using this command npx create-nx-workspace test
2. Add @nx/nest with npx nx add @nx/nest
3. Create a Nest.js app npx nx g @nx/nest:application packages/backend
4. Create a Nest.js library npx nx g @nx/nest:library libs/env
5. import module from libs/env into packages/backend
6. Run build on backend npx nx build backend
7. Check output in apps/backend/dist/main.js
8. @test/env is listed as an external dependency

### Nx Report

```shell

Node           : 20.15.0
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.7.0

nx             : 20.6.3
@nx/js         : 20.6.3
@nx/jest       : 20.6.3
@nx/eslint     : 20.6.3
@nx/workspace  : 20.6.3
@nx/devkit     : 20.6.3
@nx/nest       : 20.6.3
@nx/node       : 20.6.3
@nx/web        : 20.6.3
@nx/webpack    : 20.6.3
typescript     : 5.7.3
Registered Plugins:
@nx/js/typescript
@nx/webpack/plugin
@nx/js/typescript
Cache Usage: 109.63 KB / 100.69 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

ndcunningham:
## Current Behavior

Currently, when you have a Nest app that imports a Nest Lib that is not buildable inside a TS solution workspace the serving that application fails because the library is marked as external. 

Before we would include these projects as a part of `tsconfig.compilerOptions.paths`. However, this is no longer possible as it would not be a valid TS solution setup.

## Expected Behavior

If a library is not buildable it should be able to be resolved regardless if we are using TS solutions or the legacy paths.

## Related Issue(s)

closes: #30492, #30410, #30544

### DIFF
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
@@ -345,7 +345,35 @@ function applyNxDependentConfig(
   const externals = [];
   if (options.target === 'node' && options.externalDependencies === 'all') {
     const modulesDir = `${options.root}/node_modules`;
-    externals.push(nodeExternals({ modulesDir }));
+
+    const graph = options.projectGraph;
+    const projectName = options.projectName;
+
+    const deps = graph?.dependencies?.[projectName] ?? [];
+
+    // Collect non-buildable TS project references so that they are bundled
+    // in the final output. This is needed for projects that are not buildable
+    // but are referenced by buildable projects. This is needed for the new TS
+    // solution setup.
+    const nonBuildableWorkspaceLibs = isUsingTsSolution
+      ? deps
+          .filter((dep) => {
+            const node = graph.nodes?.[dep.target];
+            if (!node || node.type !== 'lib') return false;
+
+            const hasBuildTarget = 'build' in (node.data?.targets ?? {});
+
+            return !hasBuildTarget;
+          })
+          .map(
+            (dep) => graph.nodes?.[dep.target]?.data?.metadata?.js?.packageName
+          )
+          .filter((name): name is string => !!name)
+      : [];
+
+    externals.push(
+      nodeExternals({ modulesDir, allowlist: nonBuildableWorkspaceLibs })
+    );
   } else if (Array.isArray(options.externalDependencies)) {
     externals.push(function (ctx, callback: Function) {
       if (options.externalDependencies.includes(ctx.request)) {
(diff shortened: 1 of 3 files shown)


packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
```diff
@@ -345,7 +345,35 @@ function applyNxDependentConfig(
...
+    // in the final output. This is needed for projects that are not buildable
+    // but are referenced by buildable projects. This is needed for the new TS
+    // solution setup.
+    const nonBuildableWorkspaceLibs = isUsingTsSolution
+      ? deps
+          .filter((dep) => {
+            const node = graph.nodes?.[dep.target];
+            if (!node || node.type !== 'lib') return false;
+
+            const hasBuildTarget = 'build' in (node.data?.targets ?? {});
```
leosvelperez, line 364: Checking for `build` would probably cover most of the cases, and we should keep it as a first/fast check, but the task could also have a different name. If there's no `build` task, can we additionally check the packages' entry points to determine whether they are pointing to the source or the outputs?

FWIW, the graph nodes can contain a `metadata.js` object where the entry points are already collected, so there wouldn't be a need to read the `package.json` files.

ndcunningham: Fair point I can update to also check entry points if no build target.

leosvelperez (changes requested):

jaysoo (approved):

leosvelperez (approved):

PR #30543, opened 2025-03-30
feat(core): run parallel based on the number of cpu cores

ianzone:
BREAKING CHANGE: default parallel changed from 3 to half of the cpu cores

## Current Behavior

can not specify parallel based on the number of cpu cores. 

## Expected Behavior

`--parallel 0.5` means half of the cpu cores. 
`--parallel 3` means exact tasks to run in parallel. 

## Related Issue(s)

https://github.com/nrwl/nx/discussions/30238

### DIFF
--- a/packages/nx/src/command-line/yargs-utils/shared-options.spec.ts
+++ b/packages/nx/src/command-line/yargs-utils/shared-options.spec.ts
@@ -1,6 +1,6 @@
 import * as yargs from 'yargs';
 
-import { withAffectedOptions, withRunManyOptions } from './shared-options';
+import { withAffectedOptions, withRunManyOptions, readParallelFromArgsAndEnv } from './shared-options';
 
 const argv = yargs.default([]);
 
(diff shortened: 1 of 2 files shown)


MaxKless:
Hey @ianzone thanks for the PR!
Some points: 
- instead of a fractional number, I think it would make more sense to go with a percentage, like jest & vitest do it. It makes sense to align with other tools in order to be more intuitive. Then it's also clear that number specifies the cores directly while a string is percentage-based
- Is there a reason that the default changed from 3 to `50%`? I think it would be better to leave the default alone for now as it will affect everyone. 

Here are some links to jest & vitest:
- [vitest/packages/vitest/src/utils/workers.ts at a9d36c719f8ce5551f61da20181490d3673bdf99 · vitest-dev/vitest](https://github.com/vitest-dev/vitest/blob/a9d36c719f8ce5551f61da20181490d3673bdf99/packages/vitest/src/utils/workers.ts)
- https://jestjs.io/docs/cli#--maxworkersnumstring

MaxKless:
sure, I get it. But it's still a guess that will affect everyone and if the price is a couple of extra characters.... let's stick with the same default for now

ianzone:
> sure, I get it. But it's still a guess that will affect everyone and if the price is a couple of extra characters.... let's stick with the same default for now

ok

ianzone:
updated

MaxKless:
Looks good! Can you add a test to `shared-options.spec.ts` to verify this is and will keep behaving as it should? :) Thanks!

ianzone:
tests added

ianzone:
Hi, could you merge this?

MaxKless (approved):
thanks for the improvement! Sorry, I was on vacation in April and forgot to hand this over. Thanks for your patience

MaxKless:
It looks like there are still some formatting errors... could you run `nx format --fix` and push the changes so CI will be green?

ianzone:
I was developing on master, and it's tricky to update now. I'll do another PR with a new branch

MaxKless:
yeah I tried just doing the update and pushing it directly too, but it didn't work :P looking forward to seeing your new branch :)

ianzone:
> yeah I tried just doing the update and pushing it directly too, but it didn't work :P looking forward to seeing your new branch :)

it's here https://github.com/nrwl/nx/pull/31011

PR #30546, opened 2025-03-31
feat(core): handle yarn resolutions and patches when parsing and pruning lock file

meeroslav:
This PR introduces support for `resolutions` and `patches` for yarn lock file parsing and pruning.

Additionally, it fixes the correct dependency matching when the version has one of the two forms:
- fluid version range (e.g. `npm:*`)
- union version range (e.g. `1 || 2 || 3`)

## Related Issue(s)

Fixes #


packages/nx/src/plugins/js/lock-file/yarn-parser.ts
```diff
@@ -306,9 +306,30 @@ function getDependencies(
           (section) => {
             if (section) {
               Object.entries(section).forEach(([name, versionRange]) => {
-                const target =
+                let target =
                   keyMap.get(`${name}@npm:${versionRange}`) ||
                   keyMap.get(`${name}@${versionRange}`);
+                if (!target) {
+                  const shortRange = versionRange.replace(/^npm:/, '');
```
meeroslav, line 313: This block handles the `*` and `||` version range parsing

packages/nx/src/plugins/js/lock-file/yarn-parser.ts
```diff
@@ -400,13 +421,13 @@ function addPackageVersion(
     collection.set(packageName, new Set());
   }
   collection.get(packageName).add(`${packageName}@${version}`);
-  if (isBerry && !version.startsWith('npm:')) {
+  if (isBerry && !version.startsWith('npm:') && !version.startsWith('patch:')) {
```
meeroslav, line 424: Ensure we don't end up with `packageName@npm:patch:...` case

packages/nx/src/plugins/js/lock-file/yarn-parser.ts
```diff
@@ -461,10 +485,32 @@ function mapSnapshots(
         snapshotMap.get(snapshot).add(requestedKey);
       }
     }
+    const requestedResolutionsVersion = getPackageJsonVersion(
```
meeroslav, line 488: Version might not match the `dependencies` block but it could match the override in `resolutions`

packages/nx/src/plugins/js/lock-file/yarn-parser.ts
```diff
@@ -528,21 +574,41 @@ function reverseMapBerryKey(
...
+    const versionRange =
+      dependencies[packageName].match(patchRegex)?.[1] ||
+      dependencies[packageName];
+    if (versionRange === version || satisfies(version, versionRange)) {
+      return dependencies[packageName];
     }
   }
 }
 
+function isStandardPackage(snapshot: YarnDependency, version: string): boolean {
```
meeroslav, line 594: These have been extracted to functions for greater readability

jaysoo (approved):

PR #30549, opened 2025-03-31
fix(js): infer dependency between `typecheck` and `build` tasks and more granular outputs for `typecheck`

leosvelperez:
## Current Behavior

There is no dependency between the inferred `typecheck` and `build` tasks. Depending on their run order, this can result in duplicated processing (type-checking, `.d.ts` generation). Given there's no explicit dependency, the order would be non-deterministic.

Additionally, when `outDir` is set in the tsconfig files, it's used as-is in the currently inferred outputs for `typecheck`. This can result in extra files being cached for the task.

## Expected Behavior

For optimum performance, the inferred `typecheck` task should depend on the `build` task. The `typecheck` task's outputs should be more granular so that only the relevant files (declaration files and declaration map files if enabled) are cached.

### Explanation

Consider a typical setup with specific tsconfig file for files with different concerns:

- tsconfig.lib.json: TS configuration for the library runtime files
- tsconfig.spec.json: TS configuration for the unit test files
- tsconfig.json: TS solution configuration, a solution file that references the specific config files above

When running `tsc -b tsconfig.lib.json --verbose` (build), we can see how the `tsconfig.lib.json` TS project is built:

```bash
Projects in this build:
    * tsconfig.lib.json

Project 'tsconfig.lib.json' is out of date because output file 'dist/tsconfig.lib.tsbuildinfo' does not exist

Building project '<workspace root>/packages/pkg1/tsconfig.lib.json'...
```

After that, if we run `tsc -b tsconfig.json --emitDeclarationOnly --verbose` (typecheck), we'll see how the `tsc` output for `tsconfig.lib.json` is reused:

```bash
Projects in this build: 
    * tsconfig.lib.json
    * tsconfig.spec.json
    * tsconfig.json

Project 'tsconfig.lib.json' is up to date because newest input 'src/lib/file.ts' is older than output 'dist/tsconfig.lib.tsbuildinfo'

Project 'tsconfig.spec.json' is out of date because output file 'out-tsc/jest/tsconfig.spec.tsbuildinfo' does not exist

Building project '<workspace root>/packages/pkg1/tsconfig.spec.json'...
```

The relevant bit above is `Project 'tsconfig.lib.json' is up to date because newest input 'src/lib/file.ts' is older than output 'dist/tsconfig.lib.tsbuildinfo'`. Because the initial `build` task already typechecks and produces `.d.ts` files for the `tsconfig.lib.json`, when the `typecheck` task runs, `tsc` identifies that the outputs for that config files were already produced and can be reused.

If we were to run the tasks in the inverse order, the results would be different:

```bash
> npx tsc -b tsconfig.json --emitDeclarationOnly --verbose
Projects in this build: 
    * tsconfig.lib.json
    * tsconfig.spec.json
    * tsconfig.json

Project 'tsconfig.lib.json' is out of date because output file 'dist/tsconfig.lib.tsbuildinfo' does not exist

Building project '<workspace root>/packages/pkg1/tsconfig.lib.json'...

Project 'tsconfig.spec.json' is out of date because output file 'out-tsc/jest/tsconfig.spec.tsbuildinfo' does not exist

Building project '<workspace root>/packages/pkg1/tsconfig.spec.json'...

> npx tsc -b tsconfig.lib.json --verbose
Projects in this build: 
    * tsconfig.lib.json

Project 'tsconfig.lib.json' is out of date because buildinfo file 'dist/tsconfig.lib.tsbuildinfo' indicates there is change in compilerOptions

Building project '<workspace root>/packages/pkg1/tsconfig.lib.json'...
```

Note how when the `build` task is run, `tsc` identifies that there was a change in `compilerOptions` (`--emitDeclarationOnly`) and it requires building the project. This is because the `typecheck` task only generates declaration files and the `build` task must also emit the transpiled `.js` files.

### Benchmark

Running those two different flows in a simple (non-Nx) project with a TS configuration structure like the one mentioned above and with 5000 TS files split in half for runtime and test files yields the following:

```bash
hyperfine -r 5 -p "rm -rf dist out-tsc" \
-n "build => typecheck" "npx tsc -b tsconfig.lib.json && npx tsc -b --emitDeclarationOnly" \
-n "typecheck => build" "npx tsc -b tsconfig.json --emitDeclarationOnly && npx tsc -b tsconfig.lib.json"
Benchmark 1: build => typecheck
  Time (mean ± σ):      6.832 s ±  0.094 s    [User: 11.361 s, System: 1.060 s]
  Range (min … max):    6.734 s …  6.985 s    5 runs
 
Benchmark 2: typecheck => build
  Time (mean ± σ):      8.789 s ±  0.015 s    [User: 14.817 s, System: 1.267 s]
  Range (min … max):    8.771 s …  8.812 s    5 runs
 
Summary
  build => typecheck ran
    1.29 ± 0.02 times faster than typecheck => build
```

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/js/src/js-ts-solution.test.ts
+++ b/e2e/js/src/js-ts-solution.test.ts
@@ -115,36 +115,36 @@ ${content}`
 
     // check build
     expect(runCLI(`build ${esbuildParentLib}`)).toContain(
-      `Successfully ran target build for project @proj/${esbuildParentLib} and 5 tasks it depends on`
+      `Successfully ran target build for project @proj/${esbuildParentLib}`
     );
     expect(runCLI(`build ${rollupParentLib}`)).toContain(
-      `Successfully ran target build for project @proj/${rollupParentLib} and 5 tasks it depends on`
+      `Successfully ran target build for project @proj/${rollupParentLib}`
     );
     expect(runCLI(`build ${swcParentLib}`)).toContain(
-      `Successfully ran target build for project @proj/${swcParentLib} and 5 tasks it depends on`
+      `Successfully ran target build for project @proj/${swcParentLib}`
     );
     expect(runCLI(`build ${tscParentLib}`)).toContain(
-      `Successfully ran target build for project @proj/${tscParentLib} and 5 tasks it depends on`
+      `Successfully ran target build for project @proj/${tscParentLib}`
     );
     expect(runCLI(`build ${viteParentLib}`)).toContain(
-      `Successfully ran target build for project @proj/${viteParentLib} and 5 tasks it depends on`
+      `Successfully ran target build for project @proj/${viteParentLib}`
     );
 
     // check typecheck
     expect(runCLI(`typecheck ${esbuildParentLib}`)).toContain(
-      `Successfully ran target typecheck for project @proj/${esbuildParentLib} and 5 tasks it depends on`
+      `Successfully ran target typecheck for project @proj/${esbuildParentLib}`
     );
     expect(runCLI(`typecheck ${rollupParentLib}`)).toContain(
-      `Successfully ran target typecheck for project @proj/${rollupParentLib} and 5 tasks it depends on`
+      `Successfully ran target typecheck for project @proj/${rollupParentLib}`
     );
     expect(runCLI(`typecheck ${swcParentLib}`)).toContain(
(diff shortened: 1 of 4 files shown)


jaysoo (approved):
LGTM

PR #30558, opened 2025-04-01
fix(rspack): allow typechecking during serve #29267

### LINKED ISSUE #29267: nx:rspack does not respect the skipTypeChecking options
### Current Behavior

when I try to pass skipTypeChecking as false or typeCheck as false (even though I have impression of typeCheck as false has different meaning than skipTypeChecking as false ), the type check never been performed.

### Expected Behavior

when passing skipTypeChecking as false or typeCheck as false, type check should be performed

### GitHub Repo

_No response_

### Steps to Reproduce

1. create project with rspack executor
2. set the skipTypeChecking as false or typeCheck as true
3. run the rspack executor

### Nx Report

```shell
Node           : 18.18.2
OS             : darwin-arm64
npm            : 9.8.1

nx                 : 20.1.4
@nx/js             : 20.1.4
@nx/jest           : 20.1.4
@nx/eslint         : 20.1.4
@nx/workspace      : 20.1.4
@nx/devkit         : 20.1.4
@nx/eslint-plugin  : 20.1.4
@nx/nest           : 20.1.4
@nx/node           : 20.1.4
@nx/plugin         : 20.1.4
@nx/rspack         : 20.1.4
@nx/web            : 20.1.4
typescript         : 5.4.5
Local workspace plugins:
         generator
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

not sure if this is because of this but seems https://github.com/nrwl/nx/blob/20.1.4/packages/rspack/src/executors/rspack/rspack.impl.ts#L22 the check is using alias option instead of the actual option, should it be using the same check as the js/swc https://github.com/nrwl/nx/blob/master/packages/js/src/utils/swc/compile-swc.ts#L93 ?

Coly010:
## Current Behavior
The `NxRspackAppPlugin` and `NxWebpackAppPlugin` will not provide typechecking if using TS Solution setup because it already has a typecheck target.
However, this means that typechecking will not be run during serve.

## Expected Behavior
Allow typechecking to run during serve even with TS Solution setup

## Related Issue(s)

Fixes #29267

### DIFF
--- a/packages/rspack/src/plugins/utils/apply-base-config.ts
+++ b/packages/rspack/src/plugins/utils/apply-base-config.ts
@@ -240,8 +240,13 @@ function applyNxDependentConfig(
     plugins.push(new NxTsconfigPathsRspackPlugin({ ...options, tsConfig }));
   }
 
-  // New TS Solution already has a typecheck target
-  if (!options?.skipTypeChecking && !isUsingTsSolution) {
+  // New TS Solution already has a typecheck target but allow it to run during serve
+  if (
+    (!options?.skipTypeChecking && !isUsingTsSolution) ||
+    (isUsingTsSolution &&
+      options?.skipTypeChecking === false &&
+      process.env['WEBPACK_SERVE'])
+  ) {
     const { TsCheckerRspackPlugin } = require('ts-checker-rspack-plugin');
     plugins.push(
       new TsCheckerRspackPlugin({
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
@@ -243,8 +243,13 @@ function applyNxDependentConfig(
     plugins.push(new NxTsconfigPathsWebpackPlugin({ ...options, tsConfig }));
   }
 
-  // New TS Solution already has a typecheck target
-  if (!options?.skipTypeChecking && !isUsingTsSolution) {
+  // New TS Solution already has a typecheck target but allow it to run during serve
+  if (
+    (!options?.skipTypeChecking && !isUsingTsSolution) ||
+    (isUsingTsSolution &&
+      options?.skipTypeChecking === false &&
+      process.env['WEBPACK_SERVE'])
+  ) {
     const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
     plugins.push(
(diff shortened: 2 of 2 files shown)


packages/rspack/src/plugins/utils/apply-base-config.ts
```diff
@@ -240,8 +240,11 @@ function applyNxDependentConfig(
     plugins.push(new NxTsconfigPathsRspackPlugin({ ...options, tsConfig }));
   }
 
-  // New TS Solution already has a typecheck target
-  if (!options?.skipTypeChecking && !isUsingTsSolution) {
+  // New TS Solution already has a typecheck target but allow it to run during serve
+  if (
+    !options?.skipTypeChecking &&
+    (!isUsingTsSolution || process.env['WEBPACK_SERVE'])
+  ) {
```
leosvelperez, line 247: The default value for `skipTypeChecking` is `false` but for the TS solution setup we should consider it to be `true`. Let's explicitly check to `skipTypeChecking === false`:
```suggestion
  if (
    (!options?.skipTypeChecking && !isUsingTsSolution) ||
    (isUsingTsSolution &&
      options?.skipTypeChecking === false &&
      process.env['WEBPACK_SERVE'])
  ) {
```

packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-base-config.ts
```diff
@@ -243,8 +243,11 @@ function applyNxDependentConfig(
     plugins.push(new NxTsconfigPathsWebpackPlugin({ ...options, tsConfig }));
   }
 
-  // New TS Solution already has a typecheck target
-  if (!options?.skipTypeChecking && !isUsingTsSolution) {
+  // New TS Solution already has a typecheck target but allow it to run during serve
+  if (
+    !options?.skipTypeChecking &&
+    (!isUsingTsSolution || process.env['WEBPACK_SERVE'])
+  ) {
```
leosvelperez, line 250: Same here:
```suggestion
  if (
    (!options?.skipTypeChecking && !isUsingTsSolution) ||
    (isUsingTsSolution &&
      options?.skipTypeChecking === false &&
      process.env['WEBPACK_SERVE'])
  ) {
```

leosvelperez (approved):

PR #30567, opened 2025-04-01
fix(bundling): do not normalize tsconfig path for Windows with rollup

ghost:
## Current Behavior
Rollup build fails on Windows with errors like this:
```
[plugin rpt2] error TS6059: File '/foo/bar/baz.ts' is not under 'rootDir' 'C:/foo/bar'. 'rootDir' is expected to contain all source files.
```

This is because since https://github.com/nrwl/nx/commit/81fe7bb278826ece0bec9ff794fa5b1ef5c0c295 the `tsconfig` path passed to `rollup-plugin-typescript2` is built using `joinPathFragments`, which removes the drive letter on Windows.

## Expected Behavior
Rollup build should not fail.


ms-tng:
The failing check is not related to my change:
image

Coly010 (approved):

PR #30573, opened 2025-04-01
fix(bundling): webpack and rspack builds respect output.clean config option

jaysoo:
This PR fixes and issue where the standard `output.clean` option is ignored and replaced by the Nx-specific `deleteOutputPath` option on the `NxAppWebpackPlugin` and `NxAppRspackPlugin` plugins.

We want to allow users to use standards over our own features, so if we see that `output.clean` is set in webpack/rspack config, then we use that value.

For example, an Rspack config could be:

```js
const { NxAppRspackPlugin } = require("@nx/rspack/app-plugin");
const { join } = require("path");

module.exports = {
  output: {
    path: join(__dirname, "dist/demo"),
    clean: false, // <-- THIS DOES NOT WORK!
  },
  plugins: [
    new NxAppRspackPlugin({
      // ...
    }),
  ],
};
```

But even though `output.clean` is `false`, each build will still delete `dist`. The only way to disable that behavior is to use the Nx-specific option like this:

```js
const { NxAppRspackPlugin } = require("@nx/rspack/app-plugin");
const { join } = require("path");

module.exports = {
  output: {
    path: join(__dirname, "dist/demo"),
  },
  plugins: [
    new NxAppRspackPlugin({
      deleteOutputPath: false,
      // ...
    }),
  ],
};
```

## Current Behavior

Setting `output.clean` in Webpack/Rspack config does nothing, and we always default our own `deleteOutputPath` to `true`.

## Expected Behavior
Setting `output.clean` standard option is respected.

## Related Issue(s)

Fixes #

### DIFF
--- a/docs/shared/packages/webpack/webpack-plugins.md
+++ b/docs/shared/packages/webpack/webpack-plugins.md
@@ -95,6 +95,10 @@ Type: `boolean`
 
 Delete the output path before building.
 
+**`Deprecated`**
+
+Use [`output.clean`](https://webpack.js.org/guides/output-management/#cleaning-up-the-dist-folder) instead.
+
 ##### deployUrl
 
 Type: `string`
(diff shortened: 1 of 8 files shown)


ndcunningham (commented):
should we deprecate this?

ndcunningham (approved):

PR #30580, opened 2025-04-02
feat(webpack): set default sassImplementation to sass-embedded

Coly010:
## Current Behavior
The `sassImplementation` option for Webpack and Rspack is set to `sass` by default.
`sass-embedded` offers a faster alternative.

## Expected Behavior
Switch the default `sassImplementation` option to `sass-embedded`.


leosvelperez (commented):
We might want to call this out as a breaking change. Even though they are comparable functionality-wise, `sass-embedded` only works in selected platforms (the most used ones, but it could still break some setups).

leosvelperez (approved):

PR #30599, opened 2025-04-02
fix(remix): update setup-tailwind generator to correct import statements

ndcunningham:
## Current Behavior

There are two issues when we generate a remix application with ts solutions:

1. Running the `setup-tailwind` generator with a reference project which now is commonly with the pattern `@nx/acme` throws an error based on the schema.
```
Property 'project' does not match the schema.'@nx/acme' should match the pattern '^[a-zA-Z].*$'.
```
2. The `tailwind.css` inside of `root.tsx` causes error with Vite.
```
app/root.tsx (9:7): "default" is not exported by "app/tailwind.css", imported by "app/root.tsx".
```

## Expected Behavior

The remix project should be able to be referenced similarly to how we handle React.
The tailwind import should not throw an error.

## Related Issue(s)

Fixes #30343


isaacplmann (approved):

Hardanish-Singh (approved):

packages/remix/src/generators/setup-tailwind/__snapshots__/setup-tailwind.impl.spec.ts.snap
```diff
@@ -43,7 +43,7 @@ exports[`setup-tailwind generator should add a tailwind config to an application
   ScrollRestoration,
 } from '@remix-run/react';
 import type { MetaFunction, LinksFunction } from '@remix-run/node';
-import twStyles from './tailwind.css';
+import from './tailwind.css';
```
jaysoo, line 46: This should be just `import './tailwind.css'` if it's just a global CSS file.

Coly010 (approved):

PR #30604, opened 2025-04-03
feat: Liberate Remote Cache

Exelord:
## Current Behavior
NX allows only specific NX packages to provide RemoteCacheV2 API, forcing users to agree on licensing or (in the future) implement own Cache Service.

## Expected Behavior

Users can specify package of their choice that provides RemoteCacheV2 API without forcing them to implementing Cache Service or agreeing on licensing.

## Motivation

https://nx.dev/blog/custom-runners-and-self-hosted-caching

> Our commitment is that no major open-source feature or API will be replaced with a non-open-source alternative. Some features will inevitably become obsolete over time, but when that happens, we will provide clear deprecation notices and migration strategies. Nx will continue to evolve, but removing an old API will never be a means to move users toward a paid feature.

> We will keep developing Nx Cloud and other premium offerings for teams that need enterprise-scale solutions, but we remain focused on a strong open-source foundation that allows teams to use and scale Nx without requiring paid features.

> Nx open-source remains a primary focus for us as a company.

## Related Issue(s)

Fixes: https://github.com/nrwl/nx/discussions/30548

### DIFF
--- a/packages/nx/src/config/nx-json.ts
+++ b/packages/nx/src/config/nx-json.ts
@@ -416,6 +416,15 @@ export interface NxJsonConfiguration<T = '*' | string[]> {
     libsDir?: string;
     appsDir?: string;
   };
+
+  /**
+   * Remote Cache configuration
+   */
+  remoteCache?: {
+    provider?: string;
+    options?: any;
+  };
+
   /**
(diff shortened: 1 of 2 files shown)


Cammisuli:
Sorry, but this isn't the route we want to take with this. 

The implementation PR for the http cache is here: https://github.com/nrwl/nx/pull/30593

Exelord:
> Our commitment is that no major open-source feature or API will be replaced with a non-open-source alternative.

Empty words.

PR #30615, opened 2025-04-04
fix(module-federation): enhance remote entry handling with query parameters in paths

### LINKED ISSUE #30602: Query params in remoteEntry URL are not taken into account
### Current Behavior

We have a dynamic module module federation setup in our angular nx monorepo. 

The host application sets the following remoteDefinitions using **setRemoteDefinitions**: 

```
{
    'sample-app': 'http://localhost:4201/remoteEntry.mjs?version=1234',
    'sample-app-2': 'http://localhost:4202/remoteEntry.mjs?version=1234'
}
```

When we load the remote module using **loadRemoteModule** method, the network request is set to 'http://localhost:4201/remoteEntry.mjs?version=1234/remoteEntry.mjs'

This is because of what we have here:
https://github.com/nrwl/nx/blob/6e50a011e7cbf7ee2b73a52cc153dcd9ebe5dcea/packages/angular/mf/mf.ts#L135

If the URL does not end in .mjs or .js, we add the default 'remoteEntry.mjs' keyword. But, the query params are not considered. 

### Expected Behavior

Query params should be considered and the request should go to 'http://localthost:4201/remoteEntry.mjs?version=1234' since the version param is important for cache busting.

### GitHub Repo

_No response_

### Steps to Reproduce

Repro steps are mentioned in the current behaviour section. 

### Nx Report

```shell
Node           : 20.17.0
OS             : win32-x64
npm            : 10.8.2

nx (global)            : 20.1.4
nx                     : 20.3.1
@nx/js                 : 20.3.1
@nx/jest               : 20.3.1
@nx/eslint             : 20.3.1
@nx/workspace          : 20.3.1
@nx/angular            : 20.3.1
@nx/cypress            : 20.3.1
@nx/devkit             : 20.3.1
@nx/eslint-plugin      : 20.3.1
@nx/module-federation  : 20.3.1
@nx/plugin             : 20.3.1
@nx/web                : 20.3.1
@nx/webpack            : 20.3.1
typescript             : 5.5.4
Registered Plugins:
@nx/cypress/plugin
@nx/eslint/plugin
@nx/jest/plugin
Community plugins:
@ngrx/effects        : 18.0.2
@ngrx/operators      : 18.0.1
@ngrx/router-store   : 18.0.2
@ngrx/store          : 18.0.2
@ngrx/store-devtools : 18.0.2
ngx-build-plus       : 18.0.0
Local workspace plugins:
         @mfe/workspace-plugin
         mfe
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

ndcunningham:
## Current Behavior

In Module Federation apps, when remotes are defined using URLs that include query string or hash fragments (e.g. for cache busting), those params are not preserved after the application is built.

## Expected Behavior

This PR ensures that query strings and hash fragments are preserved when resolving or generating remote URLs. 

## Related Issue(s)

Fixes #30602

### DIFF
--- a/e2e/react/src/module-federation/core.rspack.test.ts
+++ b/e2e/react/src/module-federation/core.rspack.test.ts
@@ -5,6 +5,7 @@ import {
   killPorts,
   killProcessAndPorts,
   newProject,
+  readJson,
   runCLIAsync,
   runCommandUntil,
   runE2ETests,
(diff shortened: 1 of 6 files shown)


Coly010:
Can you also check if the MF Rspack Crystal plugin needs updated with changes? It’s under plugins in the module-federation package

ndcunningham:
> Can you also check if the MF Rspack Crystal plugin needs updated with changes? It’s under plugins in the module-federation package

I think it should be covered as it's only the remote handling that needs to be updated which is shared between webpack and rspack.

jaysoo (approved):

PR #30619, opened 2025-04-05
feat: add devDependencies support & regex ignored dependencies to `@nx/dependency-checks`

### LINKED ISSUE #30614: `@nx/dependency-checks` fails on missing dev dependencies
### Current Behavior

Behavior was changed in #18780, but the [documentation](https://nx.dev/nx-api/eslint-plugin/documents/dependency-checks#:~:text=The%20rule%20uses,the%20expected%20list.) was not updated accordingly. I do not understand why the changes in #18780 were made if I'm honest:

> if a consumer installed my published package, they will get an error because `foo` package is not installed.

Isn't that exactly the point of `foo` being a dev dependency? With this change its now impossible for spec files not to cause errors, meaning we have to completely ignore those dependencies. Enforcing the presence of a dev dependency is therefore currently infeasible.

### Expected Behavior

Revert the behavior change made in #18780.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Add `foo` to dev dependencies
2. Add a spec file and import `foo` package
3. package.json will error with no way to fix 

### Nx Report

```shell
Node           : 22.13.0
OS             : darwin-arm64
bun            : 1.2.0

nx                 : 20.7.0
@nx/js             : 20.7.0
@nx/workspace      : 20.7.0
@nx/devkit         : 20.3.1
@nx/eslint-plugin  : 20.7.0
typescript         : 5.8.2
Cache Usage: 0.00 B / 185.82 GB
  - @nx/devkit@20.3.1

To fix this, run `nx migrate nx@20.7.0`
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

nwidynski:
This (non-breaking) PR adds support for multiple configuration objects as well as a `production` flag to the `dependency-checks` eslint rule. This is similar to how https://knip.dev/features/production-mode works.

With this change, it is now possible to enforce presence of `devDependencies` for dev-only files, e.g. `.spec`, like so:

```jsonc
// ...
"@nx/dependency-checks": ["error", 
  { "ignoredFiles": ["**/*.spec.ts"] }, 
  { "production": false, "ignoredFiles": ["**/!(*.spec.ts)"] }
]
```

Furthermore, this PR also adds capability to match dependencies via regex pattern in the `ignoredDependencies`.
## Current Behavior

It is impossible to enforce dev-only dependencies.

## Related Issue(s)

Fixes #30614

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -49,8 +49,9 @@ export default ESLintUtils.RuleCreator(
       description: `Checks dependencies in project's package.json for version mismatches`,
     },
     fixable: 'code',
-    schema: [
-      {
+    schema: {
+      type: 'array',
+      items: {
         type: 'object',
         properties: {
           buildTargets: { type: 'array', items: { type: 'string' } },
@@ -66,7 +67,7 @@ export default ESLintUtils.RuleCreator(
         },
         additionalProperties: false,
       },
-    ],
+    },
     messages: {
       missingDependency: `The "{{projectName}}" project uses the following packages, but they are missing from "{{section}}":{{packageNames}}`,
       obsoleteDependency: `The "{{packageName}}" package is not used by "{{projectName}}" project.`,
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -151,12 +151,16 @@ export default ESLintUtils.RuleCreator(
 
       const rootPackageJsonDeps = getAllDependencies(rootPackageJson);
 
+      function includesMatch(value: string, arr: string[]): boolean {
+        return arr.some(pattern => new RegExp(pattern).test(value));
+      }
+
       function validateMissingDependencies(node: AST.JSONProperty) {
         if (!checkMissingDependencies) {
           return;
         }
         const missingDeps = expectedDependencyNames.filter(
-          (d) => !projPackageJsonDeps[d] && !ignoredDependencies.includes(d)
+          (d) => !includesMatch(d, Object.keys(projPackageJsonDeps)) && !includesMatch(d, ignoredDependencies)
         );
 
         if (missingDeps.length) {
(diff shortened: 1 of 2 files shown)


nwidynski:
@FrozenPandaz Any updates here?

packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -50,8 +49,9 @@ export default ESLintUtils.RuleCreator(
       description: `Checks dependencies in project's package.json for version mismatches`,
     },
     fixable: 'code',
-    schema: [
-      {
+    schema: {
+      type: 'array',
+      items: {
```
leosvelperez, line 54: The current changes would allow multiple configurations with the same value for `production` and more than two configurations, which is not ideal. The tooling wouldn't prevent users from making such a mistake. We need to limit to at most two configurations, and a way to enforce that there are not two configurations with the same value for `production`.

Additionally, it would be great if users wouldn't have to repeat common configuration properties for both configurations.

leosvelperez (changes requested):
Thanks for contributing!

Some general feedback:

- Look for operations that could be cached for all visitors. For example:
  - The call to `findNpmDependencies` could be cached using `buildTarget` as the key, most of the time all configs will have the same `buildTarget`
  - The few Regex you added could be created once instead of every time they are needed
- Please do not remove existing comments unless you're addressing them or are outdated

PowerSupply:
Any progress on this?

nwidynski:
@PowerSupply Not for now, I will see whether I can pick this back up later this month.

jaysoo (commented):
Thanks for the work on devDependencies support and regex ignoredDependencies. A few issues:

1. **Regex bug**: `@types/*` is invalid regex (`*` quantifier without a preceding element). This would throw at runtime. Should be `@types/.*` or use glob-to-regex conversion.

2. **Multi-visitor merging** is fragile — combining visitors from different rules can lead to unexpected interactions.

3. **`production` naming** is confusing since it's inverted from what you'd expect (when `production: true`, devDependencies are excluded, but the mental model for "production mode" varies).

4. **Staleness**: This PR has been open for 10+ months with no recent activity. A rebase against current master would likely be needed, and there may be overlapping changes.

If you're still interested in pursuing this, I'd suggest starting with just the regex `ignoredDependencies` feature (simpler, less controversial) as a separate PR.

jaysoo:
Closing this PR as there hasn't been any activity for a while.

PR #30623, opened 2025-04-07
feat(angular): add --bundler=rspack option to app generator

Coly010:
## Current Behavior
The `@nx/angular` app generator currently does not support generating an Angular Rspack application. This makes it slightly more difficult for users to get up and running with Angular Rspack

## Expected Behavior
Add `rspack` as a supported `--bundler` option allowing for easy generationg of new Angular Rspack apps.

### DIFF
--- a/packages/angular/src/generators/application/application.spec.ts
+++ b/packages/angular/src/generators/application/application.spec.ts
@@ -1242,6 +1242,16 @@ describe('app', () => {
         ]
       `);
     });
+    it('should generate a correct setup when --bundler=rspack including a correct config file and no build target', async () => {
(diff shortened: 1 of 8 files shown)


packages/angular/src/generators/application/application.spec.ts
```diff
@@ -1242,6 +1242,16 @@ describe('app', () => {
         ]
       `);
     });
+    it('should generate a correct build target for --bundler=rspack', async () => {
```
leosvelperez, line 1245: nit: test case title seems mileading, we're actually validating that no build target should be generated and the config is correctly generated.

leosvelperez (approved):

PR #30631, opened 2025-04-08
fix(angular): ensure crystal targets for testing when bundler=rspack

Coly010:
## Current Behavior
Angular Rspack relies on Inferred Targets however, when scaffolding the application, the unit test runners are being set up with executors.

## Expected Behavior
Ensure that when `bundler=rspack` unit test runners are being set up with inference plugins

### DIFF
--- a/packages/angular/src/generators/application/lib/add-unit-test-runner.ts
+++ b/packages/angular/src/generators/application/lib/add-unit-test-runner.ts
@@ -12,6 +12,7 @@ export async function addUnitTestRunner(host: Tree, options: NormalizedSchema) {
         projectRoot: options.appProjectRoot,
         skipPackageJson: options.skipPackageJson,
         strict: options.strict,
+        addPlugin: options.addPlugin,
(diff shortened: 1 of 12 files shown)


packages/angular/src/generators/application/lib/add-unit-test-runner.ts
```diff
@@ -4,14 +4,19 @@ import { addJest } from '../../utils/add-jest';
 import { addVitest } from '../../utils/add-vitest';
 import type { NormalizedSchema } from './normalized-schema';
 
-export async function addUnitTestRunner(host: Tree, options: NormalizedSchema) {
+export async function addUnitTestRunner(
+  host: Tree,
+  options: NormalizedSchema,
+  isRspack?: boolean
```
leosvelperez, line 10: You can remove `isRspack?: boolean`. It's a leftover of a previous change.

leosvelperez (approved):

PR #30652, opened 2025-04-09
fix(core): add keepExistingVersions to jest option to preserve dependency versions

ndcunningham:
## Current Behavior

Currently, when we use the jest configuration generator it will forcibly update the jest version if the package already exist.

## Expected Behavior

Now, the jest version will be preserved unless the option is passed to update the version.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/jest/src/generators/configuration/configuration.ts
+++ b/packages/jest/src/generators/configuration/configuration.ts
@@ -71,6 +71,7 @@ function normalizeOptions(
   return {
     ...schemaDefaults,
     ...options,
+    keepExistingVersions: options.keepExistingVersions ?? true,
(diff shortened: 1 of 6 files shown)


isaacplmann (approved):

packages/jest/src/generators/configuration/configuration.ts
```diff
@@ -71,6 +71,7 @@ function normalizeOptions(
   return {
     ...schemaDefaults,
     ...options,
+    keepExistingVersions: options.keepExistingVersions ?? true,
```
leosvelperez, line 74: This default value doesn't match the schema default value.

jaysoo (approved):

PR #30687, opened 2025-04-11
fix(core): clean up batch process after post run

xiongemi:
## Current Behavior

currently, the batch process is not being cleanup

## Expected Behavior

need to kill the batch process after post run in task orchestrator, otherwise, when kill it during the post run (store the output to cache), it will cause the process to hang in DTE

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/task-orchestrator.ts
+++ b/packages/nx/src/tasks-runner/task-orchestrator.ts
@@ -301,6 +301,7 @@ export class TaskOrchestrator {
     }
 
     await this.postRunSteps(tasks, results, doNotSkipCache, { groupId });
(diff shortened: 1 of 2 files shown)


packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -320,6 +320,8 @@ export class TaskOrchestrator {
         groupId
       );
     }
+
+    this.forkedProcessTaskRunner.cleanUpBatchProcesses();
```
FrozenPandaz, line 324: Move this up

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #30701, opened 2025-04-12
chore: remove `@nx/workspace/valid-command-object` from eslint rules …

Avivbens:
## Current Behavior
An ESLint error occurs when opening a file
   
CleanShot 2025-04-12 at 15 20 57@2x

### No setup for this rule

CleanShot 2025-04-12 at 15 16 39@2x

## Expected Behavior

ESLint will have all the required dependencies set up.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -75,8 +75,7 @@
               }
             ]
           }
-        ],
-        "@nx/workspace/valid-command-object": "error"
+        ]
       }
     },
     {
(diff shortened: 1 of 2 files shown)


leosvelperez:

Please create an issue to discuss first and with information to reproduce the problem. I can't reproduce it in the repo. The rule is working as expected, and there are no errors in the IDE, so it might be happening to you due to some environment-specific conditions or an edge case that's not apparent.

Either way, the solution wouldn't be to remove the rule from the configuration but to fix any potential issues with its resolution. I'm going to close this for the time being. If you create the issue with further information, you can tag me, and I'll take a look.

PR #30705, opened 2025-04-13
fix(vite): add logic to prevent overwriting generated package.json

### LINKED ISSUE #30312: nxViteTsPaths overwrites generated package.json
### Current Behavior

When using nxViteTsPaths as a vite plugin for bundling a library and also using rollup-plugin-generate-pacakge-json (or any other plugin to create/modify/generate a `package.json`) the project directory's `package.json` gets [copied to the outputDir](https://github.com/nrwl/nx/blob/20.5.0/packages/vite/plugins/nx-tsconfig-paths.plugin.ts#L204) during the [`writeBundle` lifecycle hook of Rollup](https://rollupjs.org/plugin-development/#writebundle) resulting in the generated `package.json` file getting overwritten.

### Expected Behavior

Several things:
1)  `nxViteTsPaths` ignores the `generatePackageJson` configuration from `@nx/vite:build`. 

2) `nxViteTsPaths` has no options related to copying a `package.json` file. Earlier versions of `nxViteTsPaths` do not copy/modify files at all. The copying of `package.json` is undocumented and the side-effect of overwriting a `package.json` that's generated during build is unexpected. 

The documentation: 

https://nx.dev/nx-api/vite/documents/overview#installation 

> add [nxViteTsPaths plugin](https://nx.dev/recipes/vite/configure-vite#typescript-paths), so your projects can use workspace libraries.

https://nx.dev/recipes/vite/configure-vite#typescript-paths

> You need to use the nxViteTsPaths() plugin to make sure that your TypeScript paths are resolved correctly in your monorepo.

3) `nxViteTsPaths` is a ['pre' enforced plugin](https://github.com/nrwl/nx/blob/4f3118d5f8bd411b06122596e97ac571ff7466d7/packages/vite/plugins/nx-tsconfig-paths.plugin.ts#L80), yet it uses the [`writeBundle` hook](https://rollupjs.org/plugin-development/#writebundle), which is the last hook in rollup runs. Consider using [`generateBundle` hook](https://rollupjs.org/plugin-development/#generatebundle) instead so other plugins can run after.  

4) IMO the functionality of copying the `package.json` file at the end of the build should be a completely separate plugin as it doesn't align with the documented use case for `nxViteTsPaths`. This functionality was added for the inference feature and was added to an existing plugin instead of creating a new one.

### Steps to Reproduce

1) Create an NX monorepo
2) Add a library bundled with vite that modifies the package.json during build
3) Use nxViteTsPaths plugin

See reproduction: https://stackblitz.com/edit/vitejs-vite-6smqcoqr?file=package.json

`npm run build:with` builds with nxViteTsPaths plugin
`npm run build:without` builds without nxViteTsPaths plugin

Notice difference in dist/libs/mylib/package.json#exports.

### Nx Report

```shell
Node           : 18.20.3
OS             : linux-x64
Native Target  : wasm32
npm            : 10.2.3

nx             : 20.5.0
@nx/js         : 20.5.0
@nx/workspace  : 20.5.0
@nx/devkit     : 20.5.0
@nx/vite       : 20.5.0
@nx/web        : 20.5.0
typescript     : 5.7.3
Registered Plugins:
@nx/js/typescript
@nx/vite/plugin
```

### Additional Information

The issue is at https://github.com/nrwl/nx/blob/20.5.0/packages/vite/plugins/nx-tsconfig-paths.plugin.ts#L196-L210

@Coly010 added [the `writeBundle` function](https://github.com/nrwl/nx/blob/20.5.0/packages/vite/plugins/nx-tsconfig-paths.plugin.ts#L196) as part of [support for the inference plugin](https://github.com/nrwl/nx/commit/64c6287b83cfd6fb1801252703c07aca7026481c).

This was introduced in `@nx/vite@19.1.0`.

JSMike:
The nxViteTsPaths plugin currently always copys the package.json file at the end of the build and does not check if the file was generated from the build process. This adds logic to check if the package.json file already exists in the dist path before copying to prevent overwriting generated files.

Closes #30312

## Current Behavior
nxViteTsPaths always copies package.json at end of build.

## Expected Behavior
nxViteTsPaths only copies package.json at end of build if package.json not generated during build.

## Related Issue(s)

Fixes #30312

### DIFF
--- a/packages/vite/plugins/nx-tsconfig-paths.plugin.ts
+++ b/packages/vite/plugins/nx-tsconfig-paths.plugin.ts
@@ -203,9 +203,9 @@ export function nxViteTsPaths(options: nxViteTsPathsOptions = {}) {
       if (isUsingTsSolutionSetup()) return;
       const outDir = options.dir || 'dist';
       const src = resolve(projectRoot, 'package.json');
-      if (existsSync(src)) {
-        const dest = join(outDir, 'package.json');
+      const dest = join(outDir, 'package.json');
 
+      if (existsSync(src) && !existsSync(dest)) {
         try {
           copyFileSync(src, dest);
         } catch (err) {


JSMike:
FYI: This was a fairly simple change, I tested this by applying the changes directly to the reproduction provided in the issue and had the expected results. Let me know if you'd like a more rigorous test.

JSMike:
> I've the same issue. Maybe it should also check if `generatePackageJson` is set to `true` when copying. I don't want any `package.json` in my `dist` folder.

@openscript I looked into that a little bit and it would likely require a larger change. I couldn't find a straight-forward way to get the NX run configuration within the scope of the Vite plugin, but maybe there's a way to do that. I would suggest to add that effort to separate pull request as this change is very simple.

I did mention listening to `generatePackageJson` in https://github.com/nrwl/nx/issues/30312

openscript:
Thanks @JSMike for looking into it. I went with just deleting the `package.json` after it's generated via an npm post script hook.

JSMike:
@Coly010 just checking that this is on your radar. Let me know if this is good or if you have any suggested changes.

Coly010 (approved):

JSMike:
@Coly010 @FrozenPandaz thank you!

PR #30736, opened 2025-04-15
feat: expand task ctx by overrides and nx args

nwidynski:
## Current Behavior

Nx tasks only receive the nx workspace root configuration.

## Expected Behavior

Task lifecycle hooks should be able to run logic depending on the overrides and current nx args.

## Related Issue(s)

Fixes #


nwidynski:
@FrozenPandaz Any updates on this?

AgentEnder:
Hey, sorry this has sat around for quite some time. I'm going to close this out as its gotten a bit outdated, but if you still want this functionality open an issue so we can discuss the use cases. In particular for nxArgs / overrides, we currently recommend inspecting process.argv directly.

PR #30743, opened 2025-04-15
fix(js): remove check that ts plugin is used for ts sync generator

jaysoo:
This PR ensures that when `@nx/js/typescript` is not used, we do not add `@nx/js:typescript-sync` sync generator to targets such as build, serve, etc.

It resolves issues where `nx init` into a repo that is compatible with TS solution will add the sync generator, even if the plugin is unused. It leads to errors everytime users run a task like build. The error is like this:

```
[@nx/js:typescript-sync] The `@nx/js/typescript` plugin is not registered

...

> Would you like to ignore the sync failures and continue running the task?
Yes
No
```

It makes it confusing for users, especially new users that don't know what sync generators are. They will always run into the error and have to choose to continue despite the failure.

**Note:** In a future follow-up, we could consider adding better info and prompts so we can let users know that Nx also helps keep workspace up to date, and can learn more about it.

## Current Behavior
Users see an error when running `nx add @nx/vite` and then `nx build <project>`.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/init/init-v2.ts
+++ b/packages/nx/src/command-line/init/init-v2.ts
@@ -69,12 +69,6 @@ export async function initHandler(options: InitArgs): Promise<void> {
   const _isMonorepo = _isNonJs ? false : isMonorepo(packageJson);
   const _isCRA = _isNonJs ? false : isCRA(packageJson);
 
-  const learnMoreLink = _isTurborepo
-    ? 'https://nx.dev/recipes/adopting-nx/from-turborepo'
-    : _isMonorepo
-    ? 'https://nx.dev/getting-started/tutorials/npm-workspaces-tutorial'
-    : 'https://nx.dev/recipes/adopting-nx/adding-to-existing-project';
-
   /**
    * Turborepo users must have set up individual scripts already, and we keep the transition as minimal as possible.
    * We log a message during the conversion process in addNxToTurborepo about how they can learn more about the power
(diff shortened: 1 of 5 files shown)


leosvelperez (changes requested):
The sync generator doesn't need the plugin to be registered. That's an outdated check in the sync generator that we should remove. Let's repurpose this to do that, and maybe the only extra thing we need in the plugins is to check whether the `@nx/js` package is resolvable from the root (otherwise, the sync generator wouldn't be found), but I don't remember if we always install it, in which case it wouldn't be needed.

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -160,7 +154,13 @@ export async function initHandler(options: InitArgs): Promise<void> {
   }
 
   printFinalMessage({
-    learnMoreLink,
+    appendLines: _isMonorepo
+      ? [
+          `- Learn how Nx helps manage your TypeScript monorepo. https://nx.dev/features/maintain-ts-monorepos`,
+        ]
+      : [
+          `- Learn how Nx works with any type of project. https://nx.dev/recipes/adopting-nx/adding-to-existing-project`,
```
leosvelperez, line 162:
```suggestion
          `- Learn how Nx helps manage your TypeScript monorepo at https://nx.dev/features/maintain-ts-monorepos`,
        ]
      : [
          `- Learn how Nx works with any type of project at https://nx.dev/recipes/adopting-nx/adding-to-existing-project`,
```

leosvelperez (approved):

PR #30746, opened 2025-04-15
feat(core): clean up unneeded continuous tasks after tasks are done

FrozenPandaz:
## Current Behavior

Continuous tasks go until the process is done.

## Expected Behavior

Continuous tasks are cleaned up after they're no longer needed. AKA once their dependent tasks are done.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/tui/lifecycle.rs
+++ b/packages/nx/src/native/tui/lifecycle.rs
@@ -114,11 +114,12 @@ impl AppLifeCycle {
     pub fn print_task_terminal_output(
         &mut self,
         task: Task,
-        status: String,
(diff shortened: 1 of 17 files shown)


packages/nx/src/native/tui/lifecycle.rs
```diff
@@ -110,19 +110,6 @@ impl AppLifeCycle {
         Ok(())
     }
 
-    #[napi]
-    pub fn print_task_terminal_output(
```
JamesHenry, line 114: This is needed for cache hits?

AgentEnder (approved):

PR #30769, opened 2025-04-17
feat(angular): remove deprecated functionalities

leosvelperez:
Remove the deprecated functionalities scheduled to be removed in Nx v21.

BREAKING CHANGE: Remove the deprecated data persistence operators previously exported in `@nx/angular` and the deprecated testing utils previously exported in `@nx/angular/testing`.


packages/angular/migrations.json
```diff
@@ -362,6 +293,15 @@
       "version": "21.0.0-beta.3",
       "description": "Set the `continuous` option to `true` for continuous tasks.",
       "factory": "./src/migrations/update-21-0-0/set-continuous-option"
+    },
+    "change-data-persistence-operators-imports-to-ngrx-router-store-data-persistence": {
```
leosvelperez, line 297: This one duplicates a migration that runs for Nx 16.2, but it's needed if someone has opted out of it in earlier versions. Since we're removing the operators in Nx v21, we must ensure we migrate those cases.

packages/angular/generators.json
```diff
@@ -86,7 +86,7 @@
       "schema": "./src/generators/move/schema.json",
       "aliases": ["mv"],
       "description": "Moves an Angular application or library to another folder within the workspace and updates the project configuration.",
-      "x-deprecated": "Use the `@nx/workspace:move` generator instead. This generator will be removed in Nx v19."
+      "x-deprecated": "Use the `@nx/workspace:move` generator instead. This generator will be removed in Nx v22."
```
Coly010, line 89: Is this bumped simply because we missed it?

Coly010 (approved):
One comment just :)

packages/angular/generators.json
leosvelperez: We didn't have a proper deprecation notice in the programmatic API, so I'm adding it to this PR and extending the notice until the next major. It's a generator we haven't touched in a while, so it's low maintenance at this point. We can live with it for another major.

Coly010: Sounds good!

packages/angular/src/migrations/update-21-0-0/change-data-persistence-operators-imports-to-ngrx-router-store-data-persistence.md
```diff
@@ -0,0 +1,139 @@
...
+
+{% tabs %}
+{% tab label="Before" %}
+
+```ts {% fileName="apps/app1/src/app/users/users.effects.ts" highlightLines=[2] %}
+import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { fetch } from '@nx/angular';
+
+@Injectable()
+class UsersEffects {
```
jaysoo, line 17: Minor: I think the `export` is missing in the before example?

jaysoo (approved):

FrozenPandaz (approved):

PR #30809, opened 2025-04-22
feat(linter): add support for package based secondary entry points

### LINKED ISSUE #18552: ESLint enforce-module-boundaries rule fails with library split in secondary entry points and mixed loading scenarios
### Current Behavior

The ESLint rule enforce-module-boundaries enforce libs to be lazy loaded as far as a dynamic import is detected. This is great for most scenarios, but not working well when a library is split into multiple secondary entry points.

Consider the following scenario where **lib A** is split into 2 secondary entry points **core** and **feature**, and **lib B** needs to lazy load **feature** and to statically load **core**:

image

Then the enforce-module-boundaries rule is failing because lib A is marked as "lazy", and everything should be lazy loaded as far as a dynamic import is detected. 

```ts
import { CommonModule } from '@angular/common';
import { NgModule } from '@angular/core';

/* Here is the line that throws "Static imports of lazy-loaded libraries are forbidden."
   but static import of the secondary entry point "core" should be allowed. */
import { CoreModule } from '@myorg/lib-a/core';

/* Only "feature" is lazy, not the whole library and all the secondary entry-points. */
async function getFeatureModule() {
  const  { FeatureModule } = await import('@myorg/lib-a/feature');
  return FeatureModule;
}

getFeatureModule();

@NgModule({
  imports: [CommonModule, CoreModule],
})
export class LibBModule {
}
```

### Expected Behavior

Statically importing **core** secondary entry points should not throw `static imports of lazy-loaded libraries are forbidden`.

Secondary entry points should be taken into account when checking for static imports when a library is considered lazy, allowing the rule to statically import secondary entry points that have not been imported dynamically.

### GitHub Repo

https://github.com/edbzn/nx-static-import-issue/tree/main

### Steps to Reproduce

1. Clone the [reproduction](https://github.com/edbzn/nx-static-import-issue)
2. Run `nx lint lib-b`
3. See the [line](https://github.com/edbzn/nx-static-import-issue/blob/e7c7dd1371a39f85edaf34143030d03c37fbb04a/libs/lib-b/src/lib/lib-b.module.ts#L6) that is failing

### Nx Report

```shell
nx 16.6.0
```

### Failure Logs

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

meeroslav:
This PR adds support for package.json based secondary entry points and implements fix for situation when package imports base entry point as dynamic dependency and secondary entry point as static dependency.

## Current Behavior
When the package is imported from itself, check for a secondary entry point checks only Angular-style secondary entry points.

When package is importing from the same library as dynamic import from root and static import from secondary entry point we still get linter errror.

## Expected Behavior
Check for secondary entry points should also support standard package.json-based entry points.

Importing from the same library as dynamic import from root and static import from secondary entry point should be allowed.

## Related Issue(s)

Fixes #18552


packages/eslint-plugin/src/utils/runtime-lint-utils.ts
```diff
@@ -533,6 +564,31 @@ function getAngularEntryPoint(file: string, projectRoot: string): string {
...
+  );
+  if (!packageContent) {
+    return [];
+  }
+  const exports = parseJson(packageContent).exports;
+  if (!exports) {
+    return [];
+  }
+  const entryPaths: Array<{ path: string; file: string }> = [];
+  for (const [key, value] of Object.entries<string>(exports)) {
```
leosvelperez, line 581: The `exports` property [could be a `string`](https://nodejs.org/docs/latest-v22.x/api/packages.html#exports-sugar).

packages/eslint-plugin/src/utils/runtime-lint-utils.ts
```diff
@@ -533,6 +564,31 @@ function getAngularEntryPoint(file: string, projectRoot: string): string {
...
+    if (key !== '.' && key !== './') {
+      entryPaths.push({
+        path: joinPathFragments(projectRoot, key),
+        file: joinPathFragments(projectRoot, value),
```
leosvelperez, line 585: The `value` here could be a `string`, but it could also be an `object` with [conditional exports](https://nodejs.org/docs/latest-v22.x/api/packages.html#conditional-exports). It can also be `null` to disable the entry point from being used.

packages/eslint-plugin/src/utils/runtime-lint-utils.ts
```diff
@@ -490,35 +499,57 @@ export function groupImports(
...
+  const entryPoint = getEntryPoint(resolvedImportFile, projectRoot);
+  return entryPoint;
+}
 
-  const importEntryPoint = getAngularEntryPoint(
-    resolvedImportFile,
-    projectRoot
+function getEntryPoint(file: string, projectRoot: string): string {
+  const packageEntryPoints = getPackageEntryPoints(projectRoot);
+  const fileEntryPoint = packageEntryPoints.find(
```
leosvelperez, line 537: The entry points could have patterns (`*`), which would require a more complex matching to find whether an import matches a given entry point.

packages/eslint-plugin/src/utils/runtime-lint-utils.ts
```diff
@@ -533,6 +564,31 @@ function getAngularEntryPoint(file: string, projectRoot: string): string {
...
+  }
+  const exports = parseJson(packageContent).exports;
+  if (!exports) {
+    return [];
+  }
+  const entryPaths: Array<{ path: string; file: string }> = [];
+  for (const [key, value] of Object.entries<string>(exports)) {
+    if (key !== '.' && key !== './') {
+      entryPaths.push({
+        path: joinPathFragments(projectRoot, key),
```
leosvelperez, line 584: The `key` could be the condition name of a [conditional export](https://nodejs.org/docs/latest-v22.x/api/packages.html#conditional-exports).

packages/eslint-plugin/src/utils/runtime-lint-utils.ts
meeroslav: Thanks, let me rework that.

packages/eslint-plugin/src/utils/runtime-lint-utils.ts
meeroslav: Let's handle this later if users request it. I'm worried this might impact performance significantly so I would not add it unless requested.

leosvelperez (approved):

PR #30818, opened 2025-04-22
feat(core)!: drop support for node versions prior to 20.19.0

ndcunningham:
BREAKING CHANGE: We are updating our minimum node version to `20.19.0` which drops support for all prior versions.


JamesHenry (changes requested):
See slack thread

docs/generated/packages/workspace/documents/nx-nodejs-typescript-version-matrix.md
```diff
@@ -13,8 +13,9 @@ and the version of NodeJS that we tested it against.
 
 | Nx Version      | Node Version     | Typescript Version |
 | --------------- | ---------------- | ------------------ |
-| 20.x (latest)   | 22.x, 20.x       | ~5.4.2             |
-| 19.x (previous) | 22.x, 20.x, 18.x | ~5.4.2             |
+| 21.x (latest)   | 22.x, ^20.19.0   | ~5.4.2             |
+| 20.x (previous) | 22.x, 20.x, 18.x | ~5.4.2             |
```
JamesHenry, line 17: We need to fix the currently live site by cherry-picking this commit onto the `website` branch so it needs to make sense for right now:

```suggestion
| 21.x (next)   | 22.x, ^20.19.0   | ~5.4.2             |
| 20.x (latest) | 22.x, 20.x, 18.x | ~5.4.2             |
```

.github/workflows/generate-embeddings.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [18]
+        node-version: ['20.19.0']
```
jaysoo, line 13: We can leave it at `20` which currently pulls in 20.19, and will automatically update to the new minor versions as they are released.

JamesHenry (approved):

packages/nx/native-packages/linux-arm64-gnu/package.json
```diff
@@ -18,7 +18,7 @@
     "nx.linux-arm64-gnu.node"
   ],
   "engines": {
-    "node": ">= 10"
+    "node": ">= 20.19.0"
```
jaysoo, line 21: Let's remove the `engine` entries. The compat matrix is good enough.

beaussan:
Hey there 👋

I suspect the node 20.19 was picked for the `require(esm)` support

If that's the case, you may want to support only node 22.12 and up instead of 22.x

Since it's not enabled under 22.12 by default, you would have node version in your support range that don't support `require(esm)`


See release notes for 22.12 https://nodejs.org/en/blog/release/v22.12.0

JamesHenry (approved):

jaysoo (approved):

PR #30822, opened 2025-04-23
feat(core): stream task output for all tasks except for direct output…

FrozenPandaz:
… tasks

## Current Behavior

Currently, only running tasks which are a PseudoTTYProcess stream outputs to the TUI.

## Expected Behavior

Most running tasks have `onOutput` even if they are not a `PseudoTTYProcess`. This method is used to stream outputs to the TUI. The only sort of running task which does not have this `onOutput` are tasks with direct output... Those remain unhandled.. for now.

## Related Issue(s)

Fixes #


JamesHenry (approved):

AgentEnder (commented):
lgtm :)

PR #30826, opened 2025-04-23
fix(webpack): interpolate process.env more verbosely to reduce bundle size with DefinePlugin

coolassassin:
## Current Behavior
When we prepare ENVs for the DefinePlugin, we are creating the `process.env` object.
For example:
```
// .env
NX_PUBLIC_VALUE1=1
NX_PUBLIC_VALUE2=2
NX_PUBLIC_VALUE3=3
```
As result we will have:
```js
{
   'process.env': {
      "NX_PUBLIC_VALUE1": "1",
      "NX_PUBLIC_VALUE2": "2",
      "NX_PUBLIC_VALUE3": "3"
   }
}
```

As a result, in the final bundle, we will replace process.env with this object.
The issue:
If I use all 3 values in my application DefinePlugin will inject this object 3 times, instead of injecting it once.
It will look like that:
```js
const a = {
      "NX_PUBLIC_VALUE1": "1",
      "NX_PUBLIC_VALUE2": "2",
      "NX_PUBLIC_VALUE3": "3"
}.NX_PUBLIC_VALUE1
const b = {
      "NX_PUBLIC_VALUE1": "1",
      "NX_PUBLIC_VALUE2": "2",
      "NX_PUBLIC_VALUE3": "3"
}.NX_PUBLIC_VALUE2
const c = {
      "NX_PUBLIC_VALUE1": "1",
      "NX_PUBLIC_VALUE2": "2",
      "NX_PUBLIC_VALUE3": "3"
}.NX_PUBLIC_VALUE3
```

## Expected Behavior
DefinePlugin injects values instead of env object in each place
```js
const a = "1"
const b = "2"
const c = "3"
```

## Fixes
- fixed this issue for webpack
- fixed this issue for storybook
- fixed this issue for rspack

TLDR:
now we have object like so:
```js
{
    "process.env.NX_PUBLIC_VALUE1": "1",
    "process.env.NX_PUBLIC_VALUE2": "2",
    "process.env.NX_PUBLIC_VALUE3": "3"
}
```

### DIFF
--- a/packages/react/plugins/storybook/index.ts
+++ b/packages/react/plugins/storybook/index.ts
@@ -51,10 +51,15 @@ function getClientEnvironment(mode) {
     );
 
   // Stringify all values so we can feed into webpack DefinePlugin
-  const stringified = Object.keys(raw).reduce((env, key) => {
-    env[`process.env.${key}`] = JSON.stringify(raw[key]);
-    return env;
-  }, {});
+  const stringified = Object.keys(raw).reduce(
+    (env, key) => {
+      env[`process.env.${key}`] = JSON.stringify(raw[key]);
+      return env;
+    },
+    // Provide a fallback for process.env itself to handle cases where code
+    // accesses process.env directly (e.g., in Cypress component testing)
+    { 'process.env': '{}' } as Record<string, string>
+  );
 
   return { stringified };
 }
--- a/packages/react/plugins/storybook/index.ts
+++ b/packages/react/plugins/storybook/index.ts
@@ -51,12 +51,10 @@ function getClientEnvironment(mode) {
     );
 
   // Stringify all values so we can feed into webpack DefinePlugin
-  const stringified = {
-    'process.env': Object.keys(raw).reduce((env, key) => {
-      env[key] = JSON.stringify(raw[key]);
-      return env;
-    }, {}),
-  };
+  const stringified = Object.keys(raw).reduce((env, key) => {
+    env[`process.env.${key}`] = JSON.stringify(raw[key]);
+    return env;
+  }, {});
 
   return { stringified };
 }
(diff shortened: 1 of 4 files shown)


Coly010 (approved):

coolassassin:
@Coly010 could you help me. Should I try to fix this failed test, or it is just something flaky?

coolassassin:
@Coly010 I can't reproduce this bug in the job on my MacOS laptop:
image

coolassassin:
@claude can you take a look at a test? I can't reproduce it on my macOS machine

packages/react/plugins/storybook/index.ts
```diff
@@ -51,10 +51,15 @@ function getClientEnvironment(mode) {
...
-    return env;
-  }, {});
+  const stringified = Object.keys(raw).reduce(
+    (env, key) => {
+      env[`process.env.${key}`] = JSON.stringify(raw[key]);
+      return env;
+    },
+    // Provide a fallback for process.env itself to handle cases where code
+    // accesses process.env directly (e.g., in Cypress component testing)
+    { 'process.env': '{}' } as Record<string, string>
```
coolassassin: I am not sure that it is safe to do so, because if `process.env` is replaced before `process.env.anyKey`, it will break the whole definition.
I guess we can do it like so:
```js
const stringified = Object.keys(raw).reduce((env, key) => {
    env[`process.env.${key}`] = JSON.stringify(raw[key]);
    return env;
  }, {});

const stringifiedWithFallback = Object.keys(stringified).length > 0 ? stringified : { 'process.env': '{}' } as Record<string, string>

return { stringified: stringifiedWithFallback}
```

Or maybe like so:
```js
const keys = Object.keys(raw);
const stringified = keys.length > 0 ? 
  Object.keys(raw).reduce((env, key) => {
    env[`process.env.${key}`] = JSON.stringify(raw[key]);
    return env;
  }, {}): 
  { 'process.env': '{}' } as Record<string, string>
  
return { stringified }
```

Or if it is an issue with Cypress, it will be better to fix it there. It is weird that the e2e-framework affects the code

Coly010: Yep good spot

coolassassin: I see that you have made a rollback. It's not working, right?

PR #30836, opened 2025-04-23
fix(graph): disable primary CTA for Migrate UI when approval is required first

jaysoo:
This PR updates the Migrate UI such that the primary CTA always performs an action. Previously, when there is a migration to approve, the primary CTA says `Run Migrations`, but clicking it does nothing since we're already in the middle of running migrations.

Screenshot 2025-04-23 at 9 36 05 AM

Now, with this PR, the primary CTA will be disabled and tell the user to approve the current migration before continuing.

## Current Behavior
Primary CTA in Migrate UI does nothing when there is a migration to approve

## Expected Behavior
Primary CTA should reflect the current action that the user must take, in this case to approve the migration

## Related Issue(s)

Fixes #

### DIFF
--- a/graph/migrate/.storybook/main.ts
+++ b/graph/migrate/.storybook/main.ts
@@ -11,10 +11,12 @@ const config: StorybookConfig = {
     options: {},
   },
 
-  viteFinal: async (config) =>
-    mergeConfig(config, {
-      plugins: [],
-    }),
+  viteFinal: async (config) => {
+    const {
+      nxViteTsPaths,
+    } = require('@nx/vite/plugins/nx-tsconfig-paths.plugin');
+    return mergeConfig(config, { plugins: [nxViteTsPaths()] });
+  },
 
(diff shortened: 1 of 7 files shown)


graph/migrate/.storybook/main.ts
```diff
@@ -11,10 +11,12 @@ const config: StorybookConfig = {
...
   },
 
-  viteFinal: async (config) =>
-    mergeConfig(config, {
-      plugins: [],
-    }),
+  viteFinal: async (config) => {
+    const {
+      nxViteTsPaths,
+    } = require('@nx/vite/plugins/nx-tsconfig-paths.plugin');
```
jaysoo, line 17: This is needed for Storybook to serve. Before it was a top-level import, which messed up when doing through `ts-node` in the publish pipeline (only for FreeBSD).

ndcunningham (approved):

PR #30838, opened 2025-04-23
feat(release)!: useLegacyVersioning is false by default, migrate config

JamesHenry:
This PR updates `nx release` to use the revamped versioning implementation by default. It also updates and adds relevant documentation, and provides an automated migration for the new configuration structure.

For anyone exclusively versioning TS/JS packages, there should be no real difference to your experience (although a number of bugs have been fixed and new features around updating multiple `package.json` files at once are now available to you with this change).

For the lifecycle of Nx v21, `release.version.useLegacyVersioning` will remain as a option that can be set to `true` to revert to the old behavior and configuration structure.

NOTE: This should only be a temporary solution, for example if one of the plugins you use does provide a `VersionActions` based versioning implementation yet. The option and legacy implementation will be removed entirely in Nx v22 (in ~6 months).

BREAKING CHANGE:

**⚠️ For any early adopters of `VersionActions` in Nx 20.8 when it was opt-in, there are breaking changes to the abstract class here as well.**

`manifestRootsToUpdate` has gone from `string[]` to `manifestsToUpdate: { manifestPath: string; preserveLocalDependencyProtocols: boolean; }[]` to support controlling the local dependency updates per manifest in order to support advanced source vs dist scenarios, and correspondingly `isLocalDependencyProtocol` has been removed from the abstract class and the method will no longer be called from the core logic. It should be thought of as an implementation detail of `updateProjectDependencies` instead.


packages/nx/src/command-line/release/config/config.spec.ts
```diff
@@ -12316,35 +12767,35 @@ describe('createNxReleaseConfig()', () => {
   describe('useLegacyVersioning', () => {
     it('should be true by default', async () => {
```
JamesHenry, line 12768:
```suggestion
    it('should be false by default', async () => {
```

FrozenPandaz (approved):

PR #30859, opened 2025-04-24
fix(core): update Homebrew cache path to /opt/Homebrew to be consistent with Apple Silicon default path

ndcunningham:
Currently, we are caching the wrong path `/usr/local/Homebrew` for homebrew on macos  it should be `/opt/Homebrew`


AgentEnder (approved):

ndcunningham:
closed in favour of https://github.com/nrwl/nx/pull/30836 which contains the same changes

PR #30878, opened 2025-04-25
feat(graph): add undo migration option when one is pending approval

jaysoo:
This PR adds a button for user to undo a migration that's already been applied and pending approval.

See: https://www.loom.com/share/97286bdc80ea4538af76a914ef8f0f8b

Also, fixes an existing issue where `migrations.json` did not record the correct git sha for each commit.

## Current Behavior
When a migration is pending approval, the only option is to accept it.

## Expected Behavior
Allow user to undo a migration if they don't want the changes.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/migrate/migrate-ui-api.ts
+++ b/packages/nx/src/command-line/migrate/migrate-ui-api.ts
@@ -171,6 +171,16 @@ export async function runSingleMigration(
         cwd: workspacePath,
         encoding: 'utf-8',
       });
+      // The revision changes after the amend, so we need to update it
+      const amendedGitRef = execSync('git rev-parse HEAD', {
+        cwd: workspacePath,
+        encoding: 'utf-8',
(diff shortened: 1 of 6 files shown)


packages/nx/src/command-line/migrate/migrate-ui-api.ts
```diff
@@ -171,6 +171,16 @@ export async function runSingleMigration(
         cwd: workspacePath,
         encoding: 'utf-8',
       });
+      // The revision changes after the amend, so we need to update it
```
jaysoo, line 174: Previously, the recorded `ref` in `migrations.json` were all wrong since they do not include the amended changes to `migrations.json`. This fixes an existing issue.

ndcunningham (commented):
LGTM

ndcunningham (approved):

PR #30880, opened 2025-04-25
feat(core): allow react-native and expo to specify that they are continuous

ndcunningham:
Update react-native and expo executor to be continuous

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,16 @@ jobs:
           pnpm install --frozen-lockfile
           pnpm playwright install --with-deps
 
+      ## TODO remove this after checking why it is needed for macos pipeline else it fails with /bin/sh: bun: command not found
+      ## See ref: https://staging.nx.app/runs/g75tGV9tMK?utm_source=pull-request&utm_medium=comment
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: '1.2.11'
+
+      - name: Verify Bun installation
+        run: bun --version
+
       - name: Install Rust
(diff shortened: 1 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -167,6 +167,14 @@ jobs:
           pnpm install --frozen-lockfile
           pnpm playwright install --with-deps
 
+      - name: Setup Bun
```
jaysoo, line 170: If this fixes the pipeline somehow, we should add a TODO to look into it more and remove it. Bun should not even be needed, and this will add some time to CI.

e2e/release/src/preserve-local-dependency-protocols.test.ts
```diff
@@ -166,7 +166,7 @@ describe('nx release preserve local dependency protocols', () => {
       Would stage files in git with the following command, but --dry-run was set:
       git add {project-name}/package.json {project-name}/package.json
     `);
-  });
+  }, 300_000);
```
FrozenPandaz, line 169: Remove this change

.nx/workflows/agents.yaml
```diff
@@ -9,8 +9,6 @@ launch-templates:
       GIT_COMMITTER_NAME: Test
       COREPACK_DEFAULT_TO_LATEST: '0'
       COREPACK_ENABLE_AUTO_PIN: '0'
-      SELECTED_PM: 'pnpm'
```
FrozenPandaz, line 12: Add this back

FrozenPandaz (changes requested):

ndcunningham:
closed in favour of: https://github.com/nrwl/nx/pull/30985

PR #30901, opened 2025-04-28
docs(nx-cloud): address typos in assignment rules docs

lourw:
## Expected Behavior

1. remove unnecessary references to `runs-on`
2. fix ordering of rule precedence example so that it matches the explanatory paragraph
3. modify the matrix github actions example so that agents of different types do not have the same name

## Related Issue(s)

Fixes #


isaacplmann (approved):

madelk (commented):
Line 239 mentions the second rule having three properties, but it the first rule with three properties

madelk (commented):
Line 188 uses target rather than targets

madelk:
`runs-on` is still used elsewhere in the doc

PR #30936, opened 2025-04-29
chore(module-federation): disable test that is causing agents to run …

FrozenPandaz:
…out of memory

## Current Behavior

Many pipelines are stuck because this task is causing agents to run out of memory.

## Expected Behavior

This test is disabled so that pipelines can complete.

## Related Issue(s)

Fixes #


ndcunningham (approved):
We need to look into reenabling this :)

PR #30947, opened 2025-04-30
chore(core): add tui layout manager

JamesHenry:


packages/nx/src/native/tui/app.rs
```diff
@@ -774,17 +939,442 @@ impl App {
...
+        // Recalculate the layout areas
+        self.recalculate_layout_areas();
+
+        // Ensure the pty instances get resized appropriately
+        let _ = self.debounce_pty_resize();
+
+        self.dispatch_action(Action::SetSpacebarMode(spacebar_mode));
+    }
+
+    fn focus_next(&mut self) {
```
FrozenPandaz, line 1039: future: Maybe create a focus manager

This is a bunch of code which was moved from tasks_list.rs..

So the story so far seems to be. app.rs was too big.. split it out into tasks_list.rs... then now we're saying tasks_list component is too big... so we've moved this back into app.rs....

I think there's enough code handling focus here to justify it's own module?

packages/nx/src/native/tui/app.rs
```diff
@@ -774,17 +939,442 @@ impl App {
...
+        self.recalculate_layout_areas();
+        // Ensure the pty instances get resized appropriately (no debounce as this is based on an imperative user action)
+        let _ = self.handle_pty_resize();
+    }
+
+    fn cycle_layout_modes(&mut self) {
+        // TODO: add visual feedback about layout modes
+        self.layout_manager.cycle_layout_mode();
+        self.recalculate_layout_areas();
+        // Ensure the pty instances get resized appropriately (no debounce as this is based on an imperative user action)
```
FrozenPandaz, line 1166: Can you explain when we need to debounce and when we don't to me one more time please?

packages/nx/src/native/tui/app.rs
```diff
@@ -774,17 +939,442 @@ impl App {
...
+            let terminal_pane_data = &mut self.terminal_pane_data[pane_idx];
+            terminal_pane_data.handle_key_event(key)
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Returns true if the currently focused pane is in interactive mode.
+    fn is_interactive_mode(&self) -> bool {
+        match self.focus {
```
FrozenPandaz, line 1255: future: I think interactive mode belongs on the app even though the terminal pane is receiving the inputs. Interactive mode means at the app level, almost all key events have to get sent into the pane. So rather than checking the pane to see if it's interactive, I think it's easier if we track interactive mode in the app.rs

packages/nx/src/native/tui/app.rs
```diff
@@ -774,17 +939,442 @@ impl App {
...
+                    rows
+                } else {
+                    0
+                };
+                let mut pty_clone = pty.as_ref().clone();
+                pty_clone.resize(pty_height, pty_width)?;
+
+                // If dimensions changed, mark for sort
+                if old_rows != pty_height {
+                    needs_sort = true;
```
FrozenPandaz, line 1315: future: This confuses me.. does the layout manager not know if it will change the size of a pty? When would the pty size change.. without the layout manager changing its size?

packages/nx/src/native/tui/components.rs
```diff
@@ -12,40 +12,110 @@ use super::{
...
+    ///
+    /// # Arguments
+    ///
+    /// * `tx` - An unbounded sender that can send actions.
+    ///
+    /// # Returns
+    ///
+    /// * `Result<()>` - An Ok result or an error.
     fn register_action_handler(&mut self, tx: UnboundedSender<Action>) -> Result<()> {
+        let _ = tx; // to appease clippy
```
FrozenPandaz, line 36: You can appease clippy by naming it _ in the signature.

packages/nx/src/native/tui/components.rs
```diff
@@ -12,40 +12,110 @@ use super::{
...
+    ///
+    /// # Arguments
+    ///
+    /// * `area` - Rectangular area to initialize the component within.
+    ///
+    /// # Returns
+    ///
+    /// * `Result<()>` - An Ok result or an error.
+    fn init(&mut self, area: Size) -> Result<()> {
+        let _ = area; // to appease clippy
```
FrozenPandaz, line 49: Same

packages/nx/src/native/tui/components.rs
```diff
@@ -12,40 +12,110 @@ use super::{
...
-            Some(Event::Mouse(mouse_event)) => self.handle_mouse_events(mouse_event)?,
+        let action = match event {
+            Some(Event::Key(key_event)) => self.handle_key_event(key_event)?,
+            Some(Event::Mouse(mouse_event)) => self.handle_mouse_event(mouse_event)?,
             _ => None,
         };
-        Ok(r)
+        Ok(action)
     }
-    #[allow(unused_variables)]
```
FrozenPandaz, line 36: Or you can bring this back

packages/nx/src/native/tui/components.rs
```diff
@@ -12,40 +12,110 @@ use super::{
...
-        let r = match event {
-            Some(Event::Key(key_event)) => self.handle_key_events(key_event)?,
-            Some(Event::Mouse(mouse_event)) => self.handle_mouse_events(mouse_event)?,
+        let action = match event {
+            Some(Event::Key(key_event)) => self.handle_key_event(key_event)?,
+            Some(Event::Mouse(mouse_event)) => self.handle_mouse_event(mouse_event)?,
             _ => None,
         };
-        Ok(r)
+        Ok(action)
```
FrozenPandaz, line 67: Is this the main way of sending out events from components? They also seem to have Action Senders? We ideally only have 1 way.

packages/nx/src/native/tui/components/help_popup.rs
```diff
@@ -9,6 +9,9 @@ use ratatui::{
     },
 };
 use std::any::Any;
+use tokio::sync::mpsc::UnboundedSender;
```
FrozenPandaz, line 12: Do the crossbeam senders work? Why use the tokio ones? We mostly use the crossbeam senders elsewhere in the codebase.

packages/nx/src/native/tui/app.rs
```diff
@@ -635,107 +680,227 @@ impl App {
...
+                                terminal_pane_data.is_continuous = task.continuous;
+                                let in_progress = task.status == TaskStatus::InProgress;
+                                terminal_pane_data.can_be_interactive = in_progress;
+                                if !in_progress {
+                                    terminal_pane_data.set_interactive(false);
+                                }
+
+                                let mut has_pty = false;
+                                if let Some(pty) = self.pty_instances.get(&relevant_pane_task) {
+                                    terminal_pane_data.pty = Some(pty.clone());
```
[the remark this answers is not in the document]
Cammisuli: This clone is fine because its cloning an `Arc` rather than the underlying struct. I made changes so that we dont clone the screen.

packages/nx/src/native/tui/components/layout_manager.rs
```diff
@@ -0,0 +1,964 @@
...
+            return false;
+        }
+
+        // If very tall and not very wide, prefer vertical
+        if aspect_ratio < 1.0 && terminal_width < self.min_horizontal_width {
+            return true;
+        }
+
+        // Otherwise, prefer horizontal for wider terminals, vertical for taller ones
+        aspect_ratio < 1.5
```
FrozenPandaz, line 415: I think this gets simpler if we look at it this way:

There are constraints which force vertical or horizontal if there's not enough space.

If the terminal area is within those constraints then an aspect ratio of 1.5 is basically where we draw the line between preferring vertical vs horizontal.

FrozenPandaz (approved):

PR #31011, opened 2025-05-02
feat(core): run parallel based on the number of cpu cores

ianzone:
Current Behavior
can not specify parallel based on the number of CPU cores.

Expected Behavior
--parallel=50% means half of the CPU cores.

Related Issue(s)
https://github.com/nrwl/nx/discussions/30238


MaxKless (approved):
thanks!!

packages/nx/src/command-line/yargs-utils/shared-options.ts
```diff
@@ -407,3 +408,15 @@ const coerceTuiAutoExit = (value: string) => {
   }
   throw new Error(`Invalid value for --tui-auto-exit: ${value}`);
 };
+
+function concurrency(str: string) {
+  let parallel = parseInt(str);
```
jaysoo, line 413: `args['parallel']` and `args['maxParallel']` can come in as a number. We need to check if it's a number of string first.

ianzone:
Can someone trigger the CI

MaxKless:
thanks again for your work :)

PR #31012, opened 2025-05-02
feat(core): enhance pnpm detection in CI workflow

ndcunningham:
This PR updates our CI workflow to detect if the repository has `pnpm` inside their package.json.

If it does it should install it at the version specified else install the default version which is `9.8.0`


ndcunningham:
Not needed for now

PR #31018, opened 2025-05-02
fix(core): update ci-workflow generator package manager installation

ndcunningham:
This pull request updates the CI workflow configuration to replace the use of `pnpm/action-setup` with `corepack enable` for package manager setup. 

- NPM versions greater than 8 has an error which has been [fixed in Node 22.5.1](https://github.com/npm/cli/issues/7639), since we are still using Node 20 we need to pin the installed npm version to 8.
- Now we are conditionally handling different package managers based on how the repo has been configured.
Should the repository specify the package manager's version for `pnpm` and `yarn` those versions will be respected, else it will fallback to a `9.8` for `pnpm` and `1.22` for `yarn`.

### DIFF
--- a/packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
+++ b/packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
@@ -2074,6 +2074,52 @@ jobs:
 "
 `;
 
+exports[`CI Workflow generator connected to nxCloud with yarn should generate github CI config when packageManager is defined in package.json 1`] = `
+"name: CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  actions: read
+  contents: read
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          filter: tree:0
+          fetch-depth: 0
+
+      - run: corepack enable
+
+      # This enables task distribution via Nx Cloud
+      # Run this command as early as possible, before dependencies are installed
+      # Learn more at https://nx.dev/ci/reference/nx-cloud-cli#npx-nxcloud-startcirun
+      # Uncomment this line to enable task distribution
+      # - run: yarn nx-cloud start-ci-run --distribute-on="3 linux-medium-js" --stop-agents-after="build"
+
+      # Cache node_modules
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'yarn'
+
+      - run: yarn install --frozen-lockfile
+      - uses: nrwl/nx-set-shas@v4
+
+      # Prepend any command with "nx-cloud record --" to record its logs to Nx Cloud
+      # - run: yarn nx-cloud record -- echo Hello World
+      # Nx Affected runs only tasks affected by the changes in this PR/commit. Learn more: https://nx.dev/ci/features/affected
+      - run: yarn nx affected -t lint test build
+"
+`;
+
 exports[`CI Workflow generator connected to nxCloud with yarn should generate github CI config with custom name 1`] = `
 "name: My custom-workflow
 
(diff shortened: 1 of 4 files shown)


jaysoo (changes requested):
I think it's better to check it in the generator itself so we don't output additional logic in the pipeline file. I'm also not sure if gitlab, circleci, bitbucket, azure has the same issue.

It's worth checking and verify the other CI, but we should update github for sure.

packages/workspace/src/generators/ci-workflow/files/github/.github/workflows/__workflowFileName__.yml__tmpl__
```diff
@@ -18,16 +18,26 @@ jobs:
         with:
           filter: tree:0
           fetch-depth: 0
-
-      <% if(packageManager == 'pnpm'){ %>
-      - uses: pnpm/action-setup@v4
```
FrozenPandaz, line 23: pnpm still recommends using their action to setup so I think we should keep this.

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -174,18 +174,23 @@ jobs:
           filter: tree:0
           fetch-depth: 0
 
-      # This enables task distribution via Nx Cloud
```
FrozenPandaz, line 177: This should not have moved down.

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -174,18 +174,23 @@ jobs:
...
-      # Uncomment this line to enable task distribution
-      # - run: npx nx-cloud start-ci-run --distribute-on="3 linux-medium-js" --stop-agents-after="e2e-ci"
-
       # Cache node_modules
       - uses: actions/setup-node@v4
         with:
           node-version: 20
           cache: 'npm'
 
+      ## NPM versions that are > v8 has an issue which has been solved in Node 22.5.1
```
FrozenPandaz, line 183: What if we update to Node 22.5+?

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -1006,22 +1024,20 @@ jobs:
           filter: tree:0
           fetch-depth: 0
 
-      - uses: pnpm/action-setup@v4
+      # Cache node_modules
```
FrozenPandaz, line 1027: The order should not have changed here and we should continue using the pnpm setup action since that's what they still recommend on the docs.

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
```diff
@@ -1006,22 +1024,20 @@ jobs:
...
+
+      - run: corepack prepare pnpm@9.8
 
       # This enables task distribution via Nx Cloud
       # Run this command as early as possible, before dependencies are installed
       # Learn more at https://nx.dev/ci/reference/nx-cloud-cli#npx-nxcloud-startcirun
       # Uncomment this line to enable task distribution
       # - run: pnpm dlx nx-cloud start-ci-run --distribute-on="3 linux-medium-js" --stop-agents-after="build"
 
-      # Cache node_modules
```
FrozenPandaz, line 1019: This should not have been removed.

packages/workspace/src/generators/ci-workflow/ci-workflow.spec.ts
```diff
@@ -337,18 +337,23 @@ describe('CI Workflow generator', () => {
                   filter: tree:0
                   fetch-depth: 0
 
-              # This enables task distribution via Nx Cloud
```
FrozenPandaz, line 340: This should not have moved.

FrozenPandaz (changes requested):

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
ndcunningham: We remove installing npm v8

packages/workspace/src/generators/ci-workflow/ci-workflow.spec.ts
ndcunningham: It hasnt been removed the line number just changed

packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
FrozenPandaz: Let's do that then..

FrozenPandaz (approved):

jaysoo (approved):

PR #31022, opened 2025-05-02
fix(gradle): fix workflow to run assemble check with batch

xiongemi:
## Current Behavior

currently, the command in ci workflow is `nx affected -t build`

## Expected Behavior

change to run `nx affected -t assemble, check` with batch turned on for gradle

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/src/generators/ci-workflow/__snapshots__/generator.spec.ts.snap
+++ b/packages/gradle/src/generators/ci-workflow/__snapshots__/generator.spec.ts.snap
@@ -6,6 +6,9 @@ exports[`ci-workflow generator connected to nxCloud circleci pipeline should mat
 orbs:
   nx: nrwl/nx@1.6.2
 
+env:
+  NX_BATCH_MODE: true
+
 jobs:
   main:
     environment:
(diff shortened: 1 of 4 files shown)


FrozenPandaz (approved):

packages/gradle/src/generators/ci-workflow/__snapshots__/generator.spec.ts.snap
```diff
@@ -6,12 +6,15 @@ exports[`ci-workflow generator connected to nxCloud circleci pipeline should mat
...
 
+env:
+  NX_BATCH_MODE: true
+
 jobs:
   main:
     environment:
       # Configure the JVM and Gradle to avoid OOM errors
       _JAVA_OPTIONS: '-Xmx3g'
-      GRADLE_OPTS: '-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2'
```
FrozenPandaz, line 14: I think you should bring back the gradle workers flag? :thinking:

FrozenPandaz (approved):

PR #31027, opened 2025-05-03
fix(rspack): must run type check with @nx/rspack:rspack when skipTypeChecking is false

dpnolte:
## Current Behavior
- Create a project that uses @nx/rspack:rspack as build target with SkipTypeChecking set to false.
- Run build target for project.
- Type check has not been run and will not bail in case of any typescript errors

## Expected Behavior

- It should run type checking as documented: https://nx.dev/nx-api/rspack/executors/rspack#skiptypechecking
I looked at the webpack executor and there it also runs the type check based on `skipTypeChecking` and not the `typeCheck` alias. So I've adjusted it accordingly for the rspack executor.

## Related Issue(s)

Fixes #31026

### DIFF
--- a/packages/rspack/src/executors/rspack/rspack.impl.ts
+++ b/packages/rspack/src/executors/rspack/rspack.impl.ts
@@ -30,7 +30,7 @@ export default async function* runExecutor(
     normalizedOptions.mode = process.env.NODE_ENV;
   }
 
-  if (normalizedOptions.typeCheck) {
+  if (!normalizedOptions.skipTypeChecking) {
     await executeTypeCheck(normalizedOptions, context);
   }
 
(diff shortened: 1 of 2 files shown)


Coly010:
@claude Can you add some normalization to the `normalizeOptions` in packages/rspack/src/plugins/utils/plugins/normalize-options.ts that will correctly return `typeCheck` and `skipTypeChecking`? `typeCheck` should be the inverse of `skipTypeChecking`. As both `typeCheck` and `skipTypeChecking` are optional in the `options` argument, we should also account for one of the options being set, while the other isn't. The default case for when neither option is provided should be `typeCheck: true` and `skipTypeChecking: false`

Coly010 (approved):

PR #31092, opened 2025-05-06
docs(nx-dev): add continuous tasks and terminal ui articles

philipjfulcher:


docs/blog/2025-05-07-continuous-tasks.md
```diff
@@ -0,0 +1,261 @@
...
+      "dev": {
+        "dependsOn": [{ "projects": ["api"], "target": "serve" }]
+      },
+      "serve": {
+        "dependsOn": [{ "projects": ["api"], "target": "serve" }]
+      }
+    }
+  }
+}
+```
```
juristr, line 141: Could we just configure it such that the frontend dev command is always `dev` and thus also the `e2e` depends on that?

I feel like this introduces unnecessary confusion because up until here we had

- frontend => `dev`
- backend => `serve`

which is nice because it allows to show the configuration options. But introducing `serve` here for the frontend and then adding another `dependsOn` for that adds unnecessary opportunities for confusion IMHO.

I'd still mention the sentence that if you're frontend also has a way of serving pre-built static assets (which is common) to make sure to include that in case.

juristr (changes requested):
Also pushed some commits to

- fix typos
- fix some confusion between the `frontend:serve` vs `frontend:dev` which was used inconsistently between paragraphs

Feel free to autosquash the fixup commits into the original ones if these edits work. I just didn't want to force-push

docs/blog/2025-05-07-continuous-tasks.md
philipjfulcher: This is due to some confusion on my part, but I think other people could be confused as well.

Vite projects have inferred dev and serve targets. I didn't realize until just now that they're aliases for the same vite command. I've worked in codebases before where dev was for serving in dev mode, serve would serve in production mode. I assumed there was a difference between the two.

When you add a Playwright config to a Vite project, the config defaults to running `serve`. This fed into my assumption that serve would serve in production mode, since you'd probably want to run e2e tests against something running in prod mode.

In this case, I think we could collapse these down and reduce some confusion. But perhaps we could also add something to the descriptions for these inferred targets to indicate they're aliases of each other?

juristr: > When you add a Playwright config to a Vite project, the config defaults to running serve. This fed into my assumption that serve would serve in production mode, since you'd probably want to run e2e tests against something running in prod mode.

yeah I see what you mean. That'd be the `vite preview` target for Vite CLI projects. I don't even know why we generate/infer the `serve-static` still for Vite tbh 🤔 . And yes the fact that our playwright setup uses `nx serve` rather than `nx dev` is confusing. If at all it should just use `dev` or even better the `preview` / `serve-static` task (at least in CI)

But tl;dr what I meant is that given that confusion I'd probably really collapse this section or remove it almost entirely. It doesn't add much to the story as we've already explained it further up I think.

philipjfulcher: Took out the section and added a paragraph to the earlier section: 

> In addition to making for a great local development experience, e2e test suites that also run `frontend:dev` will have the same experience. The frontend and backend will be served at the same time, making e2e tests easier to run locally.

philipjfulcher: Also removed that section from the video

juristr (approved):
LGTM

PR #31103, opened 2025-05-07
feat(devkit): allow requiring cts config files

nicolas-marien:
When migrating our project to esm, we encountered an issue with the playright plugin, but more generally with the `loadConfigFile` from the devkit.
 Our configuration is a `.cts` file, but it's not treated as commonjs: `__dirname` and `__filename` are not available.

CleanShot 2025-05-07 at 15 31 04@2x

## Expected Behavior
`.cts` files are interpreted as commonJS files when in a module context.


barbados-clemens (approved):

jaysoo (approved):

jaysoo (changes requested):
I think if we know that the file is specifically `.cts` or `.mts`, we can choose either `require` or dynamic `import` without having to do what `load` function currently does (which is try one, then try the other).

nicolas-marien:
@jaysoo thank you for the feedback. I'll update the code accordingly.

jaysoo (approved):

jaysoo:
I tested this locally. I had to make the `mts` use `load` instead of `loadEsm` as there might be some CJS specific contents in there like `__filename`. Depending on `tsconfig.base.json`, the modules may be transpiled as CJS anyway.

This PR now fixes the `.cts` issue, while maintaining same support for `.ts`, `.mts`. And for `mjs`, it'll go straight to loading via ESM.

PR #31107, opened 2025-05-07
feat(vite): add ajv dev dependency for vite-plugin-dts

ndcunningham:
## Current Behavior

Currently, when we generate a vite library that uses `vite-plugin-dts` it breaks the project graph and all nx commands fail. 
It happens because the wrong version of `ajv` is being hoisted in `node_modules/ajv` so the expected module path required from `vite-plugin-dts` is not found.

## Expected Behavior

Nx commands should work out of the box when we generate a library with vite.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/vite/.eslintrc.json
+++ b/packages/vite/.eslintrc.json
@@ -44,7 +44,8 @@
               // we only check if the package is installed
               "eslint",
               // we ensure it is installed and only use it when eslint is installed
-              "@nx/eslint"
+              "@nx/eslint",
+              "ajv"
             ]
           }
         ]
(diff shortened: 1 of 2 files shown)


packages/vite/src/utils/versions.ts
```diff
@@ -12,6 +12,9 @@ export const happyDomVersion = '~9.20.3';
 export const edgeRuntimeVmVersion = '~3.0.2';
 export const jitiVersion = '2.4.2';
 
+// Peer dependecies
+export const ajvVersion = '^8.0.0'; // Needed for vite-plugin-dts to be hoisted
```
jaysoo, line 16: It might be better to add it to `@nx/vite/package.json`. Emily had to do something similar for react-native. Can you try it?

jaysoo (approved):

PR #31136, opened 2025-05-08
docs(release): set proper heading indentation

nesjett:
## Current Behavior
Headings in the documentation site are wrongly arranged, making it difficult to navigate.
image

## Expected Behavior
"Options" heading should be nested inside the parent heading topic.


nesjett:
Just a friendly reminder. Do I need to change anything for this to be approved?

JamesHenry:
@nesjett You've edited a file with "generated" (in its path because it's an autogenerated file), so this isn't a valid resolution. The generation tooling will need to updated to better handle this

cc @isaacplmann

Coly010:
Given the docs refactor, if we need this we can reopen.

PR #31148, opened 2025-05-09
feat(core): add nx console messaging to TUI

Cammisuli:
## Current Behavior

There is no communication between Nx CLI and Nx Console

## Expected Behavior

This enables a connection between the Nx TUI app and Nx Console so that we can send messages to console. This is used to update MCP tools on Nx Console to assist with LLMs

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/tui/app.rs
+++ b/packages/nx/src/native/tui/app.rs
@@ -1586,7 +1586,7 @@ impl App {
             .as_ref()
             .is_some_and(|c| c.is_connected())
         {
-            self.dispatch_action(Action::ConsoleMessagesAvailable(true));
+            self.dispatch_action(Action::ConsoleMessengerAvailable(true));
         }
     }
 }
--- a/packages/nx/src/native/tui/app.rs
+++ b/packages/nx/src/native/tui/app.rs
@@ -23,7 +23,6 @@ use crate::native::{
     tasks::types::{Task, TaskResult},
 };
 
-use super::action::Action;
 use super::components::Component;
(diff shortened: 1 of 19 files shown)


packages/nx/src/native/tui/app.rs
```diff
@@ -769,8 +779,25 @@ impl App {
             trace!("{action:?}");
         }
         match &action {
+            Action::StartCommand(_) => {
+                self.console_messenger
+                    .as_ref()
+                    .and_then(|c| c.start_running_tasks());
+            }
+            Action::Tick => {
+                self.console_messenger.as_ref().and_then(|messenger| {
```
JamesHenry, line 788: Should we throttle/debounce in some way? Updating everything on every tick is quite intense, no?

JamesHenry (commented):
Please can we add some more logs in the happy path when messages are being send to console?

packages/nx/src/native/tui/app.rs
Cammisuli: Yes. That happens here:
https://github.com/nrwl/nx/pull/31148/files#diff-395a7cc58ef509a48734eb3ffee39299d9af81020dd9115329de1c4404165faeR106-R108

Cammisuli:
> Please can we add some more logs in the happy path when messages are being send to console?

Logs do show up when messages are sent. They're part of the underlying `jsonrpsee` crate. Like this:
image

FrozenPandaz (approved):

FrozenPandaz (approved):

PR #31154, opened 2025-05-09
fix(core): show all native logs for debug mode

FrozenPandaz:
## Current Behavior

TUI debug mode only shows native logs since the TUI launched.

## Expected Behavior

TUI debug mode shows logs even before the TUI launched such as task hashing logs and more.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/logger/mod.rs
+++ b/packages/nx/src/native/logger/mod.rs
@@ -109,6 +109,7 @@ pub(crate) fn enable_logger() {
     let registry = tracing_subscriber::registry()
         .with(stdout_layer)
         .with(TuiTracingSubscriberLayer);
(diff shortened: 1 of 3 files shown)


packages/nx/src/native/logger/mod.rs
```diff
@@ -109,6 +109,7 @@ pub(crate) fn enable_logger() {
     let registry = tracing_subscriber::registry()
         .with(stdout_layer)
         .with(TuiTracingSubscriberLayer);
+    tui_logger::init_logger(tui_logger::LevelFilter::Trace).unwrap();
```
Cammisuli, line 112: `expect("")` or `.ok()` please.

Cammisuli (approved):

PR #31157, opened 2025-05-09
fix(core): simplify action dispatch, use mutex locking, and avoid unnecessary clones

Cammisuli:
## Current Behavior

* There are many unwraps that we can avoid by using `parking_lot::mutex`
* There is an unnecessary `task::spawn` to dispatch actions. The channel is non-blocking and this is wasted cycles
* There are a bunch of clones that are happening with the PtyInstances

## Expected Behavior

* Uses `parking_lot::Mutex` and removes `unwraps()`
* Dispatches actions directly without going into tasks
* Uses tuple derefs to avoid clones, and uses 2 instead of 4

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/tui/app.rs
+++ b/packages/nx/src/native/tui/app.rs
@@ -1034,10 +1034,11 @@ impl App {
 
     /// Dispatches an action to the action tx for other components to handle however they see fit
     fn dispatch_action(&self, action: Action) {
-        let tx = self.action_tx.clone().unwrap();
-        tokio::spawn(async move {
-            let _ = tx.send(action);
-        });
+        if let Some(tx) = &self.action_tx {
+            tx.send(action).unwrap_or_else(|e| {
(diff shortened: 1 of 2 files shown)


packages/nx/src/native/tui/app.rs
```diff
@@ -1034,10 +1034,11 @@ impl App {
 
     /// Dispatches an action to the action tx for other components to handle however they see fit
     fn dispatch_action(&self, action: Action) {
-        let tx = self.action_tx.clone().unwrap();
-        tokio::spawn(async move {
-            let _ = tx.send(action);
-        });
+        if let Some(tx) = &self.action_tx {
+            tx.send(action).unwrap_or_else(|e| {
+                debug!("Failed to dispatch action: {}", e);
```
FrozenPandaz, line 1039: Food for thought, should this be `error!()`?

FrozenPandaz (approved):

PR #31163, opened 2025-05-10
fix(core): make running a single task more transparent

FrozenPandaz:
## Current Behavior

Running a single task is in this uncanny framed state and doesn't exit immediately.

image

## Expected Behavior

Running a single task isn't framed and exits immediately.

This also fixes scrolling in interactive mode based on whether or not the underlying terminal is in application cursor mode or not.

image

## Related Issue(s)

Fixes #


packages/nx/src/native/tui/app.rs
```diff
@@ -970,6 +984,7 @@ impl App {
                                 );
 
                                 let terminal_pane = TerminalPane::new()
+                                    .minimal(tasks_list_hidden && self.tasks.len() <= 1)
```
JamesHenry, line 987: How come not just == 1?

AgentEnder (approved):

PR #31164, opened 2025-05-12
fix(storybook): include storybook pkg version migration

dmitry-stepanenko:
Just noticed that storybook package is not being migrated to latest if it's installed explicitly


dmitry-stepanenko:
I see that rust installation has failed, should I trigger the CI again?

PR #31193, opened 2025-05-13
docs(core): disable tutorialkit for firefox

isaacplmann:
- Unrevert removing tutorialkit
- Disable interactive (right) side of tutorialkit when using Firefox
- Run npm install in the terminal instead of in the preview pane.  This makes it so that users can't enter commands until the install command finishes


Coly010:
Going to close this PR given its status and lack of attention - also considering docs site has had a refactor. If we need this we can reopen

PR #31198, opened 2025-05-13
fix(gradle): make ci inputs same as test inputs

xiongemi:
## Current Behavior

current ci test target does not contain inputs of test, it just contains the test source file

## Expected Behavior

change the ci targets to contain inputs of test
Screenshot 2025-05-13 at 1 31 22 PM

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/gradle/project-graph/src/main/kotlin/dev/nx/gradle/utils/CiTargetsUtils.kt
+++ b/packages/gradle/project-graph/src/main/kotlin/dev/nx/gradle/utils/CiTargetsUtils.kt
@@ -88,8 +88,10 @@ private fun buildTestCiTarget(
     testFile: File,
     testTask: Task,
     projectRoot: String,
-    workspaceRoot: String
+    workspaceRoot: String,
 ): MutableMap<String, Any?> {
+  val taskInputs = getInputsForTask(testTask, projectRoot, workspaceRoot, null)
+
   val target =
       mutableMapOf<String, Any?>(
           "executor" to "@nx/gradle:gradle",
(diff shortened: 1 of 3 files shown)


packages/gradle/project-graph/src/test/kotlin/dev/nx/gradle/utils/AddTestCiTargetsTest.kt
```diff
@@ -76,7 +76,6 @@ class AddTestCiTargetsTest {
     val firstTarget = targets["ci--MyFirstTest"]!!
     assertEquals(firstTarget["executor"], "@nx/gradle:gradle")
     assertEquals(true, firstTarget["cache"])
```
FrozenPandaz, line 78: Can this test actually list the inputs?

packages/gradle/project-graph/src/main/kotlin/dev/nx/gradle/utils/CiTargetsUtils.kt
```diff
@@ -88,8 +88,13 @@ private fun buildTestCiTarget(
     testFile: File,
     testTask: Task,
     projectRoot: String,
-    workspaceRoot: String
+    workspaceRoot: String,
 ): MutableMap<String, Any?> {
+  val taskInputs = getInputsForTask(testTask, projectRoot, workspaceRoot, null)
+  val testFileInput = replaceRootInPath(testFile.path, projectRoot, workspaceRoot)
```
FrozenPandaz, line 94: Is the test file not part of the other task inputs? Would this repeat it twice?

FrozenPandaz (changes requested):

packages/gradle/project-graph/src/main/kotlin/dev/nx/gradle/utils/CiTargetsUtils.kt
xiongemi: change to taskInputs

packages/gradle/project-graph/src/test/kotlin/dev/nx/gradle/utils/AddTestCiTargetsTest.kt
xiongemi: changed

FrozenPandaz (approved):

PR #31221, opened 2025-05-14
fix(core): prefer vertical layout categorically if there are less tha…

FrozenPandaz:
…n 75 characters of width

## Current Behavior

Auto layout is always based on an aspect ratio which lead to horizontal layout even when there is very little horizontal space.

## Expected Behavior

Auto layout prefers vertical layout with less than 75 characters of width which will allow 50 characters (75 * 2/3) for terminal output. It should be enough to show a URL in most cases:

image
image

## Related Issue(s)

Fixes #


AgentEnder (approved):

JamesHenry (dismissed):
It makes sense, we can definitely refine the auto detection a lot further, the initial work just got the "plumbing" in place. Please add a snapshot test for this

PR #31249, opened 2025-05-16
fix(core): ensure that global nx works with dot-nx workspace after init

jaysoo:
This PR fixes an issue with the global `nx` command, where running `nx init` in a non-JS folder (i.e. using dot-nx setup) results in a workspace that doesn't work with the global command.

e.g.

```
brew install nx
cd some-java-project
nx init
nx report
```

https://www.loom.com/share/e8dbb2fb7a084300a4dd1e5dff2c0db1

## Current Behavior
`nx report` or any other command fails

## Expected Behavior
`nx` commands should work

## Related Issue(s)

Fixes #


FrozenPandaz (approved):

gioragutt:
@jaysoo random question 😅 why does your change fix this? Does running the local version of NX like this perform some side-effect that then makes the global version work properly?

PR #31294, opened 2025-05-21
chore(angular): ensure e2es with rspack builds are minimized

Coly010:
## Current Behavior
The e2e tests running Angular Rspack are not having their builds minimized currently.

## Expected Behavior
Ensure `NODE_ENV=production` is set when calling the build commands

### DIFF
--- a/e2e/angular/src/module-federation.rspack.test.ts
+++ b/e2e/angular/src/module-federation.rspack.test.ts
@@ -46,7 +46,7 @@ describe('Angular Module Federation', () => {
     let rspackConfigFileContents = readFile(join(hostApp, 'rspack.config.ts'));
     let updatedConfigFileContents = rspackConfigFileContents.replace(
       `maximumError: '1mb'`,
-      `maximumError: '11mb'`
+      `maximumError: '3mb'`
     );
     updateFile(join(hostApp, 'rspack.config.ts'), updatedConfigFileContents);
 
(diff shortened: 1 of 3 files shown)


e2e/angular/src/module-federation.rspack.test.ts
```diff
@@ -43,23 +43,23 @@ describe('Angular Module Federation', () => {
...
-      `maximumError: '1mb'`,
-      `maximumError: '11mb'`
-    );
-    updateFile(join(hostApp, 'rspack.config.ts'), updatedConfigFileContents);
+    // let rspackConfigFileContents = readFile(join(hostApp, 'rspack.config.ts'));
+    // let updatedConfigFileContents = rspackConfigFileContents.replace(
+    //   `maximumError: '1mb'`,
+    //   `maximumError: '11mb'`
+    // );
+    // updateFile(join(hostApp, 'rspack.config.ts'), updatedConfigFileContents);
```
leosvelperez, line 51: Should we remove this?

e2e/angular/src/module-federation.rspack.test.ts
```diff
@@ -43,23 +43,23 @@ describe('Angular Module Federation', () => {
...
-      `maximumError: '1mb'`,
-      `maximumError: '11mb'`
-    );
-    updateFile(join(remoteApp1, 'rspack.config.ts'), updatedConfigFileContents);
+    // rspackConfigFileContents = readFile(join(remoteApp1, 'rspack.config.ts'));
+    // updatedConfigFileContents = rspackConfigFileContents.replace(
+    //   `maximumError: '1mb'`,
+    //   `maximumError: '11mb'`
+    // );
+    // updateFile(join(remoteApp1, 'rspack.config.ts'), updatedConfigFileContents);
```
leosvelperez, line 62: Should we remove this?

leosvelperez (approved):

e2e/angular/src/module-federation.rspack.test.ts
Coly010: I will - I just want to verify with MF that the budget does not need to change first because the MF runtime will be injected into the entrypoint(s)

PR #31321, opened 2025-05-23
feat(nx-dev): add PHP landing page

jaysoo:
This PR adds a `/php` landing page that will eventually be used to promote PHP support via `@nx/php`.


jaysoo:
Closing until we figure out a strategy around this.

PR #31372, opened 2025-05-28
fix(rspack): Add missing return for rspack composoable plugin

nadavshatz:
This was broken in https://github.com/nrwl/nx/pull/29691

it's missing this return.

### DIFF
--- a/packages/rspack/src/utils/read-rspack-options.ts
+++ b/packages/rspack/src/utils/read-rspack-options.ts
@@ -20,9 +20,8 @@ export async function readRspackOptions(
   const resolveConfig = async (
     config: unknown
   ): Promise<Configuration | Configuration[]> => {
-    let resolvedConfig: Configuration;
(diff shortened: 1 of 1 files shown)


ndcunningham (changes requested):
Hey, thanks for the PR!

Can we instead return here: https://github.com/nrwl/nx/blob/1f0acc6cd31c26ba88eb8bfefe378e1e5aa4ca0c/packages/rspack/src/utils/read-rspack-options.ts#L25

and remove https://github.com/nrwl/nx/blob/1f0acc6cd31c26ba88eb8bfefe378e1e5aa4ca0c/packages/rspack/src/utils/read-rspack-options.ts#L23

nadavshatz:
sure thing - was trying to do the least amount of changes. but i agree this is nicer. on it

ndcunningham (approved):

PR #31395, opened 2025-05-30
docs(core): add cacheKeyPrefix option to s3 remote cache options for cache organization flexibility

ndcunningham:
This pull request updates the documentation for the S3 cache plugin to introduce the `cacheKeyPrefix` setting. 

This new option allows users to add a prefix to cache keys, enabling better organization and separation of cached data across projects and environments.


Coly010:
Given docs refactor, I'm going to close this PR. If we need it we can reopen correctly

PR #31414, opened 2025-06-01
feat(repo): Add golden list of projects to our Nightly CI

ndcunningham:
This pull request introduces several updates to the CI/CD workflows and matrix configuration files. 

The aim is to highlight critical Nx failures contained in each project for maintainers to address.

### Changes
- Improvements to workflow caching.
- Improvements to macOS simulator handling.
- Updates to Slack notifications.
- Update matrix data processing for golden projects. 
- Support for Windows has been temporarily disabled due to build issues.


.github/workflows/e2e-matrix.yml
```diff
@@ -36,8 +37,8 @@ jobs:
             node_version: 22
           # - os: macos-latest
           #   node_version: 23
-          - os: windows-latest
-            node_version: 22
+          # - os: windows-latest
```
jaysoo: It'd be great to fix up the Gradle issue in this PR, but if not there should be a `TODO` comment and a fast-follow on it so we're not missing Windows in nightlies. @xiongemi can help out

.github/workflows/nightly/process-matrix.ts
```diff
@@ -69,7 +71,8 @@ const matrix: Array<{
   package_manager: string,
   os: string,
   os_name: string,
-  os_timeout: number
+  os_timeout: number,
+  isGolden?: boolean,
```
jaysoo: Nit: since the pipeline files and existing keys use underscores, we should use `is_golden`.

We should also explain what this property does somewhere in this file.

.github/workflows/nightly/process-result.ts
```diff
@@ -0,0 +1,192 @@
+import * as fs from 'fs';
+
+interface MatrixResult {
```
jaysoo: This is the same interface as before, right? We can use it again here.

.github/workflows/nightly/process-result.ts
```diff
@@ -0,0 +1,192 @@
...
+    Object.entries(results).forEach(([key, value]) => {
+      setOutput(key, value);
+    });
+  } catch (error) {
+    console.error('Error processing results:', error);
+    process.exit(1);
+  }
+}
+
+export { processResults, MatrixResult, ProcessedResults };
```
jaysoo: I don't think these are imported anywhere? I would remove them and then we can remove the `require.main` check since it's assumed this is a script to be run, and there are no exports.

.github/workflows/e2e-matrix.yml
ndcunningham: Added TODO also assigned a task for @xiongemi to follow-up

FrozenPandaz (approved):

PR #31427, opened 2025-06-02
feat(core): migrate to pnpm 10

ndcunningham:
- Added .pnpmrc for pnpm 10 configuration, enabling peer dependencies and lifecycle scripts.
- Updated package.json to reflect pnpm version change to 10.11.1 and added onlyBuiltDependencies.
- Update pipelines to reflect pnpm version update to 10.11.1

## Upgrading your pnpm version
Now to upgrade your `pnpm` version you can run `pnpm migrate-to-pnpm-version 10`. Which would upgrade your `pnpm` and it will run the upgrade script.

Later on if you want to upgrade to pnpm v11 you can run `pnpm migrate-to-pnpm-version 11`.
Additionally, if you just want to upgrade to the version that is inside of `package.json` you would run
`pnpm migrate-to-pnpm-version` without passing in a major version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
-          version: 9.8.0
+          version: 10.11.1
           run_install: false
 
       - uses: actions/setup-node@v4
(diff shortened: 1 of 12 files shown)


scripts/migrate-to-pnpm-10.js
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env node
+
+const { execSync } = require('child_process');
```
FrozenPandaz: Read the target version from `package.json` so that this can take it and do the rest.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #31444, opened 2025-06-03
fix(webpack): handle package.json exports field for non-buildable libs

Coly010:
Current Behavior
The webpack and rspack plugins for handling non-buildable libraries don't properly process the exports field in package.json. They incorrectly assume libraries have only a single entry point, typically through a barrel file (index.ts).

When a library defines multiple export paths using the exports field (e.g., "./*": "./src/*.ts"), the plugins fail to generate the correct allowlist patterns for webpack externals. This causes build failures when trying to use non-buildable libraries that expose multiple entry points without barrel files.

Expected Behavior
The webpack and rspack plugins should properly parse the exports field from package.json and generate appropriate allowlist patterns for all exported subpaths. This includes:

Handling wildcard patterns ("./*": "./src/*.ts")
Processing conditional exports (import/require/development)
Supporting exact subpath exports ("./utils": "./src/utils.ts")
Escaping special characters in package names for regex patterns
Gracefully falling back to reading package.json directly when metadata is unavailable

### DIFF
--- a/packages/rspack/src/plugins/utils/get-non-buildable-libs.spec.ts
+++ b/packages/rspack/src/plugins/utils/get-non-buildable-libs.spec.ts
@@ -1,0 +1,154 @@
+import { logger } from '@nx/devkit';
+import { createAllowlistFromExports } from './get-non-buildable-libs';
+
+describe('createAllowlistFromExports', () => {
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


Coly010:
@claude provide a good description for this PR

leosvelperez (approved):

PR #31455, opened 2025-06-04
fix(core): scope typescript resolution cache correctly when processing the graph

### LINKED ISSUE #31223: Subpath import aliases break Nx graph
### Current Behavior

In some of our projects, we'd like to use `package.json#imports` to alias private import paths as an alternative to relative path imports:
```jsonc
# package.json
  "imports": {
    "#*": "./src/*"
  }
```
```tsx
// SomeComponent.tsx
import { MyComponent } from '#components/MyComponent'
```

We stopped using this pattern because it can cause the Nx graph to mistakenly detect dependencies between projects that use subpath import aliases, sometimes even reporting circular dependencies for projects that don't depend on each other in any way. If this is due to some invalid config on our end, it is not clear how to resolve it.

### Expected Behavior

Projects should be able to leverage `package.json#imports` to alias imports within a project.

### GitHub Repo

https://github.com/laneysmith/nx-package-imports-bug-repro

### Steps to Reproduce

1. Clone https://github.com/laneysmith/nx-package-imports-bug-repro
2. `git checkout 37970a3` to checkout a working example without any `package.json#imports`
3. `nx reset && nx graph`
    - ✅ graph does not show any dependencies between `mylib` and `sidenav` 
    - Image
4. `git checkout main` to checkout changes that use `package.json#imports`
5. `nx reset && nx graph`
    - ❌ graph now shows incorrect dependency between `mylib` and `sidenav`
    - Image
    - Nx will also inconsistently report circular dependencies:
    - Image

### Nx Report

```shell

Node           : 22.12.0
OS             : darwin-arm64
pnpm           : 9.14.2

nx                     : 21.0.3
@nx/js                 : 21.0.3
@nx/eslint             : 21.0.3
@nx/workspace          : 21.0.3
@nx/devkit             : 21.0.3
@nx/eslint-plugin      : 21.0.3
@nx/module-federation  : 21.0.3
@nx/react              : 21.0.3
@nx/vite               : 21.0.3
@nx/web                : 21.0.3
Registered Plugins:
@nx/js/typescript
@nx/react/router-plugin
@nx/eslint/plugin
@nx/vite/plugin
```

### Failure Logs

```shell

```

### Package Manager Version

pnpm@9.15.4

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

leosvelperez:
## Current Behavior

When using Node.js subpath imports with the same name in different projects, the Nx graph incorrectly picks up seemingly random dependencies between projects that shouldn't exist.

This happens because the result of the resolution performed with TypeScript is cached using the import path as the cache key. The problem with that is that multiple projects can have the same subpath import name pointing to internal files of the project, so when the resolution is made for the first project (say `project1`), the result will be cached and incorrectly reused for other projects with the same subpath import name. So, all projects with the same subpath import name would resolve the dependency to the first project (`project1`).

The same could happen to projects with TS path mappings defined in the project's tsconfig file. These TS path mappings would only apply to the project internally and therefore, other unrelated projects could also define them with the same name pointing to different files.

## Expected Behavior

The Node.js subpath imports should be handled correctly. The TypeScript resolution result should be cached safely and scoped to the project from which the import is being done.

## Related Issue(s)

Fixes #31223

### DIFF
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
@@ -143,9 +143,8 @@ export class TargetProjectLocator {
     }
 
     if (this.tsConfig.config) {
-      // TODO(meeroslav): this block is probably obsolete
-      // and existed only because of the incomplete `paths` matching
-      // if import cannot be matched using tsconfig `paths` the compilation would fail anyway
+      // TODO: this can be removed once we rework resolveImportWithRequire below
+      // to properly handle ESM (exports, imports, conditions)
       const resolvedProject = this.resolveImportWithTypescript(
         importExpr,
         filePath
@@ -369,8 +368,15 @@ export class TargetProjectLocator {
     filePath: string
   ): string | undefined {
     let resolvedModule: string;
-    if (this.typescriptResolutionCache.has(normalizedImportExpr)) {
-      resolvedModule = this.typescriptResolutionCache.get(normalizedImportExpr);
+    const projectName = findProjectForPath(filePath, this.projectRootMappings);
+    const cacheScope = projectName
+      ? // fall back to the project name if the project root can't be determined
+        this.nodes[projectName]?.data?.root || projectName
+      : // fall back to the file path if the project can't be determined
+        filePath;
+    const cacheKey = `${normalizedImportExpr}__${cacheScope}`;
+    if (this.typescriptResolutionCache.has(cacheKey)) {
+      resolvedModule = this.typescriptResolutionCache.get(cacheKey);
     } else {
       resolvedModule = resolveModuleByImport(
         normalizedImportExpr,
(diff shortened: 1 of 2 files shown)


packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.ts
```diff
@@ -369,28 +368,58 @@ export class TargetProjectLocator {
     filePath: string
   ): string | undefined {
     let resolvedModule: string;
-    if (this.typescriptResolutionCache.has(normalizedImportExpr)) {
-      resolvedModule = this.typescriptResolutionCache.get(normalizedImportExpr);
+    let cacheKey: string;
+    if (normalizedImportExpr.startsWith('#')) {
```
FrozenPandaz: Is there a reason why this is only done for imports that start with `#`? I know there's a convention that frameworks set this but is there something special about # in the typescript spec?

leosvelperez: It's part of the [Node.js specification for subpath imports](https://nodejs.org/api/packages.html#subpath-imports). These imports are aliases that can only be used internally for the project where they are specified. So, they don't need to be unique across projects, and that's why we were getting the wrong cache hits and needed to special case their cache key to scope it to the project.

If your question is whether other imports without `#` could also be internal to a project, I guess it could happen if you have TS path mappings defined in the project's tsconfig file. In that case, those TS path mappings only apply to the project and therefore, you could have TS path mappings with the same name in other projects.

I'll update this to always generate the cache key after the project root. It's a safer approach.

FrozenPandaz (approved):

laneysmith:
Thank you, @leosvelperez & @FrozenPandaz ! 🙌

PR #31472, opened 2025-06-05
docs(core): add zephyr cloud to partners page

jordanpowell88:
## Expected Behavior
PR Adds Zephyr Cloud to Partners Page

## Related Issue(s)

Fixes #


nx-dev/ui-icons/src/lib/partners/zephry-cloud.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC, SVGProps } from 'react';
+
+export const ZephyrCloudIcon: FC<SVGProps<SVGSVGElement>> = (props) => {
```
Hardanish-Singh, line 3: nit: use inline return

Hardanish-Singh (approved):

barbados-clemens (approved):
small nit I noticed, seems like the partner list is arbitrarily listed. maybe we should move to making it alphabetical. but not something to change in this PR as list isn't too unruly yet. just wanted to call it out.

Issue #31454, opened 2025-06-04
New Versioning protocol incorrectly bumps downstream pinned versions


comp615:
### Current Behavior

Picture three package.jsons:

```
{
  "name": "@company/upstream",
  "version": "8.1.0"
}

{
  "name": "@company/downstream-pinned",
  "version": "0.0.1"
  "dependencies": {
    "@company/upstream": "7.3.4"
  }
}

{
  "name": "@company/downstream-workspace",
  "version": "0.0.1"
  "dependencies": {
    "@company/upstream": "workspace:*"
  }
}
```

Currently, when we make a change to `upstream`, it will replace the dependency in `@company/downstream-pinned` with 8.1.0.

### Expected Behavior

In the legacy versioning implementation, bumping upstream would result in no dependencies being incremented in the example above, and pnpm would handle things via the workspace using `preserveLocalDependencyProtocols`.

We have a specific usage for pinning dependencies that also exist in the local workspace and need a way to control this. It helps us migrate packages from outside the repo into the repo (we use syncpack to allow exceptions and work our way through aligning them to current).

### Steps to Reproduce

1. Create a repo where a package depends on the pinned version of another in-workspace package
2. Update that package
3. Compare results with both sets of nx.json settings (see below)

### Nx Report

```shell
Node           : 20.19.1
OS             : darwin-arm64
Native Target  : aarch64-macos
pnpm           : 10.11.0

nx                     : 21.2.0-beta.1
@nx/js                 : 21.1.2
@nx/jest               : 21.2.0-beta.1
@nx/eslint             : 21.2.0-beta.1
@nx/workspace          : 21.2.0-beta.1
@nx/cypress            : 21.2.0-beta.1
@nx/devkit             : 21.2.0-beta.1
@nx/eslint-plugin      : 21.2.0-beta.1
@nx/module-federation  : 21.2.0-beta.1
@nx/plugin             : 21.2.0-beta.1
@nx/react              : 21.2.0-beta.1
@nx/storybook          : 21.2.0-beta.1
@nx/vite               : 21.2.0-beta.1
@nx/web                : 21.2.0-beta.1
typescript             : 5.5.4

---------------------------------------
Nx key licensed packages
Licensed to Block Inc..

@nx/powerpack-s3-cache  : 2.0.3

---------------------------------------
Registered Plugins:
@nx/vite/plugin
@squareup/service-dashboard-web-runtime/nx-plugin
@nx/storybook/plugin
[...]

PR #31474, opened 2025-06-05
fix(js): WIP stop bumping pinned versions when proposed version bump does not satisfy current semver version

laneysmith:
## Current Behavior

In the legacy versioning logic, you could use workspace protocol and pinned versions for the same upstream dependency, and bumping the upstream package would only affect downstream projects that depended on it via workspace protocol:
```
{
  "name": "@company/upstream",
  "version": "8.1.0"
}

{
  "name": "@company/downstream-pinned",
  "version": "0.0.1"
  "dependencies": {
    "@company/upstream": "7.3.4" // <--- pinned
  }
}

{
  "name": "@company/downstream-workspace",
  "version": "0.0.1"
  "dependencies": {
    "@company/upstream": "workspace:^"
  }
}
```
The new versioning logic introduced a regression where the pinned version would be bumped to the latest version of the upstream package. So bumping `@company/upstream` to `8.2.0` replaces `"@company/upstream": "7.3.4"` with `"@company/upstream": "8.2.0"` in `@company/downstream-pinned`.

## Expected Behavior

Bumping `"@company/upstream"` to `8.2.0` should only affect `"@company/downstream-workspace"` and not `"@company/downstream-pinned"`.

In our Nx monorepo, we generally adhere to a single version policy, but we have a few exceptions where we need to pin versions during migrations. Given that Nx supports both single version policy and independently maintained dependencies, I'd expect that we could continue to use pinned versions in this scenario.

## Related Issue(s)

Fixes # https://github.com/nrwl/nx/issues/31454

### DIFF
--- a/packages/nx/src/command-line/release/version/release-group-processor.ts
+++ b/packages/nx/src/command-line/release/version/release-group-processor.ts
@@ -817,7 +817,7 @@ export class ReleaseGroupProcessor {
               this.bumpedProjects.add(project);
 
               // Update any dependencies in the manifest
-              await this.updateDependenciesForProject(project);
+              await this.updateDependenciesForProject(project, 'fixed');
             }
           }
         }
@@ -836,7 +836,7 @@ export class ReleaseGroupProcessor {
             );
             // Ensure the bump for remaining projects
             this.bumpedProjects.add(project);
-            await this.updateDependenciesForProject(project);
+            await this.updateDependenciesForProject(project, 'fixed');
           }
         }
       } else {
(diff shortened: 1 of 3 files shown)


packages/nx/src/command-line/release/version/release-group-processor.ts
```diff
@@ -1332,6 +1332,17 @@ Valid values are: ${validReleaseVersionPrefixes
           if (!currentDependencyVersion) {
             continue;
           }
+          const targetMatchesRequirement =
```
laneysmith: Adapted from [this line](https://github.com/nrwl/nx/blob/33bfc51ec2ccb5208760f86f5d4f45886a2e7a64/packages/js/src/generators/release-version/utils/resolve-local-package-dependencies.ts#L101-L104C9) in `release-version/utils/resolve-local-package-dependencies.ts`.

PR #31480, opened 2025-06-05
fix(core): disable TUI on ai agents

FrozenPandaz:
## Current Behavior

The TUI (Terminal User Interface) is enabled for all environments, including AI agents like Claude Code, which can cause issues with AI-driven development workflows.

## Expected Behavior

When an AI agent is detected through environment variables, the TUI should be automatically disabled to prevent interference with AI-driven interactions.

## Related Issue(s)

This change improves the developer experience when using AI agents by automatically detecting their presence and adjusting the interface accordingly.


packages/nx/src/native/utils/ai.rs
```diff
@@ -0,0 +1,76 @@
...
+    let is_ai = !detected_vars.is_empty();
+
+    if is_ai {
+        debug!("AI agent detected with variables: {:?}", detected_vars);
+    } else {
+        debug!("No AI agent environment variables found");
+    }
+
+    is_ai
+}
```
Cammisuli:
```suggestion
pub fn is_ai_agent() -> bool {
    debug!("Checking for AI agent environment variables");

    let is_ai = AI_ENV_VARS.iter().any(|var| match env::var(var) {
        Ok(_) => {
            debug!("Found AI environment variable: {}", var);
            true
        }
        Err(_) => {
            debug!("AI environment variable not set: {}", var);
            false
        }
    });

    if is_ai {
        debug!("AI agent detected");
    } else {
        debug!("No AI agent environment variables found");
    }

    is_ai
}
```

Cammisuli (dismissed):

packages/nx/src/native/utils/ai.rs
```diff
@@ -0,0 +1,73 @@
...
+            // Restore original state
+            match original {
+                Some(orig_val) => unsafe { env::set_var(var, orig_val) },
+                None => unsafe { env::remove_var(var) },
+            }
+            assert!(
+                !is_ai_agent(),
+                "Should not detect AI after resetting {}",
+                var
+            );
```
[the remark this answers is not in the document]
FrozenPandaz: @claude can you help me implement this please?

packages/nx/src/native/utils/ai.rs
FrozenPandaz: @claude Can you help me implement this?

packages/nx/src/native/utils/ai.rs
```diff
@@ -0,0 +1,81 @@
+use std::env;
+use tracing::debug;
+
+const AI_ENV_VARS: &[&str] = &[
+    "CLAUDECODE",
+    "CURSOR_TRACE_ID",
```
FrozenPandaz: @claude this environment does not work because it is set in user's terminals when they are using Cursor as well. I don't want the TUI to be disabled in the user's terminal. I only want the TUI to be disabled when Cursor's AI Agent is running Nx.

AgentEnder (approved):

PR #31482, opened 2025-06-05
fix(core): improve local registry setup and make it more flexible

FrozenPandaz:
## Current Behavior

The E2E global setup hardcodes local registry configuration and waits for a specific localhost registry to be available, which can cause issues when the registry setup is managed differently or when using external registries.

## Expected Behavior

The E2E setup should be more flexible and work with different registry configurations:
- Wait for the registry configuration to change from the default instead of assuming local registry
- Properly manage local registry processes with cleanup
- Work with both local and external registries
- Provide better error handling and logging

## Related Issue(s)

This PR improves the E2E local testing setup to be more robust and flexible.

## Changes Made

- Changed from hardcoded local registry waiting to detecting registry configuration changes
- Added proper process management for local registry startup
- Improved error handling and cleanup with process termination  
- Made registry configuration work with both local and external registries
- Refactored code into smaller, focused functions for better maintainability
- Added better logging for debugging setup issues

### DIFF
--- a/e2e/utils/global-setup.ts
+++ b/e2e/utils/global-setup.ts
@@ -38,7 +38,7 @@ export default async function (globalConfig: Config.ConfigGlobals) {
       // Keep checking until the configured registry changes from default
       while (currentRegistry === defaultRegistry) {
         console.log(
-          'Waiting for registry configuration to change from default...'
+          `Waiting for registry configuration to change from default (${defaultRegistry})...`
         );
         await new Promise((resolve) => setTimeout(resolve, 250));
         currentRegistry = getCurrentRegistry();
--- a/e2e/utils/global-setup.ts
+++ b/e2e/utils/global-setup.ts
@@ -38,7 +38,7 @@ export default async function (globalConfig: Config.ConfigGlobals) {
       // Keep checking until the configured registry changes from default
       while (currentRegistry === defaultRegistry) {
         console.log(
-          `Waiting for registry configuration to change from default (${defaultRegistry})...`
+          `Waiting for registry configuration to change from ${defaultRegistry}...`
         );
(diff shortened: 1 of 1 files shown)


e2e/utils/global-setup.ts
```diff
@@ -20,46 +22,38 @@ export default async function (globalConfig: Config.ConfigGlobals) {
...
+    if (currentRegistry === defaultRegistry) {
+      console.log('Default registry detected, starting local registry...');
+
+      // Start the local registry
+      localRegistryProcess = await startLocalRegistry(isVerbose);
+
+      // Keep checking until the configured registry changes from default
+      while (currentRegistry === defaultRegistry) {
+        console.log(
+          'Waiting for registry configuration to change from default...'
```
JamesHenry: Maybe include what the default is in the log to avoid any confusion

FrozenPandaz: @claude can you include the default in the log please?

FrozenPandaz: @claude Instead of printing `default (https://registry.npmjs.org/)`, just print the default URL.

e2e/utils/global-setup.ts
```diff
@@ -96,18 +87,123 @@ export default async function (globalConfig: Config.ConfigGlobals) {
...
+      hostname: url.hostname,
+      port: url.port || (url.protocol === 'https:' ? '443' : '80'),
+    };
+  } catch {
+    // Fallback for invalid URLs
+    return {
+      hostname: 'localhost',
+      port: '4873',
+    };
+  }
```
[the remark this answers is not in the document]
FrozenPandaz: It's okay, the registry URL will for sure have the protocol.

PR #31492, opened 2025-06-08
fix(core): add secondary entry point 'nx/release'

RobbyRabbitman:
## Current Behavior
Docs say that you should use the programmatic nx release api from `nx/release` - see [docs](https://nx.dev/features/manage-releases#using-the-programmatic-api-for-nx-release)

## Expected Behavior
The package.json of the `nx` package should expose the nx release api via a secondary entry point to match the docs or adjust the docs if `nx/release` is not desired.

## Related Issue(s)

Fixes #


RobbyRabbitman:
The build task has changed apparently in the meantime, see https://github.com/nrwl/nx/pull/31654. The old build task had this `additionalEntryPoints` option, so I tried something else in the PR - from what I can see: projects know use a standard tsc build task and just define packageJson.exports, however some projects apparently rely on some .ts files during build time - idk, but I dont get the build pipeline at all in this repo. So i added those exports in the package.json which is not ideal, since they shouldnt be present when the package gets published. There is apparently a _legacy post build_ task which can do stuff, maybe hardcode there a solution for the nx package, since this package seems to be special compared to the rest. Also what i dont get: there is a dist and build folder and there are some slight difference in the package.json files. Can someone help pls?

JamesHenry:
You just need to apply the relevant extension in your case, adding exports to the nx package does _way_ more than address your issue and is not something to be done lightly

RobbyRabbitman:
ahh nice! Thank you for explaining, my bad for PR.

PR #31508, opened 2025-06-09
fix(js): recognize NodeNext as ESM

JonathanWilbur:
## Current Behavior
NodeNext is not recognized as ESM. This causes this warning message to be logged, even when you have `"type": "module",` in the `package.json` file and are compile TypeScript to `"module": "NodeNext"`.

```
Package type is set to "module" but "cjs" format is included. Going to use "esm" format instead. You can change the package type to "commonjs" or remove type in the package.json file.
```

## Expected Behavior
Don't log this message. It is incorrect.


Coly010:
I wonder if there is, or if there needs to be, a further check to verify that the adjacent package.json contains `type:module` before we set it as esm

mandarini:
@Coly010 let's remove me as the reviewer 😅 hehe I have enough `supabase-js` PRs to review 😝

Coly010 (approved):

PR #31513, opened 2025-06-09
docs(core): update Nx installation steps to prefer global installs

jaysoo:
This PR makes the global install more prominent. This is important for non-JS workspaces since they may not have node or npm available. The installation flow should be:

1. Install Nx (e.g. `brew install nx`)
2. Run `nx init` in your project

The `create-nx-workspace` is removed since it's already mentioned on the intro page, and it is covered in tutorials. It is also not technically an "install" command.

Preview: https://nx-dev-git-docs-update-installation-nrwl.vercel.app/getting-started/installation

**Note:** The cross-platform NPM install is still featured first because it'll work for everyone. It shouldn't be a concern for non-JS devs since other tools like Claude Code also has `npm install --global` as their install step.

### DIFF
--- a/docs/shared/getting-started/installation.md
+++ b/docs/shared/getting-started/installation.md
@@ -5,146 +5,73 @@ description: Learn how to install Nx in a new workspace, add it to an existing r
 
 # Installation
 
-Create a new Nx workspace using the following command:
+To install Nx on your machine, choose one of the following methods based on your operating system and package manager. You can also use `npx` to run Nx without installing it globally.
 
 {% tabs %}
 {% tab label="npm" %}
 
 ```shell
-npx create-nx-workspace
+npm add --global nx
 ```
+
+**Note:** You can also use yarn, pnpm, or bun
 
 {% /tab %}
-{% tab label="yarn" %}
-
-```shell
-npx create-nx-workspace --pm yarn
-```
-
-{% /tab %}
-{% tab label="pnpm" %}
-
-```shell
-npx create-nx-workspace --pm pnpm
-```
-
-{% /tab %}
-{% tab label="Bun" %}
-
-```shell
-bunx create-nx-workspace --pm bun
-```
-
-{% /tab %}
-{% /tabs %}
-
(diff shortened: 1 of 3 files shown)


docs/shared/getting-started/installation.md
```diff
@@ -7,144 +7,64 @@ description: Learn how to install Nx in a new workspace, add it to an existing r
 
 Create a new Nx workspace using the following command:
```
juristr: I think we should remove this line. Probably a left-over.

docs/shared/getting-started/installation.md
```diff
@@ -7,144 +7,64 @@ description: Learn how to install Nx in a new workspace, add it to an existing r
...
-```shell
-npx create-nx-workspace --pm pnpm
-```
-
-{% /tab %}
-{% tab label="Bun" %}
-
-```shell
-bunx create-nx-workspace --pm bun
+# You can also use yarn, pnpm, or bun
```
juristr: We can move this outside the `shell` block, just as a normal text within the tab. Otherwise the rendering looks weird

Screenshot 2025-06-10 at 11 21 06

docs/shared/getting-started/installation.md
```diff
@@ -7,144 +7,64 @@ description: Learn how to install Nx in a new workspace, add it to an existing r
...
+npm add --global nx
 ```
 
 {% /tab %}
-{% /tabs %}
-
-This will guide you through the setup, asking whether you want a monorepo or a standalone app and whether you want to start with a blank or a preconfigured setup.
-
-```{% command="npx create-nx-workspace@latest" path="~" %}
```
juristr, line 43: did you remove this on purpose? I agree that it shouldn't be as prominent (like showing log output etc), but should we merge it with the "nx init" one? Like maybe two tabs or just a sentence + shell box for adding to an existing and right after the command for creating a new workspace?

juristr (changes requested):

juristr (approved):

PR #31533, opened 2025-06-10
fix(js): improve typescript plugin build detection

### LINKED ISSUE #29670: TypeScript plugin doesn't find buildable project unless all package.json exports point to built source files
### Current Behavior

With [this change](https://github.com/nrwl/nx/pull/29431/files#diff-915b2db94922f651a865b86277057f88f023fe54d5617808f1688248770b1ecfR602-R607) that landed in v20.3.2, the TypeScript plugin no longer infers a build target if the package.json `exports` field exposes any module that isn't located within the resolved TypeScript `outDir`.

### Expected Behavior

I've got JavaScript modules in the project that live outside the outDir and aren't part of the TypeScript build that I'd like to export in addition to built files. From the code comment, it sounds like the current behavior is intentional, but I'm not sure how I can proceed with my intention. Thanks!

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create a module outside the TypeScript outDir, i.e. a file that is not part of the TypeScript build. For example, if I've got TypeScript configured to build everything in the `src` directory in my project, add a file like `test.js` in the root of the project.
2. Add that file to package.json `exports`.
3. Observe the TypeScript plugin no longer infers a `build` target.

### Nx Report

```shell
Node           : 20.18.0
OS             : darwin-arm64
yarn           : 4.5.1

nx             : 20.3.2
@nx/js         : 20.3.2
@nx/jest       : 20.3.2
@nx/workspace  : 20.3.2
@nx/devkit     : 20.3.2
typescript     : 5.5.4
Registered Plugins:
@nx/js/typescript
@nx/jest/plugin
@nx/jest/plugin
@nx/jest/plugin
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

ndcunningham:
This PR enhances the typescript plugin build detection. Notably the changes include:
- Add absolute path handling for package.json entry points.
- Simplify the buildable verification
- Add glob pattern matching for `include` patterns.

MISC: 
- Also fixes `e2e-rollup` test failing due to the typescript plugin having the same build target as rollup so it created an unexpected output.

closes: #29670

### DIFF
--- a/packages/js/src/plugins/typescript/plugin.spec.ts
+++ b/packages/js/src/plugins/typescript/plugin.spec.ts
@@ -95,8 +95,8 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
                     "^typecheck",
                   ],
                   "inputs": [
-                    "production",
-                    "^production",
+                  "production",
+                  "^production",
                     {
                       "externalDependencies": [
                         "typescript",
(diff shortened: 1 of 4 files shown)


packages/js/src/plugins/typescript/plugin.spec.ts
```diff
@@ -5219,6 +5219,791 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
...
+            "libs/my-lib": {
+              "projectType": "library",
+              "targets": {},
+            },
+          },
+        }
+      `);
+    });
+
+    it('should handle mixed path scenarios with exports field', async () => {
```
jaysoo: I don't think we realistically can have this, we should just not infer `build` I think because it's likely not a setup we support.

packages/js/src/plugins/typescript/plugin.spec.ts
```diff
@@ -5219,6 +5219,791 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
...
+      `);
+    });
+
+    it('should handle allowJs option and consider JavaScript files as source files', async () => {
+      await applyFilesToTempFsAndContext(tempFs, context, {
+        'libs/my-lib/tsconfig.json': '{}',
+        'libs/my-lib/tsconfig.lib.json': JSON.stringify({
+          compilerOptions: {
+            outDir: 'dist',
+            allowJs: true,
```
jaysoo: I don't think this needed, if the `.js` file should be considered source, then the source must exist in the repo.

For the default behavior when `outDir` is not set, tsc would create `.js` sibling files, like

```
src/
  index.ts
  index.js
```

and in this case, pointing to `src/index.js` means it's not source, thus must be buildable.

packages/js/src/plugins/typescript/util.ts
```diff
@@ -77,22 +78,52 @@ export function isValidPackageJsonBuildConfig(
...
-    return ['.ts', '.tsx', '.cts', '.mts'].includes(ext);
+    // If no include patterns, TypeScript includes all TS files by default
+    const include = tsConfig.raw?.include;
+    if (!include || !Array.isArray(include)) {
+      const ext = extname(path);
+      const tsExtensions = ['.ts', '.tsx', '.cts', '.mts'];
+      if (tsExtensions.includes(ext)) {
+        return true;
+      }
+    }
```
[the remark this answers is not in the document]
ndcunningham: no should not be added

jaysoo (approved):

fregante:
Thank you! Will try it out when it's published

PR #31545, opened 2025-06-11
feat(js): add copy-workspace-modules executor

Coly010:
## Current Behavior

When building applications that depend on workspace libraries for deployment (particularly in containerized environments like Docker), developers must manually handle copying workspace dependencies and updating package.json references. 

This creates friction when trying to deploy applications that consume workspace libraries, as the build output doesn't contain the necessary workspace dependencies and the package.json still references them with `workspace:` protocol which doesn't work outside the workspace context.

  ## Expected Behavior

  With the new `@nx/js:copy-workspace-modules` executor, developers can automatically prepare their built applications for deployment by:

  1. **Automatically copying workspace dependencies**: The executor scans the application's package.json for workspace dependencies (those with `workspace:` or `file:` version specifiers) and copies the source code of these dependencies into a `workspace_modules` directory within the build output

### DIFF
--- a/e2e/js/src/js-executor-copy-workspace-modules.test.ts
+++ b/e2e/js/src/js-executor-copy-workspace-modules.test.ts
@@ -53,7 +53,10 @@ describe('@nx/js:copy-workspace-modules', () => {
 }`
     );
 
-    updateFile(`${nodelib}/src/index.ts`, `export * from './lib/${nodelib}';`);
+    updateFile(
+      `${nodelib}/src/index.ts`,
+      `export * from './lib/${nodelib}.js';`
+    );
 
     // Add workspace dependency to the app's package.json
     updateJson(`${nodeapp}/package.json`, (json) => {
@@ -63,6 +66,7 @@ describe('@nx/js:copy-workspace-modules', () => {
       };
       return json;
     });
+    runCommand(`pnpm install`);
 
     // Update the app to use the library
     updateFile(
@@ -94,8 +98,6 @@ console.log(${nodelib}());`
       };
       return json;
     });
-
-    runCommand(`pnpm install`);
 
     // Run the copy-workspace-modules executor
     const result = runCLI(`run ${nodeapp}:copy-workspace-modules`);
@@ -137,7 +139,7 @@ console.log(${nodelib}());`
     );
     updateFile(
       `${nodelib1}/src/index.ts`,
-      `export * from './lib/${nodelib1}';`
+      `export * from './lib/${nodelib1}.js';`
     );
 
     updateFile(
@@ -148,7 +150,7 @@ console.log(${nodelib}());`
     );
     updateFile(
       `${nodelib2}/src/index.ts`,
-      `export * from './lib/${nodelib2}';`
+      `export * from './lib/${nodelib2}.js';`
     );
 
     // Add workspace dependencies to the app's package.json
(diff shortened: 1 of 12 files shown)


e2e/js/src/js-executor-copy-workspace-modules.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  let scope: string;
+  let previousPackageManager: string;
+
+  beforeAll(() => {
+    previousPackageManager = process.env.SELECTED_PM;
+    // Use pnpm to properly test workspace: protocol dependencies
+    process.env.SELECTED_PM = 'pnpm';
+
+    scope = newProject({
+      packages: ['@nx/node', '@nx/js'],
```
jaysoo, line 19: I think package manager option should work without setting `SELECTED_PM?

packages/js/src/executors/copy-workspace-modules/copy-workspace-modules.ts
```diff
@@ -0,0 +1,152 @@
...
+  projectGraph: ProjectGraph
+) {
+  if (!packageJson.dependencies) {
+    return;
+  }
+
+  for (const [pkgName, pkgVersion] of Object.entries(
+    packageJson.dependencies
+  )) {
+    if (pkgVersion.startsWith('workspace:') || pkgVersion.startsWith('file:')) {
```
jaysoo: What if users use `*`? Considering NPM doesn't support `workspace:` I think it's down to `*` or `file:`. Or perhaps the version is valid semver, e.g. `1.0.0`, and it just matches what's in the monorepo.

packages/js/src/executors/copy-workspace-modules/copy-workspace-modules.ts
```diff
@@ -0,0 +1,152 @@
...
+  readJsonFile,
+  workspaceRoot,
+} from '@nx/devkit';
+import { interpolate } from 'nx/src/tasks-runner/utils';
+import { type CopyWorkspaceModulesOptions } from './schema';
+import { cpSync, existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import { dirname, join } from 'path';
+import { lstatSync } from 'fs';
+
+export default async function copyWorkspaceModules(
```
jaysoo, line 16: After the workspace module is copied, there needs to be something outside of the this executor to make sure they are transpiled, right? If the source is `.ts`, then it either needs `ts-node`/`tsx`/etc. to transpile it at runtime, or preferably the build step would do this, like esbuild does.

packages/js/src/executors/copy-workspace-modules/copy-workspace-modules.ts
Coly010: If valid semver is acceptable in npm - the design of this will need to change somewhat 🤔

packages/js/src/executors/copy-workspace-modules/copy-workspace-modules.ts
Coly010: Yes. It's a separate challenge I think. 
If `copy-workspace-modules` is set to depend on `build` and `build` depends on `^build` then we can safely assume that buildable workspace libraries will be transpiled when they're copied across.

The challenge comes with non buildable workspace libraries. I think it requires more thought and not necessarily within the scope of this executor in particular.

jaysoo (approved):

PR #31550, opened 2025-06-11
fix(core): complete child processes when stdout includes readyWhen

jordanpowell88:
## Current Behavior
Currently in version 21 a command using `readyWhen` will never exit

**Steps to reproduce:**

Clone this repo => [https://github.com/jordanpowell88/readyWhen-bug](https://github.com/jordanpowell88/readyWhen-bug)

`run nx serve core`

The serve target `dependsOn` `watch-tailwind` which never exits even after it echos "Done".

This works in version 20 but stopped working in 21. If you upgrade this project to 21 it will return the following error: 

```Warning: command "npx tailwindcss -i ./apps/core/src/styles.css -o ./dist/apps/core/styles.css --watch" exited with non-zero status code```

## Expected Behavior
When the `stdout` of the run command includes the string from `options.readyWhen` the child process should complete

## Related Issue(s)

Fixes #

### CODE REVIEW

File: packages/nx/src/tasks-runner/pseudo-terminal.ts [hunk 1]
```diff
@@ -67,12 +68,14 @@ export class PseudoTerminal {
...
       quiet,
       tty,
+      readyWhenStatus,
     }: {
       cwd?: string;
       execArgv?: string[];
       jsEnv?: Record<string, string>;
       quiet?: boolean;
       tty?: boolean;
+      readyWhenStatus?: RunCommandsOptions['readyWhenStatus'];
```
meeroslav: RunCommandsOptions does not contain `readyWhenStatus`. If it should contain it, please extend the interface in `run-commands.impl.ts`.

packages/nx/src/tasks-runner/pseudo-terminal.ts
```diff
@@ -151,20 +156,37 @@ export class PseudoTtyProcess implements RunningTask {
 
   constructor(
     public rustPseudoTerminal: RustPseudoTerminal,
-    private childProcess: ChildProcess
+    private childProcess: ChildProcess,
+    private readyWhenStatus?: RunCommandsOptions['readyWhenStatus']
```
meeroslav: Same as above, ensure the `RunCommandsOptions` has the necessary property defined in the interface.

meeroslav (changes requested):
Left one minor comment regarding the `RunCommandsOptions` interface

packages/nx/src/tasks-runner/pseudo-terminal.ts
```diff
@@ -5,6 +5,7 @@ import { ChildProcess, IS_WASM, RustPseudoTerminal } from '../native';
 import { PseudoIPCServer } from './pseudo-ipc';
 import { RunningTask } from './running-tasks/running-task';
 import { codeToSignal } from '../utils/exit-codes';
+import { NormalizedRunCommandsOptions } from '../executors/run-commands/run-commands.impl';
```
leosvelperez: This introduces a circular dep. `run-commands.impl` already imports from this file. We can either extract the type to a shared file or duplicate it here.

packages/nx/src/tasks-runner/pseudo-terminal.ts
```diff
@@ -151,20 +156,37 @@ export class PseudoTtyProcess implements RunningTask {
...
+      this.childProcess.onExit((message) => {
+        this.isAlive = false;
 
-      const code = messageToCode(message);
-      childProcess.cleanup();
+        const code = messageToCode(message);
+        this.childProcess.cleanup();
 
-      this.exitCallbacks.forEach((cb) => cb(code));
+        this.exitCallbacks.forEach((cb) => cb(code));
```
leosvelperez: We also need to resolve the promise here. This is needed in cases where the process exits before the readyWhen condition is met.

packages/nx/src/tasks-runner/pseudo-terminal.ts
```diff
@@ -202,6 +224,18 @@ export class PseudoTtyProcess implements RunningTask {
   }
 }
 
+function isReady(readyWhenStatus, output) {
```
leosvelperez: Let's type the arguments.

leosvelperez (changes requested):

packages/nx/src/tasks-runner/pseudo-terminal.ts
```diff
@@ -143,6 +151,7 @@ export class PseudoTerminal {
 
 export class PseudoTtyProcess implements RunningTask {
   isAlive = true;
+  private initChildProcessesPromise: Promise<any>;
```
leosvelperez: Sorry, I missed this in the previous review: this `Promise` shouldn't be needed. We're storing it but not awaiting or using it. We can set up the listeners right away.

leosvelperez (approved):

PR #31557, opened 2025-06-12
feat(js): add prune-lockfile executor

Coly010:
## Current Behavior
In the new TS Soln Setup with Package Manager Workspaces, there is currently no builtin method for creating a pruned lockfile that is agnostic across bundlers and compilers.
This is problematic for apps that must be containerized.

## Expected Behavior
Add a new `@nx/js:prune-lockfile` executor that can be used to produce a pruned lockfile in the build artifact output directory, ready for containerization.


e2e/js/src/js-executor-prune-lockfile.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        };
+        json.nx.targets['prune-lockfile'] = {
+          executor: '@nx/js:prune-lockfile',
+          options: {
+            buildTarget: 'build',
+          },
+        };
+        return json;
+      });
+      console.log(tmpProjPath());
```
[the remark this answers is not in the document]
Coly010: @claude Can you remove this unneeded console.log

e2e/js/src/js-executor-prune-lockfile.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  describe.each([
+    ['pnpm', 'pnpm-lock.yaml'],
+    ['yarn', 'yarn.lock'],
+    ['npm', 'package-lock.json'],
+  ])('package manager %s', (packageManager, lockfile) => {
+    let scope;
+    let previousPackageManager: string;
+
+    beforeAll(() => {
+      previousPackageManager = process.env.SELECTED_PM;
```
jaysoo: `newProject` has a package manager argument. Does it not work here?

packages/js/src/executors/prune-lockfile/prune-lockfile.ts
```diff
@@ -0,0 +1,130 @@
...
+    packageJson,
+    context.projectGraph
+  );
+  writeFileSync(join(outputDirectory, lockfileName), lockFile);
+  writeFileSync(
+    join(outputDirectory, 'package.json'),
+    JSON.stringify(packageJson, null, 2)
+  );
+  logger.log('Lockfile pruned.');
+  logger.log('Success!');
```
jaysoo: Minor: I think prune take that long, so reducing the log to just one at the end, like `Lockfile pruned: ${output}` makes more sense.

packages/nx/src/plugins/js/lock-file/npm-parser.ts
```diff
@@ -407,21 +409,58 @@ export function stringifyNpmLockfile(
...
 
+function mapWorkspaceModules(
+  packageJson: NormalizedPackageJson,
+  rootLockFile: NpmLockFile
+) {
+  const output: Record<string, NpmDependencyV3 & NpmDependencyV1> = {};
+  for (const [pkgName, pkgVersion] of Object.entries(
+    packageJson.dependencies
+  )) {
+    if (pkgVersion.startsWith('workspace:') || pkgVersion.startsWith('file:')) {
```
jaysoo: For npm, only `file:` or `*` would work for workspace modules right? `workspace:` is pnpm/yarn.

Coly010: This might take more thought for npm. 
`*` is also valid for non workspace packages too right? 
So we're likely going to have to identify workspace packages initially through a different manner rather than just the pkgVersion.

jaysoo (approved):

JamesHenry (approved):

PR #31579, opened 2025-06-13
Parallel with fraction number

ianzone:
Before:
parallel accept 3, "3" or "30%"
After:
parallel accept 3, "3", "30%", 0.3 or "0.3"


ianzone:
Hi, can someone give it a look?

ianzone:
Hi, it's been a while. Could this be merged?

ianzone:
The error log seems irrelevant to my commits. Is that something I need to fix?

MaxKless:
Hey @ianzone! Thanks for the effort in improving Nx. 
As we talked about in your last PR, we prefer the %-based syntax because it's more in line with what other tools do.
Having a fractional number doesn't really add anything beyond that and we don't think there's an upside to supporting multiple different ways of doing the same thing - it will just be more confusing and harder to maintain.

I'll close this PR for now but if you have a clear use case that specifying `0.4` vs `40%` unlocks, I'd love to hear about it.

Thanks!

ianzone:
> Hey @ianzone! Thanks for the effort in improving Nx. As we talked about in your last PR, we prefer the %-based syntax because it's more in line with what other tools do. Having a fractional number doesn't really add anything beyond that and we don't think there's an upside to supporting multiple different ways of doing the same thing - it will just be more confusing and harder to maintain.
> 
> I'll close this PR for now but if you have a clear use case that specifying `0.4` vs `40%` unlocks, I'd love to hear about it.
> 
> Thanks!

A real customer walks into the bar and asks where the bathroom is. The bar goes up in flames. 

30% equals 0.3. It's intuitive, thus people will use it, and it would be wrong to result in one thread. 

Another upside, 0.3 presses fewer keys than 30%.

ianzone:
I don't think it's a different syntax since 0.3 and 30% are all arithmetic expressions. 

"thirty percent" and 30% are different ways, though.

ianzone:
Hi @MaxKless , are the reasons good enough?

PR #31596, opened 2025-06-14
fix(core): nx.json should not be required to run tasks

juristr:
## Current Behavior

You get a " NX   Could not find Nx modules in this workspace." error when there's no "nx.json"

## Expected Behavior

It should just work. Nx should still be able to run tasks.

## Related Issue(s)

Fixes #


juristr:
The implementation works as expected. There was an issue on my own machine (e.g. a lost `nx.json` higher up in a parent folder) 🤦‍♂️

PR #31622, opened 2025-06-17
feat(js): support publishing with registryConfigKey when pnpm >=9.15.7 <10.0.0 || >=10.5.0

jogelin:
### Description

This PR adds support for using a custom scoped registry with `pnpm` during the release process.

### Problem

When using `pnpm < v9.15.7`, and your workspace `.npmrc` includes a scoped registry config like:

```ini
@my-orga:registry=https://npm.pkg.my-orga.io/
//npm.pkg.my-orga.io/:_auth=${MY_NPM_APIKEY}
```

...attempting to override the registry dynamically (e.g., to use a local Verdaccio instance) like this:

```ts
await releasePublish({
  firstRelease: true,
  registry: 'http://localhost:4873',
});
```

...results in the following command:

```bash
pnpm publish "/Users/user/my-orga/packages/library" --json --registry="http://localhost:4873/" --tag=latest --no-git-checks
```

However, due to limitations documented in the [Nx release-publish executor](https://github.com/nrwl/nx/blob/d0d62846a24ffdb4be8506be49c5d4d68503c601/packages/js/src/executors/release-publish/release-publish.impl.ts#L272) and the [pnpm issue #9084](https://github.com/pnpm/pnpm/issues/9084), **`pnpm publish` does not allow overriding scoped registries via `--registry` CLI option**.

### Solution

Starting with `pnpm@10.5` and [patched on `pnpm@9.15.7`](https://github.com/pnpm/pnpm/releases/tag/v9.15.7), it’s possible to override a scoped registry via CLI using:

```bash
--@my-orga:registry=http://localhost:4873/
```

This PR updates the release process to support this by injecting `--@scope:registry` for recent pnpm versions like:
```bash
pnpm publish "/Users/user/org/packages/library" --json --@my-orga:registry="http://localhost:4873/" --tag=latest --no-git-checks
```

### DIFF
--- a/packages/nx/src/utils/package-manager.ts
+++ b/packages/nx/src/utils/package-manager.ts
@@ -11,7 +11,7 @@ import {
 } from 'yaml';
 import { rm } from 'node:fs/promises';
 import { dirname, join, relative } from 'path';
-import { gte, lt, parse } from 'semver';
+import { gte, lt, parse, satisfies } from 'semver';
 import { dirSync } from 'tmp';
 import { promisify } from 'util';
 
@@ -44,6 +44,12 @@ export interface PackageManagerCommands {
   run: (script: string, args?: string) => string;
   // Make this required once bun adds programatically support for reading config https://github.com/oven-sh/bun/issues/7140
   getRegistryUrl?: string;
+  publish: (
+    packageRoot: string,
+    registry: string,
+    registryConfigKey: string,
+    tag: string
+  ) => string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mpsanchis:
🔥

packages/nx/src/utils/package-manager.ts
```diff
@@ -130,17 +136,26 @@ export function getPackageManagerCommand(
...
+      let modernPnpm: boolean,
+        includeDoubleDashBeforeArgs: boolean,
+        allowRegistryConfigKey: boolean;
       try {
         const pnpmVersion = getPackageManagerVersion('pnpm', root);
         modernPnpm = gte(pnpmVersion, '6.13.0');
         includeDoubleDashBeforeArgs = lt(pnpmVersion, '7.0.0');
+        allowRegistryConfigKey = satisfies(
+          pnpmVersion,
+          '>=9.15.7 <10.0.0 || >=10.5.0'
```
JamesHenry, line 155: What happened here? `>= 10.0.0 < 10.5.0`?

jogelin: So they [fixed it initially on `v10.5`](https://github.com/pnpm/pnpm/releases?page=3) on the 25 of February
image

[then patched `v9.15.7`](https://github.com/pnpm/pnpm/releases/tag/v9.15.7) on the 6th of March
image

So all versions `>= 10.0.0 < 10.5.0` are not containing the fix

JamesHenry: Oh interesting, so it was actually a v10 feature, but they chose to back port it to the latest v9 patch...

Please can we add that as a comment to clarify for posterity? Otherwise LGTM

jogelin: Comment added

JamesHenry (approved):

PR #31626, opened 2025-06-17
feat(graph): enhance migration state management and introduce migration stopping functionality.

ndcunningham:
## Overview
This PR enhances the Migrate UI by adding a "Stop" button that allows users to halt in-progress migrations at any point during execution.
It works in tandem with https://github.com/nrwl/nx-console/pull/2567

### Currently
When a migration is running and needs to be stopped for any reason:
- User accidentally triggered the migration
- Migration is taking longer than expected
- Changes are needed before completion

Users must wait for the migration to complete before using the available "Undo" or "Skip" options.

### Expected
Users should be able to stop a currently running migration at any time before it completes, providing immediate control over the migration process.

### Key Features:

- Refactor guards to improve migration state checks and add conditions for running and completing migrations.
- Update the state machine to handle new states for running, stopped, and evaluating migrations.
- Implement logic to track running migrations and allow for stopping them gracefully.
- Introduce a new process for running migrations in a separate child process to support cancellation.
- Enhance metadata management to include stopped migrations and update UI accordingly. (The UI is completely driven by the backend now aka Nx Console)
- Add tests to cover new migration states and behaviours.

### DIFF
--- a/graph/migrate/src/lib/components/automatic-migration.tsx
+++ b/graph/migrate/src/lib/components/automatic-migration.tsx
@@ -8,6 +8,7 @@ import {
   currentMigrationHasChanges,
   currentMigrationHasFailed,
   currentMigrationHasSucceeded,
+  currentMigrationIsStopped,
 } from '../state/automatic/selectors';
 import { MigrationTimeline } from './migration-timeline';
 import { Interpreter } from 'xstate';
@@ -45,15 +46,13 @@ export function AutomaticMigration(props: {
     (migration) => migration.id === currentMigration?.id
   );
 
-  const currentMigrationRunning = useSelector(
-    props.actor,
-    (state) => state.context.currentMigrationRunning
-  );
-
   const currentMigrationFailed = useSelector(props.actor, (state) =>
     currentMigrationHasFailed(state.context)
   );
 
+  const isCurrentMigrationStopped = useSelector(props.actor, (state) =>
+    currentMigrationIsStopped(state.context)
+  );
   const currentMigrationSuccess = useSelector(props.actor, (state) =>
     currentMigrationHasSucceeded(state.context)
   );
@@ -75,15 +74,16 @@ export function AutomaticMigration(props: {
 
   return (
     <MigrationTimeline
+      actor={props.actor}
       migrations={props.migrations}
       nxConsoleMetadata={props.nxConsoleMetadata}
       currentMigrationIndex={
         currentMigrationIndex >= 0 ? currentMigrationIndex : 0
       }
-      currentMigrationRunning={currentMigrationRunning}
       currentMigrationFailed={currentMigrationFailed}
       currentMigrationSuccess={currentMigrationSuccess}
       currentMigrationHasChanges={currentMigrationChanges}
+      currentMigrationStopped={isCurrentMigrationStopped}
       isDone={isDone}
       isInit={isInit}
       onRunMigration={props.onRunMigration}
--- a/graph/migrate/src/lib/components/automatic-migration.tsx
+++ b/graph/migrate/src/lib/components/automatic-migration.tsx
@@ -6,9 +6,7 @@ import type { MigrationsJsonMetadata } from 'nx/src/command-line/migrate/migrate
 import { useSelector } from '@xstate/react';
 import {
   currentMigrationHasChanges,
-  currentMigrationHasFailed,
-  currentMigrationHasSucceeded,
-  currentMigrationIsStopped,
+  getCurrentMigrationType,
 } from '../state/automatic/selectors';
 import { MigrationTimeline } from './migration-timeline';
 import { Interpreter } from 'xstate';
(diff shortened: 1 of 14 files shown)


packages/nx/src/command-line/migrate/migrate-ui-api.ts
```diff
@@ -356,3 +441,47 @@ export function undoMigration(workspacePath: string, id: string) {
...
+): boolean {
+  try {
+    if (workspacePath) {
+      modifyMigrationsJsonMetadata(workspacePath, stopMigration(migrationId));
+    }
+
+    // Check if this is the currently running migration and kill the process
+    if (currentMigrationId === migrationId && currentMigrationProcess) {
+      if (process.platform === 'win32') {
+        execSync(`taskkill /f /t /pid ${currentMigrationProcess.pid}`, {
```
jaysoo: I don't think we need taskkill since it's just one process. There shouldn't be child processes, but if there were then `p.kill()` should work as well.

graph/migrate/src/lib/components/migration-card.tsx
```diff
@@ -99,14 +120,31 @@ export const MigrationCard = forwardRef<
...
+  );
+  const isSkipped = useSelector(actor, (state) =>
+    isMigrationSkipped(state.context, migration.id)
+  );
+  const isStopped = useSelector(actor, (state) =>
+    isMigrationStopped(state.context, migration.id)
+  );
+  const hasChanges = useSelector(actor, (state) =>
+    currentMigrationHasChanges(state.context)
+  );
```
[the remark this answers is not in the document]
ndcunningham: The selector is only used for the `currentMigration` card so we should be fine.

graph/migrate/src/lib/state/automatic/selectors.ts
```diff
@@ -4,58 +4,102 @@ import { MigrationDetailsWithId } from 'nx/src/config/misc-interfaces';
...
+  ctx: AutomaticMigrationState,
+  migrationId: string
+) {
+  return (
+    ctx.nxConsoleMetadata?.completedMigrations?.[migrationId]?.type ===
+    'successful'
+  );
+}
+
+export function isMigrationFailed(
```
jaysoo: Instead of having one selector for each type, I'd just use `getMigrationType` and then in the component you can select the type and do the `===` comparison there. Otherwise we'd have to add more and more of these selectors if we add more types.

jaysoo (approved):

graph/migrate/src/lib/components/automatic-migration.tsx
```diff
@@ -45,17 +44,18 @@ export function AutomaticMigration(props: {
...
-  const currentMigrationRunning = useSelector(
+  const currentMigrationFailed = useSelector(
     props.actor,
-    (state) => state.context.currentMigrationRunning
+    (state) => getCurrentMigrationType(state.context) === 'failed'
   );
 
-  const currentMigrationFailed = useSelector(props.actor, (state) =>
-    currentMigrationHasFailed(state.context)
+  const isCurrentMigrationStopped = useSelector(
```
MaxKless, line 52: nit but why does this one start with `is...` when the others don't?

graph/migrate/src/lib/components/migration-card.tsx
```diff
@@ -99,14 +117,35 @@ export const MigrationCard = forwardRef<
...
-  );
 
-  const madeChanges = succeeded && !!migrationResult?.changedFiles.length;
+  const filesChanged =
+    migrationResult?.type === 'successful' ? migrationResult.changedFiles : [];
+
+  const nextSteps =
+    migrationResult?.type === 'successful' ? migrationResult.nextSteps : [];
+
+  const isSucceeded = useSelector(
```
MaxKless, line 127: aren't all these selectors duplicated from `automatic-migration.tsx`?

graph/migrate/src/lib/components/migration-timeline.tsx
```diff
@@ -538,44 +574,67 @@ function TimelineButton({ icon: Icon, onClick }: TimelineButtonProps) {
...
   onClick,
 }: MigrationStateCircleProps) {
   let bgColor = '';
   let textColor = '';
   let Icon = null;
 
-  // Check if this migration is in the completed migrations
-  const completedMigration =
-    nxConsoleMetadata.completedMigrations?.[migration.id];
+  const isSkipped = useSelector(
```
MaxKless, line 599: again, maybe we can find a way to reuse these selectors somehow?

MaxKless (approved):
some small stuff but I'm okay if it goes in like this

graph/migrate/src/lib/components/migration-card.tsx
ndcunningham: I don't think they are for the same purpose. 
In `automatic-migration.tsx` most of the derived selectors are for the _current migration_ while these selectors are for the migration passed in which can be  current, future or past.

PR #31634, opened 2025-06-18
feat(docker): add docker plugin

Coly010:
## Current Behavior
We do not currently have a plugin for docker to infer targets such as
- build
- run

## Expected Behavior
The Docker Plugin should remain agnostic between tech stacks, but should allow for the containerization of projects.

It should support initially the ability to build and run images by inferring targets on projects that contain a Dockerfile.

`nx release` should be expanded to support releasing Docker images to registries


packages/docker/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -0,0 +1,93 @@
...
+} from './schema';
+import { join } from 'path';
+import { existsSync, readFileSync } from 'fs';
+
+export default async function dockerReleasePublish(
+  schema: DockerReleasePublishSchema,
+  context: ExecutorContext
+) {
+  logger.warn(
+    `Docker support is experimental. Breaking changes may occur and not adhere to semver versioning.`
```
FrozenPandaz: Let's only warn once maybe during nx-release rather than once for every single project.

packages/docker/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -0,0 +1,93 @@
...
+      return imageRef;
+    }
+    throw new Error(
+      `Could not find Docker Image ${imageRef}. Did you run 'nx release version'?`
+    );
+  }
+}
+
+function readVersionFromFile(projectRoot: string) {
+  const versionFilePath = join(projectRoot, '.docker-version');
```
FrozenPandaz: Let's move this into tmp

packages/docker/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -0,0 +1,93 @@
...
+    );
+    return result.trim().length > 0;
+  } catch {
+    return false;
+  }
+}
+
+function dockerPush(imageReference: string) {
+  try {
+    const result = execSync(`docker push ${imageReference} --quiet`, {
```
FrozenPandaz: Remove --quiet so that we're more transparent to what docker users would be used to.

If they pass --quiet, we should pass it through. aka Add an option and pass it through here. Users can configure it easily with targetDefaults

packages/docker/src/generators/init/init.ts
```diff
@@ -0,0 +1,71 @@
...
+    {},
+    {
+      '@nx/docker': nxVersion,
+    },
+    undefined,
+    schema.keepExistingVersions
+  );
+}
+
+export function initGenerator(tree: Tree, schema: InitGeneratorSchema) {
```
FrozenPandaz: This can be removed and the Internal one can be renamed to be the one and only generator function.

packages/docker/src/generators/init/init.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export async function initGeneratorInternal(
+  tree: Tree,
+  schema: InitGeneratorSchema
+) {
+  logger.warn(
+    `Docker support is experimental. Breaking changes may occur and not adhere to semver versioning.`
+  );
+  const nxJson = readNxJson(tree);
+  const addPluginDefault =
```
FrozenPandaz: Please remove the option to not add the plugin... it would be very hard to configure without it.

packages/docker/src/plugins/plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+function writeTargetsCache(
+  cachePath: string,
+  results?: Record<string, DockerTargets>
+) {
+  writeJsonFile(cachePath, results ?? {});
+}
+
+const dockerfileGlob = '**/Dockerfile';
```
FrozenPandaz, line 40: I think most of our other plugins have something like

`Dockerfile,**/Dockerfile`

packages/docker/src/plugins/plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+
+async function createNodesInternal(
+  configFilePath: string,
+  options: DockerPluginOptions,
+  context: CreateNodesContext,
+  targetsCache: Record<string, DockerTargets>
+) {
+  const projectRoot = dirname(configFilePath);
+  const normalizedOptions = normalizePluginOptions(options);
+  const hash = await calculateHashForCreateNodes(
```
FrozenPandaz: We can use calculateHashesForCreateNodes

packages/docker/src/plugins/plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+      ...('production' in namedInputs
+        ? ['production', '^production']
+        : ['default', '^default']),
+    ],
+    metadata: {
+      technologies: ['docker'],
+      description: `Run Docker build`,
+      help: {
+        command: `docker build --help`,
+        example: {},
```
FrozenPandaz: Add some examples for --cache-from and --cache-to

packages/docker/src/plugins/plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+      ...('production' in namedInputs
+        ? ['production', '^production']
+        : ['default', '^default']),
+    ],
+    metadata: {
+      technologies: ['docker'],
+      description: `Run Docker run`,
+      help: {
+        command: `docker run --help`,
+        example: {},
```
FrozenPandaz: Add some examples here. Probably use the docker run reference as a starter.

packages/docker/src/plugins/plugin.ts
```diff
@@ -0,0 +1,164 @@
...
+function writeTargetsCache(
+  cachePath: string,
+  results?: Record<string, DockerTargets>
+) {
+  writeJsonFile(cachePath, results ?? {});
+}
+
+const dockerfileGlob = '**/Dockerfile';
+
+export const createNodesV2: CreateNodesV2<DockerPluginOptions> = [
```
FrozenPandaz, line 42: Add tests for this.

FrozenPandaz (changes requested):

packages/docker/src/plugins/plugin.spec.ts
```diff
@@ -0,0 +1,652 @@
...
+                        "command": "docker build --help",
+                        "example": {
+                          "options": {
+                            "cache-from": "type=s3,region=eu-west-1,bucket=mybucket .",
+                            "cache-to": "type=s3,region=eu-west-1,bucket=mybucket .",
+                          },
+                        },
+                      },
+                      "technologies": [
+                        "docker",
```
FrozenPandaz, line 74: @Coly010 

Can we add an icon for our project details view please?

Can you work with the Nx Cloud team to get an icon for us please?

packages/docker/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "executors": "./executors.json",
+  "generators": "./generators.json",
+  "dependencies": {
+    "@nx/devkit": "workspace:*",
+    "enquirer": "~2.3.6",
+    "tslib": "^2.3.0"
+  },
+  "devDependencies": {
+    "nx": "workspace:*"
```
JamesHenry, line 51: Should not be necessary and isn't done by any other packages

packages/docker/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "./executors.json": "./executors.json",
+    "./migrations.json": "./migrations.json"
+  },
+  "nx-migrations": {
+    "migrations": "./migrations.json"
+  },
+  "executors": "./executors.json",
+  "generators": "./generators.json",
+  "dependencies": {
+    "@nx/devkit": "workspace:*",
```
JamesHenry, line 46: Let's use file to be consistent with all other packages for now, we can migrate them all at some point

packages/docker/project.json
```diff
@@ -0,0 +1,52 @@
...
+  "name": "docker",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "packages/docker",
+  "projectType": "library",
+  "targets": {
+    "build": {
+      "command": "node ./scripts/copy-readme.js docker",
+      "outputs": ["{workspaceRoot}/build/packages/docker/README.md"]
+    },
+    "legacy-post-build": {
```
JamesHenry, line 11: I don't recognize this, what is going to invoke it?

packages/docker/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -0,0 +1,99 @@
...
+  return {
+    quiet: schema.quiet ?? false,
+    imageReference: findImageReference(projectConfig, schema),
+    dryRun: process.env.NX_DRY_RUN === 'true' || schema.dryRun || false,
+  };
+}
+
+function findImageReference(
+  projectConfig: ProjectGraphProjectNode,
+  schema: DockerReleasePublishSchema
```
JamesHenry, line 54: Does not seem to be used here

packages/docker/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -0,0 +1,99 @@
...
+    return result.trim().length > 0;
+  } catch {
+    return false;
+  }
+}
+
+function dockerPush(imageReference: string, quiet: boolean) {
+  try {
+    const result = execSync(
+      `docker push ${imageReference}${quiet ? ' --quiet' : ''}`,
```
JamesHenry, line 94: We usually increase the maxBuffer on these kind of operations just in case. I am not sure how much output docker push produces but might be good just to be safe

packages/docker/src/plugins/plugin.ts
```diff
@@ -0,0 +1,181 @@
...
+    const optionsHash = hashObject(options);
+    const cachePath = join(
+      workspaceDataDirectory,
+      `docker-${optionsHash}.hash`
+    );
+    const targetsCache = readTargetsCache(cachePath);
+    const projectRoots = configFilePaths.map((c) => dirname(c));
+    const normalizedOptions = normalizePluginOptions(options);
+    const hashes = await calculateHashesForCreateNodes(
+      projectRoots,
```
FrozenPandaz, line 55: nit for later, we don't need to hash the entire project directory, only the dockerfile. See if there's a way we can do that.

packages/docker/src/executors/release-publish/schema.d.ts
```diff
@@ -0,0 +1,10 @@
+export interface DockerReleasePublishSchema {
+  dryRun?: boolean;
+  quiet?: boolean;
+}
+
+export interface NormalizedDockerReleasePublishSchema {
```
JamesHenry: Can we inline this in the file? I was initially confused by its presence here and why imageReference was not on the schema.json, I would say this type is really just implementation details related and not the schema directly

packages/docker/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/nrwl/nx/issues"
+  },
+  "keywords": [
+    "Monorepo",
+    "Docker",
+    "Containers"
+  ],
+  "author": "Colum Ferry",
```
JamesHenry: We do not usually attribute personal authorship, although I see you are the only other person who has done so in the repo...

packages/docker/src/release/version-pattern-utils.ts
JamesHenry, line 1: This seems like a good candidate for unit tests, please add a spec file

packages/docker/src/release/version-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+async function promptForNewVersion(
+  versionSchemes: Record<string, string>,
+  projectName: string
+) {
+  const { versionScheme } = await prompt<{ versionScheme: string }>({
+    name: 'versionScheme',
+    type: 'select',
+    message: `What type of release would you like to make for project "${projectName}"?`,
```
JamesHenry: Might be clearer where it's originating from?

```suggestion
    message: `What type of docker release would you like to make for project "${projectName}"?`,
```

packages/nx/src/command-line/release/version/release-group-processor.ts
```diff
@@ -736,6 +759,59 @@ export class ReleaseGroupProcessor {
...
+    try {
+      const {
+        handleDockerVersion: _handleDockerVersion,
+        // nx-ignore-next-line
+      } = require(// @ts-ignore
+      '@nx/docker/release/version-utils');
+      handleDockerVersion = _handleDockerVersion;
+    } catch (e) {
+      console.error(
+        'Could not find `@nx/docker`. Please run `nx add @nx/docker` before attempting to releas Docker images.'
```
FrozenPandaz: nit: There's a typo here 
```suggestion
        'Could not find `@nx/docker`. Please run `nx add @nx/docker` before attempting to release Docker images.'
```

packages/docker/src/release/version-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function updateProjectVersion(
+  newVersion: string,
+  workspaceRoot: string,
+  projectRoot: string,
+  repositoryName?: string,
+  registry?: string
+): string[] {
+  const isDryRun = process.env.NX_DRY_RUN && process.env.NX_DRY_RUN !== 'false';
```
JamesHenry, line 98: Why not

```suggestion
  const isDryRun = process.env.NX_DRY_RUN === 'true';
```

FrozenPandaz (approved):

packages/docker/src/release/version-utils.ts
```diff
@@ -0,0 +1,131 @@
...
+  const imageRef = getDefaultImageReference(projectRoot);
+  const newImageRef = getImageReference(projectRoot, repositoryName, registry);
+  const fullImageRef = `${newImageRef}:${newVersion}`;
+  if (!isDryRun) {
+    execSync(`docker tag ${imageRef} ${fullImageRef}`);
+  }
+  const logs = [`Image tagged with ${fullImageRef}.`];
```
JamesHenry: Only a nit: In standard versioning we tend to change the tense of the action `would tag image with ".." but dry is enabled` etc

packages/docker/src/utils/versions.ts
```diff
@@ -0,0 +1 @@
+export const nxVersion = require('../../package.json').version;
```
JamesHenry, line 1: The diff is a bit too big I can't find it, what is this used for?

packages/nx/schemas/nx-schema.json
```diff
@@ -995,6 +1019,59 @@
...
+          "description": "Repository name for the image on the configured registry"
+        },
+        "registryUrl": {
+          "type": "string",
+          "description": "Url of the Docker Image/Container Registry to push images to. Defaults to Docker Hub."
+        }
+      },
+      "additionalProperties": false
+    },
+    "NxReleaseGroupDockerConfiguration": {
```
JamesHenry, line 1065: Isn't this missing the other additional config properties?

packages/docker/package.json
Coly010: This is done by other packages:
https://github.com/nrwl/nx/blob/master/packages/angular/package.json#L70
https://github.com/nrwl/nx/blob/master/packages/cypress/package.json#L49
https://github.com/nrwl/nx/blob/master/packages/js/package.json#L66
etc

packages/docker/package.json
Coly010: N/A

packages/docker/project.json
Coly010: N/A

packages/nx/src/command-line/release/release.ts
```diff
@@ -351,7 +351,8 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
     // null means that all projects are versioned together but there were no changes
     if (versionResult.workspaceVersion !== null) {
       hasNewVersion = Object.values(versionResult.projectsVersionData).some(
-        (version) => version.newVersion !== null
+        (version) =>
+          version.newVersion !== null || version.dockerVersion !== null
```
JamesHenry, line 360: Maybe add a comment covering the scenario where only dockerVersion is set for posterity

packages/nx/src/command-line/release/utils/shared.ts
```diff
@@ -229,6 +234,33 @@ function stripPlaceholders(str: string, placeholders: string[]): string {
   return str;
 }
 
+export function shouldPreferDockerVersionForReleaseGroup(
```
JamesHenry, line 237: This function seems to behave identically in both code branches

packages/nx/src/command-line/release/utils/shared.ts
```diff
@@ -229,6 +234,33 @@ function stripPlaceholders(str: string, placeholders: string[]): string {
...
+}
+
+export function shouldSkipVersionActions(
+  dockerOptions: { skipVersionActions?: string[] | boolean },
+  projectName: string
+): boolean {
+  return (
+    dockerOptions.skipVersionActions === true ||
+    (Array.isArray(dockerOptions.skipVersionActions) &&
+      dockerOptions.skipVersionActions.includes(projectName))
```
JamesHenry, line 253: Let's use proper project matches here not just exact strings, so that we have the same flexibility as in `projects`

packages/nx/src/command-line/release/version.ts
```diff
@@ -258,6 +258,61 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
...
+          releaseGroup,
+          true
+        );
+      }
+    }
+
+    // TODO(colum): Remove when Docker support is no longer experimental
+    if (nxReleaseConfig.docker || releaseGroups.some((rg) => rg.docker)) {
+      output.warn({
+        title: 'Warning!',
```
JamesHenry: Can maybe lose the exclamation mark to align with the tone of our other warnings etc

packages/nx/src/command-line/release/version/release-group-processor.spec.ts
```diff
@@ -1675,7 +1675,7 @@ describe('ReleaseGroupProcessor', () => {
   });
 
   describe('versionData', () => {
-    it('should populate versionData even when projects are not versioned', async () => {
+    it('should populate versionData even ~when projects are not versioned', async () => {
```
JamesHenry: 😄 
```suggestion
    it('should populate versionData even when projects are not versioned', async () => {
```

packages/nx/src/command-line/release/version/release-group-processor.ts
```diff
@@ -65,6 +72,7 @@ export const BUMP_TYPE_REASON_TEXT = {
     ', because a dependency project belonging to another release group was bumped, ',
   OTHER_PROJECT_IN_FIXED_GROUP_WAS_BUMPED_DUE_TO_DEPENDENCY:
     ', because of a dependency-only bump to another project in the same fixed release group, ',
+  NOOP_VERSION_ACTIONS: ', because this project uses docker, ',
```
JamesHenry: Should this explain that it uses docker _and_ that the docker config has caused it to skip version actions? (More succinctly than that obviously)

packages/nx/src/config/nx-json.ts
```diff
@@ -82,6 +82,39 @@ export type ManifestRootToUpdate =
   | string
   | { path: string; preserveLocalDependencyProtocols: boolean };
 
+export interface NxReleaseDockerConfiguration {
+  /**
+   * A command to run after validation of nx release configuration, but before versioning begins.
```
JamesHenry:
```suggestion
   * A command to run after validation of nx release configuration, but before docker versioning begins.
```

JamesHenry: This runs after standard versioning right? Should probably clarify

packages/nx/src/config/nx-json.ts
```diff
@@ -82,6 +82,39 @@ export type ManifestRootToUpdate =
...
+   * Useful for preparing docker build artifacts. If --dry-run is passed, the command is still executed,
+   * but with the NX_DRY_RUN environment variable set to 'true'.
+   */
+  preVersionCommand?: string;
+  /**
+   * Projects which should use a no-op VersionActions implementation rather than any potentially inferred by default or via Inference Plugins.
+   * Can be an array of project names (subset of projects in the release setup/release group) or a boolean (true means all projects).
```
JamesHenry, line 94: Might be good to give an example of JS projects skipping updating a package.json because right now this comment requires some understanding of the internals

packages/nx/src/config/nx-json.ts
```diff
@@ -341,6 +379,21 @@ export interface NxReleaseConfiguration {
...
        */
       projects: string[] | string;
+      /**
+       * Configure options to handle versioning docker projects for this group.
+       * Set to `true` to enable with default settings, or provide a configuration object for custom settings.
+       */
+      docker?:
+        | (NxReleaseDockerConfiguration & {
+            /**
+             * A command to run after validation of nx release configuration, but before versioning begins.
```
JamesHenry: Same comment as above

packages/docker/src/executors/release-publish/release-publish.impl.ts
Coly010: Used in normalizeOptions

packages/docker/src/release/version-utils.ts
Coly010: I just matched what was used in other areas

packages/docker/src/utils/versions.ts
Coly010: For `init` generator to install the correct nx version. Common pattern in our packages.

packages/nx/schemas/nx-schema.json
Coly010: I matched this with how NxReleaseGroupVersionConfiguration did it

packages/nx/src/command-line/release/utils/shared.ts
Coly010: I'm not sure where we are doing this for `projects` in nx release currently. It seems like we're matching by full name for `args.projects` too?

Coly010: Found it

packages/nx/src/config/nx-json.ts
Coly010: good point

PR #31658, opened 2025-06-20
feat(nx-plugin): refactor plugin options normalization to common method with defaultOptions (normalizeOptions)

mxfh:
- Move all plugin option normalization to a shared utility
- Add formatter and fallback logic, with tests for normalizeOptions/normalizeExtensions
- Webpack plugin now is no longer overriding user supplied build/watch deps strings, it fills them from defaults if unset
- Note: consider whether internal options like buildDepsTargetName and watchDepsTargetName should be exposed to users at all

This is a refactor for plugin option handling. Further plugin-specific improvements.

found some likely inconsistencies in the process.

## Current Behavior

in nx.json

ommiting options
```
{
  "plugin": "@nx/storybook/plugin",
}
```

causes failure in processin project graph

```
npx nx run-many -t=build --verbose

 NX   Failed to process project graph.

     - TypeError: Cannot read properties of undefined (reading 'buildStorybookTargetName')
      at normalizeOptions (workspaceRoot/node_modules/@nx/storybook/src/plugins/plugin.js:240:43)
      at exports.createNodesV2 (workspaceRoot/node_modules/@nx/storybook/src/plugins/plugin.js:31:35)

```

this works:
```
{
  "plugin": "@nx/storybook/plugin",
  "options": {}
}
```
## Expected Behavior
```
{
  "plugin": "@nx/storybook/plugin",
}
```
should just run with all default Options

## Related Issue(s)
could not find any open issues in that context

Fixes #

### DIFF
--- a/packages/next/src/plugins/plugin.ts
+++ b/packages/next/src/plugins/plugin.ts
@@ -18,6 +18,7 @@ import { getNamedInputs } from '@nx/devkit/src/utils/get-named-inputs';
 import { existsSync, readdirSync } from 'fs';
 
 import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
+import { normalizeOptions } from 'nx/src/utils/normalize-options';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
 import { getLockFileName } from '@nx/js';
 import { loadConfigFile } from '@nx/devkit/src/utils/config-utils';
@@ -35,6 +36,12 @@ export interface NextPluginOptions {
   buildDepsTargetName?: string;
   watchDepsTargetName?: string;
 }
+
+const defaultOptions: NextPluginOptions = {
+  buildTargetName: 'build',
+  devTargetName: 'dev',
+  startTargetName: 'start',
+};
 
 const pmc = getPackageManagerCommand();
 
@@ -67,16 +74,16 @@ export const createDependencies: CreateDependencies = () => {
 export const createNodesV2: CreateNodesV2<NextPluginOptions> = [
   nextConfigBlob,
   async (configFiles, options, context) => {
-    const optionsHash = hashObject(options);
+    const normalizedOptions = normalizeOptions(options, defaultOptions);
+    const optionsHash = hashObject(normalizedOptions);
     const cachePath = join(workspaceDataDirectory, `next-${optionsHash}.json`);
     const targetsCache = readTargetsCache(cachePath);
-
     try {
       return await createNodesFromFiles(
         (configFile, options, context) =>
           createNodesInternal(configFile, options, context, targetsCache),
         configFiles,
-        options,
+        normalizedOptions,
         context
       );
     } finally {
@@ -96,13 +103,14 @@ export const createNodes: CreateNodes<NextPluginOptions> = [
       '`createNodes` is deprecated. Update your plugin to utilize createNodesV2 instead. In Nx 21, this will change to the createNodesV2 API.'
     );
 
-    const optionsHash = hashObject(options);
+    const normalizedOptions = normalizeOptions(options, defaultOptions);
+    const optionsHash = hashObject(normalizedOptions);
     const cachePath = join(workspaceDataDirectory, `next-${optionsHash}.json`);
     const targetsCache = readTargetsCache(cachePath);
 
     const result = await createNodesInternal(
       configFilePath,
-      options,
+      normalizedOptions,
       context,
       targetsCache
     );
@@ -130,7 +138,6 @@ async function createNodesInternal(
   ) {
     return {};
   }
-  options = normalizeOptions(options);
 
   const hash = await calculateHashForCreateNodes(
     projectRoot,
@@ -273,15 +280,6 @@ function getNextConfig(
   const resolvedPath = join(context.workspaceRoot, configFilePath);
 
   return loadConfigFile(resolvedPath);
-}
-
-function normalizeOptions(options: NextPluginOptions): NextPluginOptions {
-  options ??= {};
-  options.buildTargetName ??= 'build';
-  options.devTargetName ??= 'dev';
-  options.startTargetName ??= 'start';
-  options.serveStaticTargetName ??= 'serve-static';
-  return options;
 }
 
 function getInputs(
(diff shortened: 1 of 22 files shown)


packages/storybook/src/plugins/plugin.ts
```diff
@@ -404,20 +412,6 @@ function getOutputs(): string[] {
   return outputs;
 }
 
-function normalizeOptions(
-  options: StorybookPluginOptions
-): Required<StorybookPluginOptions> {
-  return {
```
mxfh, line 410: all other options pretty much had this pattern 
`options ??= {};`
but here no empy object is filled with the defaults

packages/webpack/src/plugins/plugin.ts
```diff
@@ -331,15 +339,20 @@ function normalizeOutputPath(
...
-    buildTargetName: options?.buildTargetName ?? 'build',
-    serveTargetName: options?.serveTargetName ?? 'serve',
-    serveStaticTargetName: options?.serveStaticTargetName ?? 'serve-static',
-    previewTargetName: options?.previewTargetName ?? 'preview',
+  const partial = normalizeOptions(options, defaults);
+
+  // TODO: clarify if this behavior is desired, just replicating prior behavior with types now
+  const required: Required<WebpackPluginOptions> = {
+    ...partial,
     buildDepsTargetName: 'build-deps',
```
mxfh: was this override part intentional? 
should these even be in the exported WebpackPluginOptions then?
pretty much only outlier i've come accross in the option handling in the plugins.

packages/next/src/plugins/plugin.ts
```diff
@@ -36,6 +37,12 @@ export interface NextPluginOptions {
   watchDepsTargetName?: string;
 }
 
+const defaultOptions: NextPluginOptions = {
+  buildTargetName: 'build',
+  devTargetName: 'dev',
+  startTargetName: 'start',
+};
```
mxfh, line 44: ommited the deprecated value here since the startTargetName is also in the defaults
serveStaticTargetName?: string;

packages/nuxt/src/plugins/plugin.ts
```diff
@@ -16,12 +16,29 @@ import { loadConfigFile } from '@nx/devkit/src/utils/config-utils';
...
 import { loadNuxtKitDynamicImport } from '../utils/executor-utils';
 import { addBuildAndWatchDepsTargets } from '@nx/js/src/plugins/typescript/util';
 
+export interface NuxtPluginOptions {
+  buildTargetName?: string;
+  serveTargetName?: string;
+  serveStaticTargetName?: string;
+  buildStaticTargetName?: string;
+  buildDepsTargetName?: string;
+  watchDepsTargetName?: string;
```
mxfh, line 32: buildDepsTargetName
watchDepsTargetName

are missing from the defaults, see also webpack options behavior

mxfh: they got added in this commit but only webpack got an update to the defaultOptions:

https://github.com/nrwl/nx/commit/0ae8665a885935982ecbfc3683b3d38bc4d56f44#diff-8ea87d5fdfbfcec0fc69136da5a959e64f7e2747a01611c16227183d06b24eabR33-R331

packages/webpack/src/plugins/plugin.ts
mxfh: see also nuxt and other plugins with these options

buildDepsTargetName
watchDepsTargetName

coming from this commit
https://github.com/nrwl/nx/commit/0ae8665a885935982ecbfc3683b3d38bc4d56f44#diff-8ea87d5fdfbfcec0fc69136da5a959e64f7e2747a01611c16227183d06b24eabR33-R331

packages/vite/src/plugins/plugin.ts
```diff
@@ -46,6 +47,15 @@ export interface VitePluginOptions {
...
 }
 
+const defaultOptions: VitePluginOptions = {
+  buildTargetName: 'build',
+  testTargetName: 'test',
+  devTargetName: 'dev',
+  previewTargetName: 'preview',
+  serveStaticTargetName: 'serve-static',
+  typecheckTargetName: 'typecheck',
+};
```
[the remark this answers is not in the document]
mxfh: /**
   * @deprecated Use devTargetName instead. This option will be removed in Nx 22.
   */
   
intentionally removed so the deprecated option stays unset.

mxfh:
@AgentEnder @FrozenPandaz this looks like a bit what happened here to https://github.com/nrwl/nx/pull/28838 here too.

My PR explicitly doesn't handle nested cases.

The typescript plugin now uses a per sub-option manual iterative approach, due to the false | object logic.

I think no would greatly benefit from a consistent handling of options and options merging strategy, not every new plugin comes with a new way of implementing this.

Any comment on if the proposed an approach here has other implementations elsewhere already or is conflicting with other paradigms? Thanks.

fregante:
I just bumped into this issue: `@nx/storybook/plugin` does not have a default task name as [the docs suggest](https://nx.dev/technologies/test-tools/storybook/introduction#nxstorybook-configuration):

> The default names are build-storybook, storybook, test-storybook and static-storybook.

this isn't possible:

```json

  "plugins": [
    {
      "options": {
        "build": {
          "configName": "tsconfig.lib.json"
        }
      }
    },
    "@nx/storybook/plugin",
  ],
```

```
An error occurred while processing files for the @nx/storybook/plugin plugin (Defined at nx.json#plugins[2])
.
  - Cannot read properties of undefined (reading 'buildStorybookTargetName')
```

So great to see there's already a PR that fixes this 🙏

mxfh:
@fregante I had a no luck finding an issue for this, maybe it helps to add one for this PR to get some visibility? I think it would help if you could create one with your case and add the one from the PR description about empty vs missing options. And then add this PR as fix.

Didn't get any feedback from the maintainers yet, it's touching quite a lot of files but this is just about implementing a pattern with typed default options on top of the file. If this is too much of a scope, could also just cherry-pick into a new PR with just the storybook and typescript plugin fixes.

This PR mixes DX with a fix, happy to implement it gradually one by one, but since nothing should functionally change here I went all in on the plugins to showcase the idea of having a clear default section in every plugin.

As said, I didn't change anything about the default values themselves and some of them still seem quite inconsistent.

PR #31665, opened 2025-06-20
fix(core): allow any framework value in preset generator schema

FrozenPandaz:
## Current Behavior

When using third-party presets with `npx create-nx-workspace`, passing framework values other than the predefined backend frameworks fails with a schema validation error:

```
Error: Property 'framework' does not match the schema. 'angular' should be one of express,koa,fastify,nest,none.
```

For example:
```bash
npx create-nx-workspace --preset=@rig/cli@alpha --framework=angular --theme=pcb --name=hello-world
```

## Expected Behavior

Third-party presets should be able to pass any framework value without schema validation errors, allowing for greater flexibility in the Nx ecosystem.

## Related Issue(s)

This fixes the issue where third-party presets cannot pass framework values like 'angular', 'react', 'vue', etc., due to overly restrictive schema validation.

## Changes

- Removed the `enum` constraint from the `framework` property in the workspace preset generator schema
- The property remains a string type but now accepts any value
- Maintains backward compatibility with existing functionality

The built-in Nx preset generators still only handle frameworks they know about, but the schema validation no longer blocks third-party presets from passing other framework values.

## Test plan

- [x] Workspace package builds successfully
- [x] Schema validation allows any string value for framework property
- [x] Backward compatibility maintained for existing presets
- [ ] Manual test: `npx create-nx-workspace --preset=@rig/cli@alpha --framework=angular` should work without schema validation errors


AgentEnder (approved):

FrozenPandaz:
@claude fix these documentation issues.

PR #31722, opened 2025-06-25
nx release version does not propagate changes to release groups with "projectsRelationship": "independent"

mpsanchis:
Test that shows the problem mentioned in https://github.com/nrwl/nx/issues/31724


JamesHenry:
@mpsanchis and I are discussing this, assigned myself

PR #31756, opened 2025-06-27
feat(release): releaseTagPatternStrictPreid option to match git tag based on semver preid

### LINKED ISSUE #31740: `nx release` current version selection wrong when using tags and independent release
### Current Behavior

I discovered that when using `nx release` in a project configured to use `conventionalCommits: true` (so `git-tag` as source for current version) and `projectsRelationship: "independent"`, the tag filtering breaks, resulting in `nx release` (no `--preid`) using the wrong previous version (tag).

Based on the implementation, the tags are sorted using the `versionsort.suffix=-` option so that a version `0.1.2` appears before `0.1.3-bla.0`. This assumes that the tag will be in the form of `{version}-{preid}.{rev}`. When doing independent releases, tags are in the form of `{project}@{version}-{preid}.{rev}`, therefore breaking this sorting.

As a result, running `nx release` with no `--preid` option when there are existent pre-release tags will not result in the creation of a stable release, in fact it might lead to no release being created, depending on the setup.

### Expected Behavior

`nx release` tag sorting logic should respect the intended logic of putting stable releases before pre-releases, as described in a comment in the codebase:

> // Apply git config to take version suffixes into account when sorting, e.g. 1.0.0 is newer than 1.0.0-beta.1

### GitHub Repo

_No response_

### Steps to Reproduce

1. Set up independent releases
2. Create initial release with `nx release` -> creates `package@0.0.1`
3. Release package with `nx release --preid=test` -> creates `package@0.0.2-test.0`
4. Run `nx release`
5. Acknowledge that no release is created, because the prerelease tag is used instead of the latest, previous, stable release

### Nx Report

```shell
Node           : 22.16.0
OS             : darwin-arm64
npm            : 10.9.2

nx (global)            : 21.1.1
nx                     : 21.2.0
@nx/js                 : 21.2.0
@nx/jest               : 21.2.0
@nx/eslint             : 21.2.0
@nx/workspace          : 21.2.0
@nx/devkit             : 21.2.0
@nx/esbuild            : 21.2.0
@nx/eslint-plugin      : 21.2.0
@nx/module-federation  : 21.2.0
@nx/node               : 21.2.0
@nx/plugin             : 21.2.0
@nx/react              : 21.2.0
@nx/vite               : 21.2.0
@nx/web                : 21.2.0
Registered Plugins:
@nx/js/typescript
@nx/react/router-plugin
@nx/eslint/plugin
@nx/vite/plugin
Cache Usage: 57.91 MB / 46.04 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

iuliancmarcu:
When running nx release without a preid with a most recent pre-release version tag, its commit was used as base for calculating the new version for a stable release. This was incorrect in most cases, and the behaviour is corrected by this fix, prioritising stable version tags when doing stable releases, and prioritising tags for the same preid when doing pre-releases.

closed #31740

## Current Behavior

We have tags:
```
my-lib@1.2.4-beta.0
my-lib@1.2.3
my-lib@1.2.3-beta.0
```

When running `nx release --preid=beta`, the current version tag is `my-lib@1.2.4-beta.0`, which is correct, as we are releasing another beta pre-release.

When running `nx release`, the source tag is also `my-lib@1.2.4-beta.0`, which is incorrect, because we are looking to release a stable version, therefore we want to use `my-lib@1.2.3` (latest stable) as source tag.

## Expected Behavior

To prioritise selecting the latest stable release tag as current version source when a non-prerelease release is created.

The new behaviour is:

**Pre-release**
|Command|Tags|Selected Tag|New Version|
|-|-|-|-|
|nx release --preid=beta|my-lib@0.0.1|my-lib@0.0.1|my-lib@0.0.2-beta.0|
|nx release --preid=beta|my-lib@0.0.1, my-lib@0.0.2-alpha.0|my-lib@0.0.1|my-lib@0.0.2-beta.0|
|nx release --preid=beta|my-lib@0.0.2-alpha.0|my-lib@0.0.2-alpha.0|my-lib@0.0.2-beta.0|

**Release**
|Command|Tags|Selected Tag|New Version|
|-|-|-|-|
|nx release|my-lib@0.0.1|my-lib@0.0.1|my-lib@0.0.2|
|nx release|my-lib@0.0.1, my-lib@0.0.2-alpha.0|my-lib@0.0.1|my-lib@0.0.2|
|nx release|my-lib@0.0.2-alpha.0|my-lib@0.0.2-alpha.0|my-lib@0.0.2|

The table above shows a patch pump - conventional commits will decide the type of version bump.

## Related Issue(s)

Fixes #31740


iuliancmarcu:
@FrozenPandaz I saw you triaging a related issue, could you please have a look? 🙏🏼

packages/nx/src/command-line/release/utils/git.ts
```diff
@@ -44,11 +49,65 @@ function escapeRegExp(string) {
 const SEMVER_REGEX =
   /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/g;
 
+/**
+ * Check if a semver string is for a stable release (i.e. no prerelease or build metadata)
+ * @param semver - The semver string to check
+ * @returns True if the semver string is stable, false otherwise
+ */
+export function isStableSemver(semver: string): boolean {
```
JamesHenry: There is already a `isPrerelease` utility in `packages/nx/src/command-line/release/utils/shared.ts`, let's not duplicate the logic

packages/nx/src/command-line/release/utils/git.ts
```diff
@@ -149,15 +208,34 @@ export async function getLatestGitTagForPattern(
...
+
+      if (preIdReleaseTags.length > 0) {
+        return extractTagAndVersion(preIdReleaseTags[0], tagRegexp);
+      }
+    } 
+    
+    // Then try to find the latest stable release tag
+    const stableReleaseTags = matchingSemverTags.filter((tag) => {
+      return !!tag.match(tagRegexp) &&
+        tag.match(tagRegexp).some(isStableSemver)
```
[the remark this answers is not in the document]
JamesHenry: Whilst the explanation isn't valid at least the suggestion avoided the double `.match` call, so given you're updating this anyway, please go with this

packages/nx/src/command-line/release/command-object.ts
```diff
@@ -65,6 +65,7 @@ export type ChangelogOptions = NxReleaseArgs &
     from?: string;
     interactive?: string;
     createRelease?: false | 'github' | 'gitlab';
+    preid?: VersionOptions['preid'];
```
JamesHenry: Supporting a new CLI argument into the release command takes this beyond a "fix" for versioning, this is a feature PR IMO

JamesHenry (changes requested):
Thanks for this @iuliancmarcu!

We need to reconcile this with https://github.com/nrwl/nx/pull/31794 because fundamentally (from my colleague) there should not be a requirement for semver in these tags as we move towards first-class application support.

I will try and land that PR first ASAP and then we can rebase this one on top of it

JamesHenry:
Ok https://github.com/nrwl/nx/pull/31794 is in, I'm sorry that it will require some trickier rebasing now, but it's important that we only build this out on top of the "semver path"

iuliancmarcu:
Hi @JamesHenry 👋🏼 Thanks a lot for looking over this, I will do the rebase and address your comments today, and let you know when it's ready for another look 👍🏼

packages/nx/src/command-line/release/command-object.ts
iuliancmarcu: `nx release version` options and `nx release` options already had `preId`, my feeling is that changelog was intended to have this option too, but it was omitted because it wasn't used in the actual logic until now.

Let me know if you'd like me to change the commit to be a `feat`, but I believe supporting `preId` is more of a fix to the `changelog` command (as a result of correctly applying it through `release`).

iuliancmarcu:
@JamesHenry this is ready for a second look. I rebased the latest changes, the AI suggestion and used the existing `isPrerelease` in the logic instead of the custom `isStableSemver`.

packages/nx/src/command-line/release/command-object.ts
iuliancmarcu: Will remove this, as I can infer the preid automatically from the release version 👍🏼

iuliancmarcu:
Hey @JamesHenry,

I need a bit of guidance here, if you could help.

I spent the last couple of days investigating the failing e2e tests (gotta be honest, running those tests and iterating on them is a pain 😅 ).

I traced the failures to the change I made in the `git` util. Previously, the `getLatestGitTagForPattern` would just return the most recent tag in the default (last resort) case, and I made it so that it follows these rules:
- if preid is specified, look for tags matching the pattern, that have the same preid. if none are found, look for latest stable tag. if none are found return null (fallback to disk or whatever)
- if no preid is specified, look for latest stable tag. if none are found return null (fallback to disk or whatever)

To make it more clear, here are some examples:

---

For these tags:
```
my-lib@0.0.1-alpha.0
```

|preid|Current logic tag|New logic tag|
|-|-|-|
|<none>|my-lib@0.0.1-alpha.0|null|
|alpha|my-lib@0.0.1-alpha.0|my-lib@0.0.1-alpha.0|
|beta|my-lib@0.0.1-alpha.0|null|

---

For these tags:
```
my-lib@0.0.2-alpha.0
my-lib@0.0.1
my-lib@0.0.1-alpha.0
```

|preid|Current logic tag|New logic tag|
|-|-|-|
|<none>|my-lib@0.0.2-alpha.0|my-lib@0.0.1|
|alpha|my-lib@0.0.2-alpha.0|my-lib@0.0.2-alpha.0|
|beta|my-lib@0.0.2-alpha.0|my-lib@0.0.1|

---

To me, this seems like the logical way to have it: stable looks for stable, pre-release looks for same pre-release or stable. The perspective that I am looking at this from is that if you have 2 devs working on PRs that use the branch name as `preid`, for them to not look at tags created by the other PRs, but at the latest stable.


What do you think about this? I am wondering if there are any hidden implications of this, and to treat this change:
- as is -> breaking change resulting in minor (major?) Nx release
- add a config for release to enable the new logic (`enforceReleaseTypeTag` or something) of always selecting same pre-release or stable
- other suggestions?

JamesHenry:
Thanks @iuliancmarcu yes it is indeed worth classing as a breaking change.  In this case, the reality is we don't have a bunch of people complaining about us currently doing the wrong thing. That could mean that the current behaviour is actually desirable or at least working well enough for those people.

Having said that, I do agree with your rationale for this change, so having it be behind an option which defaults to disabled (to make it non-breaking) is the way to go. We can then strongly consider making it the default behaviour in Nx v22, with anyone who depends on the old logic still able to leverage that if they want to.

Nx release is very flexible and dynamic, and so the configuration has to be too. For that reason, adding a new config option for this is a little involved, so I will push up the config option to the branch for you on Monday and then you can leverage it from there.

Sound good?

iuliancmarcu:
Hi @JamesHenry, that sounds great to me. Let me know when the config option is pushed and I'll update the PR to support both the old and new behaviour (defaulting to the old, as you mentioned). Many thanks!

JamesHenry:
@iuliancmarcu sorry for the delay pushed the config option now.

Please check you agree with my description of the flag and how it should work:

https://github.com/nrwl/nx/pull/31756/commits/739bc1b195ab0d27624852060e26b519ab025ce9

It does not yet apply the flag to any of your changes

iuliancmarcu:
It looks perfect, I'll work on integrating it today. Thank you for the help, @JamesHenry!

iuliancmarcu:
Hey @JamesHenry, I believe the PR is ready now.
Could you please approve the run for the e2e tests?

packages/nx/src/command-line/release/config/config.ts
```diff
@@ -345,6 +349,9 @@ export async function createNxReleaseConfig(
     releaseTagPatternRequireSemver:
       userConfig.releaseTagPatternRequireSemver ??
       defaultReleaseTagPatternRequireSemver,
+    releaseTagPatternStrictPreid:
```
meeroslav, line 352: Can we rename this to `releaseTagPatternStrictPreId`?

iuliancmarcu: I can do that, of course, just note that the CLI flag is also `--preid`, not `--preId`. Would you like me to go ahead with it?

JamesHenry: @meeroslav Our CLI option is `preid` which is consistent with the npm CLI and the `semver` library, so this option also uses `Preid` casing to be consistent again

iuliancmarcu:
@JamesHenry there seems to be an unrelated test failing for `storybook`. Do you have any suggestion on how we can move this forward?

Coly010:
@iuliancmarcu This is an issue coming from master. A fix will be merged soon and then you can rebase on top :)

e2e/release/src/release-tag-pattern.test.ts
```diff
@@ -61,6 +61,7 @@ describe('nx release releaseTagPattern', () => {
     updateJson<NxJsonConfiguration>('nx.json', (nxJson) => {
       nxJson.release = {
         releaseTagPattern: 'v{version}',
+        releaseTagPatternStrictPreid: false,
```
JamesHenry: Please remove this, it should not be necessary to opt out of the new behaviour

e2e/release/src/source-tag-selection.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    pkg1 = uniq('my-pkg-1');
+    runCLI(`generate @nx/workspace:npm-package ${pkg1}`);
+
+    await runCommandAsync(`git add .`);
+    await runCommandAsync(`git commit -m "chore: initial commit"`);
+  }, 60000);
+
+  afterEach(() => cleanupProject());
+
+  describe('when releaseTagPatternStrictPreid is false', () => {
```
JamesHenry, line 108: Please duplicate this block and cover when `releaseTagPatternStrictPreid` is not set (should be the same as false right now but will provide an expected if/when we change the default)

packages/nx/src/command-line/release/changelog.ts
```diff
@@ -1446,3 +1487,18 @@ function versionPlanSemverReleaseTypeToChangelogType(bump: ReleaseType): {
       throw new Error(`Invalid semver bump type: ${bump}`);
   }
 }
+
+function extractPreId(version: string): string | undefined {
```
JamesHenry: Your casing is inconsistent with the option, let's change that please (for the same reason Miro wanted it changed and we said no), we need to commit either way

packages/nx/src/command-line/release/changelog.ts
```diff
@@ -671,8 +703,17 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
               await getLatestGitTagForPattern(
                 releaseGroup.releaseTagPattern,
                 {},
-                releaseGroup.releaseTagPatternCheckAllBranchesWhen,
-                releaseGroup.releaseTagPatternRequireSemver
+                {
+                  checkAllBranchesWhen:
+                    releaseGroup.releaseTagPatternCheckAllBranchesWhen,
+                  preId:
```
JamesHenry: Casing

packages/nx/src/command-line/release/changelog.ts
```diff
@@ -243,6 +244,21 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
     const toSHA = await getCommitHash(to);
     const headSHA = to === 'HEAD' ? toSHA : await getCommitHash('HEAD');
 
+    /**
+     * Extract the preid from the workspace version and the project versions
+     */
+    const workspacePreId: string | undefined = workspaceChangelogVersion
```
JamesHenry: I won't keep repeating on every occurrence, but casing

packages/nx/src/command-line/release/changelog.ts
```diff
@@ -1446,3 +1487,18 @@ function versionPlanSemverReleaseTypeToChangelogType(bump: ReleaseType): {
...
+  }
+
+  const preId = prerelease(version)[0];
+  if (typeof preId === 'string') {
+    return preId;
+  }
+  if (typeof preId === 'number') {
+    return preId.toString();
+  }
+  return undefined;
```
JamesHenry: Can we protect against `prerelease(version)` not being defined or being null here (it shouldn't happen based on isPrerelease but better to not blow up if there was a logic mismatch.

We should probably also convert an empty string to `undefined` explicitly?

packages/nx/src/command-line/release/utils/git.ts
```diff
@@ -44,12 +57,61 @@ function escapeRegExp(string) {
...
+ * @param tagRegexp - The regex to use to extract the tag and version from the tag string
+ *
+ * @returns The tag and version
+ */
+export function extractTagAndVersion(
+  tag: string,
+  tagRegexp: string,
+  options: GetLatestGitTagForPatternOptions
+): GitTagAndVersion {
+  const { releaseTagPatternRequireSemver = true } = options;
```
JamesHenry: Important: There should not be any default values to config anywhere other than `config.ts` that is a fundamental design decision and why config.ts is so involved. The rest of the code can be assured that default values have been applied upfront and in one place

packages/nx/src/command-line/release/utils/git.ts
```diff
@@ -44,12 +57,61 @@ function escapeRegExp(string) {
...
-  releaseTagPatternRequireSemver: boolean = true
-): Promise<{ tag: string; extractedVersion: string } | null> {
+  options: GetLatestGitTagForPatternOptions
+): Promise<GitTagAndVersion | null> {
+  const {
+    checkAllBranchesWhen,
+    releaseTagPatternRequireSemver = true,
+    releaseTagPatternStrictPreid,
+    preId = '',
+  } = options;
```
JamesHenry: This is the same point as above, and I had missed it on a previous PR review, we shouldn't have default values in here

iuliancmarcu:
@JamesHenry Thanks for the review! I just pushed commits addressing your comments 👍🏼

JamesHenry (approved):
Thanks a lot for you work on this!

PR #31759, opened 2025-06-27
fix(misc): add fallback for missing source root

### LINKED ISSUE #30638: [createGlobPatternsForDependencies] Failed to generate glob pattern
### Current Behavior

#30111 and specifically the one line change [here](https://github.com/nrwl/nx/commit/ada8be473d5ddd50117145bb1c96c536170baf6a#diff-e5e0798fc4c3a06f0b677f8a582330e66bccff59d2c372a7904e1f87789867f7L212) that removes `sourceRoot` from the package-json plugins projectConfiguration response has resulted in all calls to `createGlobPatternsForDependencies` from `@nx/react/tailwind` failing with below log for us.

This is because createGlobPatternsForDependencies seems to require sourceRoot be set for all nodes as seen below. Without `sourceRoot`, createGlobPatternsForDependencies will add all projects to the warning response and not generate globs for them.

https://github.com/nrwl/nx/blob/7b85d912bab825691cbfd1fc11cf88bb8e0c056c/packages/nx/src/utils/project-graph-utils.ts#L40

### Expected Behavior

Glob patterns are returned rather than an empty array.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Setup a package.json based repo.
2. Call createGlobPatternsForDependencies.
3. Observe no globs returned.

### Nx Report

```shell
Node           : 22.1.0
OS             : darwin-arm64
pnpm           : 10.5.1

nx                     : 20.7.1
@nx/js                 : 20.7.1
@nx/eslint             : 20.7.1
@nx/workspace          : 20.7.1
@nx/devkit             : 20.7.1
@nx/eslint-plugin      : 20.7.1
@nx/module-federation  : 20.7.1
@nx/react              : 20.7.1
@nx/web                : 20.7.1
typescript             : 5.7.2

Nx key licensed packages
Licensed to _.

@nx/azure-cache  : 1.3.1

Registered Plugins:
@nx/eslint/plugin
Community plugins:
@nx/azure-cache : 1.3.1
Cache Usage: 543.89 KB / 92.64 GB
```

### Failure Logs

```shell
[createGlobPatternsForDependencies] Failed to generate glob pattern for the following:
@company/package1
 -@company/package2
due to missing "sourceRoot" in the dependencies' project configuration
```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Have fixed this temporarily by pnpm patching the `sourceRoot` line back in to `packages/nx/src/plugins/package-json/create-nodes.ts`.

```diff
const projectConfiguration: ProjectConfiguration & { name: string } = {
     root: projectRoot,
+    sourceRoot: projectRoot,
     name,
```

leosvelperez:
## Current Behavior

The `sourceRoot` project configuration property is optional. Several places in the codebase do not properly handle this, which can result in issues.

## Expected Behavior

A missing `sourceRoot` project configuration property should be handled correctly throughout the codebase.

## Related Issue(s)

Fixes #30638

### DIFF
--- a/packages/js/src/utils/typescript/ts-solution-setup.ts
+++ b/packages/js/src/utils/typescript/ts-solution-setup.ts
@@ -3,6 +3,7 @@ import {
   joinPathFragments,
   offsetFromRoot,
   output,
+  type ProjectConfiguration,
   readJson,
   readNxJson,
   type Tree,
@@ -290,14 +291,15 @@ export function getProjectType(
 }
 
 export function getProjectSourceRoot(
-  tree: Tree,
-  projectSourceRoot: string | undefined,
-  projectRoot: string
-): string | undefined {
+  project: ProjectConfiguration,
+  tree?: Tree
+): string {
+  tree ??= new FsTree(workspaceRoot, false);
+
   return (
-    projectSourceRoot ??
-    (tree.exists(joinPathFragments(projectRoot, 'src'))
-      ? joinPathFragments(projectRoot, 'src')
-      : projectRoot)
+    project.sourceRoot ??
+    (tree.exists(joinPathFragments(project.root, 'src'))
+      ? joinPathFragments(project.root, 'src')
+      : project.root)
   );
 }
--- a/packages/js/src/utils/typescript/ts-solution-setup.ts
+++ b/packages/js/src/utils/typescript/ts-solution-setup.ts
@@ -10,6 +10,7 @@ import {
   updateJson,
   workspaceRoot,
 } from '@nx/devkit';
+import { existsSync } from 'node:fs';
 import { basename, dirname, join } from 'node:path/posix';
 import { FsTree } from 'nx/src/generators/tree';
 import {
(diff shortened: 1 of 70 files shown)


Coly010 (approved):

packages/js/src/utils/typescript/ts-solution-setup.ts
```diff
@@ -290,14 +291,15 @@ export function getProjectType(
...
 
 export function getProjectSourceRoot(
-  tree: Tree,
-  projectSourceRoot: string | undefined,
-  projectRoot: string
-): string | undefined {
+  project: ProjectConfiguration,
+  tree?: Tree
+): string {
+  tree ??= new FsTree(workspaceRoot, false);
```
FrozenPandaz: Not really a fan of creating a FsTree just for this. Is this `tree` even needed? I think we can use 2 different `exists` functions here rather than creating a tree?

FrozenPandaz (changes requested):

packages/js/src/utils/typescript/ts-solution-setup.ts
leosvelperez: The `tree` argument is needed when generators consume this function because some changes might be on the virtual tree and have not yet been committed to the file system. In that case, it's better to just forward the `tree` rather than requiring generators to provide a function that uses the `tree` to check if a file exists.

When a tree is not provided, we don't "need" it. In that case, it's being created and used as an abstraction over the file system. I'll change that so the function internally defines the two different `exists` functions.

packages/workspace/src/generators/utils/project-config.ts
```diff
@@ -0,0 +1,21 @@
+import {
+  joinPathFragments,
+  workspaceRoot,
+  type ProjectConfiguration,
+  type Tree,
+} from '@nx/devkit';
+import { FsTree } from 'nx/src/generators/tree';
+
+export function getProjectSourceRoot(
```
FrozenPandaz, line 10: Please fix this function too and it's good to go.

FrozenPandaz (approved):

packages/workspace/src/generators/utils/project-config.ts
leosvelperez: Ah, missed that one 🤦🏻

PR #31780, opened 2025-06-30
fix(core): ignore type imports in generatePackageJson

### LINKED ISSUE #31263: generatePackageJson includes type imports in package.json
### Current Behavior

If I have a package in my root package.json, and in my service I only import from it using type imports (`import type { A } from “package”`) the package will still be included in the generated package json, even though it’s not needed at runtime.

I thought this was fixed as this issue was resolved: https://github.com/nrwl/nx/issues/13462

### Expected Behavior

NX should only include dependencies that are needed at runtime, not ones that are imported using type imports.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create a simple nx monorepo and install react for example
2. in your app, configure the project.json to have `generatePackageJson: true`
3.  have this line in your main.ts: `import type { ReactElement } from “react”;`
4. build your app and check if react is present in your dist’s package.json

### Nx Report

```shell
Node           : 18.14.1
OS             : win32-x64
npm            : 9.3.1

nx (global)    : 21.0.3
nx             : 20.8.1
@nx/js         : 20.8.1
@nx/jest       : 20.8.1
@nx/eslint     : 20.8.1
@nx/workspace  : 20.8.1
@nx/devkit     : 20.8.1
@nx/nest       : 20.8.1
@nx/node       : 20.8.1
@nx/web        : 20.8.1
@nx/webpack    : 20.8.1
typescript     : 5.7.3
Registered Plugins:
@nx/webpack/plugin
Cache Usage: 0.00 B / 95.25 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Adig0:
## Current Behavior
When using generatePackageJson, the resulting package.json includes packages that are only referenced as type-only imports (e.g., import type { X } from 'some-package'). These type-only imports should not appear in the final package.json since they are erased during compilation and not required at runtime.
This leads to unnecessary dependencies being installed in production, which bloats deployments and increases cold start time.

## Expected Behavior
generatePackageJson should omit packages that are only used in type-only imports. This ensures the package.json includes only the actual runtime dependencies, avoiding unnecessary installs and potential dependency mismatches.

## Related Issue(s)
Fixes #31263


Adig0:
@FrozenPandaz Sorry for the tag, can we have anyone take a look at this PR?
Would love to see it merged!

jaysoo (changes requested):
Good direction — filtering type-only imports from generated `package.json` is the right goal. A few issues to address:

1. **Misses inline type imports**: This only handles `import type { Foo } from 'pkg'` but not the very common `import { type Foo } from 'pkg'` (inline type specifiers, available since TS 4.5). Both forms should be treated equivalently.

2. **Public API change**: Adding a new `DependencyType.type` enum value is a public API change that needs team buy-in, as external tools may rely on the existing enum values.

3. **Filtering logic**: The dependency filtering is brittle — a package that has both type and value imports should still be included. The logic needs to check whether ALL imports from a package are type-only before excluding it.

Adig0:
Thanks for the suggestion @jaysoo !
After digging a bit deeper, I realized this change may be problematic due to how TypeScript treats type imports under `verbatimModuleSyntax`.

According to the [TypeScript docs](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax), type-only named imports are not always erased at runtime. For example:

```ts
// Erased entirely
import type { A } from "a";

// Rewritten to: import { b } from "bcd";
import { b, type c, type d } from "bcd";

// Rewritten to: import {} from "xyz";
import { type xyz } from "xyz";
```

So even in cases like:
```ts
import { type A } from "pkg";
```

TypeScript may still emit an `import {} from "pkg"` to preserve module side effects.

Because of that, implementing the requested change would diverge from TypeScript’s actual emit behavior and could result in dropping dependencies that are still required at runtime.

While this behavior would be convenient for some use cases (including ours), I’m not confident it would be correct or safe for all Nx users.

Happy to discuss further if you think there’s a different angle I’m missing 👍

jaysoo:
I'll close this for now. Happy to discuss further if there is a good solution here.

PR #31785, opened 2025-06-30
fix(node): reorder addPlugin in normalizeOptions return object

ndcunningham:
When creating a node app If you want plugins to be added and  `applicationGeneratorInternal` is called instead of `applicationGenerator` the option will be overwritten by the default `false`.

### DIFF
--- a/packages/node/src/generators/application/lib/normalize-options.ts
+++ b/packages/node/src/generators/application/lib/normalize-options.ts
@@ -33,8 +33,9 @@ export async function normalizeOptions(
 
   const nxJson = readNxJson(host);
   const addPlugin =
-    process.env.NX_ADD_PLUGINS !== 'false' &&
-    nxJson.useInferencePlugins !== false;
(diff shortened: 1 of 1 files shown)


Coly010 (approved):

packages/node/src/generators/application/lib/normalize-options.ts
```diff
@@ -44,8 +44,8 @@ export async function normalizeOptions(
   const useProjectJson = options.useProjectJson ?? !isUsingTsSolutionConfig;
 
   return {
-    addPlugin,
     ...options,
+    addPlugin,
```
FrozenPandaz, line 49: How do you overwrite this based on `options` then? :thinking: Isn't the way it was before the way it is everywhere?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #31824, opened 2025-07-03
feat(core): allow {args} to be fully interpolated in run-commands

Coly010:
## Current Behavior
Currently, there is no way for a target using `run-commands` to define where in the command args are attached.
This is problematic in some tooling cases where args positional location matters

## Expected Behavior
Placing `{args}` into the command should allow for interpolation of any and all args provided.
Therefore commands can be written such as `docker run {args} imageRef`

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.ts
@@ -231,6 +231,11 @@ export function interpolateArgsIntoCommand(
   >,
   forwardAllArgs: boolean
 ): string {
+  if (command.indexOf('{args.') > -1 && command.indexOf('{args}') > -1) {
+    throw new Error(
+      'Command should not contain both {args} and {args.*} values. Please choose one to use.'
+    );
+  }
   if (command.indexOf('{args.') > -1) {
     const regex = /{args\.([^}]+)}/g;
     return command.replace(regex, (_, group: string) =>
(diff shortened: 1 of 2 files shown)


packages/nx/src/executors/run-commands/run-commands.impl.ts
```diff
@@ -236,18 +236,18 @@ export function interpolateArgsIntoCommand(
...
       opts.parsedArgs[group] !== undefined ? opts.parsedArgs[group] : ''
     );
+  } else if (command.indexOf('{args}') > -1) {
+    const regex = /{args}/g;
+    const args = [
+      ...unknownOptionsToArgsArray(opts),
+      ...unparsedOptionsToArgsArray(opts),
+    ];
+    const argsString = `${args.join(' ')} ${opts.args ?? ''}`;
+    return command.replace(regex, argsString);
```
FrozenPandaz, line 251: Do you think it makes sense to have both `{args.something} {args}`?

If not, we should probably error somewhere.

If so, we should probably do both? 🤷 So this would have to change from an `if + if else`.

FrozenPandaz (changes requested):

packages/nx/src/executors/run-commands/run-commands.impl.ts
Coly010: We should error - passing `{args}` explicitly states you want to print all of them. having additional `{args.*}` would cause repeated values and could be breaking

FrozenPandaz (approved):

PR #31839, opened 2025-07-06
fix(react-native): app creation should sync deps

xiongemi:
## Current Behavior

currently, when options.install=true (need to do pod install), it only syncs deps.

## Expected Behavior

make it sync deps regardless of options.install, so when developers choose to run `pod-install` after, it will work right away.

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/react-native/src/react-native.test.ts
+++ b/e2e/react-native/src/react-native.test.ts
@@ -10,6 +10,7 @@ import {
   checkFilesExist,
   runE2ETests,
   updateFile,
+  readJson,
 } from 'e2e/utils';
 
 describe('@nx/react-native', () => {
@@ -43,6 +44,25 @@ describe('@nx/react-native', () => {
   it('should test and lint', async () => {
     expect(() => runCLI(`test ${appName}`)).not.toThrow();
     expect(() => runCLI(`lint ${appName}`)).not.toThrow();
+  });
+
+  it('should have dependencies synced after React Native app creation', () => {
+    // Check that the app's package.json exists
+    checkFilesExist(`${appName}/package.json`);
+
+    // Read the app's package.json
+    const appPackageJson = readJson(`${appName}/package.json`);
+
+    // Verify that the app package.json has dependencies section
+    expect(appPackageJson.dependencies).toBeDefined();
+
+    // Verify that React Native specific dependencies are synced
+    expect(appPackageJson.dependencies).toEqual(
+      expect.objectContaining({
+        react: '*',
+        'react-native': '*',
+      })
+    );
   });
 
   it('should bundle the app', async () => {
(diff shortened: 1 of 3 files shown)


packages/react-native/src/generators/application/application.ts
```diff
@@ -170,3 +141,47 @@ export async function reactNativeApplicationGeneratorInternal(
 }
 
 export default reactNativeApplicationGenerator;
```
ndcunningham: Would it be better if the export is the last?

packages/react-native/src/generators/application/application.ts
```diff
@@ -108,37 +108,8 @@ export async function reactNativeApplicationGeneratorInternal(
...
-    tasks.push(podInstallTask);
-  } else {
-    output.log({
-      title: 'Skip `pod install`',
-      bodyLines: [
-        `run 'nx run ${options.name}:pod-install' to install native modules before running iOS app`,
-      ],
-    });
-  }
+  tasks.push(addSyncDepsTask(host, options));
```
ndcunningham, line 111: What will be the purpose of `options.install`? 
Do we no longer need it? Also, would this make the installation process longer for things like CI for example if its not behind a flag?

packages/react-native/src/generators/application/application.ts
xiongemi: moved

packages/react-native/src/generators/application/application.ts
xiongemi: i moved it inside addPodInstallTask. this still controls pod install. just sync deps will always run.

e2e/react-native/src/react-native.test.ts
```diff
@@ -45,6 +46,25 @@ describe('@nx/react-native', () => {
...
+    // Verify that the app package.json has dependencies section
+    expect(appPackageJson.dependencies).toBeDefined();
+
+    // Verify that React Native specific dependencies are synced
+    expect(appPackageJson.dependencies).toEqual(
+      expect.objectContaining({
+        react: '*',
+        'react-native': '*',
+      })
+    );
```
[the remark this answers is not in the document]
xiongemi: version is * in this case.

packages/react-native/src/generators/application/application.ts
```diff
@@ -170,3 +141,47 @@ export async function reactNativeApplicationGeneratorInternal(
...
+    'package.json'
+  );
+
+  const workspacePackageJson = readJson<PackageJson>(host, 'package.json');
+  const projectPackageJson = readJson<PackageJson>(
+    host,
+    projectPackageJsonPath
+  );
+
+  return async () => {
```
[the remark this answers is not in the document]
xiongemi: changed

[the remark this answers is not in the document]
xiongemi: changed

ndcunningham (approved):

PR #31845, opened 2025-07-07
feat(angular): support angular v20.1

leosvelperez:
## Current Behavior

Angular v20.1 is not supported.

## Expected Behavior

Angular v20.1 should be supported.


packages/nx/file.txt
```diff
@@ -0,0 +1 @@
+hello world
```
FrozenPandaz: @claude please remove this.

FrozenPandaz (approved):

PR #31865, opened 2025-07-09
docs(nx-dev): remove problematic GitHub parallelization

meeroslav:
The existing parallelization example does not work correctly due to the encapsulated run_command swallowing the status codes and passing the runs with a failed command.

If users are interested in how to parallelize the tasks they can easily find it today. We should provide minimal example like we do for other CI providers.

## Related Issue(s)

Fixes #


rarmatei (approved):
yey 👍

PR #31889, opened 2025-07-10
fix(core): upgrade glob version to 10

xiongemi:
## Current Behavior

current got this warning:
```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```

also there is duplicate entries of tinyglobby, one in dependencies and one in devDependencies

## Expected Behavior

upgrade glob version to 10
remove tinyglobby from devDependencies, keep the one in depedencies

## Related Issue(s)

Fixes #


JamesHenry (changes requested):
Can we instead use this opportunity to just exclusively be using tiny globby and picomatch?

We are most of the way through transitioning:

image

Is there any reason to have two globbing libraries?

PR #31897, opened 2025-07-11
feat(core): add live durations for running tasks in the tui

leosvelperez:
## Current Behavior

The TUI doesn't display live durations for running tasks.

image

## Expected Behavior

The TUI should display live durations for running tasks.

image


packages/nx/src/native/tui/components/tasks_list.rs
```diff
@@ -716,9 +718,18 @@ impl TasksList {
 
     /// Updates their status to InProgress and triggers a sort.
     pub fn start_tasks(&mut self, tasks: Vec<Task>) {
+        let current_time = get_current_time_ms();
         for task in tasks {
             if let Some(task_item) = self.tasks.iter_mut().find(|t| t.name == task.id) {
                 task_item.update_status(TaskStatus::InProgress);
+                if task_item.start_time.is_none() {
+                    // It should be set, but just in case
+                    task_item.start_time = Some(current_time);
```
FrozenPandaz, line 727: Move defining current time into here.

packages/nx/src/native/tui/components/tasks_list.rs
```diff
@@ -716,9 +718,18 @@ impl TasksList {
...
+                }
+                // Update duration to show "..." initially for non-continuous tasks
+                if !task_item.continuous {
+                    task_item.duration = DURATION_NOT_YET_KNOWN.to_string();
```
FrozenPandaz, line 731: I don't think we need to store this value in state. I think we only need to store the start time.. and ensure that it is there. (it should be as you said).

I think the rest of this becomes obsolete if we calculate the elapsed time during the drawing of the duration column. So it naturally will not happen if we do not draw the column.

FrozenPandaz (changes requested):

AgentEnder (approved):

packages/nx/src/native/tui/components/tasks_list.rs
FrozenPandaz: @claude can you move defining current_time into here please?

FrozenPandaz:
@claude please fix the commit message so that it conforms to standards.

packages/nx/src/native/tui/components/tasks_list.rs
leosvelperez: Like we discussed offline, we'll keep it as is because moving the calculation inside the rendering would result in many needless calculations.

FrozenPandaz (approved):

PR #31943, opened 2025-07-16
chore(core): use carrot semver constraint for `@napi-rs/wasm-runtime`

aqeelat:
Just a simple change.
changelog: https://github.com/napi-rs/napi-rs/blob/main/wasm-runtime/CHANGELOG.md

I could change it to `@^0.2.12` if you think it is better.


Coly010:
I think this is fine as-is and we can change it new versions in a more nuanced fashion rather than potentially allowing unexpected versions to be resolved.

PR #31968, opened 2025-07-18
chore(repo): update nx to 21.3.0-rc.0

FrozenPandaz:
Updating Nx from 21.3.0-beta.7 to 21.3.0-rc.0


package.json
```diff
@@ -64,7 +64,7 @@
     "@jest/test-result": "30.0.2",
     "@jest/types": "30.0.1",
     "@ltd/j-toml": "^1.38.0",
-    "@module-federation/enhanced": "^0.17.0",
+    "@module-federation/enhanced": "0.17.0",
     "@module-federation/sdk": "^0.17.0",
```
Julien-Marcou: Is there a reason to no longer pin `@module-federation/enhanced` with a caret `^` like we do for `@module-federation/sdk`

For alpha releases (<1.0.0), the caret `^` will only allow to update the package for minor fixes, so `^0.17.0` is equal to `>=0.17.0 <0.18.0`

I think it would be nice to keep the caret here

FrozenPandaz: Good point.

@claude can you please restore the caret here and update the pnpm lock file?

AgentEnder (approved):

PR #31989, opened 2025-07-20
fix(gradle): add dep with build file

xiongemi:
## Current Behavior

got and error:
```
Error: project spring-boot-integration-tests not found
```

the issue was that for some gradle project, it does not have a build file, but it is a still valid gradle project.
in create nodes, we only look at the build file to create a node for each build file detected

## Expected Behavior

- do not add dep for project with build file

## Related Issue(s)

Fixes #


lourw:
Already merged as of https://github.com/nrwl/nx/pull/32463

PR #31999, opened 2025-07-21
feat(nx): use platform certificate trust store

wereHamster:
## Current Behavior

Certificates in the local trust store are ignored, because reqwest (rustls) is using webpki-roots.

## Expected Behavior

Certificates from the local trust store are trusted.

This PR solves this the same way as turborepo - by turning on the `rustls-tls-native-roots` feature in reqwest.


wereHamster:
@FrozenPandaz @vsavkin I don't understand the CI failures, they don't look related to my change.

AgentEnder (approved):

AgentEnder (approved):

PR #32021, opened 2025-07-22
fix(core): move bin and scripts fields to nxConfig in package.json

ndcunningham:
### Currently

When we run `pnpm i` install the Nx repo we get logs and warnings due to the local nx package running it's post-install script

```shell
 WARN  Failed to create bin at /../nx/node_modules/.bin/nx. ENOENT: no such file or directory, open '../packages/nx/bin/nx.js'
 WARN  Failed to create bin at ../e2e/plugin/node_modules/.bin/nx-cloud. ENOENT: no such file or directory, open '../packages/nx/bin/nx-cloud.js'
 WARN  Failed to create bin at /../e2e/angular/node_modules/.bin/nx. ENOENT: no such file or directory, open '/../packages/nx/bin/nx.js'
```

### Expected

We should restrict local packages post-install from being run in this context.

### DIFF
--- a/packages/create-nx-workspace/package.json
+++ b/packages/create-nx-workspace/package.json
@@ -20,9 +20,6 @@
     "Playwright",
     "CLI"
   ],
-  "bin": {
-    "create-nx-workspace": "./bin/index.js"
-  },
   "author": "Victor Savkin",
   "license": "MIT",
   "bugs": {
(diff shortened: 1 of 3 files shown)


Coly010:
Let's revisit this later if needed, changing the package.json bin locations might cause adverse affects in the future

PR #32028, opened 2025-07-22
chore(repo): isolate angular tests which seem to take more memory

FrozenPandaz:
## Current Behavior

Angular tests run with other tests in parallel causing memory issues.

## Expected Behavior  

Angular tests are isolated to run on their own agent to prevent memory issues.

## Related Issue(s)

This is a maintenance improvement to prevent CI memory issues.

Fixes nrwl/nx internal issue


nx.json
```diff
@@ -283,7 +283,7 @@
   "nxCloudId": "62d013ea0852fe0a2df74438",
   "nxCloudUrl": "https://staging.nx.app",
   "parallel": 1,
-  "bust": 3,
+  "bust": 6,
```
FrozenPandaz: @claude Remove this change please and amend the last commit.

AgentEnder (approved):

PR #32035, opened 2025-07-23
feat(bundling)!: remove stylus support

ndcunningham:
BREAKING CHANGE: Stylus (.styl) files are no longer supported.

Stylus has been deprecated and was slated to be removed in Nx 20. 
Now we are removing it.

- Remove `stylus` and `stylus-loader` dependencies
- Remove `deprecated-stylus-loader` file
- Remove `stylus` configuration from `webpack` and `rspack` plugins
- Remove `'styl'` option from `rspack` generator schemas and TypeScript definitions
- Remove `.styl`/`.stylus` extensions from executor schema file completion globs

MISC
  Even though we removed stylus from our repo as a dependency other projects in the workspace still have stylus as an optional dependency:
  - Vite
  - Astro
  
  So we add a placeholder for it since on npm registry there is a _security_ placeholder.


jaysoo (approved):

packages/webpack/src/plugins/nx-webpack-plugin/lib/apply-web-config.ts
fedesa92, line 1: oh shit, here we go again

PR #32063, opened 2025-07-24
feat(js): ignore scoped export conditions

aidant:
There is an emerging pattern of using scoped custom export conditions for pointing to source. See:
- [Announcing TypeScript 5.7](https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#path-rewriting-for-relative-paths:~:text=As%20a%20result%2C%20if%20you%E2%80%99ve%20been%20using%20a%20workspace%2Dstyle%20layout%20with%20multiple%20packages%20referencing%20each%20other%2C%20you%20might%20need%20to%20use%20conditional%20exports%20with%20scoped%20custom%20conditions%20to%20make%20this%20work%3A)
- [Live types in a TypeScript monorepo](https://colinhacks.com/essays/live-types-typescript-monorepo#:~:text=Here%27s%20how%20a%20custom%20condition%20might%20look%20in%20your%20package.json.%20There%27s%20nothing%20special%20about%20the%20string%20%22%40colinhacks/zod%22%20here!%20It%20could%20be%20anything.)
- [Loading from Source](https://github.com/isaacs/tshy?tab=readme-ov-file#loading-from-source)

## Current Behavior

When including an export condition that points to the source code the typescript plugin does not add a build target. However, if this export condition is named `development` the build target is added.

## Expected Behavior

Scoped export conditions pointing at source should not remove the build target.

### DIFF
--- a/packages/js/src/plugins/typescript/util.ts
+++ b/packages/js/src/plugins/typescript/util.ts
@@ -113,7 +113,8 @@ export function isValidPackageJsonBuildConfig(
           return !isPathSourceFile(dotExport);
         } else if (typeof dotExport === 'object') {
           const hasMatch = Object.entries(dotExport).some(([key, value]) => {
-            if (key === 'types' || key === 'development') return false;
+            if (key === 'types' || key === 'development' || key.startsWith('@'))
+              return false;
             return typeof value === 'string' && isPathSourceFile(value);
           });
           return !hasMatch;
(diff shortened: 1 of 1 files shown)


jeffora:
We are encountering this issue where nx's typescript plugin is not adding the build target for our projects that are using a scoped export condition.

Is this a pattern nx would consider adopting @ndcunningham ? Any chance of a review or feedback on this approach?

Coly010 (approved):

PR #32072, opened 2025-07-24
feat(core): log out-of-sync details returned by sync generators when running `nx sync:check`

leosvelperez:
## Current Behavior

The `@nx/js:typescript-sync` generator does not provide details about the out-of-sync files. It only returns a generic message.

image

## Expected Behavior

The `nx sync:check` command should display additional details about the out-of-sync files if provided by the sync generators. The `@nx/js:typescript-sync` generator should provide details about the out-of-sync files.

image

Note:  the `nx sync` command will still only display the generic message to avoid cluttering the logs and the details can be seen in the changes made to files.

### DIFF
--- a/packages/nx/src/utils/sync-generators.ts
+++ b/packages/nx/src/utils/sync-generators.ts
@@ -25,6 +25,7 @@ import chalk = require('chalk');
 export type SyncGeneratorResult = void | {
   callback?: GeneratorCallback;
   outOfSyncMessage?: string;
+  outOfSyncDetails?: string[];
 };
 
 export type SyncGenerator = (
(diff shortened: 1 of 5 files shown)


packages/nx/src/utils/sync-generators.ts
```diff
@@ -272,6 +278,10 @@ export function getSyncGeneratorSuccessResultsMessageLines(
         result.outOfSyncMessage ?? `Some files are out of sync.`
       }`
     );
+
+    if (logOutOfSyncDetails && result.outOfSyncDetails?.length) {
+      messageLines.push(...result.outOfSyncDetails);
+    }
```
FrozenPandaz, line 284: This is sort of unrelated but can we also add some instruction to run `nx sync` to actually correct these check failures?

1. I think it's good to lead users.
2. I think AI will pick it up and make this easier to fix for AI :shrug:

FrozenPandaz (approved):

PR #32114, opened 2025-07-28
feat(testing): add atomizedInputs for playwright executor when running with atomizer

jordanpowell88:
## Current Behavior
Today, when we split our tests into separate targets with atomizer they all depend the other split tests because they are using `default` inputs. Their are 2 pretty common scenarios where I see this as a problem:

1. When you open a PR which has a failing test. Either Self Healing CI or a developer locally fix the test and push a fix for that single test it will rerun all the other tests even though the others weren't affected.

2. If you have to rerun a PR because of flakiness in any of those tests it would need to rerun all of them instead of just the flaky tests.

## Expected Behavior
If I have 2 different tests that get atomized and don't have a dependency between each other they should be separate targets that are not affected by each other.

_**Note**: This adds recursive checking of each specs relative imports and their relative imports, etc. We need to validate this doesn't have a negative impact on graph performance_

## Related Issue(s)

Fixes #


JamesHenry (commented):
I don't have a perfect answer, but I don't love the name `granularInputs` to me it sounds like a core feature rather than something specific to atomized tests...

Maybe `specInputs` or `atomizedInputs`? Or something that ties it a bit more closely to this case

jordanpowell88:
> I don't have a perfect answer, but I don't love the name `granularInputs` to me it sounds like a core feature rather than something specific to atomized tests...
> 
> Maybe `specInputs` or `atomizedInputs`? Or something that ties it a bit more closely to this case

I went ahead with `atomizedInputs` but I'm open to any other recommendations on naming. IMO the ideal state is for this to be the default behavior but wanted to play it safe in the short term

Issue #1992, opened 2019-10-25
[feature] nest hot reload


yassernasc:
## Expected Behavior

Out-of-box webpack hot reload on created nest apps.

## Current Behavior

Needed to add webpack manually following the tutorial: https://docs.nestjs.com/techniques/hot-reload.

flogh:
This feature would be incredible

FrozenPandaz:
Hi, sorry about this.

This was mislabeled as stale. We are testing ways to mark _not reproducible_ issues as stale so that we can focus on actionable items but our initial experiment was too broad and unintentionally labeled this issue as stale.

Nightbr:
Hey, would love to see a solution for this.
BTW I found this on Stackoverflow -> https://stackoverflow.com/questions/60476245/how-to-setup-webpacks-hmr-in-nestjs-project-in-nx-monorepo

webberwang:
any updates on this?

rubennaatje:
@yassernascimento Any way you could show how you did it manually?

yassernascimento:
@rubennaatje didn't tried yet, sorry

pacexy:
any news?

jensbodal:
I’ve tried a few times to get this working and it’s just no go.  I’m not even sure how well the feature works outside of the nest cli, does anyone have any experience?

mehrad-rafigh:
Unstale please

MinJungHyun:
please.. help

MinJungHyun:
https://github.com/typeorm/typeorm/issues/3017

DeepaPrasanna:
Are there any updates regarding this? I am facing this same issue.

antischematic:
Until this is added here's how to get HMR working with Nx 16.1.4.

Create a `webpack.hmr.config` file like this:

```js
const { composePlugins, withNx } = require('@nx/webpack');
const nodeExternals = require('webpack-node-externals');
const webpack = require('webpack')

// Set true if you don't want type checking
const skipTypeChecking = false

// Nx plugins for webpack.
module.exports = composePlugins(withNx({ skipTypeChecking }), (config) => {
  // Update the webpack config as needed here.
  // e.g. `config.plugins.push(new MyPlugin())`
  return {
    ...config,
    entry: ['webpack/hot/poll?100', ...config.entry.main],
    externals: [
      nodeExternals({
        allowlist: ['webpack/hot/poll?100'],
      }),
    ],
    plugins: [
[...]

PR #32133, opened 2025-07-29
fix(rspack): abide by rspacks watch configuration

yharaskrik:
## Current Behavior

As it currently stands, there is no (easy) way to get HMR support working with Rspack + Nest by using the `@nx/rspack:rspack` build executor and `@nx/js:node` serve executor because we need to turn `watch: false` in the serve executor (as we don't want it to watch and call the build again) but that in turn makes the rspack build executor not use `compiler.watch` this only building once.

Due to the way options are passed through to the executor (via `runExecutor`) even if we do:

```
"options": {
    "buildTarget": "myapp:build",
    "buildTargetOptions": {
        "watch": true
    },
    "watch": false
}
```

The `watch: false` overwrites it.

## Expected Behavior

We should be able to have the Rspack executor abide by whatever the `watch` configuration is that is returned from our `rspack.config.ts` file for the application.

There is precedent for abiding by the tools configuration file vs the executor option, the Webpack executor already does it.

The Webpack compiler is started in watch mode if the `watch` flag is true in the webpack config itself vs. strictly abiding by the executors options. Based on https://github.com/nrwl/nx/issues/1992#issuecomment-2826664777 I could do what I want, but I am using Rspack not Webpack.

## Related Issue(s)

https://github.com/nrwl/nx/issues/1992

Fixes #

### DIFF
--- a/packages/rspack/src/executors/rspack/rspack.impl.ts
+++ b/packages/rspack/src/executors/rspack/rspack.impl.ts
@@ -47,7 +47,9 @@ export default async function* runExecutor(
     success: boolean;
     outfile?: string;
   }>(async ({ next, done }) => {
-    if (options.watch) {
+    const watch = compiler instanceof Compiler ? compiler.options.watch : compiler.options[0].watch;
+
+    if (watch) {
       const watcher = compiler.watch(
         {},
         async (err, stats: Stats | MultiStats) => {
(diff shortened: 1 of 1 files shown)


Coly010 (approved):

Coly010:
@yharaskrik Made one minor adjustment to continue to evaluate `options.watch` in the circumstances when the rspack config file does not contain the `watch` option at all but users still pass `build --watch`

If this requires a further update to the serve, then we can update that too.

yharaskrik:
> @yharaskrik Made one minor adjustment to continue to evaluate `options.watch` in the circumstances when the rspack config file does not contain the `watch` option at all but users still pass `build --watch`
> 
> If this requires a further update to the serve, then we can update that too.

Amazing thank you!

PR #32169, opened 2025-07-31
fix(release): support "nx relase changelog" for docker images

jaysoo:
## Current Behavior
Running `nx release changelog [version]` fails for docker images

## Expected Behavior
it should work

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/release/changelog.ts
+++ b/packages/nx/src/command-line/release/changelog.ts
@@ -831,6 +831,7 @@ function resolveChangelogVersions(
       for (const projectName of releaseGroupProjectNames) {
         if (!args.versionData) {
           versionData[projectName] = {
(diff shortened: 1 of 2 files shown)


Coly010 (approved):

JamesHenry (changes requested):
Let's update `resolveChangelogVersions()`

We are in control of the project version data being generated in this case:

```ts
        if (!args.versionData) {
          versionData[projectName] = {
            newVersion: args.version,
            currentVersion: '', // not relevant within changelog/commit generation
            dependentProjects: [], // not relevant within changelog/commit generation
          };
          continue;
        }
```

So we shouldn't work around it, we should just set `dockerVersion: args.version`

JamesHenry (approved):

PR #32191, opened 2025-08-01
fix(docker): generate dockerfile compliant with package manager

Coly010:
## Current Behavior
We currently generate a Dockerfile for node applications that only takes into account npm.
This means that if the workspace is using yarn or pnpm then when we prune the lockfile, we have an incorrect lockfile in the output folder for the docker build.

## Expected Behavior
Ensure the generated Dockerfile is compliant for the package manager used in the workspace


jaysoo (approved):
Let's remove the `adduser` commands for now since they aren't set up correctly. We can add them later. Otherwise LGTM.

e2e/node/src/__snapshots__/node-server.test.ts.snap
Coly010, line 1: Removed as it can start to be problematic dependent on the package manager being used in CI. 
Dockerfile is also tested in the new `node-docker.test.ts` e2es

e2e/node/src/__snapshots__/node.test.ts.snap
Coly010, line 1: Removed as it can start to be problematic dependent on the package manager being used in CI. 
Dockerfile is also tested in the new `node-docker.test.ts` e2es

PR #32199, opened 2025-08-01
docs(misc): update sidebar for astro docs

jaysoo:
## Related Issue(s)

Fixes #


astro-docs/sidebar.mjs
barbados-clemens, line 1: In an ideal world we can leverage the directory structure to auto make the sidebar (at least for ally the markdown base files) 

but for now this is okay to get the PoC out the door.

barbados-clemens (approved):

PR #32220, opened 2025-08-04
fix(gradle): keep original gradle name

xiongemi:
## Current Behavior

currently, the project name is just last part to full gradle path, for example:
- gradle path for :parent-path:project-name, it currently takes only project-name
this might create duplicates in a large project

## Expected Behavior

- change the name to match gradle path exactly, then the project name would just be the gradle path
- for example, the gradle command would be `./gradlew :parent-path:project-name:build`, nx command would be `nx run :parent-path:project-name:build`

https://linear.app/nxdev/issue/NXC-2917/fix-terminal-ui-compatibility-with-gradle-naming need to be fixed before this pr

## Related Issue(s)

Fixes #

### DIFF
--- a/e2e/gradle/project.json
+++ b/e2e/gradle/project.json
@@ -3,7 +3,11 @@
   "$schema": "../../node_modules/nx/schemas/project-schema.json",
   "sourceRoot": "e2e/gradle",
   "projectType": "application",
-  "implicitDependencies": ["gradle"],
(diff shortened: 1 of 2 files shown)


FrozenPandaz (approved):

meeroslav:
This sadly doesn't seem to be working.

We need to fix this and test it again.

PR #32252, opened 2025-08-07
fix(release): do not require the npm CLI to be installed for publishing JS packages with bun

### LINKED ISSUE #32126: Publishing packages fails if using bun and the npm executable is not available in the environment
### Current Behavior

I want to build and publish a nx lib to my GitLab registry while being in my GitLab ci, using bun (v1.2.19) as package manager with the `oven/bun:1.2.19` image.

Unfortunately somewhere seems to be a `npm view` command hard-coded. The bun replacement would be `bun info`.

Since I'm using the bun container image, there's no npm executable:

```
bunx --bun nx release publish --verbose
 NX   Running target nx-release-publish for project @myScope/my-cool-lib-name:
- @myScope/my-cool-lib-name
> nx run @myScope/my-cool-lib-name:nx-release-publish
Something unexpected went wrong when checking for existing dist-tags.
 108 |      * perform the npm view step, and just show npm/pnpm publish's dry-run output.
109 |      */
110 |     if (!isDryRun && !options.firstRelease) {
111 |         const currentVersion = packageJson.version;
112 |         try {
113 |             const result = (0, child_process_1.execSync)(npmViewCommandSegments.join(' '), {
                                                     ^
error: Command failed: npm view @myScope/my-cool-lib-name versions dist-tags --json --"@myScope:registry=undefined"
/bin/sh: 1: npm: not found
 signal: null,
 status: 127,
 output: [ null,
  "", "/bin/sh: 1: npm: not found\n"
],
    pid: 395,
 stdout: "",
 stderr: "/bin/sh: 1: npm: not found\n",
      at genericNodeError (node:child_process:959:13)
      at checkExecSyncError (node:child_process:458:27)
      at execSync (node:child_process:278:31)
      at runExecutor (/builds/binsky/my-cool-repo-name/node_modules/@nx/js/src/executors/release-publish/release-publish.impl.js:113:48)
 NX   Running target nx-release-publish for project @myScope/my-cool-lib-name failed
Failed tasks:
- @myScope/my-cool-lib-name:nx-release-publish
Cleaning up project directory and file based variables 00:00
ERROR: Job failed: exit code 1
```

### Expected Behavior

I would like to have support for different package managers when using the `nx release publish` command. Or at least a way to inject a custom command to replace `npm view`.

### GitHub Repo

https://github.com/binsky08/nx-bun-test-repo

### Steps to Reproduce

1. `docker run --rm -it oven/bun:1.2.19 bash`
2. run all following commands within the container (to definitely not having npm installed)
3. `apt update && apt -y install git`
4. `git clone https://github.com/binsky08/nx-bun-test-repo`
5. `cd nx-bun-test-repo/`
6. `bun install`
7. `bunx nx release publish --verbose`

Since we do not really want to publish this test repo, we would be happy with getting something like `404 Not Found: https://registry.npmjs.org/mylib` which indicated a successful registry check.
Instead we get the attached failure logs, because npm is (correctly) not installed.

### Nx Report

```shell

Node           : 24.3.0
OS             : linux-x64
Native Target  : x86_64-linux
bun            : 1.2.19

nx             : 21.3.9
@nx/js         : 21.3.9
@nx/workspace  : 21.3.9
@nx/devkit     : 21.3.9
typescript     : 5.8.3
Cache Usage: 0.00 B / 89.63 GB
```

### Failure Logs

```shell
# bunx nx release publish --verbose

 NX   Running target nx-release-publish for project mylib:

- mylib

————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

> nx run mylib:nx-release-publish

Something unexpected went wrong when checking for existing dist-tags.
 108 |      * perform the npm view step, and just show npm/pnpm publish's dry-run output.
109 |      */
110 |     if (!isDryRun && !options.firstRelease) {
111 |         const currentVersion = packageJson.version;
112 |         try {
113 |             const result = (0, child_process_1.execSync)(npmViewCommandSegments.join(' '), {
                                                     ^
error: Command failed: npm view mylib versions dist-tags --json --"registry=undefined"
/bin/sh: 1: npm: not found

 signal: null,
 status: 127,
 output: [ null,
  "", "/bin/sh: 1: npm: not found\n"
],
    pid: 5819,
 stdout: "",
 stderr: "/bin/sh: 1: npm: not found\n",

      at genericNodeError (node:child_process:959:13)
      at checkExecSyncError (node:child_process:458:27)
      at execSync (node:child_process:278:31)
      at runExecutor (/home/bun/app/nx-bun-test-repo/node_modules/@nx/js/src/executors/release-publish/release-publish.impl.js:113:48)


————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Running target nx-release-publish for project mylib failed

Failed tasks:

- mylib:nx-release-publish
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

binsky08:
## Current Behavior
`nx release publish` requires the npm package manager to run the hard-coded `npm view ...` command even in a project that doesn't use npm as package manager

## Expected Behavior
`nx release publish` uses the detected package manager of the project (e.g. bun) to run a package manager specific command like `npm view ...` / `bun info ...`

## Related Issue(s)

Fixes #32126

## Notes
While `nx affected --target=e2e` ("No tasks were run") and `nx affected --target=test` ("Successfully ran target test for 43 projects and 2 tasks they depend on") were successful, I was not able to publish a local test version to a local registry - so unfortunately I couldn't manually verify whether this actually fixes the linked issue.

### DIFF
--- a/packages/js/src/executors/release-publish/release-publish.impl.ts
+++ b/packages/js/src/executors/release-publish/release-publish.impl.ts
@@ -127,8 +127,17 @@ Please update the local dependency on "${depName}" to be a valid semantic versio
   );
 
   const npmViewCommandSegments = [
-    `npm view ${packageName} versions dist-tags --json --"${registryConfigKey}=${registry}"`,
+    getPackageManagerCommand(pm).view,
+    packageName,
   ];
+
+  // Since only yarn does not support multiple specific fields, we'll then fetch the complete data
+  if (pm !== 'yarn') {
+    npmViewCommandSegments.push('versions dist-tags');
+  }
+  npmViewCommandSegments.push(`--json --"${registryConfigKey}=${registry}"`);
+
+  // Unlike npm, none of the other package managers appear to have a dedicated equivalent to the "npm dist-tag" command for managing tags after publication
   const npmDistTagAddCommandSegments = [
     `npm dist-tag add ${packageName}@${packageJson.version} ${tag} --"${registryConfigKey}=${registry}"`,
   ];
--- a/packages/js/src/executors/release-publish/release-publish.impl.ts
+++ b/packages/js/src/executors/release-publish/release-publish.impl.ts
@@ -1,6 +1,7 @@
 import {
   detectPackageManager,
   getPackageManagerCommand,
+  getPackageManagerVersion,
   ExecutorContext,
   readJsonFile,
 } from '@nx/devkit';
@@ -33,6 +34,19 @@ export default async function runExecutor(
   context: ExecutorContext
 ) {
   const pm = detectPackageManager();
+
+  let isNpmInstalled = false;
+  try {
+    // If npm is installed, we should get a version string, otherwise an error will be thrown.
+    isNpmInstalled =
+      getPackageManagerVersion('npm', process.cwd(), true) !== '';
+  } catch (e) {
+    // Allow missing npm only when using bun (https://github.com/nrwl/nx/pull/32252#discussion_r2262481795)
+    if (pm !== 'bun') {
+      throw e;
+    }
+  }
+
   /**
    * We need to check both the env var and the option because the executor may have been triggered
    * indirectly via dependsOn, in which case the env var will be set, but the option will not.
@@ -178,7 +192,7 @@ Please update the local dependency on "${depName}" to be a valid semantic versio
 
       if (versions.includes(currentVersion)) {
         try {
-          if (!isDryRun) {
+          if (!isDryRun && isNpmInstalled) {
             execSync(npmDistTagAddCommandSegments.join(' '), {
               env: processEnv(true),
               cwd: context.root,
@@ -188,11 +202,15 @@ Please update the local dependency on "${depName}" to be a valid semantic versio
             console.log(
               `Added the dist-tag ${tag} to v${currentVersion} for registry ${registry}.\n`
             );
-          } else {
+          } else if (isDryRun) {
             console.log(
               `Would add the dist-tag ${tag} to v${currentVersion} for registry ${registry}, but ${chalk.keyword(
                 'orange'
               )('[dry-run]')} was set.\n`
+            );
+          } else {
+            console.log(
+              `Since no npm installation can be detected, we skip the 'npm dist-tag add' command.\n`
             );
           }
           return {
--- a/packages/js/src/executors/release-publish/release-publish.impl.ts
+++ b/packages/js/src/executors/release-publish/release-publish.impl.ts
@@ -39,11 +39,20 @@ export default async function runExecutor(
   try {
     // If npm is installed, we should get a version string, otherwise an error will be thrown.
     isNpmInstalled =
-      getPackageManagerVersion('npm', process.cwd(), true) !== '';
+      execSync('npm --version', {
+        cwd: process.cwd(),
+        encoding: 'utf-8',
+        windowsHide: true,
+      }).trim() !== '';
   } catch (e) {
     // Allow missing npm only when using bun (https://github.com/nrwl/nx/pull/32252#discussion_r2262481795)
     if (pm !== 'bun') {
-      throw e;
+      console.error(
+        `npm was not found in the current environment. This is only supported when using \`bun\` as a package manager, but your detected package manager is "${pm}"`
+      );
+      return {
+        success: false,
+      };
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -127,8 +127,17 @@ Please update the local dependency on "${depName}" to be a valid semantic versio
...
+    packageName,
   ];
+
+  // Since only yarn does not support multiple specific fields, we'll then fetch the complete data
+  if (pm !== 'yarn') {
+    npmViewCommandSegments.push('versions dist-tags');
+  }
+  npmViewCommandSegments.push(`--json --"${registryConfigKey}=${registry}"`);
+
+  // Unlike npm, none of the other package managers appear to have a dedicated equivalent to the "npm dist-tag" command for managing tags after publication
```
JamesHenry, line 163: We can't just leave this right? This would still break your use-case of no `npm` being installed?

We should safely check for the existence of npm before performing this step.

My feeling is, if you are using `npm`, `yarn` or `pnpm` - you should have `npm` installed because it's bundled with node. Therefore the only acceptable case for not having `npm` is if you are running `bun` only in a particular environment.

So with that in mind, let's add a check next to where we resolve the package manager and safely check for whether or not npm is installed. If it is not and `pm !== 'bun'`, we should hard error.

Then here at the `dist-tag add` we check `isNpmInstalled` and if it's false, skip the `dist-tag` command.

Sound good?

JamesHenry (changes requested):

packages/js/src/executors/release-publish/release-publish.impl.ts
binsky08: good catch, please take a look at the recent changes [c300694](https://github.com/nrwl/nx/pull/32252/commits/c300694e11c6d9b0645c2379f511dc1cb30c07a2)

packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -188,12 +202,16 @@ Please update the local dependency on "${depName}" to be a valid semantic versio
             console.log(
               `Added the dist-tag ${tag} to v${currentVersion} for registry ${registry}.\n`
             );
-          } else {
+          } else if (isDryRun) {
```
JamesHenry: It doesn't make sense to print this dry-run message given it doesn't corresponding to an action when not in dry-run.

It's this entire dist tag related section that should be in an 

```ts
if (isNpmInstalled) {

}
```

i.e. roughly lines 188-284 from what I can tell from this view

packages/js/src/executors/release-publish/release-publish.impl.ts
```diff
@@ -33,6 +34,19 @@ export default async function runExecutor(
...
+
+  let isNpmInstalled = false;
+  try {
+    // If npm is installed, we should get a version string, otherwise an error will be thrown.
+    isNpmInstalled =
+      getPackageManagerVersion('npm', process.cwd(), true) !== '';
+  } catch (e) {
+    // Allow missing npm only when using bun (https://github.com/nrwl/nx/pull/32252#discussion_r2262481795)
+    if (pm !== 'bun') {
+      throw e;
```
JamesHenry: I'd rather have this be a controlled user-facing message and handled error state please:

```ts
console.error(`npm was not found in the current environment. This is only supported when using \`bun\` as a package manager, but your detected package manager is "${pm}"`);
return {
  success: false,
};
```

packages/nx/src/utils/package-manager.ts
```diff
@@ -242,10 +242,11 @@ export function getPackageManagerCommand(
  */
 export function getPackageManagerVersion(
   packageManager: PackageManager = detectPackageManager(),
-  cwd = process.cwd()
+  cwd = process.cwd(),
+  forceExecPackageManager = false
```
JamesHenry: Let's not add this and keep things simple, let's just run execSync(`npm --version`) in the implementation where we need it. It doesn't need a generic solution

JamesHenry (changes requested):
Thanks! Few comments

binsky08:
> Thanks! Few comments

I've added a fixup with the requested changes

JamesHenry:
Thanks @binsky08 I've triggered a PR release for this, it's a real npm release that we can test out in your environment and make sure it's fully working before merging. There will be an automated comment on here when it's ready to try, please let me know once you've verified that release in your bun only environment

JamesHenry:
Ah looks like that is not possible due to your fork not being up to date with latest nx master (it's a precaution we have in the release script). I think we can go ahead and merge when it's green, it LGTM

JamesHenry (approved):

JamesHenry (changes requested):
@binsky08 You have some legit test failures on here.

Our AI fixer has correctly identified one of them (obviously you can use a more modern value for the npm version but it doesn't really matter):

image

binsky08:
> @binsky08 You have some legit test failures on here.
> 
> Our AI fixer has correctly identified one of them (obviously you can use a more modern value for the npm version but it doesn't really matter):
> 
> image

thanks, should be fixed with [52737d3](https://github.com/nrwl/nx/pull/32252/commits/52737d313ec25431d7c1b222cc502cb5829017ff)

JamesHenry:
@binsky08 Nx Cloud AI nailed the final failure too I think 😁 can you apply it via the link above?

(This)

https://github.com/nrwl/nx/pull/32252#issuecomment-3188044312

image

binsky08:
> @binsky08 Nx Cloud AI nailed the final failure too I think 😁 can you apply it via the link above?
> 
> (This)
> 
> [#32252 (comment)](https://github.com/nrwl/nx/pull/32252#issuecomment-3188044312)
> 
> image

I'm not entirely sure about the error pattern, partly because I don't fully understand the specific test (flow). Bun says something like:
```
404 Not Found: http://localhost:4873/@proj%2f{project-name}
    + - “@proj/{project-name}@latest” does not exist in this registry
```
which seems correct to me, if the package (with the packageName from the test) does not actually exist in the registry, because the command should only be executed if `!options.firstRelease` applies, i.e. a package with that name already exists in the registry.

This condition already existed in the previous version of the file: https://github.com/nrwl/nx/blob/master/packages/js/src/executors/release-publish/release-publish.impl.ts#L144-L147

Therefore, I am wondering:
- 1. Why is there no error with pnpm?
- 2. Does the requested package really not exist in the test registry?
    - If so, the error would be correct and we may miss the firstRelease option in the test?

Unfortunately I can't reproduct the error (needed to validate my second point ^) since it (earlier) fails to run the e2e test in the [proposed devcontainer](https://github.com/nrwl/nx/tree/master/.devcontainer).

image

Seems to be an issue with connecting to the automatically started local registry. I tried it several times and have already waited >10 minutes. Does not work.
I am not deep enough in nx to fix it when I can't test it.


I'm not sure how to proceed here.

jaysoo (commented):
Thanks for the work on Bun publish support. A few items from the review history that still need to be addressed:

1. **Reviewer feedback**: @JamesHenry specifically said "Let's not add `view` to `PackageManagerCommands`" but the PR still adds it. This needs to be resolved — either refactor to avoid extending that interface, or get explicit agreement on the approach.

2. **JSON output compatibility**: The output formats of `bun info`, `yarn info`, and `npm view` differ significantly. The PR should include tests verifying that the JSON parsing works correctly across package managers.

3. **CI**: Tests are still failing and need to pass before this can merge.

JamesHenry:
Sorry for the delay on this @binsky08, I merged in #34835 to address it

PR #32257, opened 2025-08-07
docs(misc): update styling and layout for astro docs

jaysoo:
It should match the original docs in terms of color, layout, etc. more closely.


barbados-clemens (approved):
only nit is in dark mode the code block colors are a touch hard to read bc same as BG color it looks like
WKMac 2025-08-07T20-32-42

PR #32261, opened 2025-08-07
docs(nx-dev): setup content from old nx-dev into new site. 

barbados-clemens:
TODOs:
- [x] intros
- [x] features
- [x] core guides
- [x] concepts (previously moved)
- [x] technologies (in progress)
    - [ ] module-federation pages are extra special
- [x] enterprise
- [x] ~showcase~ decided not to move since very outdated
- [x] references
- [x] troubleshooting

closes: DOC-93
contributes to: DOC-94


astro-docs/astro.config.mjs
```diff
@@ -1,13 +1,12 @@
...
 import { defineConfig } from 'astro/config';
 import starlight from '@astrojs/starlight';
-import { sectionSidebarPlugin } from './src/plugins/section-sidebar.plugin';
 import netlify from '@astrojs/netlify';
 import linkValidator from 'starlight-links-validator';
 import react from '@astrojs/react';
 import markdoc from '@astrojs/markdoc';
 import tailwindcss from '@tailwindcss/vite';
-import { sidebar } from './sidebar.mjs';
+import { sidebar } from './sidebar';
```
jaysoo: Might be good to keep `.mts` in the filename, and leave `.mjs` as that is technically the correct way to import moving forward.

nx-dev/ui-common/src/lib/youtube.component.tsx
```diff
@@ -124,7 +124,7 @@ export function YouTube(props: YouTubeProps): JSX.Element {
               {
                 'rounded-lg shadow-lg': !props.disableRoundedCorners,
               },
-              'border-2 border-slate-200 hover:border-slate-500 dark:border-slate-700/40 dark:hover:border-slate-700'
+              'aspect-video border-2 border-slate-200 hover:border-slate-500 dark:border-slate-700/40 dark:hover:border-slate-700'
```
jaysoo, line 127: Does this have  any effect on the existing Next.js docs rendering?

jaysoo (approved):

nx-dev/ui-common/src/lib/youtube.component.tsx
barbados-clemens: no this was just fixing the styles I noticed where the youtube video didn't have the right ratio. it seems that the previous impl was done via a global css instead of tailwinds which is why it didn't have the right aspect ratio initially when porting.

astro-docs/astro.config.mjs
barbados-clemens: updated to use mts file

PR #32266, opened 2025-08-08
feat(docker): add --dockerVersion flag for specifying exact version

Coly010:
## Current Behavior
There is no method for specifying an exact version for docker version during `nx release`

## Expected Behavior
Add `--dockerVersion` to provide the user with a manner of providing an exact docker version


packages/nx/src/command-line/release/command-object.ts
```diff
@@ -504,9 +505,15 @@ function withFirstReleaseOptions<T>(
 function withDockerVersionSchemeOptions<T>(
   yargs: Argv<T>
 ): Argv<T & DockerVersionSchemeArgs> {
-  return yargs.option('dockerVersionScheme', {
-    type: 'string',
-    describe:
-      'Exact docker version scheme to apply to the selected release group. Warning: Docker support is experimental. Breaking changes may occur and not adhere to semver versioning.',
-  });
+  return yargs
```
JamesHenry, line 508: Let's error in yargs validation if you provide both, it would be nonsensical

JamesHenry (changes requested):

packages/nx/src/command-line/release/command-object.ts
Coly010: @claude Can you apply validation to error when both `--dockerVersionScheme` and `--dockerVersion` are supplied

JamesHenry (approved):

PR #32271, opened 2025-08-08
fix(core): adding `customProxyConfigPath` support to default runner

ZackDeRose:
## Current Behavior

Users migrating to Nx 21 that were using custom task runners had support for a `customProxyConfigPath` option that is not supported at the top level of nx.json config (however most others now are)

## Expected Behavior

Users should be able to list a `customProxyConfigPath` that is now applied as expected to the default runner.

## Related Issue(s)

Fixes #


ZackDeRose:
Closing in favor of continuing to support this through `taskRunnerOptions`.

See: https://github.com/nrwl/nx-cloud-helm/blob/main/PROXY-GUIDE.md#nx-cloud-proxy-issues

PR #32282, opened 2025-08-10
fix(core): make watch command work with all and initialRun specified

### LINKED ISSUE #32281: 'nx watch --all --initialRun' does not work
### Current Behavior

If you specify both the `all` and `initialRun` options when running `nx watch`, `initialRun` will have no effect.

### Expected Behavior

The command should be called once at the beginning even if there are no file changes.

### GitHub Repo

Use any repo

### Steps to Reproduce

1. Run `nx watch --initialRun --all --verbose -- echo executed`

### Nx Report

```shell
Node           : 22.15.0
OS             : linux-x64
Native Target  : x86_64-linux
pnpm           : 10.10.0

nx          : 21.3.11
typescript  : 5.9.2
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

omasakun:
## Current Behavior

If you specify both the `all` and `initialRun` options when running `nx watch`, `initialRun` have no effect.

## Expected Behavior

The command should be called once at the beginning even if there are no file changes.

## Related Issue(s)

Fixes #32281

### DIFF
--- a/packages/nx/src/command-line/watch/watch.ts
+++ b/packages/nx/src/command-line/watch/watch.ts
@@ -48,11 +48,11 @@ export class BatchFunctionRunner {
       this.pendingFiles.add(fileChange.path);
     });
 
-    return this.process();
+    return this.process(true);
   }
 
-  private async process() {
-    if (!this.running && this.hasPending) {
+  private async process(runAnyway: boolean) {
+    if (!this.running && (this.hasPending || runAnyway)) {
       this.running = true;
 
       // Clone the pending projects and files before clearing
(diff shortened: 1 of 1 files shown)


omasakun:
Hi, @FrozenPandaz
Can you review this PR?
Unless this bug is resolved we cannot use nx without a patch on our project.

AgentEnder (approved):

jaysoo (approved):
Fix is correct — `enqueue([], [])` yields `hasPending=false` so the callback never fires for `--initialRun --all`.

One minor observation: `runAnyway=true` is passed on all `enqueue()` calls, not just the initial one. If the daemon ever sends an event with empty changed files/projects, the callback would fire unnecessarily. A dedicated `runInitial()` method would be cleaner, but this is minor and the practical impact is negligible since the daemon doesn't send empty events today.

Already approved by @AgentEnder. LGTM.

PR #32300, opened 2025-08-11
fix(release): add missing git config options to nx-schema

A-EbrahimRSA:
## Current Behavior
The current tooltip options displayed for the `nx.json` config file don't correctly reflect some underlying options that are available

## Expected Behavior
All config options that are available should be exposed on the schema file


A-EbrahimRSA:
On a slightly separate note, when I was reviewing the Configuration interfaces in `packages/nx/src/config/nx-json.ts` I found it quite difficult to compare the interface options against what was inside `nx-schema.json`. The primary reason being the attributes in the interface definition and the json are not ordered in any particular way. I wonder if having some alphabetic ordering would help aid in situations like these and keep things aligned better

A-EbrahimRSA:
Additionally, I'm struggling to get the repo cloned and setup locally, so if someone could format the file for me I'd appreciate it

packages/nx/schemas/nx-schema.json
```diff
@@ -823,6 +831,10 @@
             "enum": ["disk"],
             "description": "The fallback version resolver to use when the configured currentVersionResolver fails to resolve the current version."
           },
+          "firstRelease":{
```
JamesHenry: It doesn’t make sense to set this in config on an ongoing basis and we don’t want to encourage its usage

JamesHenry:
@A-EbrahimRSA please explain more about why you can’t clone the repo, on your last PR I linked you to the article that showed you how to do a shallow checkout, did that not work? In what way?

A-EbrahimRSA:
> @A-EbrahimRSA please explain more about why you can’t clone the repo, on your last PR I linked you to the article that showed you how to do a shallow checkout, did that not work? In what way?

When running the `pnpm install`/`pnpm install --frozen-lockfile` I get a OOM error:

```
<--- Last few GCs --->

[17608:000001DB3BFEF000]   229829 ms: Scavenge (interleaved) 399.0 (528.6) -> 399.0 (528.6) MB, pooled: 0 MB, 222.87 / 0.00 ms  (average mu = 0.988, current mu = 0.731) allocation failure;
[17608:000001DB3BFEF000]   230310 ms: Scavenge (interleaved) 676.6 (806.2) -> 676.6 (806.2) MB, pooled: 0 MB, 129.08 / 0.00 ms  (average mu = 0.988, current mu = 0.731) allocation failure;

FATAL ERROR: invalid array length Allocation failed - JavaScript heap out of memory
----- Native stack trace -----
Progress: resolved 4248, reused 303, downloaded 1875, added 1463
 1: 00007FF69525CCD9 node::OnFatalError+1193
 2: 00007FF695E91617 v8::Function::NewInstance+423
 3: 00007FF695C93927 v8::base::AddressSpaceReservation::AddressSpaceReservation+320967
 4: 00007FF6958894FE v8::String::Utf8Value::~Utf8Value+135838
 5: 000001DC1E03733A
```

It seems the issue was linked to a regression in Node v24.4.0. I've just upgraded to Node v24.5.0 and I seem to be able to move past that issue. https://github.com/nodejs/node/issues/59057

However, I am now getting a different error at the end of installation. The bottom one seems to indicate Rust is needed? I'm not sure if this is true or a false positive.
```
packages/nx postinstall$ node ./bin/post-install || exit 0
[6 lines collapsed]                                                                                                                                                             
│     at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1056:22)
│     at Module._load (node:internal/modules/cjs/loader:1219:37)
│     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
│     at wrapModuleLoad (node:internal/modules/cjs/loader:238:24)
│     at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
│     at node:internal/main/run_main_module:33:47 {
│   code: 'MODULE_NOT_FOUND',
│   requireStack: []
│ }
│ Node.js v24.5.0
└─ Done in 159ms
. preinstall$ node ./scripts/preinstall.js
│ 'rustc' is not recognized as an internal or external command,
│ operable program or batch file.
│ Could not find the Rust compiler on this system. Please make sure that it is installed with https://rustup.rs
└─ Failed in 657ms at \nx
 ELIFECYCLE  Command failed with exit code 1.
```

Regardless of these errors, I've been able to actually run the format command. I've updated the PR so hopefully the formatting issues are resovled now

packages/nx/schemas/nx-schema.json
A-EbrahimRSA: Ah I see your meaning, I personally prefer that all config should be discoverable through tooltips, even if the intention is as a once off. Nevertheless, I've removed the above item

JamesHenry (approved):

JamesHenry:
@A-EbrahimRSA yes many parts of nx core are written in rust, it is therefore required for development, thanks for updating the PR

PR #32323, opened 2025-08-12
fix(js): recognize tsconfig module `nodenext` as ESM

### LINKED ISSUE #31295: Warning shown about cjs format when NodeNext module used
### Current Behavior

When building with `@nx/js:tsc` I get the following warning:

> Package type is set to "module" but "cjs" format is included. Going to use "esm" format instead. You can change the package type to "commonjs" or remove type in the package.json file.

Despite me not setting `format: "cjs"` anywhere.

My module format is set to `NodeNext` in my `tsconfig.json`

### Expected Behavior

When building an ESM package with module set to `NodeNext` this warning should not be shown

### GitHub Repo

https://github.com/morganstanley/fdc3-web

### Steps to Reproduce

Clone above repo

```bash
nx build fdc3-web
```

### Nx Report

```shell
Node           : 22.13.1
OS             : win32-x64
npm            : 10.9.2

nx (global)    : 21.0.3
nx             : 21.0.3
@nx/js         : 21.0.3
@nx/jest       : 21.0.3
@nx/eslint     : 21.0.3
@nx/workspace  : 21.0.3
@nx/devkit     : 21.0.3
@nx/esbuild    : 21.0.3
@nx/node       : 21.0.3
@nx/vite       : 21.0.3
@nx/web        : 21.0.3
typescript     : 5.8.3
Registered Plugins:
@nx/eslint/plugin
@nx/vite/plugin
@nx/vite/plugin
Cache Usage: 726.05 KB / 47.69 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

the format that is used is not set in the options of the `@nx/js:tsc` executor (this caused quite a bit of confusion as when I tried to set it it had no effect). It is instead generated from the tsconfig file that is used. The format comes from the function `determineModuleFormatFromTsConfig` which only considers:

```
ts.ModuleKind.ES2015
ts.ModuleKind.ES2020
ts.ModuleKind.ES2022
ts.ModuleKind.ESNext
```

to be esm. `NodeNext` should be added

https://github.com/nrwl/nx/blob/43a20e2ecc104d072eb1071e24783890b6ecbcfe/packages/js/src/executors/tsc/tsc.impl.ts#L28

JoshuaCWebDeveloper:
## Current Behavior
A tsconfig `module` value of `nodenext` is treated as a CJS format, which produces a warning when built due to it being used in ESM projects.

## Expected Behavior
A tsconfig `module` value of `nodenext` is now teated as an ESM format, producing no warning when built.

## Related Issue(s)
Fixes https://github.com/nrwl/nx/issues/31295


JoshuaCWebDeveloper:
Failed check is due to failing test in `angular:test` on `master`, unclear whether or not the fix is to just update the snapshot.

Coly010:
Hey @JoshuaCWebDeveloper I'm not sure if this is 100% correct. 

`NodeNext` will be ESM _if_ the nearest package.json type is module, and/or depending on the file extension of files.

JoshuaCWebDeveloper:
> Hey @JoshuaCWebDeveloper I'm not sure if this is 100% correct.
> 
> `NodeNext` will be ESM _if_ the nearest package.json type is module, and/or depending on the file extension of files.

@Coly010 Good point. Looked at this a bit closer, `format` is an optional option to `updatePackageJson()` because it is already checking the `package.json` type. `getUpdatedPackageJsonContent()` also has logic for handling no given format.

My new solution passes a format of undefined to `updatePackageJson()` for the `node*` module types. This allows `updatePackageJson()` to determine the correct format (currently checks `package.json`, and can check file extensions in the future if it needs to). I also check every single tsconfig module option for completeness and to make it super clear what the method returns now that I'm introducing a third return type.

packages/js/src/executors/tsc/tsc.impl.ts
```diff
@@ -27,17 +27,27 @@ import { createEntryPoints } from '../../utils/package-json/create-entry-points'
...
+      return 'esm';
+    case ts.ModuleKind.CommonJS:
+    case ts.ModuleKind.AMD:
+    case ts.ModuleKind.UMD:
+    case ts.ModuleKind.System:
+    case ts.ModuleKind.None:
+    case ts.ModuleKind.Preserve:
+        return 'cjs';
+    case ts.ModuleKind.Node16:
+    case ts.ModuleKind.Node18:
```
leosvelperez, line 46: Please add `Node20`.

leosvelperez (changes requested):
Thanks for contributing! Apologies for the delay.

Please rebase the changes and update the call sites (`tsc.impl.ts` and `tsc.batch-impl.ts`) to handle the `undefined` result and avoid providing `[undefined]` to the `updatePackageJson` function. Usages should be something like:

```ts
const format = determineModuleFormatFromTsConfig(options.tsConfig);
updatePackageJson(
  {
    ...options,
    additionalEntryPoints: createEntryPoints(
      options.additionalEntryPoints,
      context.root
    ),
    format: format ? [format] : undefined,
  },
  context,
  target,
  dependencies
);
```

Coly010:
Should be handled now by: https://github.com/nrwl/nx/pull/31508

PR #32342, opened 2025-08-13
feat(gradle): enable migrations to occur with workspaces using version catalogs

lourw:
## Current Behavior

Our project graph gradle migrations currently only bump up the gradle plugin if its versions are set in `build.gradle.kts` files. This means that migrations will not work with workspaces that use version catalogs and aliases. 

## Expected Behavior

Given a workspace using the project graph gradle plugin on a legacy version, migrations should bump up the version of the Nx gradle plugin if its specified in `libs.versions.toml` instead of `build.gradle.kts`

## Related Issue(s)

Fixes #NXC-2952


packages/gradle/src/utils/version-catalog-utils.ts
```diff
@@ -0,0 +1,189 @@
+import { Tree, globAsync } from '@nx/devkit';
+import * as TOML from 'smol-toml';
+import { gradleProjectGraphPluginName } from './versions';
+
+export async function findVersionCatalogFiles(tree: Tree): Promise<string[]> {
+  const versionCatalogPaths: string[] = [];
+
+  tree.listChanges().forEach((change) => {
```
FrozenPandaz: Do you know why we need to look through the changes in the tree when we're already doing the glob?

The glob below should also pick up files which were written to the tree but not on disk yet.

cc @AgentEnder

packages/gradle/src/utils/version-catalog-utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  return false;
+}
+
+export function extractPluginVersionFromCatalog(
+  catalogContent: string,
+  pluginName: string
+): string | null {
+  const toml = TOML.parse(catalogContent);
```
FrozenPandaz: Each of these util functions parses the catalog content...

Can the utils take a parsed toml file? so it can be parsed once?

FrozenPandaz (changes requested):

FrozenPandaz (commented):
Could you also update the migration for `0.1.5` to point to Nx `21.4.0-beta.12` where this change will be released please?

packages/gradle/src/utils/version-catalog-utils.ts
lourw: Woop. I think you-know-what dropped that there. Yeah the glob itself is enough.

FrozenPandaz (approved):

PR #32344, opened 2025-08-13
chore(misc): remove circular dep from @nx/cypress to itself and clarify why other remaining packages are ignored

jaysoo:
This removes the need to ignore `@nx/cypress` in `packages/cypress/.eslintrc.json`.  This was never really a ciruclar dep, as `packages/cypress/src/generators/configuration/files/v9/__directory__/plugins/index.js` is a template for `generateFiles`, thus appending `__tmpl__` fixes it.

Other packages have valid reasons so a comment is added to clarify for the future.


leosvelperez (approved):

packages/cypress/src/generators/configuration/files/v9/__directory__/plugins/index.js__tmpl__
Coly010, line 1: We definitely substitute `__tmpl__` right?

Coly010 (approved):
One question just to double check something

packages/cypress/src/generators/configuration/files/v9/__directory__/plugins/index.js__tmpl__
jaysoo: yep! I checked :)

PR #32351, opened 2025-08-14
feat(angular): support angular v20.2

leosvelperez:
## Current Behavior

Angular v20.2 is not supported.

## Expected Behavior

Angular v20.2 is supported.


package.json
```diff
@@ -316,14 +316,14 @@
...
-    "typescript-eslint": "^8.29.0",
+    "typescript": "~5.9.2",
+    "typescript-eslint": "^8.40.0",
     "unist-builder": "^4.0.0",
     "use-sync-external-store": "^1.2.0",
     "verdaccio": "6.0.5",
     "vite": "6.2.0",
     "vitest": "3.0.5",
-    "webpack": "5.99.9",
+    "webpack": "5.101.1",
```
Coly010: This will likely conflict with https://github.com/nrwl/nx/pull/32387/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R326

Also, Angular has updated their version to `5.101.3`: https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/package.json#L58

Coly010 (approved):

packages/nx/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
+{"fileNames":[],"fileInfos":[],"root":[],"options":{"allowJs":true,"composite":true,"declaration":true,"declarationMap":true,"emitDecoratorMetadata":true,"experimentalDecorators":true,"importHelpers":true,"module":1,"skipLibCheck":true,"target":8},"version":"5.8.3"}
```
FrozenPandaz: This should not get committed

packages/angular/src/utils/backward-compatible-versions.ts
```diff
@@ -13,8 +13,14 @@ export type PackageVersionNames =
   | CompatPackageVersionNames;
 
 export type VersionMap = {
-  angularV18: Record<CompatPackageVersionNames, string>;
-  angularV19: Record<CompatPackageVersionNames, string>;
+  angularV18: Record<
+    Exclude<CompatPackageVersionNames, 'angularRspackVersion'>,
```
FrozenPandaz: This doesn't have to be excluded anymore.

FrozenPandaz (approved):

PR #32388, opened 2025-08-18
fix(release): publish only projects that have a new version

mpsanchis:
## Current Behavior

IF
there is a change in the repository, committed in a way that results in a new version for any project in the graph
THEN
`releasePublish` is called for all projects in the graph

For example:
IF
- a repo has only two independent projects A, B
- both A and B have already been released and published, and the repo is tagged with *A@0.0.0* and *B@0.0.0*
- there is a new commit that modifies B: *feat(B): New feature only for B*

THEN
- calling `nx release --yes` will end up calling `nx-release-publish` for both A and B 
- because *A* has not changed, it will attempt to (re)publish it with version 0.0.0

## Expected Behavior

Only the projects that have a new version should be published by default. This is more similar to the "affected" behavior.
Other behaviors are always possible by using the programmatic API.

For example:
In the example from above, only *B* would be published.

### DIFF
--- a/packages/nx/src/command-line/release/release.ts
+++ b/packages/nx/src/command-line/release/release.ts
@@ -403,8 +403,15 @@ export function createAPI(overrideReleaseConfig: NxReleaseConfiguration) {
     }
 
     if (shouldPublish) {
+      const projectsToPublish = Object.entries(
+        versionResult.projectsVersionData
+      )
+        .filter(([, versionDataEntry]) => versionDataEntry.newVersion !== null)
+        .map(([name]) => name);
+
       const publishResults = await releasePublish({
         ...args,
+        projects: projectsToPublish,
         versionData: versionResult.projectsVersionData,
       });
       const allExitOk = Object.values(publishResults).every(


JamesHenry (changes requested):
This would be a breaking change and prevent people from rerunning a failed release when e.g. NPM was down
Right now you can happily do that knowing it will skip your project during the publishing phase if it's been published already (thanks to an npm view check), or publish if not
So at the very least this behaviour would have to be opt in and not the default

mpsanchis:
Hi @JamesHenry ,

First of all, thanks for taking a look 🙏. 

I will have to roll this back, and maybe in the end I don’t do a PR at all. I’m evaluating my options.

However, here’s some comments about your reply:

1. In a repo where manifests don’t get updated (say, they say “0.0.0”, or “workspace”, “local” or whatever), `releaseVersion` will say something like `🚫 No changes were detected using git history and the conventional commits standard`. This will result in the manifest not being modified. Then, `nx-release-publish` will attempt to publish an artifact whose manifest is `0.0.0`, `workspace` or similar. That’s not what I would expect.

2. Not all technologies behave like NPM. A colleague has a repo which pushes OCI images to a Nexus registry, and if you attempt to re-publish an image that already exists, it will return an HTTP 4xx and fail with `blob upload invalid`. 

My PR has some issues and it’s not adequate to all cases, but I wanted to highlight that we shouldn’t think about NPM or JS when dealing with core Nx code such as this one.

I’ll re-think what we need. If I find it to be generic, I will attempt a next PR 😄  Otherwise, I’ll just code a release script for us using the programmatic API.

Cheers!

JamesHenry:
@mpsanchis It's absolutely true that not all technologies behave like npm, but that doesn't mean we can break the npm use-case in minor releases, not least because it's still by far the majority usage of `nx release`. There's always going to be a gray area in terms of bug fixes and breaking changes, but I think in this case it's certainly the latter and so again I am definitely open to it but it needs to be opt in for 21.x

Point (1) _does_ sound like an unquestionable bug but I'm curious how that situation is reproducible, please could you provide some steps? Are you explicitly releasing a project that doesn't have any applicable changes?

mpsanchis:
@JamesHenry 

Sure. Unfortunately it’s a private repo, but I give you the main idea:

- We had a repo working with `@jscutlery/semver`, whose developers were always releasing something (pre-releases in PRs, and releases in *main*). 
- Versioning was one more task of the graph, which was triggered at every PR with `—affected`, so only affected projects would be versioned/built/tested/…/published
- Now, we migrated to `nx release`, wrongly assuming that `nx release --yes` (for main) would only attempt the publish step on projects with new version. However, it calls publish for all projects

The `nx-release-publish` executors/commands are very dummy in this repo, as they simply take the version from the manifest and attempt to publish. Our registry complains when re-attempting to publish `0.0.0` (which is our “local” keyword).

If you need more details don’t hesitate to ask- here or Discord

Cheers

A-EbrahimRSA:
I'm not sure if this is also related but what is the expected behaviour of running the `nx release` command within a monorepo that uses independent versioning and something like conventional Commits versioning? 
My understanding with the above context would be that running the `release` command would only running the publish step for packages which need it. However I see that all packages in my monorepo get updated. This sounds quite similar to the issue above although I am unsure if that is true

mpsanchis:
@A-EbrahimRSA the issue is:

> which need it

For me, it means the ones that have a new version. But the out-of-the box implementation calls version+changelog+publish for all projects. The only way to custom this right now is to use the programmatic API: https://nx.dev/features/manage-releases#using-the-programmatic-api-for-nx-release

A-EbrahimRSA:
Yeah IMO, I think that shouldn't be the expected behaviour. Especially in situations where the new version is being derived by things such as git tags/conventional commits etc.

mpsanchis:
Closing because this won't be addressed any time soon. Programmatic API is the way

PR #32401, opened 2025-08-19
fix(graph): some misc fixes to the graph

nartc:
- Expanded inputs with placeholder to prevent content shifting
- Bump `@nx/graph` to fix task graph missing transitive deps issue
- Disabling cache will throw runtime errors for expanded inputs operation so putting in a couple of safe fallbacks when things are undefined

### DIFF
--- a/graph/shared/src/lib/project-graph-data-service/fetch-project-graph-service.ts
+++ b/graph/shared/src/lib/project-graph-data-service/fetch-project-graph-service.ts
@@ -82,17 +82,17 @@ export class FetchProjectGraphService implements ProjectGraphService {
     if (!this.taskInputsUrl) {
       return {};
     }
-    
+
     const request = new Request(`${this.taskInputsUrl}?taskId=${taskId}`, {
       mode: 'no-cors',
     });
 
     const response = await fetch(request);
     const result = (await response.json())[taskId];
-    
+
     // Cache the result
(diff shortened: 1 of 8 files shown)


graph/shared/src/lib/project-graph-data-service/fetch-project-graph-service.ts
```diff
@@ -73,14 +74,25 @@ export class FetchProjectGraphService implements ProjectGraphService {
   async getExpandedTaskInputs(
     taskId: string
   ): Promise<Record<string, string[]>> {
+    // Check cache first
+    if (this.expandedTaskInputsCache.has(taskId)) {
```
MaxKless: is there some way for clearing the cache? If the project graph changes

MaxKless (approved):

graph/shared/src/lib/project-graph-data-service/fetch-project-graph-service.ts
nartc: Remove the caching on this level and keep the one in `graph.ts` command directly. There's not a good place to invalidate this cache on the `graph-client` side.
- we don't currently have any polling for new task graph and because we're lazy loading task graph or project/target on demand, it's tricky to find a place to poll
- we definitely don't want to poll for expanded inputs

PR #32416, opened 2025-08-19
fix(gradle): allow test target name to be configured from nx.json

lourw:
## Current Behavior

We had hardcoded test target names which meant that users who specify custom test suites would not get their atomized tests correctly picked up. 

## Expected Behavior

Allow users to configure the `testTargetName` field in their `nx.json` and correctly identify which tasks to atomize.

## Related Issue(s)

Fixes #


packages/gradle/project-graph/src/main/kotlin/dev/nx/gradle/utils/ProjectUtils.kt
```diff
@@ -107,13 +107,16 @@ fun processTargetsForProject(
...
+      // Add task to its Gradle group (e.g., "build", "verification") if it has one
+      val gradleGroup = task.group
+      if (!gradleGroup.isNullOrBlank()) {
+        val tasksInGroup = targetGroups.getOrPut(gradleGroup) { mutableListOf() }
+        tasksInGroup.add(task.name)
+      }
 
+      // Set the target name that will be used with Nx if the override is set, else use the task
+      // name
+      val taskName = targetNameOverrides.getOrDefault("${task.name}TargetName", task.name)
```
MaxKless: this looks like it's not used later on?

MaxKless (approved):

PR #32418, opened 2025-08-19
feat(graph): task graph support multiple targets

nartc:
- [x] graph-client supports multiple targets with multiselect and query params `?targets`
- [x] `nx graph` supports multiple targets (i.e: `nx run-many -t build,test --graph`

image


graph/client/src/app/task-graph-cache.ts
```diff
@@ -57,90 +56,55 @@ class TaskGraphCache {
...
 
     // Update fetched projects tracking
     if (isAllProjects) {
-      // Mark that we have all projects for this target
+      // Mark that we have all projects for this combination
       merged.hasAllProjects = true;
       // Add all projects from the response to fetchedProjects
-      Object.keys(response.taskGraphs).forEach((taskId) => {
+      Object.keys(response.taskGraph.tasks).forEach((taskId) => {
         const [project] = taskId.split(':');
```
FrozenPandaz: This is not always valid.. projects can have `:` in their name.

packages/nx/src/command-line/graph/graph.ts
```diff
@@ -261,10 +261,13 @@ export async function generateGraph(
     args.file &&
     !(args.file === 'stdout' || args.file.endsWith('.json'))
   ) {
-    output.warn({
-      title: 'Showing Multiple Targets is not supported yet',
+    output.note({
```
FrozenPandaz: Let's remove this note.. we don't need to tell them that it's supported.. since they're already running it.

packages/nx/src/command-line/graph/graph.ts
```diff
@@ -962,44 +967,81 @@ async function createTaskGraphClientResponse(
...
-  );
-  performance.mark('task hash plan generation:start');
-  const plans: Record<string, string[]> = {};
-  for (const individualTaskGraph of Object.values(taskGraphs.taskGraphs)) {
-    for (const task of Object.values(individualTaskGraph.tasks)) {
-      if (plans[task.id]) {
-        continue;
-      }
+  // Get all projects and all targets for legacy full workspace support
+  const projects = Object.keys(graph.nodes);
```
FrozenPandaz, line 954: Can you investigate if we can remove this?

packages/nx/src/command-line/run-many/run-many.ts
```diff
@@ -58,7 +58,7 @@ export async function runMany(
         watch: true,
         open: true,
         view: 'tasks',
-        all: nxArgs.all,
+        all: nxArgs.all && !nxArgs.projects,
```
FrozenPandaz, line 61: Add this to affected as well

FrozenPandaz (approved):

PR #32427, opened 2025-08-20
docs(nx-dev): fix broken markdoc components in documentation

jaysoo:
This PR re-enables the escaped markdoc tags. Note that graph and PDV tags that pointed to an external JSON file is changed to inline the JSON instead so we don't need that extra file loading logic.

## Current Behavior
Multiple pages in the Astro documentation site display broken components due to:
- Escaped template blocks (\{% ... %\}) that prevent proper rendering
- Component name mismatches using hyphens instead of underscores
- Missing side-by-side component causing "Undefined tag" errors
- Graph components showing "Could not parse JSON" due to markdown code blocks
- External JSON file references not being inlined properly

## Expected Behavior
All documentation pages render correctly with:
- Proper template block syntax without backslash escaping
- Component names matching markdoc.config.mjs registrations
- All required components available and properly configured
- Graph and project_details components displaying JSON data inline
- All Markdoc components functioning as intended

Screenshot 2025-08-20 at 10 55 53 AM

Screenshot 2025-08-20 at 12 33 19 PM

Also add back the `side_by_side` tag for now since it's used in a couple of places:

Screenshot 2025-08-20 at 12 33 19 PM
Screenshot 2025-08-20 at 12 55 23 PM

## Related Issue(s)
Fixes DOC-148

### DIFF
--- a/astro-docs/markdoc.config.mjs
+++ b/astro-docs/markdoc.config.mjs
@@ -378,5 +378,14 @@ export default defineMarkdocConfig({
         },
       },
     },
+    side_by_side: {
+      render: component('./src/components/markdoc/SideBySide.astro'),
+      attributes: {
+        align: {
+          type: 'String',
+          default: 'center',
+        },
+      },
+    },
   },
 });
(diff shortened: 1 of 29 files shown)


astro-docs/markdoc.config.mjs
```diff
@@ -378,5 +378,14 @@ export default defineMarkdocConfig({
         },
       },
     },
+    side_by_side: {
```
barbados-clemens, line 381: IMO we probs should just get rid of this one, but can do in future PRs if we want to remove it.

barbados-clemens (approved):

PR #32432, opened 2025-08-20
docs(nx-cloud): add new nx cloud intro page

barbados-clemens:


docs/map.json
```diff
@@ -2511,6 +2511,19 @@
       "id": "ci",
       "description": "Learn how to enable Distributed Tasks Executions, remote caching and more.",
       "itemList": [
+        {
```
jaysoo, line 2514: Weird that I cannot see the sidebar link even though it's in the `menu.json`. We must be filtering it out in the sidebar. (maybe because it's not top level?) We could make it `Getting Started -> Intro` like before.

Screenshot 2025-08-21 at 8 23 35 AM

barbados-clemens: forgot there was a sidebar allowlist. updated 👍

jaysoo (approved):

PR #32435, opened 2025-08-20
fix(core): exclude non-cacheable tasks from flaky detection

FrozenPandaz:
## Current Behavior

Currently, the flaky task detection considers ALL tasks that have mixed success/failure results as "flaky", including non-cacheable tasks like development servers (`nx serve`) which are expected to have inconsistent behavior.

## Expected Behavior

With this PR, only explicitly cacheable tasks (`cache: true`) are considered for flaky detection. Non-cacheable tasks with mixed results will no longer trigger flaky warnings.

## Related Issue(s)

This addresses the issue where non-cacheable tasks like development servers were incorrectly being flagged as flaky, creating noise in the flaky task detection system.

### Changes Made

1. **Database Schema Updates**
   - Added `cache` column to `task_details` table 
   - Updated `HashedTask` struct to include `cache: Option<bool>` field

2. **Query Logic Enhancement**
   - Modified flaky task detection query to JOIN with `task_details` table
   - Added filter condition `AND task_details.cache = 1` to exclude non-cacheable tasks
   - Added clear comments explaining the query logic

3. **Data Flow Updates**
   - Updated TypeScript code in `hash-task.ts` to pass `cache` information from Task objects to native layer
   - Both batch and individual task detail recording now include cache property

4. **Comprehensive Testing**
   - Updated existing tests to include cache information
   - Added new test "should not consider non-cacheable tasks as flaky" that verifies:
     - Tasks with `cache: false` are NOT marked as flaky
     - Tasks with `cache: undefined` are NOT marked as flaky  
     - Only explicitly cacheable tasks with mixed results are marked as flaky

### Test Results
- **Rust tests**: 202 passed ✅
- **TypeScript tests**: 4 passed ✅ (all NxTaskHistory tests)

## Benefits

- **Improved Accuracy**: Flaky detection now focuses only on tasks where consistency is expected
- **Reduced Noise**: Development servers and other non-cacheable tasks no longer create false positives
- **Better Developer Experience**: More relevant and actionable flaky task warnings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,8 +103,8 @@ node_modules/
 .cursorindexingignore
 # OS specific
 # Task files
-tasks.json
-tasks/
+/tasks.json
+/tasks
 
 # Upstream docs local configuration (machine-specific)
 .upstreamdocs.local.json
--- a/packages/nx/src/hasher/hash-task.ts
+++ b/packages/nx/src/hasher/hash-task.ts
@@ -64,6 +64,7 @@ export async function hashTasksThatDoNotDependOnOutputsOfOtherTasks(
         project: task.target.project,
         target: task.target.target,
         configuration: task.target.configuration,
+        cache: task.cache,
       }))
     );
   }
@@ -111,6 +112,7 @@ export async function hashTask(
         project: task.target.project,
         target: task.target.target,
         configuration: task.target.configuration,
+        cache: task.cache,
       },
     ]);
   }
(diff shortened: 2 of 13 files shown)


AgentEnder (approved):

packages/react/plugins/component-testing/index.ts
```diff
@@ -142,7 +142,7 @@ export function nxComponentTestingPreset(
           target: ctTargetName,
           configuration: ctConfigurationName,
         },
-        ctExecutorContext
+        ctExecutorContext as any
```
Coly010: Why the as any casting in this file?

Coly010 (approved):

PR #32452, opened 2025-08-21
fix(core): repair nx mcp in pnpm/yarn

MaxKless:
This pull request updates the `mcp.ts` command handler to improve compatibility with multiple package managers and enhance argument handling for command execution. The changes ensure that the correct flags are used for each package manager and that commands with spaces are executed properly.

* Adjusted arguments passed to the `dlx` command for `npm`, `yarn`, and `bun` to use their respective flags (`-y`, `--quiet`, `--silent`), ensuring consistent behavior across different package managers.
* Improved handling of `dlx` commands that may contain spaces (e.g., `pnpm dlx`) by splitting the command and passing arguments appropriately to `spawnSync`.
* Updated both `mcpHandler` and `showHelp` functions to use the new argument handling logic for executing commands and displaying help output.

### DIFF
--- a/packages/nx/src/command-line/mcp/mcp.ts
+++ b/packages/nx/src/command-line/mcp/mcp.ts
@@ -1,33 +1,59 @@
 import { spawnSync } from 'child_process';
-import { getPackageManagerCommand } from '../../utils/package-manager';
+import {
+  detectPackageManager,
+  getPackageManagerCommand,
+} from '../../utils/package-manager';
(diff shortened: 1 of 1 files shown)


FrozenPandaz (changes requested):
We should have helpers for this in `package-manager.ts`

AgentEnder (approved):

FrozenPandaz (approved):

PR #32484, opened 2025-08-22
docs(misc): fix issues with base href, rewrites, and connecting next.js and astro docs

jaysoo:
This PR fixes issues with connecting new Astro docs with the `/docs` rewrite rule on Next.js side.

https://www.loom.com/share/150deedd25a24259b0aa8cb141cec1d3

The redirects now go through Netlify since it's a lot more flexible, and we need to do that later when porting Next.js to Netlify, etc.

## Not done / follow-up:
Other pages and UI components that use `<Link>` to a doc page need to check against `NEXT_PUBLIC_ASTRO_URL` and make sure to use the updated `/docs` path for that. I only handled the homepage for now just to show it working.


astro-docs/netlify.toml
barbados-clemens, line 1: I think we can use the the `*` to clean up the redirects a little. 

```toml
[[redirects]]
from = "/showcase/example-repos/*"
to = "/docs/quickstart"

[[redirects]]
from = "/showcase/benchmarks/*"
to = "/docs/reference/benchmarks/:splat"
```

or we can just do that as a follow up next week if this file becomes too long.

barbados-clemens (approved):
does this only work on the canary deploy or will docs previews also use the NEXT_PUBLIC_ASTRO_URL var?

nx-dev/nx-dev/redirect-rules.js
```diff
@@ -892,6 +890,14 @@ const missingAndCatchAllRedirects = {
   '/packages/:path*': '/nx-api/:path*',
 };
 
+if (process.env['NEXT_PUBLIC_ASTRO_URL']) {
+  missingAndCatchAllRedirects['/getting-started/:path*'] =
+    '/docs/getting-started/intro';
```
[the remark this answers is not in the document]
jaysoo: This is not true

astro-docs/netlify.toml
jaysoo: Only for canary for now because I set it for `master` branch only, but we can set it for all previews as well.

PR #32501, opened 2025-08-25
fix(nx-dev): task graph should not break docs

nartc:
There was a recent change to `TaskGraphClientResponse` where we drop the multiple `taskGraphs` and replace with single `taskGraph`. Goal was to improve the performance for `nx graph` task graph in general and support multiple targets. Docs pages that render task graph were not updated to adhere to the new single task graph model. This PR fixes it.

This PR also adjusts the project graph to handle missing props more gracefully.

### DIFF
--- a/astro-docs/src/assets/nx/connected-tasks.json
+++ b/astro-docs/src/assets/nx/connected-tasks.json
@@ -31,34 +31,32 @@
       }
     }
   ],
-  "taskId": "lib:test",
-  "taskGraphs": {
-    "lib:test": {
-      "roots": ["app1:test", "app2:test", "lib:test"],
-      "tasks": {
-        "app1:test": {
-          "id": "app1:test",
-          "target": { "project": "app1", "target": "test" },
-          "projectRoot": "apps/app1",
-          "overrides": {}
-        },
-        "app2:test": {
-          "id": "app2:test",
-          "target": { "project": "app2", "target": "test" },
-          "projectRoot": "apps/app2",
-          "overrides": {}
-        },
-        "lib:test": {
-          "id": "lib:test",
-          "target": { "project": "lib", "target": "test" },
-          "projectRoot": "libs/lib",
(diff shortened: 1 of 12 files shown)


MaxKless (approved):

astro-docs/src/assets/nx/connected-tasks.json
```diff
@@ -31,34 +31,32 @@
...
-          "projectRoot": "apps/app2",
-          "overrides": {}
-        },
-        "lib:test": {
-          "id": "lib:test",
-          "target": { "project": "lib", "target": "test" },
-          "projectRoot": "libs/lib",
-          "overrides": {}
-        }
+  "taskIds": ["app1:test", "app2:test"],
```
nartc, line 34: Note to reviewer: This is for "connected tasks" so we should see the tasks being connected somehow (i.e: task dependencies)

Previously, `taskId` was set to `lib:test` and the graph rendered "correctly". However, it was a bug. What `taskId` do is that it will send an event to `@nx/graph` 

```ts
send({ type: 'show', taskIds: [taskId] });
```

This tells the Task Graph to "render the task with said taskId". The thought process here is: "the graph will show you the task as if you run it via nx". This means that if the task doesn't have any dependencies (i.e: running the task will not run any other tasks), task graph will only render the task itself and not any of the task's dependents.

With this in mind, sending just `lib:test` to the Task Graph will just show the `lib:test` task instead of all 3: `app1:test`, `app2:test`, and `lib:test`. This was a bug in the legacy task graph cytoscape logic.

Currently, we set it to `taskIds: ["app1:test", "app2:test"]` to say that: if nx runs `app1:test` and `app2:test`, nx will also run `lib:test` first because `lib:test` is a dependency of the apps' `test` tasks. This is so the Task Graph will render accordingly to how `nx` behaves.

barbados-clemens (approved):

PR #32531, opened 2025-08-27
fix(release): ensure versionData is accurate for all transitive dependencies

mpsanchis:
## Current Behavior

In a repo with transitive dependencies, manifests are correctly bumped with `patch` versions. However, the `versionData` object is not correctly populated with the `newVersion`s that are caused by such transitive dependencies.

This is particularly harmful when using a custom release script, which leverages the programmatic API of nx release, and expects `await releaseVersion(...)` to return a `VersionData` object which reflects all versions, new and current.

## Expected Behavior

What we find in the manifests matches exactly what we find in `versionData`.

## Note

One test showing the behaviour has been updated. Without the code change, `jest` logs:
```
    expect(received).toEqual(expected) // deep equality

    Expected: "1.0.1"
    Received: null

      546 |
      547 |       const versionData = processor.getVersionData();
    > 548 |       expect(versionData['pkg-a'].newVersion).toEqual('1.0.1');
```


mpsanchis:
CC @JamesHenry

JamesHenry (commented):
Thanks for catching this! I've updated the branch as it seems the last CI run had some temporary issues

JamesHenry:
✅ Thanks for this @mpsanchis

PR #32534, opened 2025-08-27
fix(release): optimize release version internals

mpsanchis:
ReleaseGroupProcessor already iterates through all release groups, and sometimes change is anyways not propagated.

## Current Behavior

All release groups are iterated in a way that, if group A depends on group B:
- A will [propagate](https://github.com/nrwl/nx/blob/53bb276a39a2de09938ed1161fadbbe92a96e596/packages/nx/src/command-line/release/version/release-group-processor.ts#L832) changes to B
- B will [check](https://github.com/nrwl/nx/blob/53bb276a39a2de09938ed1161fadbbe92a96e596/packages/nx/src/command-line/release/version/release-group-processor.ts#L894) if it should be bumped by its dependencies 

This means there is redundant double-checking in the code.

However, propagation doesn't always work. The linked [snippet](https://github.com/nrwl/nx/blob/53bb276a39a2de09938ed1161fadbbe92a96e596/packages/nx/src/command-line/release/version/release-group-processor.ts#L1626-L1661) selects a random project of group *A*, and if that project doesn't have dependencies to projects in group *B*, it doesn't consider *B* as bumped.

### Ideally
There would be no double loops. 

### Otherwise 
If there are, they should consider all projects when deciding what to bump. For instance, with something like:
```typescript
      const hasDependencyInChangedGroupV2 = Array.from(releaseGroupFilteredProjects).some(
        (project) => {
          const dependencies = this.projectGraph.dependencies[project] || [];
          return dependencies.some(
            (dep) =>
              this.getReleaseGroupNameForProject(dep.target) ===
              changedDependencyGroup
          )
      })
```

## Notes

- I know it's a bold move to simply delete code. I wanted to highlight that, if tests pass, we are either missing tests or we are doing unnecessary work. I would like to replicate locally with a bit more confidence, but `jest` tests seem to be flaky 
- I would need to pick @JamesHenry 's brain about what's the intention with this "double" propagation in order to have the full context here

### DIFF
--- a/packages/nx/src/command-line/release/version/release-group-processor.ts
+++ b/packages/nx/src/command-line/release/version/release-group-processor.ts
@@ -827,21 +827,6 @@ export class ReleaseGroupProcessor {
       const projectLogger = this.getProjectLoggerForProject(project);
       projectLogger.flush();
     }
-
-    if (bumped) {
-      await this.propagateChangesToDependentGroups(releaseGroupName);
-    }
-  }
-
-  private async propagateChangesToDependentGroups(
-    changedReleaseGroupName: string
-  ): Promise<void> {
-    const changedGroupNode = this.groupGraph.get(changedReleaseGroupName)!;
-    for (const depGroupName of changedGroupNode.dependents) {
-      if (!this.processedGroups.has(depGroupName)) {
-        await this.propagateChanges(depGroupName, changedReleaseGroupName);
-      }
-    }
   }
 
   private async bumpVersions(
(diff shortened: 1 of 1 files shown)


JamesHenry (changes requested):
Definitely news to me that release unit tests are flaky? I've never seen that and nobody has ever reported it. You do have a failure in release unit tests here, are you saying you cannot reproduce it?

mpsanchis:
@JamesHenry I have rebased my branch, and locally I get no failing tests (only test-native sometimes, but not important for us now). It doesn't look like I re-triggered CI, though

JamesHenry:
Well... it's green. That's pretty wild. It's been almost exactly a year since this code was first written (it was prototyped a couple of months before it actually got merged in) and I simply don't remember any specific requirements here, so it's possible that further improvements and refactorings that came later simply removed the need for this behaviour.

I'm very grateful to you for finding this and addressing it. Like you, I am also wary about us potentially missing test cases for this behaviour and it actually being needed for certain use-cases, but it's going to be very hard for us to "prove a negative".

I think the best way for us to proceed is merge this in as a _fix_ even though it theoretically does not change anything. This will allow us to clearly trace it on the changelog and revert it/refine it if we do discover issues in practice.

Thanks again!

JamesHenry (approved):

mpsanchis:
Thanks to you @JamesHenry 

I'll keep an eye on this as usual ;)

JamesHenry:
21.5.2 was released containing this change 🤞

PR #32557, opened 2025-08-28
fix(core): check nx packages for provenance config before running nx migrate

MaxKless:
## Current Behavior
The nx migrate command installs nx@latest without verifying package provenance, which could expose users to security risks if a malicious version were published (as noted in the security advisory).

## Expected Behavior
The migrate command now verifies that nx@latest has valid provenance attestations before installation to ensure package security. If the package doesn't have provenance or the check fails, the migration is aborted with a clear error message.

Also, when installing packages or executing them, we should default to using `--ignore-scripts`.

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/32550


jaysoo (approved):

packages/nx/src/utils/provenance.ts
```diff
@@ -0,0 +1,95 @@
...
+      )
+    ).stdout.trim()
+  );
+
+  const attURL = npmViewResult.dist?.attestations?.url;
+
+  if (!attURL) return { success: false, error: 'No attestation URL found' };
+
+  // refer to https://slsa.dev/spec/v1.1/provenance#schema for the shape of this object
+  const attestations = (await (await fetch(attURL)).json()) as any;
```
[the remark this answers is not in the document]
FrozenPandaz: @MaxKless what do you think of this?

packages/nx/src/utils/provenance.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+      )
+    ).stdout.trim()
+  );
+
+  const attURL = npmViewResult.dist?.attestations?.url;
+
+  if (!attURL) return { success: false, error: 'No attestation URL found' };
+
+  // refer to https://slsa.dev/spec/v1.1/provenance#schema for the shape of this object
```
FrozenPandaz: Can we type this?

packages/nx/src/utils/provenance.ts
```diff
@@ -0,0 +1,95 @@
...
+      ?.workflow;
+
+  // verify that provenance was actually generated from the right publishing workflow
+  if (workflowParameters?.repository !== 'https://github.com/nrwl/nx') {
+    return { success: false, error: 'Repository does not match nrwl/nx' };
+  }
+  if (workflowParameters?.path !== '.github/workflows/publish.yml') {
+    return {
+      success: false,
+      error: 'Publishing workflow does not match .github/workflows/publish.yml',
```
FrozenPandaz: This would be more idiomatic if it threw an Error vs not throw an Error (indicating success).

Also, I think the error messages should show remediation steps. Maybe double-check if the version is actually proper. And also show that this check can be skipped so users have a way around it if for some reason the version if proper.

I also know.. sometimes I visit npmjs.com and it has trouble fetching provenance? 🤔 If the attestation servers are down, users might want to skip this 🤷

FrozenPandaz (changes requested):

packages/nx/src/utils/provenance.ts
MaxKless: yes, I updated it to throw an error, it's cleaner.
Also added an instruction to check the package at npmjs.org.

I don't know about the last bit, I would be very hesitant about adding an instruction to skip here as people don't really read error messages and if they don't know what provenance is, they might just ignore the warning and skip the check.

packages/nx/src/utils/provenance.ts
```diff
@@ -0,0 +1,177 @@
...
+    return;
+  }
+
+  const execFileAsync = promisify(execFile);
+
+  const npmViewResult = JSON.parse(
+    (
+      await execFileAsync(
+        'npm',
+        ['view', `${packageName}@${packageVersion}`, '--json', '--silent'],
```
FrozenPandaz, line 29: ` npm view nx dist.attestations.url` This should give you the result as a string.

packages/nx/src/utils/provenance.ts
```diff
@@ -0,0 +1,177 @@
...
+ */
+export async function ensurePackageHasProvenance(
+  packageName: string,
+  packageVersion: string
+): Promise<void> {
+  // this is used for locally released versions without provenance
+  // do not set this for other reasons or you might be exposed to security risks
+  if (process.env.NX_MIGRATE_SKIP_PROVENANCE_CHECK) {
+    return;
+  }
```
FrozenPandaz: Hm, we should make this variable more generic because it might not just be for `migrate`.

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -990,6 +994,9 @@ async function getPackageMigrationsUsingRegistry(
   packageName: string,
   packageVersion: string
 ): Promise<ResolvedMigrationConfiguration> {
+  if (packageName.startsWith('@nx/') || packageName === 'nx') {
```
FrozenPandaz: This will probably break things like `nx migrate @nx/php@latest` since that's not in the `nrwl/nx` repo.

packages/nx/src/utils/package-manager.ts
```diff
@@ -128,20 +128,21 @@ export function getPackageManagerCommand(
         useBerry = true;
       }
 
+      // new versions of yarn only support ignoring scripts via .yarnrc.yml
       return {
         preInstall: `yarn set version ${yarnVersion}`,
-        install: 'yarn',
+        install: useBerry ? 'yarn install' : `yarn install --ignore-scripts`,
```
FrozenPandaz: Hm `dlx` esque commands make sense.. but I'm not entirely sure about `install`. What if we keep this isolated to `dlx` for now? 🤔

packages/workspace/src/generators/ci-workflow/ci-workflow.spec.ts
```diff
@@ -392,22 +392,22 @@ describe('CI Workflow generator', () => {
...
+              # - run: npx --ignore-scripts nx start-ci-run --distribute-on="3 linux-medium-js" --stop-agents-after="build"
 
               # Cache node_modules
               - uses: actions/setup-node@v4
                 with:
                   node-version: 20
                   cache: 'npm'
 
-              - run: npm ci --legacy-peer-deps
+              - run: npm ci --legacy-peer-deps --ignore-scripts
```
FrozenPandaz: I can see this causing issues.

packages/workspace/src/generators/ci-workflow/ci-workflow.spec.ts
```diff
@@ -392,22 +392,22 @@ describe('CI Workflow generator', () => {
...
               - uses: nrwl/nx-set-shas@v4
 
               # Prepend any command with "nx-cloud record --" to record its logs to Nx Cloud
-              # - run: npx nx-cloud record -- echo Hello World
-              - run: npx nx affected -t lint test build typecheck
+              # - run: npx --ignore-scripts nx-cloud record -- echo Hello World
+              - run: npx --ignore-scripts nx affected -t lint test build typecheck
               # Nx Cloud recommends fixes for failures to help you get CI green faster. Learn more: https://nx.dev/ci/features/self-healing-ci
-              - run: npx nx fix-ci
+              - run: npx --ignore-scripts nx fix-ci
```
FrozenPandaz: Not sure this makes sense 🤔 This shouldn't install nx at this point. I'm thinking we might have to make this an optional.. or a separate command so that we don't add this where we don't intend to?

packages/nx/src/utils/package-manager.ts
```diff
@@ -128,20 +128,21 @@ export function getPackageManagerCommand(
...
           ? 'yarn install --mode update-lockfile'
           : 'yarn install',
-        add: useBerry ? 'yarn add' : 'yarn add -W',
-        addDev: useBerry ? 'yarn add -D' : 'yarn add -D -W',
+        add: useBerry ? 'yarn add' : 'yarn add --ignore-scripts -W',
+        addDev: useBerry ? 'yarn add -D' : 'yarn add -D --ignore-scripts -W',
         rm: 'yarn remove',
         exec: 'yarn',
-        dlx: useBerry ? 'yarn dlx' : 'yarn',
+        dlx: useBerry ? 'yarn dlx' : 'npx --ignore-scripts',
```
FrozenPandaz: I'm thinking it might make more sense to only use `--ignore-scripts` when we expect the package is not installed.. would not make more sense? I'm just not as much of a fan of this showing up in all the `--help` information and in the generated workflow and stuff. I don't even see where this is needed for `migrate`... did I miss something?

FrozenPandaz (changes requested):

packages/nx/src/utils/provenance.ts
MaxKless: we need more than just the url further down

packages/nx/src/command-line/migrate/migrate.ts
MaxKless: okay, let's not check for those two then as they're separate

FrozenPandaz (approved):

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -990,6 +994,13 @@ async function getPackageMigrationsUsingRegistry(
   packageName: string,
   packageVersion: string
 ): Promise<ResolvedMigrationConfiguration> {
+  if (
+    (packageName.startsWith('@nx/') || packageName === 'nx') &&
+    packageName !== '@nx/php' &&
```
FrozenPandaz: Hm, this is also a bad way to do it... shouldn't we have a list somewhere?

FrozenPandaz (approved):

PR #32559, opened 2025-08-28
fix(nx-dev): issue with index pages not showing content lists

barbados-clemens:
also fixed issue where links contained 'CI Features' instead of
'ci-features' in URL breaking the markdown to html link rendering

Fixes: DOC-181
Fixes: DOC-186

### DIFF
--- a/astro-docs/src/components/markdoc/IndexPageCards.astro
+++ b/astro-docs/src/components/markdoc/IndexPageCards.astro
@@ -7,7 +7,10 @@ export interface Props {
   path: string;
 }
 
-const { path } = Astro.props;
+const { path: rawPath } = Astro.props;
+// make sure paths are normalized to url slugs 
+// i.e. features/CI Features -> features/ci-features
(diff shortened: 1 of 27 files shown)


astro-docs/src/components/markdoc/IndexPageCards.astro
```diff
@@ -7,7 +7,10 @@ export interface Props {
   path: string;
 }
 
-const { path } = Astro.props;
+const { path: rawPath } = Astro.props;
+// make sure paths are normalized to url slugs 
+// i.e. features/CI Features -> features/ci-features
+const path = rawPath.replace(/\s/, '-').toLowerCase();
```
jaysoo, line 13: Do we need to strip non-alphanumeric characters like `&`, or is that already not in the path?

jaysoo: Oh wait, I got confused, it's out own path. Ignore.

jaysoo (approved):

PR #32565, opened 2025-08-28
fix(nx-dev): correctly link to url fragments for devkit

barbados-clemens:
switch to making typedoc to the rendering and then pull in the generated file contents instead of manually parsing the reflection metadata.

Fixes: DOC-63


astro-docs/src/plugins/utils/markdown-formatters.ts
```diff
@@ -0,0 +1,348 @@
...
+  ) {
+    // astro/starlight lowercases all url fragments and is how the ToC links to headers
+    return `${baseUrl}/${part.target.parent.name.replace(
+      /\s/g,
+      ''
+    )}#${normalizedUrlText.toLowerCase()}`;
+  }
+
+  return `${baseUrl}/${normalizedUrlText}`;
+}
```
barbados-clemens: this is functionally the fix

astro-docs/src/plugins/utils/devkit-generation.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    const childReflections = reflections.filter(
+      (r) => r.parent && r.parent.id === reflection.id
+    );
+    const { content, ...metadata } = formatDeclarationToMarkdown(
+      reflection as DeclarationReflection,
+      childReflections as DeclarationReflection[],
+      packageType
+    );
```
barbados-clemens: apparently `reflection.children` does not contain all the children. so we manually filter for them and pass them through. this is why some of the devkit pages were empty.

astro-docs/src/plugins/utils/markdown-formatters.ts
```diff
@@ -0,0 +1,348 @@
...
+      }
+      content += '\n';
+    }
+  }
+
+  // we don't use the reflection.children since it seems to be a partial list of children
+  // and merging the two lists can result in type lost, where one reflection.children intrinsic types seem almost always be `any`
+  const allChildren = childRefs.sort((a, b) => a.name.localeCompare(b.name));
+  const seenProperties = new Set<number>();
+  const seenMethods = new Set<number>();
```
barbados-clemens: I was sometimes getting duplicated properties/methods showing up so I track the ids i've seen. this could be an artifact of some other settings/logic that I have incorrectly impl'd but leaving it for now and we can sanity check it on some cleanup passes.

astro-docs/e2e/devkit-urls-fragments.spec.ts
barbados-clemens, line 1: added e2e test for the behavior of link goes to same page as URL fragment to prevent any regressions in the future since it can be hard to know if/when you break it.

astro-docs/src/plugins/utils/devkit-generation.ts
```diff
@@ -0,0 +1,161 @@
...
+    const documentRecord: CollectionEntry<'nx-reference-packages'> = {
+      id: `devkit_${slug.replace(/\//g, '-')}`,
+      body: content,
+      // @ts-expect-error astro auto gen types don't align
+      rendered,
+      collection: 'nx-reference-packages',
+      data: {
+        title: title,
+        packageType: 'devkit',
+        docType: slug,
```
jaysoo: Should we just add `slug` to the schema here? Looks like it has both `docType` and `category` in the schema, but I'm unsure what the difference is, as both seem to be for `generators`, `overview`, etc. But in this context the `docType` is being used as a slug.

barbados-clemens: Yes we should. I meant to refactor the schema and just forgot. I’ll fix it up

jaysoo (approved):

PR #32583, opened 2025-09-02
feat(core): add id, start and end time to lifecycle hooks

meeroslav:
## Current Behavior
Pre and post hooks lack corellation Id which might make it difficult to culculate stats for task using pre and post hooks.
Additionally, post hook does not expose the duration or time span for the task. Using pre/post hook might not be precise enough.

## Expected Behavior
Hooks expose the unique taskId so we can corellate pre hook of a task to the post hook of the same task. Post hook should expose start and end time of the task run.

## Related Issue(s)

Resolves discussion [31076](https://github.com/nrwl/nx/discussions/31076)


Hardanish-Singh (approved):

packages/nx/src/project-graph/plugins/loaded-nx-plugin.ts
```diff
@@ -1,5 +1,5 @@
 import type { ProjectGraph } from '../../config/project-graph';
-import { readNxJson, type PluginConfiguration } from '../../config/nx-json';
```
meeroslav, line 2: These are unused

packages/nx/src/tasks-runner/life-cycle.ts
```diff
@@ -22,14 +22,6 @@ export interface TaskMetadata {
   groupId: number;
 }
 
-interface RustRunningTask extends RunningTask {
```
meeroslav, line 25: This was unused

FrozenPandaz (changes requested):
This seems very easy to get if users run `Date.now()` in their pre and post hooks. I don't think we need to be the one providing this.

meeroslav:
> This seems very easy to get if users run Date.now() in their pre and post hooks. I don't think we need to be the one providing this.

@FrozenPandaz, it might not be trivial when you run several commands in parallel. Providing this information out of the box instead of forcing the user to track the collection of start times in memory per args combo is better, in my opinion. Also, it prevents other (potentially long-running) Pre and Post hooks from skewing the numbers.

packages/nx/src/command-line/release/publish.ts
```diff
@@ -272,11 +273,14 @@ async function runPublishOnProjects(
...
     );
   }
+  const taskId = hashArray([...process.argv, Date.now().toString()]);
   await runPreTasksExecution({
+    taskId,
     workspaceRoot,
     nxJsonConfiguration: nxJson,
     argv: process.argv,
   });
+  const startTime = Date.now();
```
FrozenPandaz, line 283: Might as well move this up and use it as both `startTime` and part of the hash for the ID

packages/nx/src/command-line/release/publish.ts
```diff
@@ -272,11 +273,14 @@ async function runPublishOnProjects(
       ].join('\n')}\n`
     );
   }
+  const taskId = hashArray([...process.argv, Date.now().toString()]);
```
FrozenPandaz: Please call this a `tasksExecutionId` or `executionId`... or even just id. Just definitely not `task.id`

packages/nx/src/project-graph/plugins/public-api.ts
```diff
@@ -190,15 +190,19 @@ export type NxPluginV2<TOptions = unknown> = {
 };
 
 export type PreTasksExecutionContext = {
+  readonly taskId: string;
```
FrozenPandaz: We can make this just `id`

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -68,6 +68,7 @@ import { TasksRunner, TaskStatus } from './tasks-runner';
 import { shouldStreamOutput } from './utils';
 import { signalToCode } from '../utils/exit-codes';
 import chalk = require('chalk');
+import { randomUUID } from 'node:crypto';
```
FrozenPandaz: Doesn't look to be used.

FrozenPandaz (approved):

packages/nx/src/command-line/release/publish.ts
meeroslav: I thought about that, but if would diminish the preciseness of the information since it would also include time spent on running the pre tasks hooks.

packages/nx/src/project-graph/plugins/public-api.ts
meeroslav: I changed it to just `id`.

packages/nx/src/tasks-runner/run-command.ts
meeroslav: Oops, AI autocomplete leftover. Let me kick that out.

packages/nx/src/command-line/release/publish.ts
```diff
@@ -272,11 +273,14 @@ async function runPublishOnProjects(
       ].join('\n')}\n`
     );
   }
+  const id = hashArray([...process.argv, Date.now().toString()]);
```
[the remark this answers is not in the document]
meeroslav: The same id value is used for both pre and post hook so they will match. We intentionally want to use different Id for different invocation of the same command, thus argv itself is not enough.

PR #32626, opened 2025-09-05
fix(core): invalidate project graph when external nodes change

leosvelperez:
## Current Behavior

When external nodes change, the project graph cache is not invalidated.  This can sometimes lead to missing dependencies due to reusing cached dependencies for nodes that no longer exist.

## Expected Behavior

When external nodes change, the project graph cache should be invalidated.

### DIFF
--- a/packages/nx/src/project-graph/nx-deps-cache.ts
+++ b/packages/nx/src/project-graph/nx-deps-cache.ts
@@ -26,6 +26,7 @@ import {
 } from './error-types';
 import { isOnDaemon } from '../daemon/is-on-daemon';
 import { serverLogger } from '../daemon/server/logger';
+import { hashObject } from '../hasher/file-hasher';
 
 export interface FileMapCache {
   version: string;
@@ -34,7 +35,7 @@ export interface FileMapCache {
   nxJsonPlugins: PluginData[];
   pluginsConfig?: any;
   fileMap: FileMap;
-  externalNodes?: Record<string, string>;
+  externalNodes?: string;
 }
 
 export const nxProjectGraph = join(
@@ -187,7 +188,7 @@ export function createProjectFileMapCache(
   externalNodes: Record<string, ProjectGraphExternalNode>
 ) {
   const nxJsonPlugins = getNxJsonPluginsData(nxJson, packageJsonDeps);
-  const externalNodesData = getExternalNodesData(externalNodes);
+  const externalNodesHash = hashObject(getExternalNodesData(externalNodes));
   const newValue: FileMapCache = {
     version: '6.0',
     nxVersion: nxVersion,
@@ -196,7 +197,7 @@ export function createProjectFileMapCache(
     nxJsonPlugins,
     pluginsConfig: nxJson?.pluginsConfig,
     fileMap,
-    externalNodes: externalNodesData,
+    externalNodes: externalNodesHash,
   };
   return newValue;
 }
@@ -325,10 +326,7 @@ export function shouldRecomputeWholeGraph(
   }
 
   // Check if external nodes have changed
-  if (
-    hashExternalNodes(getExternalNodesData(externalNodes)) !==
-    hashExternalNodes(cache.externalNodes)
-  ) {
+  if (hashObject(getExternalNodesData(externalNodes)) !== cache.externalNodes) {
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -445,3 +458,27 @@ function getNxJsonPluginsData(
     };
   });
 }
+
+function getExternalNodesData(
+  externalNodes: Record<string, ProjectGraphExternalNode>
+): Record<string, string> {
+  return Object.entries(externalNodes).reduce((acc, [name, node]) => {
+    acc[name] = node.data.version;
```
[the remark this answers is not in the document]
leosvelperez: The definition of the external node has data and version as mandatory, so they shouldn't be missing:

https://github.com/nrwl/nx/blob/27e8ce077184472347ddb4c0a1be9d75d4f405ec/packages/nx/src/config/project-graph.ts#L122-L130

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -445,3 +458,27 @@ function getNxJsonPluginsData(
...
+    return acc;
+  }, {});
+}
+
+function hashExternalNodes(
```
FrozenPandaz: There's a `hashObject` function which sorts.. does that work? 

https://github.com/nrwl/nx/blob/76580dbd93a581f27c123cfdaaa6ac0292877fdc/packages/nx/src/hasher/file-hasher.ts#L7

FrozenPandaz (changes requested):

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -193,6 +196,7 @@ export function createProjectFileMapCache(
     nxJsonPlugins,
     pluginsConfig: nxJson?.pluginsConfig,
     fileMap,
+    externalNodes: externalNodesData,
```
FrozenPandaz: I don't think we should write the entire `externalNodesData`.. can we write just the hash?

It'll save us disk space used by the file.. and it'll save us computing the hash again.

FrozenPandaz (changes requested):

packages/nx/src/project-graph/nx-deps-cache.ts
leosvelperez: Yeah, with the hash, it makes sense. With the output of `JSON.stringify` I had, less so, though we would still reduce computation for sure.

If we're interested in reducing disk space used by the file, we should also consider writing an unformatted JSON to disk with no spaces. If anyone wants to manually read it, they can easily format it anyway.

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -182,9 +184,11 @@ export function createProjectFileMapCache(
   nxJson: NxJsonConfiguration<'*' | string[]>,
   packageJsonDeps: Record<string, string>,
   fileMap: FileMap,
-  tsConfig: { compilerOptions?: { paths?: { [p: string]: any } } }
+  tsConfig: { compilerOptions?: { paths?: { [p: string]: any } } },
+  externalNodes: Record<string, ProjectGraphExternalNode>
 ) {
   const nxJsonPlugins = getNxJsonPluginsData(nxJson, packageJsonDeps);
+  const externalNodesHash = hashObject(getExternalNodesData(externalNodes));
```
FrozenPandaz: Are we hashing it twice? One to compare and once to write it? We should only hash it once.

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -445,3 +456,12 @@ function getNxJsonPluginsData(
     };
   });
 }
+
+function getExternalNodesData(
```
FrozenPandaz: Do we have to reduce this or can we hash the whole thing?

FrozenPandaz (changes requested):

packages/nx/src/project-graph/nx-deps-cache.ts
leosvelperez: I think we should. Benchmarking both approaches in this repo:

```bash
Hash full external nodes: 27.798042000000002ms
Reduce and hash external nodes: 4.348999999999997ms
```

That's a ~6x time reduction, which is significant.

FrozenPandaz (approved):

PR #32666, opened 2025-09-09
fix(nx): `nx affected` docs options shorthand should use single dash

ericbf:
The options shorthand uses single dash, not double dash. See [docs](https://nx.dev/reference/core-api/nx/documents/affected#options).

## Current Behavior
The documentation has a double dash for options shorthand in `nx affected` docs.

## Expected Behavior
Should be single dashes.


JamesHenry (changes requested):
Thanks @ericbf but our docs are auto-generated (you can see the file path says `generated`), so you need to please update the source of truth and then run `pnpm nx documentation`

ericbf:
Don’t know where that is. I’ll just close this and create an issue instead.

ericbf:
Opened #32723.

PR #32698, opened 2025-09-11
docs(misc): add sidebar icons for Technologies section

jaysoo:
Add SVG icons to the Technologies section sidebar items to improve visual hierarchy and navigation. Icons are displayed next to technology labels like TypeScript, Angular, React, Vue, Node.js, Java, etc.

`Sidebar` is now using a custom `SidebarSublist` that uses `data-icon` (if it exists) to render the SVG images. A middleware is used since `attrs` cannot be set on sidebar items with children. Alternatively, we could cast the sidebar config to avoid type errors.

image

image

Closes #DOC-197


astro-docs/src/plugins/sidebar-icons.middleware.ts
```diff
@@ -0,0 +1,37 @@
...
+  Vue: 'vuedotjs',
+  'Node.js': 'nodedotjs',
+  Java: 'java',
+  'Module Federation': 'module-federation',
+  ESLint: 'eslint',
+  'Build Tools': 'buildtools',
+  'Test Tools': 'testtools',
+};
+
+function addIconsToSidebar(sidebar: any[]): void {
```
jaysoo, line 20: Can try casting if it passes zod validation.

astro-docs/public/images/icons/nx.svg
jaysoo, line 1: Don't need this or nuxt.

astro-docs/src/plugins/sidebar-icons.middleware.ts
jaysoo: Doesn't work since it leads to validation error anyway.

astro-docs/src/components/layout/SidebarSublist.astro
```diff
@@ -0,0 +1,197 @@
...
+        <li class={icon ? 'p-0 border-none' : ''}>
+          {entry.type === 'link' ? (
+            <a
+              href={entry.href}
+              aria-current={entry.isCurrent && 'page'}
+              class:list={[{large: !nested}, entry.attrs?.class]}
+              {...entry.attrs}
+            >
+              {icon && (
+                <img
```
barbados-clemens, line 65: we probs want `aria-hidden="true"` on this since it's just decoration and shouldn't get picked up by screen readers. 
```suggestion
                <img
                aria-hidden="true"
```

astro-docs/src/components/layout/SidebarSublist.astro
```diff
@@ -0,0 +1,197 @@
...
+                <Badge variant={entry.badge.variant} class={entry.badge.class}
+                       text={entry.badge.text}/>}
+            </a>
+          ) : (
+            <details
+              open={flattenSidebar(entry.entries).some((i: any) => i.isCurrent) || !entry.collapsed}>
+              <summary class={icon ? 'pl-0 py-2' : ''}>
+                <div class="group-label">
+                  {icon && (
+                    <img src={icon} alt="" class="sidebar-icon w-4 h-4 dark:invert"/>
```
barbados-clemens: same above
```suggestion
                    <img src={icon} alt="" aria-hidden="true" class="sidebar-icon w-4 h-4 dark:invert"/>
```

barbados-clemens (approved):
1 small comment.

Issue #26466, opened 2024-06-07
`nx release`: changelog of independent project include workspace changes


Badisi:
### Current Behavior

Consider:
- the following project structure:
```
.
├── .github
│  └── workflows
│     └── ci_tests.yml
├── projects
│  └── package-a
│     └── ...
├── ...
```
- the following commit (that is the only one that exists after `package-a@1.2.0` was released):
```
+ fix: tests
  .github/workflows/ci_tests.yml
```
- the following `nx.json` configuration:
```json
"release": {
    "projectsRelationship": "independent",
    "version": {
        "conventionalCommits": true
    },
    "changelog": {
        "projectChangelogs": {
            "createRelease": "github"
        }
    }
}
```

----

Running `nx release` will trigger the following changelog for **package-a**:
*(even if nothing changed under `projects/package-a/` directory)*
```md
UPDATE projects/package-a/CHANGELOG.md [dry-run]

+ ## 1.2.1 (2024-06-07)
+
+ ### 🩹 Fixes
+
+ - tests ([c33cbae](https://github.com/org/repo/commit/c33cbae))
+ ...
```

----

The issue relies in `getCommitsRelevantToProjects()`:
https://github.com/nrwl/nx/blob/55197fb33fb2e52f7bb37da44b0298d981ff4682/packages/nx/src/command-line/release/utils/shared.ts#L322-L336
I don't know why commits that `"touch non-project files and are not scoped"` are intentionally marked as `"relevant"` here, but this is clearly the issue..

### Expected Behavior

Workspace changes should not be included in independent project changelog.

### Steps to Reproduce

None

### Nx Report

```shell
Node   : 18.18.2
OS     : darwin-x64
npm    : 10.8.0

nx (global)    : 19.0.4
nx             : 19.2.0
@nx/js         : 19.2.0
@nx/workspace  : 19.2.0
@nx/devkit     : 19.2.0
@nrwl/tao      : 19.2.0
typescript     : 4.7.4
```

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

thdk:
@Badisi if your commit convention allows you to use other types than `fix` or `feat` you can configure your nx release command to ignore certain types of commits when the release command needs to determine the next version.

See the documentation here: https://nx.dev/recipes/nx-release/customize-conventional-commit-types

[...]

PR #32732, opened 2025-09-15
feat(release)!: process non-release dependencies of release projects

Coly010:
## Current Behavior
When processing projects within Nx Release, only projects that are also configured within Nx Release are considered when determining dependencies of projects.

## Expected Behavior
All dependencies of projects within Nx Release, whether they are included in the Nx Release Config or not, should be added.

BREAKING CHANGE: All dependencies of projects within Nx Release Config are considered and processed now when determining version bumps and changelog generation.

### DIFF
--- a/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts
+++ b/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts
@@ -11,6 +11,7 @@ export async function resolveSemverSpecifierFromConventionalCommits(
   from: string,
   projectGraph: ProjectGraph,
   projectNames: string[],
+  projectToDependencies: Map<string, Set<string>>,
   conventionalCommitsConfig: NxReleaseConfig['conventionalCommits']
 ): // Map of projectName to semver bump type
 Promise<Map<string, SemverSpecifier | null>> {
@@ -19,7 +20,8 @@ Promise<Map<string, SemverSpecifier | null>> {
   const relevantCommits = await getCommitsRelevantToProjects(
     projectGraph,
     parsedCommits,
-    projectNames
+    projectNames,
+    projectToDependencies
   );
   return determineSemverChange(relevantCommits, conventionalCommitsConfig);
 }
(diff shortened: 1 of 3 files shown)


studds:
Hi @Coly010, thanks for working on this issue, just sat down to implement automatic versioning and a bit blocked at the moment, so very excited to see a PR in the works.

It strikes me that the outcome here should line up with the results of `nx show projects --affected`, which is not quite the case at present. This PR will include dependent projects, but will still miss shared global named inputs (see also #26466). The key logic seems to be in `filterAffected`, which uses `TouchedProjectLocator`s to assess an array of `FileChange`.

This is the process that `nx show projects --affected` follows:

1. Find the list of commits
2. Get the `FileChange[]` for **all** commits.
3. Call `filterAffected` to get affected projects.

By contrast, I believe what needs to happen here is:

1. Find the list of commits
2. Filter out commits which don't affect the target projects, by getting `FileChange[]` for **each** commit, and calling `filterAffected`.

Here's a naive implementation:

```typescript

export async function getCommitsRelevantToProjects(
  projectGraph: ProjectGraph,
  commits: GitCommit[],
  projects: string[]
): Promise<GitCommit[]> {
  const allWorkspaceFiles = await allFileData();
  const affectedCommits: GitCommit[] = [];
  for (const commit of commits) {
    // assess file changes for the individual commit
    const fileChangesForCommit = calculateFileChanges(
      commit.affectedFiles,
      allWorkspaceFiles
    );
    // find which projects this commit affects
    const affectedProjectsForCommit = await filterAffected(
      projectGraph,
      fileChangesForCommit
    );
    // include the commit if it impacts any of the targeted projects
    if (projects.some((project) => affectedProjectsForCommit.nodes[project])) {
      affectedCommits.push(commit);
    }
  }
  return affectedCommits;
}
```

I've (manually) tested this implementation in these scenarios, involving App A, App B, Shared Lib, and named global inputs:

- Update to App A only (only App A bumped)
- Update to App B only (only App B bumped)
- Updated to Shared Lib (App A and App B bumped)
- Update to named global input (App A and App B bumped)
- Update to other global file -ie. file not included in dependency graph at all (neither App A nor App B bumped)

Coly010:
Hi @studds 

Thanks for the comment. Yes, we're aware that this doesn't solve the issue where non-project files affect a project. 
However, it's more nuanced than just the update to a global named input. 

For example, a project may have only targets that do not use any named inputs. We also potentially don't want to mark a project for release if a named input only affected a non-build related task (linting, testing etc).

We're going to solve this by providing an `--affected-target` CLI flag that will default to `build`.

We can then use this to check all the input's to the provided target of a project and compare the files in a commit with the inputs to determine if project is affected.

PR #32742, opened 2025-09-16
docs(nx-dev): add biome plugin creation article

philipjfulcher:
https://nx-dev-git-philip-biome-lint-article-nrwl.vercel.app/blog/integrate-biome-in-20-minutes


docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
+---
+title: 'Build Your Own Nx Plugin: Integrating Biome in 20 Minutes'
```
juristr: I wouldn't necessarily lead with building an Nx plugin which might turn people off. Rather just mention using Biome with Nx or something like that.


Similarly in the intro paragraphs. Can we emphasize more that you don't **have to have** a plugin, but that simply will only make things smoother and your life easier if you have one?

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+
+3. **Run the linter**:
+
+```shell
+npx @biomejs/biome lint
+```
+
+Without thinking about Nx at all, we've successfully installed Biome and run a lint across our entire workspace. That's a good starting point, but in a monorepo, **we don't want to lint all files all the time**. We want to lint individual projects.
+
+## Using npm scripts
```
juristr: some of the titles are not coherent in the casing that's being used (sentence casing vs title casing)

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+{
+  "targets": {
+    "biome-lint": {
+      "command": "npx @biomejs/biome lint {projectRoot}"
+    }
+  }
+}
+```
+
+You'll notice one small difference: `{projectRoot}` appended to the end of the command. We'll explain this part when we're creating our inferred task plugin later, because this configuration is _exactly_ what you'll be creating as part of the plugin.
```
juristr: Should this just be a subsection of the using NPM scripts? Because I feel like I'm reading that using NPM script section, but I might not have an NPM workspace, and so I could just have a subsection there for those that use product.json files.

We should probably also link to the NPM workspace's docs, because there might still be people that are not familiar with what an NPM workspace is.

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+
+## Adding Caching Support
+
+One thing missing with this simple setup is **caching**. Nx doesn't know the `biome-lint` command is cacheable, and it doesn't know what to cache.
+
+Right now, if we rerun `nx biome-lint biome-example` twice, it actually runs the lint every time. Ideally, we shouldn't rerun this lint if we haven't changed anything between successful runs.
+
+### Configuring Target Defaults
```
juristr: I would just remove this subsection and keep it in the adding caching support section. Basically just have a sentence that says to configure caching, open your Nx JSON and add a target defaults, etc.

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+}
+```
+
+Now if we move into that directory and run `npm run biome-lint`, we're only linting that specific directory. When you run `biome lint` with no other options, it lints that directory and directories below it.
+
+## The Nx Integration you didn't realize was there
+
+Here's something you might not realize: **we've already started an Nx integration**. If you open Nx Console, you'll see that `biome-lint` is available as a target under the "npm scripts" area. Nx automatically incorporates npm scripts into available tasks.
+
+![Screenshot of Nx Console showing npm scripts available as targets](/blog/images/2025-09-16/npm-scripts.avif)
```
juristr: Is it possible that the target audience might not know how to reach this screen? Should we mention it somewhere? Either in a comment below the image or a callout box or something?

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+
+So while `npm run biome-lint` works, we can also run this through Nx:
+
+```shell
+npx nx biome-lint biome-example
+```
+
+This gives us the exact same result. We can copy this script to other packages, like our navigation library, and suddenly we can run `biome-lint` on multiple projects individually.
```
juristr, line 95: I would include a shell command that shows the run_many command with the -t biome-lint.

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+  // Project configuration to be merged into the rest of the Nx configuration
+  return {
+    projects: {
+      [root]: projectConfiguration,
+    },
+  };
+}
+```
+
+Let's take this example and make some changes:
```
juristr, line 222: Can you use the line highlighting feature to emphasize what parts changed?

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
```diff
@@ -0,0 +1,448 @@
...
+
+### Development Tips
+
+During plugin development, Nx caches plugin compilation. Use `NX_DAEMON=false` to bypass this cache:
+
+```shell
+NX_DAEMON=false nx run-many --target=biome-lint
+```
+
+When you're ready for production, run `nx reset` to clear the cache and the plugin will work normally.
```
juristr: I would keep this just in a call out box

juristr (changes requested):

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
philipjfulcher: Do we have a doc that's just focused on using npm workspaces? Everything I can find right now focuses on TypeScript Project References more than workspaces. I don't really want to complicate things by pointing to project references.

docs/blog/2025-09-16-creating-an-nx-plugin-for-biome.md
philipjfulcher: I added a short video showing how to open the details panel

juristr (approved):

PR #32744, opened 2025-09-16
fix(angular): install a compatible version of jest for angular

leosvelperez:
## Current Behavior

Generating Angular projects with Jest results in Jest v30 being used, which is not supported by Angular and causes peer dependency errors.

## Expected Behavior

Generating Angular projects with Jest should install a compatible version of Jest (Angular currently supports `^29.5.0`).


packages/angular/migrations.json
```diff
@@ -1915,21 +1915,6 @@
         }
       }
     },
-    "21.3.5-jest": {
```
leosvelperez, line 1918: We cannot change published versions, but with this change, if users migrate to the latest from a version lower than Nx v21.3.5, it will not bump the `jest-preset-angular` package.

I considered adding a migration to lower the version of `jest-preset-angular` (and Jest), but users in this situation are already handling it somehow, and it can get complex to revert some migration changes done for Jest v30.

packages/jest/migrations.json
```diff
@@ -19,9 +19,6 @@
     },
     "rename-test-path-pattern": {
       "version": "21.3.0-beta.3",
-      "requires": {
```
leosvelperez, line 22: This was incorrectly requiring Jest v30. Regardless of the Jest version, we renamed the executor option from that version of Nx onwards, so we need the migration to run.

Coly010 (approved):

jaysoo (approved):

PR #32754, opened 2025-09-17
feat(bundling)!: change default useLegacyTypescriptPlugin to false for v22

Coly010:
The default TypeScript plugin for Rollup has changed from rollup-plugin-typescript2 to @rollup/plugin-typescript. To continue using the legacy plugin, explicitly set useLegacyTypescriptPlugin: true in your configuration.

Resolves NXC-3094

### DIFF
--- a/e2e/rollup/src/rollup.test.ts
+++ b/e2e/rollup/src/rollup.test.ts
@@ -235,64 +235,6 @@ export default config;
   });
 
   describe('useLegacyTypescriptPlugin', () => {
-    it('should use @rollup/plugin-typescript by default when useLegacyTypescriptPlugin is not specified', async () => {
-      const myPkg = uniq('my-pkg');
-      runCLI(
-        `generate @nx/js:lib ${myPkg} --directory=libs/${myPkg} --bundler=rollup --no-interactive`
-      );
-      updateFile(
-        `libs/${myPkg}/src/index.ts`,
-        `export const hello = 'default';\n`
-      );
-
-      // Change TypeScript plugin build target to avoid conflict with Rollup
-      updateJson('nx.json', (nxJson) => {
-        nxJson.plugins.forEach((plugin) => {
-          if (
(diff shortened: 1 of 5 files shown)


e2e/rollup/src/rollup.test.ts
```diff
@@ -235,6 +235,64 @@ export default config;
   });
 
   describe('useLegacyTypescriptPlugin', () => {
+    it('should use @rollup/plugin-typescript by default when useLegacyTypescriptPlugin is not specified', async () => {
```
leosvelperez: nit/question: Do we need this test case?

I suggest not adding an extra e2e test case (which takes time to run) to cover the default value, which we can't explicitly check for the plugin usage, and we already check the build of a similar setup in other test cases.

packages/rollup/src/plugins/with-nx/with-nx.ts
```diff
@@ -250,14 +250,14 @@ export function withNx(
       image(),
       json(),
       // TypeScript compilation and declaration generation
-      // TODO(v22): Change default value of useLegacyTypescriptPlugin to false for Nx 22
-      options.useLegacyTypescriptPlugin !== false
+      options.useLegacyTypescriptPlugin === true
```
leosvelperez, line 253: Can switching the plugin cause issues or unexpected results? Just to confirm whether we need a migration to keep the previous behavior.

leosvelperez (approved):

e2e/rollup/src/rollup.test.ts
Coly010: We can remove it.

packages/rollup/src/plugins/with-nx/with-nx.ts
Coly010: There's a lot of nuance around it. It should be fine in majority of cases. There will of course be edge cases. 
I think this is fine without a migration, and we've had the deprecation warning for this reason.

PR #32756, opened 2025-09-17
feat(rspack)!: remove deprecated deleteOutputPath and sassImplementation options

Coly010:
## Current Behavior

  The rspack package contains deprecated options that were marked with TODO(v22) comments for
  removal:
  - deleteOutputPath option in the rspack executor schema (Line 79 in models.ts)
  - sassImplementation option in the rspack executor schema (Line 164 in models.ts)

  These deprecated options were still being referenced in the codebase and schema files,
  potentially causing confusion for users.

 ## Expected Behavior

  Remove the deprecated options from the rspack package to clean up the API for v22:
  - Remove deleteOutputPath option from the rspack executor and related configurations
  - Remove sassImplementation option from the rspack executor and related configurations
  - Add a migration to automatically update existing workspaces that use these deprecated options

##  Related Issue(s)
Resolves NXC-3112


packages/rspack/src/migrations/update-22-0-0/remove-deprecated-options.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+
+      // ACT
+      await removeDeprecatedOptions(tree);
+
+      // ASSERT
+      const configContent = tree.read(
+        'apps/test-app/rspack.config.ts',
+        'utf-8'
+      );
+      expect(configContent).toContain(
```
jaysoo: Nit: if we use inline snapshots with the full config it would be easier to see that it is valid after update.

jaysoo (approved):

packages/rspack/src/migrations/update-22-0-0/remove-deprecated-options.spec.ts
Coly010: +1

PR #32765, opened 2025-09-17
feat(testing): forward `e2e-ci` task options to their atomized tasks

leosvelperez:
## Current Behavior

Options configured in the `e2e-ci` tasks are not forwarded to their atomized tasks.

## Expected Behavior

Options configured in the `e2e-ci` tasks can be forwarded to their atomized tasks.

The `@nx/cypress/plugin` and `@nx/playwright/plugin` plugins will infer their `e2e-ci` dependencies to atomized tasks with `"options": "forward"`.

### DIFF
--- a/packages/playwright/src/plugins/plugin.ts
+++ b/packages/playwright/src/plugins/plugin.ts
@@ -123,6 +123,9 @@ async function createNodesInternal(
     {
       ...normalizedOptions,
       CI: process.env.CI,
+      // change this to bust the cache when making changes that would yield
+      // different results for the same hash
+      bust: 1,
     },
     context,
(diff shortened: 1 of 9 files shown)


packages/playwright/src/plugins/plugin.ts
```diff
@@ -123,6 +123,9 @@ async function createNodesInternal(
     {
       ...normalizedOptions,
       CI: process.env.CI,
+      // change this to bust the cache when making changes that would yield
+      // different results for the same hash
+      bust: 1,
```
FrozenPandaz: You don't need the bust as long as the lockfile is hashed which it is below.

FrozenPandaz (approved):

PR #32768, opened 2025-09-17
docs(nx-dev): initial round of search improvements

barbados-clemens:
update pagefind config to improve search hits
add weight and filter properties for docs to control individual pages

filter` allows selecting "type" of docs being searched
wm_2025-09-18T16-48-23@2x

where weight can set the given pages search weight
unfortunately this will set the weight of the whole page content instead which can make performance worse so important to use it sparingly. current set values are trial and error based. and subject to change over time really just depends on feedback over time of the search performance. 

fixes DOC-159


astro-docs/src/content/docs/concepts/common-tasks.mdoc
```diff
@@ -1,9 +1,12 @@
...
-title: Common Tasks
+title: Common Task Configurations
 description: Learn about standard task naming conventions in Nx projects, including build, serve, test, and lint tasks, for consistent project configuration.
 keywords: [build, serve, test, lint]
 sidebar:
   order: 1
+  label: Common Tasks
+filter: "type:Concepts"
+weight: 5.0
 ---
```
barbados-clemens, line 10: updated title here to try to add some more weight in relevancy but any more weight and it quickly overpowers the search results for unrelated things this one we'll need to play with more I think in increasing the content to help saturate

astro-docs/src/content/docs/reference/Deprecated/affected-graph.mdoc
```diff
@@ -3,6 +3,8 @@ title: 'affected:graph - CLI command'
 description: 'Graph dependencies affected by changes'
 sidebar:
   hidden: true
+weight: .1
+filter: "type:References"
```
barbados-clemens, line 7: basically made the deprecated items never show up unless you hit the exact title phrase as they were over taking the results. which I think is okay as no reason to really return deprecated commands often.

astro-docs/src/content/docs/reference/Powerpack/owners/Generators.mdoc
```diff
@@ -2,6 +2,8 @@
 title: '@nx/owners - Generators'
 sidebar:
   label: Generators
+weight: .5
+filter: 'type:References'
```
barbados-clemens, line 6: short docs with exact matching titles soar to the top of the search results. so going for `init` results in this page being at the top. since the `init` generators are called for you automatically with `nx add` not super helpful doc page anyways. so toned it down.

jaysoo (approved):

PR #32785, opened 2025-09-19
chore(repo): add missing tsx

meeroslav:
## Current Behavior
The `tsx` package is missing causing TUI to hang in interactive mode.

## Expected Behavior
The `tsx` package is installed so commands running `npx tsx` don't need to ask for a permission to install it.

## Related Issue(s)

Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -315,6 +315,7 @@
     "ts-node": "10.9.1",
     "tsconfig-paths": "^4.1.2",
     "tsconfig-paths-webpack-plugin": "4.2.0",
+    "tsx": "^4.20.5",
     "typedoc": "0.25.12",
     "typedoc-plugin-markdown": "3.17.1",
     "typescript": "~5.9.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -315,6 +315,7 @@
     "ts-node": "10.9.1",
     "tsconfig-paths": "^4.1.2",
     "tsconfig-paths-webpack-plugin": "4.2.0",
+    "tsx": "^4.20.5",
```
barbados-clemens, line 318: We should probs just use ts-node since it’s already installed. I just out of habit use tsx

meeroslav: We also use it on two CI pipelines. I think it makes sense as it's cleaner than `ts-node`

barbados-clemens (approved):
my approval has no power here, but wanted to show my support 🫡

FrozenPandaz (approved):

PR #32786, opened 2025-09-19
feat(react): remove support for non-isolated webpack config in `nxComponentTestingPreset`

jaysoo:
This PR removes the unnecessary support for non-isolated webpack config for React component testing via Cypress. We had deprecated `@nx/webpack:webpack` usage without explicit webpack config a long time ago, and that deprecated feature was already removed.

The logic used `nxComponentTestingPreset` is not necessary for latest Nx versions and should be removed in v22.

### DIFF
--- a/packages/react/plugins/component-testing/index.ts
+++ b/packages/react/plugins/component-testing/index.ts
@@ -116,7 +116,7 @@ export function nxComponentTestingPreset(
     };
   }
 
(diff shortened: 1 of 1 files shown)


jaysoo:
Eh, I had pushed some commits from master and resulted in tagging too many people for reviews. Let me close and re-open.

PR #32793, opened 2025-09-19
fix(core): update tui title text structure

JamesHenry:
## Current Behavior

```
Running test for 81 projects, and 121 requisite tasks
```

## Expected Behavior

```
Running 81 test tasks, and 121 others they depend on
```

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/run-command.ts
+++ b/packages/nx/src/tasks-runner/run-command.ts
@@ -145,19 +145,24 @@ async function getTerminalOutputLifeCycle(
       if (mainContinuousDependencies.length > 0) {
         pinnedTasks.push(mainContinuousDependencies[0]);
       }
-      const [project, target] = mainTaskId.split(':');
-      titleText = `${target} ${project}`;
+      const [, target] = mainTaskId.split(':');
+      titleText = `Running 1 ${target} task`;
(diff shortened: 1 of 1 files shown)


packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -145,19 +145,24 @@ async function getTerminalOutputLifeCycle(
       if (mainContinuousDependencies.length > 0) {
         pinnedTasks.push(mainContinuousDependencies[0]);
       }
-      const [project, target] = mainTaskId.split(':');
-      titleText = `${target} ${project}`;
+      const [, target] = mainTaskId.split(':');
```
meeroslav, line 148: This would not work for projects with colons in their name like `:common:utils`

meeroslav (dismissed):
Let's use splitTarget function here instead of explicit splitting by ":". This would enable my change on how we handle targets and projects with ":" in name to be automatically applied once it's merged.

packages/nx/src/tasks-runner/run-command.ts
FrozenPandaz: Can you send a separate PR for this please?

FrozenPandaz (approved):

PR #32888, opened 2025-09-30
feat(release)!: remove deprecated legacy versioning

Coly010:
Remove deprecated legacy versioning support

BREAKING CHANGE


graph/client/src/app/console-project-details/project-details.app.tsx
```diff
@@ -87,7 +87,7 @@ export function ProjectDetailsApp({
       <>
         <ExpandedTargetsProvider>
           <ProjectDetails
-            project={project}
+            project={project as any}
```
FrozenPandaz, line 90: Why does this need `as any` now? 👀

packages/nx/src/command-line/release/config/config.ts
```diff
@@ -352,35 +348,21 @@ export async function createNxReleaseConfig(
...
-                ?.updateDependents ?? 'auto',
-          }),
+
+      versionActions: DEFAULT_VERSION_ACTIONS_PATH,
+      versionActionsOptions: {},
+      currentVersionResolver: defaultGeneratorOptions.currentVersionResolver,
+      specifierSource: defaultGeneratorOptions.specifierSource,
+      preserveLocalDependencyProtocols:
+        userConfig.version?.preserveLocalDependencyProtocols ?? true,
+      // TODO(v22): flip default to true
```
FrozenPandaz, line 361: This will be a separate PR?

FrozenPandaz (approved):

packages/nx/src/command-line/release/config/config.ts
Coly010: yes

packages/next/package.json
```diff
@@ -52,7 +52,9 @@
     "@phenomnomnominal/tsquery": "~5.0.1"
   },
   "devDependencies": {
-    "nx": "workspace:*"
+    "nx": "workspace:*",
+    "@nx/cypress": "workspace:*",
```
JamesHenry, line 56: The scoped packages should really come first

graph/client/src/app/console-project-details/project-details.app.tsx
JamesHenry: What was the answer? :)

packages/js/src/generators/library/utils/add-release-config.ts
```diff
@@ -132,30 +131,16 @@ export async function addReleaseConfigForNonTsSolution(
...
-      version: {
-        manifestRootsToUpdate: [packageRoot],
-        // using git tags to determine the current version is required here because
-        // the version in the package root is overridden with every build
-        currentVersionResolver: 'git-tag',
-        fallbackCurrentVersionResolver: 'disk',
-      },
-    };
-  }
+  // TODO: re-evaluate this config in new versions
```
JamesHenry, line 134: I think we can strip this TODO, it's lost meaning now that a lot of time has passed and this code path (non TS solution) is already kind of legacy

packages/nx/schemas/nx-schema.json
JamesHenry, line 884: The comment was removed here but the change wasn't addressed - this is currently an under typed object with `"additionalProperties": true` which should not be the case. It was previously impossible with the complex dynamic schema, it should now be possible for this to be exhaustive, we need to revisit this

packages/nx/schemas/nx-schema.json
```diff
@@ -1107,7 +1062,6 @@
     },
     "NxReleaseGroupDockerConfiguration": {
```
JamesHenry, line 1063: Same thing

packages/nx/src/command-line/release/config/config.ts
```diff
@@ -1330,14 +1228,8 @@ function hasInvalidConventionalCommitsConfig(
       if (
         group.version?.conventionalCommits === true &&
         // v2 config - directly on version config
```
JamesHenry, line 1230: I think we can strip these comments now that there is no alternative

packages/nx/src/command-line/release/config/filter-release-groups.spec.ts
```diff
@@ -193,9 +175,22 @@ describe('filterReleaseGroups()', () => {
             "resolvedVersionPlans": false,
             "version": {
               "conventionalCommits": false,
-              "generator": "",
-              "generatorOptions": {},
+              "currentVersionResolver": undefined,
```
JamesHenry, line 178: These values are not correct, the should not be `undefined`, they should be their default values to match how all other config construction has been set up to work

packages/nx/src/command-line/release/config/filter-release-groups.spec.ts
```diff
@@ -473,12 +526,7 @@ describe('filterReleaseGroups()', () => {
           projects: ['lib-a'],
           changelog: false,
           docker: undefined,
-          version: {
-            conventionalCommits: false,
-            generator: '',
-            generatorOptions: {},
-            groupPreVersionCommand: '',
-          },
+          version: createVersionConfig(),
```
JamesHenry, line 529: I'm not a fan of this abstraction, every test previously was standalone and self-documenting, I don't think this is adding any value but has that cost

packages/nx/src/command-line/release/version/test-utils.ts
```diff
@@ -41,7 +41,6 @@ export async function createNxReleaseConfigAndPopulateWorkspace(
       groups,
       version: {
```
JamesHenry, line 42: This extra assignment is now unnecessary as no modification is happening, it's covered already by `...additionalNxReleaseConfig` above

packages/react/package.json
```diff
@@ -50,7 +50,9 @@
     "semver": "^7.6.3"
   },
   "devDependencies": {
-    "nx": "workspace:*"
+    "nx": "workspace:*",
+    "@nx/cypress": "workspace:*",
```
JamesHenry, line 54: Scoped packages first

packages/nx/src/command-line/release/config/filter-release-groups.spec.ts
JamesHenry: Wait sorry I see this is actually because of this `createVersionConfig` abstraction, rather than the config logic

PR #32903, opened 2025-09-30
fix(core): improve provenance error with custom registry information

MaxKless:
## Current Behavior
The provenance error pops up frequently for custom registries and it can be confusing. 

## Expected Behavior
With an updated error message, it's clearer what's going on and how to fix it. This is what it looks like for a custom registry: 
``` 
NX   The migrate command failed.


NX   An error occurred while checking the provenance of nx@21.10.5-provenance.
This might be due to a custom registry configuration (http://localhost:4874/). Please check whether provenance is correctly configured for your registry. 
To disable this check at your own risk, you can set the NX_SKIP_PROVENANCE_CHECK environment variable to true.
 ```

### DIFF
--- a/packages/nx/src/utils/provenance.ts
+++ b/packages/nx/src/utils/provenance.ts
@@ -1,10 +1,11 @@
-import { exec } from 'child_process';
+import { execSync } from 'child_process';
 import { join } from 'path';
 import { promisify } from 'util';
 import { readJsonFile } from './fileutils';
 import {
   detectPackageManager,
   getPackageManagerCommand,
+  packageRegistryView,
 } from './package-manager';
 
 /*
@@ -24,20 +25,18 @@ export async function ensurePackageHasProvenance(
     return;
   }
 
-  const execAsync = promisify(exec);
   try {
-    const result = await execAsync(
-      `npm view ${packageName}@${packageVersion} --json --silent`,
-      {
-        timeout: 20000,
-      }
+    const result = await packageRegistryView(
+      packageName,
+      packageVersion,
+      '--json --silent'
     );
-    const npmViewResult = JSON.parse(result.stdout.trim());
+    const npmViewResult = JSON.parse(result);
 
     const attURL = npmViewResult.dist?.attestations?.url;
 
     if (!attURL)
-      throw await createProvenanceError(
+      throw new ProvenanceError(
         packageName,
         packageVersion,
         'No attestation URL found'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/nx/src/utils/provenance.ts [hunk 1]
```diff
@@ -118,18 +121,69 @@ export async function ensurePackageHasProvenance(
...
 
+async function createProvenanceError(
+  packageName: string,
+  packageVersion: string,
+  error?: string
+): Promise<ProvenanceError> {
+  let customRegistry: string | undefined;
+
+  try {
+    const packageManager = detectPackageManager();
```
FrozenPandaz: Pass this in

packages/nx/src/utils/provenance.ts
```diff
@@ -118,18 +121,69 @@ export async function ensurePackageHasProvenance(
...
-  constructor(packageName: string, packageVersion: string, error?: string) {
+  constructor(
+    packageName: string,
+    packageVersion: string,
+    customRegistry?: string,
+    error?: string
+  ) {
+    const registryNote = customRegistry
+      ? `This might be due to a custom registry configuration (${customRegistry}). Please check whether provenance is correctly configured for your registry.`
+      : 'This could indicate a security risk. Please double check https://www.npmjs.com/package/${packageName} to see if the package is published correctly or file an issue at https://github.com/nrwl/nx/issues.';
```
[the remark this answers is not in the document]
FrozenPandaz: Graphite is absolutely right! 😉

packages/nx/src/utils/provenance.ts
```diff
@@ -118,18 +121,69 @@ export async function ensurePackageHasProvenance(
...
+): Promise<ProvenanceError> {
+  let customRegistry: string | undefined;
+
+  try {
+    const packageManager = detectPackageManager();
+    const commands = getPackageManagerCommand(packageManager);
+
+    // Try to get registry from current package manager, fall back to npm
+    const registryCommand =
+      commands.getRegistryUrl ?? 'npm config get registry';
```
FrozenPandaz: Shouldn't this be in the `commands` itself? 👀 

Like if `yarn` doesn't have a `getRegistryUrl`, it should use `npm`?

packages/nx/src/utils/provenance.ts
```diff
@@ -118,18 +121,69 @@ export async function ensurePackageHasProvenance(
...
-    throw new ProvenanceError(
+    throw await createProvenanceError(
       packageName,
       packageVersion,
       error.message || error
     );
   }
 }
 
+async function createProvenanceError(
```
FrozenPandaz: This should be possible to incorporate into the `ProvenanceError constructor` right?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #32906, opened 2025-09-30
docs(nx-dev): integrate Calendly forms with Hubspot forms

bcabanes:
Added support for Calendly form integration within Hubspot forms across multiple components.

### DIFF
--- a/nx-dev/nx-dev/app/global-scripts.tsx
+++ b/nx-dev/nx-dev/app/global-scripts.tsx
@@ -71,6 +71,22 @@ export default function GlobalScripts({ gaMeasurementId, gtmMeasurementId }) {
       hsFormsScript.async = true;
       hsFormsScript.defer = true;
       document.head.appendChild(hsFormsScript);
+
+      // Load Calendly Forms
+      const calendlyFormsScript = document.createElement('script');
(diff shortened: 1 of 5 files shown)


nx-dev/nx-dev/app/global-scripts.tsx
```diff
@@ -71,6 +71,22 @@ export default function GlobalScripts({ gaMeasurementId, gtmMeasurementId }) {
       hsFormsScript.async = true;
       hsFormsScript.defer = true;
       document.head.appendChild(hsFormsScript);
+
+      // Load Calendly Forms
+      const calendlyFormsScript = document.createElement('script');
+      calendlyFormsScript.src =
+        '//assets.calendly.com/assets/external/forms.js';
```
barbados-clemens, line 78: is this supposed to have two slashes?

vsavkin (approved):

nx-dev/nx-dev/app/global-scripts.tsx
bcabanes: Yep, it means "load with the current protocol".
`http` or `https` whatever you are on (local VS live).

PR #32908, opened 2025-09-30
docs(nx-dev): require description for doc pages

barbados-clemens:
add description to all docs and make required to make index views not render differently

image

closes: DOC-246


juristr (approved):
Awesome, thank you!

PR #32946, opened 2025-10-03
feat(core): collect resource usage

leosvelperez:
Adds resource usage collection while running tasks.


packages/nx/src/native/metrics/collector.rs
```diff
@@ -0,0 +1,663 @@
...
+        sys.refresh_processes_specifics(
+            sysinfo::ProcessesToUpdate::Some(&[target_pid]),
+            false, // don't remove dead processes
+            ProcessRefreshKind::nothing().with_cpu(),
+        );
+    }
+
+    /// The main collection loop that runs in a separate thread
+    /// Discovers current metrics and directly notifies subscribers (true push-based)
+    fn collection_loop(
```
FrozenPandaz, line 309: Check if `&mut self` can be used here

packages/nx/src/native/metrics/collector.rs
```diff
@@ -0,0 +1,663 @@
...
+                                    metadata,
+                                };
+
+                                // Clone ThreadsafeFunction (cheap - internally Arc-based)
+                                (state.callback.clone(), update)
+                            })
+                            .collect::<Vec<_>>()
+                    }; // Lock released here!
+
+                    // Notify subscribers WITHOUT holding lock (fixes RUST-8)
```
FrozenPandaz: clean up

packages/nx/src/native/metrics/collector.rs
```diff
@@ -0,0 +1,663 @@
...
+        system: &Mutex<System>,
+        individual_tasks: &DashMap<String, IndividualTaskRegistration>,
+        batches: &DashMap<String, BatchRegistration>,
+        main_cli_pid: &Mutex<Option<i32>>,
+        daemon_pid: &Mutex<Option<i32>>,
+        metadata_store: &MetadataStore,
+    ) -> Result<MetricsCollectionResult, Box<dyn std::error::Error>> {
+        // Capture timestamp FIRST for accuracy
+        let timestamp = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
```
FrozenPandaz: This seems weird. There should be other methods to do this easily.

packages/nx/src/native/metrics/collector.rs
```diff
@@ -0,0 +1,663 @@
...
+            batches.len() * 5, // batch tasks × avg processes per batch
+        );
+
+        // Track new metadata discovered during collection (conditionally populated)
+        let mut new_metadata = HashMap::new();
+
+        // Collect main CLI process
+        let main_cli_metrics = {
+            let cli_pid = main_cli_pid.lock();
+            if let Some(pid) = *cli_pid {
```
FrozenPandaz, line 576: If `NX_ISOLATE_PLUGINS=true NX_DAEMON=false` then the main process will have the plugin workers as child processes.

packages/nx/src/native/metrics/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod types;
+
+#[cfg(not(target_arch = "wasm32"))]
+#[cfg(not(target_os = "wasi"))]
+pub mod collector;
```
FrozenPandaz, line 1: pnpm build:wasm will either fail or suceed and you can try.

packages/nx/src/native/metrics/napi_bindings.rs
```diff
@@ -0,0 +1,132 @@
...
+/// NAPI wrapper for the process metrics collector
+/// Provides a JavaScript-friendly interface for metrics collection
+#[napi]
+pub struct ProcessMetricsCollector {
+    /// The actual metrics collector instance
+    collector: Arc<RwLock<MetricsCollector>>,
+}
+
+#[napi]
+impl ProcessMetricsCollector {
```
FrozenPandaz: You shouldn't need a separate napi binding struct

packages/nx/src/native/metrics/types.rs
```diff
@@ -0,0 +1,178 @@
...
+            task_ids: Arc::new(task_ids),
+            anchor_pid,
+        }
+    }
+}
+
+/// Process snapshot with full metadata and metrics
+#[napi(object)]
+#[derive(Debug, Clone)]
+pub struct ProcessSnapshot {
```
FrozenPandaz: This does not match typescript?

packages/nx/src/native/pseudo_terminal/child_process.rs
```diff
@@ -52,6 +52,11 @@ impl ChildProcess {
         External::new((self.parser.clone(), self.writer_arc.clone()))
     }
 
+    #[napi]
+    pub fn get_pid(&self) -> i32 {
+        self.process_killer.get_pid()
```
FrozenPandaz: This is weird to get the pid from the killer. Whatever instantiates this should also pass in a PID

packages/nx/src/native/metrics/napi_bindings.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Create a new metrics collector with default configuration
+    #[napi(constructor)]
+    pub fn new() -> anyhow::Result<Self> {
```
FrozenPandaz: This can return `Self`

packages/nx/src/native/metrics/collector.rs
leosvelperez: We can't use `&mut self` here because the function runs in a separate thread that needs to own its data. We can't pass references across thread boundaries due to Rust's ownership rules.

packages/nx/src/native/metrics/collector.rs
```diff
@@ -0,0 +1,751 @@
...
+        batches.retain(|_, batch_reg| {
+            sys.process(Pid::from(batch_reg.anchor_pid as usize))
+                .is_some()
+        });
+
+        // Clean up dead CLI subprocesses
+        main_cli_subprocess_pids.retain(|pid, _| sys.process(Pid::from(*pid as usize)).is_some());
+
+        // Pre-allocate HashSet for tracking all live PIDs to reduce allocations
+        let mut live_pids = HashSet::with_capacity(
```
FrozenPandaz: Can we create a set once and then maintain it for each loop?

packages/nx/src/native/utils/time.rs
```diff
@@ -0,0 +1,8 @@
+use std::time::{SystemTime, SystemTimeError, UNIX_EPOCH};
+
+/// Returns the current time in milliseconds since Unix epoch
+pub fn current_timestamp_millis() -> Result<i64, SystemTimeError> {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
```
FrozenPandaz, line 6: Add an `unwrap_or(0)` here so that this can return just an `i64` and we don't have to worry about `unwrap` or whatever when we consume this.

FrozenPandaz (approved):

PR #32953, opened 2025-10-03
docs(linter): show flat config in docs

jaysoo:
We show legacy eslintrc format that's been deprecated and shouldn't be used anymore. We still want to document the legacy format in case users haven't switched yet.

### DIFF
--- a/astro-docs/src/content/docs/concepts/Decisions/project-dependency-rules.mdoc
+++ b/astro-docs/src/content/docs/concepts/Decisions/project-dependency-rules.mdoc
@@ -114,7 +114,7 @@ export { formatCurrency } from './src/format-currency';
 
 In order to enforce the dependency constraints that were listed for each type, you can add the following rule in your ESLint configuration:
 
-{% tabs %}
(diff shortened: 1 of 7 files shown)


astro-docs/src/content/docs/concepts/Decisions/project-dependency-rules.mdoc
```diff
@@ -112,9 +112,59 @@ export { formatCurrency } from './src/format-currency';
 
 ## Enforce Project Dependency Rules
 
-In order to enforce the dependency constraints that were listed for each type, you can add the following rule in the root `.eslintrc.json` file:
+In order to enforce the dependency constraints that were listed for each type, you can add the following rule in your ESLint configuration:
+
+{% tabs %}
```
barbados-clemens: small nit non-blocking: we can add a `syncKey` so people can keep the tabs in sync across pages for their current setup. 
```suggestion
{% tabs syncKey="eslint-config-preference" %}
```

barbados-clemens (approved):
1 comment, non blocking.

PR #32963, opened 2025-10-06
fix(remix): package json has mispelling in exports section

ts-ign0re:
This pull request fixes a misspelling in packages/remix/package.json file

There is "defautl" prop inside "exports" sections instead of "default"

Thanks


spolesciuc (approved):

AZ604 (approved):

ts-ign0re:
okay guys I think I should use turborepo instead

Coly010:
This seems to have been closed accidentally, or incorrectly, and I have no idea how.
And GitHub will not let me reopen 🤯 

EDIT: after futher investigation, this commit has already been applied, and the issue was fixed

PR #32978, opened 2025-10-06
feat(core): add support for pnpm catalogs

### LINKED ISSUE #29772: Generators don't work when using pnpm catalogs
### Current Behavior

Running the example code:
`nx g plugin my-plugin --directory=libs --importPath=@squareup/my-plugin`

yields:
`Error: The package.json lists a version of typescript that Nx is unable to validate - (catalog:)`

This is because the generators use a check of the TS version which simply gets the entry from package.json and does not resolve catalogs from pnpm:
https://github.com/nrwl/nx/blob/e0c49d3627db8a02914ca90a764765b9778bd561/packages/js/src/generators/init/init.ts#L34

### Expected Behavior

A helper should be employed in the various places NX tries to resolve a version string from the package.json to resolve the "real" version using pnpm for checks like this to enable fuller support for pnpm as the package manager.

### GitHub Repo

_No response_

### Steps to Reproduce

`package.json`:
```
"devDependencies": {
    "typescript": "catalog:",
}
```

`pnpm-workspace.yaml`:
```
catalog:
  typescript: ~5.5.2
```

### Nx Report

```shell

Node           : 22.9.0
OS             : darwin-arm64
pnpm           : 9.11.0

nx                     : 20.3.2
@nx/js                 : 20.3.2
@nx/jest               : 20.3.2
@nx/eslint             : 20.3.2
@nx/workspace          : 20.3.2
@nx/cypress            : 20.3.2
@nx/devkit             : 20.3.2
@nx/eslint-plugin      : 20.3.2
@nx/module-federation  : 20.3.2
@nx/plugin             : 20.3.2
@nx/react              : 20.3.2
@nx/storybook          : 20.3.2
@nx/vite               : 20.3.2
@nx/web                : 20.3.2
typescript             : 5.5.4
Registered Plugins:
@nx/vite/plugin
@nx/storybook/plugin
```

### Failure Logs

```shell

```

### Package Manager Version

9.11.0

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

### LINKED ISSUE #30035: PNPM Catalogs are not supported

leosvelperez:
## Current Behavior

Pnpm Catalogs is not supported.

## Expected Behavior

Pnpm Catalogs should be supported.

## Related Issue(s)

Fixes #30035 
Fixes #29772


e2e/nx/src/misc.test.ts
```diff
@@ -447,7 +448,7 @@ describe('Nx Commands', () => {
 // TODO(colum): Change the fetcher to allow incremental migrations over multiple versions, allowing for beforeAll
 describe('migrate', () => {
   beforeEach(() => {
-    newProject({ packages: [] });
```
FrozenPandaz, line 450: Why did this get removed?

e2e/nx/src/misc.test.ts
```diff
@@ -756,7 +766,7 @@ describe('migrate', () => {
     expect(output).toContain(`Migrations file 'migrations.json' doesn't exist`);
   });
 
-  it('should handle Nx tokens correctly in Angular CLI migration schematics', () => {
+  it('hhhhhhshould handle Nx tokens correctly in Angular CLI migration schematics', () => {
```
FrozenPandaz: clean up

e2e/nx/src/misc.test.ts
```diff
@@ -903,6 +913,171 @@ describe('migrate', () => {
       ],
     });
   });
+
+  if (getSelectedPackageManager() === 'pnpm') {
+    it('hhhhhhshould handle pnpm catalog references and update catalog definitions during migration', () => {
```
FrozenPandaz: clean up

packages/angular/src/generators/init/init.ts
```diff
@@ -57,7 +57,7 @@ function installAngularDevkitCoreIfMissing(
   tree: Tree,
   options: Schema
 ): GeneratorCallback {
-  const packageVersion = getInstalledPackageVersion(
```
FrozenPandaz, line 60: This one reads from the installed one in `node_modules`. I think that's different from the new util right?

packages/angular/src/generators/ngrx-root-store/lib/normalize-options.ts
```diff
@@ -23,11 +23,12 @@ export function normalizeOptions(
   let rxjsVersion: string;
   try {
     rxjsVersion = checkAndCleanWithSemver(
+      tree,
       'rxjs',
       readJson(tree, 'package.json').dependencies['rxjs']
```
FrozenPandaz: Is this correct? Don't we have to resolve this with the util?

packages/nx/src/utils/catalog/pnpm-manager.ts
```diff
@@ -0,0 +1,309 @@
...
+  parseCatalogReference(version: string): CatalogReference | null {
+    if (!this.isCatalogReference(version)) {
+      return null;
+    }
+
+    const catalogName = version.substring(this.catalogProtocol.length);
+
+    return {
+      catalogName: catalogName || undefined,
+      isDefaultCatalog: catalogName === '',
```
aidant: The shorthand `catalog:` is an alias for `catalog:default` and similarly in the `pnpm-workspace.yaml` file the `catalog` property is an alias for the named variant `catalogs.default`. You can see the normalisation logic in [`getCatalogsFromWorkspaceManifest.ts`](https://github.com/pnpm/pnpm/blob/1b15e45ae949485ba9b9bfd4e079b551ab7d8819/catalogs/config/src/getCatalogsFromWorkspaceManifest.ts#L20-L27) as well as a reference to this in the documentation:
> [These version ranges can be referenced through catalog:default. For the default catalog only, a special catalog: shorthand can also be used. Think of catalog: as a shorthand that expands to catalog:default.](https://pnpm.io/catalogs#:~:text=These%20version%20ranges%20can%20be%20referenced%20through%20catalog%3Adefault.%20For%20the%20default%20catalog%20only%2C%20a%20special%20catalog%3A%20shorthand%20can%20also%20be%20used.%20Think%20of%20catalog%3A%20as%20a%20shorthand%20that%20expands%20to%20catalog%3Adefault.)

I would recommend performing a similar normalisation logic to prevent issues arising where a named default catalog is used in a `pnpm-workspace.yaml` file but the shorthand `catalog:` is used in a `package.json`.

You should also keep note of the transformations applied to ensure consistency when writing back to the `pnpm-workspace.yaml` or a `package.json` to prevent the short and long variants from replacing each other.

packages/devkit/src/utils/catalog/pnpm-manager.ts
aidant, line 1: Is this directory (`packages/devkit/src/utils/catalog/`) intentionally duplicated from `packages/nx/src/utils/catalog/`?

packages/angular/src/generators/init/init.ts
leosvelperez: That one doesn't read from `node_modules`. The change is correct.

packages/angular/src/generators/ngrx-root-store/lib/normalize-options.ts
leosvelperez: This is safe. The `checkAndCleanWithSemver` helper handles Pnpm Catalogs.

packages/devkit/src/utils/catalog/pnpm-manager.ts
leosvelperez: It is, `@nx/devkit` depends on a range of `nx` versions, so it can't re-export the functionality until Nx v22 is the lowest version supported by `@nx/devkit`. It will eventually only exist in the `nx` package.

packages/nx/src/utils/catalog/pnpm-manager.ts
leosvelperez: TIL. Thanks for the info. I'll make the relevant changes.

packages/nx/src/utils/catalog/manager-factory.ts
```diff
@@ -0,0 +1,29 @@
...
+  switch (packageManager) {
+    case 'pnpm':
+      return new PnpmCatalogManager();
+    case 'npm':
+      return new NpmCatalogManager();
+    case 'yarn':
+      return new YarnCatalogManager();
+    case 'bun':
+      return new BunCatalogManager();
+    default:
```
FrozenPandaz: These should all return `null`

packages/nx/src/utils/catalog/manager-factory.ts
```diff
@@ -0,0 +1,29 @@
...
+      return new UnknownCatalogManager();
```
FrozenPandaz: There should not be an unknown catalog manager

packages/devkit/src/utils/semver.ts
```diff
@@ -1,22 +1,64 @@
...
+    const validation = tree
+      ? manager.validateCatalogReference(tree, pkgName, newVersion)
+      : manager.validateCatalogReference(root, pkgName, newVersion);
+    if (!validation.isValid) {
+      throw new Error(
+        `The catalog reference for ${pkgName} is invalid - (${newVersion})\n${formatCatalogError(
+          validation.error!
+        )}`
+      );
+    }
```
FrozenPandaz: Throw the error from validation

packages/devkit/src/utils/catalog/manager.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Interface for catalog managers that handle package manager-specific catalog implementations.
+ */
+export interface CatalogManager {
+  readonly name: string;
+  /**
+   * Check if this package manager supports catalogs.
+   */
+  supportsCatalogs(): boolean;
```
FrozenPandaz: This is unneeded

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #32979, opened 2025-10-06
docs(testing): add generic guide over feature based testing

barbados-clemens:
- add feature based testing guide
- add option to show all nodes w/ affected styles

### DIFF
--- a/astro-docs/src/content/docs/guides/Tips-n-Tricks/feature-based-testing.mdoc
+++ b/astro-docs/src/content/docs/guides/Tips-n-Tricks/feature-based-testing.mdoc
@@ -10,13 +10,14 @@ Feature-based testing is a strategy that co-locates tests with the features they
 
 ## The Problem: Monolithic Test Projects
 
-Many projects start with a single, large test project that depends on the entire application. While this setup ensures tests run when any dependency changes, it also means **all tests run even when only one feature changes**.
+Many projects have a single, large test project that depends on the entire application, such as e2e projects. While this setup ensures tests run when any dependency changes, it also means **all tests run even when only one subset of the app changes**.
 
 {% aside type="tip" title="Beyond E2E Tests" %}
-While this guide uses end-to-end (e2e) tests as examples, the same strategy applies to any type of testing—integration tests, component tests, or even large unit test suites. The principles of feature-based testing work wherever you have monolithic test projects that could benefit from being split by feature.
+While this guide uses end-to-end (e2e) tests as examples, the same strategy applies to any type of testing—integration tests, component tests, or even large unit test suites.
+The principles of feature-based testing work wherever you have monolithic test projects that could benefit from being split by feature.
 {% /aside %}
 
-Consider a typical setup where all e2e tests live in a single project:
+Consider a typical setup where all e2e tests live in a single project at the top of the graph:
 {% graph height="400px" %}
 
 ```json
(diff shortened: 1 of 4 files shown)


nx-dev/ui-markdoc/src/lib/graphs/project-graph.tsx
```diff
@@ -76,7 +81,11 @@ function NxDevProjectGraphInner({
       updater: (config) => ({
         mode: composite ? 'composite' : 'individual',
         autoExpand: composite ? config.autoExpand : 0,
-        showMode: affectedProjects.length ? 'affected' : 'all',
+        showMode: showAffectedWithNodes
```
barbados-clemens, line 84: allows showing all nodes. before we only showed all nodes if there were not affected nodes, but to illustate the concept of affected path in a graph, I needed a way to show both affected and non-affected nodes.
which required brining over some styles from the graph package that weren't exported.

astro-docs/src/content/docs/guides/Tips-n-Tricks/feature-based-testing.mdoc
```diff
@@ -0,0 +1,171 @@
+---
+title: Optimize Testing with Feature-Based Testing
+sidebar:
+  label: Feature-Based Testing
+description: Learn how to optimize your testing strategy by breaking down monolithic test projects into feature-scoped tests that run only when needed.
+filter: 'type:Guides'
```
barbados-clemens, line 6: Honestly this is more of a concept than a guide, but it doesn't really fit into a 'core concept' cleanly, where we have kind of this 'dumping ground' of tips-n-tricks to place it in.

astro-docs/src/content/docs/guides/Tips-n-Tricks/feature-based-testing.mdoc
```diff
@@ -0,0 +1,171 @@
...
+# Run all affected tests based on your changes
+nx affected -t test
+
+# Run affected e2e tests
+nx affected -t e2e
+```
+
+This ensures you're only testing what changed, whether locally or in CI.
+
+### Combining with Automated Task Splitting (Atomizer)
```
jaysoo: I wonder if we should move the Atomizer earlier in the guide since it is much easier to enable, and works with both `run-many` and `affected` to speed up tests in CI.

Even if both feature would help speed up CI, it's faster to enable per-file distribution first, and I think it's simpler conceptually.

astro-docs/src/content/docs/guides/Tips-n-Tricks/feature-based-testing.mdoc
jaysoo: Yeah, I think we can even move stuff into their technologies section and keep core slimmer. Like this one and some bundling concepts/guides can go into `Technologies > Testing Tools` or `Technologies > Build Tools`. We'll need to look through everything we have an how we want to reorganize.

astro-docs/src/content/docs/guides/Tips-n-Tricks/feature-based-testing.mdoc
barbados-clemens: Pulled it up to be the first item in the getting started and mention turning it on is typically the easier win initially.

jaysoo (approved):

PR #32998, opened 2025-10-07
fix(nx-dev): update docs code blocks usage

barbados-clemens:
by default code blocks will get text as the lang if not set or the lang is not supported
but we should be so I don't have to look at the warnings in the terminal output 😅 
also remove the frame=none since the team preferred to have the terminal frame

Example of invalid codeblock messages
wm_2025-10-15T11-26-05@2x

Also added conformance rule for validating image pages for public and src/assets/ directory

examples of incorrect absolute path ref from "public" folder and incorrect absolute path: 

wm_2025-10-15T10-23-41@2x

fixes DOC-242
fixes DOC-259


package.json
```diff
@@ -84,7 +84,7 @@
     "@nuxt/kit": "^3.10.0",
     "@nuxt/schema": "^3.10.0",
     "@nx/angular": "22.0.0-beta.4",
-    "@nx/conformance": "3.0.0",
+    "@nx/conformance": "4.0.0",
```
barbados-clemens, line 86: bump to conformance v4 so I can use the `tree` and `fileMapCache`

pnpm-lock.yaml
barbados-clemens, line 1: rebasing w/ main added extra esbuild related deps after installing. not sure if expected or not. but might be fine

jaysoo (approved):

FrozenPandaz (approved):

PR #33002, opened 2025-10-08
docs(misc): update description for @berenddeboer/nx-aws-cdk

berenddeboer:
Description update for @berenddeboer/nx-aws-cdk plugin: this has become self-inferring.


leosvelperez (changes requested):
Please update the corresponding entry in the `astro-docs/src/content/approved-community-plugins.json` file as well.

leosvelperez (approved):

PR #33013, opened 2025-10-08
feat(core): apply parent env to atomized target

meeroslav:
## Current Behavior
Env variables from the parent target are not propagated to the atomized target e.g. 
`.test-ci.env` will be only applied to **no-op** `test-ci` but not to `test-ci--path/to/test/file`.

## Expected Behavior
Running an atomized target will load from the parent's env files.

## Related Issue(s)

Fixes #


packages/nx/src/tasks-runner/task-env.ts
```diff
@@ -166,50 +166,51 @@ export function unloadDotEnvFile(
 }
 
 function getEnvFilesForTask(task: Task): string[] {
+  const target = task.target.target.split('--')[0];
```
MaxKless: I think this would be less brittle/hacky if we passed the project graph into here and used the atomization metadata to determine the parent task vs basing it purely on the task name

meeroslav: I agree. I wasn't sure about modifying the public caller with graph reference but i guess it's the only way to be sure.

packages/nx/src/tasks-runner/task-env.ts
```diff
@@ -165,51 +166,64 @@ export function unloadDotEnvFile(
   });
 }
 
-function getEnvFilesForTask(task: Task): string[] {
+function getOwnerTargetForTask(task: Task, graph: ProjectGraph): string {
+  const project = graph.nodes[task.target.project];
+  if (project.data.metadata?.targetGroups) {
+    for (const targets of Object.values(project.data.metadata.targetGroups)) {
+      if (targets.includes(task.target.target)) {
+        return targets[0];
```
FrozenPandaz: This is not correct.. can we fix this in the plugins themselves since `run-commands` has a `envFile` option?

FrozenPandaz (dismissed):

packages/nx/src/tasks-runner/task-env-paths.ts
```diff
@@ -0,0 +1,49 @@
+export function getEnvPathsForTask(
+  projectRoot: string,
+  target: string,
+  configuration?: string,
+  nonAtomizedTarget?: string
+): string[] {
+  const indentifiers = [];
```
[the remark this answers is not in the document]
meeroslav: This broke my code, since that variable was of course used further in the function.

AgentEnder (changes requested):
The docs still need to be updated to reflect this, we should do it in same PR

meeroslav:
@AgentEnder @FrozenPandaz can you check this PR?

AgentEnder (approved):

barbados-clemens (approved):

PR #33015, opened 2025-10-08
docs(misc): update Tailwind CSS guides for React and Angular to not mention the generator

jaysoo:
The generator `setup-tailwind` is outdated and not necessary. You can easily set up Tailwind yourself in 1 minute. Keep the pages around because people do search for tailwind in our docs.


barbados-clemens (approved):
slight nit, non-blocking and not with PR changes but the pages. 

react is titled 
"Using Tailwind CSS in React"

where angular is 
"Using Tailwind CSS with Angular Projects"

we should probs standardize on 1 or the other. 

we can leave the filenames the same just update the title frontmatter.

PR #33035, opened 2025-10-09
docs(misc): allow embeddings to generate from new docs

jaysoo:
This PR adds support for generating embeddings using docs in the `astro-docs` folder. The embeddings are used for `docs_search` MCP tool, and we currently do not populate new content into it.


.github/workflows/generate-embeddings.yml
```diff
@@ -4,6 +4,15 @@ on:
   schedule:
     - cron: "0 5 * * 0,4" # sunday, thursday 5AM
   workflow_dispatch:
+    inputs:
```
jaysoo: This is needed now to test, we could remove it once this pipeline is verified.

pnpm-workspace.yaml
```diff
@@ -3,6 +3,7 @@ packages:
   - 'e2e/*'
   - 'graph/*'
   - 'nx-dev/*'
+  - 'tools/documentation/*'
```
jaysoo: We missed this earlier, so the deps weren't installed--we had dups in the root `package.json`

tools/documentation/create-embeddings/src/main.mts
```diff
@@ -383,13 +486,68 @@ async function generateEmbeddings() {
...
 function delay(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
+/**
+ * Recursively find all index.html files in Astro dist directory
+ */
+async function getAstroPaths(): Promise<WalkEntry[]> {
+  console.log('Using Astro mode - reading from astro-docs/dist');
+  const repoRoot = join(import.meta.dirname, '../../../../');
```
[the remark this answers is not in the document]
barbados-clemens: can't we just import workspaceRoot from devkit? at least that's what I always do to get repo root

barbados-clemens (approved):

jabari-max (approved):

PR #33043, opened 2025-10-10
docs(nx-cloud): update release notes

rarmatei:
## Related Issue(s)

Fixes #

### DIFF
--- a/astro-docs/src/content/docs/reference/Nx Cloud/release-notes.mdoc
+++ b/astro-docs/src/content/docs/reference/Nx Cloud/release-notes.mdoc
@@ -37,7 +37,7 @@ filter: 'type:References'
 ##### Onboarding
 
 - Feat: Revamped end-to-end UI-only onboarding flow for GitHub workspaces
-    - For this to work, please enable the "Repository Administration Write" permission in your GitHub, to allow NxCloud to create new repositories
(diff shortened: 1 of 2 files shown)


astro-docs/src/content/docs/reference/Nx Cloud/release-notes.mdoc
```diff
@@ -4,6 +4,60 @@ description: 'Release notes for Nx Cloud Enterprise'
...
+
+- Feat: Revamped end-to-end UI-only onboarding flow for GitHub workspaces
+
+##### Enterprise Features
+
+- Feat: SCIM support for Okta, PingIdentity and Azure AD
+
+##### Conformance
+
+- Feat: Richer conformance violation notification emails
```
barbados-clemens: On prem doesn’t have email client configs? So I don’t think this will work for them

rarmatei: good spot 👍

docs/nx-cloud/reference/release-notes.md
barbados-clemens, line 1: You don't have to update this file BTW, this is no longer used and we're aiming to remove them after a bit, but if you want to keep them in sync you can.

barbados-clemens (approved):

PR #33042, opened 2025-10-10
docs(nx-dev): self-healing ci update blog post

juristr:
## Related Issue(s)

Fixes #

PR #33036, opened 2025-10-09
docs(nx-dev): add 21.6 changelog and october highlights article

philipjfulcher:
https://nx-dev-git-philip-october-update-nrwl.vercel.app/changelog#nx-216
https://nx-dev-git-philip-october-update-nrwl.vercel.app/blog/nx-highlights-oct-2025

PR #33045, opened 2025-10-10
fix(nx-dev): update nextjs internal-link-checker to include astro urls

barbados-clemens:
blog and marketing pages will still link to astro docs, so depend on the
astro sitemap to check all links available to be used.
NOTE: the specific header links (url fragments) are skipped for pages
that come from astro docs (prefixed with /docs) due to not having a
simple way to create these lists without parsing all the build html
files out and that's not the part we really care about

confirmed working https://github.com/nrwl/nx/pull/33042 and https://github.com/nrwl/nx/pull/33036 which are blocked until this change merges

Fixes DOC-264

### DIFF
--- a/scripts/documentation/internal-link-checker.ts
+++ b/scripts/documentation/internal-link-checker.ts
@@ -88,10 +88,8 @@ function readSiteMapIndex(directoryPath: string, filename: string): string[] {
     };
   } = parser.parse(readFileContents(join(directoryPath, filename)));
 
-  const internalSitemap = sitemapIndex.sitemapindex.sitemap.find(
-    // astro sitemap adds a new item into the sitemap entries with <domain>/docs/sitemap-index.xml.
-    // we already validate the sitemap links insite astro. no need to do it in nextjs
-    (s) => !s.loc.endsWith('/docs/sitemap-index.xml')
+  const internalSitemap = sitemapIndex.sitemapindex.sitemap.find((s) =>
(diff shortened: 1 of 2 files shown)


bcabanes (approved):

scripts/documentation/internal-link-checker.ts
```diff
@@ -88,10 +88,8 @@ function readSiteMapIndex(directoryPath: string, filename: string): string[] {
     };
   } = parser.parse(readFileContents(join(directoryPath, filename)));
 
-  const internalSitemap = sitemapIndex.sitemapindex.sitemap.find(
-    // astro sitemap adds a new item into the sitemap entries with <domain>/docs/sitemap-index.xml.
-    // we already validate the sitemap links insite astro. no need to do it in nextjs
-    (s) => !s.loc.endsWith('/docs/sitemap-index.xml')
+  const internalSitemap = sitemapIndex.sitemapindex.sitemap.find((s) =>
+    s.loc.endsWith('sitemap-0.xml')
```
bcabanes, line 92: You might have multiple sitemaps, only targeting the `0` index might not work in the long term.

jaysoo (approved):

scripts/documentation/internal-link-checker.ts
barbados-clemens: That’ll be a problem for future me if we hit 7000 pages in the docs

PR #33056, opened 2025-10-13
feat(module-federation): support TS Soln and Package Manager workspaces paradigms

Coly010:
## Current Behavior
Using TS Soln Workspaces and/or Packaage Manager Workspaces, handling of certain workflows and scenarios is not correct.

Detecting and Sharing Workspace Libraries relies entirely on TS Path Aliases existing in the base TSConfig file.
All guidance also points towards adding Workspace Libraries as dependencies or devDependencies within the consuming application's package.json file.
This also does not allow correct configuration of sharing.

Meanwhile, TS Path Aliases are added for remote applications such that TS can find them in consuming applications, while also being able to provide Typing Support. 
However, this has an increased build-time cost for TS compilation as it will follow the path in source.  

## Expected Behavior
Allow attaching Workspace Libraries as deps in the package.json of host and remote applications.
Configure packages added in such a manner correctly for share scope in Module Federation.

Attach Remote applications to Host applications via devDependencies in package.json.
Configure the `exports` and `main, types` properties in the remote application's package.json to point to the `src/remote-entry.ts` file such that node resolution can correctly follow the paths.
Bundler will continue to strip this out of compilation and replace with Module Federation Module Loading code.

### DIFF
--- a/packages/module-federation/src/utils/share.ts
+++ b/packages/module-federation/src/utils/share.ts
@@ -40,9 +40,6 @@ export function shareWorkspaceLibraries(
   }
 
   const tsconfigPathAliases = readTsPathMappings(tsConfigPath);
-  if (!Object.keys(tsconfigPathAliases).length) {
-    return getEmptySharedLibrariesConfig();
-  }
 
   // Nested projects must come first, sort them as such
   const sortedTsConfigPathAliases = {};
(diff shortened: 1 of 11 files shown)


packages/module-federation/src/utils/share.ts
```diff
@@ -143,6 +151,53 @@ export function shareWorkspaceLibraries(
...
+
+      // Add workspace libs from package.json dependencies
+      // This supports TS Solution + PM Workspaces
+      for (const libraryName of workspaceLibrariesAsDeps) {
+        let version =
+          pkgJson?.dependencies?.[libraryName] ??
+          pkgJson?.devDependencies?.[libraryName];
+
+        // Normalize workspace protocol versions (workspace:*, workspace:^, *, etc.)
+        if (version && (version === '*' || version.startsWith('workspace:'))) {
```
jaysoo: Also `file:` I think?

jaysoo (approved):

Issue #27842, opened 2024-09-09
[Angular] [Module Federation] nx-angular-mf duplicated in vendor.js and library module


pcurrivan:
### Current Behavior

After upgrading an Nx MF Angular app to Nx 19.6.6, Angular 18, etc., I see this error when calling `loadRemoteModule`:

`TypeError: Cannot read properties of undefined (reading 'endsWith')`

Looking in the dev tools, it looks like the `nx-module-mf.mjs` module is included twice: in vendor.js as well as in the chunk for the Angular library project that uses the loadRemoteModule function.

In my app, I call `setRemoteDefinitions` before Angular bootstrap, and then I call `loadRemoteModule` after bootstrap (dynamic plugin system).

It seems like `remoteUrlDefinitions` is correctly populated before bootstrap, but then the library chunk loads after bootstrap and resets `remoteUrlDefinitions` to `{}`.  So when I call `loadRemoteModule`, `remoteUrl` ends up undefined, causing the error.

### Expected Behavior

`nx-module-mf.mjs` shouldn't be duplicated at runtime and `loadRemoteModule` should find the remote definitions set before bootstrap with `setRemoteDefinitions`

### Steps to Reproduce

1. Create an MFE host and remote
2. call `setRemoteDefinitions` in main.ts before bootstrap
3. create a library that uses `loadRemoteModule`
4. call `loadRemoteModule` via the library after bootstrap

### Nx Report

```shell
Node: 20.16.0
npm: 10.8.1

nx: 19.6.6
typescript: 5.5.4

Community Plugins:
ngx-build-plus: 18.0.0
```

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

hakalb:
I tried the fix but I don't think the main issue is solved. Now there's just another console error.
![image](https://github.com/user-attachments/assets/9337767a-7a06-4d90-8163-a1a092ce74b3)

Still there must be duplicate instances.

A workaround is to make sure `setRemoteDefinitions` (or `setRemoteUrlResolver`) is located in the same project as `loadRemoteModule` to end up in the same bundle file.

Looks like it's also possible to add `@nx/angular/mf` to `additionalShared` in `module-federation.config.js` for the host application.

pcurrivan:
[...]

PR #33059, opened 2025-10-13
fix(react): dynamic module federation should not initialize remoteUrlDefinitions to empty object

### LINKED ISSUE #27793: MFE resolveRemoteUrl no longer working
### Current Behavior

```
    const remoteUrl = remoteUrlDefinitions
        ? remoteUrlDefinitions[remoteName]
        : await resolveRemoteUrl(remoteName);
```

remoteUrlDefinitions was changed to being an object by default not null in https://github.com/nrwl/nx/pull/27051, so the provided check always passes and resolveRemoteUrl won't be called.

### Expected Behavior

resolveRemoteUrl should be called by the code

### GitHub Repo

_No response_

### Steps to Reproduce

1. Make code that uses only setRemoteUrlResolver

### Nx Report

```shell
N/A
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

NX 19.6.5

Coly010:
## Current Behavior

  The `remoteUrlDefinitions` variable in `packages/react/mf/dynamic-federation.ts` is
  initialized as an empty object (`{}`), which causes:
  1. The `resolveRemoteUrl` callback is never called because `remoteUrlDefinitions` is
  always truthy
  2. When `loadRemoteModule` is called before `setRemoteDefinitions`, users get a cryptic
   error "Cannot read properties of undefined (reading 'endsWith')" instead of the
  helpful error message

  ## Expected Behavior

  - `resolveRemoteUrl` should be called when provided and `remoteUrlDefinitions` hasn't
  been set
  - The helpful error message should be shown when `loadRemoteModule` is called before
  setup

  ## Changes Made

  Applied the same fix from PR #27927 (for Angular) to the React implementation:
  - Changed `remoteUrlDefinitions` initialization to be `undefined` by default
  - Added nullish coalescing operator (`??=`) in `setRemoteDefinition` to initialize only
   when needed

  ## Related Issue(s)

  Fixes #33055

  This follows the same pattern as PR #27927 which fixed #27793 and #27842 for Angular.

### DIFF
--- a/packages/react/mf/dynamic-federation.ts
+++ b/packages/react/mf/dynamic-federation.ts
@@ -17,7 +17,7 @@ declare const document: {
 
 declare const __webpack_init_sharing__: (scope: 'default') => Promise<void>;
 declare const __webpack_share_scopes__: { default: unknown };
-let remoteUrlDefinitions: Record<string, string> = {};
+let remoteUrlDefinitions: Record<string, string>;
 let resolveRemoteUrl: ResolveRemoteUrlFunction;
 const remoteModuleMap = new Map<string, unknown>();
 const remoteContainerMap = new Map<string, unknown>();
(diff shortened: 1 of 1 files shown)


AMWJ:
I appreciate the rapid response!

jaysoo (approved):

PR #33060, opened 2025-10-13
feat(core): support providing schema in sync generators

jogelin:
## Summary

  Align sync generator invocation with regular generators by providing schema-based options from `nx.json` and project configuration.

## Current Behavior

  Currently, sync generators are invoked with only a `Tree` parameter and do not receive any configuration options from the generator's schema, `nx.json`, or project configuration:

  ```typescript
  // Before
  export type SyncGenerator = (tree: Tree) => SyncGeneratorResult;
```

  This means sync generators cannot:
  - Access generator-specific options defined in their schema
  - Use options configured in nx.json via generators configuration
  - Use project-level generator defaults
  - Apply schema defaults and validations

## Expected Behavior

  Sync generators should receive combined options from their schema, just like regular generators do, but without command-line arguments (since they're not invoked directly by users).

  ```typescript
  // After
  export type SyncGenerator<T = unknown> = (
    tree: Tree,
    schema: T
  ) => SyncGeneratorResult | Promise<SyncGeneratorResult>;
```

## Related Discussion

Related to https://github.com/nrwl/nx/discussions/32663

### DIFF
--- a/packages/nx/src/utils/sync-generators.ts
+++ b/packages/nx/src/utils/sync-generators.ts
@@ -2,16 +2,16 @@ import { performance } from 'perf_hooks';
 import { parseGeneratorString } from '../command-line/generate/generate';
 import { getGeneratorInformation } from '../command-line/generate/generator-utils';
 import type { GeneratorCallback } from '../config/misc-interfaces';
-import { readNxJson, type NxJsonConfiguration } from '../config/nx-json';
+import { type NxJsonConfiguration, readNxJson } from '../config/nx-json';
 import type { ProjectGraph } from '../config/project-graph';
 import type { TaskGraph } from '../config/task-graph';
 import type { ProjectConfiguration } from '../config/workspace-json-project-json';
 import { daemonClient } from '../daemon/client/client';
 import { isOnDaemon } from '../daemon/is-on-daemon';
 import {
+  type FileChange,
   flushChanges,
   FsTree,
-  type FileChange,
   type Tree,
 } from '../generators/tree';
 import {
@@ -20,6 +20,14 @@ import {
 } from '../project-graph/project-graph';
 import { updateContextWithChangedFiles } from './workspace-context';
 import { workspaceRoot } from './workspace-root';
+import {
+  coerceTypesInOptions,
+  convertAliases,
+  convertSmartDefaultsIntoNamedParams,
+  getNxJsonGeneratorDefaults,
+  setDefaults,
+  validateOptsAgainstSchema,
+} from './params';
 import chalk = require('chalk');
 
 export type SyncGeneratorResult = void | {
@@ -28,8 +36,9 @@ export type SyncGeneratorResult = void | {
   outOfSyncDetails?: string[];
 };
 
-export type SyncGenerator = (
-  tree: Tree
+export type SyncGenerator<T = unknown> = (
+  tree: Tree,
+  schema: T
 ) => SyncGeneratorResult | Promise<SyncGeneratorResult>;
 
 export type SyncGeneratorRunSuccessResult = {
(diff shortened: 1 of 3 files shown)


mpsanchis:
👍 interested

packages/nx/src/utils/sync-generators.ts
```diff
@@ -136,19 +138,35 @@ export async function collectAllRegisteredSyncGenerators(
...
-    const { implementationFactory } = getGeneratorInformation(
+    const { implementationFactory, schema, normalizedGeneratorName } =
+      getGeneratorInformation(
+        collection,
+        generator,
+        workspaceRoot,
+        projectsConfigurations.projects
+      );
+
+    // Combine options from nx.json and project configuration (no command line options for sync generators)
```
AgentEnder: This is really just reading the config from the nx.json config right? Sync generators are ran once per workspace, not once per project, so there's not really a way to narrow down which project's options should be read.

jogelin: Actually, since we’re inferring all project configurations — including sync generators — this feature is essential for us.

In some cases, we configure a sync generator on an existing target, but more commonly, we generate a dedicated sync target that contains a sync generator specific to the project type.

We also configure sync generators at the root project level.

jogelin: @AgentEnder any thoughts related to my reply? :)

jogelin: @AgentEnder I misunderstood your remark.

- I removed the usage of project generators configuration
- I re-used some part of the `combineOptionsForGenerator`

AgentEnder: Oh, cool - I was coming back to respond. Sorry for the delay :)

Is there a reason we can re-use combineOptionsForGenerator completely? Looking at the signature:

```
export async function combineOptionsForGenerator(
  commandLineOpts: Options,
  collectionName: string,
  generatorName: string,
  projectsConfigurations: ProjectsConfigurations,
  nxJsonConfiguration: NxJsonConfiguration,
  schema: Schema,
  isInteractive: boolean,
  defaultProjectName: string | null,
  relativeCwd: string | null,
  isVerbose = false
) 
```

We should be able to just pass null for defaultProjectName and relativeCwd. `isInteractive` would be false, commandLineOpts would be an empty object I believe

jogelin: @AgentEnder 
I am confused, I was using it before with the defaultProjectName to null  but you made that comment:

> This is really just reading the config from the nx.json config right? Sync generators are ran once per workspace, not once per project, so there's not really a way to narrow down which project's options should be read.

And in fact, `combineOptionsForGenerator` will load generators config from project configurations which does not make sense for sync generators.

I can re-used `combineOptionsForGenerator` but I have to provide `projectsConfigurations` otherwise it will not load generator defaults from `nx.json` because of that condition:

https://github.com/nrwl/nx/blob/0014ae66d86cecb306c261bb7c37303537e69420/packages/nx/src/utils/params.ts#L656

This is why I split that function to be able to re-use the code that loads Nx Configuration only.

mpsanchis: What I understand is that we have 2 options:

(a) We load all configs from everywhere (nx.json + projects), like you're saying now @AgentEnder . This will work as expected, as long as plugins don't start inferring different options for different projects.

(b) We only load configs from nx.json, like @jogelin interpreted you were suggesting before. This will work as long as configuration is only declared in nx.json (if plugins infer some options for projects, they will be silently ignored).

Both can work. The former is more powerful (because we can use Crystal for sync generators, and infer the options), but it can also generate gotcha situations. Of course @AgentEnder you make the calls here :D

jogelin:
Hey @AgentEnder @FrozenPandaz, any chance we can move forward on this? It would be great to have it ready for 22.1 if possible.

jogelin:
In the [Nx documentation ](https://nx.dev/docs/reference/nx-json#sync) I just discovering:

image

This is only used by the `typescript-sync` generator which is reading NxConf directly:
https://github.com/nrwl/nx/blob/081bec53cf23beab0bae332bd589d24a4745f638/packages/js/src/generators/typescript-sync/typescript-sync.ts#L201

Should we expect sync generators to read and apply options manually without any schemas?

I’m confused why this appears in the docs as a general way to configure sync generators…

mpsanchis:
Hi folks!
Anything we can do to move this forward? I'm happy to contribute with anything 😄

Gowza68:
💖

Scott reacted via Gmail
<https://www.google.com/gmail/about/?utm_source=gmail-in-product&utm_medium=et&utm_campaign=emojireactionemail#app>

PR #33166, opened 2025-10-21
fix(release)!: better respect version plan file contents for changelog entries

JamesHenry:
## Current Behavior

The rendered changelog is not faithful enough to the original version plan markdown contents.

## Expected Behavior

The version plan author maintains maximum control over final changelog contents.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/release/changelog-renderer/index.ts
+++ b/packages/nx/release/changelog-renderer/index.ts
@@ -184,13 +184,6 @@ export default class DefaultChangelogRenderer {
     }
 
     if (this.isVersionPlans) {
-      this.conventionalCommitsConfig = {
-        types: {
-          feat: DEFAULT_CONVENTIONAL_COMMITS_CONFIG.types.feat,
-          fix: DEFAULT_CONVENTIONAL_COMMITS_CONFIG.types.fix,
-        },
-      };
-
       for (let i = this.relevantChanges.length - 1; i >= 0; i--) {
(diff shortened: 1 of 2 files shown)


packages/nx/release/changelog-renderer/index.ts
```diff
@@ -184,13 +184,6 @@ export default class DefaultChangelogRenderer {
     }
 
     if (this.isVersionPlans) {
-      this.conventionalCommitsConfig = {
```
JamesHenry, line 187: I am honestly not sure why this was hardcoded to defaults, it means custom config isn't respected even when present. It definitely makes sense to get rid of this, the defaults will have been set in `config.ts` in the case the user is not customizing

packages/nx/src/command-line/release/changelog.ts
```diff
@@ -1116,7 +1116,7 @@ async function generateChangelogForWorkspace({
     changes,
     changelogEntryVersion: releaseVersion.rawVersion,
     project: null,
-    isVersionPlans: false,
+    isVersionPlans: !!nxReleaseConfig.versionPlans,
```
JamesHenry, line 1119: Again I don't have a good explanation for why this was previously hardcoded to `false`... it should respect the config

FrozenPandaz (approved):

Coly010 (approved):

PR #33187, opened 2025-10-22
feat(gradle): add custom gradle executable directory to options

lourw:
## Current Behavior

We search for the gradle wrapper of a project by first looking at the project root, then traversing upwards to the workspace root. If a workspace has a separate gradle project defined that does not contain a wrapper, then Nx will error. 

## Expected Behavior

Add a field to the gradle plugin that will allow users to specify a custom gradle installation within their workspace. Nx will check for a gradle wrapper that the specified location when executing gradle tasks.

## Related Issue(s)

Fixes NXC-3147


MaxKless (approved):

packages/gradle/src/utils/exec-gradle.ts
```diff
@@ -93,14 +134,49 @@ export function findGradlewFile(
...
+  } else {
+    if (existsSync(customGradlewPath)) {
+      // Return path relative to workspace root if it was relative, otherwise return absolute
+      return isAbsolute(customGradleInstallationPath)
+        ? customGradlewPath
+        : join(customGradleInstallationPath, 'gradlew');
+    }
+  }
+
+  return undefined;
```
FrozenPandaz: If we can't find one.. I say we throw an error since it's misconfiguration.

packages/gradle/src/plugin/utils/gradle-plugin-options.ts
```diff
@@ -1,6 +1,7 @@
 export interface GradlePluginOptions {
   testTargetName?: string;
   ciTestTargetName?: string;
+  customGradleInstallation?: string;
```
FrozenPandaz:
```suggestion
  gradleExecutableDirectory?: string;
```

I think this works better?

FrozenPandaz (approved):

barbados-clemens (approved):

PR #33224, opened 2025-10-23
feat(nx-dev): add GitHub stars count to header

DipakHalkude:
## Description
This PR adds a GitHub stars count button to the Nx marketing site header to showcase the project's community traction and provide social proof.

## Changes
- Added a GitHub stars button in the header's secondary navigation section
- Uses consistent styling with existing header buttons
- Includes proper accessibility attributes and security rel tags
- Displays current star count (27.2k)

## Implementation Details
- Button links to https://github.com/nrwl/nx
- Uses GitHub octicon SVG for the star icon
- Currently displays static count of "27.2k" (can be made dynamic in future)
- Follows existing Tailwind CSS patterns and color schemes
- Positioned before the existing GitHub icon in the header

## Benefits
- Highlights Nx's growing community (27.2k+ stars)
- Encourages more GitHub stars through visibility
- Adds credibility for new visitors exploring Nx
- Consistent with common open-source project patterns

## Testing
- [ ] Button appears in header on desktop view
- [ ] Clicking button opens Nx GitHub repo in new tab
- [ ] Styling matches existing header buttons

Closes #33028


DipakHalkude:
@barbados-clemens @FrozenPandaz @jaysoo Please have a look :)

nx-dev/ui-common/src/lib/headers/header.tsx
```diff
@@ -328,6 +328,25 @@ export function Header({
...
+              <svg
+                width="16"
+                height="16"
+                viewBox="0 0 16 16"
+                fill="currentColor"
+                className="flex-shrink-0"
+              >
+                <path d="M8 0C3.58 0 0 3.58 0 8c0 3.54 2.29 6.53 5.47 7.59.4.07.55-.17.55-.38 0-.19-.01-.82-.01-1.49-2.01.37-2.53-.49-2.69-.94-.09-.23-.48-.94-.82-1.13-.28-.15-.68-.52-.01-.53.63-.01 1.08.58 1.23.82.72 1.21 1.87.87 2.33.66.07-.52.28-.87.51-1.07-1.78-.2-3.64-.89-3.64-3.95 0-.87.31-1.59.82-2.15-.08-.2-.36-1.02.08-2.12 0 0 .67-.21 2.2.82.64-.18 1.32-.27 2-.27.68 0 1.36.09 2 .27 1.53-1.04 2.2-.82 2.2-.82.44 1.1.16 1.92.08 2.12.51.56.82 1.27.82 2.15 0 3.07-1.87 3.75-3.65 3.95.29.25.54.73.54 1.48 0 1.07-.01 1.93-.01 2.2 0 .21.15.46.55.38A8.013 8.013 0 0016 8c0-4.42-3.58-8-8-8z" />
+              </svg>
+              <span>27.2k</span>
```
julekgwa: Could we please fetch the GitHub star count dynamically from the GitHub API instead of hardcoding the value (27.2k)?

api: https://api.github.com/repos/nrwl/nx

`"stargazers_count": 27302,`

DipakHalkude: Yes, Just I am waiting for the review from member. After that I will proceed as they tell.

barbados-clemens:
Hey @DipakHalkude sorry for the late reply here, but yeah style looks good. like @julekgwa mentioned if we can get the live value so we don't have to maintain the value. the data should already be in the content collection since we fetch the plugin stats for the nx repo. logic is in the plugin.loader.ts. but should be able to get the nx overview entry in the header component. you can use the largeNumberFormatter from the formatters.ts file to format the value. as well.

barbados-clemens (changes requested):
style looks fine just want to make it dynamic so we don't need to manually update the value

DipakHalkude:
Ok I am on it :)

DipakHalkude:
@barbados-clemens Please have a look :)

DipakHalkude:
Done from my side :) Thank you for the suggestion @barbados-clemens @julekgwa

juristr:
we should also be aware of the space in the header nav. So I'd just opt for having either of the two GitHub buttons there
image

DipakHalkude:
We can remove the second one as when we are clicking on any of the button it goes to github repo only. So we can remove the 2nd one

DipakHalkude:
@juristr What you think? 
```
<a
title="Nx is open source, check the code on GitHub"
href="https://github.com/nrwl/nx"
target="_blank"
rel="noreferrer"
className="inline-flex items-center px-3 py-2 opacity-60 hover:opacity-90"
>
<span className="sr-only">Nx on GitHub</span>
<GitHubIcon aria-hidden="true" className="h-5 w-5" />
</a>
```

We can remove this one!
Just waiting for your response then I make changes. Thank you :)

DipakHalkude:
@juristr  I guess now the implementation looks great :)

DipakHalkude:
@juristr @barbados-clemens  Can someone review this :)

DipakHalkude:
Can anyone please review this PR :)

nx-dev/ui-common/src/lib/headers/header.tsx
```diff
@@ -52,9 +52,38 @@ export function Header({
...
+      }
+    } catch (error) {
+      console.error('Failed to fetch GitHub stars:', error);
+    }
+  };
+  getStarCount();
+  return () => {
+    isMounted = false;
+  };
+}, []);
```
barbados-clemens: this should not be making external API calls. we already have that info when generating the sites. pass this value into the component to render.

barbados-clemens (changes requested):

nx-dev/ui-common/src/lib/headers/header.tsx
DipakHalkude: Hi, I removed the external API call and updated the code to use the pre-generated value as suggested. But I’m still not seeing the GitHub stars on the page. Could you please help me check what I might be missing?

DipakHalkude:
GitHub stars are now showing up on the changelog page, which is great! But they're still missing from the other pages like the homepage, AI page, and community section.

The reason is that only the changelog page currently has the code to fetch the star count during build time. The other pages don't have this setup yet.

We can't just put the star-fetching code directly in the header component because the header runs in the browser, but fetching GitHub stars requires server-side code and secure tokens that can't be exposed to the client.

What we really need is a way to fetch the stars once and share that data across all pages, rather than having to copy the same code into every single file. That would be much cleaner and more maintainable than updating 20+ individual pages.


image

DipakHalkude:
Please have a look @barbados-clemens

barbados-clemens:
Hey @DipakHalkude sorry about this. I thought your changes had already landed, but turned out it was this PR I wasn't aware of: https://github.com/nrwl/nx/pull/33289

The team decided we didn't want to put the stars count in the header as it cluttered up too much, so left it only on the docs portion of the docs.
So i'll close this PR for now. sorry this missed my radar for so long, didn't mean to keep you waiting for so long.

DipakHalkude:
Not an issue at all :)
I actually learned a lot while working on this PR and through the discussions on the issue.
Thank you for giving me the opportunity and for the guidance throughout!

PR #33232, opened 2025-10-24
chore(repo): dogfood pnpm catalogs

leosvelperez:
Dogfoods the Pnpm Catalogs feature in the Nx repo. This is the first step to move all dependencies to Pnpm Catalogs definitions. More work will be done incrementally later as we consolidate package versions across the repository.

The initial list of dependencies moved to Pnpm Catalogs is:

- Angular packages
- React packages
- TypeScript packages
- Jest packages
- Rspack packages
- Some common utilities


Coly010 (approved):

jaysoo (approved):

packages/gradle/src/utils/exec-gradle.ts
```diff
@@ -34,14 +34,19 @@ export function execGradleAsync(
...
+      args,
+      {
+        cwd: dirname(gradleBinaryPath),
+        shell: true,
+        windowsHide: true,
+        env: process.env,
+        maxBuffer: LARGE_BUFFER,
+        ...execOptions,
+      },
+      undefined
```
FrozenPandaz, line 53: Why was this added? 👀

packages/jest/src/executors/jest/jest.impl.ts
```diff
@@ -33,7 +33,7 @@ export async function jestExecutor(
 
   const config = await parseJestConfig(options, context);
 
-  const { results } = await runCLI(config, [options.jestConfig]);
+  const { results } = await runCLI(config as any, [options.jestConfig]);
```
FrozenPandaz: Why `as any`? 👀

FrozenPandaz (approved):

packages/gradle/src/utils/exec-gradle.ts
leosvelperez: TypeScript version changed when it was consolidated, and it's stricter. The inference didn't work without including `undefined` for the callback.

packages/jest/src/executors/jest/jest.impl.ts
leosvelperez: There's a type mismatch due to different resolved versions of `@jest/core` and `@types/jest`. It's only a typing issue that doesn't affect runtime, and it doesn't affect any other part of the executor code.

leosvelperez: The type mismatch was actually caused by accidentally forcing an update for the `@types/jest` resolved version. It's fixed now.

PR #33233, opened 2025-10-24
chore(repo): enable verbose logging

rarmatei:
Enables verbose logging and ensures tests use a separate env var for verbosity, so that we can switch verbose CI logs on/off without worrying that it will affect the tests.


Cammisuli (approved):

e2e/utils/get-env-info.ts
```diff
@@ -141,7 +137,7 @@ export {
 } from './ensure-browser-installation';
 
 export function getStrippedEnvironmentVariables() {
-  return Object.fromEntries(
+  const env = Object.fromEntries(
```
jaysoo: Nit: This can still just `return`

jaysoo (approved):

rarmatei:
closing after talking with Jason - we're with enabling verbose logging as needed, even if some e2e tests fail

PR #33235, opened 2025-10-24
fix(docker): handle undefined options when creating graph

Coly010:
## Current Behavior
When `@nx/docker` is registered via string only (`nxJson.plugins: ["@nx/docker"]`, project graph creation fails because we try to access `options.buildTarget`.

## Expected Behavior
Handle undefined `options` gracefully, and still create the default target.

## Related Issue(s)

Fixes NXC-3320

### DIFF
--- a/packages/docker/src/plugins/plugin.ts
+++ b/packages/docker/src/plugins/plugin.ts
@@ -348,7 +348,7 @@ function normalizePluginOptions(
   };
 
   return {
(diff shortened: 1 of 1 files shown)


jaysoo (approved):

emiliosheinz:
@Coly010 it would be really nice to have this released under a beta tag so that I can unblock myself. Thanks!

PR #33239, opened 2025-10-24
fix(maven): fix mvn not available in child process

jogelin:
## Description

  Fixes the "spawn mvn ENOENT" error that occurs when the Maven executable is not found in the PATH of the spawned child process.

  ## Problem

  When using the `@nx/maven` plugin, users encounter the following error:

  NX   Failed to process project graph.

   - Error: Failed to spawn Maven process: spawn mvn ENOENT
    at ChildProcess.<anonymous> (/node_modules/@nx/maven/dist/plugins/maven-analyzer.js:126:20)

  ### Root Cause

  The `maven-analyzer.js` file spawns a child process to run Maven commands, but it doesn't pass environment variables to the spawned process. This means the child process doesn't inherit the parent's `PATH` environment variable, which is needed to locate the `mvn` executable.

  This issue is particularly problematic when:
  - Maven is managed by version managers like `mise`, `asdf`, or `sdkman`
  - The Nx daemon is running (daemon processes have a different environment)
  - Maven is installed in a non-standard location that's in the user's PATH

  ## Solution

  Add `env: process.env` to the `spawn` options to inherit environment variables from the parent process, including the critical `PATH` variable.

  **Before:**
  ```javascript
  const child = (0, child_process_1.spawn)(mavenExecutable, mavenArgs, {
      cwd: workspaceRoot,
      stdio: 'pipe',
  });
  ```
  
  **After:**
  ```javascript
  const child = (0, child_process_1.spawn)(mavenExecutable, mavenArgs, {
      cwd: workspaceRoot,
      stdio: 'pipe',
      env: process.env, // Inherit environment variables from parent process
  });
```
  
  ##  Testing

  Tested with:
  - Maven 3.9.5 managed by mise
  - Java 17 managed by mise
  - Nx 22.0.1 with @nx/maven 22.0.1

  The fix allows the plugin to successfully locate and execute Maven commands.

  ## Impact

  - ✅ Fixes Maven detection when using version managers
  - ✅ No breaking changes - purely additive
  - ✅ Improves compatibility with various development environments
  - ✅ Aligns with Node.js best practices for spawning child processes

  This PR description clearly explains the issue, the root cause, and the solution in a way that maintainers and other developers can easily understand.

### DIFF
--- a/packages/maven/src/plugins/maven-analyzer.ts
+++ b/packages/maven/src/plugins/maven-analyzer.ts
@@ -100,7 +100,8 @@ export async function runMavenAnalysis(
       cwd: workspaceRoot,
       windowsHide: true,
       shell: true,
-      stdio: 'pipe', // Always use pipe so we can control output
+      stdio: 'pipe', // Always use pipe so we can control output,
+      env: process.env,
     });
 
     logger.verbose(`[Maven Analyzer] Process spawned with PID: ${child.pid}`);

### CODE REVIEW

**Comment by jogelin** [comment]
@FrozenPandaz can you check

File: packages/maven/src/plugins/maven-analyzer.ts [hunk 1]
```diff
@@ -100,7 +100,8 @@ export async function runMavenAnalysis(
       cwd: workspaceRoot,
       windowsHide: true,
       shell: true,
-      stdio: 'pipe', // Always use pipe so we can control output
+      stdio: 'pipe', // Always use pipe so we can control output,
+      env: process.env,
```
FrozenPandaz, line 104: This change technically does nothing because `spawn` defaults to `process.env`... 🤔 

https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

Did this somehow fix things for you?

FrozenPandaz (changes requested):

FrozenPandaz:
Thanks for trying out `@nx/maven`! :pray:

packages/maven/src/plugins/maven-analyzer.ts
jogelin: yes it does...
let me reproduce it on clean repo

jogelin:
> Thanks for trying out `@nx/maven`! 🙏

you are welcome, we followed similar approach in internal plugin so it looks promising

packages/maven/src/plugins/maven-analyzer.ts
jogelin: you can try on this repo

https://github.com/jogelin/test-nx-maven

I am using [mise](https://mise.jdx.dev/installing-mise.html) which is providing mvn in the path

jogelin: @FrozenPandaz did you try it?

jogelin:
Look like it is only happening on my local env ;)

PR #33264, opened 2025-10-27
feat(gradle): allow specifying project and task configuration from gradle build files

MaxKless:
## Current Behavior 
Gradle projects in Nx workspaces can only configure metadata through Gradle's built-in mechanisms. There's no way to specify Nx-specific project metadata (like tags) or customize task target configurations other than overriding with `project.json`

 ## Expected Behavior

  Developers can now configure Nx-specific metadata for both projects and tasks using a type-safe Kotlin/Groovy DSL:

 ### Project-level metadata (in build.gradle.kts):
  ```
nx {
    set("name", "my-service")
    array("tags", "scope:backend", "type:api")
    set("description", "Payment processing service")
  }
```

 ### Task-level metadata (in build.gradle.kts):
```
  tasks.named("integrationTest") {
    nx {
      set("cache", false)
      array("tags", "integration", "slow")
    }
  }
```

### DIFF
--- a/astro-docs/src/content/docs/technologies/java/gradle/introduction.mdoc
+++ b/astro-docs/src/content/docs/technologies/java/gradle/introduction.mdoc
@@ -421,13 +421,13 @@ tasks.named('integrationTest') {
 
 **Available DSL methods for task-level configuration:**
 
-- `dependsOn` - ListProperty for adding Nx task dependencies (see important note below)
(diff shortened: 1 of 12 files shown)


lourw:
should we do a docs change in this PR?

lourw (approved):

FrozenPandaz (approved):

PR #33265, opened 2025-10-27
feat(vite): add atomizer support for vitest

JamesHenry:
## Current Behavior

The `vitest` plugin within `@nx/vite` does not support Nx Atomizer.

## Expected Behavior

The `vitest` plugin within `@nx/vite` supports Nx Atomizer.

The contents of this PR are applied as a patch on a real repo here: https://github.com/angular-eslint/angular-eslint/pull/2720

Everything performs and displays correctly in Nx Cloud:

image

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -6,10 +6,10 @@ import {
   detectPackageManager,
   getPackageManagerCommand,
   joinPathFragments,
-  normalizePath,
   ProjectConfiguration,
   readJsonFile,
   TargetConfiguration,
+  workspaceRoot,
   writeJsonFile,
 } from '@nx/devkit';
 import { calculateHashesForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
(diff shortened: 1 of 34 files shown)


packages/vite/src/plugins/plugin.ts
```diff
@@ -565,6 +666,7 @@ function normalizeOptions(options: VitePluginOptions): VitePluginOptions {
   options.testTargetName ??= 'test';
   options.serveStaticTargetName ??= 'serve-static';
   options.typecheckTargetName ??= 'typecheck';
+  options.ciTargetName ??= 'e2e-ci';
```
Coly010: Are we always atomizing because we’re setting a default regardless if the user has requested it?

packages/vite/src/plugins/plugin.ts
```diff
@@ -583,3 +685,39 @@ function checkIfConfigFileShouldBeProject(
 
   return true;
 }
+
+async function getTestPaths(
```
Coly010: I don’t think this is accurate enough. It doesn’t take into account vitest’s projects syntax for test

Coly010: We should see if we can invoke this programmatically: https://vitest.dev/guide/cli.html#vitest-list

Coly010 (changes requested):

packages/vite/src/plugins/plugin.ts
Coly010: Primary concern being that we are globing and increasing project graph creation time as a result

packages/vite/src/plugins/plugin.ts
JamesHenry: Thanks, done

packages/vite/src/plugins/plugin.ts
JamesHenry: Thanks, done

Coly010 (approved):

packages/vite/src/plugins/plugin.ts
```diff
@@ -583,3 +676,20 @@ function checkIfConfigFileShouldBeProject(
...
+  const { createVitest } = await import('vitest/node');
+  const vitest = await createVitest('test', {
+    dir: fullProjectRoot,
+    filesOnly: true,
+    watch: false,
+  });
+  const relevantTestSpecifications =
+    await vitest.getRelevantTestSpecifications();
+  return relevantTestSpecifications.map((ts) =>
+    relative(projectRoot, ts.moduleId)
```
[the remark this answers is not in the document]
JamesHenry: This already happens in the consuming code

PR #33272, opened 2025-10-27
chore(nx-dev): remove link checker from build so it is only checked during CI not deploy

jaysoo:
This PR removes the need to check links during deploy and instead enforces it in CI. This removes the need to build astro-docs when building next.js app.

Reduces Vercel build from 10-11 mins to 6.5 mins.

image

### DIFF
--- a/nx-dev/nx-dev/next.config.js
+++ b/nx-dev/nx-dev/next.config.js
@@ -4,7 +4,10 @@ const redirectRules = require('./redirect-rules');
 
 if (!process.env.NEXT_PUBLIC_ASTRO_URL && !global.NX_GRAPH_CREATION) {
   // If we're building for production throw error as each env must set this value.
-  if (process.env.NODE_ENV === 'production') {
+  if (
+    process.env.NODE_ENV === 'production' &&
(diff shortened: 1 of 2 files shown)


nx-dev/nx-dev/next.config.js
```diff
@@ -4,7 +4,10 @@ const redirectRules = require('./redirect-rules');
 
 if (!process.env.NEXT_PUBLIC_ASTRO_URL && !global.NX_GRAPH_CREATION) {
   // If we're building for production throw error as each env must set this value.
-  if (process.env.NODE_ENV === 'production') {
+  if (
+    process.env.NODE_ENV === 'production' &&
+    (process.env.VERCEL || process.env.NETLIFY)
```
jaysoo, line 9: Since we're making the build/check-links required for `test`, it kinda sucks locally when the build fails due to a variable only needed for prod.

barbados-clemens: should we make these env vars as env var inputs?

barbados-clemens (approved):

PR #33275, opened 2025-10-28
fix(docker): guard commitSha null in plugin interpolation

emiliosheinz:
## Current Behavior
Docker plugin assumed `commitSha` was always non-null; when `null`, `shortCommitSha.slice` caused a runtime error during target interpolation.

## Expected Behavior
Plugin should succeed even if latest commit SHA cannot be resolved, simply omitting shortCommitSha-based substitutions.

## Changes
- Added null guard: `shortCommitSha` now set to `commitSha ? commitSha.slice(0,7) : null`.
- Added test "should not throw when commitSha is null" verifying node / target creation succeeds.
- No breaking changes; only broadens safe input surface.

## Additional notes
Logic only executes when `commitSha` was previously null (error case); normal paths unchanged. If consumers interpolate `{shortCommitSha}`, they should handle possible null (unchanged if interpolation is already optional).

### DIFF
--- a/packages/docker/src/plugins/plugin.spec.ts
+++ b/packages/docker/src/plugins/plugin.spec.ts
@@ -776,6 +776,20 @@ describe('@nx/docker', () => {
   });
 
   describe('pattern interpolation', () => {
+    it('should not throw when commitSha is null', async () => {
+      mockGetLatestCommitSha.mockReturnValue(null);
+      await tempFs.createFiles({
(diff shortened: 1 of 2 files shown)


Coly010 (approved):

emiliosheinz:
Hey @Coly010, it would be lovely if we could get this merged and released on a beta channel. It's currently blocking me from moving forward on a personal project 😬

PR #30259, opened 2025-03-04
fix(nextjs): ensure `eslint-config-next` matches Next.js 14 and 15 versions

tuffz:
This PR builds on the previous fix that ensured `eslint-config-next` was correctly installed when using Next.js 15. #30258 

Now, the logic has been further refined to dynamically determine the installed Next.js version and install the corresponding `eslint-config-next` version accordingly.

### Current Behavior

- `next@~15.1.4` is installed
- `eslint-config-next@14.2.16` is installed (incorrect for Next.js 15)

### Expected Behavior

If Next.js 15 is detected -> eslint-config-next@15.1.4 is installed
If Next.js 14 is detected -> eslint-config-next@14.2.16 is installed

### GitHub Repo

https://github.com/tuffz/new-nx-with-preset-nextjs

### Steps to Reproduce

1. Run the following command from the official documentation:
`npx create-nx-workspace@latest --preset=next`
3. Open `package.json` and check the installed dependencies

- `next@~15.1.4` is installed
- `eslint-config-next@14.2.16` is installed

## Related Issue(s)

Fixes #30257 (& #30258)

PR #33296, opened 2025-10-29
feat(nextjs): add support for next 16

### LINKED ISSUE #33207: (next): upgrade to next 16
### Current Behavior

Currently Nx is using next 15

### Expected Behavior

Now that Next 16 is available, generators/executors should use this version for new apps

### GitHub Repo

_No response_

### Steps to Reproduce

1.

### Nx Report

```shell
N/A
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Coly010:
## Current Behavior
We currently support Next 14 and 15.

## Expected Behavior
Add support for Next 16, bringing support to 14, 15, 16.
Existing workspaces will continue to use the version they are on.
New workspaces will use Next 16.

Refer to Next 16 Migration Guide for migrating from Next 15 to 16

## Related Issue(s)

Fixes #33207

### DIFF
--- a/e2e/next/src/next-appdir.test.ts
+++ b/e2e/next/src/next-appdir.test.ts
@@ -31,6 +31,8 @@ describe('Next.js App Router', () => {
     );
     runCLI(`generate @nx/js:lib ${jsLib} --no-interactive`);
 
+    // Turbopack interprets the TS source incorrectly assuming ESM despite package.json type stating module
+    // TODO(Colum): remove this when JS Lib generator switches to ESM
     updateJson(`${jsLib}/package.json`, (json) => {
       delete json.type;
       return json;
--- a/e2e/next/src/next-appdir.test.ts
+++ b/e2e/next/src/next-appdir.test.ts
@@ -7,6 +7,7 @@ import {
   runE2ETests,
   uniq,
   updateFile,
+  updateJson,
 } from '@nx/e2e-utils';
 
 describe('Next.js App Router', () => {
@@ -29,6 +30,11 @@ describe('Next.js App Router', () => {
       `generate @nx/next:app ${appName} --e2eTestRunner=playwright --appDir=true`
     );
     runCLI(`generate @nx/js:lib ${jsLib} --no-interactive`);
+
+    updateJson(`${jsLib}/package.json`, (json) => {
+      delete json.type;
+      return json;
+    });
 
     checkFilesExist(`${appName}/src/app/page.tsx`);
     checkFilesExist(`${appName}-e2e/src/example.spec.ts`);
@@ -69,7 +75,7 @@ describe('Next.js App Router', () => {
         `e2e ${appName}-e2e --configuration=production`
       );
       expect(e2eResults).toContain('Successfully ran target e2e for project');
-      expect(await killPorts()).toBeTruthy();
+      await killPorts();
     }
   }, 300_000);
 });
--- a/packages/next/docs/server-next-executor-examples.md
+++ b/packages/next/docs/server-next-executor-examples.md
@@ -62,19 +62,24 @@ This is the default configuration for Next.js standalone projects. Our `@nx/next
 ```
 
 {% /tab %}
-{% tab label="Enable turbo" %}
+{% tab label="Choosing your bundler" %}
 
-Turbopack (beta) is a cutting-edge bundler designed for JavaScript and TypeScript. To read more about support features see [Next.js Turbopack Documentation](https://turbo.build/pack/docs/features)
+Turbopack is a cutting-edge bundler designed for JavaScript and TypeScript. To read more about supported features see [Next.js Turbopack Documentation](https://turbo.build/pack/docs/features)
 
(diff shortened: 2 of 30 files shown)


tuffz:
Hi Nx-Team,

could you please consider that the `eslint-config-next` needs to be the same major version as the Next.js version: https://github.com/nrwl/nx/pull/30259 🙏

ianldgs:
Nit: I think it should also handle the turbo/webpack flags.

In next <= 15, for `nx serve` you could pass `--turbo` to use turbopack.
Seems like for `nx build --turbo` it never worked.

In next >= 16, they've flipped it: turbopack is enabled by default.
You can still use webpack with the flag `next dev --webpack` / `next build --webpack`.

The executor still only has `turbo`, and for `serve`:
https://github.com/nrwl/nx/blob/20d31fd6c085950eb2940cf3421ff2caf264d497/packages/next/src/executors/server/schema.json#L53-L56

ianldgs (approved):

packages/next/plugins/with-nx.ts
```diff
@@ -243,12 +244,21 @@ export function getNextConfig(
...
   const userWebpack = nextConfig.webpack || ((x) => x);
   const { nx, ...validNextConfig } = nextConfig;
-  return {
-    eslint: {
+
+  const baseConfig: NextConfig = {
+    ...validNextConfig,
+  };
+
+  const nextJsVersion = require('next/package.json')?.version ?? '16.0.1';
```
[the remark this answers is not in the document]
Coly010: If next is not installed, this path wouldn't be hit as the config file wouldn't load

packages/next/src/generators/application/application.spec.ts
```diff
@@ -654,6 +654,33 @@ describe('app', () => {
...
+        const packageJson = readJson(tree, '/package.json');
+        expect(packageJson).toMatchObject({
+          devDependencies: {
+            'eslint-config-next': '^16.0.1',
+            '@next/eslint-plugin-next': '^16.0.1',
+          },
+        });
+      });
+
+      it('should install eslint-config-next@14 when an existing Next.js 15 project is detected', async () => {
```
tuffz:
```suggestion
      it('should install eslint-config-next@15 when an existing Next.js 15 project is detected', async () => {
```

packages/next/src/utils/versions.ts
```diff
@@ -1,9 +1,13 @@
 export const nxVersion = require('../../package.json').version;
 
-export const nextVersion = '~15.2.4';
+export const next16Version = '~16.0.1';
+export const next15Version = '~15.2.4';
 export const next14Version = '~14.2.26';
```
tuffz, line 5:
```suggestion
export const next16Version = '^16.0.1';
export const next15Version = '^15.2.4';
export const next14Version = '^14.2.26';
```

Shouldn’t we use carets (`^`) here to allow receiving minor and patch updates within the same major version?

packages/next/src/utils/versions.ts
```diff
@@ -1,9 +1,13 @@
...
-export const eslintConfigNextVersion = '^15.2.4';
+export const nextVersion = next16Version;
+export const eslintConfigNext16Version = '^16.0.1';
+export const eslintConfigNext15Version = '^15.2.4';
 export const eslintConfigNext14Version = '~14.2.26';
+export const eslintConfigNextVersion = eslintConfigNext16Version;
```
tuffz, line 10:
```suggestion
export const nextVersion = next16Version;

export const eslintConfigNext16Version = next16Version;
export const eslintConfigNext15Version = next15Version;
export const eslintConfigNext14Version = next14Version;
export const eslintConfigNextVersion = eslintConfigNext16Version;

```

As the `eslintConfigNext` versions currently match the `next` versions, should we assign them directly to `nextVersion` and retain the separate variables for potential future differentiation?

e2e/next/src/next-appdir.test.ts
```diff
@@ -30,6 +31,11 @@ describe('Next.js App Router', () => {
     );
     runCLI(`generate @nx/js:lib ${jsLib} --no-interactive`);
 
+    updateJson(`${jsLib}/package.json`, (json) => {
```
jaysoo, line 36: A comment here would help explain why this was added. We can also remove it once JS library is `type: module` by default.

packages/next/docs/server-next-executor-examples.md
```diff
@@ -89,14 +94,37 @@ Updating the build options to include `turbo`.
...
-        //
+        }
+      }
     }
-  }
 ```
 
-```bash
-nx run acme:serve
+### Using webpack in Next.js 16 and above
```
jaysoo: Nit: `webpack` -> `Webpack` for consistent cases. `Turbopack` is already capitalized.

jaysoo (approved):

FrozenPandaz (approved):

PR #33299, opened 2025-10-29
cleanup(repo): use mise for agent toolchain installs

barbados-clemens:
also dedupe agent launch templates with yaml anchors for better readability


mise.toml
```diff
@@ -3,3 +3,4 @@ dotnet = "9"
 java = "24"
 node = "24"
```
barbados-clemens: note this was already here for node 24, but previously we weren't running node 24, but node 20.19. we should probs drop this to node 20.19 to be consistent

mise.toml
```diff
@@ -3,3 +3,4 @@ dotnet = "9"
...
 rust = "1.90.0"
```
barbados-clemens, line 5: note: this was already here, and previously we were using rust 1.70 in CI, but figured it's probs fine to bump to v1.90

mise.toml
```diff
@@ -3,3 +3,4 @@ dotnet = "9"
...
+bun = "1.3"
```
barbados-clemens, line 6: Bun was installed in the pipeline manually, switched to use mise with the version that was being installed v1.3.x

mise.toml
```diff
@@ -3,3 +3,4 @@ dotnet = "9"
 java = "24"
```
barbados-clemens, line 3: note: this was already here for java 24 but the pipeline was using java 21. probs should move back to java 21?

mise.toml
```diff
@@ -1,5 +1,6 @@
 [tools]
 dotnet = "9"
-java = "24"
-node = "24"
+java = "21"
```
FrozenPandaz: Did you run into issues with Java 24?

FrozenPandaz (approved):

PR #33311, opened 2025-10-30
feat(vitest): split vitest into @nx/vitest plugin

Coly010:
Split `vitest` out of `@nx/vite` and create a new `@nx/vitest`.

This allows for each plugin to have a single responsibility. One for bundling, one for testing.

It should allow for lighter setups where Vite for Bundling is not required, but users still want to use vitest for testing.


Stanzilla:
Will this then also be used by the nx Storybook plugin for tests?

e2e/react/src/react-rsbuild.test.ts
```diff
@@ -41,27 +41,6 @@ describe('Build React applications and libraries with Rsbuild', () => {
...
-
-    const appLintResults = await runCLIAsync(`lint ${rsbuildApp}`);
-    expect(appLintResults.combinedOutput).toContain(
-      'Successfully ran target lint'
-    );
-
-    await runCLIAsync(`build ${rsbuildApp}`);
-    checkFilesExist(`apps/${rsbuildApp}/dist/index.html`);
-  }, 300_000);
-
```
Coly010, line 64: This was a duplicated set of tests

e2e/vite/src/vite.test.ts
```diff
@@ -301,14 +301,11 @@ describe('@nx/vite/plugin', () => {
...
     });
 
-    it('should contain targets build, test and lint', () => {
+    it('should contain targets test', () => {
       const nxJson = readJson('nx.json');
 
-      const vitePlugin = nxJson.plugins.find(
-        (p) => p.plugin === '@nx/vite/plugin'
-      );
+      const vitePlugin = nxJson.plugins.find((p) => p.plugin === '@nx/vitest');
```
Coly010, line 308: This is only using `unitTestRunner=vitest` while the bundler is `webpack`. It should only have the test target

packages/vite/migrations.json
```diff
@@ -10,11 +10,6 @@
       "description": "Add gitignore entry for temporary vite config files and remove previous incorrect glob.",
       "implementation": "./src/migrations/update-20-0-4/add-vite-temp-files-to-git-ignore"
     },
-    "update-20-3-0": {
-      "version": "20.3.0-beta.2",
-      "description": "Add gitignore entry for temporary vitest config files.",
-      "implementation": "./src/migrations/update-20-3-0/add-vitest-temp-files-to-git-ignore"
-    },
```
Coly010, line 17: Old migration that is not needed anymore, but was causing issues after vitest was split. This migration has been duplicated in the `vitest` package such that it's not completely lost.

packages/vitest/README.md
```diff
@@ -0,0 +1,71 @@
+# @nx/vitest
+
+The Nx Plugin for Vitest contains executors and generators for managing Vitest tests within your Nx workspace. Vitest is a fast, modern unit test framework for JavaScript and TypeScript projects.
+
```
Coly010: This needs updated

e2e/vite/src/vite.test.ts
jaysoo: Aside: This is existing logic, but I just realize how weird it is to have `vite.test.ts` but the bundler is webpack.

packages/vitest/src/executors/test/schema.json
```diff
@@ -0,0 +1,36 @@
...
+      "type": "array",
+      "items": { "type": "string" }
+    },
+    "watch": {
+      "description": "Watch files for changes and rerun tests related to changed files.",
+      "type": "boolean"
+    }
+  },
+  "required": [],
+  "examplesFile": "../../../docs/test-examples.md"
```
jaysoo: This doesn't exist.

packages/vite/docs/test-examples.md
```diff
@@ -3,6 +3,12 @@ title: Examples for the Vitest executor
 description: This page contains examples for the Vitest @nx/vite:test executor.
 ---
 
+{% callout type="warning" title="Deprecated" %}
+The `@nx/vite:test` executor is deprecated. Please use `@nx/vitest:test` instead. This executor will be removed in Nx 23.
+
+See the [migration guide](/packages/vitest/documents/migrating-from-nx-vite) for more information.
```
jaysoo: It's better to use an actual guide rather than API docs. e.g. `/docs/technologies/testingtools/vitest/guides/migrating-from-nx-vite`

The API docs setup is fine for generators/executors/migrations examples, but let's avoid it for other use cases.

packages/remix/src/generators/application/application.impl.ts
```diff
@@ -325,7 +329,7 @@ export default {...nxPreset};
 
       updateJestTestMatch(
         tree,
-        'jest.config.cts',
+        'jest.config.ts',
```
jaysoo: I think this will fail with `jest.config.cts` that just merged. Should check for existence of `jest.config.cts` or `jest.config.ts` and update the one that exists.

jaysoo (approved):

jaysoo:
Just need to add `vitest` to `astro-docs/sidebar.mts` after `storybook` then LGTM.

jaysoo:
LGTM!

packages/vitest/generators.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Nx Vitest",
+  "version": "0.1",
+  "generators": {
+    "init": {
+      "factory": "./src/generators/init/init",
+      "schema": "./src/generators/init/schema.json",
```
FrozenPandaz: Can you switch this project to build to `packages/vite/dist` so it's easier to link? 🙏

FrozenPandaz (approved):

PR #33315, opened 2025-10-30
feat(maven): upgrade to version 0.0.8 with automated migration

FrozenPandaz:
## Current Behavior

The Maven plugin is at version 0.0.7 with no automated migration path for users to upgrade their pom.xml files.

## Expected Behavior

Users can upgrade to Maven plugin 0.0.8 and have their pom.xml files automatically updated via the Nx migration system.

## Changes Made

- Updated Maven plugin version from 0.0.7 to 0.0.8 in:
  - `packages/maven/package.json`
  - `packages/maven/src/utils/versions.ts`
  - `packages/maven/maven-plugin/pom.xml`

- Created `updateNxMavenPluginVersion()` utility function with:
  - Proper XML parsing using `@xmldom/xmldom` DOM API
  - Targeted updates only for `dev.nx.maven:nx-maven-plugin` elements
  - Safe handling of all other version elements (dependencies, parent, project, etc.)
  - Comprehensive error handling

- Added migration `0-0-8/update-pom-xml-version.ts` that:
  - Automatically runs when users upgrade to 0.0.8
  - Updates root pom.xml files via the utility function
  - Logs status of migrations performed

- Registered migration in `migrations.json`

- Added comprehensive unit tests (9 test cases):
  - Updates only the nx-maven-plugin version
  - Does not update other plugin versions
  - Handles missing files gracefully
  - Preserves XML formatting and structure
  - Handles multiple plugin references
  - Handles whitespace correctly
  - Only updates plugins, not parent/project versions
  - All tests passing (27/27)

## Test Plan

- [x] Unit tests pass (27 passing tests)
- [x] XML parsing correctly identifies and updates only `dev.nx.maven:nx-maven-plugin`
- [x] Other version elements remain untouched
- [x] Migration registration validated

### DIFF
--- a/packages/maven/src/utils/pom-xml-updater.ts
+++ b/packages/maven/src/utils/pom-xml-updater.ts
@@ -1,0 +1,54 @@
+import { Tree } from '@nx/devkit';
+import { DOMParser, XMLSerializer } from '@xmldom/xmldom';
+
+/**
+ * Updates the version of dev.nx.maven:nx-maven-plugin in pom.xml files.
+ * Only updates the version element that belongs to the nx-maven-plugin.
+ *
+ * @param tree - Nx Tree instance
+ * @param version - The new version to set
+ */
+export function updateNxMavenPluginVersion(tree: Tree, version: string): void {
+  const pomPath = 'pom.xml';
+
+  if (!tree.exists(pomPath)) {
+    return;
+  }
+
+  const content = tree.read(pomPath, 'utf-8');
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(content);
+
+  let hasChanges = false;
+
+  // Find all plugin elements in the document
+  const plugins = Array.from(doc.getElementsByTagName('plugin'));
+
+  for (const plugin of plugins) {
+    // Check if this is the nx-maven-plugin by looking for groupId and artifactId
+    const groupId = plugin.getElementsByTagName('groupId')[0];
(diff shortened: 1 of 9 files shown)


packages/maven/src/utils/pom-xml-updater.ts
```diff
@@ -0,0 +1,54 @@
...
+      groupId?.textContent?.trim() === 'dev.nx.maven' &&
+      artifactId?.textContent?.trim() === 'nx-maven-plugin';
+
+    // If this is the nx-maven-plugin, update its version
+    if (isNxMavenPlugin) {
+      const versionElement = plugin.getElementsByTagName('version')[0];
+      const currentVersion = versionElement?.textContent?.trim();
+
+      if (currentVersion && currentVersion !== version) {
+        versionElement.textContent = version;
```
MaxKless, line 42: Are we ok with just always overriding with the current version? What if they have manually upgraded to a higher version and would be downgraded by a migration?

Or is that just not supported because maven plugin & nx are so tightly coupled?

MaxKless (approved):

packages/maven/src/utils/pom-xml-updater.ts
FrozenPandaz: It should be ok for now. I'm going to merge this one first and we'll see if we need to revisit it later.

Migrate also just overwrites `package.json` so the behavior is consistent.

PR #33356, opened 2025-11-03
feat(core): update formatting of agent rules documentation

Stanzilla:
Markdown documents should only have one h1 header, it is pretty unlikely that an existing agents config does not already have one, switching to h2 instead should make most linters happy.

In theory you could make this a little smarter by checking beforehand..

I am also considering adding a line akin to 

```
Never use `nx graph` command - This spawns a web server that you don't have access to. 
Use the `nx_visualize_graph` tool or other nx tools instead.
```

since mine keeps doing that. thoughts?

## Current Behavior

The header is inserted as an h1

## Expected Behavior

The header is inserted as an h2

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
+++ b/packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
@@ -399,10 +399,13 @@ function writeAgentRules(tree: Tree, path: string, writeNxCloudRules: boolean) {
   const existingNxConfiguration = existing.match(regex);
 
   if (existingNxConfiguration) {
-    // Use h1 for comparison since we're replacing existing content
+    // Check the rest of the file (outside nx block) for an h1 header
+    // to ensure only one h1 exists in the document
+    const contentWithoutNxBlock = existing.replace(regex, '');
+    const hasExternalH1 = /^# /m.test(contentWithoutNxBlock);
     const expectedRules = getAgentRulesWrapped({
       writeNxCloudRules,
-      useH1: true,
+      useH1: !hasExternalH1,
     });
     const contentOnly = (str: string) =>
       str
--- a/packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
+++ b/packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
@@ -385,7 +385,10 @@ export async function setupAiAgentsGeneratorImpl(
 function writeAgentRules(tree: Tree, path: string, writeNxCloudRules: boolean) {
   if (!tree.exists(path)) {
     // File doesn't exist - create with h1 header (standalone content)
-    const expectedRules = getAgentRulesWrapped(writeNxCloudRules, true);
+    const expectedRules = getAgentRulesWrapped({
+      writeNxCloudRules,
+      useH1: true,
+    });
     tree.write(path, expectedRules);
     return;
   }
@@ -397,7 +400,10 @@ function writeAgentRules(tree: Tree, path: string, writeNxCloudRules: boolean) {
 
   if (existingNxConfiguration) {
     // Use h1 for comparison since we're replacing existing content
-    const expectedRules = getAgentRulesWrapped(writeNxCloudRules, true);
+    const expectedRules = getAgentRulesWrapped({
+      writeNxCloudRules,
+      useH1: true,
+    });
     const contentOnly = (str: string) =>
       str
         .replace(nxRulesMarkerCommentStart, '')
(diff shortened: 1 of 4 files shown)


MaxKless:
I see what you mean... we decided for an h1 originally because then we can make sure that we're not messing with anyone's layout by nesting something where they don't want to. 
I wasn't aware of this best practice to only ever use one h1 though, what kind of markdown linter are you using? Does it have that preconfigured?
I'm open to changing this in general though.

Stanzilla:
> I see what you mean... we decided for an h1 originally because then we can make sure that we're not messing with anyone's layout by nesting something where they don't want to. I wasn't aware of this best practice to only ever use one h1 though, what kind of markdown linter are you using? Does it have that preconfigured? I'm open to changing this in general though.

I'm on https://github.com/DavidAnson/markdownlint-cli2 (or rather the tooling around it) the rule is https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md025.md

Stanzilla:
there are also spacing issues, currently it has two blank links before and after the snippet, which also triggers linters. so if we want to make it smarter it should

- check if it is the only content of the file, if yes, no space before, use an h1 and then 1 or 0 empty lines after
- if it is not the only content, use h2 with one space before and one after

Stanzilla:
I now addressed that with an additional commit so it now dynamically adjusts spacing and header

Stanzilla:
@MaxKless what do you think?

MaxKless:
Hey sorry this fell through the cracks - I like the solution more. 
The only thing that I'm worried about is if someone adds any content, the next update when using `configure-ai-agents` will be just a single `##` -> `#` change which is so trivial.

Stanzilla:
> Hey sorry this fell through the cracks - I like the solution more. The only thing that I'm worried about is if someone adds any content, the next update when using `configure-ai-agents` will be just a single `##` -> `#` change which is so trivial.

Hah okay, so how do you want to proceed here?

MaxKless:
I think I'm okay with merging this if we add one more condition: the existing text has to have a `# Header` statement. This way it wouldn't change with just any addition to `AGENTS.md` but only if the user actually adds a h1 themselves.

Also, can you switch the implementation to be an object with options instead of adding a new option? I think this will become relevant in the future when we change some things around there in the codebase.

Also rebase pls. Thanks!

Stanzilla:
> I think I'm okay with merging this if we add one more condition: the existing text has to have a `# Header` statement. This way it wouldn't change with just any addition to `AGENTS.md` but only if the user actually adds a h1 themselves.
> 
> Also, can you switch the implementation to be an object with options instead of adding a new option? I think this will become relevant in the future when we change some things around there in the codebase.
> 
> Also rebase pls. Thanks!

how does this look?

packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
```diff
@@ -374,6 +377,11 @@ function writeAgentRules(tree: Tree, path: string, writeNxCloudRules: boolean) {
   const existingNxConfiguration = existing.match(regex);
 
   if (existingNxConfiguration) {
+    // Use h1 for comparison since we're replacing existing content
```
MaxKless: I don't think this assumption is right... if we have an nx cloud block with h2 in the file, we have an existing configuration - but this comparison will compare it with the h1 version which will break things.

MaxKless: or am I misunderstanding something here? Definitely possible, please add a test to confirm that calling it in an .md file that has the h2 version this doesn't produce any changes.

packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
```diff
@@ -391,6 +399,13 @@ function writeAgentRules(tree: Tree, path: string, writeNxCloudRules: boolean) {
     const updatedContent = existing.replace(regex, expectedRules);
     tree.write(path, updatedContent);
   } else {
+    // Appending to existing content - use h2 only if the file already has an h1 header
+    // This prevents unnecessary changes when users add content without their own h1
+    const hasExistingH1 = /^# /m.test(existing);
```
MaxKless, line 429: this will also be triggered by an h1 in the existing nx part of the rules, right? so another test is needed for both cases: when running it twice on the same files, no changes should ever be made

MaxKless (changes requested):

Stanzilla:
Addressed the review comments in b3e419b1d5:

- The update path now strips the nx block from the file and checks the rest of the document for an h1. This means it correctly determines the header level based on the surrounding content rather than blindly preserving what was there before.
- If a user adds their own h1 header after initial setup, the next run will switch the nx block from h1 to h2 automatically.
- Added tests covering: h2 idempotency (with external h1), h1-to-h2 switching when user adds their own h1, and run-twice-no-changes for fresh files, files with h1, and files without h1. All 33 tests pass.

MaxKless (approved):
LGTM! Thanks for your patience and the fast feedback cycles on reviews/discussions :) appreciate your efforts

MaxKless:
can you push your branch again after a rebase? I think there have been some changes to netflify config which caused this to hang @Stanzilla

Stanzilla:
> can you push your branch again after a rebase? I think there have been some changes to netflify config which caused this to hang @Stanzilla

done!

PR #33382, opened 2025-11-05
feat(release): add option to opt-out commit scope filter

L2jLiga:
## Current Behavior

PR #32915 changed how conventional commits determine version, making them rely on commit scope:
commits with types configured to bump minor / major version bumps only patch if commit scope exists and it does not include project name

## Expected Behavior

In our project we do not use projectName as commit scope, so we would like to bring back old behavior, this can be achieved by adding option to opt-out such behavior

### DIFF
--- a/packages/nx/schemas/nx-schema.json
+++ b/packages/nx/schemas/nx-schema.json
@@ -1253,6 +1253,10 @@
     "NxReleaseConventionalCommitsConfiguration": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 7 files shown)


Vitya23:
I have the same problem. @Cammisuli please review this

Coly010 (approved):

leosvelperez (approved):

jaysoo (approved):

PR #33385, opened 2025-11-05
fix(nx-dev): fix GitHub star button styling in mobile view

jaysoo:
This PR fixes the GitHub star widget in the left sidebar such that it only shows in smaller screens (e.g. mobile). For desktop we have it in the ToC.

Also styles it to be consistent with the login and sign up buttons.

Large screen, don't show:

image

Smaller screen, show:

image

Mobile view, show:

image

Fixes DOC-325


juristr (approved):
Looks good. 
Might be a personal preference, but I'd place it below the TOC, which is where most pages have additional info/links/page edit buttons etc

Image

PR #33389, opened 2025-11-06
feat(core): add dependents property for reverse dependency execution

### LINKED ISSUE #9322: Run Nx dependency graph backwards (leaves first)
## Description

I need to run certain command for my packages in a dependency graph, but backwards. So the tree of dependencies will have commands run around the edges at the leaves first the moving to the root. 

## Motivation

To successfully remove the deployed stacks on AWS we need to destroy stacks that aren't depended on first. So if I have stacks that depend on each other like: stack1 -> stack2 -> [stack3, stack4] I need stacks 3, 4 to be destroyed first then stack2 then stack1. This is the reverse of deployment in which we want stack1 to be deployed first and so on.

seriouscoderone:
## Current Behavior

Currently, Nx only supports forward dependency execution where dependencies must be built/executed before their dependents. For example, if `app` depends on `lib`, the `build` target runs `lib:build` before `app:build`.

There's no built-in way to execute tasks in reverse dependency order (leaves-first), which is critical for infrastructure-as-code teardown workflows.

## Expected Behavior

With this PR, users can now configure targets to run in reverse dependency order using the `dependents: true` property in `TargetDependencyConfig`.

**Usage Example:**
```json
{
  "targetDefaults": {
    "destroy": {
      "dependsOn": [{"target": "destroy", "dependents": true}]
    }
  }
}
```

When destroying infrastructure, dependent resources (e.g., AWS CloudFormation stacks that import values) must be removed before their dependencies.

## Related Issue(s)

Fixes #9322

### Changes Made:
1. **Type Definition** (`workspace-json-project-json.ts`):
   - Added `dependents?: boolean` property to `TargetDependencyConfig` interface
   - Updated documentation for mutual exclusivity with `projects` and `dependencies`

2. **Configuration Utilities** (`utils.ts`):
   - Added validation to ensure `projects`, `dependencies`, and `dependents` are mutually exclusive
   - Updated `expandWildcardTargetConfiguration()` to preserve `dependents` property

3. **Task Graph Creation** (`create-task-graph.ts`):
   - Implemented `processTasksForDependents()` using the existing `reverse()` function from `project-graph/operators`
   - Leverages the reversed project graph to reuse `processTasksForDependencies()` logic

4. **Comprehensive Tests**:
   - Simple reverse dependency scenarios
   - Multiple dependents
   - Multi-level reverse dependencies
   - Mix of dependencies and dependents
   - Edge case handling (missing targets)
   - Validation tests for mutual exclusivity

### Key Design Decisions:
- **Reuses existing utilities**: Uses the `reverse()` function from devkit instead of implementing custom reverse traversal logic
- **Clean implementation**: ~23 lines of new core logic by leveraging existing infrastructure
- **Consistent API**: Follows the same pattern as `dependencies: true`

### Use Cases:
- **AWS CloudFormation/Pulumi stack destruction**: Stacks with `Fn::ImportValue` dependencies must be destroyed in reverse order
- **Database schema teardown**: Drop dependent tables before dropping referenced tables
- **Service shutdown**: Stop dependent services before shutting down core services

Co-Authored-By: Claude <noreply@anthropic.com>


seriouscoderone:
@leosvelperez It's a relatively small PR. Let me know your thoughts.

leosvelperez (changes requested):
@seriouscoderone apologies for the delay here!

Please get the PR in a working state. It's currently failing the CI checks with legit failures. I haven't reviewed too deeply yet, but some notes:

- The new property must be added to the relevant schema files: `packages/nx/schemas/nx-schema.json`, `packages/nx/schemas/project-schema.json`, and `packages/nx/schemas/workspace-schema.json`
- Tests don't seem accurate, and at least some of them (I didn't review them all) have the expectations inverted.
- There might be some an issue with the graph swap and the recursion causing a cycle. The existing tests might surface it once the type issue is fixed and the tests run.

jaysoo (commented):
This is a highly requested feature (issue #9322, 3+ years old). The implementation approach is elegant — using `reverse(this.projectGraph)` then delegating to existing traversal is clean and maintainable.

Before this can move forward:

1. **CI needs to pass** — tests are currently failing.
2. **Documentation** — a feature of this significance needs docs explaining the `dependents: true` option in `dependsOn` config, with use cases (e.g., deploying dependents after a library change).
3. **API design buy-in** — this adds a new property to the task pipeline configuration. The Nx team should weigh in on whether `dependents: true` in `dependsOn` is the right API surface (vs. a separate `dependedOnBy` property, or a target-level `runDependents` flag).

The core implementation looks solid. Looking forward to seeing this land.

seriouscoderone:
Sorry about all the thrashing and force pushing (blush)

AgentEnder:
@leosvelperez to take another look, since he left the earlier review. CI is green now, which was the first of the three things @jaysoo listed — what's left from his comment is docs for `dependents: true` and team buy-in on the API shape. He also called it a highly requested feature and #9322 is over three years old, so the hold-up here isn't interest.

<details>
<summary>

#### Triage notes — automated triage, reviewed before posting.

</summary>

> Assignee is continuity: leosvelperez left the CHANGES_REQUESTED review, and the owners table sends a re-review back to whoever did the first one.
>
> jaysoo called this a long-requested feature against #9322 and listed three blockers — CI, docs for `dependents: true`, and buy-in on the API shape. CI rollup is `SUCCESS` on the head commit, so two of the three remain.
>
> Linked to #9322.
>
> The `stale` label is left alone — it is managed by the stale workflow, and this comment resets its clock.

</details>

PR #33456, opened 2025-11-12
fix(core): remove system metrics collection and reporting

leosvelperez:
Removes system metrics collection and reporting.

### DIFF
--- a/packages/nx/src/native/metrics/types.rs
+++ b/packages/nx/src/native/metrics/types.rs
@@ -18,19 +18,6 @@ impl Default for CollectorConfig {
     }
 }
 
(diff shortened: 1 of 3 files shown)


packages/nx/src/native/metrics/types.rs
```diff
@@ -18,19 +18,6 @@ impl Default for CollectorConfig {
     }
 }
 
-/// Error types specific to metrics collection
-#[derive(Debug, thiserror::Error)]
-pub enum MetricsError {
```
leosvelperez, line 23: This was a leftover from a previous refactor

FrozenPandaz (approved):

PR #33459, opened 2025-11-12
feat(angular): support angular cli v21.0.0-rc.4

leosvelperez:
Update Angular CLI to v21.0.0-rc.4.


guysenpai:
Hello @leosvelperez is it planned to add `@angular/build` package support for `@nx/angular:extract-i18n` executor?

PR #33482, opened 2025-11-13
fix(core): filter out automated release commits in getCommitsRelevant…

laneysmith:
…ToProjects

## Current Behavior

While testing the new release flow, I noticed that some projects we're being version bumped excessively and unexpectedly by automated `chore(release): ...` commits that affected projects for various reasons.

For example, as reported in https://github.com/nrwl/nx/issues/33374, if a `chore(release): ...` commit touches a `package.json`/`CHANGELOG.md` that has since been deleted, that release commit will be returned in `getCommitsRelevantToProjects` and evaluated as a patch bump in `determineSemverChange`.

Another example I'm seeing is that if the commit updates the `package.json`/`CHANGELOG.md` for a package that also happens to be installed as a dependency in the workspace root `package.json` then all projects in the repo are affected and patch bumped. (I'm not sure if this is an anti-pattern, but to provide more context, we have an eslint config package that we version & publish, but also have installed in the repo itself using pnpm workspace protocol).

## Expected Behavior

I could be missing something, but I can't think of a valid scenario where an automated release commit containing only `package.json#version` bumps and/or `CHANGELOG.md` updates should trigger a new release for any projects.

## Related Issue(s)

I did not file an issue for this and couldn't find an existing one that matched.

Fixes #

### DIFF
--- a/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts
+++ b/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts
@@ -11,7 +11,7 @@ export async function resolveSemverSpecifierFromConventionalCommits(
   from: string,
   projectGraph: ProjectGraph,
   projectNames: string[],
-  conventionalCommitsConfig: NxReleaseConfig['conventionalCommits']
+  releaseConfig: NxReleaseConfig
 ): // Map of projectName to semver bump type
 Promise<Map<string, SemverSpecifier | null>> {
   const commits = await getGitDiff(from);
@@ -19,9 +19,13 @@ Promise<Map<string, SemverSpecifier | null>> {
   const relevantCommits = await getCommitsRelevantToProjects(
     projectGraph,
     parsedCommits,
-    projectNames
+    projectNames,
+    releaseConfig
   );
-  return determineSemverChange(relevantCommits, conventionalCommitsConfig);
+  return determineSemverChange(
+    relevantCommits,
+    releaseConfig.conventionalCommits
+  );
 }
 
 export async function resolveSemverSpecifierFromPrompt(
--- a/packages/nx/src/command-line/release/utils/shared.spec.ts
+++ b/packages/nx/src/command-line/release/utils/shared.spec.ts
@@ -13,6 +13,8 @@ jest.mock('../../../config/nx-json', () => ({
   readNxJson: jest.fn(),
 }));
 import { createVersionConfig } from './test/test-utils';
+import { createNxReleaseConfig, NxReleaseConfig } from '../config/config';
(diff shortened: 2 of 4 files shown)


Coly010 (approved):

FrozenPandaz (approved):

MaxKless (approved):

JamesHenry:
Thanks for this @laneysmith!

Please can you expand a bit more on the below?

> Another example I'm seeing is that if the commit updates the package.json/CHANGELOG.md for a package that also happens to be installed as a dependency in the workspace root package.json then all projects in the repo are affected and patch bumped. (I'm not sure if this is an anti-pattern, but to provide more context, we have an eslint config package that we version & publish, but also have installed in the repo itself using pnpm workspace protocol).

If you are using the `workspace:` protocol then there should be no diff on the root `package.json` in that case?

laneysmith:
> Thanks for this @laneysmith!
> 
> Please can you expand a bit more on the below?
> 
> > Another example I'm seeing is that if the commit updates the package.json/CHANGELOG.md for a package that also happens to be installed as a dependency in the workspace root package.json then all projects in the repo are affected and patch bumped. (I'm not sure if this is an anti-pattern, but to provide more context, we have an eslint config package that we version & publish, but also have installed in the repo itself using pnpm workspace protocol).
> 
> If you are using the `workspace:` protocol then there should be no diff on the root `package.json` in that case?

@JamesHenry Correct, there's no diff in the root `package.json` in this scenario I'm describing. What I mean is that if I have a `chore(release): ...` commit that only has a single line change to `libs/our-custom-eslint-config/package.json#version`, the `affectedGraph` returned [here](https://github.com/nrwl/nx/blob/6b3b0c679b5f03bcf4157ea6c47f0da8a48e886b/packages/nx/src/command-line/release/utils/shared.ts#L404) includes all projects nodes for the entire workspace. All nodes get added by `addAffectedNodes`, and I'm not totally following the logic there, but have confirmed this only occurs for the couple of workspace libraries that happen to be installed as `devDependencies` in the workspace root `package.json`, so I think that's why.

I don't know if the assumption here is that this particular `chore(release): ...` commit should also always have release tags for all projects, in which case it wouldn't be included in the commit range being evaluated, but that's not the case during the transition period away from legacy versioning. Also (this starts to get into the weeds of my team's bespoke continuous publishing flow) I'm experimenting with resolving a bug in our release process by putting the GitHub release tags on the commit that triggered the publishing action, which in our case is at least 1 commit before the `chore(release): ...` commit.

Coly010:
@laneysmith FYI, I pushed a commit to this to fix merge conflict. If you try to make any other changes, make sure to pull first

PR #33542, opened 2025-11-19
chore(repo): add Node 24 to e2e nightly matrix

jaysoo:
- E2E nightly tests only run on Node 20 and 22.
- E2E nightly tests now also run on Node 24, 22, and 20 for linux. For Windows and Mac run only 24.

Did a test run from this branch, and Node 24 passes where other versions pass, and fails where other versions fail. It shouldn't make any golden tests fail just due to Node 24, but it's possible that things will flake more since the matrix has expanded. We just have to make them more robust.

image

Close NXC-3491

### DIFF
--- a/.github/workflows/e2e-matrix.yml
+++ b/.github/workflows/e2e-matrix.yml
@@ -28,19 +28,20 @@ jobs:
           # - windows-latest Windows fails to build gradle wrapper which always runs when we build nx.
           ## https://staging.nx.app/runs/LgD4vxGn8w?utm_source=pull-request&utm_medium=comment
         node_version:
+          # TODO(v23): remove node 20 - EOL April 2026
           - 20
           - 22
-          # - 23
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-matrix.yml
```diff
@@ -28,19 +28,16 @@ jobs:
           # - windows-latest Windows fails to build gradle wrapper which always runs when we build nx.
           ## https://staging.nx.app/runs/LgD4vxGn8w?utm_source=pull-request&utm_medium=comment
         node_version:
+          # TODO(v23): remove node 20 - EOL April 2026
           - 20
           - 22
-          # - 23
+          - 24
```
jaysoo, line 34: We don't care about off versions like 23, just remove it rather than commenting it out.

FrozenPandaz (approved):

PR #33543, opened 2025-11-19
feat(linter): add util to load eslint rules from a directory

jaysoo:
This PR adds the ability for users to import ESLint rules from arbitrary location in the workspace rather than storing them in `tools/eslint-rules`. This is useful for monorepo not using npm/yarn/pnpm workspaces and need a mechanism to load from any custom rules location without them being installed/symlinked.

It also handles TS files automatically.

Demo: https://www.loom.com/share/3c32af4555614eeab4f81fce8db0c955

Example:

```js
import baseConfig from "../../eslint.config.mjs";
import { loadWorkspaceRules } from "@nx/eslint-plugin";

const customRules = await loadWorkspaceRules("foo/bar/eslint-rules");

export default [
  ...baseConfig,
  {
    ignores: ["**/out-tsc"],
  },
  {
    files: ["**/*.ts", "**/*.tsx", "**/*.js", "**/*.jsx"],
    plugins: {
      custom: { rules: customRules },
    },
    rules: {
      "custom/valid-command-object": "error",
    },
  },
];
```

### DIFF
--- a/packages/devkit/src/utils/config-utils.ts
+++ b/packages/devkit/src/utils/config-utils.ts
@@ -118,7 +118,7 @@ export function clearRequireCache(): void {
   }
 }
 
-export async function load(path: string): Promise<any> {
+async function load(path: string): Promise<any> {
   try {
     // Try using `require` first, which works for CJS modules.
     // Modules are CJS unless it is named `.mjs` or `package.json` sets type to "module".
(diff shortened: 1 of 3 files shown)


packages/devkit/src/utils/config-utils.ts
```diff
@@ -118,7 +118,7 @@ export function clearRequireCache(): void {
   }
 }
 
-async function load(path: string): Promise<any> {
+export async function load(path: string): Promise<any> {
```
leosvelperez: Are the changes in this file needed? I don't see any usage of them in the rest of the files. Or is that we want to export this in devkit?

leosvelperez (approved):

packages/devkit/src/utils/config-utils.ts
jaysoo: I can undo that... I had an idea earlier for `loadWorkspaceRulesSync` but abandoned it.

llwt:
Resolves #32668

PR #33550, opened 2025-11-19
fix(graph): align exclude flag with others by using findMatchingProjects

MaxKless:
## Current Behavior
- you can only exclude project names from `graph` commands
- the exclusion is also only honored in cli versions of the command (for example `--print`) but not in the web client.

## Expected Behavior
- You can use any matching logic that other commands support (like globs, paths, tags)
- exclusions are also honored in the graph client and projects aren't shown 

Example for running `npx nx graph --exclude='!tag:type:ui' `
image

### DIFF
--- a/packages/nx/src/command-line/graph/graph.ts
+++ b/packages/nx/src/command-line/graph/graph.ts
@@ -53,6 +53,7 @@ import { getAffectedGraphNodes } from '../affected/affected';
 import { readFileMapCache } from '../../project-graph/nx-deps-cache';
 import { filterUsingGlobPatterns } from '../../hasher/task-hasher';
 import { ConfigurationSourceMaps } from '../../project-graph/utils/project-configuration-utils';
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/graph/graph.ts
```diff
@@ -351,19 +352,29 @@ export async function generateGraph(
...
-    const invalidExcludes: string[] = [];
-
-    args.exclude.forEach((project) => {
-      if (!projectExists(projects, project)) {
-        invalidExcludes.push(project);
+  let excludePatterns: string[] = [];
+  if (args.exclude && args.exclude.length > 0) {
+    try {
+      // Convert projects array to the format expected by findMatchingProjects
+      const projectsMap = projects.reduce((acc, project) => {
```
nartc: pretty sure prunedGraph.nodes is technically a projectsMap so I don't think this is needed

nartc (approved):

AgentEnder (approved):

PR #33551, opened 2025-11-19
fix(core): include PNPM patches in externalDependencies hash computations

comp615:
## Current Behavior
If a workspace uses pnpm and adds a local patch to a dependency (e.g. `vitest`), this patch is not taken into account when computing that dependency's hash for purposes of determining cache changes. In practice, you could patch vitest locally, and tests would pull from the cache.

## Expected Behavior
Patches can alter behavior in the same way that updating the version could, it's just that the version is not created and the patch is applied locally.

This updates the pnpm lockfile parsing functionality to read the patches field into a map and then combine the patch hash with the integrity hash. The integrity hash ONLY represents the remote / tarball intergrity, so these have to be combined in order to create a proper key.

## Related Issue(s)
(Could not find any, but saw this in my work today)

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
+++ b/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
@@ -135,13 +135,19 @@ function matchedDependencyName(
   );
 }
 
-function createHashFromSnapshot(snapshot: PackageSnapshot) {
-  return (
+function createHashFromSnapshot(snapshot: PackageSnapshot, patchHash?: string) {
+  const baseHash =
     snapshot.resolution?.['integrity'] ||
     (snapshot.resolution?.['tarball']
       ? hashArray([snapshot.resolution['tarball']])
-      : undefined)
-  );
+      : undefined);
+
+  // If there's a patch hash, combine it with the base hash
+  if (patchHash && baseHash) {
+    return hashArray([baseHash, patchHash]);
+  }
+
+  return baseHash;
 }
 
 function isAliasVersion(depVersion: string) {
@@ -157,6 +163,18 @@ function getNodes(
 } {
   const keyMap = new Map<string, Set<ProjectGraphExternalNode>>();
   const nodes: Map<string, Map<string, ProjectGraphExternalNode>> = new Map();
+
+  // Extract patch hashes from patchedDependencies section
+  const patchHashes = new Map<string, string>();
+  if (data.patchedDependencies) {
+    for (const [pkgName, patchInfo] of Object.entries(
+      data.patchedDependencies
+    )) {
+      if (patchInfo && typeof patchInfo === 'object' && 'hash' in patchInfo) {
+        patchHashes.set(pkgName, patchInfo.hash);
+      }
+    }
+  }
 
   const maybeAliasedPackageVersions = new Map<string, string>(); // <version, alias>
 
@@ -200,7 +218,11 @@ function getNodes(
     if (!originalPackageName) {
       continue;
     }
-    const hash = createHashFromSnapshot(snapshot);
+
+    // Compute hash once, including patch hash if available
+    const patchHash = patchHashes.get(originalPackageName);
+    const hash = createHashFromSnapshot(snapshot, patchHash);
+
     // snapshot already has a name
     if (snapshot.name) {
       packageNameObj = {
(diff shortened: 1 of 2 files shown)


packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
```diff
@@ -179,7 +197,11 @@ function getNodes(
     if (!originalPackageName) {
       continue;
     }
-    const hash = createHashFromSnapshot(snapshot);
+
+    // Compute hash once, including patch hash if available
+    const patchHash = patchHashes.get(originalPackageName);
```
leosvelperez: `patchedDependencies` keys can be package names with an exact version, a version range, or just the name (https://pnpm.io/settings#patcheddependencies). Using `originalPackageName` here would handle the case where the key is just the package name, but it wouldn't handle the other cases. We should try to find the potential patch hash with the package name (what you have here) and with the unparsed `key`, which contains the version, and the format will match according to the lockfile version.

leosvelperez (changes requested):
Thanks for the contribution!

It's mostly fine, but we need to handle the different formats the keys in `patchedDependencies` can have.

packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
comp615: Ah of course, I should have checked that instead of assuming everyone did it like us.

I've updated the code to match the same priority order pnpm uses which is I think the most accurate for complex cases. It adds a bit to the code. As an alternative, if we wanted a less complex, but more conservative approach...we could also ignore the version stuff and just match any base package (essentially as before). It'd be less accurate, but might keep the codebase more comprehensible. Up to you! Should have two working options now :)

packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
```diff
@@ -759,17 +863,21 @@ function extractNameFromKey(key: string, isV5: boolean): string {
...
     if (isV5) {
       const startFrom = key.indexOf('/');
-      return key.slice(0, key.indexOf('/', startFrom + 1));
+      const endIndex = key.indexOf('/', startFrom + 1);
+      return endIndex === -1 ? key : key.slice(0, endIndex);
     } else {
       // find the position of the '@'
-      return key.slice(0, key.indexOf('@', 1));
+      const atIndex = key.indexOf('@', 1);
+      return atIndex === -1 ? key : key.slice(0, atIndex);
```
comp615: As a note, I updated this so that it works with "packageName" specifiers to return the original. I felt that made the helper a little more predictable and cleaned up my code a bit :)

packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
```diff
@@ -158,6 +221,33 @@ function getNodes(
...
+        const versionSpecifier = getVersion(specifier, packageName) || null;
+
+        patchEntries.push({
+          packageName,
+          versionSpecifier,
+          hash: patchInfo.hash,
+        });
+      }
+    }
+  }
```
leosvelperez: Let's change this to a `Map` for easier lookups and slightly better perf:
```suggestion
  const patchEntriesByPackage = new Map<
    string,
    Array<{
      versionSpecifier: string | null; // null for name-only patches
      hash: string;
    }>
  >();
  if (data.patchedDependencies) {
    for (const [specifier, patchInfo] of Object.entries(
      data.patchedDependencies
    )) {
      if (patchInfo && typeof patchInfo === 'object' && 'hash' in patchInfo) {
        const packageName = extractNameFromKey(specifier, false);
        const versionSpecifier = getVersion(specifier, packageName) || null;

        if (!patchEntriesByPackage.has(packageName)) {
          patchEntriesByPackage.set(packageName, []);
        }
        patchEntriesByPackage.get(packageName).push({
          versionSpecifier,
          hash: patchInfo.hash,
        });
      }
    }
  }
```

leosvelperez (approved):
Thanks again for your contribution! LGTM!

I went ahead with some minor tweaks (refactored the `patchEntries` array to use a `Map`, reverted some unrelated lock file changes, etc.) to speed up the feedback and get this merged.

comp615:
> Thanks again for your contribution! LGTM!
> 
> I went ahead with some minor tweaks (refactored the `patchEntries` array to use a `Map`, reverted some unrelated lock file changes, etc.) to speed up the feedback and get this merged.

Ah thanks @leosvelperez. Sorry, I have to remove the lock file stuff every time I run pnpm and forgot. Probably a repo difference or ambiguous config. Appreciate it!

PR #33562, opened 2025-11-20
fix(core): include continuous tasks in taskResults

### LINKED ISSUE #33561: postTasksExecution hook does not consistently include continuous tasks
### Current Behavior

```
  console.log(devTask);
};
```

### Expected Behavior

### GitHub Repo

_No response_

### Steps to Reproduce

3. Run a continuous task
4. Cancel/kill the task
5. Observe output

### Nx Report

```shell
Node           : 24.11.0
OS             : darwin-arm64
pnpm           : 10.18.3

Nx key licensed packages
Licensed to Block Inc..

Registered Plugins:
@nx/vite/plugin
@nx/storybook/plugin
@squareup/nx-lint-plugin
@squareup/nx-stats-plugin
Community plugins:
Local workspace plugins:
@squareup/service-i18n
@squareup/migrate
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

comp615:
## Open Questions for Reviewers

- Added logic to not hash outputs for continuous tasks. I wasn't sure what lay down that code path, but continuous tasks cannot cache, and this seemed extra safe

## Current Behavior
Continuous tasks are not included in the list of run tasks because we don't capture their results + output. This leads to an incomplete set of results

## Expected Behavior
All run tasks should be included in the task results.

## Related Issue(s)
Fixes #33561

### DIFF
--- a/packages/nx/src/tasks-runner/task-orchestrator.ts
+++ b/packages/nx/src/tasks-runner/task-orchestrator.ts
@@ -528,26 +528,165 @@ export class TaskOrchestrator {
   private async runTask(
     task: Task,
     streamOutput: boolean,
-    env: NodeJS.ProcessEnv,
+    env: { [p: string]: string | undefined; TZ?: string },
     temporaryOutputPath: string,
     pipeOutput: boolean
-  ) {
-    const runTaskStart = performance.mark('TaskOrchestrator-run-task:start');
-    try {
-      return await this.runTaskInForkedProcess(
+  ): Promise<RunningTask> {
+    const shouldPrefix =
+      streamOutput && process.env.NX_PREFIX_OUTPUT === 'true';
+    const targetConfiguration = getTargetConfigurationForTask(
+      task,
+      this.projectGraph
+    );
+    if (
+      process.env.NX_RUN_COMMANDS_DIRECTLY !== 'false' &&
+      targetConfiguration.executor === 'nx:run-commands' &&
+      !shouldPrefix
+    ) {
+      try {
+        const { schema } = getExecutorForTask(task, this.projectGraph);
+        const combinedOptions = combineOptionsForExecutor(
+          task.overrides,
+          task.target.configuration ?? targetConfiguration.defaultConfiguration,
+          targetConfiguration,
+          schema,
+          task.target.project,
+          relative(task.projectRoot ?? workspaceRoot, process.cwd()),
+          process.env.NX_VERBOSE_LOGGING === 'true'
+        );
+        if (combinedOptions.env) {
+          env = {
+            ...env,
+            ...combinedOptions.env,
+          };
+        }
+        if (streamOutput) {
+          const args = getPrintableCommandArgsForTask(task);
+          output.logCommand(args.join(' '));
+        }
+        const runCommandsOptions = {
+          ...combinedOptions,
+          env,
+          usePty:
+            this.tuiEnabled ||
+            (!this.tasksSchedule.hasTasks() &&
+              this.runningContinuousTasks.size === 0),
+          streamOutput,
+        };
+
+        const runningTask = await runCommands(
+          runCommandsOptions,
+          {
+            root: workspaceRoot, // only root is needed in runCommands
+          } as any,
+          task.id
+        );
+
+        this.runningRunCommandsTasks.set(task.id, runningTask);
+        runningTask.onExit(() => {
+          this.runningRunCommandsTasks.delete(task.id);
+        });
+
+        if (this.tuiEnabled) {
+          if (runningTask instanceof PseudoTtyProcess) {
+            // This is an external of a the pseudo terminal where a task is running and can be passed to the TUI
+            this.options.lifeCycle.registerRunningTask(
+              task.id,
+              runningTask.getParserAndWriter()
+            );
+            runningTask.onOutput((output) => {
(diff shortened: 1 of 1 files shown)


danbucholtz (approved):

comp615:
In some scenarios this appears to still not fully capture: dev depends on watch-deps...it now gets watch-deps, but not dev

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -732,14 +732,19 @@ export class TaskOrchestrator {
...
+      runningTask.onExit(async (code) => {
         if (this.tuiEnabled) {
           this.options.lifeCycle.setTaskStatus(
             task.id,
             NativeTaskStatus.Stopped
           );
         }
         this.runningContinuousTasks.delete(task.id);
+
+        const status = code === 0 ? 'success' : 'failure';
```
leosvelperez: An unexpected exit (`!this.cleaningUp`) should always be considered a failure. Continuous tasks are intended to run continuously until all dependent tasks have finished.

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -802,13 +807,16 @@ export class TaskOrchestrator {
...
-    childProcess.onExit(() => {
+    childProcess.onExit(async (code) => {
       if (this.tuiEnabled) {
         this.options.lifeCycle.setTaskStatus(task.id, NativeTaskStatus.Stopped);
       }
       if (this.runningContinuousTasks.delete(task.id)) {
         this.runningTasksService.removeRunningTask(task.id);
       }
+
+      const status = code === 0 ? 'success' : 'failure';
```
leosvelperez: An unexpected exit (`!this.cleaningUp`) should always be considered a failure. Continuous tasks are intended to run continuously until all dependent tasks have finished.

leosvelperez (changes requested):
Thanks for the contribution! Apologies for the delay.

There have been some changes in the orchestrator since this PR was created. Please rebase and reconsider the needed changes based on the updated task orchestrator.

Some notes:

>Now when you ctrl-c a continuous task, it gets marked as failure (because it's a non-zero exit code). I'm not sure it'd ever be possible to have a successful continuous task. So that's something to think about.

When a continuous task exits unexpectedly (not as part of cleaning them up), they must be considered failures. Continuous tasks are designed to run continuously and only stop when all dependent tasks have finished.

"Stopped" continuous tasks can be considered successful. They continuously ran and were only stopped when the orchestrator decided to do so. Unexpected exits, on the other hand, mean that the task was stopped either on its own or by a user's action, and not by the orchestrator when all its dependent tasks had finished.

packages/nx/src/tasks-runner/task-orchestrator.ts
comp615: I've updated the logic to leverage the new this.cleaningUp you added. I think it should be nice and clean now

comp615:
@leosvelperez I've rebased and reupdated. I think it meshes nicely with the new boolean you added.

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -749,8 +749,11 @@ export class TaskOrchestrator {
           console.error(
             `Task "${task.id}" is continuous but exited with code ${code}`
           );
-          this.complete([{ taskId: task.id, status: 'failure' }]);
         }
+        const status = this.cleaningUp ? 'success' : 'failure';
+        await this.postRunSteps([task], [{ task, status }], false, {
```
comp615, line 767: Just a note that `postRunSteps` is a super-set of `this.complete` and does ultimately [still run it](https://github.com/nrwl/nx/blob/b584d735c972adf73389732ee6b71c79103d4937/packages/nx/src/tasks-runner/task-orchestrator.ts#L941), just with lifecycle hooks around it

comp615:
@leosvelperez are we good to go here?

leosvelperez:
@comp615, sorry for the delay, and thanks for keeping this updated.

I'm currently looking at another issue related to the handling of continuous tasks, and I'm taking the chance to implement a more comprehensive fix that would improve overall handling, including changes to ensure continuous tasks are correctly reported. The changes restructure how we complete tasks a bit and wouldn't require invoking `postRunSteps` for continuous tasks, so it will most likely supersede this one.

comp615:
@leosvelperez I did another in-depth pass today on this too (hope we're not stepping on toes) and I think I have a more root cause solution: https://github.com/nrwl/nx/pull/34132

See what you can see in that PR though. I think there's some wonkiness with just how the signals and interrupts are getting passed around. So avoiding a more comprehensive refactor there which you might be doing this hook seems like it fixes things in a lower risk way

leosvelperez:
@comp615 thanks for your work on these issues and the investigation. As mentioned, I'm fixing this a bit more comprehensively, and it fixes that as well. It's not like a massive refactor, but it does change some bits to make the flow more robust and properly handle things like the summary. It takes a bit of testing and making sure no regressions are introduced, but once I have a PR open, I'll update here.

leosvelperez:
@comp615 thanks again for your work here and investigations! This will be superseded by https://github.com/nrwl/nx/pull/34226.

PR #33572, opened 2025-11-21
fix(core): run post-task lifecycle for run-command parallel exections

### LINKED ISSUE #33586: Differing behavior for continuous tasks with and without nx daemon
### Current Behavior

I have a continuous task I run.

When I run `NX_DAEMON=false nx dev <my-project>`, an nx plugin can receive successfully receive the `postTasksExecution` hook call from the @nx/devkit package.

When I run `nx dev <my-project>` with the daemon enabled, the plugin never receives the `postTasksExecution` hook call.

### Expected Behavior

I would expect identical behaviors in this regard with and without the daemon.

### GitHub Repo

https://github.com/danbucholtz/nx-plugin-post-task-bug-repro

### Steps to Reproduce

1. Run `NX_DAEMON=false nx dev nx dev my-vite-lib` and observe a series of `postTasksExecution` related messages being logged.
2. Run `nx dev nx dev my-vite-lib` and **don't** observe a series of `postTasksExecution` related messages being logged.

### Nx Report

```shell
Node           : 24.11.1
OS             : darwin-arm64
pnpm           : 10.7.1

nx          : 22.1.1
typescript  : 5.9.3
Registered Plugins:
./libs/nx-log-plugin/src/index.js
```

### Failure Logs

```shell

```

### Package Manager Version

pnpm

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

comp615:
## Current Behavior
When using a plugin with `postTasksExecution`, this is never called in some complex scenarios with continuous tasks. 

For instance, we have a task `dev` which runs the server, and then depends on `watch-deps` (both continuous). In these scenarios, we never get stats output when the NX run is killed. However, if we remove the watch-deps it seems to work.

This seems to happen because tasks which have a slow cleanup can cause the overall NX task to finish and thus report results before they are actually killed:
```
Time →

SIGINT received
    │
    ├─► ForkedProcessTaskRunner: cleanup() + process.exit() scheduled
    │
    └─► TaskOrchestrator.cleanup() starts
            │
            ├─► cleaningUp = true
            │
            ├─► kill(prerequisite task)  ──────► Process dies quickly
            │                                         │
            │                                         └─► onExit fires
            │                                               │
            │                                               └─► postRunSteps() ✅
            │
            └─► kill(initiating task)  ────────► Process still dying...
                                                      │
        Promise.all resolves                          │ (takes longer)
        run() returns                                 │
        endCommand() writes stats  ◄──────────────────┼── Stats written HERE
                                                      │
                                                      └─► onExit fires (too late)
                                                            │
                                                            └─► postRunSteps() ❌ missed
```

In short the kill command is not awaited and even though we have async handling for onExit, that does nothing since it's a fire and forget .

## Expected Behavior
All invocations of NX which run tasks should ultimately output task data.

NOTE: This changes the output in non-TUI mode to include the second command which was lost before:
```
 NX   Successfully ran target build-deps for project mylib and 56 tasks it depends on

Nx read the output from the cache instead of running the command for 56 out of 57 tasks.

Licensed to Block Inc..
^CWarning: command "pnpm exec nx watch --projects mylib --includeDependentProjects -- pnpm exec nx build-deps mylib" exited with non-zero status code
———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Successfully ran target dev for project banking-mfe and 58 tasks it depends on

Nx read the output from the cache instead of running the command for 56 out of 59 tasks.

Licensed to Block Inc..

```

## Related Issue(s)
This is related to https://github.com/nrwl/nx/pull/33562, but potentially upstream of that as well. I've tried to test the changes together and ensure that not only do we get invocations, but we see the continuous tasks as well

Closes #33586

### DIFF
--- a/packages/nx/src/executors/run-commands/running-tasks.ts
+++ b/packages/nx/src/executors/run-commands/running-tasks.ts
@@ -520,7 +520,7 @@ class RunningNodeProcess implements RunningTask {
     process.on('SIGINT', () => {
       this.kill('SIGTERM');
       // we exit here because we don't need to write anything to cache.
-      process.exit(signalToCode('SIGINT'));
+      process.exitCode = signalToCode('SIGINT');
     });
     process.on('SIGTERM', () => {
       this.kill('SIGTERM');
@@ -702,7 +702,7 @@ function registerProcessListener(
   process.on('SIGINT', () => {
     runningTask.kill('SIGTERM');
     // we exit here because we don't need to write anything to cache.
-    process.exit(signalToCode('SIGINT'));
+    process.exitCode = signalToCode('SIGINT');
   });
   process.on('SIGTERM', () => {
     runningTask.kill('SIGTERM');
--- a/packages/nx/src/tasks-runner/forked-process-task-runner.ts
+++ b/packages/nx/src/tasks-runner/forked-process-task-runner.ts
@@ -441,7 +441,7 @@ export class ForkedProcessTaskRunner {
       this.cleanup('SIGTERM');
       process.off('message', messageHandler);
       // we exit here because we don't need to write anything to cache.
-      process.exit(signalToCode('SIGINT'));
+      process.exitCode = signalToCode('SIGINT');
     });
     process.once('SIGTERM', () => {
       this.cleanup('SIGTERM');
--- a/packages/nx/src/tasks-runner/task-orchestrator.ts
+++ b/packages/nx/src/tasks-runner/task-orchestrator.ts
@@ -761,7 +761,7 @@ export class TaskOrchestrator {
           runningTask.onExit((code) => {
             if (!this.tuiEnabled) {
               if (code > 128) {
-                process.exit(code);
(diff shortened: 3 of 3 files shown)


comp615:
@leosvelperez this is related to the other PR #33562. That one fixes a single continuous task. This one fixes when there's multiple continuous tasks (e.g. `dev` + `watch-deps`). This one I'm less confident in as it goes even further outside of the orchestrator. If you have revisions or thoughts here, I'm happy to jump on a call and pair...or just feel free to take the problem statement and run with it. But would love your deeper understanding of the forking management.

It also seems like it's probably generally not right?... It does make it work, but breaks tests perhaps. And the AI suggested fix unsolves the problem.

comp615:
Closing since this feels quite wrong and was superseded.

PR #33612, opened 2025-11-25
feat(expo): support Expo 54

Coly010:
## Current Behavior
We only support Expo 53

## Expected Behavior
Add support for Expo 54
Allow existing workspaces wishing to remain on Expo 53 to continue to be supported
Add migrations allowing LLMs to handle migrating from Expo 53 to Expo 54

## Related Issue(s)

Closes NXC-3526


packages/expo/src/migrations/update-22-0-0/update-jest-for-expo-54.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (
+      !jestConfigContent.includes('jest-expo') ||
+      !jestConfigContent.includes('jest.resolver.js')
+    ) {
+      continue;
+    }
+
+    // Remove resolver line from jest.config
+    const updatedJestConfig = jestConfigContent.replace(
+      /\s*resolver:\s*require\.resolve\(['"]\.\/jest\.resolver\.js['"]\),?\n?/,
```
jaysoo: It should be easy to do this with AST as well.

packages/expo/src/migrations/update-22-2-0/files/ai-instructions-for-expo-54.md
```diff
@@ -0,0 +1,495 @@
+# Expo SDK 54 Migration Instructions for LLM
```
jaysoo, line 1: Not sure if we should mention something about Detox being unsupported? Like if they are using that in the workpace we may need to confirm with the user that they want to continue the migration knowing that it wouldn't work?

jaysoo (approved):

packages/expo/migrations.json
```diff
@@ -169,6 +190,94 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "22.0.0": {
```
jaysoo: Should be 22.2, right?

PR #33623, opened 2025-11-26
fix(core): optimize bun lockfile parser

berenddeboer:
Key optimisations:

1. Pre-computed PackageIndex - Built once during lockfile parsing:
   - byName: Map from package name → array of versions (O(1) lookup)
   - workspaceNames: Set of workspace package names (O(1) lookup)
   - workspacePaths: Set of workspace paths (O(1) lookup)
   - packagesWithWorkspaceVariants: Set of packages with workspace-specific variants (O(1) lookup)
   - patchedPackages: Set of patched package names (O(1) lookup)
2. findResolvedVersion: Changed from O(n) scan through all packages to O(1) map lookup + O(k) where k = number of versions for that package (typically 1-3)
3. isWorkspacePackage: Changed from O(n) scan to O(1) set lookup
4. hasWorkspaceSpecificVariant: Changed from O(n) scan to O(1) set lookup
5. isNestedPackageKey: Now uses pre-computed workspace paths/names sets instead of computing them each call

On my 40 project typescript monorepo my time goes from about 30 seconds to 4.5s, a speed-up of 6-7x.


yharaskrik:
🥺

leosvelperez (approved):
LGTM! 🚀 

Thanks for the contribution!

PR #33657, opened 2025-11-28
fix(testing): remove --no-experimental-strip-types flag from @nx/jest/plugin + migrate to jest.config.cts if needed

jaysoo:
## Current Behavior
On Node.js v24+, the `@nx/jest/plugin` sets `--no-experimental-strip-types` in NODE_OPTIONS which causes an error: "node: --no-experimental-strip-types is not allowed in NODE_OPTIONS".

Additionally, `jest.config.ts` files using ESM syntax (`export default`, `import`) fail to load correctly under Node.js type-stripping when the project is configured for CommonJS.

## Expected Behavior
- Remove the NODE_OPTIONS manipulation that adds `--no-experimental-strip-types`
- Add a migration (22.2.0-beta.2) that converts `jest.config.ts` files from ESM to CJS syntax for projects using `@nx/jest/plugin`
- The migration only runs when `@nx/jest/plugin` is registered in `nx.json`
- Projects with `type: module` are warned as they're incompatible with the plugin
- Files using ESM-only features (import.meta, top-level await) are skipped with a warning for manual conversion

## Demo

https://www.loom.com/share/8a157a0b01d144ae8d6ae48b9b0cd0e4

## Related Issue(s)
Closes NXC-3541

### DIFF
--- a/packages/jest/src/migrations/update-22-2-0/convert-jest-config-to-cjs.ts
+++ b/packages/jest/src/migrations/update-22-2-0/convert-jest-config-to-cjs.ts
@@ -49,9 +49,7 @@ export default async function convertJestConfigToCjs(tree: Tree) {
       '@nx/jest/plugin',
       configPath
     );
-    if (!pluginRegistration) {
-      continue;
-    }
+    if (!pluginRegistration) continue;
 
     const projectRoot = dirname(configPath);
     const packageJsonPath = joinPathFragments(projectRoot, 'package.json');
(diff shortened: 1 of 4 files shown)


packages/jest/src/migrations/update-22-2-0/convert-jest-config-to-cjs.ts
```diff
@@ -0,0 +1,270 @@
...
+    let parent = awaitExpr.parent;
+    let isInsideFunction = false;
+
+    while (parent) {
+      if (
+        parent.kind === require('typescript').SyntaxKind.FunctionDeclaration ||
+        parent.kind === require('typescript').SyntaxKind.FunctionExpression ||
+        parent.kind === require('typescript').SyntaxKind.ArrowFunction ||
+        parent.kind === require('typescript').SyntaxKind.MethodDeclaration
+      ) {
```
leosvelperez: nit: we could use `ts.isFunctionLike(parent)`

leosvelperez (approved):

PR #33660, opened 2025-11-28
fix(core): hash git-ignored files when explicitly listed as inputs

ZackDeRose:
## Current Behavior

When a file is listed as an input for a task but is git-ignored (e.g., `.env`, `secrets.json`), its contents are not included in the hash computation. This causes incorrect cache hits when these git-ignored input files change.

**Example:**
```json
{
  "targets": {
    "build": {
      "inputs": ["{projectRoot}/.env"],
      "cache": true
    }
  }
}
```

If `.env` is in `.gitignore`, changes to `.env` do NOT invalidate the cache (bug).

## Expected Behavior

Files explicitly listed as inputs should be hashed regardless of git-ignore status. When a git-ignored input file changes, the cache should be invalidated.

### Core Fix (Rust)
- Modified `hash_project_files()` in `packages/nx/src/native/tasks/hashers/hash_project_files.rs`
  - Added `hash_missing_files()` helper that walks the filesystem without git-ignore
  - Finds files matching input patterns that were filtered out by git-ignore
  - Hashes these files directly from the filesystem

- Modified `hash_workspace_files()` in `packages/nx/src/native/tasks/hashers/hash_workspace_files.rs`
  - Same approach for workspace-level inputs

### Tests
- Added Rust unit test: `should_hash_git_ignored_files_when_explicitly_specified`
- Added e2e test in `cache.test.ts`: `should invalidate cache when git-ignored input file changes`

## Important Note for Reviewers ⚠️

This implementation hashes **all git-ignored files that match the input patterns**, including broad wildcards like `{projectRoot}/**/*`.

**Potential concerns:**
1. **Performance**: Walks the directory twice (once with gitignore, once without)
2. **Broad patterns**: If someone uses `{projectRoot}/**/*` (the default), ALL git-ignored files will now be hashed

**Should we refine this to:**
- Only hash git-ignored files for explicit patterns (no wildcards)?  
- Add caching to avoid double directory walking?
- Optimize walk scope to specific subdirectories?

Would appreciate team feedback on whether the current approach is acceptable or if we should implement the refinements mentioned above.

## Test Plan

- [x] Rust unit test passes
- [x] E2e test added (demonstrates the fix)
- [ ] Run full e2e test suite
- [ ] Performance testing with large monorepos

### DIFF
--- a/e2e/nx/src/cache.test.ts
+++ b/e2e/nx/src/cache.test.ts
@@ -599,6 +599,47 @@ describe('cache', () => {
     );
   }
 
+  it('should invalidate cache when git-ignored input file changes', async () => {
+    const projectName = uniq('myapp');
+
+    // Create a project with a target that uses a git-ignored file as input
+    updateFile(
+      `projects/${projectName}/project.json`,
+      JSON.stringify({
+        name: projectName,
+        targets: {
+          build: {
+            command: 'echo "building"',
+            inputs: ['{projectRoot}/.env'],
+            cache: true,
+          },
+        },
+      })
+    );
+
+    // Create the .env file
+    updateFile(`projects/${projectName}/.env`, 'SECRET=initial-value');
+
+    // Add .env to gitignore
+    updateFile('.gitignore', (content) => `${content}\n.env\n`);
+
+    // First run - should not be cached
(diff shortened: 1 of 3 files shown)


JamesHenry (changes requested):
Pending discussion

PR #33669, opened 2025-12-01
feat(vite): add migration to add @nx/vitest

Coly010:
## Current Behavior
We don't have a migration to migrate users to use the new @nx/vitest package if they're currently using @nx/vite
Given that we have Vitest-related operations marked as deprecated when used via @nx/vite, we should have a migration.

## Expected Behavior
Add a migration that:
1. Installs @nx/vitest
2. Switches @nx/vite:test executor usage to use @nx/vitest:test executor usage
3. Splits `@nx/vite/plugin` in nx.json that sets up vitest test targets to use `@nx/vitest` instead

### DIFF
--- a/packages/vite/src/migrations/update-22-2-0/migrate-vitest-to-vitest-package.ts
+++ b/packages/vite/src/migrations/update-22-2-0/migrate-vitest-to-vitest-package.ts
@@ -1,0 +1,193 @@
+import {
+  addDependenciesToPackageJson,
+  formatFiles,
+  type GeneratorCallback,
+  readJson,
+  readNxJson,
+  readProjectConfiguration,
+  type Tree,
+  updateNxJson,
+  updateProjectConfiguration,
+} from '@nx/devkit';
+import { forEachExecutorOptions } from '@nx/devkit/src/generators/executor-options-utils';
+import { nxVersion } from '../../utils/versions';
+
+interface ViteTestExecutorOptions {
+  configFile?: string;
+  reportsDirectory?: string;
+  mode?: string;
+  testFiles?: string[];
+  watch?: boolean;
+}
+
+/**
+ * Migrates Vitest usage from @nx/vite to @nx/vitest package.
+ *
+ * This migration:
+ * 1. Installs @nx/vitest package if not present
+ * 2. Converts @nx/vite:test executor usages to @nx/vitest:test
+ * 3. Splits @nx/vite/plugin configurations to add @nx/vitest plugin
+ * 4. Migrates targetDefaults from @nx/vite:test to @nx/vitest:test
+ */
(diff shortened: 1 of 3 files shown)


packages/vite/src/migrations/update-22-2-0/migrate-vitest-to-vitest-package.ts
```diff
@@ -0,0 +1,193 @@
...
+    newPlugins.push(vitestPlugin);
+  }
+
+  nxJson.plugins = newPlugins;
+  updateNxJson(tree, nxJson);
+}
+
+function migrateTargetDefaults(tree: Tree): void {
+  const nxJson = readNxJson(tree);
+  if (!nxJson?.targetDefaults?.['@nx/vite:test']) {
```
leosvelperez: This is what we generate, but folks could also have `nxJson.targetDefaults.[<some target name, e.g. 'test'>].executor === '@nx/vite:test'`. It wouldn't hurt to check both to handle potential manual setups.

packages/vite/src/migrations/update-22-2-0/migrate-vitest-to-vitest-package.ts
```diff
@@ -0,0 +1,193 @@
...
+      continue;
+    }
+
+    const options = (plugin.options as Record<string, unknown>) || {};
+    const { testTargetName, ciTargetName, ciGroupName, ...viteOptions } =
+      options;
+
+    // Check if this plugin has test-related options
+    if (testTargetName || ciTargetName || ciGroupName) {
+      // Only collect test options once (from first entry found)
```
leosvelperez: What would happen with multiple plugin registrations with different `include`/`exclude` and options? If I'm reading this correctly, we would strip the test options from the second registration, but we wouldn't have a corresponding `@nx/vitest` registration with those options for a different set of `include`/`exclude`.

Coly010: Good point 🤔 I didn’t consider the implication of differing options, just collected the included and excluded

packages/vite/src/migrations/update-22-2-0/migrate-vitest-to-vitest-package.ts
Coly010: Thinking more on this, it'll be fine. We're not changing the target name. Only executor name.

Coly010: further discussion has shown that it could be configured as:

```
targetDefaults: {
  "test": {
    "executor": "@nx/vite:test"
  }
}
```

leosvelperez (approved):

PR #33691, opened 2025-12-03
fix(core): exit with error when generator prompts fail

mpsanchis:
## Current Behavior

If there are issues with values passed to generators via prompt, we still see a green output and exit code 0.

A colleague found out about this by pressing `Ctrl+C` when being prompted for parameters for a generator, and this led to the CLI simply continuing execution and showing no issue.

## Expected Behavior

CLI fails.

### DIFF
--- a/packages/nx/src/utils/params.ts
+++ b/packages/nx/src/utils/params.ts
@@ -923,7 +923,7 @@ async function promptForValues(
     .then((values) => ({ ...opts, ...values }))
     .catch((e) => {
       console.error(e);
-      process.exit(0);
+      process.exit(1);
     });
 }
 


packages/nx/src/utils/params.ts
```diff
@@ -922,8 +922,10 @@ async function promptForValues(
     .prompt(getPromptsForSchema(opts, schema, projectsConfigurations))
     .then((values) => ({ ...opts, ...values }))
     .catch((e) => {
-      console.error(e);
-      process.exit(0);
+      console.error(
+        `Could not properly parse prompts. Error:\n${e.message || e}`
+      );
+      process.exit(1);
```
leosvelperez: We don't know for certain that it would be a parsing error. It could be that, or it could also be an `enquirer` specific error. Let's keep the previous, more generic error logging, and only change the exit code.
```suggestion
      console.error(e);
      process.exit(1);
```

leosvelperez (changes requested):

packages/nx/src/utils/params.ts
mpsanchis: Thanks @leosvelperez 🙏  Change added

leosvelperez (approved):
Thanks!

PR #33699, opened 2025-12-03
fix(core): cnw sends correct selectedRepositoryName; prints instructions when user Ctrl+C

jaysoo:
This PR adds `SIGINT` handling when user kills the process via `Ctrl+C` during CNW. This only prints when the workspace setup is complete, and we also print the Cloud onboarding URL if it has been set up.

Also fixes an issue where `selectedRepositoryName` is never sent during CNW.

### DIFF
--- a/packages/create-nx-workspace/src/create-workspace.ts
+++ b/packages/create-nx-workspace/src/create-workspace.ts
@@ -23,6 +23,17 @@ import { Preset } from './utils/preset/preset';
 import { cloneTemplate } from './utils/template/clone-template';
 import { execAndWait } from './utils/child-process-utils';
 
+// State for SIGINT handler - only set after workspace is fully installed
+let workspaceDirectory: string | undefined;
+let cloudConnectUrl: string | undefined;
+
+export function getInterruptedWorkspaceState(): {
+  directory: string | undefined;
+  connectUrl: string | undefined;
+} {
+  return { directory: workspaceDirectory, connectUrl: cloudConnectUrl };
+}
+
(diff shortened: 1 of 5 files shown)


packages/nx/src/nx-cloud/utilities/url-shorten.ts
```diff
@@ -10,9 +10,10 @@ export async function createNxCloudOnboardingURL(
   accessToken?: string,
   meta?: string,
   forceManual = false,
-  forceGithub = false
+  forceGithub = false,
+  directory?: string
```
jaysoo, line 14: This is needed during CNW, since the `process.cwd()` is not where the Nx workspace is. During `nx connect` or `nx init` it was fine but CNW never worked to find git remote info.

packages/create-nx-workspace/src/create-workspace.ts
```diff
@@ -162,7 +166,22 @@ export async function createWorkspace<T extends CreateWorkspaceOptions>(
     }
   }
 
-  if (connectUrl) {
+  // Create onboarding URL AFTER git operations so getVcsRemoteInfo() can detect the repo
+  let connectUrl: string | undefined;
```
jaysoo, line 170: Had to move the `createNxCloudOnboardingUrl` function to after GitHub repo is (potentially) pushed. Otherwise, the remote info will never exist.

In the future we can see if passing the directory name and GitHub username (if we can get it from `gh`) are good fallback options.

FrozenPandaz (approved):

PR #33702, opened 2025-12-04
feat(core): add packageManager field when setting up new workspaces

juristr:
## Current Behavior

Currently we don't add the `packageManager` property when generating a new workspace. 

## Expected Behavior

We should have it as more and more tooling relies on it to make decisions. (e.g. on CI etc..)

## Related Issue(s)

Fixes #


JamesHenry (changes requested):
@juristr I think we'll want a custom snapshot serializer for these otherwise they will be subject to change every time our CI environment changes by even a patch version of npm

juristr:
@JamesHenry oh yeah good catch

AgentEnder:
I'm going to close this out as its sat for 6 months - we can come back to it if it feels important.

PR #33717, opened 2025-12-05
chore(core): skip connect to cloud prompt during migration if `neverConnectToCloud` is set

mpsanchis:
## Current Behavior

User is always prompted to connect to Nx cloud, unless already connected, in some scenarios (such as jumping major version).

## Expected Behavior

Under some conditions, like explicitly setting `NX_NO_CLOUD`, the prompt should not appear.

I am aware that setting `CI=true` also prevents that, but it also forces other behaviours that users might not want. When programmatically migrating repos, the best way is to simply execute a command, without having to interact with its options.

### DIFF
--- a/packages/nx/src/command-line/migrate/migrate.ts
+++ b/packages/nx/src/command-line/migrate/migrate.ts
@@ -78,7 +78,7 @@ import {
 import { readNxJson } from '../../config/configuration';
 import { runNxSync } from '../../utils/child-process';
 import { daemonClient } from '../../daemon/client/client';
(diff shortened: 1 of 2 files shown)


AgentEnder (approved):

mpsanchis:
Hi @AgentEnder when I saw the notification with your approval and force push, I thought this had been merged, but apparently not. Can I do something here? Just rebase? (doesn't even seem necessary, as GitHub mentions "changes can be cleanly merged".
Let me know, thanks!

AgentEnder:
Hey, sorry - some CI checks were failing. I rebased + formatted it so we
will see if it passes this time around.

Issue #32265, opened 2025-08-08
Slow project graph calculation with bun


rosslo:
### Current Behavior

I was previously using `nx@21.1.2` and noticed that it does not parse externalNodes when inspecting the `.nx/workspace-data/project-graph.json` file.

To address this, I attempted to upgrade to `nx@21.4.0-beta.7` to see if the new version resolves this issue.

While externalNodes parsing works correctly in `nx@21.4.0-beta.7`, I am now experiencing extremely slow project graph calculations. It takes approximately 10 minutes to complete.

When running an Nx task, I see the following log message:

```bash
⠙ Calculating the project graph on the Nx Daemon is taking longer than expected. Re-run with NX_DAEMON=false to see more details.
```

And the following is `.nx/workspace-data/d/daemon.log`:

```log
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.505Z - Started listening on: /var/folders/k3/p73n85qx2bv2qhft425zd4cc0000gn/T/794c3a57da334cf7bfaa/d.sock
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.514Z - [WATCHER]: Subscribed to changes within: /Users/user/Documents/repo (native)
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.515Z - Established a connection. Number of open connections: 1
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.516Z - Established a connection. Number of open connections: 2
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.517Z - Closed a connection. Number of open connections: 1
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.521Z - [REQUEST]: Client Request for Project Graph Received
Time for 'plugin worker 22522 code loading' 7.648125
Time for 'plugin worker 22523 code loading' 7.727083999999998
Time for 'plugin worker 22524 code loading' 26.829417
[NX v21.4.0-beta.7 Daemon Server] - 2025-08-08T08:51:25.748Z - Time taken for 'Load Nx Plugin: /Users/user/Documents/repo/node_modules/nx/src/plugins/package-json' 225.86283399999996ms
[...]

PR #33736, opened 2025-12-08
perf(core): optimize project graph cache validation

adwait1290:
## Current Behavior

The `shouldRecomputeWholeGraph` function runs on every Nx command that touches the project graph (build, test, run, etc.). The current implementation uses `JSON.stringify` for comparison:

```typescript
// Path mappings - O(n) JSON.stringify calls in a loop
Object.keys(cache.pathMappings).some((t) => {
  const cached = JSON.stringify(cache.pathMappings[t]);
  const notCached = JSON.stringify(tsConfig.compilerOptions.paths[t]);
  return cached !== notCached;
});

// Plugins - 2 JSON.stringify calls
JSON.stringify(getNxJsonPluginsData(nxJson, packageJsonDeps)) !== JSON.stringify(cache.nxJsonPlugins)

// Config - 2 JSON.stringify calls
JSON.stringify(nxJson?.pluginsConfig) !== JSON.stringify(cache.pluginsConfig)
```

Additionally, `deepClone` uses the slower `JSON.parse(JSON.stringify())` pattern.

## Expected Behavior

Use pre-computed hashes for O(1) comparison instead of O(n) JSON.stringify calls. Use `structuredClone` which is 2-3x faster than JSON round-trip for deep cloning.

### 1. Hash-based cache validation (`nx-deps-cache.ts`)

- Added optional hash fields to `FileMapCache` interface:
  - `pathMappingsHash?: string`
  - `nxJsonPluginsHash?: string`
  - `pluginsConfigHash?: string`
- Pre-compute hashes when creating cache in `createProjectFileMapCache`
- Compare hashes in `shouldRecomputeWholeGraph` instead of using JSON.stringify
- Maintain backward compatibility: legacy caches without hashes fall back to original logic

### 2. Faster deep clone (`project-configuration-utils.ts`)

- Replace `JSON.parse(JSON.stringify(obj))` with `structuredClone(obj)`
- Added TypeScript generic for type safety

## Performance Impact

| Cache validation | O(n) JSON.stringify per path | O(1) hash comparison | Significant for large monorepos |
| Deep clone | JSON round-trip | structuredClone | 2-3x faster |
| Hash computation | N/A | Uses native xxhash | Very fast |

The native hasher (`hashObject` from `hasher/file-hasher.ts`) uses Rust's xxhash implementation which is significantly faster than JavaScript's JSON.stringify.

## Backward Compatibility

- Optional hash fields ensure old caches work seamlessly
- Missing hashes trigger fallback to original JSON.stringify comparison
- Cache version remains `6.0` (no breaking change)

## Related Issue(s)

Contributes to #32265

## Testing

All 129 project-graph tests pass, including the 17 cache-specific tests.

## Merge Dependencies

**Must be merged BEFORE:** #33739, #33746

---

### DIFF
--- a/packages/nx/src/project-graph/nx-deps-cache.ts
+++ b/packages/nx/src/project-graph/nx-deps-cache.ts
@@ -17,6 +17,7 @@ import {
   writeJsonFile,
 } from '../utils/fileutils';
 import { nxVersion } from '../utils/versions';
+import { hashObject } from '../hasher/file-hasher';
 import { ConfigurationSourceMaps } from './utils/project-configuration-utils';
 import {
   ProjectGraphError,
@@ -34,6 +35,10 @@ export interface FileMapCache {
   pluginsConfig?: any;
   fileMap: FileMap;
   externalNodesHash?: string;
+  // Pre-computed hashes for fast comparison (added in 6.1)
+  pathMappingsHash?: string;
+  nxJsonPluginsHash?: string;
+  pluginsConfigHash?: string;
 }
 
 export const nxProjectGraph = join(
@@ -186,15 +191,22 @@ export function createProjectFileMapCache(
   externalNodesHash: string
 ) {
   const nxJsonPlugins = getNxJsonPluginsData(nxJson, packageJsonDeps);
+  const pathMappings = tsConfig?.compilerOptions?.paths || {};
+  const pluginsConfig = nxJson?.pluginsConfig;
+
   const newValue: FileMapCache = {
     version: '6.0',
     nxVersion: nxVersion,
     // compilerOptions may not exist, especially for package-based repos
-    pathMappings: tsConfig?.compilerOptions?.paths || {},
+    pathMappings,
     nxJsonPlugins,
-    pluginsConfig: nxJson?.pluginsConfig,
+    pluginsConfig,
     fileMap,
     externalNodesHash,
+    // Pre-compute hashes for fast comparison in shouldRecomputeWholeGraph
+    pathMappingsHash: hashObject(pathMappings),
+    nxJsonPluginsHash: hashObject(nxJsonPlugins as object),
+    pluginsConfigHash: pluginsConfig ? hashObject(pluginsConfig) : undefined,
   };
   return newValue;
 }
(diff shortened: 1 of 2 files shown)


packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -861,8 +861,8 @@ function targetDefaultShouldBeApplied(
   return !plugin?.startsWith('nx/');
 }
 
-function deepClone(obj) {
-  return JSON.parse(JSON.stringify(obj));
+function deepClone<T>(obj: T): T {
+  return structuredClone(obj);
```
JamesHenry, line 865: Looks like we're safe to use it now based on our minimum node floor

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -34,6 +35,10 @@ export interface FileMapCache {
   pluginsConfig?: any;
   fileMap: FileMap;
   externalNodesHash?: string;
+  // Pre-computed hashes for fast comparison (added in 6.1)
```
JamesHenry: 6.1 isn't referenced anywhere else, please remove

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -186,15 +191,22 @@ export function createProjectFileMapCache(
   externalNodesHash: string
 ) {
   const nxJsonPlugins = getNxJsonPluginsData(nxJson, packageJsonDeps);
+  const pathMappings = tsConfig?.compilerOptions?.paths || {};
+  const pluginsConfig = nxJson?.pluginsConfig;
+
   const newValue: FileMapCache = {
     version: '6.0',
```
JamesHenry: I'll need to defer to @FrozenPandaz on if this version should change based on this PR, but otherwise the change LGTM, thank you!

adwait1290: I've bumped the version to 6.1 since we're adding new optional fields (pathMappingsHash, nxJsonPluginsHash, pluginsConfigHash) to the FileMapCache interface. This ensures older caches
  without these fields will be invalidated. Happy to revert if @FrozenPandaz prefers otherwise.

packages/nx/src/project-graph/utils/project-configuration-utils.ts
adwait1290: Thanks for confirming! structuredClone is available in Node 17+ and our minimum is higher than that now.

packages/nx/src/project-graph/nx-deps-cache.ts
adwait1290: Done - removed the version reference from the comment.

AgentEnder:
There are conflicts and we are working on some more perf improvements, these seem relatively small in comparison so we are going to close these out. Afterwards, if you'd like to open please include information from running `nx run benchmarks`

Issue #32254, opened 2025-08-07
TypeError: Cannot destructure property 'specifiers' of 'projectSnapshot' during build – pruned lockfile error


jsdev-robin:
### Current Behavior

Hi team, I'm encountering a build-time issue related to pruned lockfile generation when running a project using Nx with pnpm and webpack-cli.

Even though the project builds and runs successfully (thanks to cache), this error consistently appears:

`NX   An error occured while creating pruned lockfile
Please open an issue at `https://github.com/nrwl/nx/issues/new?template=1-bug.yml` and provide a reproduction.
To prevent the build from breaking we are returning the root lock file.
If you run `pnpm install --lockfile-only` in your output folder it will regenerate the correct pruned lockfile.
Original error: Cannot destructure property 'specifiers' of 'projectSnapshot' as it is undefined.
TypeError: Cannot destructure property 'specifiers' of 'projectSnapshot' as it is undefined.
    at convertProjectSnapshotToInlineSpecifiersFormat (...)
`

Full stack trace included below.

Reproduction:
Nx Version: 21.3.11

pnpm Version: (please include your version here)

Node Version: (please include your version here)

OS: Windows 10

Steps:
I ran: webpack-cli build --node-env=production

Nx shows success from cache

But the pruned lockfile generation fails with the above error

### Expected Behavior

The build should not throw errors duThe build should not throw errors during pruned lockfile creation, especially when cache is used.ring pruned lockfile creation, especially when cache is used.

### Steps to Reproduce

1.

### Nx Report

```shell
> nx run @server/auth:build  [existing outputs match the cache, left as is]
> webpack-cli build --node-env=production
 NX   An error occured while creating pruned lockfile
Please open an issue at `https://github.com/nrwl/nx/issues/new?template=1-bug.yml` and provide a reproduction.
To prevent the build from breaking we are returning the root lock file.
If you run `pnpm install --lockfile-only` in your output folder it will regenerate the correct pruned lockfile.
Original error: Cannot destructure property 'specifiers' of 'projectSnapshot' as it is undefined.
TypeError: Cannot destructure property 'specifiers' of 'projectSnapshot' as it is undefined.
[...]

PR #33750, opened 2025-12-08
fix(core): optimize pnpm lockfile parsing with pre-built indexes

adwait1290:
## Current Behavior

When parsing pnpm lockfiles to build the project graph:

### matchPropValue function
- Uses separate `Object.values()` + `Object.keys()` iterations
- Two passes over the same data to find a value and get its key

### Hoisted Dependencies Lookup
- For each package, calls `Object.keys(hoistedDeps).find(k => k.startsWith(...))`
- O(n) search performed m times = O(n*m) total

### Single-pass Iteration

```
BEFORE: matchPropValue
│  const index = Object.values(record).findIndex(v === key)   │
│  if (index > -1)                                            │
│    return Object.keys(record)[index]  ← Another iteration!  │
│  = 2 iterations over the same data                          │

AFTER: Object.entries() single pass
│  for (const [name, version] of Object.entries(record)) {    │
│    if (version === key) return name  ← Early exit           │
│  }                                                          │
│  = 1 iteration with early exit on match                     │
```

### Pre-built Index for Hoisted Dependencies

```
BEFORE: O(n*m) Hoisted Lookup
│  for each package (n packages):                             │
│    Object.keys(hoistedDeps).find(k => k.startsWith(...))    │
│    ← O(m) search where m = hoisted deps                     │
│  Total: O(n * m)                                            │
│  Example: 500 packages × 200 hoisted deps                   │
│         = 100,000 string comparisons                        │

AFTER: Pre-built Index Map O(n+m)
│  Build hoistedKeysByPackage Map once:  O(m)                 │
│    Map<packageName, hoistedKey>                             │
│  for each package (n packages):                             │
│    hoistedKeysByPackage.get(packageName)  O(1)              │
│  Total: O(n + m)                                            │
│  Example: 500 packages + 200 hoisted deps                   │
│         = 700 operations (vs 100,000)                       │
```

## Impact

For large pnpm monorepos:

| matchPropValue iterations | 2 | 1 | 50% fewer iterations |
| Hoisted lookup complexity | O(n×m) | O(n+m) | ~100x for large repos |
| String comparisons | n×m | n+m | Dramatic reduction |

## Related Issue(s)

Contributes to #32669, #32254

## Merge Dependencies

**Must be merged BEFORE:** #33751

---

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
+++ b/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
@@ -367,17 +367,13 @@ function getHoistedVersion(
   hoistedDependencies: Record<string, any>,
   packageName: string,
   isV5: boolean,
-  hoistedKeysByPackage?: Map<string, string>
+  hoistedKeysByPackage: Map<string, string>
 ): string {
   let version = getHoistedPackageVersion(packageName);
 
   if (!version) {
-    // Use pre-built index for O(1) lookup if available, otherwise fallback to O(n) search
-    const key = hoistedKeysByPackage
-      ? hoistedKeysByPackage.get(packageName)
-      : Object.keys(hoistedDependencies).find((k) =>
-          k.startsWith(`/${packageName}/`)
-        );
+    // Use pre-built index for O(1) lookup
+    const key = hoistedKeysByPackage.get(packageName);
     if (key) {
       version = parseBaseVersion(getVersion(key.slice(1), packageName), isV5);
     } else {
--- a/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
+++ b/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
@@ -89,9 +89,11 @@ function matchPropValue(
   if (!record) {
     return undefined;
   }
-  const index = Object.values(record).findIndex((version) => version === key);
-  if (index > -1) {
-    return Object.keys(record)[index];
+  // Use Object.entries() for single-pass iteration instead of separate Object.values() + Object.keys()
+  for (const [name, version] of Object.entries(record)) {
+    if (version === key) {
+      return name;
+    }
   }
   // check if non-aliased name is found
   if (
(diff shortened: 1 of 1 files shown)


packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
```diff
@@ -340,14 +366,18 @@ function getNodes(
 function getHoistedVersion(
   hoistedDependencies: Record<string, any>,
   packageName: string,
-  isV5: boolean
+  isV5: boolean,
+  hoistedKeysByPackage?: Map<string, string>
```
leosvelperez: This parameter shouldn't be optional, and we can simplify its usage. This function is private and has a single usage, where the parameter is provided.

```suggestion
  hoistedKeysByPackage: Map<string, string>
```

packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
```diff
@@ -340,14 +366,18 @@ function getNodes(
...
   if (!version) {
-    const key = Object.keys(hoistedDependencies).find((k) =>
-      k.startsWith(`/${packageName}/`)
-    );
+    // Use pre-built index for O(1) lookup if available, otherwise fallback to O(n) search
+    const key = hoistedKeysByPackage
+      ? hoistedKeysByPackage.get(packageName)
+      : Object.keys(hoistedDependencies).find((k) =>
+          k.startsWith(`/${packageName}/`)
+        );
```
leosvelperez:
```suggestion
    const key = hoistedKeysByPackage.get(packageName);
```

leosvelperez (changes requested):
Thanks for contributing! Please make `hoistedKeysByPackage` a required parameter.

packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
adwait1290: Good catch! Simplified; made the parameter non-optional and removed the fallback logic since it's always provided.

leosvelperez (approved):
Thanks!

meeroslav:
The hoisted calculation improvement was a good catch. On a test repo we measured 5x improvement on that code section.

Sadly, the `matchPropValue` is now performing worse. On test repo it brought the total duration of matching versions from ~650 to 900+ milliseconds.

Here are the reasons why:
- Native methods are heavily optimized: `findIndex()` is implemented in optimized C++ code in the JavaScript engine, while `for...of` runs in the JavaScript layer with more overhead per iteration.
- `Object.entries()` is expensive: It creates an array of arrays (tuples), which involves more object allocation than `Object.values()`, which creates a flat array. Even though you call both `Object.values()` and `Object.keys()`, creating two flat arrays can be faster than creating one array of tuple objects.
- Destructuring has overhead: The `[name, version]` destructuring happens on every loop iteration, adding cost.
- Array access is extremely fast: `Object.keys(record)[index]` is a simple array index access, which is highly optimized. Modern engine makes this nearly free.
- The `record` is small, so the "double iteration" penalty is negligible due to native method optimizations.

PR #33766, opened 2025-12-08
feat(core): detect VS Code Copilot as AI agent

### LINKED ISSUE #33698: TUI Auto Exit does not work in VSCode
### Current Behavior

Even when manually specifying `--tuiAutoExit` AND with `"tui": { "autoExit": true }` in my `nx.json`, it still does not auto exit when called from Copilot Chat/ an agent.

### Expected Behavior

Either of those options should be enough and honestly, it should be the default when called from an agent.

### GitHub Repo

no repo

### Steps to Reproduce

1. tell the agent to run something like `npx nx run-many -t type-check --tuiAutoExit` and watch the TUI stay up once it is done

### Nx Report

```shell

Node           : 24.11.1
OS             : darwin-arm64
yarn           : 1.22.22

nx (global)            : 22.1.3
@nx/js                 : 22.0.2
@nx/next               : 22.0.2
@nx/webpack            : 22.0.2

Nx key licensed packages
Licensed to EFG.

@nx/gcs-cache  : 5.0.0

Registered Plugins:
@nx/next/plugin
@nx/eslint/plugin
@nx/jest/plugin
Community plugins:
@faceit/nx-helpers : 0.0.1
@nx/gcs-cache      : 5.0.0
nx-stylelint       : 18.0.2
Local workspace plugins:
@faceit/nx-helpers
Cache Usage: 75.66 MB / 92.64 GB
```

### Failure Logs

```shell

```

### Package Manager Version

yarn 1.22.22

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I raised this previousy to @juristr on Twitter and he said there might be an issue with VS Code specifically

Stanzilla:
## Current Behavior

When running Nx commands from VS Code Copilot's chat panel, Nx does not detect that it's running inside an AI agent. This means the TUI is enabled, which doesn't work well with AI agents that expect plain text output.

## Expected Behavior

Nx should detect when it's running from VS Code Copilot's chat panel and disable the TUI accordingly, similar to how it handles Claude Code, Cursor, and Repl.it.

## Related Issue(s)

Fixes #33698

## Implementation

Added detection for VS Code Copilot by checking if the `PATH` environment variable contains `github.copilot-chat`. When VS Code Copilot runs terminal commands, it adds its CLI tool path (e.g., `.../globalStorage/github.copilot-chat/copilotCli`) to the PATH.

Changes:
- Added `is_vscode_copilot()` function to detect VS Code Copilot
- Added `is_vscode_copilot_from_path()` helper for testability
- Updated `is_ai_agent()` to include VS Code Copilot detection
- Added unit tests for the new detection logic

### DIFF
--- a/packages/nx/src/native/utils/ai.rs
+++ b/packages/nx/src/native/utils/ai.rs
@@ -12,6 +12,24 @@ fn is_claude_ai() -> bool {
     }
 }
 
+/// Detects if the current process is being run by VS Code Copilot chat
+fn is_vscode_copilot() -> bool {
+    is_vscode_copilot_from_path(env::var("PATH").ok())
+}
+
+fn is_vscode_copilot_from_path(path: Option<String>) -> bool {
+    match path {
+        Some(path) => {
+            let is_copilot = path.contains("github.copilot-chat");
+            if is_copilot {
+                debug!("VS Code Copilot detected via PATH containing github.copilot-chat");
+            }
+            is_copilot
+        }
+        None => false,
+    }
+}
+
 /// Detects if the current process is being run by Repl.it
 fn is_replit_ai() -> bool {
     match env::var("REPL_ID") {
@@ -45,7 +63,7 @@ fn is_cursor_ai() -> bool {
 /// Detects if the current process is being run by an AI agent
 #[napi]
 pub fn is_ai_agent() -> bool {
-    let is_ai = is_claude_ai() || is_replit_ai() || is_cursor_ai();
+    let is_ai = is_claude_ai() || is_replit_ai() || is_cursor_ai() || is_vscode_copilot();
 
     if is_ai {
         debug!("AI agent detected");
@@ -179,5 +197,29 @@ mod tests {
                 env::set_var("COMPOSER_NO_INTERACTION", val);
             }
         }
+    }
+
+    #[test]
+    fn test_vscode_copilot_detection() {
+        // Test with no PATH
+        assert!(
+            !is_vscode_copilot_from_path(None),
+            "Should not detect VS Code Copilot without PATH"
+        );
+
+        // Test with PATH not containing copilot
+        assert!(
+            !is_vscode_copilot_from_path(Some("/usr/bin:/usr/local/bin".to_string())),
+            "Should not detect VS Code Copilot without github.copilot-chat in PATH"
+        );
+
+        // Test with PATH containing copilot-chat
+        assert!(
+            is_vscode_copilot_from_path(Some(
+                "/usr/bin:/some/path/globalStorage/github.copilot-chat/copilotCli:/usr/local/bin"
+                    .to_string()
+            )),
+            "Should detect VS Code Copilot with github.copilot-chat in PATH"
+        );
     }
 }


MaxKless (changes requested):
Hey @Stanzilla I appreciate the PR!
I went into VSCode to verify and it looks like the copilot string is also added to `PATH` in the regular vscode terminal, not just when running through copilot itself (see below).
This would mean that the TUI won't be shown in the vscode terminal which isn't what we want. So I think we'll need another approach here...

Image

Stanzilla:
Thanks for catching that @MaxKless! You're right - I just verified and the `github.copilot-chat` path is present in the regular VS Code terminal as well, not just in Copilot agent mode.

I've checked the environment more thoroughly, and unfortunately there doesn't appear to be a unique environment variable that distinguishes Copilot agent mode from regular VS Code terminal usage:
- `GIT_PAGER=cat` is set, but this might also be common in regular VS Code
- No `COPILOT_*` or similar specific env vars are set
- TTY status is the same in both cases (stdin/stdout/stderr all report as terminals)
- Parent process is the same VS Code helper

Do you have any suggestions for how we might detect Copilot agent mode specifically? Some possibilities:
1. Is there perhaps a VS Code extension API that could set a specific env var when running from agent mode?
2. Could Copilot itself set something like `GITHUB_COPILOT_AGENT=1` when invoking terminal commands?
3. Should we file an issue with the Copilot team to request an environment variable for this use case?

Alternatively, would it make sense to approach this from a different angle - perhaps detecting that the terminal is not truly interactive in some way, even if it reports as a TTY?

packages/nx/src/native/utils/ai.rs
```diff
@@ -12,6 +12,24 @@ fn is_claude_ai() -> bool {
...
 
+/// Detects if the current process is being run by VS Code Copilot chat
+fn is_vscode_copilot() -> bool {
+    is_vscode_copilot_from_path(env::var("PATH").ok())
+}
+
+fn is_vscode_copilot_from_path(path: Option<String>) -> bool {
+    match path {
+        Some(path) => {
+            let is_copilot = path.contains("github.copilot-chat");
```
Tyriar, line 23: Having this in the PATH just means that the copilot extension is installed. There's actually intentionally not a way right now to determine whether the session is being driven by AI because I worry about tools starting to behave differently if we offer that and therefore start working differently when ideally the agent would be able to handle everything.

What do you think? https://github.com/microsoft/vscode/issues/253945#issuecomment-3337637793

Do most other terminals driven by AI include environment variables now?

Tyriar (changes requested):

packages/nx/src/native/utils/ai.rs
Stanzilla: From the code present here, at least Claude, Replit and Cursor are. That's what NX detects right now to do exactly what you were worried about. But is that still a problem for you? Do you see any other good way this could be handled in tools like NX? You don't actually have access to the harness etc, the only thing you could do is try to force it via agent instructions?

MaxKless: @Tyriar thanks for chiming in here! At least there's a solid workaround which is setting `NX_TUI=false` in the terminal profile setting @Stanzilla (you could also add an instruction to your AGENTS.md or whatever to always run tasks with NX_TUI=false). 

@Tyriar I get where you're coming from with not wanting to have tools adapt too much to whether they're being driven by an agent or human. Right now TUIs kind of don't work well with the text-based approach agents have to input/output, so this being broken is a side effect of that decision. 
Other tools we looked into do have a way of telling agent invocations apart from regular ones and it's helpful for us. 
Seems like either choice will have its drawbacks - one of them immediately relevant today though while the other one is more theoretical in the future. Tricky for sure :D

Stanzilla:
I found another case where this could help, outside of the terminal, currently Agents keep trying to us `nx show` which opens a browser and then just waits for something to happen which..doesn't. So this could also be gated behind such detections.

@MaxKless thoughts?

packages/nx/src/native/utils/ai.rs
Stanzilla: @Tyriar anything I can do to make you reconsider? :)

MaxKless:
I still think it makes sense to have this but the current approach isn't the right one. 
@Tyriar any updates? The decision to not have an env var like this from the VSCode side is final?

jaysoo (changes requested):
As @MaxKless and @Tyriar identified, `github.copilot-chat` appears in `PATH` for **all** VS Code terminal sessions, not just Copilot agent mode. This would falsely disable the TUI for every VS Code user.

This needs to wait for VS Code to provide a dedicated environment variable that distinguishes agent mode from regular terminal sessions. Until then, this detection method will produce too many false positives to ship.

PR #33780, opened 2025-12-10
fix(core): invalidate sync generator cache on file changes and use up-to-date project graph

leosvelperez:
## Current Behavior

Sync generators are processed in the background by the daemon server. Their results are cached and reprocessed when the project graph is recomputed. There are currently two issues:

- The cache is only invalidated after the project graph finishes recomputing, which means that there's an interval between files changed (triggering the project graph recomputation) and the recomputation finishes, where the cache is not invalidated, and it's stale. During that interval, any request to get the sync generator changes will use the stale cache.
- Sync generators are scheduled to be processed after the project graph is recomputed, so a quick succession of recomputations can be coalesced. The problem is that the scheduled closure uses the project graph from the initial scheduling, rather than the latest available project graph at the time it runs. This results in the usage of stale data to process the sync generators.

## Expected Behavior

Getting sync generators changes should always return up-to-date information.

### DIFF
--- a/packages/nx/src/daemon/server/sync-generators.ts
+++ b/packages/nx/src/daemon/server/sync-generators.ts
@@ -38,6 +38,11 @@ const log = (...messageParts: unknown[]) => {
   serverLogger.log('[SYNC]:', ...messageParts);
 };
 
+export function clearSyncGeneratorsCache(): void {
+  log('clearing sync generators cache due to file changes');
+  syncGeneratorsCacheResultPromises.clear();
+}
+
 export async function getCachedSyncGeneratorChanges(
   generators: string[]
 ): Promise<SyncGeneratorRunResult[]> {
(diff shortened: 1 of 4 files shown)


packages/nx/src/daemon/server/sync-generators.ts
```diff
@@ -136,8 +141,16 @@ export function collectAndScheduleSyncGenerators(
       return;
     }
 
-    const { projects } =
-      readProjectsConfigurationFromProjectGraph(projectGraph);
+    // Fetch the latest graph instead of using the captured one which might be stale
+    const { projectGraph: latestGraph, error } =
```
FrozenPandaz, line 145: The `projectGraph` has been used plenty in this function... wouldn't those also be outdated? What is sending this function an outdated graph?

FrozenPandaz (changes requested):

packages/nx/src/daemon/server/sync-generators.ts
FrozenPandaz: Nevermind, I get it now

FrozenPandaz (approved):

PR #33783, opened 2025-12-10
fix(core): fix record stat on initial CNW call

jaysoo:
This PR fixes an issue with CNW where the initial `recordStat` call is not working due to a logic error on passing `directory` that isn't initialized yet.


packages/create-nx-workspace/src/utils/nx/ab-testing.ts
jaysoo, line 273: This is already no hardcoded anymore.

FrozenPandaz (approved):

PR #33791, opened 2025-12-10
fix(core): improve performance of buildExplicitPackageJsonDependencies

meeroslav:
Further improvement of the graph creation.

Improving the `buildExplicitPackageJsonDependencies`.

| | Before | After |
| ---- | ----- | ---- |
| buildExplicitPackageJsonDependencies | 128 | 12 |
| isPackageJsonAtProjectRoot | 114 | 1 |

Fixes #


packages/nx/src/plugins/js/project-graph/build-dependencies/build-dependencies.ts
```diff
@@ -19,16 +18,14 @@ export function buildExplicitDependencies(
...
-    Object.entries(ctx.projects).map(([key, config]) => [
-      key,
-      {
-        name: key,
-        type: null,
-        data: config,
-      },
-    ])
-  );
+  const nodes = {};
```
leosvelperez: We should keep the type-safety for the creation below:
```suggestion
  const nodes: Record<string, ProjectGraphProjectNode> = {};
```

leosvelperez (approved):

PR #33821, opened 2025-12-11
feat(js): add option for using tsgo compiler over tsc for build, typecheck

ajkarell:
Adds a 'compiler' option to the @nx/js/typescript plugin configuration, with options 'tsc' and 'tsgo'. Affects both typecheck and build targets.

## Current Behavior

The `@nx/js/typescript` plugin always uses `tsc` as the compiler, with no way to use the native `tsgo` preview.

## Expected Behavior

The `@nx/js/typescript` plugin can be configured to use `tsgo` for building and typechecking.

## Related Issue(s)

Related discussion #32591.


leosvelperez (approved):
Thank you!

felipeplets (approved):

PR #33822, opened 2025-12-11
fix(js): allow copying generated Prisma client (asset) from 'node_modules'

parostatkiem:
Removed 'node_modules' from the ignore list for asset copying.

## Current Behavior

Updating Nx to v22 breaks my app. Some projects use generated Prisma clients which, because of how Prisma binaries work, have to be generated to `node_modules` in order to work both locally and in Docker context.
With Nx v22 Prisma client is not copied when project is built using `@nx/esbuild:esbuild` (or any other executor supporting `assets` property) without any error or warning. It took mi couple of hours to pinpoint the exact line of code responsible for this. `node_modules` dir is hardcoded there without any possibility to be overridden.

## Expected Behavior

Prisma client can be copied from `node_modules` to project output directory when building. 
Ignoring `node_modules` is removed **or can be overridden**. I'm open to any solution which will let me update Nx in my repository.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/angular/src/plugins/plugin.spec.ts
+++ b/packages/angular/src/plugins/plugin.spec.ts
@@ -1,5 +1,5 @@
 import type { CreateNodesContextV2 } from '@nx/devkit';
-import { mkdirSync, rmdirSync } from 'node:fs';
+import { mkdirSync, rmSync } from 'node:fs';
 import { TempFs } from 'nx/src/internal-testing-utils/temp-fs';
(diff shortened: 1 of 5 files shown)


jaysoo:
Thanks. There were some problem, mainly with performance regression if `node_modules` isn't needed. I pushed a commit to check if we need to include `node_modules` or not.

jaysoo (approved):

Issue #30359, opened 2025-03-13
Docker container won't restart when serving app with Nx and NX_ISOLATE_PLUGINS = true


decline:
### Current Behavior

When serving an app with `nx serve` inside a docker container, it will keep hanging (without any log output) when restarting the container using `docker compose restart`.

I noticed this starting with Nx 20, but it could have been the case earlier maybe.

While researching, I came across this comment https://github.com/nrwl/nx/issues/27494#issuecomment-2297519561 for a similar issue, that suggest to set `NX_ISOLATE_PLUGINS` to false.

**That fixes the issue.** 

I don't know what changing this variable actually causes, since I cannot find any documentation about it. Are there any negative side-effects maybe?

Nevertheless, I guess it should be fixed so that docker containers can be restarted successfully without adjusting this variable.

### Expected Behavior

Docker containers should be able to restart without setting `NX_ISOLATE_PLUGINS` to `false`.

### GitHub Repo

https://github.com/decline/docker-nx-plugin-issue

### Steps to Reproduce

Checkout the repository, then:

## How to reproduce

1. Start docker container with `docker compose up` and wait until app gets served.
2. Go to another terminal and run `docker compose restart`.
3. Go back to the first terminal => the application will hang

## How to fix

1. Stop docker container again with `docker compose down`
2. Uncomment environment variable `NX_ISOLATE_PLUGINS: false` in docker-compose.yml
3. Start docker container `docker compose up` and wait until app gets served.
4. Go to another terminal and run `docker compose restart`.
5. Go back to the first terminal => the application restarts successfully

### Nx Report

```shell
Node           : 22.13.1
OS             : linux-arm64
Native Target  : aarch64-linux
npm            : 10.9.2

nx                     : 20.5.0
@nx/js                 : 20.5.0
@nx/jest               : 20.5.0
@nx/eslint             : 20.5.0
@nx/workspace          : 20.5.0
@nx/angular            : 20.5.0
@nx/devkit             : 20.5.0
@nx/eslint-plugin      : 20.5.0
@nx/module-federation  : 20.5.0
@nx/web                : 20.5.0
@nx/webpack            : 20.5.0
typescript             : 5.7.3
---------------------------------------
Community plugins:
angular-eslint : 19.2.1
```

### Failure Logs

```shell

```

### Operating System

- [x] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

[...]

Issue #14126, opened 2023-01-04
Nx daemon not available in docker containers


hytromo:
### Current Behavior

Currently the Nx daemon usage is completely disabled when Nx is running in a docker container ([source](https://github.com/nrwl/nx/blob/19abb973f0b78733c1fbd3d9ee05576a42a602e5/packages/nx/src/daemon/client/client.ts#L75)).

This causes Nx commands to recalculate the project graph each time instead of using the daemon's cache ([source](https://github.com/nrwl/nx/blob/19abb973f0b78733c1fbd3d9ee05576a42a602e5/packages/nx/src/project-graph/project-graph.ts#L127)).

So, for every nx command that is run, this includes hashing all the monorepo files each time, which can be a really time consuming operation for larger monorepos ([source](https://github.com/nrwl/nx/blob/19abb973f0b78733c1fbd3d9ee05576a42a602e5/packages/nx/src/project-graph/build-project-graph.ts#L51)).

We are developing inside a docker container to ensure common tooling between the team members, and this is really slowing us down (5+ seconds to _start_ running each nx command, like `nx run <project>:<target>`)

Hacking the code on the fly to enable the daemon to work inside a docker container seems to work fine, so not sure why this was introduced [here](https://github.com/nrwl/nx/commit/1a31018ea2bc9f415268636b2ad834ff02a8d147#diff-ac77bbed6ea29034032e92619972b1ad989e0accad1b05b5462a276707549da0).

### Expected Behavior

Nx Daemon should work inside docker containers to avoid recalculating the project graph each time which can be a very expensive operation.

### Steps to Reproduce

1. Start a docker container and mount your large monorepo in it
2. Try to run an nx target to see the slow startup time
3. Run `nx daemon` to see that the daemon hasn't started
4. Run another nx target and notice how slow the startup is

### Nx Report

```shell
Node : 18.12.1
   OS   : linux x64
   npm  : 8.19.2
   
   nx : 15.4.2
   @nrwl/angular : Not Found
   @nrwl/cypress : Not Found
   @nrwl/detox : Not Found
   @nrwl/devkit : 15.4.2
   @nrwl/esbuild : Not Found
   @nrwl/eslint-plugin-nx : Not Found
   @nrwl/expo : Not Found
   @nrwl/express : Not Found
   @nrwl/jest : Not Found
   @nrwl/js : 15.4.2
   @nrwl/linter : 15.4.2
   @nrwl/nest : Not Found
   @nrwl/next : Not Found
[...]

PR #33848, opened 2025-12-13
docs(core): add Docker development guidance

kazukinakai:
## Summary

Adds comprehensive Docker development guidance to the Nx Daemon documentation.

Related to #33263, #30359, #14126

## What's included

This PR expands the existing Nx Daemon docs with a new "Running Nx in Docker" section that covers:

- **Why the daemon often fails in Docker** - ephemeral filesystems, inode/mtime changes from volume mounts, container restarts, IPC issues
- **Recommended approach** - disable daemon with `NX_DAEMON=false`
- **Example docker-compose setup** - minimal reproducible configuration for local development
- **CI/CD best practices** - when to prefer stateless builds over daemon caching

## Context

Several issues have been opened around daemon behavior in containers, but the existing docs only briefly mention socket location customization. Users are left wondering:
- Why doesn't the daemon work reliably in Docker?
- What's the recommended workflow for containerized development?
- How should CI pipelines handle this?

Issue #33263 specifically describes daemon crashes when running nx between Docker and non-docker environments - this PR documents the recommended workaround (`NX_DAEMON=false`) and explains why.

## Preview

The new section appears under the existing "Customizing the socket location" heading and includes:
- Explanation of Docker-specific challenges
- Code examples for Dockerfile, docker-compose, and CLI
- Rule of thumb callout for quick reference

Open to feedback on structure or placement.

### DIFF
--- a/astro-docs/src/content/docs/concepts/nx-daemon.mdoc
+++ b/astro-docs/src/content/docs/concepts/nx-daemon.mdoc
@@ -51,72 +51,30 @@ To see information about the running Nx Daemon (such as its background process I
 The Nx Daemon uses a unix socket to communicate between the daemon and the Nx processes. By default this socket gets placed in a temp directory. If you are using Nx in a docker-compose environment, however, you may want to run the daemon manually
 and control its location to enable sharing the daemon among your docker containers. To do so, simply set the NX_DAEMON_SOCKET_DIR environment variable to a shared directory.
 
-## Running Nx in Docker
+## Daemon Behavior in Containers
 
-Docker containers present unique challenges for the Nx Daemon due to their ephemeral nature and filesystem behavior.
+Nx automatically disables the daemon in Docker containers and CI environments. The daemon's performance benefits come from maintaining state between commands and watching for file changes—in short-lived environments where each command runs in a fresh container, this state cannot be reused, so the overhead of starting a background process provides no benefit.
 
-### Why the daemon often fails in Docker
+### Automatic Detection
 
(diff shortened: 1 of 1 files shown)


distante:
This isn't a correct workaround. Without the NX daemon, you are stuck with services that doesn't reload after changes. You need to manually stop the container and start it again.

kazukinakai:
Good point - you're right that disabling the daemon loses the automatic file watching.

For hot reload without the daemon, the app's bundler (webpack/vite) needs to poll for changes instead:

```js
// webpack.config.js
devServer: {
  watchOptions: {
    poll: 1000,
    ignored: /node_modules/
  }
}
```

I can add a section covering this trade-off and the polling workaround. Would that address your concern?

distante:
I am not on my laptop right now but I think that will not work with projects that runs NestJs applications with the node executor (which is our problem)

jaysoo:
I updated the wording a bit, and while it does not provide specific guidance on enabling daemon in containers, it at least clarifies why it is automatically disabled in CI and Docker, with the points that @kazukinakai provided.

We can look into a guide for Docker + Daemon in a follow-up. This at least provides users with some information that are currently missing.

jaysoo (approved):

jaysoo:
I merged the content since it provide more information for anyone wondering about Docker + Daemon. As I said earlier, we should have a separate guide for using Dev Containers with Daemon, as the CI/production use case without file-watching is fine without daemon, but local development is not.

PR #33862, opened 2025-12-15
feat(core): delegate db operations to daemon when enabled

leosvelperez:
Delegates all DB operations to the daemon when enabled.


packages/nx/src/command-line/report/report.ts
```diff
@@ -420,12 +420,14 @@ export async function getReportData(): Promise<ReportData> {
...
-      }
-    : null;
+  let cache = null;
+  if (dbCacheEnabled()) {
+    const cacheInstance = new DbCache({ nxCloudRemoteCache: null });
+    cache = {
+      max: resolveMaxCacheSize(nxJson),
+      used: await cacheInstance.getUsedCacheSpace(),
+    };
+  }
```
AgentEnder: Nit: lets just move this to a getCacheInfo function or something so cache can be const and it matches the other data bits

packages/nx/src/tasks-runner/cache.ts
```diff
@@ -137,15 +164,39 @@ export class DbCache {
     }
   }
 
-  getUsedCacheSpace() {
-    return this.cache.getCacheSize();
+  getUsedCacheSpace(): number | Promise<number> {
```
AgentEnder: Maybe it would make more sense to move to two separate classes... A `daemon-enabled` cache which always returns promises, and a `in-process` cache that is essentially the same as before. This would mirror the tasks hasher setup to some extent, but would mean our return types are either always a promise or retain their current typings.

IMHO it'd be worth it?

packages/nx/src/utils/consume-messages-from-socket.ts
```diff
@@ -41,9 +41,10 @@ export function isJsonMessage(message: string): boolean {
   return (
     // json objects
     ['[', '{'].some((prefix) => message.startsWith(prefix)) ||
-    // booleans
+    // booleans and null
     message === 'true' ||
     message === 'false' ||
+    message === 'null' ||
```
AgentEnder: Why is this needed? We probably shouldn't have messages that are 'null'

packages/nx/src/utils/running-tasks.ts
```diff
@@ -0,0 +1,110 @@
...
+      performance.measure(
+        'db:runningTasks.get',
+        'db:runningTasks.get:start',
+        'db:runningTasks.get:end'
+      );
+      return result;
+    });
+  }
+
+  addRunningTask(taskId: string): void | Promise<void> {
```
AgentEnder: Honestly on these I'd just mark them async and always return a promise

packages/nx/src/utils/task-history.ts
```diff
@@ -1,10 +1,18 @@
+import { performance } from 'perf_hooks';
 import { daemonClient } from '../daemon/client/client';
 import { isOnDaemon } from '../daemon/is-on-daemon';
 import { IS_WASM, NxTaskHistory, TaskRun, TaskTarget } from '../native';
 import { getDbConnection } from './db-connection';
 
 export class TaskHistory {
```
AgentEnder: DaemonTaskHistory + InProcessTaskHistory? Not 100% convinced it'd be necessary but maybe.

The perf marks + measures matter a whole lot less when the code is on the demon... since the daemon logs the message timings. Not sure what if its worth removing them, but it is a bit weird that they are being added here.

AgentEnder (changes requested):

packages/nx/src/utils/consume-messages-from-socket.ts
leosvelperez: An early iteration of the changes returned some of the data directly and could be null, but I changed that since then.

Hardanish-Singh (approved):

PR #33906, opened 2025-12-17
fix(testing): remove `declare global` wrapper from cypress component test support file

leosvelperez:
## Current Behavior

When generating Cypress component testing configuration, the generated `support/component.ts` file wraps the Cypress namespace declaration inside a `declare global` block. This can cause TypeScript compatibility issues in certain project configurations.

## Expected Behavior

The generated `support/component.ts` file now uses `declare namespace Cypress` directly without the `declare global` wrapper, which is the recommended pattern for extending Cypress's type definitions and provides better TypeScript compatibility.

## Related Issue(s)

Fixes #33629


jaysoo (approved):

tomalaforge:
This doesn't work for me, I have this error

```
TS2345: Argument of type 'mount' is not assignable to parameter of type keyof Chainable<any>
```

When I change the declare global with declare namespace

PR #34006, opened 2025-12-30
fix(core): handle resizing a bit better for inline_tui

AgentEnder:
## Current Behavior
Resizing the TUI while in inline view kinda breaks things. Its unfortunate, I'm not sure there's a ton to be done, but this PR explores some solutions

## Expected Behavior
The TUI is less sensitive to resize events with inline mode

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/tui/inline_app.rs
+++ b/packages/nx/src/native/tui/inline_app.rs
@@ -848,9 +848,6 @@ impl InlineApp {
                 )
             };
 
-        // Get cloud message from state (using trait default implementation)
-        let cloud_message = TuiApp::get_cloud_message(self);
-
         // Get status message (e.g., "Output copied")
         let status_msg = self.status_message.as_ref().map(|(msg, _)| msg.as_str());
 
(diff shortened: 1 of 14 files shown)


packages/nx/src/native/tui/inline_app.rs
```diff
@@ -215,35 +216,38 @@ impl InlineApp {
     /// Called when switching to inline mode via init().
     /// This ensures PTY output fits the available space, pushing excess content
     /// into scrollback.
-    fn resize_all_ptys(&mut self) {
-        let (rows, cols) = self.calculate_pty_dimensions();
+    fn resize_selected_pty(&mut self, forced_dimensions: Option<(u16, u16)>) {
```
FrozenPandaz:
```suggestion
    fn resize_selected_pty(&mut self, dimensions: (u16, u16)) {
```

FrozenPandaz (changes requested):
This is 90% of the way there. I'm able to enter a broken state when.. either resizing a lot... or... sometimes just resizing.

FrozenPandaz (approved):

PR #34031, opened 2026-01-06
feat(release): special-case 0.x versions for semver bumps

Coly010:
For 0.x versions, shift semver bump types down to follow the common
convention where breaking changes bump minor, and new features bump patch:

- major -> minor
- premajor -> preminor
- minor -> patch
- preminor -> prepatch
- patch -> patch (unchanged)

This ensures that `nx release` with a breaking change on a 0.x package
(e.g., 0.1.0) bumps to 0.2.0 instead of 1.0.0.

Fixes NXC-3638

### DIFF
--- a/packages/nx/schemas/nx-schema.json
+++ b/packages/nx/schemas/nx-schema.json
@@ -1036,7 +1036,7 @@
         },
         "adjustSemverBumpsForZeroMajorVersion": {
           "type": "boolean",
-          "description": "Whether to apply the common convention for 0.x versions where breaking changes bump the minor version (instead of major), and new features bump the patch version (instead of minor). When enabled: 'major' bumps become 'minor' bumps for 0.x versions, 'minor' bumps become 'patch' bumps for 0.x versions. Versions 1.0.0 and above are unaffected. This is false by default for backward compatibility.",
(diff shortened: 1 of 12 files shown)


jaysoo (approved):

packages/nx/schemas/nx-schema.json
```diff
@@ -1034,6 +1034,11 @@
             ]
           }
         },
+        "adjustSemverBumpsForZeroMajorVersion": {
+          "type": "boolean",
+          "description": "Whether to apply the common convention for 0.x versions where breaking changes bump the minor version (instead of major), and new features bump the patch version (instead of minor). When enabled: 'major' bumps become 'minor' bumps for 0.x versions, 'minor' bumps become 'patch' bumps for 0.x versions. Versions 1.0.0 and above are unaffected. This is false by default for backward compatibility.",
```
JamesHenry: It's documented in the latest v2 semver spec: https://semver.org/#spec-item-4 so I think it's more than common convention tbh, the big thing is that the node-semver library doesn't follow that spec sadly

Coly010: What would you suggest the wording be instead? "Whether to strictly follow SemVer V2 spec for 0.x versions..."

JamesHenry (approved):

PR #34046, opened 2026-01-08
fix(maven): maven batch runner to use tcp-based communication

altaiezior:
## Current Behavior
Maven plugin executions can sometimes lead to stackoverflow error, if a plugin is writing to stdout

In my current setup, I am facing this issue however I am not able to add a reproducer
It is a multi-module project where in one of the modules I am using maven-shade-plugin

When I am including this module, I see this error while excluding this plugin works fine. However I can't even pin-point that shading is the issue here. But I observed this and it can affect other plugins too.

With batch implementation either the execution usually gets stuck with this error

## Expected Behavior
Stackover error should not come even when any plugin is writing using System.out / System.err as maven logger tries to redirect these to its maven logger

### Root Cause
1. `NxMavenBatchRunner` configures Maven's logger to write to `System.out` via `System.setProperty("maven.logger.logFile", "System.out")`.
2. When running in resident/embedded mode, Maven 4's `LookupInvoker` attempts to capture all `System.out` output by redirecting it to a `LoggingOutputStream`.
3. `LoggingOutputStream` turns everything written to it into a log event (`logger.info()`).
4. `MavenBaseLogger` (the logger implementation) handles this event by writing to its configured log file, which is `System.out`.
5. Since `System.out` is now the `LoggingOutputStream` from step 2, it causes infinite recursion.

### Solution
The fix involves disabling Maven 4's automatic redirection of standard streams when running in the resident executor. This is achieved by adding the `--raw-streams` flag to the Maven arguments. In "raw streams" mode, Maven 4 leaves `System.out` and `System.err` untouched, allowing the runner to manage them directly (which it already does by passing `stdOut` and `stdErr` to the `ParserRequest`).

Moreover currently we are using System.err for IPC which is also seems like a fragile setup. Moreover when using `--raw-parser`, the NX_RESULT logs are also printed on console which adds noise.

Introduced `TcpResultEmitter` for transmitting task result data via a TCP socket instead of using `System.err`. This along with `--raw-parser` eliminates the issue without adding noise to the console and makes setup more robust.

### Changes
- Modified `packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/ResidentMavenExecutor.kt` to automatically add `--raw-streams` to the arguments if not already present.
- Updated `packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/NxMavenBatchRunner.kt` with a comment explaining the logging configuration and its interaction with `--raw-streams`.
- Introduced `TcpResultEmitter` for transmitting task result data via a TCP socket
- Replaced `stderr`-based result streaming with a TCP server in Node.js batch implementation
- Updated CLI and argument parsing to require a `--communicationPort` for the connection
- Ensured resource cleanup by properly managing socket and server lifecycle

## Related Issue(s)

Fixes #34038

### DIFF
--- a/packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/NxMavenBatchRunner.kt
+++ b/packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/NxMavenBatchRunner.kt
@@ -16,7 +16,7 @@ private val log = run {
     System.setProperty("maven.logger.showLogName", "false")
     System.setProperty("maven.logger.levelInBrackets", "true")
     System.setProperty("maven.logger.defaultLogLevel", "info")
-    // Log to stdout (stderr is used for NX_RESULT streaming)
+    // NOTE: We use --raw-streams in ResidentMavenExecutor to prevent recursion
     System.setProperty("maven.logger.logFile", "System.out")
     // Enable colored output (jansi.force needed when stdout isn't a TTY)
     System.setProperty("style.color", "always")
--- a/packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/cli/ArgParser.kt
+++ b/packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/cli/ArgParser.kt
@@ -12,6 +12,7 @@ object ArgParser {
     fun parseArgs(args: Array<String>): MavenBatchOptions {
         var workspaceRoot = ""
         var verbose = false
+        var communicationPort: Int? = null
 
         var i = 0
         while (i < args.size) {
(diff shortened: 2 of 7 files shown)


jaysoo (commented):
The move from stderr-based `NX_RESULT:` parsing to TCP socket communication is the right approach for Maven 4 compatibility (avoiding the recursive `System.out` StackOverflowError). The `--raw-streams` flag is a good addition.

One concern: if the TCP connection fails, results are silently dropped — the writer becomes null and subsequent `emit()` calls only log a warning. This could lead to hard-to-debug silent failures in CI. Consider propagating the failure (e.g., throwing or returning an error status) so the task is properly reported as failed rather than appearing to succeed with no output.

Otherwise the approach is sound. Good domain expertise on the Maven 4 internals.

altaiezior:
@FrozenPandaz @jaysoo I will be able to pick up the changes by next week as I am travelling and catching up with my work lately

PR #34052, opened 2026-01-08
docs(nx-cloud): add new scopes for BB when using Nx Cloud onboarding

barbados-clemens:


astro-docs/src/content/docs/guides/Nx Cloud/Source Control Integration/bitbucket.mdoc
```diff
@@ -32,8 +32,8 @@ An API token is a secure credential that allows scripts and other processes to a
    - `write:pullrequest:bitbucket` - to write comments on pull requests
    - `read:repository:bitbucket` - to read repository contents
    - `write:repository:bitbucket` - to write files to the repository
-
-![Required Scopes](../../../../../assets/nx-cloud/set-up/bitbucket-api-tokens-selected-scopes.jpg)
```
llwt, line 36: should we delete the image from assets as well? Or will that break other versions of the docs that point to it?

llwt (approved):

jaysoo (approved):

PR #34061, opened 2026-01-09
fix(maven): use project from options instead of target

### LINKED ISSUE #34060: Maven Runner to use `options.project` instead of `target.projectName`
### Current Behavior

When inferring maven projects and targets the names generated are of java packages and a little bit complex to type when building a specific project / target

To simply I can just modify the name in the project.json in project specific folder

That works for the normal graph building and execution, but passes incorrect project name to maven, because of which it breaks.

We also have the original project name in the options.project which we can use instead.

### Expected Behavior

We should be able to customize the project names generated and inferred by @nx/maven plugin

### GitHub Repo

_No response_

### Steps to Reproduce

1. Create a project with pom.xml
2. Let the plugin inference work
3. Add a project.json file and add custom name
4. The build will break with respect to maven but the rest of the dependency graph will work fine

### Nx Report

```shell
Node           : 24.12.0
OS             : darwin-arm64
npm            : 11.6.2

nx (global)  : 22.4.0-beta.0
nx           : 22.4.0-beta.0
@nx/maven    : 22.4.0-beta.0
Registered Plugins:
@nx/maven
@naxodev/gonx
Community plugins:
@naxodev/gonx : 2.1.0
Cache Usage: 11.07 MB / 46.04 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

altaiezior:
## Current Behavior
Customising project name doesn't work for java projects, it uses the one inferred by the plugin

It doesn't impact the graph building logic of nx, just the part where the options are sent to maven for execution

## Expected Behavior
We should be allowed to customize the project names generated by the maven plugin as they can be too complex to be used in real-world for building daily

### Changes
* Used project from options which is added by nx-maven-plugin version > 0.0.11, this is the original project name generated by the plugin
* Fallback to the older format in case someone is not using the latest version of nx-maven-plugin

## Related Issue(s)

Fixes #34060

### DIFF
--- a/packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/MavenInvokerRunner.kt
+++ b/packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/MavenInvokerRunner.kt
@@ -164,8 +164,8 @@ class MavenInvokerRunner(private val workspaceRoot: File, private val options: M
       }
 
       // Extract unique project selectors from executed tasks
-      val uniqueProjectSelectors = options.taskGraph?.tasks?.values
-        ?.map { it.target.project }
+      val uniqueProjectSelectors = options.taskGraph?.tasks
+        ?.map { (taskId, task) -> options.taskOptions[taskId]?.project ?: task.target.project }
         ?.toSet() ?: emptySet()
 
       if (uniqueProjectSelectors.isEmpty()) {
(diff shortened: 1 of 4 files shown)


altaiezior:
Closing becuase of https://github.com/nrwl/nx/issues/34060#issuecomment-3863478327

PR #34094, opened 2026-01-13
fix(core): only init inline view if able to run

AgentEnder:
## Current Behavior
The inline tui runs some terminal escape codes to check cursor position, these break when stdin isn't a tty (like in a git hook)

## Expected Behavior
The inline tui is disabled if stdin isn't a tty

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/tui/tui.rs
+++ b/packages/nx/src/native/tui/tui.rs
@@ -10,7 +10,7 @@ use crossterm::{
 use futures::{FutureExt, StreamExt};
 use ratatui::backend::CrosstermBackend;
 use std::{
-    io::Write,
+    io::{IsTerminal, Write},
     ops::{Deref, DerefMut},
     time::Duration,
 };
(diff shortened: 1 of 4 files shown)


packages/nx/src/native/tui/tui.rs
```diff
@@ -227,6 +255,12 @@ impl Tui {
                 execute!(std::io::stderr(), EnterAlternateScreen, cursor::Hide)?;
             }
             TuiMode::Inline => {
+                // Inline terminal must exist (created upfront if stdin is TTY)
+                if self.inline_terminal.is_none() {
+                    return Err(color_eyre::eyre::eyre!(
```
FrozenPandaz, line 260: WHy use color_eyre here?

packages/nx/src/native/tui/tui.rs
```diff
@@ -300,8 +334,14 @@ impl Tui {
             }
         }
 
-        // Switch the mode - NO terminal recreation needed!
-        // Both terminals were created upfront, we just delegate to the right one via Deref
+        // Ensure inline terminal exists before switching to it
+        if new_mode == TuiMode::Inline && self.inline_terminal.is_none() {
+            return Err(color_eyre::eyre::eyre!(
```
FrozenPandaz, line 339: Same as above

FrozenPandaz (approved):

PR #34110, opened 2026-01-15
feat(bundling): replace rollup-plugin-postcss with inlined version

Coly010:
## Current Behavior
The `rollup-plugin-postcss` has not released a new version in 4 years.
The deps it depends on are outdated and starting to cause problems with peer-dep conflicts.

## Expected Behavior
Recreate the plugin within the `@nx/rollup` package to maintain the functionality/behaviour and manage dependencies ourselves.

## Related Issue(s)

Closes NXC-3644

### DIFF
--- a/packages/rollup/.eslintrc.json
+++ b/packages/rollup/.eslintrc.json
@@ -46,6 +46,8 @@
               "typescript",
               "@swc/core", // Installed to workspace and only used in swc() plugin
               "postcss",
+              "postcss-modules", // Dynamically loaded in postcss plugin
+              "source-map-js", // Used for types only
               "@rollup/plugin-typescript"
             ]
           }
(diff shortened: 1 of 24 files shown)


packages/rollup/.eslintrc.json
```diff
@@ -46,6 +46,8 @@
               "typescript",
               "@swc/core", // Installed to workspace and only used in swc() plugin
               "postcss",
+              "postcss-modules", // Dynamically loaded in postcss plugin
+              "source-map-js", // Used for types only
```
leosvelperez, line 50: Do these need to be ignored if we're adding them as `dependencies` in the `package.json` file?

If `source-map-js` is used for types only, should it be listed in `devDependencies` instead? We'd keep it ignored here in that case.

Coly010: yep, it absolutely should have been in devDependencies

Coly010: pushing a fix now

leosvelperez (approved):

PR #34115, opened 2026-01-15
feat(core): add failed project listing environment variable

AdamMcCormickOMF:
This adds `NX_FAILED_TASKS_TO_PRINT` as an environment variable controlling how many projects are shown when a command fails. 0 or less removes the limit

## Current Behavior
Right now, when some number of projects fail an `nx` command at most 5 items are shown and the rest are hidden

## Expected Behavior
Now, by specifying `NX_FAILED_TASKS_TO_PRINT` environment variable as a positive number (or 0 for no limit) that limit can be increased


AgentEnder:
Hey, I'm going to close this out for now. Feel free to open a discussion detailing why you want to be able to tune this, currently you can get all of the outputs by using a different output style and we feel its not terribly useful to print a larger number.

PR #34132, opened 2026-01-16
fix(core): record continuous task stats before process cleanup

### LINKED ISSUE #33561: postTasksExecution hook does not consistently include continuous tasks

comp615:
# fix(core): record continuous task stats before process cleanup

## Problem

When running continuous tasks (e.g., `nx serve app`) and sending Ctrl-C (SIGINT/SIGTERM), the `postTasksExecution` lifecycle hook does not receive continuous tasks in its `taskResults`. This breaks plugins and tooling that rely on post-run statistics to track all executed tasks.

**Root cause:** Signal handlers in `ForkedProcessTaskRunner` call `cleanup()` which kills child processes. The child processes exit, triggering `onExit` handlers in `TaskOrchestrator`. However, by the time `postTasksExecution` runs, the continuous tasks have already been removed from tracking maps and were never recorded via `lifeCycle.endTasks()`.

The critical issue is **timing**: `ForkedProcessTaskRunner` handles SIGINT/SIGTERM signals independently and kills processes before `TaskOrchestrator` has a chance to record stats.

Related issues:

- Fixes #33561
- Related to #33586

Older attempts:
https://github.com/nrwl/nx/pull/33562
https://github.com/nrwl/nx/pull/33572

## Solution

Add an `onBeforeCleanup` callback mechanism to `ForkedProcessTaskRunner` that allows `TaskOrchestrator` to record continuous task stats **before** processes are killed.

### Changes

**forked-process-task-runner.ts:**

- Add `onBeforeCleanup(callback)` method to register a pre-cleanup callback
- Call the callback at the start of `cleanup()` before sending kill signals

**task-orchestrator.ts:**

- Register callback in `init()` via `forkedProcessTaskRunner.onBeforeCleanup()`
- Add `recordContinuousTaskStatsSync()` method that:
  - Sets `cleaningUp = true` (prevents duplicate work in existing `onExit` handlers)
  - Iterates running continuous and run-commands tasks
  - Sets `endTime` and marks tasks as completed
  - Calls `lifeCycle.endTasks()` to populate `taskResults`
- Add idempotency guard to prevent double-invocation

### Why not modify `onExit` handlers? (Previous attempt in #33562)

The previous PR (#33562) attempted to fix this by modifying `onExit` handlers to call `postRunSteps()` when continuous tasks exit:

```typescript
childProcess.onExit(async (code) => {
  const status = this.cleaningUp ? 'success' : 'failure';
  await this.postRunSteps([task], [{ task, status }], false, { groupId });
});
```

**This approach fails because:**

1. `ForkedProcessTaskRunner`'s signal handlers call their own `cleanup()` directly, bypassing `TaskOrchestrator.cleanup()`
2. When `onExit` fires, `this.cleaningUp` is still `false` because `TaskOrchestrator.cleanup()` hasn't run yet
3. Tasks are incorrectly marked as `failure` instead of `success`
4. The async `onExit` handler may not complete before the process exits

### Why `onBeforeCleanup` is the right solution

1. **Correct timing:** Stats are recorded _before_ processes are killed, while we still know which tasks are running
2. **Single source of truth:** The callback fires from `ForkedProcessTaskRunner.cleanup()`, which is the actual point where shutdown begins, regardless of which signal handler triggered it
3. **Sets `cleaningUp = true` early:** Any subsequent `onExit` handlers will correctly see that cleanup is in progress
4. **Synchronous and reliable:** The callback runs synchronously before any async kill operations, guaranteeing execution
5. **Minimal changes:** Only ~35 lines across 2 files, no changes to existing control flow
6. **Idempotent:** Guard prevents issues if `cleanup()` is called multiple times

### Why this is the simplest method

| Approach                          | Files Changed | Lines Changed | Reliable? | Issues                                        |
| --------------------------------- | ------------- | ------------- | --------- | --------------------------------------------- |
| Modify `onExit` handlers (#33562) | 1             | ~20           | ❌        | `cleaningUp` not set; tasks marked as failure |
| Add guards to `postRunSteps`      | 1             | ~30           | ❌        | Complex filtering; doesn't fix root cause     |
| Refactor signal handling          | 3+            | 100+          | ⚠️        | Major architectural change                    |
| **`onBeforeCleanup` callback**    | **2**         | **~35**       | **✅**    | **None**                                      |

The callback approach provides a clean extension point that solves the timing problem without modifying existing signal handling or lifecycle management.

## Testing

Tested with a reproduction repo containing:

- Continuous task (`serve`) depending on another continuous task (`build-watch`)
- Both depending on a regular task (`task-a`)
- Plugin that logs `postTasksExecution` results

Before fix: Only `task-a` appears in `taskResults`
After fix: All 3 tasks (`task-a`, `build-watch`, `serve`) appear with `status: 'success'`

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if applicable)
- [x] Follows Nx coding conventions

### DIFF
--- a/packages/nx/src/tasks-runner/forked-process-task-runner.ts
+++ b/packages/nx/src/tasks-runner/forked-process-task-runner.ts
@@ -31,6 +31,7 @@ export class ForkedProcessTaskRunner {
   private processes = new Set<RunningTask | BatchProcess>();
   private finishedProcesses = new Set<BatchProcess>();
   private pseudoTerminals = new Set<PseudoTerminal>();
+  private onBeforeCleanupCallback?: () => void;
 
   constructor(
     private readonly options: DefaultTasksRunnerOptions,
@@ -410,10 +411,17 @@ export class ForkedProcessTaskRunner {
   }
 
   cleanup(signal?: NodeJS.Signals) {
+    if (this.onBeforeCleanupCallback) {
+      this.onBeforeCleanupCallback();
+    }
     this.processes.forEach((p) => {
       p.kill(signal);
     });
     this.cleanUpBatchProcesses();
+  }
+
+  onBeforeCleanup(callback: () => void) {
+    this.onBeforeCleanupCallback = callback;
   }
 
   private setupProcessEventListeners() {
--- a/packages/nx/src/tasks-runner/task-orchestrator.ts
+++ b/packages/nx/src/tasks-runner/task-orchestrator.ts
@@ -89,6 +89,7 @@ export class TaskOrchestrator {
 
   private runningContinuousTasks = new Map<string, RunningTask>();
   private runningRunCommandsTasks = new Map<string, RunningTask>();
+  private recordedShutdownStats = false;
 
   // endregion internal state
 
@@ -115,6 +116,11 @@ export class TaskOrchestrator {
       }),
       'init' in this.cache ? this.cache.init() : null,
     ]);
+
+    // Register callback to record continuous task stats before process cleanup
+    this.forkedProcessTaskRunner.onBeforeCleanup(() => {
+      this.recordContinuousTaskStatsSync();
+    });
 
     // Pass estimated timings to TUI after TasksSchedule is initialized
     if (this.tuiEnabled) {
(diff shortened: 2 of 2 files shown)


comp615:
@leosvelperez FYI...pulling this out of my previous comment: https://github.com/nrwl/nx/pull/33562#issuecomment-3761604799, take a look here too. Again...no obligation to use this if you are going down a different path, but I thought the information and some of the pitfalls I hit might be useful as you investigate (essentially problem being that we can't "await" onExit to happen before nx shuts down---so stats can be recorded before the continuous tasks callback)

leosvelperez:
@comp615 thanks again for your work here and investigations! This will be superseded by https://github.com/nrwl/nx/pull/34226.

PR #34148, opened 2026-01-19
fix(testing): add timeout to runCommandUntil to prevent hanging tests

FrozenPandaz:
## Current Behavior

The `runCommandUntil` e2e utility function waits indefinitely for the expected output to appear. If the output never appears (e.g., server fails to start, different output format, port conflict), the test hangs forever, causing CI jobs to run for hours before being killed.

## Expected Behavior

The function should timeout after a configurable duration and fail with a clear error message showing what output was received.

## Related Issue(s)

Fixes hanging e2e tests observed in CI (e.g., `e2e-node:e2e-ci--src/node-server.test.ts` hung for 1h 21m).

## Changes

- Added optional `timeout` parameter to `runCommandUntil` opts (default: 5 seconds)
- On timeout: kills the process, logs the collected output, and rejects with a clear error
- Existing call sites work unchanged; tests needing more startup time can pass `{ timeout: 30000 }`

### DIFF
--- a/.nx/workflows/dynamic-changesets.yaml
+++ b/.nx/workflows/dynamic-changesets.yaml
@@ -8,7 +8,6 @@ assignment-rules:
   - projects:
       - e2e-gradle
       - e2e-node
-      - e2e-angular
       - e2e-react
       - e2e-next
       - e2e-plugin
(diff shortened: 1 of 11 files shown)


FrozenPandaz:
@copilot Can you update the global-setup.ts file in e2e-utils to use environment variables rather than the npm config commands so that the setup and teardown do not polute other e2e tests which may be running?

PR #34151, opened 2026-01-20
fix(core): handle multibyte UTF-8 characters in socket message consumption

Chanki-Min:
## Current Behavior

When socket data chunks split a multibyte UTF-8 character (e.g., CJK characters like Korean, Chinese, Japanese) at an arbitrary byte boundary, `Buffer.toString()` decodes incomplete byte sequences as replacement characters (�), causing message corruption.

This can occur when:
- File paths contain non-ASCII characters
- Project names include multibyte characters
- Any JSON message contains international text

## Expected Behavior

Multibyte UTF-8 characters should be properly decoded even when split across multiple socket data chunks. The fix uses Node.js `StringDecoder` which buffers incomplete multibyte sequences until the remaining bytes arrive.

## Related Issue(s)

Fixes socket message corruption for paths/names containing multibyte characters.


lmalma2004 (approved):

p-iknow (approved):

AgentEnder (approved):

Chanki-Min:
Hello @AgentEnder . Thanks for reviewing my PR!

I just rebased onto main branch to resolve conflict. (should be merge..)So I need another approval to run CI workflow. Thanks

PR #34160, opened 2026-01-20
feat(core): add negation pattern support for plugin include/exclude

llwt:
## Current Behavior
Negation patterns are ignored in plugin configuration for the `include` and `exclude` properties.

## Expected Behavior

- Negation patterns should work in the same way that they do for other `include`/`exclude` configurations

**Example: Excluding all e2e projects except one**

```jsonc
// nx.json
{
  "plugins": [
    {
      "plugin": "@nx/jest/plugin",
      "exclude": ["**/*-e2e/**/*", "!**/toolkit-workspace-e2e/**/*"],
    },
  ],
}
```

This will exclude all e2e projects except `toolkit-workspace-e2e`.

**Example: Including packages except legacy ones**

```jsonc
// nx.json
{
  "plugins": [
    {
      "plugin": "@nx/vite/plugin",
      "include": ["packages/**/*", "!packages/legacy/**/*"],
    },
  ],
}
```

**How negation patterns work:**

- Patterns are processed in order from first to last
- A pattern starting with `!` removes files from the match set
- A pattern without `!` adds files to the match set
- The last matching pattern determines if a file is included
- If the first pattern is a negation, all files are matched initially


packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -592,20 +622,16 @@ export function findMatchingConfigFiles(
...
   for (const file of projectFiles) {
     if (minimatch(file, pattern, { dot: true })) {
-      if (include) {
-        const included = include.some((includedPattern) =>
-          minimatch(file, includedPattern, { dot: true })
-        );
-        if (!included) {
+      // Process include patterns with negation support
+      if (include && include.length > 0) {
+        if (!matchesPatternArray(file, include)) {
```
llwt: @AgentEnder So to do this we have to move from `.some` to always checking all patterns. 

This feels like it should be ok, but I worry that someone might have a project config doing something weird like enumerating all projects they want a plugin enabled for 😅

Thoughts?

AgentEnder: Hmm, in a repo I've not tuned at all, just using our generators to create some buildable and some non-buildable projects, the ts plugin is configureded as below:

```json
{
  "plugins": [
    {
      "options": {
        "typecheck": {
          "targetName": "typecheck"
        },
        "build": {
          "targetName": "build",
          "configName": "tsconfig.lib.json",
          "buildDepsName": "build-deps",
          "watchDepsName": "watch-deps"
        }
      },
      "exclude": [
        "packages/model/*",
        "packages/env-schemas/*",
        "packages/schema-utils/*",
        "e2e/*"
      ]
    },
    {
      "include": [
        "packages/model/*",
        "packages/env-schemas/*",
        "packages/schema-utils/*"
      ],
      "options": {
        "typecheck": {
          "targetName": "typecheck"
        }
      }
    },
  ]
}

Comment (llwt) [reply packages/nx/src/project-graph/utils/project-configuration-utils.ts]: Added a check for negations in the patterns before doing the more expensive check. With this the current perf should at least be maintained for anyone not using negations.

**Review by barbados-clemens — APPROVED** [review]
docs are fine 👍

File: packages/nx/src/project-graph/utils/project-configuration-utils.ts [hunk 2]
```diff
@@ -590,29 +640,25 @@ export function findMatchingConfigFiles(
...
   for (const file of projectFiles) {
     if (minimatch(file, pattern, { dot: true })) {
-      if (include) {
-        const included = include.some((includedPattern) =>
-          minimatch(file, includedPattern, { dot: true })
-        );
-        if (!included) {
-          continue;
-        }
+      if (!includes(file)) {
```
AgentEnder, line 650: I honestly kind of hate that we are doing this whole `if matches, check matches, then check other matches` thing, but its existing code and I don't know of a good way to combine them off the top of my head, nor that it'd be faster if we did... In any case, that's not your PRs fault. I think this is fine?

AgentEnder (approved):

PR #34180, opened 2026-01-21
feat(maven): load Maven classes at runtime for version-agnostic batch execution

FrozenPandaz:
## Current Behavior

The Maven batch runner bundles Maven 4 classes at compile time, which:
- Creates a large JAR file (~50+ MB)
- Only works with Maven 4
- Has classloader conflicts with Maven's own SLF4J

## Expected Behavior

The batch runner loads Maven classes at runtime from `MAVEN_HOME`, which:
- Creates a small JAR (~2 MB) with no bundled Maven dependencies
- Works with both Maven 3.x and Maven 4.x
- Avoids classloader conflicts by isolating Maven in its own ClassRealm
- Outputs clean Maven-style logs (`[INFO]`, `[WARNING]`, etc.)

### Architecture

```
batch-runner.jar (NO Maven dependencies)
├── MavenClassRealm         → Loads Maven JARs from MAVEN_HOME at runtime
├── ResidentMavenExecutor   → Maven 4 executor (reflection-based)
├── CachingMaven3Invoker    → Maven 3 executor (reflection-based)
└── nx-maven-adapters/      → Pre-compiled adapter JARs (embedded as resources)
    ├── batch-runner-adapters-maven3.jar
    └── batch-runner-adapter-maven4.jar
```

### Key Changes

1. **Removed compile-time Maven dependencies** from batch-runner module
2. **Created batch-runner-adapters** modules for Maven 3 and Maven 4 specific code
3. **Implemented MavenClassRealm** to load Maven JARs from MAVEN_HOME at runtime
4. **Implemented reflection-based executors** that load adapter JARs into ClassRealm
5. **Fixed SLF4J logging** with System.out redirection for clean Maven-style output
6. **Added shared module** for BuildStateManager, BuildStateApplier, and BuildStateRecorder

### Benefits

- **Version agnostic**: Same JAR works with Maven 3.x and 4.x
- **Graph caching**: Project dependency graph built once, reused across tasks
- **Build state persistence**: compile → package → install works correctly
- **No classloader conflicts**: Maven's classes isolated in their own ClassRealm
- **Clean output**: Standard Maven log format without SLF4J noise

## Related Issue(s)

N/A - Internal refactoring for better Maven version support


packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/ResidentMavenExecutor.kt
```diff
@@ -1,295 +1,89 @@
...
+    init {
+        log.debug("Initializing ResidentMavenExecutor for Maven $mavenMajorVersion")
+        configureSystemProperties()
+
+        mavenRealm = MavenClassRealm.create(mavenHome)
+        try {
+            mavenRealm.loadAdapterJar(mavenMajorVersion)
+            invoker = createInvoker()
+        } catch (e: Exception) {
+            mavenRealm.close()
```
lourw, line 37: Would it make more sense to move this into a finally so that you don't swallow errors from the load if close throws?

packages/maven/batch-runner-adapters/maven3/src/main/kotlin/dev/nx/maven/adapter/maven3/CachingMaven3Invoker.kt
```diff
@@ -0,0 +1,188 @@
...
+
+    init {
+        initialize()
+    }
+
+    private fun initialize() {
+        log.debug("Initializing CachingMaven3Invoker...")
+
+        System.setProperty("maven.multiModuleProjectDirectory", workspaceRoot.absolutePath)
+        System.setProperty("maven.home", System.getenv("MAVEN_HOME") ?: "")
```
lourw: No downstream impacts of us setting this system property to "" if the env is not found?

packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/MavenClassRealm.kt
```diff
@@ -0,0 +1,99 @@
...
+    fun loadAdapterJar(mavenMajorVersion: String) {
+        val adapterJar = findAdapterJar(mavenMajorVersion)
+            ?: throw RuntimeException("Adapter JAR not found for Maven $mavenMajorVersion")
+
+        realm.addURL(adapterJar.toURI().toURL())
+        log.debug("Loaded adapter JAR: ${adapterJar.name}")
+    }
+
+    private fun findAdapterJar(mavenMajorVersion: String): File? {
+        val codeSource = MavenClassRealm::class.java.protectionDomain.codeSource
```
lourw, line 23: Seems like issues here just get hidden by returning null. Could be surfaced?

packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/ResidentMavenExecutor.kt
FrozenPandaz: No, the Maven Realm has to remain open if it succeeds.. it only needs to close if there's an error. The error from load is rethrown below.

lourw (approved):

PR #34182, opened 2026-01-21
fix(maven): fix set the pom file without changing base directory

altaiezior:
## Current Behavior
nx-maven-plugin 0.0.12 is also changing the base directory along with the pom file when the plugins like flatten-maven-plugin / maven-shade-plugin produces pomFile in different directory other than the base directory

## Expected Behavior
plugin should only update the pom file and not the base directory location

## Related Issue(s)

Fixes #34181

https://github.com/mojohaus/flatten-maven-plugin/issues/50


altaiezior:
@FrozenPandaz would request your review on this one as a fix to the current nx-maven-plugin

lourw (approved):

altaiezior:
@lourw @FrozenPandaz this is a simple change, can we plan for its release in the next release?

jaysoo (approved):
Confirmed this is the right fix. Maven's `setPomFile()` updates both the file reference and the `basedir`, which `setFile()` (via direct assignment) does not do. This is critical for plugins like `flatten-maven-plugin` that resolve paths relative to the POM's base directory.

Single-line, no regression risk. LGTM.

PR #34193, opened 2026-01-22
fix(core): cloud commands are noop when not connected rather than errors

jaysoo:
This PR makes it so Cloud commands like `npx nx record` and `npx nx fix-ci` still work without `nxCloudId`. We'll log a warning so that `ci.yml` using these commands will still work. The warning let's users know that these do not work without being connected.

Closes #NXC-3753

### DIFF
--- a/packages/nx/src/command-line/nx-cloud/utils.ts
+++ b/packages/nx/src/command-line/nx-cloud/utils.ts
@@ -2,6 +2,23 @@ import { verifyOrUpdateNxCloudClient } from '../../nx-cloud/update-manager';
 import { getCloudOptions } from '../../nx-cloud/utilities/get-cloud-options';
 import { handleErrors } from '../../utils/handle-errors';
 import { findAncestorNodeModules } from '../../nx-cloud/resolution-helpers';
+import { readNxJson } from '../../config/nx-json';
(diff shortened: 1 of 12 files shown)


Coly010 (approved):

packages/nx/src/command-line/nx-cloud/utils.ts
```diff
@@ -2,6 +2,23 @@ import { verifyOrUpdateNxCloudClient } from '../../nx-cloud/update-manager';
 import { getCloudOptions } from '../../nx-cloud/utilities/get-cloud-options';
 import { handleErrors } from '../../utils/handle-errors';
 import { findAncestorNodeModules } from '../../nx-cloud/resolution-helpers';
+import { readNxJson } from '../../config/nx-json';
+import { output } from '../../utils/output';
+
+export function isConnectedToNxCloud(): boolean {
```
FrozenPandaz: Use isNxCloudUsed() in packages/nx/src/utils/nx-cloud-utils.ts instead. Remove this function.

That one checks if the auth token is in the env variables and stuff.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

Issue #32649, opened 2025-04-01
Angular tries to compile package.json file


Tallyb:
**Describe the bug**
In our code we import the package.json file to get version number. 
Angular compiler is trying to process package.json file (and potentially any json file?) with babel transformer. 
```
SyntaxError:../package.json: Missing semicolon. (2:8)

  1 | {
> 2 |   "name": "xxx",
    |         ^
  3 |   "version": "13.0.0-15",

    at constructor (node_modules/@babel/parser/lib/index.js:360:19)
    at Parser.raise (node_modules/@babel/parser/lib/index.js:3338:19)
    at Parser.semicolon (node_modules/@babel/parser/lib/index.js:3664:10)
    at Parser.parseExpressionStatement (node_modules/@babel/parser/lib/index.js:12949:10)
    at Parser.parseStatementContent (node_modules/@babel/parser/lib/index.js:12563:19)
```
(full path was truncated)

**To Reproduce**
I would love to have a starter project, on which I can work. 
Probably add something like: 
```
const version = await import('../package.json').then(m => m.default.version)
```
**Expected behavior**
Handle json file correctly, not thru babel. 

**Desktop (please complete the following information):**
 - OS: MacOS 15.3.2
 - Browser N/A
 - Version nx/angualr-rspack 20.6.1 angular 19+

**Additional context**
this is happening in the angular "transformWithBabel" method found in node_modules/@angular/build/src/tools/esbuild/javascript-transformer-worker.js. 
Probably the issue is that .json files are moved to being processed by angular build

vz-tl:
...that's a really odd bug ;-)

My app is running on rspack and in several places I'm also importing JSON files, which works flawlessly. 

But now I just tried to import the root package.json and guess what? The build crashed immediately.

The culprit is any string occurrence of `@angular`!

Replacing all such occurrences in package.json with e.g. "xxxx" and the import will work without issues.

@Coly010 , in order to reproduce just add e.g. `@angular/compiler` as dependency to https://github.com/nrwl/angular-rspack/blob/main/e2e/fixtures/rspack-csr-css/package.json and then you should see the build crashing.

Coly010:
🤦 i think I might know why this is happening.

Thanks for the investigation @vz-tl

faileon:
Was there any progress on this? I'm yet again trying to move from webpack to rspack a year later but still running into the same issues :)

faileon:
Current workaround as pnpm patch:

```patch
[...]

PR #34195, opened 2026-01-23
fix(angular-rspack): exclude .json files from JS/TS regex patterns

faileon:
## Current Behavior

When importing a `package.json` file in an Angular application built with `@nx/angular-rspack`, the build fails with a Babel syntax error if the `package.json` contains `@angular/*` dependencies:

```
SyntaxError: /path/to/package.json: Missing semicolon. (2:10)

  1 | {
> 2 |     "name": "@org/app",
    |           ^
  3 |     "version": "4.0.2",
  4 |     "dependencies": {
  5 |         "@angular/platform-browser": "20.3.7",
```

This happens because the `JS_ALL_EXT_REGEX` pattern `/\.[cm]?(js)[^x]?\??/` incorrectly matches `.json` files. When the JSON file content contains `@angular` strings, the `angular-partial-transform-loader` attempts to process it through Babel, which fails because JSON is not valid JavaScript.

**Root cause:** The regex `[^x]?` (optional character that is NOT 'x') allows `.json` to match because 'o' is not 'x'.

## Expected Behavior
- `.json` files should NOT match `JS_ALL_EXT_REGEX` or `TS_ALL_EXT_REGEX`
- Importing `package.json` in Angular applications should work correctly
- All existing matches for `.js`, `.jsx`, `.mjs`, `.cjs` (and TypeScript equivalents) should continue to work

## Related Issue(s)
https://github.com/nrwl/nx/issues/32649

### DIFF
--- a/packages/angular-rspack-compiler/src/utils/regex-filters.spec.ts
+++ b/packages/angular-rspack-compiler/src/utils/regex-filters.spec.ts
@@ -16,7 +16,7 @@ describe('TS_ALL_EXT_REGEX', () => {
     expect(filename).toMatch(TS_ALL_EXT_REGEX);
   });
 
-  it.each([['file.js'], ['file.cjs'], ['file.mjs'], ['file']])(
+  it.each([['file.js'], ['file.cjs'], ['file.mjs'], ['file'], ['file.json']])(
     'should not match other files',
     (filename) => {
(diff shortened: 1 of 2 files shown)


leosvelperez (approved):
Thanks!

Issue #32647, opened 2025-06-03
angular-rspack: configured fileReplacements not applied


vz-tl:
Hi @Coly010 ,

first a big, big thank you for the great efforts you spent on implementing this plug-in! I watched the progress already for a while, before I finally decided to migrate our enterprise-scale app from custom webpack to angular-rspack. This more or less worked (still, biggest blocker is rebuild slowness - I already commented [here](https://github.com/nrwl/angular-rspack/issues/18#issuecomment-2930990682) ), but I had to apply a LOT of fixes and workarounds, I'd like to share now here, in order to help improving the plugin. I'll file them all in separate issues to be able for tracking them individually.

The first one, is that apparently the `fileReplacements` aren't applied. I just configured them like it was in the project.json, but this didn't work. I tried all sort of filename variations (relative from current directory, relative from project root directory, etc. ), nothing worked. Finally I worked around by dropping this prop and configuring it via `NormalModuleReplacementPlugin` plugin. This worked as in that case as first param the file to replace has to be applied as a RegExp. 

Hence my suspicion is that, the error relates to applying a full file-path (under the hood it'll be converted to an absolute one) but actually the imports to be replaced never will match with this absolute path.

devanshalite:
Hi @Coly010 ,

We're also facing the same issue described by @vz-tl regarding fileReplacements not being applied when using Rspack. We've configured them in project.json just like with Webpack, but during the build, the replacements are not taking effect.

We also verified the same setup with ESBuild, where fileReplacements work flawlessly, so this does appear to be specific to Rspack.

Hoping this gets some attention, and happy to help verify any fixes or suggestions.

Coly010:
@vz-tl Can you provide a reproduction?

@devanshalite If you've configured them within project.json, it wont be applied. Angular Rspack reads everything from the `rspack.config` file

vz-tl:
@Coly010 , here we go: https://github.com/vz-tl/rspack-issues-repro

faileon:
Current workaround as a pnpm patch until the PR is resolved. We are slowly migrating our production setup to rspack and this is one of the last few blockers.
```patch
diff --git a/dist/lib/config/config-utils/common-config.js b/dist/lib/config/config-utils/common-config.js
index 9f63e2bec1f8477dc6555ab5e3b3e76521edbf74..bf6dcaf978b896e9cd589d8961f646f7549b51d6 100644
--- a/dist/lib/config/config-utils/common-config.js
[...]

PR #34197, opened 2026-01-23
fix(angular-rspack): add fileReplacements to resolve.alias

faileon:
## Current Behavior

The `fileReplacements` option in `@nx/angular-rspack` is passed to the Angular AOT compiler but not added to rspack's `resolve.alias` configuration. This means file replacements only work during TypeScript compilation, not during module resolution/bundling.

## Expected Behavior

File replacements should work consistently, replacing modules both during compilation and bundling - matching the behavior of `@nx/rspack`.

## Related Issue(s)

https://github.com/nrwl/nx/issues/32647


leosvelperez (approved):
Thank you!

PR #34224, opened 2026-01-27
fix(core): do not throw error if worker.stdout is not instanceof socket

drewteachout:
deno worker.stdout is a Readable/Writeable. To provide better deno support an error should not be thrown

## Current Behavior

If `worker.stdout` or `worker.stderr` are not instanceof Socket then nx throws an error. This is problematic in Deno where `stdout` and `stderr` are Readable/Writable and not Socket.

## Expected Behavior

`startupPluginWorker` function should work in Deno runtime

## Related Issue(s)

- https://github.com/denoland/deno/issues/31961
- https://github.com/oven-sh/bun/issues/26505

### DIFF
--- a/packages/nx/src/project-graph/plugins/isolation/plugin-pool.ts
+++ b/packages/nx/src/project-graph/plugins/isolation/plugin-pool.ts
@@ -506,17 +506,9 @@ async function startPluginWorker(name: string) {
   worker.unref();
   if (worker.stdout instanceof Socket) {
     worker.stdout.unref();
-  } else {
-    throw new Error(
-      `Expected worker.stdout to be an instance of Socket, but got ${getTypeName(worker.stdout)}`
-    );
   }
   if (worker.stderr instanceof Socket) {
     worker.stderr.unref();
-  } else {
-    throw new Error(
-      `Expected worker.stderr to be an instance of Socket, but got ${getTypeName(worker.stderr)}`
-    );
   }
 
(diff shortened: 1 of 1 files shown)


AgentEnder:
Hello 👋  - this is a bug in both Bun and Deno it turns out. I've created reproductions and issue reports for them. I'll update this PR to dodge the check where necessary, but there could still be a few dragons as neither are officially supported runtimes.

Bun and Deno are typically treated in the same way .NET CLI is treated for Nx. Nx itself runs in `node`, but if you have tasks that should be ran with bun or deno just configure them to do so at the task level. `bun` goes a bit further as its an officially supported **_package manager_**, but still not for runtime. 

This isnt to say it won't work, or that it shouldn't work, but generally the stance is that if something's breaking when using them to run nx itself, that points to a bug in bun or deno, not Nx.

- https://github.com/denoland/deno/issues/31961
- https://github.com/oven-sh/bun/issues/26505

drewteachout:
Hey 👋, thanks for the response.

Makes perfect sense to me that nx doesn't officially support deno or bun. Thanks for making those issues.

I like what you did checking the runtime and warning instead of strictly erroring.

Am new here, what does it take to get this merged and released? Other than fix the 3 failing checks

AgentEnder:
I'll get it the rest of the way over the line, thanks for kicking it off :)

AgentEnder (approved):

PR #34226, opened 2026-01-27
fix(core): show the correct status for stopped continuous tasks

### LINKED ISSUE #33561: postTasksExecution hook does not consistently include continuous tasks

leosvelperez:
### 1. Continuous tasks missing from `postTasksExecution` hook

When running continuous tasks (e.g., `nx serve app`) and stopping them with Ctrl+C, the `postTasksExecution` lifecycle hook does not include them in `taskResults`. This breaks plugins that rely on post-run statistics (e.g., uploading task stats to DataDog).

### 2. Confusing TUI status when sibling continuous task exits

When multiple continuous tasks run together and one exits unexpectedly, its sibling is marked as "failed" even though it was intentionally terminated/stopped by the task orchestrator.

## Expected Behavior

1. All tasks, including continuous ones, are included in `taskResults` for the `postTasksExecution` hook
2. Continuous tasks that are intentionally stopped (because dependent tasks completed or during graceful shutdown) report as `success` with `Stopped` display status
3. Continuous tasks that exit unexpectedly (crash) report as `failure`
4. TUI summary shows correct status: success when all tasks completed successfully, square icon for stopped tasks

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/33561

Supersedes:

- https://github.com/nrwl/nx/pull/33562
- https://github.com/nrwl/nx/pull/34132

### DIFF
--- a/packages/nx/src/executors/run-commands/running-tasks.ts
+++ b/packages/nx/src/executors/run-commands/running-tasks.ts
@@ -507,7 +507,10 @@ class RunningNodeProcess implements RunningTask {
         cb(1, terminalOutput);
       }
     });
-    this.childProcess.on('exit', (code) => {
+    this.childProcess.on('exit', (code, signal) => {
+      if (code === null) {
+        code = signalToCode(signal);
+      }
       if (!this.readyWhenStatus.length || isReady(this.readyWhenStatus)) {
         const terminalOutput = this.terminalOutputChunks.join('');
         this.terminalOutputChunks = [];
--- a/packages/nx/src/native/tasks/task_history.rs
+++ b/packages/nx/src/native/tasks/task_history.rs
@@ -75,7 +75,7 @@ impl NxTaskHistory {
         self.db
             .prepare(
                 "SELECT hash from task_history
-                    WHERE hash IN rarray(?1)
+                    WHERE hash IN rarray(?1) AND status != 'stopped'
                     GROUP BY hash
                     HAVING COUNT(DISTINCT code) > 1
                 ",
--- a/packages/nx/src/native/tui/components/tasks_list.rs
+++ b/packages/nx/src/native/tui/components/tasks_list.rs
@@ -166,6 +166,7 @@ impl std::str::FromStr for TaskStatus {
             "success" => Ok(Self::Success),
             "failure" => Ok(Self::Failure),
             "skipped" => Ok(Self::Skipped),
+            "stopped" => Ok(Self::Stopped),
             "local-cache-kept-existing" => Ok(Self::LocalCacheKeptExisting),
             "local-cache" => Ok(Self::LocalCache),
             "remote-cache" => Ok(Self::RemoteCache),
--- a/packages/nx/src/tasks-runner/forked-process-task-runner.ts
+++ b/packages/nx/src/tasks-runner/forked-process-task-runner.ts
@@ -3,8 +3,6 @@ import { writeFileSync } from 'fs';
 import { join } from 'path';
 import { ProjectGraph } from '../config/project-graph';
 import { Task, TaskGraph } from '../config/task-graph';
-import { RustPseudoTerminal } from '../native';
-import { signalToCode } from '../utils/exit-codes';
 import { output } from '../utils/output';
 import { stripIndents } from '../utils/strip-indents';
 import { BatchMessageType } from './batch/batch-messages';
@@ -437,23 +435,6 @@ export class ForkedProcessTaskRunner {
       this.cleanup();
       process.off('message', messageHandler);
     });
-    process.once('SIGINT', () => {
-      this.cleanup('SIGTERM');
-      process.off('message', messageHandler);
-      // we exit here because we don't need to write anything to cache.
-      process.exit(signalToCode('SIGINT'));
-    });
-    process.once('SIGTERM', () => {
-      this.cleanup('SIGTERM');
-      process.off('message', messageHandler);
-      // no exit here because we expect child processes to terminate which
-      // will store results to the cache and will terminate this process
-    });
-    process.once('SIGHUP', () => {
-      this.cleanup('SIGTERM');
-      process.off('message', messageHandler);
-      // no exit here because we expect child processes to terminate which
-      // will store results to the cache and will terminate this process
-    });
+    // Signal handlers removed - TaskOrchestrator owns signal handling
   }
 }
--- a/packages/nx/src/tasks-runner/life-cycles/tui-summary-life-cycle.spec.ts
+++ b/packages/nx/src/tasks-runner/life-cycles/tui-summary-life-cycle.spec.ts
@@ -246,6 +246,18 @@ describe('getTuiTerminalSummaryLifeCycle', () => {
 
       lifeCycle.startTasks?.([target], null as unknown as TaskMetadata);
       lifeCycle.appendTaskOutput?.(target.id, 'I was a happy dev server', true);
+      // With the new flow, continuous tasks are properly ended via endTasks
+      lifeCycle.endTasks?.(
+        [
+          {
+            code: 0,
+            status: 'success',
+            task: target,
+            terminalOutput: 'I was a happy dev server',
+          },
+        ],
+        null as unknown as TaskMetadata
+      );
       lifeCycle.setTaskStatus?.(target.id, NativeTaskStatus.Stopped);
       lifeCycle.printTaskTerminalOutput?.(
         target,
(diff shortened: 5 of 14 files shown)


comp615:
I tried to have amp apply this as a patch to our repo, and wasn't seeing the right behavior still. It's entirely possible it didn't apply right since it's larger, but was seeing:
- `Task "blah:watch-deps" is continuous but exited with code 128` (works as "Cancelled" in TUI, but still doesn't output `watch-deps` in the result)
- No post hook output when multiple continuous tasks running (`watch-deps -> serve`)

AI claims that perhaps the handler for non-TUI interrupt codes is missing now...

Is it possible to make a prerelease version that I can try plugging into the repo so it simplifies and removes some variables?

comp615:
@leosvelperez thanks! Did a bit more testing with that version and generally confirmed:
- With TUI enabled, in our prod repo, I still do not see (ex) `watch-deps` in the task list as I would expect to, I only see the `build-deps` task
  - However, my playground/test repo does seem happy with the change although it doesn't have real tasks (just simulated continuous things so the cleanup may be less complicated).
- It does not seem to work with interrupt signals **at all now** (Ctrl-C in non-TUI mode), no tasks are generated and the post hook is never called now (previously it was called but missing tasks) including in the test repo
- I still see the `Task "pacakge:watch-deps" is continuous but exited with code 128` in some cases.

Amp once again attempted to fix this by removing various interrupt exit handlers and returning control to the orchestrator which improves the test repo, but not our prod repo (and may still cause issues):
<details>

<summary>AI suggested changes / flow control issues</summary>

NOTE: This has that previously mentioned issue of not showing them as cancelled correctly. So still not right.
```
➜  nx-playground cat patches/nx.patch
diff --git a/src/executors/run-commands/running-tasks.js b/src/executors/run-commands/running-tasks.js
index 1d6d40dcbd9ee75e604370f4bf689d2918c3ceb6..baee4046edf13dcc6341035eb06a3e40054a24c3 100644
--- a/src/executors/run-commands/running-tasks.js
+++ b/src/executors/run-commands/running-tasks.js
@@ -341,8 +341,7 @@ class RunningNodeProcess {
         });
         process.on('SIGINT', () => {
             this.kill('SIGTERM');
-            // we exit here because we don't need to write anything to cache.
-            process.exit((0, exit_codes_1.signalToCode)('SIGINT'));
+            // Don't exit - let orchestrator handle cleanup and call postTasksExecution
         });
         process.on('SIGTERM', () => {
             this.kill('SIGTERM');
@@ -474,8 +473,7 @@ function registerProcessListener(runningTask, pseudoTerminal) {
     });
     process.on('SIGINT', () => {
         runningTask.kill('SIGTERM');
-        // we exit here because we don't need to write anything to cache.
-        process.exit((0, exit_codes_1.signalToCode)('SIGINT'));
+        // Don't exit - let orchestrator handle cleanup and call postTasksExecution
     });
     process.on('SIGTERM', () => {
         runningTask.kill('SIGTERM');
diff --git a/src/tasks-runner/forked-process-task-runner.js b/src/tasks-runner/forked-process-task-runner.js
index 49e385ae8a0c3aa369cc6affbbfd5f9f630dbe68..82480b9d7644139def954412b5a2b760f780ba7a 100644
--- a/src/tasks-runner/forked-process-task-runner.js
+++ b/src/tasks-runner/forked-process-task-runner.js
@@ -157,9 +157,7 @@ class ForkedProcessTaskRunner {
             terminalOutput += msg;
         });
         p.onExit((code) => {
-            if (!this.tuiEnabled && code > 128) {
-                process.exit(code);
-            }
+            // Don't exit on signal codes - let orchestrator handle cleanup
             this.pseudoTerminals.delete(pseudoTerminal);
             this.processes.delete(p);
             if (!streamOutput) {
diff --git a/src/tasks-runner/task-orchestrator.js b/src/tasks-runner/task-orchestrator.js
index ab7b52212b0c2d3b27172c7c0c26aa009f286a5f..113c58de91aacb74a97d86357367c8b1b59a8a3f 100644
--- a/src/tasks-runner/task-orchestrator.js
+++ b/src/tasks-runner/task-orchestrator.js
@@ -134,11 +134,7 @@ class TaskOrchestrator {
                 if (this.initializingTaskIds.has(task.id)) {
                     await new Promise((res) => {
                         runningTask.onExit((code) => {
-                            if (!this.tuiEnabled) {
-                                if (code > 128) {
-                                    process.exit(code);
-                                }
-                            }
+                            // Don't exit on signal codes - let orchestrator handle cleanup
                             res();
                         });
                     });
@@ -802,7 +798,9 @@ class TaskOrchestrator {
     }
     setupSignalHandlers() {
         process.once('SIGINT', () => {
-            this.cleanup().finally(() => process.exit((0, exit_codes_1.signalToCode)('SIGINT')));
+            // Don't exit here - just cleanup and let the orchestrator flow return
+            // so that run-command.js can call postTasksExecution
+            this.cleanup();
         });
         process.once('SIGTERM', () => {
             this.cleanup();
```
</details>

leosvelperez:
@comp615 those `process.exit` are intentional. They are meant to respect those termination signals and interrupt everything else (you can see the comment in those AI diffs). That didn't change with the PR. It's existing and intentional behavior.

That said, I understand it's a bit nuanced. I'll bring this up internally, and we'll discuss it. I'll update here with the outcome.

>It does not seem to work with interrupt signals at all now (Ctrl-C in non-TUI mode), no tasks are generated, and the post hook is never called now (previously it was called but missing tasks), including in the test repo

I'm sure that didn't work before either. As mentioned above, that shouldn't have changed with this PR. Could you provide a repro with the latest version of Nx where hitting Ctrl-C in non-TUI mode results in the post tasks execution hook being called?

>- With TUI enabled, in our prod repo, I still do not see (ex) watch-deps in the task list as I would expect to, I only see the build-deps task
>    - However, my playground/test repo does seem happy with the change although it doesn't have real tasks (just simulated continuous things so the cleanup may be less complicated).

I'll also need some repro for this. So far, things have worked on my testing, but I might be missing a scenario/setup you have. Probably similar to how your playground is working, but not your production repo. Please, try to identify what's different and what it would take to reproduce it.

 > I still see the Task "pacakge:watch-deps" is continuous but exited with code 128 in some cases.

I have a fix for this one, which I'll be pushing soon. Continuous tasks terminated by signals (SIGTERM, SIGINT, SIGHUP, SIGQUIT) were being reported as failures because their exit codes (128 + signal) weren't recognized as intentional termination. The fix treats those exit codes as expected stops. I also fixed messageToCode to properly handle macOS signal message formats that were falling through to a generic code 128.

---

Thanks again for taking the time to check this! In general, we'll need some repros for these issues so we can properly address them. Please make sure to specify in each report the exact flow in which you see the issues. This is an important detail that would help us track things more quickly, since they all have different exit paths.

comp615:
> I'm sure that didn't work before either. 

I'm on an airplane so it's slow going, but I think I have one: https://github.com/comp615/nx-playground/tree/clean-vite-setup, it's just a simple vite task where it depends on `^build`. When I run the vite task and ctrl-c it, I do see all the build stuff logged in a version of main.
`NX_TUI=false nx run vite-app:serve`

Underscoring that as below, we don't use TUI, and it's crucial for me that we can track stats either way to give devs flexibility based on their preferences (or just for AI-powered workflow).

Grain of salt as ever:
```
Why you WERE seeing post-hooks before (inaccurately)

If post-hooks were being called before, it was likely because:

The discrete tasks had already finished and triggered postRunSteps() → endTasks()
Only the continuous task cleanup was racing with process.exit()
```

> TUI-enabled still doesn't have watch-deps

The issue here appears to be in just our current stats plugin (sorry!), it appears the continuous tasks do not have an `endTime` and we filter those out because we log duration of tasks. Unsure if this is the behavior you want or not. More below...this is probably OK, and explains the disconnect. I verified this with the new branch, in TUI only, whilst fixing our stats logic:
```
    if (task.continuous && !task.endTime) {
      task.endTime = Date.now(); // TODO: use command end time?
    }
```

> Also

Stepping back a bit, I wanted to recenter on my core problem which is that I want to be able to see stats on dev machines for how often, and how long it takes them to "boot" their vite servers (continuous tasks). This is usually done by running `app:dev` or similar. 

We essentially map the list of tasks to stat points. However, we don't really log the overall run. This left me two options:
- Map the argv command to something, and then use first task start and last task end as a guess (because command end is meaningless)
- or what I was envisioning as a sort of easy workaround, is basically look for the "latest" to start TS of a continuous task (guess at the root command, because there's a million argv ways to get the same "task"), and diff that against the earliest start of any task, or just the overall command start. This should yield "time from command start, to dev'ing"
  - I mention this because if the best we can do is to have continuous tasks with start times but no end due to interrupt...that's still valuable!

...obviously there's some problems or fudging still either way (does it include graph gen when needed?), but it gives us something that feels very accurate and close to the dev experience.

FrozenPandaz (approved):

PR #34227, opened 2026-01-27
fix(core): hide already-installed nx packages from suggestion list during nx import

MaxKless:
## Current Behavior
If something is in `package.json#dependencies`, we still suggest it to be `nx add`-ed during `nx import`

## Expected Behavior
If a plugin is already installed, we don't suggest it anymore


juristr (approved):

AgentEnder (approved):

JacobMccaakill (commented):
- - **@**

JacobMccaakill:
> - - **@**

Issue #34228, opened 2026-01-27
@nx/js takes precedence over my plugins, even if not added to `plugins`


mpsanchis:
### Current Behavior

The current flow, when inferring targets for a project, is the following:
1. My nx-react-project plugin runs first and infers nx-release-publish with my custom executor 
2. The default `package-json` plugin runs later 
3. For non-private packages, it adds nx-release-publish with `@nx/js:release-publish` 
4. Since the executors differ, the targets are incompatible 
5. The later target (from `package-json`) overwrites my plugin's target

This happens even if I don't add "@nx/js" to my "plugins" section in `nx.json`

### Expected Behavior

If my plugin infers something, the default behaviour should not override it.

### Steps to Reproduce

1. Create a plugin that infers `nx-release-publish` for projects with a `package.json`
2. Keep "private" to "false" (or do not define it in `package.json`)
3. `nx show project my-project`, and verify targets

### Nx Report

```shell
Node           : 24.0.0
OS             : darwin-arm64
Native Target  : aarch64-macos
pnpm           : 10.12.1

nx                     : 22.2.5
@nx/js                 : 22.2.5
@nx/jest               : 22.2.5
@nx/eslint             : 22.2.5
@nx/workspace          : 22.2.5
@nx/devkit             : 22.2.5
@nx/eslint-plugin      : 22.2.5
@nx/module-federation  : 22.2.5
@nx/playwright         : 22.2.5
@nx/plugin             : 22.2.5
@nx/react              : 22.2.5
@nx/rollup             : 22.2.5
@nx/vite               : 22.2.5
@nx/vitest             : 22.2.5
@nx/web                : 22.2.5
typescript             : 5.9.3
```

### Failure Logs

```shell

```

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

More details:

The issue is in `nx/packages/nx/src/utils/package-json.ts` at lines 236-255.

Plugin Order (`packages/nx/src/project-graph/plugins/get-plugins.ts:58`): 
```typescript
loadedPlugins = specifiedPlugins.concat(defaultPlugins); 
```
User plugins are processed first, then default plugins (including package-json) come after. 

Default Plugins (`get-plugins.ts:217-226`): 
```typescript
function getDefaultPlugins(root: string) { 
  return [ 
    join(__dirname, '../../plugins/js'), 
    // ... 
    join(__dirname, '../../plugins/package-json'), // <-- here
[...]

PR #34229, opened 2026-01-27
fix(core): User specifiedPlugins override defaultPlugins

mpsanchis:
## Current Behavior

See https://github.com/nrwl/nx/issues/34228

## Expected Behavior

See https://github.com/nrwl/nx/issues/34228

## Notes

I would need an expert's eye on this one. Building locally the Nx repo is somehow painful for me (e.g.: needs Gradle, but it's not in `mise.toml`, so building the graph fails), so I have to trust the change without being able to run much locally.

## Fixes

 https://github.com/nrwl/nx/issues/34228

### DIFF
--- a/packages/nx/src/project-graph/plugins/get-plugins.ts
+++ b/packages/nx/src/project-graph/plugins/get-plugins.ts
@@ -55,7 +55,7 @@ export async function getPlugins(
     )),
   ]);
 
-  loadedPlugins = specifiedPlugins.concat(defaultPlugins);
+  loadedPlugins = defaultPlugins.concat(specifiedPlugins);
 
   return loadedPlugins;
 }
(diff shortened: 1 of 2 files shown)


AgentEnder (changes requested):
Hello @mpsanchis - this is an interesting one. I understand your use case, but the package.json / project.json plugins we are adding here override inference plugins intentionally because they are sources of static configuration vs inferred configuration. In the case of the package.json though, the release target is inferred instead of being part of the static configuration so it doesn't necesssarily feel correct that its overriding your plugin's code.

We'll have to think through how to best handle this, but this PR is heading down the wrong direction so I'm going to close this out for now. Thanks for putting in the effort though, if there weren't the larger implications this could concievably be  the correct approach and it does fix your observed issue, just breaks a lot of other things.

mpsanchis:
Thanks for the rapid feedback @AgentEnder and don't hesitate to reach out if I can give a hand with something related

PR #34237, opened 2026-01-27
fix(misc): fix gitlab ci workflow for new repos and merge requests

AgentEnder:
## Current Behavior

The generated GitLab CI workflow (`nx generate @nx/workspace:ci-workflow --ci=gitlab`) fails on new repositories with errors like:

fatal: ambiguous argument 'HEAD~1': unknown revision or path not in the working tree.

This happens because:
1. GitLab's default shallow clone doesn't have full git history
2. `CI_COMMIT_BEFORE_SHA` is always `0000000000000000000000000000000000000000` for merge requests and first commits
3. Environment variables `NX_HEAD` and `NX_BASE` weren't exported, so Nx couldn't read them

## Expected Behavior

The GitLab CI workflow should work correctly on:
- New repositories with no previous commits
- Merge request pipelines
- Main branch push pipelines

## Related Issue(s)

Fixes https://linear.app/nxdev/issue/NXC-3707

## Changes Made

1. **Added `GIT_DEPTH: 0`** - Fetches full git history instead of shallow clone
2. **Changed `only` branch to use template variable** - Uses `<%= mainBranch %>` instead of hardcoded `main`
3. **Fixed `NX_BASE` and `NX_HEAD` exports** - Added `export` keyword and improved fallback logic:
   - Uses `CI_MERGE_REQUEST_DIFF_BASE_SHA` for merge requests
   - Falls back to `HEAD~1` for regular commits
   - Falls back to `HEAD` for initial commits (no parent)

## Testing

Verified on real GitLab CI at `gitlab.com/AgentEnder/nx-gitlab-ci-test`:
- ✅ Main branch push pipeline passed
- ✅ Merge request pipeline passed

### DIFF
--- a/packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
+++ b/packages/workspace/src/generators/ci-workflow/__snapshots__/ci-workflow.spec.ts.snap
@@ -722,6 +722,7 @@ exports[`CI Workflow generator connected to nxCloud with bun should generate git
 "image: node:20
 variables:
   CI: 'true'
+  GIT_DEPTH: 0
 
 # Main job
 CI:
@@ -739,8 +740,8 @@ CI:
     # - bunx nx start-ci-run --distribute-on="3 linux-medium-js" --stop-agents-after="build"
 
     - bun install --no-cache
-    - NX_HEAD=$CI_COMMIT_SHA
-    - NX_BASE=\${CI_MERGE_REQUEST_DIFF_BASE_SHA:-$CI_COMMIT_BEFORE_SHA}
+    - export NX_HEAD=$CI_COMMIT_SHA
+    - export NX_BASE=\${CI_MERGE_REQUEST_DIFF_BASE_SHA:-$(git rev-parse HEAD~1 2>/dev/null || git rev-parse HEAD)}
 
     # Prepend any command with "nx record --" to record its logs to Nx Cloud
     # - bun nx record -- echo Hello World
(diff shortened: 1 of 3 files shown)


jansauer:
Hi @AgentEnder I debugged the same problem yesterday and then found your Issue as I was planning to create a MR for it. Two questions regarding your changes:

**1. How would your change work for first commits?**
If you fall back to HEAD as the base wouldn't affected just compare the commit with it self, find no changes and do nothing? My understanding is that at least the pipeline would not fail but also do nothing.

**2. Why the `GIT_DEPTH: 0`?**
The default depth should be 20 commits. I understand that having MR's with more then that amount of commits would result in failing pipelines but how common is that problem? Shouldn't best practice be to have shorter MR's. This could make builds slower for a lot of teams that would never encounter this problem.

FrozenPandaz (approved):

Stanzilla:
> Hi @AgentEnder I debugged the same problem yesterday and then found your Issue as I was planning to create a MR for it. Two questions regarding your changes:
> 
> **1. How would your change work for first commits?** If you fall back to HEAD as the base wouldn't affected just compare the commit with it self, find no changes and do nothing? My understanding is that at least the pipeline would not fail but also do nothing.
> 
> **2. Why the `GIT_DEPTH: 0`?** The default depth should be 20 commits. I understand that having MR's with more then that amount of commits would result in failing pipelines but how common is that problem? Shouldn't best practice be to have shorter MR's. This could make builds slower for a lot of teams that would never encounter this problem.

Agree on `0` being too extreme. Can do some large random number like `1000` though.

meeroslav (approved):

PR #34249, opened 2026-01-28
feat(core): add Nx Cloud connect URL to template README

jaysoo:
## Current Behavior
Template-generated workspaces use a generic link in the README instead of a per-workspace short link for Nx Cloud setup.

## Expected Behavior
When users opt into Nx Cloud (or are auto-connected via variant 2), the template README is updated with a personalized connect URL section that helps them finish setting up their workspace.

---
BEFORE: 

image

AFTER: 

image

---

## Related Issue(s)
Closes NXC-3783

### DIFF
--- a/packages/create-nx-workspace/src/utils/template/update-readme.ts
+++ b/packages/create-nx-workspace/src/utils/template/update-readme.ts
@@ -1,0 +1,118 @@
+import { existsSync, readFileSync, writeFileSync } from 'node:fs';
+import { join } from 'path';
+import { execAndWait } from '../child-process-utils';
+
+const BEGIN_MARKER = '<!-- BEGIN: nx-cloud -->';
+const END_MARKER = '<!-- END: nx-cloud -->';
+
+const CLOUD_SETUP_SECTION = `## Finish your Nx platform setup
+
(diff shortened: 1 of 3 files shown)


packages/create-nx-workspace/src/utils/template/update-readme.ts
```diff
@@ -0,0 +1,67 @@
...
+  const endIndex = readmeContent.indexOf(END_MARKER);
+
+  if (beginIndex === -1 || endIndex === -1 || beginIndex >= endIndex) {
+    return readmeContent;
+  }
+
+  const section = CLOUD_SETUP_SECTION.replace('{{CONNECT_URL}}', connectUrl);
+
+  return (
+    readmeContent.slice(0, beginIndex + BEGIN_MARKER.length) +
```
FrozenPandaz: Did you intend to keep the markers? I don't think we should keep the markers.. looks weird in people's readmes?

FrozenPandaz (approved):

packages/create-nx-workspace/src/utils/template/update-readme.ts
jaysoo: Yeah, I thought I'd leave it the same as the unconnected case, but it doesn't matter much to me.

jaysoo: I'll push an update to strip this plus make sure commit is after the README updates.

PR #34259, opened 2026-01-29
feat(gradle): add debug env var to gradle batch executor

lourw:
## Current Behavior

- No way to run the batch executor in debug mode
- Any flags passed into the nx gradle batch command get forwarded into the `gradlew` command. 

## Expected Behavior

- Allow for an env variable to be set for debug flags so that the batch runner jar can be run with a debugger hooked in.

## Related Issue(s)

Fixes #NXC-3797


packages/gradle/src/executors/gradle/schema.d.ts
```diff
@@ -5,4 +5,6 @@ export interface GradleExecutorSchema {
   excludeDependsOn: boolean;
   includeDependsOnTasks?: string[];
   __unparsed__?: string[];
+  debugMode?: boolean;
```
FrozenPandaz: Do we need 2 properties for this? if debugPort is undefined or null, debugMode is false right?

packages/gradle/src/executors/gradle/gradle-batch.impl.ts
```diff
@@ -185,14 +182,20 @@ async function runTasksInBatch(
...
   args: string[],
-  root: string
+  root: string,
+  debugMode?: boolean,
+  debugPort?: number
 ): Promise<BatchResults> {
   const gradlewBatchStart = performance.mark(`gradlew-batch:start`);
 
+  const debugArgs = debugMode
+    ? `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:${debugPort ?? 5005}`
```
FrozenPandaz: This is mostly for us right? Not other people's tasks? Maybe we should set this via something like..

```
NX_GRADLE_BATCH_DEBUG="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"
```

FrozenPandaz (changes requested):

packages/gradle/src/executors/gradle/gradle-batch.impl.ts
```diff
@@ -192,7 +192,8 @@ async function runTasksInBatch(
   const usePseudoTerminal =
     process.env.NX_NATIVE_COMMAND_RUNNER !== 'false' &&
     PseudoTerminal.isSupported();
-  const command = `java -jar ${batchRunnerPath} --tasks='${JSON.stringify(
+  const debugOptions = process.env.NX_GRADLE_BATCH_DEBUG ?? '';
+  const command = `java ${debugOptions} -jar ${batchRunnerPath} --tasks='${JSON.stringify(
```
FrozenPandaz:
```suggestion
  const debugOptions = ' ' + process.env.NX_GRADLE_BATCH_DEBUG ?? '';
  const command = `java${debugOptions} -jar ${batchRunnerPath} --tasks='${JSON.stringify(
```

FrozenPandaz (approved):

PR #34267, opened 2026-01-30
fix(nx-dev): make headers and table options linkable

barbados-clemens:
- fix(nx-dev): always link headers regardless of mdoc or markdown content source (generated vs static file)
- fix(nx-dev): make option/property columns in table linkable
    - the table column header is matched on `options`, `option`, `properties`, and property` (case insensitive)

https://github.com/user-attachments/assets/7250b9d5-1030-4ebc-9e21-0a05f295bbf5

Note bc mdoc and `renderMarkdown` go through 2 different rendering pipelines, this logic must bc within the markdoc config and rehype (markdown) processing logic. tried to shared logic where I could


patches/@astrojs__starlight.patch
barbados-clemens, line 1: for content being generated dynamically from nx source, this breaks generating header links bc of the path check. in this case path we allow for a null path and assume it's of 'en' locale to allow the auto link headers of starlight to work.

jaysoo (approved):

AgentEnder (approved):

PR #34271, opened 2026-01-30
docs(gradle): add reference to batch mode

lourw:
## Expected Behavior

Add documentation for batch mode. Remove references to removed custom overrides for intTest.

## Related Issue(s)

Fixes #


astro-docs/src/content/docs/technologies/java/gradle/introduction.mdoc
```diff
@@ -172,6 +172,50 @@ Once a Gradle configuration file has been identified, the targets are created wi
...
+{% tabitem label="CircleCI" %}
+
+```yaml
+env:
+  NX_BATCH_MODE: true
+
+jobs:
+  main:
+    # ... rest of your workflow
+```
```
barbados-clemens: I don't think you need to provide examples of how to set env vars per CI provider. ppl know how to google their CI providers syntax. you can just say "set this env var for your CI job" 

only "gotcha" will be people using per step env vars vs job env vars when using start-ci-run. e.g. not set for start-ci-run but is for the nx run-many command. it wouldn't be sent to Nx Agents right?

jaysoo (approved):

PR #34276, opened 2026-02-01
fix(js): correctly validate buildable packages when outDir is outside…

### LINKED ISSUE #34243: Nx JS TypeScript Plugin Not Inferring Tasks
## Environment
- **Nx version:** 22.3.3
- **@nx/js version:** 22.3.3
- **TypeScript version:** 5.8.3-5.9.3
- **Workspace:** npm workspaces with TypeScript project references
- **Packages:** 11 TypeScript packages (buildable libraries)

## Problem
The `@nx/js/typescript` plugin is not inferring build or typecheck tasks. All tasks continue to use `nx:run-script` executor (inferred from package.json scripts) instead of `nx:run-commands` executor with proper caching configuration.

## Non-Standard Configuration
Our workspace uses `.config/` directory for Nx and TypeScript configs:
- `.config/nx.json` (not `./nx.json`)
- `.config/tsconfig.base.json` (not `./tsconfig.base.json`)
- Packages extend: `"extends": "../../.config/tsconfig.base.json"`

**Question:** Does the plugin require tsconfig.base.json at workspace root, or should it work with custom config directory structure?

## Workaround (Confirms Expected Behavior)
Creating manual `project.json` with `nx:run-commands` executor DOES work:
- Caching functions correctly
- Cache-hit/cache-miss behaves as expected
- Proves the issue is plugin inference, not caching mechanism

## Expected Behavior
Plugin should automatically:
1. Detect buildable libraries (exports → dist/)
2. Create build/typecheck targets with `nx:run-commands` executor
3. Configure inputs/outputs for caching
4. Register `@nx/js:typescript-sync` generator

### GitHub Repo

_No response_

## Evidence
1. **Plugin is registered** in `.config/nx.json`:
   ```json
   {
     "plugins": [
       {
         "options": {
           "typecheck": { "targetName": "check", "configName": "tsconfig.check.json" },
           "build": { "targetName": "build", "configName": "tsconfig.json" }
         }
       }
     ]
   }
   ```

2. **All packages have required setup:**
   - ✅ `tsconfig.json` exists (runtime config)
   - ✅ `package.json` has `exports` pointing to `./dist/*` (buildable)
   - ✅ `composite: true` and `declaration: true` in base config
   - ✅ TypeScript project references configured

3. **But targets show wrong executor:**
   ```bash
   npx nx show project domains --json | jq '.targets.build.executor'
   # Returns: "nx:run-script"  (should be "nx:run-commands")
   ```

4. **No sync generator registered:**
   ```bash
   npx nx sync
   # Output: "There are no sync generators to run"
   # Expected: @nx/js:typescript-sync should run
   ```

5. **Test: Removing build script doesn't trigger plugin:**
   - Removed `"build": "tsc -b"` from package.json
   - Result: NO build target created at all
   - Expected: Plugin should infer build target from tsconfig.json

### Nx Report

```shell

Node           : 22.18.0
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.9.0

nx             : 22.3.3
@nx/js         : 22.3.3
@nx/workspace  : 22.3.3
@nx/devkit     : 22.3.3
typescript     : 5.9.3
Cache Usage: 0.00 B / 95.23 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

## Non-Standard Configuration
Our workspace uses `.config/` directory for Nx and TypeScript configs:
- `.config/nx.json` (not `./nx.json`)
- `.config/tsconfig.base.json` (not `./tsconfig.base.json`)
- Packages extend: `"extends": "../../.config/tsconfig.base.json"`

**Question:** Does the plugin require tsconfig.base.json at workspace root, or should it work with custom config directory structure?

## Workaround (Confirms Expected Behavior)
Creating manual `project.json` with `nx:run-commands` executor DOES work:
- Caching functions correctly
- Cache-hit/cache-miss behaves as expected
- Proves the issue is plugin inference, not caching mechanism

Nitin75408:
… project root

Previously, `isAnyEntryPointPointingToOutDir` in the TypeScript plugin assumed that if `outDir` resolves to a path outside the project root, the package is automatically buildable. This caused incorrect behavior when users have base tsconfig in non-standard directories (like `.config/`), because TypeScript resolves `outDir` relative to the config file containing it.

The fix removes the short-circuit logic and always validates that package.json exports actually point to the resolved outDir.

Closes #34243

## Current Behavior
The `isAnyEntryPointPointingToOutDir` function in the `@nx/js/typescript` plugin assumes that if `outDir` resolves to a path outside the project root, the package is automatically buildable. This causes incorrect behavior when:

1. Users have base `tsconfig.json` in non-standard directories (like `.config/`)
2. The base config has `outDir: "dist"` which TypeScript resolves relative to the base config directory
3. The resolved `outDir` becomes `.config/dist` instead of `packages/mylib/dist`
4. The plugin incorrectly marks packages as buildable even when their `package.json` exports don't point to the actual output directory

## Expected Behavior
The plugin should always validate that `package.json` exports actually point to files inside the resolved `outDir`, regardless of whether the `outDir` is inside or outside the project root.

With this fix:
- Packages with exports pointing to the correct `outDir` are correctly identified as buildable
- Packages with exports pointing to source files (not `outDir`) are correctly identified as non-buildable
- Users with non-standard config directories (like `.config/`) get correct build target inference when they properly override `outDir` in their project's tsconfig

## Related Issue(s)

Fixes #34243

### DIFF
--- a/packages/js/src/plugins/typescript/plugin.spec.ts
+++ b/packages/js/src/plugins/typescript/plugin.spec.ts
@@ -6092,6 +6092,232 @@ describe(`Plugin: ${PLUGIN_NAME}`, () => {
       `);
     });
   });
+
+  describe('non-standard config directories (issue #34243)', () => {
+    it('should infer build target when tsconfig extends from a non-standard directory like .config/', async () => {
+      configFiles = await applyFilesToTempFsAndContext(tempFs, context, {
+        // Base tsconfig in .config directory (non-standard location)
+        '.config/tsconfig.base.json': JSON.stringify({
+          compilerOptions: {
+            target: 'ES2022',
+            module: 'NodeNext',
+            moduleResolution: 'NodeNext',
+            composite: true,
+            declaration: true,
+            outDir: 'dist',
+            rootDir: 'src',
+            strict: true,
+          },
+        }),
+        // Project tsconfig that extends the base
+        'packages/domains/tsconfig.json': JSON.stringify({
+          extends: '../../.config/tsconfig.base.json',
+          compilerOptions: {
+            outDir: 'dist',
+            rootDir: 'src',
+          },
+          include: ['src/**/*'],
+        }),
+        'packages/domains/package.json': JSON.stringify({
+          name: '@repro/domains',
+          exports: {
+            '.': {
+              import: './dist/index.js',
+              types: './dist/index.d.ts',
+            },
+          },
+        }),
+        'packages/domains/src/index.ts':
+          'export const hello = () => "Hello World";',
+      });
+
+      const result = await invokeCreateNodesOnMatchingFiles(
+        configFiles,
+        context,
+        {
+          typecheck: { targetName: 'check' },
+          build: { targetName: 'build', configName: 'tsconfig.json' },
+        }
+      );
+
+      // Should have both typecheck and build targets
+      expect(result.projects['packages/domains'].targets.check).toBeDefined();
+      expect(result.projects['packages/domains'].targets.build).toBeDefined();
+      expect(result.projects['packages/domains'].targets.build.command).toBe(
+        'tsc --build tsconfig.json'
+      );
+    });
+
+    it('should NOT infer build target when outDir is only defined in extended base config (TypeScript resolves outDir relative to base config)', async () => {
+      // When outDir is defined in a base config at a different directory (like .config/),
+      // TypeScript resolves it relative to that directory. So outDir: "dist" in .config/tsconfig.base.json
+      // resolves to ".config/dist", NOT "packages/lib/dist". Users must override outDir in their project config.
+      configFiles = await applyFilesToTempFsAndContext(tempFs, context, {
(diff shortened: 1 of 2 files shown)


jaysoo (changes requested):
I understand the issue — `isAnyEntryPointPointingToOutDir()` incorrectly returns `true` when `outDir` is outside the project root (path starts with `..`).

However, removing the `if (relativePath.startsWith('..')) return true` shortcut entirely is too broad. This regresses legitimate cases where monorepos use shared output directories (e.g., `outDir: "../../dist/packages/lib"`), which is a common pattern.

A more targeted fix would be to check whether the `outDir` actually resolves to the project's build output directory rather than just checking if the relative path goes up. For example, comparing the resolved absolute paths of the entry points against the resolved `outDir`.

jaysoo:
Closing this as it is not the right fix.

PR #34285, opened 2026-02-02
feat(core): add support for '...' as a spread token when merging target config

AgentEnder:
## Current Behavior

When merging project configurations (e.g., from plugins, `project.json`, target defaults in `nx.json`), array and object properties are completely replaced by the new value. There is no way to extend or merge with the base value.

For example, if a plugin infers:
```json
{
  "targets": {
    "build": {
      "inputs": ["default", "{projectRoot}/**/*"]
    }
  }
}
```

And `nx.json` has target defaults:
```json
{
  "targetDefaults": {
    "build": {
      "inputs": ["production"]
    }
  }
}
```

The result would be `["production"]` — completely replacing the inferred inputs rather than combining them.

## Expected Behavior

This PR adds support for `"..."` as a spread token when merging configurations. Users can now control how arrays and objects are merged by specifying where the base value should be inserted.

**Array spread:**
```json
{
  "inputs": ["production", "...", "{workspaceRoot}/.eslintrc.json"]
}
```
Results in: `["production", "default", "{projectRoot}/**/*", "{workspaceRoot}/.eslintrc.json"]`

**Object spread:**
```json
{
  "options": {
    "env": {
      "NEW_VAR": "value",
      "...": true,
      "OVERRIDE_VAR": "overridden"
    }
  }
}
```
Spreads the base object's properties at the position of `"..."`, with keys defined after the spread taking precedence.

This works in:
- Top-level target properties (`inputs`, `outputs`, `dependsOn`)
- Target `options` and nested option objects (one layer deep)
- Target `configurations` and their options (one layer deep)
- Both `project.json` merging and `nx.json` target defaults

## Related Issue(s)

Fixes #


packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -986,18 +1087,34 @@ export function mergeTargetConfigurations(
...
-  if (projectConfigSourceMap) {
-    projectConfigSourceMap[targetIdentifier] = sourceInformation;
+  const result: Partial<TargetConfiguration> = {};
+  const mergeBase = isCompatible ? baseTargetProperties : {};
+
+  const keys = isCompatible
+    ? [
+        ...Array.from(
+          new Set<string>([
+            ...Object.keys(isCompatible ? baseTargetProperties : {}),
```
FrozenPandaz: simplify this since we're already compatible here

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -986,18 +1087,34 @@ export function mergeTargetConfigurations(
...
+          ])
+        ),
+      ]
+    : Object.keys(target);
+
+  for (const key of keys) {
+    // Only apply merge logic if the key exists in target (has potential spread syntax)
+    // Otherwise, use base value or target value depending on which has it
+    if (key in target) {
+      result[key] = getMergeValueResult(mergeBase[key], target[key]);
```
FrozenPandaz: We should not do this for options and configurations which don't get completely overridden

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -1125,15 +1243,23 @@ function mergeOptions(
   sourceInformation?: SourceInformation,
   targetIdentifier?: string
 ): Record<string, any> | undefined {
-  const mergedOptions = {
+  const mergedOptionKeys = Object.keys({
```
FrozenPandaz: this could be a util function

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -865,6 +865,102 @@ function deepClone(obj) {
...
+    return baseValue;
+  }
+  if (
+    // Different types
+    typeof baseValue !== typeof newValue ||
+    // one is truthy, one is falsy
+    !!baseValue !== !!newValue
+  ) {
+    return newValue;
+  }
```
FrozenPandaz: Please leave a better comment here.. and if it is possible... simplify it because I am struggling to understand it.

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -865,6 +865,102 @@ function deepClone(obj) {
...
+
+  const newIsArray = Array.isArray(newValue);
+  if (newIsArray) {
+    const spreadIndex = newValue.findIndex((v) => v === '...');
```
FrozenPandaz: Have a variable like `NX_SPREAD_TOKEN`

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -865,6 +865,102 @@ function deepClone(obj) {
...
+      const baseArray = baseIsArray ? baseValue : [];
+      const clone = [...newValue];
+      // delete 1 element at spreadIndex and insert all elements from baseArray in its place
+      clone.splice(spreadIndex, 1, ...baseArray);
+      return clone;
+    }
+  } else if (
+    typeof newValue === 'object' &&
+    newValue != null &&
+    newValue['...']
```
FrozenPandaz:
```suggestion
    newValue['...'] === true
```

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -865,6 +865,102 @@ function deepClone(obj) {
...
+  ) {
+    const baseIsObject = typeof baseValue === 'object' && baseValue != null;
+    const baseObj = baseIsObject ? (baseValue as any) : {};
+
+    const res: any = {};
+    for (const newKey in newValue) {
+      if (newValue[newKey] === '...') {
```
FrozenPandaz:
```suggestion
      if (newKey !== '...') {
```

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -865,6 +865,102 @@ function deepClone(obj) {
   return JSON.parse(JSON.stringify(obj));
 }
 
+function getMergeValueResult(baseValue: any, newValue: any): any {
```
FrozenPandaz: Consider also updating the source information here since this seems like a clear spot where we know whether the base or target value is being used.

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -865,6 +865,102 @@ function deepClone(obj) {
...
+  sourceMapKeyBase: string,
+  sourceMap: Record<string, SourceInformation>
+) {
+  const result = {
+    ...optionsFromTargetDefinition,
+  };
+  for (const optionKey in optionsFromTargetDefaults) {
+    const sourceMapKey = `${sourceMapKeyBase}.${optionKey}`;
+    // Note: definedSourceInfoComesFromCorePlugins returns true when the value was defined
+    // by a NON-core plugin (i.e., plugins that don't start with 'nx/'). In that case,
```
FrozenPandaz: This is contradictory... it seems to read that being defined by a core plugin means.. it is not by a core plugin.

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -882,16 +978,22 @@ export function mergeTargetDefaultWithTargetDefinition(
...
-        for (const optionKey in normalizedDefaults) {
-          const sourceMapKey = `targets.${targetName}.options.${optionKey}`;
-          if (
-            targetDefinition.options[optionKey] === undefined ||
-            targetDefaultShouldBeApplied(sourceMapKey, sourceMap)
-          ) {
-            result.options[optionKey] = targetDefault.options[optionKey];
-            sourceMap[sourceMapKey] = ['nx.json', 'nx/target-defaults'];
+        const keysFromCorePlugins = new Set<string>();
+        const keysFromOtherPlugins = new Set<string>();
```
FrozenPandaz: Are these unused?

packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
```diff
@@ -457,6 +457,297 @@ describe('project-configuration-utils', () => {
...
+            }
+          );
+          expect(result.inputs).toEqual([
+            'new-input',
+            'base-input-1',
+            'base-input-2',
+          ]);
+        });
+
+        it('should handle "..." when base array is empty', () => {
```
FrozenPandaz: Should we have a case where the base "array" is undefined?

packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
```diff
@@ -457,6 +457,297 @@ describe('project-configuration-utils', () => {
...
+
+      describe('object spread with "..." in nested options', () => {
+        it('should spread base object properties in nested option objects', () => {
+          const result = mergeTargetConfigurations(
+            {
+              executor: 'target',
+              options: {
+                env: {
+                  NEW_VAR: 'new',
+                  '...': '...',
```
FrozenPandaz: This is.. not what we want people to do.. it should be 
```suggestion
                  '...': true,
```

That said, also add a test when `"...": "This is a comment"`. It should keep the comment value

packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
```diff
@@ -2233,6 +2524,383 @@ describe('project-configuration-utils', () => {
       // not be replaced by target default
       expect(result.dependsOn).toEqual([]);
     });
+
+    describe('spread syntax', () => {
+      it('should spread arrays in target default options using "..."', () => {
+        const sourceMap: Record<string, SourceInformation> = {
```
FrozenPandaz: Double check that this sourceMap looks accurate

packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
```diff
@@ -2233,6 +2524,383 @@ describe('project-configuration-utils', () => {
       // not be replaced by target default
       expect(result.dependsOn).toEqual([]);
     });
+
+    describe('spread syntax', () => {
```
FrozenPandaz: Add tests for when the project configuration (from project.json) should override target defaults

FrozenPandaz (changes requested):

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -892,25 +892,154 @@ export function validateProject(
...
+  for (const optionKey in optionsFromTargetDefaults) {
+    const sourceMapKey = `${sourceMapKeyBase}.${optionKey}`;
+    if (!definedSourceInfoComesFromCorePlugins(sourceMapKey, sourceMap)) {
+      // Property was defined by a non-core plugin, so target defaults take precedence
+      result[optionKey] = getMergeValueResult(
+        optionsFromTargetDefinition[optionKey], // base
+        optionsFromTargetDefaults[optionKey] // overrides
+      );
+      // Update source map since target defaults is winning
+      sourceMap[sourceMapKey] = ['nx.json', 'nx/target-defaults'];
```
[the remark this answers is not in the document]
AgentEnder: This is wrong, but points out a real and kinda interesting problem - the `...` token in package.json / project.json would expand th values from the last specified plugin, not target defaults, because of the way target defaults are applied as an interim post processing step after the root map is fully built.... this means the ... tokens are already resolved? We will need to find a way around this...

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
+  ): NameRef {
+    const referencedRoot = this.getNameMap()[referencedName]?.root;
+    const ref: NameRef =
+      referencedRoot !== undefined
+        ? { __type: 'root', value: referencedRoot, parent, key, targetPart }
+        : { __type: 'usage', value: referencedName, parent, key, targetPart };
+
+    const weak = new WeakRef(ref);
+    this.allRefs.add(weak);
+    this.finalizer.register(ref, weak);
```
FrozenPandaz: this isn't actually being used so let's remove it... we can see if there actually is a perf optimization to be done here.

astro-docs/src/content/docs/reference/project-configuration.mdoc
```diff
@@ -631,6 +631,75 @@ In the case of an explicit target using an executor, you can specify the executo
...
+| `inputs`, `outputs`, `dependsOn`, `syncGenerators` | `"inputs": ["my-input", "..."]`                                             |
+| `options`                                          | `"options": { "...": true, "outputPath": "dist/custom" }`                   |
+| `options[x]`                                       | `"options": { "env": { "MY_VAR": "val", "...": true } }`                    |
+| `configurations`                                   | `"configurations": { "my-config": { ... }, "...": true }`                   |
+| `configurations[x]`                                | `"configurations": { "prod": { "...": true, "sourceMap": false } }`         |
+| `configurations[x][y]`                             | `"configurations": { "prod": { "env": { "MY_VAR": "val", "...": true } } }` |
+
+`options[x]` and `configurations[x][y]` are equivalent depths — both are the innermost level the merge pipeline processes. The spread token has no effect nested inside those values (for example, `options.webpack.plugins: { "...": true }` is ignored because Nx treats `options.webpack` as an opaque value that fully replaces the base).
```
FrozenPandaz: Callout that it will not work in deeply nested options

astro-docs/src/content/docs/reference/project-configuration.mdoc
```diff
@@ -631,6 +631,75 @@ In the case of an explicit target using an executor, you can specify the executo
...
+
+#### Placing `"..."` at the end of a target
+
+Placing `"..."` at the end of a target object switches the merge to "add-only" mode: properties defined before `"..."` can be overridden by the base, so only keys the base does not define survive from your configuration.
```
FrozenPandaz: rm

packages/nx/src/project-graph/utils/project-configuration-utils/target-defaults.ts
```diff
@@ -0,0 +1,237 @@
...
+  defaultPluginRootMap: Record<string, ProjectConfiguration>,
+  nxJsonConfiguration: NxJsonConfiguration
+): CreateNodesResultEntry[] {
+  if (!nxJsonConfiguration.targetDefaults) {
+    return [];
+  }
+
+  const syntheticProjects: Record<string, ProjectConfiguration> = {};
+
+  const readAndPrepareTargetDefaults = (
```
FrozenPandaz: can this be hoisted into the module scope function?

packages/nx/src/project-graph/utils/project-configuration-utils/target-defaults.ts
```diff
@@ -0,0 +1,237 @@
...
+  };
+
+  // Get all unique project roots that have targets from either specified or default plugins
+  const allRoots = new Set<string>([
+    ...Object.keys(specifiedPluginRootMap).filter(
+      (root) => specifiedPluginRootMap[root]?.targets
+    ),
+    ...Object.keys(defaultPluginRootMap).filter(
+      (root) => defaultPluginRootMap[root]?.targets
+    ),
```
FrozenPandaz: don't filter here... skip them later

packages/nx/src/project-graph/utils/project-configuration-utils/target-defaults.ts
```diff
@@ -0,0 +1,237 @@
...
+      const resolvedSpecified = specifiedTarget
+        ? resolveCommandSyntacticSugar(specifiedTarget, root)
+        : undefined;
+      const resolvedDefault = defaultTarget
+        ? resolveCommandSyntacticSugar(defaultTarget, root)
+        : undefined;
+
+      // Determine which executor to use for target defaults lookup
+      let executorForDefaults: string | undefined;
+      let syntheticTarget: TargetConfiguration | undefined;
```
FrozenPandaz: Please pull this out into a function

packages/nx/src/project-graph/utils/project-configuration-utils/target-defaults.ts
```diff
@@ -0,0 +1,237 @@
...
+      }
+
+      // Add synthetic target if we determined one should be created
+      if (syntheticTarget) {
+        if (!syntheticProjects[root]) {
+          syntheticProjects[root] = {
+            root,
+            targets: {},
+          };
+        }
```
FrozenPandaz:
```suggestion
        syntheticProjects[root] ?= {
          root,
          targets: {},
        };
```

packages/nx/src/project-graph/utils/project-configuration-utils/target-defaults.ts
```diff
@@ -0,0 +1,237 @@
...
+        if (!syntheticProjects[root].targets) {
+          syntheticProjects[root].targets = {};
+        }
```
FrozenPandaz:
```suggestion
```

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+export function uniqueKeysInObjects(...objs: Record<string, any>[]): string[] {
+  const keys = new Set<string>();
+  for (const obj of objs) {
+    if (obj) {
+      for (const key of Object.keys(obj)) {
+        keys.add(key);
+      }
+    }
+  }
+  return Array.from(keys);
```
FrozenPandaz: return the set

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+ * When `sourceMapContext` is provided, the source map is updated per property
+ * to track which source each element/key came from. The existing source map
+ * entry for `key` is used as the base source information.
+ *
+ * When `preserveUnknownSpreads` is true and the base value is undefined (no
+ * earlier layer contributed anything to merge against), the spread sentinel is
+ * kept in the result so a later merge pass can resolve it against the
+ * eventual base. Used by intermediate merges that must defer spread
+ * resolution until the final layer order is known.
+ */
```
FrozenPandaz: simplify the comment

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+export function getMergeValueResult(
```
FrozenPandaz: Add a generic

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+): any {
+  // no new value, use old value — source map stays as-is (base source)
+  if (newValue === undefined && baseValue !== undefined) {
+    return baseValue;
+  }
+
+  // Check for spread syntax before type-mismatch bailout, since spread
+  // should work even when baseValue is undefined (treated as empty)
+  if (Array.isArray(newValue)) {
+    const spreadIndex = newValue.findIndex((v) => v === NX_SPREAD_TOKEN);
```
FrozenPandaz:
```suggestion
    const newSpreadIndex = newValue.findIndex((v) => v === NX_SPREAD_TOKEN);
```

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+          if (keepSpread) {
+            result.push(NX_SPREAD_TOKEN);
+            if (sourceMapContext) {
+              sourceMapContext.sourceMap[
+                `${sourceMapContext.key}.${resultIdx}`
+              ] = sourceMapContext.sourceInformation;
+            }
+            resultIdx++;
+            continue;
+          }
```
FrozenPandaz: make this an else

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (spreadIndex !== -1) {
+      const baseArray = Array.isArray(baseValue) ? baseValue : [];
+      const baseSourceInfo = sourceMapContext?.sourceMap[sourceMapContext.key];
+      const keepSpread = preserveUnknownSpreads && baseValue === undefined;
+
+      // Build the result array element by element so we can track each
+      // element's source in the source map.
+      const result: any[] = [];
+      let resultIdx = 0;
+      for (let i = 0; i < newValue.length; i++) {
```
FrozenPandaz:
```suggestion
      for (let newValueIndex = 0; newValueIndex < newValue.length; newValueIndex++) {
```

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+        sourceMapContext.sourceMap[sourceMapContext.key] =
+          sourceMapContext.sourceInformation;
+      }
+      return result;
+    }
+    // no spread token, and looking at an array, so return new array.
+    if (sourceMapContext) {
+      sourceMapContext.sourceMap[sourceMapContext.key] =
+        sourceMapContext.sourceInformation;
+    }
```
FrozenPandaz: this is missing the source map context per item... we'll want to always loop over it here.

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+export function getMergeValueResult(
+  baseValue: any,
+  newValue: any,
+  sourceMapContext?: {
+    sourceMap: Record<string, SourceInformation>;
+    key: string;
+    sourceInformation: SourceInformation;
+  },
+  preserveUnknownSpreads?: boolean
+): any {
```
FrozenPandaz: break this function up please

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+  } else if (
+    // new value is a non-null object, that we know is not an array
+    typeof newValue === 'object' &&
+    newValue != null &&
+    // '...' is explicitly set to `true`
+    newValue[NX_SPREAD_TOKEN] === true
+  ) {
+    const baseObj =
+      typeof baseValue === 'object' && baseValue != null ? baseValue : {};
+    const baseSourceInfo = sourceMapContext?.sourceMap[sourceMapContext.key];
```
FrozenPandaz: this should be per key?

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+    const keepSpread = preserveUnknownSpreads && baseValue === undefined;
```
FrozenPandaz: inline this... this name is confusing

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+          sourceMapContext.sourceMap[`${sourceMapContext.key}.${newKey}`] =
+            sourceMapContext.sourceInformation;
+        }
+      } else if (keepSpread) {
+        res[NX_SPREAD_TOKEN] = true;
+        if (sourceMapContext) {
+          sourceMapContext.sourceMap[
+            `${sourceMapContext.key}.${NX_SPREAD_TOKEN}`
+          ] = sourceMapContext.sourceInformation;
+        }
```
FrozenPandaz: it's weird to have this spread token source information because the spread token won't make it to the end of the project merging.

packages/nx/src/project-graph/utils/project-configuration-utils/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+      } else {
+        for (const baseKey in baseObj) {
```
FrozenPandaz: make this the first if block

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
+}
+export type NameRef = RootRef | UsageRef;
+
+export function isNameRef(value: unknown): value is NameRef {
+  if (typeof value !== 'object' || value === null) return false;
+  const type = (value as { __type?: unknown }).__type;
+  return (
+    (type === 'root' || type === 'usage') &&
+    typeof (value as { value?: unknown }).value === 'string'
+  );
```
FrozenPandaz: This seems like it can be simplified

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
+    targetPart?: string
+  ): NameRef {
+    const referencedRoot = this.getNameMap()[referencedName]?.root;
+    const ref: NameRef =
+      referencedRoot !== undefined
+        ? { __type: 'root', value: referencedRoot, parent, key, targetPart }
+        : { __type: 'usage', value: referencedName, parent, key, targetPart };
+
+    const weak = new WeakRef(ref);
+    this.allRefs.add(weak);
```
FrozenPandaz:
```suggestion
    this.allRefs.add(ref);
```

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
+    this.finalizer.register(ref, weak);
+
+    if (isUsageRef(ref)) {
+      let set = this.pendingByName.get(referencedName);
+      if (!set) {
+        set = new Set();
+        this.pendingByName.set(referencedName, set);
       }
+      set.add(weak);
```
FrozenPandaz:
```suggestion
      set.add(ref);
```

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
-      }
+  applySubstitutions(rootMap: Record<string, ProjectConfiguration>): void {
+    const nameByRoot: Record<string, string | undefined> = {};
+    for (const root in rootMap) {
+      nameByRoot[root] = rootMap[root]?.name;
+    }
+
+    for (const weak of this.allRefs) {
+      const ref = weak.deref();
+      if (!isNameRef(ref)) continue;
```
FrozenPandaz: there should be nothing else in here

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
 
-      const substitutors = this.substitutorsByReferencedRoot.get(root);
-      if (!substitutors) {
+      const finalName = this.resolveFinalName(ref, nameByRoot);
+      if (finalName === undefined) {
+        // Can't resolve (orphaned root ref or unknown usage ref): leave
+        // the owning slot alone and drop our back-reference.
+        ref.parent = undefined;
         continue;
       }
```
FrozenPandaz: this doesn't seem necessary

packages/nx/src/project-graph/utils/project-configuration/name-substitution-manager.ts
```diff
@@ -1,667 +1,371 @@
...
+      const replacement =
+        ref.targetPart !== undefined
+          ? `${finalName}:${ref.targetPart}`
+          : finalName;
+
+      this.writeReplacement(ref, replacement);
+
+      // Drop the back-reference to break the sentinel ↔ array cycle so
+      // both can be collected if no one else holds them.
+      ref.parent = undefined;
```
FrozenPandaz: droip this

packages/nx/src/project-graph/utils/project-configuration/project-nodes-manager.ts
```diff
@@ -335,18 +337,35 @@ export class ProjectNodesManager {
...
-      pluginResultProjects
-    );
+    if (!pluginResultProjects) return;
+    const scoped: Record<string, ProjectConfiguration> = {};
+    for (const root in pluginResultProjects) {
+      if (mergedRootMap[root]) {
+        scoped[root] = mergedRootMap[root];
+      }
+    }
+    this.nameSubstitutionManager.registerSubstitutorsForNodeResults(scoped);
```
FrozenPandaz: rename this

packages/nx/src/project-graph/utils/project-configuration-utils.ts
```diff
@@ -194,121 +231,201 @@ export async function createProjectConfigurationsWithPlugins(
...
+      intermediateDefaultRootMap,
+      project,
+      defaultConfigurationSourceMaps,
+      sourceInfo,
+      false,
+      true
+    );
+  };
+
+  const processBatch = (
```
FrozenPandaz: please describe this better... and maybe also hoist it up to the module scope?

FrozenPandaz (changes requested):

packages/nx/schemas/nx-schema.json
```diff
@@ -655,6 +655,7 @@
     },
     "inputs": {
       "type": "array",
+      "description": "Inputs that this target depends on. Include `\"...\"` as an array element to splice in base inputs at that position; see https://nx.dev/reference/project-configuration#spread-token.",
```
FrozenPandaz: open a follow up pr for this...

These descriptions should probably point to where the actual properties are described rather than all pointing to spread tokens

FrozenPandaz (approved):

PR #34302, opened 2026-02-03
fix(testing): set openHtmlReport to on-failure if options not specified

appsandwich:
## Current Behavior

Using `nxE2EPreset` without the `options` parameter, or omitting the `openHtmlReport` property causes Playwright tests to fail with a config parsing error:
```typescript
nxE2EPreset(__filename); // <-- fails

const nxPlaywrightOptions: NxPlaywrightOptions = {
    openHtmlReport: 'on-failure'
};
nxE2EPreset(__filename, options); // <-- succeeds
```

## Expected Behavior

Omitting the `options` or the `openHtmlReport` property should default to `openHtmlReport: 'on-failure'` as per the docs.


leosvelperez:
Closing as fixed by https://github.com/nrwl/nx/pull/34750.

PR #34306, opened 2026-02-03
fix(core): fix CNW git amend and README marker handling

jaysoo:
This PR fixes two issues:
1. When the README changes are amended, there's an edge case where we don't have a commit to amend (e.g. `--skipGit`), and this fails the entire CNW flow.
2. When user opts out of Cloud, we strip the entire `<!-- BEGIN: nx-cloud -->` block in README rather than just the comments and leaving the content.

Closes NXC-3812

### DIFF
--- a/packages/create-nx-workspace/src/utils/template/update-readme.ts
+++ b/packages/create-nx-workspace/src/utils/template/update-readme.ts
@@ -13,7 +13,7 @@ const CLOUD_SETUP_SECTION = `## Finish your Nx platform setup
 /**
  * Updates README content by processing the nx-cloud marker section.
  * - If connectUrl is provided: replaces markers and content with the connect URL section
- * - If connectUrl is undefined: removes markers and their content entirely
+ * - If connectUrl is undefined: keeps the content between markers but strips the marker comments
  *
  * This ensures users never see the raw <!-- BEGIN/END: nx-cloud --> markers in their README.
  *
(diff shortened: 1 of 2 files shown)


packages/create-nx-workspace/src/utils/template/update-readme.ts
```diff
@@ -106,13 +108,30 @@ export async function amendOrCommitReadme(
   alreadyPushed: boolean
 ): Promise<void> {
   await execAndWait('git add README.md', directory);
+
+  // Check if there are any staged changes to commit
+  const { stdout: stagedChanges } = await execAndWait(
```
FrozenPandaz: We should be able to tell without this whether or not there's changes to made. Starting with if the content had the start and end markers.

FrozenPandaz (changes requested):

packages/create-nx-workspace/src/utils/template/update-readme.ts
jaysoo: We already check the result of calling `addConnectUrlToReadme` (https://github.com/nrwl/nx/blob/master/packages/create-nx-workspace/src/utils/template/update-readme.ts#L93). 

If it returns `false`, then we don't try to amend the commit.

https://github.com/nrwl/nx/blob/master/packages/create-nx-workspace/src/create-workspace.ts#L223-L226

I tested this in different scenarios, and could not hit the amend error.

I think it's possible that the git init didn't work, so we tried to amend when there's nothing commited. Here, for example, we log the init error but still proceed:

https://github.com/nrwl/nx/blob/master/packages/create-nx-workspace/src/create-workspace.ts#L183-L191

If we remove the check for staged changes here in `amendOrCommitReadme` then we should at least throw

FrozenPandaz (approved):

PR #34310, opened 2026-02-03
feat(core): support dependency filesets with ^{projectRoot} syntax

AgentEnder:
## Current Behavior
`^` and `dependencies: true` only work for fileset inputs

## Expected Behavior
Adds support for inputs of the form `^{projectRoot}/**/*.ts` as syntactic sugar for specifying a fileset input that should be collected from dependency projects.

Previously, only named inputs could use the `^` prefix to include dependencies (e.g., `^production`). Now filesets can also use this syntax directly without needing to define a named input first.

Examples:
- `^{projectRoot}/**/*.ts` - include .ts files from all dependencies
- `^{workspaceRoot}/tools/**/*` - include workspace tools from dependencies
- `{ fileset: '{projectRoot}/**/*.ts', dependencies: true }` - object form

Detection is deterministic: if the string after `^` starts with `{projectRoot}` or `{workspaceRoot}`, it's treated as a dependency fileset; otherwise, it's treated as a named input reference.

## Related Issue(s)

Fixes #


FrozenPandaz (approved):

FrozenPandaz (approved):
Please add docs for this in this PR.

PR #34320, opened 2026-02-04
feat(core): add AI agent detection and NDJSON output for CNW

jaysoo:
AI agents are detected via environment variables (CLAUDECODE, OPENCODE) and receive NDJSON streaming output, non-interactive mode, structured JSON results with explicit GitHub setup instructions.

Related NXC-3628


packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -336,6 +359,36 @@ export const commandsObject: yargs.Argv<Arguments> = yargs
     nxVersion
   ) as yargs.Argv<Arguments>;
 
+// Add AI-specific examples and epilogue only when AI agent detected
+// This keeps --help clean for human users
+if (isAiAgent()) {
+  commandsObject
+    .example(chalk.green('AI AGENTS (RECOMMENDED):'), '')
+    .example(
+      '  npx create-nx-workspace@latest myorg --template=nrwl/react-template --nxCloud=yes --interactive=false',
```
Coly010: I think we should make empty template the default. Not sold on it, but empty will cover more use cases

packages/create-nx-workspace/src/utils/ai/ai-output.ts
```diff
@@ -0,0 +1,387 @@
...
+  // Future: CURSOR_AGENT, GEMINI_CLI
+] as const;
+
+let _isAiAgent: boolean | undefined;
+
+/**
+ * Detect if the current process is being run by an AI agent.
+ * Checks environment variables in priority order.
+ */
+export function isAiAgent(): boolean {
```
Coly010, line 15: Let's also add logic for cursor and replit since we know these:

```
function isReplitAi(): boolean {
  return !!process.env.REPL_ID;
}

function isCursorAi(): boolean {
  const pagerMatches = process.env.PAGER === 'head -n 10000 | cat';
  const hasCursorTraceId = !!process.env.CURSOR_TRACE_ID;
  const hasComposerNoInteraction = !!process.env.COMPOSER_NO_INTERACTION;
  return pagerMatches && hasCursorTraceId && hasComposerNoInteraction;
}
```

Coly010 (approved):

packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -435,10 +504,99 @@ async function normalizeArgsMiddleware(
...
+    // Check if --template or --preset was EXPLICITLY provided via CLI
+    const templateProvided = Boolean(rawArgs.template);
+    const presetProvided = Boolean(rawArgs.preset);
+
+    // If no template/preset provided, output help and exit
+    // AI agent must explicitly choose a template after asking the user
+    if (!templateProvided && !presetProvided) {
+      const workspaceName = (argv.name as string) || (argv._[0] as string);
+      writeAiOutput(buildTemplateRequiredResult(workspaceName));
+      process.exit(0); // Exit 0 - JSON output has success: false, AI parses that
```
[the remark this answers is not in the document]
jaysoo: This is only for AI agents so the output isn't so jarring.

FrozenPandaz (approved):

PR #34329, opened 2026-02-04
fix(core): reduce daemon inotify watch count by upgrading watchexec

### LINKED ISSUE #2468: High memory/file-system-watcher usage
## Current Behavior

For a while we have been having issues with high memory usage. A month ago we also noticed error messages about "file watcher count too low". And also we started to notice that the disk would quickly fill up. 
We noticed dump files being created in the folder: `C:\Users\me\AppData\Local\Temp\wsl-crashes`. Sometimes the dump files have a size of 4gb. Although during this snapshot the dmp files were quite small:

Image

I also noticed this out-of-memory notice running `dmesg`:
Image

## Expected Behavior

We would not expect our workspace to cause these kinds of problems

### Failure Logs / Images / Videos

When running `htop` i saw high memory usage:
Image

I check the limit:

Image

Especially this snapshot makes me think of NX causing the high file watcher count errors. On the left you see that i retrieved a list of the pids with the number of watchers. On the right you can see the processes. Most of them are vscode/nx related:

Image

### Environment

Windows 11, wslg

Image

```

Node           : 22.13.0
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.9.2

nx                     : 20.4.4
@nx/js                 : 20.4.4
@nx/jest               : 20.4.4
@nx/eslint             : 20.4.4
@nx/workspace          : 20.4.4
@nx/angular            : 20.4.4
@nx/cypress            : 20.4.4
@nx/devkit             : 20.4.4
@nx/eslint-plugin      : 20.4.4
@nx/express            : 20.4.4
@nx/module-federation  : 20.4.4
@nx/node               : 20.4.4
@nx/plugin             : 20.4.4
@nx/storybook          : 20.4.4
@nx/web                : 20.4.4
@nx/webpack            : 20.4.4

Nx Powerpack
Nx key is expired. Renew license, or contact support (powerpack-support@nrwl.io) for more assistance.

Registered Plugins:
nx-stylelint/plugin
@nx/jest/plugin
Community plugins:
@compodoc/compodoc     : 1.1.26
@ngrx/component-store  : 19.0.1
@ngrx/effects          : 19.0.1
@ngrx/entity           : 19.0.1
@ngrx/eslint-plugin    : 18.1.1
@ngrx/operators        : 19.0.1
@ngrx/router-store     : 19.0.1
@ngrx/schematics       : 19.0.1
@ngrx/signals          : 19.0.1
@ngrx/store            : 19.0.1
@ngrx/store-devtools   : 19.0.1
@spaceribs/nx-betterer : 5.0.0
@storybook/angular     : 8.6.4
ng-mocks               : 14.13.2
nx-stylelint           : 18.0.0
Local workspace plugins:
         @athora-nx/storybook-test-runner
         @athora-nx/static-server
         @athora-nx/azuredevops
         @athora-nx/chromatic
         @athora-nx/workspace-plugin
         @athora-nx/coverage
```

wsl image:
```
root@dtl2024-q2:~/AthoraNx# uname -r
5.15.167.4-microsoft-standard-WSL2
```

- Nx Console version: `18.44.2`
- VSCode version:
```
Version: 1.98.2 (system setup)
Commit: ddc367ed5c8936efe395cffeec279b04ffd7db78
Date: 2025-03-12T13:32:45.399Z
Electron: 34.2.0
ElectronBuildId: 11161602
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.22631
```

### LINKED ISSUE #33781: Daemon file watcher slow to start causing stale cache and project graph in large repos using pnpm workspace
### Current Behavior

When starting the Nx daemon in a large project (in my case over 150 libraries) that uses PNPM workspaces, the file watching behavior doesn't activate until multiple minutes have passed. I presume this occurs due to the file watcher or hasher going through every library, each of which containing symlinks to each other inside `node_modules`. Having deleted most of the library-specific node_modules (specifically the nested workspace libraries), the file watching begins much quicker. Once it begins, one can run `pnpm install` to restore the library-specific node_modules and utilize the daemon normally.

This causes Nx cache to not work correctly (because file hashes aren't being recomputed). Probably also causes the project graph to become invalid if dependencies are modified.

### Expected Behavior

The daemon file watcher should activate quickly after daemon startup, preventing stale cache and project graph.

### GitHub Repo

_No response_

### Steps to Reproduce

Hard to reproduce, requires large monorepo using PNPM workspaces with lots of libraries and dependencies. This problem is prevalent on Linux. My colleague using macOS on Apple Silicon doesn't seem to experience this.

### Nx Report

```shell
Node           : 22.21.1
OS             : linux-x64
Native Target  : x86_64-linux
pnpm           : 10.22.0

nx                     : 22.2.0
@nx/js                 : 22.2.0
@nx/jest               : 22.2.0
@nx/eslint             : 22.2.0
@nx/workspace          : 22.2.0
@nx/cypress            : 22.2.0
@nx/devkit             : 22.2.0
@nx/esbuild            : 22.2.0
@nx/eslint-plugin      : 22.2.0
@nx/module-federation  : 22.2.0
@nx/node               : 22.2.0
@nx/playwright         : 22.2.0
@nx/react              : 22.2.0
@nx/rollup             : 22.2.0
@nx/storybook          : 22.2.0
@nx/vite               : 22.2.0
@nx/vitest             : 22.2.0
@nx/web                : 22.2.0
@nx/docker             : 22.2.0

Nx key licensed packages

@nx/azure-cache  : 5.0.0

Registered Plugins:
@nx/eslint/plugin
@nx/vite/plugin
@nx/js/typescript
@nx/vitest
@nx/js/typescript
@nx/storybook/plugin
@nx/docker
Community plugins:
@nx/azure-cache : 5.0.0
```

### Failure Logs

```shell

```

### Package Manager Version

pnpm 10.22.0

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

I've checked that all `node_modules` directories are in `.gitignore` and `.nxignore`. Modifying a file inside any node_modules directory doesn't result in any daemon log output, contrary to any other non-ignored file. I'm also not hitting any inotify watcher limits, and there are no error logs overall. The same issue occurs with Nx 22.1.3.

My current workaround is this:
1. Delete all workspace dependencies from subdirectory node_modules:
`find . -type d -path './*/**node_modules/@my-workspace' -not -path './node_modules**' -exec rm -rf {} +`
2. Start the Nx daemon: `pnpm nx`
3. Wait and test that file watching is working, e.g. watch logs with `tail -F .nx/workspace-data/d/daemon.log` and modify a file, you should see something like `[WATCHER]: libs/my-lib/src/index.ts was modified`
4. Once the watcher works, restore all dependencies with `pnpm install`

Example script (replace @acme with correct name):
```
#!/usr/bin/env bash

# Workaround for https://github.com/nrwl/nx/issues/33781

# Reset the daemon
echo "Resetting the daemon..."
pnpm nx reset

# Remove library specific node_modules
echo "Removing library specific node_modules..."
find . -type d -path './*/**node_modules/@acme' -not -path './node_modules**' -exec rm -rf {} +

# Start the daemon
echo "Starting the daemon..."
pnpm nx daemon

# Wait for file watcher to run
echo "Waiting for file watcher to run..."
while true; do
  echo "Waiting for file watcher to run..." > tmp_wait_for_daemon
  sleep 0.5
  # Check that daemon log contains 'Processing file changes in outputs'
  if grep -q 'Processing file changes in outputs' .nx/workspace-data/d/daemon.log; then
    break
  fi
  # Check that daemon log contains 'tmp_wait_for_daemon'
  if grep -q 'tmp_wait_for_daemon' .nx/workspace-data/d/daemon.log; then
    break
  fi
done

rm tmp_wait_for_daemon

# Restore dependencies
echo "Restoring dependencies..."
pnpm install
```

FrozenPandaz:
## Current Behavior

The daemon's file watcher uses watchexec 3.0.1 which hardcodes `RecursiveMode::Recursive` when registering inotify watches. This means **every** directory gets an inotify watch — including all of `node_modules`, `.git`, and other ignored trees.

On a typical workspace with a large `node_modules`, this can consume thousands of inotify watches, eating kernel memory and CPU. The `WatchFilterer` only filters **events** after watches are already registered — the watches themselves are never prevented.

## Expected Behavior

Only non-ignored directories (workspace source code) get inotify watches. Ignored directories like `node_modules`, `.git`, `.nx/cache`, `.nx/workspace-data`, and `.yarn/cache` are skipped entirely at the watch registration level.

This dramatically reduces:
- **inotify watch count** (from thousands to hundreds)
- **Memory usage** (each watch consumes kernel memory)
- **CPU overhead** (fewer watches = less kernel bookkeeping)

### How it works

- Upgraded watchexec 3.0.1 → 8.0.1 which supports `WatchedPath::non_recursive()`
- Added `create_watch_walker()` using `ignore::WalkBuilder` (same pattern as `walker.rs`) to enumerate only non-ignored directories
- Each directory is watched with `NonRecursive` mode — like putting security cameras only in the rooms you care about instead of every room in the building
- New directories created at runtime are dynamically added to the watch set via the `on_action` handler
- Event-level filtering via `WatchFilterer` is unchanged — same behavior for gitignore/nxignore patterns

### macOS Support for Dynamic Directory Registration

The initial implementation worked on Linux and Windows but failed tests on macOS because macOS FSEvents doesn't always provide the same `FileEventKind` tags as Linux inotify or Windows ReadDirectoryChangesW.

**Three changes to support macOS:**

1. **watcher.rs**: On macOS, check all events for directory creation (not just events with specific FileEventKind tags) and verify via filesystem
2. **types.rs**: Filter directory events from JavaScript callbacks on macOS (similar to Windows behavior)
3. **watch_filterer.rs**: Allow macOS directory events (`Create(Folder)` and `Modify(Metadata)`) through the filter so the action handler can register them

All changes use `#[cfg(target_os = "macos")]` for compile-time conditional compilation, so Linux/Windows behavior is completely unchanged and there's zero runtime overhead.

### Additional notes

- Pinned `serde` to `<1.0.220` because serde 1.0.220+ moved `__private` to `serde_core`, breaking `swc_common 0.31.22`
- No TypeScript changes — the napi interface is identical
- `watch_filterer.rs`, `types.rs`, `utils.rs` required no changes (APIs are compatible)

## Related Issue(s)
Fixes https://github.com/nrwl/nx/issues/33781
Fixes https://github.com/nrwl/nx-console/issues/2468


packages/nx/src/native/watch/watcher.rs
```diff
@@ -111,16 +236,50 @@ impl Watcher {
...
+                let new_pathset: Vec<WatchedPath> = path_set
+                    .iter()
+                    .map(|p| WatchedPath::non_recursive(p))
+                    .collect();
+                trace!(
+                    count = new_pathset.len(),
+                    "updating pathset with new directories"
+                );
+                watch_exec_for_action.config.pathset(new_pathset);
+            }
```
AgentEnder: Can we combine these 2? It feels like `register` should either actually do the registration, or it should reurn the new ones and then we handle it below. Right now it feels like its split into two

packages/nx/src/native/watch/watcher.rs
```diff
@@ -111,16 +236,50 @@ impl Watcher {
...
             }
             trace!(?origin_path);
 
+            trace!(
+                event_count = action.events.len(),
+                "on_action received events"
+            );
+            for ev in action.events.iter() {
+                trace!(?ev, "raw event");
+            }
```
AgentEnder: Can we inline into either par_iter or a tap somewhere in there? Feels weird to iter to trace and the iter again  to do the real work... The events are realistically not going to be some massive collection so it probably doesn't really matter but feels odd

packages/nx/Cargo.toml
```diff
@@ -65,7 +65,7 @@ tui-term = { git = "https://github.com/JamesHenry/tui-term", rev = "88e3b61425c9
 walkdir = '2.3.3'
 xxhash-rust = { version = '0.8.5', features = ['xxh3', 'xxh64'] }
 vt100-ctt = { git = "https://github.com/JamesHenry/vt100-rust", rev = "b15dc3b0f7db94167a9c584f1d403899c0cc871d" }
-serde = "1.0.219"
+serde = ">=1.0.219, <1.0.220"
```
AgentEnder: This looks weird?

AgentEnder (approved):

AgentEnder (approved):

PR #34345, opened 2026-02-05
feat(core): implement non-recursive watcher to avoid unnecessary watchers being registered

MaxKless:
### Current Behavior                                                            
                                                                                                                                                                             
  The Nx native file watcher uses watchexec (v3.0.1) which internally calls notify::INotifyWatcher with RecursiveMode::Recursive. This walks every subdirectory in the
  workspace and registers an OS-level watch on each one. The ignore patterns (node_modules/, .git/, .nx/, etc.) are only applied as event filters via the watchexec Filterer
  trait — they don't prevent the watch registration itself.

  In practice, a typical node_modules contains 30-50k+ subdirectories. All of these get watches registered, even though every event from them is immediately filtered out and
   discarded. This causes two problems:

  1. Linux (inotify): Blows past the default max_user_watches limit (usually 8192-65536), causing the watcher to fail with ENOSPC: System limit for number of file watchers
  reached.
  2. All platforms: Wastes OS resources tracking events for tens of thousands of directories whose events will never be used.

  On the Nx repo itself: 59,568 directories get watches, of which 46,086 are ignored (node_modules, .git, .nx).

  ### Expected Behavior

  The watcher only registers watches for directories that aren't ignored. Using the Nx repo as a benchmark:

  - Before: ~59,568 watches (every directory)
  - After: ~13,482 watches (only non-ignored directories)
  - Reduction: 77%

  For typical user projects with larger node_modules trees, the reduction will be even greater.

  The NAPI API surface is unchanged — TypeScript consumers see no difference.

  ### Approach

  Upgrade watchexec from 3.0.1 to 4.1.0, which introduces WatchedPath::non_recursive(). Instead of passing a single recursive root path to watchexec, we walk the directory
  tree ourselves using ignore::WalkBuilder (the same crate and pattern already used in walker.rs) and pass each non-ignored directory as a WatchedPath::non_recursive().


 ### Key design decisions

  1. All platforms, not just Linux. While inotify's hard max_user_watches limit makes Linux the most acute case, registering watches for 50k+ ignored directories is wasteful
   everywhere. Same code path on all platforms keeps things simple.
  2. walk_directories() reuses the walker.rs pattern. Uses ignore::WalkBuilder with the same gitignore/nxignore/glob handling as the existing file walker —
  parent_gitignore_files() for git boundary detection, .nxignore support, and filter_entry for the watcher's static ignore globs.
  3. Dynamic directory creation handling. With non-recursive watches, newly created directories aren't automatically watched. The on_action callback detects
  CreateKind::Folder events and re-walks the workspace to update the watch set. This required removing the #[cfg(target_os = "linux")] guard on directory events in the
  filterer — previously directory events were only allowed through on Linux (for inotify race condition handling), but now all platforms need them to detect new directories.
  4. Weak<Watchexec> to avoid circular references. The on_action closure needs access to watchexec.config.pathset() to update watches. Using Arc::downgrade() avoids a
  circular Arc reference (Watchexec → Config → on_action closure → Arc<Watchexec>) that would prevent cleanup.
  5. Full re-walk on directory creation rather than incremental updates. config.pathset() replaces the entire path set (no "add" API). Rather than maintaining a separate
  tracking structure, we re-walk the workspace on directory creation. This is simple, correct, and fast since WalkBuilder skips ignored directories.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -163,9 +163,9 @@ dependencies = [
 
 [[package]]
 name = "async-priority-channel"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c21678992e1b21bebfe2bc53ab5f5f68c106eddab31b24e0bb06e9b715a86640"
+checksum = "acde96f444d31031f760c5c43dc786b97d3e1cb2ee49dd06898383fe9a999758"
 dependencies = [
  "event-listener",
 ]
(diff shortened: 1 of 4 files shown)


FrozenPandaz:
I'll handle this in https://github.com/nrwl/nx/pull/34329

PR #34347, opened 2026-02-05
fix(js): guard against undefined closest node in rehoistNodes

kaigritun:
## Current Behavior

When running `@nx/js:prune-lockfile` on a monorepo with transitive dependencies that have multiple versions where neither version is reachable from a direct dependency in package.json, the executor throws:

```
NX   An error occurred while creating pruned lockfile

Original error: Cannot read properties of undefined (reading 'name')

TypeError: Cannot read properties of undefined (reading 'name')
    at switchNodeToHoisted (node_modules/nx/src/plugins/js/lock-file/project-graph-pruning.js:165:31)
```

## Expected Behavior

The lockfile pruning should complete without crashing, even when some transitive dependencies cannot be traced back to a direct dependency.

## Root Cause

In `rehoistNodes()`, when there are multiple nested nodes for a package, the code finds the "closest" node by computing `pathLengthToIncoming()` for each. However, when none of the nested nodes have a path to any direct dependency in package.json, `pathLengthToIncoming()` returns `undefined` for all of them. Since `undefined < Infinity` is `false` in JavaScript, `closest` remains `undefined`, and then `switchNodeToHoisted(undefined, ...)` crashes.

## Fix

Add a guard to only call `switchNodeToHoisted()` when a closest node was actually found:

```typescript
if (closest) {
  switchNodeToHoisted(closest, builder, invBuilder);
}
```

This allows the pruning to continue - the nested nodes simply won't be rehoisted if no closest node can be determined.

## Related Issue

Fixes #34322

## Test Added

Added a unit test that verifies `rehoistNodes()` doesn't crash when nested nodes have no path to package.json dependencies.

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/project-graph-pruning.spec.ts
+++ b/packages/nx/src/plugins/js/lock-file/project-graph-pruning.spec.ts
@@ -424,6 +424,48 @@ describe('project-graph-pruning', () => {
         '4.17.20'
       );
     });
+
+    it('should not crash when nested nodes have no path to package.json deps', () => {
+      // This tests the fix for https://github.com/nrwl/nx/issues/34322
+      // When transitive dependencies have multiple versions where neither version
+      // is reachable from a direct dependency, the code should not crash.
+
+      // Setup: Add multiple nested versions that are NOT in packageJsonDeps
+      builder.addExternalNode({
+        type: 'npm',
+        name: 'npm:lodash@4.17.20',
+        data: {
+          packageName: 'lodash',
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

leosvelperez (approved):

PR #34356, opened 2026-02-05
fix(core): use dependent task hash instead of re-reading output files…

Exelord:
## Current Behavior

When a target configures `dependentTasksOutputFiles: "**/*"`, Nx always reads the dependency's output files from disk to compute the parent task's hash. This means:

1. Dependencies must run first to produce output files before the parent task can be hashed
2. During cache restoration, Nx restores child outputs and re-reads them to recalculate the parent hash — even though the child's hash already represents all of its outputs
3. Tasks with `dependentTasksOutputFiles` cannot be hashed upfront, adding latency to the task pipeline

## Expected Behavior

When `dependentTasksOutputFiles` is `"**/*"` (meaning "depend on all output files"), Nx uses the dependency task's pre-computed hash as a proxy instead of reading files from disk. This:

- Enables upfront hashing — tasks with `"**/*"` no longer need to wait for dependencies to build first
- Avoids redundant disk I/O during cache restoration
- Preserves existing behavior for specific globs (e.g. `"**/*.d.ts"`) which still read from disk for precise cache invalidation

### Implementation

- **`types.rs`**: Added `hash` field to `Task` struct and `TaskHash(task_id, task_hash)` variant to `HashInstruction` enum
- **`dep_outputs.rs`**: When glob is `"**/*"` and the dependency task has a pre-computed hash, emits `TaskHash` instead of `TaskOutput`. Falls back to `TaskOutput` when hash is not yet available
- **`task_hasher.rs`**: Handles `TaskHash` instruction by combining task hash + task ID
- **`hash-task.ts`**: Allows tasks with `"**/*"` `depsOutputs` to be hashed upfront in `hashTasksThatDoNotDependOnOutputsOfOtherTasks()`
- Includes Rust unit tests and TypeScript integration tests

## Related Issue(s)

Fixes #


Exelord:
Hi @FrozenPandaz !
any chance to get a bit of your time review this PR? I would very appreciate it. This seams to a be an area that was missed in NX, while delivering enormous speed improvement in our project.

AgentEnder:
Thanks for the PR. It conflicts with `master` at the moment, so could you rebase it? It will get a review as time allows once that is done.

<details>
<summary>

</summary>

> Assignee is a subsystem override on the `scope: core` rotation. Task hashing and cache inputs sit with FrozenPandaz, which is where this lands.
>
> No GitHub Actions run exists for the head commit, so nothing has built this. The `PENDING` rollup is a third-party check, not our CI. That is not a red run, so no draft is proposed, and a rebased push is what would trigger a real build.
>
>
> No GitHub closing reference and no Linear key, so this is unlinked. Kept rather than closed, since it is a real optimisation to `dependentTasksOutputFiles` hashing with the argument written out, not a speculative change.
>
> The `stale` label is left alone. It is managed by the stale workflow, and this comment resets its clock.
>
> Whether substituting a dependency's task hash for its output files is correct in every case, including when outputs change without the hash changing, is a call on the shape of the change rather than something a line-by-line review settles. That wants a yes before review effort goes into it.

</details>

PR #34358, opened 2026-02-05
fix(core): ensure verbose logs go to stderr 

AgentEnder:
## Current Behavior
`logger.[x]` doesn't get decorated despite being emitted to the daemon log, perf logs aren't decorated, etc

## Expected Behavior
daemon logs are decorated

## Related Issue(s)

Fixes #


e2e/utils/command-utils.ts
```diff
@@ -466,7 +466,10 @@ export function runCLI(
     }
     if (opts.silenceError) {
       runCLI.lastExitCode = (e.status ?? 1) as number;
-      return stripVTControlCharacters(e.stdout + e.stderr);
+      const combinedOutput = opts.redirectStderr
```
FrozenPandaz: We don't need this because it's being redirected to stdout via bash

e2e/utils/command-utils.ts
FrozenPandaz, line 417: Try flipping this to default to true.... and all you should have to do after that is revert the changes in e2e tests... and update snapshots

FrozenPandaz (changes requested):

e2e/gradle/src/gradle-plugin-v1.test.ts
AgentEnder, line 1: Nothing in this file seems to have meaningfully changed?

e2e/nx/src/extras.test.ts
AgentEnder, line 1: This change doesn't make sense, I'd still expect this to go to stderr?

e2e/nx/src/run.test.ts
AgentEnder, line 1: Seemingly no meaningful change, can we revert?

e2e/nx/src/workspace.test.ts
AgentEnder, line 1: Not sure why this changed, we still want to assert that it errored

e2e/release/src/version-plans-check.test.ts
AgentEnder, line 1: These seem largely unnecessary too, we could keep the inlining + the snapshots would already match the .toContain string bits so its 2 expectations for the same thing.

e2e/utils/command-utils.ts
```diff
@@ -466,7 +466,7 @@ export function runCLI(
     }
     if (opts.silenceError) {
       runCLI.lastExitCode = (e.status ?? 1) as number;
-      return stripVTControlCharacters(e.stdout + e.stderr);
+      return stripVTControlCharacters(e.stdout);
```
AgentEnder: This probably still needs e.stderr for the cases where we explicitly pass false?

packages/nx/src/daemon/logger.ts
AgentEnder, line 1: Check the graphite comments here, we'll want to validate them

packages/nx/src/utils/output.ts
```diff
@@ -264,16 +282,16 @@ class CLIOutput {
...
     this.addNewline();
-    this.writeToStdOut(output);
+    this.writeToStream(output);
 
     if (
       process.env.NX_SKIP_LOG_GROUPING !== 'true' &&
       process.env.GITHUB_ACTIONS
     ) {
-      this.writeToStdOut(GH_GROUP_SUFFIX);
+      this.writeToStream(GH_GROUP_SUFFIX);
```
AgentEnder, line 294: I believe these logs that are meant to directly affect github actions **must** go to stdout or github won't parse them

e2e/release/src/version-plans-check.test.ts
```diff
@@ -204,8 +204,13 @@ describe('nx release version plans check command', () => {
...
-      .toMatchInlineSnapshot(`
+    const verboseResult = runCLI(
+      'release plan:check --base=HEAD~1 --head=HEAD~1 --verbose'
+    );
+    expect(verboseResult).toContain(
+      'No changed files found based on resolved "base" and "head"'
+    );
+    expect(verboseResult).toMatchInlineSnapshot(`
 
       NX   No changed files found based on resolved "base" and "head"
```
AgentEnder, line 219: Part of the motivation here was that these snapshots are overly broad and adding new verbose logs causes them to break. It'd be nice to not assert on the combined output with a snapshot, and if using a snapshot to use toContain or similar to ensure adding new logs doesn't break things. Without verbose its more sensible, as you are less likely to break them and are asserting on the normal user UX

packages/nx/src/utils/perf-logging.ts
AgentEnder, line 1: How about `NX_PERF_LOGGING=true` == "NX_PERF_LOGGING=stderr", but we support stdout or relative file paths as other values?

AgentEnder (commented):
Spot checked some stuff

e2e/utils/command-utils.ts
AgentEnder: Tried it, it kinda worked, but I didn't love the final results and it caused issues in other tests that already asserted against stderr, so it seemed to even out and I decided to leave it as is.

packages/nx/src/utils/output.ts
```diff
@@ -103,22 +104,30 @@ class CLIOutput {
...
-    this.writeToStdOut(`${this.applyNxPrefix(color, title)}${EOL}`);
+  private writeOutputTitle(
+    {
+      color,
+      title,
+    }: {
+      color: string;
+      title: string;
+    },
+    stream: WriteStream = process.stdout
```
FrozenPandaz, line 115: Make it so you always have to pass this in... it actually seems like this is only called with `stderr`

packages/nx/src/utils/output.ts
```diff
@@ -103,22 +104,30 @@ class CLIOutput {
...
+  ): void {
+    this.writeToStream(`${this.applyNxPrefix(color, title)}${EOL}`, stream);
   }
 
-  private writeOptionalOutputBody(bodyLines?: string[]): void {
+  private writeOptionalOutputBody(
+    bodyLines?: string[],
+    stream: WriteStream = process.stdout
```
FrozenPandaz, line 122: This one too

FrozenPandaz (approved):

PR #34362, opened 2026-02-06
docs(maven): minor tweaks and include targetNamePrefix

MaxKless:
this makes the maven documentation more complete for the latest changes


astro-docs/src/content/docs/technologies/java/maven/introduction.mdoc
```diff
@@ -107,13 +107,31 @@ nx add @nx/maven
...
 }
 ```
 
 Once the plugin has been added, `@nx/maven` will automatically retrieve information about projects from Maven and create targets for each phase, goal, and some additional targets for CI.
+
+### Target Name Prefix
+
+In a polyglot workspace supporting projects in different languages, you may have targets with the same name (e.g., `build`, `test`) but different behaviors. By default, Maven phase and goal names are used directly as Nx target names. Use the `targetNamePrefix` option to prefix all Maven targets with a custom string to distinguish them from other targets in your workspace.
+
+For example, with `"targetNamePrefix": "mvn-"`, Nx will recognize your targets as `mvn-compile`, `mvn-test`, `mvn-package`, etc.
```
barbados-clemens: I don’t think “recognize” is right here. Bc Nx isn’t recognized targets with the prefix, it’s creating targets with that prefix right?

barbados-clemens (approved):

PR #34376, opened 2026-02-09
cleanup(core): use picocolors in eslint-plugin

43081j:
Switches to picocolors in the ESLint plugin.

Doesn't use native `styleText` yet as we'd need to bump the minimum Node
version I think. Also unfortunately needs its own `orange`
implementation for now.

@JamesHenry we could use `ansis` here if we want a third party `orange` but it seems simple enough i'd just keep our own implementation here. one day this could all switch to native `styleText` too.

## Related Issue(s)

N/A

### DIFF
--- a/packages/eslint-plugin/src/utils/project-graph-utils.ts
+++ b/packages/eslint-plugin/src/utils/project-graph-utils.ts
@@ -13,10 +13,6 @@ import { readNxJson } from 'nx/src/config/configuration';
 import { TargetProjectLocator } from '@nx/js/internal';
 import { readFileMapCache } from 'nx/src/project-graph/nx-deps-cache';
 
-// TODO (43081j): if `styleText` or `picocolors` ever supports RGB, use
-// that instead.
-const orangeText = (text: string) => `\x1B[38;2;255;165;0m${text}\x1B[39m`;
-
 export function ensureGlobalProjectGraph(ruleName: string) {
   /**
    * Only reuse graph when running from terminal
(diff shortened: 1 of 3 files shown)


gperdomor:
what about tinyrainbow? It's piccolo's but ESM 🤔

43081j:
i do have an open issue to move picocolors to ESM but its slow going

we use picocolors throughout the monorepo already too, so this is more of a consistency thing.

ideally we can switch to native `styleText` across the board eventually once the engines constraint can be bumped.

packages/eslint-plugin/src/utils/project-graph-utils.ts
```diff
@@ -13,6 +13,10 @@ import { readNxJson } from 'nx/src/config/configuration';
 import { TargetProjectLocator } from '@nx/js/src/internal';
 import { readFileMapCache } from 'nx/src/project-graph/nx-deps-cache';
 
+// TODO (43081j): if `styleText` or `picocolors` ever supports RGB, use
+// that instead.
+const orangeText = (text: string) => `\x1B[38;2;255;165;0m${text}\x1B[39m`;
```
JamesHenry: We should reuse the existing one right? `import { orange } from 'nx/src/utils/output';`

JamesHenry (dismissed):

AgentEnder (approved):

FrozenPandaz (approved):

FrozenPandaz:
Thank you for your contribution! I made things easier by changing orange to yellow :+1:

PR #34400, opened 2026-02-11
fix(maven): write output after each task in batch mode to ensure correct files are cached

MaxKless:
## Current Behavior
When running in maven 4 batch mode, the build state is recorded only after the full batch is done. 
This means that nx caching records the state of a task before build state is recorded to disk.
When running another maven task that depends on this partially recorded cache, the build state file is missing and we get errors.

## Expected Behavior
build state should be recorded after every task is done and before nx caching can kick in. This way we can ensure that nx cache is correct.

### DIFF
--- a/e2e/maven/src/maven-batch-v4.test.ts
+++ b/e2e/maven/src/maven-batch-v4.test.ts
@@ -50,6 +50,25 @@ wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-w
     );
   });
 
+  it('should install successfully after restoring cached package outputs', () => {
+    // Step 1: Delete target directories to simulate a clean CI checkout
+    for (const mod of ['app', 'lib', 'utils']) {
+      rmSync(tmpProjPath(`${mod}/target`), { recursive: true, force: true });
+    }
+
+    // Step 2: Run package in batch mode — cache hit restores outputs (including nx-build-state.json)
+    runBatchCLI('run-many -t package');
+    checkFilesExist(
+      'app/target/app-1.0.0-SNAPSHOT.jar',
+      'lib/target/lib-1.0.0-SNAPSHOT.jar',
+      'utils/target/utils-1.0.0-SNAPSHOT.jar'
+    );
+
+    // Step 3: Run install in batch mode — this requires build state from the package phase
+    // to know about the main artifact.
+    runBatchCLI('run-many -t install');
+  });
+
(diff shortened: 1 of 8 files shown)


e2e/maven/src/maven-batch-v4.test.ts
```diff
@@ -81,4 +83,48 @@ class AppApplicationTests {
     expect(error).toBeDefined();
     expect(error.stdout || error.stderr).toContain('thisTestShouldFail');
   });
+
+  it('should install successfully after restoring cached package outputs', () => {
+    // Clean up from previous test: reset Nx cache to avoid stale cached state,
+    // delete target directories to remove stale compiled classes,
+    // and restore the original passing test file (previous test added a failing test)
+    runCLI('reset');
```
MaxKless: this is because of another caching bug - hashes are computed upfront in batch mode even though there are dependencies between the tasks

e2e/maven/src/maven-batch-v4.test.ts
```diff
@@ -48,7 +50,28 @@ wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-w
     );
   });
 
+  it('should install successfully after restoring cached package outputs', () => {
+    // Step 1: Delete target directories to simulate a clean CI checkout
+    for (const mod of ['app', 'lib', 'utils']) {
+      rmSync(tmpProjPath(`${mod}/target`), { recursive: true, force: true });
+    }
```
FrozenPandaz: Try calling `nx run-many -t clean` or `./mvnw clean` because that's a more idiomatic way for people to get back to a clean state.

packages/maven/batch-runner/src/main/kotlin/dev/nx/maven/runner/MavenInvokerRunner.kt
```diff
@@ -132,9 +137,6 @@ class MavenInvokerRunner(private val workspaceRoot: File, private val options: M
 
       // Wait for all tasks to complete
       completionLatch.await()
-
-      // Record build states for all projects that had tasks executed
-      recordBuildStatesForExecutedTasks()
```
FrozenPandaz, line 137: I had this as recording build states for all projects because there are cases where `projA:taskA` will alter the build state for `projB`. So this needs to be taken into account still.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #34407, opened 2026-02-11
feat(js): add deps-sync generator

meeroslav:
This PR introduces the `deps-sync` generator which pairs with `typescript-sync`generator to ensure internal dependencies are correctly mapped via `devDependencies` of corresponding `package.json`.

## Current Behavior
When dependency to local package is created (via import for example) the existing typescript-sync generator updates the tsconfig but package.json is left untouched which might cause issues when referencing transitive dependencies.

## Expected Behavior
The typescript-sync should be accompanied by deps-sync generator that would update dependencies in package.json for workspaces.

## Related Issue(s)

Fixes #


pnpm-lock.yaml
leosvelperez, line 1: Why the large diff here?

AgentEnder: IIRC this is almost always an out of sync pnpm version?

meeroslav: It probably sneaked in through merge of master. I'll remove it as nothing in this PR needs lock file change.

astro-docs/src/content/docs/concepts/sync-generators.mdoc
```diff
@@ -54,7 +54,7 @@ Use the project details view to **find registered sync generators** for a given
 nx show project <name>
 ```
 
-The above command opens up the project details view, and the registered sync generators are under the **Sync Generators** for each target. Most sync generators are inferred when using an [inference plugin](/docs/concepts/inferred-tasks). For example, the `@nx/js/typescript` plugin registers the `@nx/js:typescript-sync` generator on `build` and `typecheck` targets.
+The above command opens up the project details view, and the registered sync generators are under the **Sync Generators** for each target. Most sync generators are inferred when using an [inference plugin](/docs/concepts/inferred-tasks). For example, the `@nx/js/typescript` plugin registers the `@nx/js:typescript-sync` and `@nx/js:deps-sync` generator on `build` and `typecheck` targets.
```
leosvelperez:
```suggestion
The above command opens up the project details view, and the registered sync generators are under the **Sync Generators** for each target. Most sync generators are inferred when using an [inference plugin](/docs/concepts/inferred-tasks). For example, the `@nx/js/typescript` plugin registers the `@nx/js:typescript-sync` and `@nx/js:deps-sync` generators on `build` and `typecheck` targets.
```

leosvelperez (approved):
Just left a comment for a grammar issue to fix

FrozenPandaz (approved):

Coly010 (approved):

PR #34423, opened 2026-02-12
cleanup(repo): avoid unnecessary project graph recomputations

leosvelperez:
## Current Behavior

- Cypress `start-dev-server.ts` file creates a port lock file next to the source code
- Native temp DB files created by tests are not properly ignored
- Astro config timestamp file is not ignored

These all trigger watch file change events, which cause the project graph to be recomputed unnecessarily.

## Expected Behavior

Output files should not trigger watch file change events. The project graph should not be recomputed unnecessarily.


.nxignore
```diff
@@ -3,3 +3,10 @@ nx-dev/**/jest.config.js
 _files
 _solution
 nx-dev/tutorial/**/templates
+
+# Generated by napi-rs (outputs of build-native)
+packages/nx/src/native/index.d.ts
+packages/nx/src/native/native-bindings.js
+
+# Workaround for ignore-files crate bug with prefix matching
```
FrozenPandaz: We merged the fix for `ignore-files`

.nxignore
```diff
@@ -3,3 +3,10 @@ nx-dev/**/jest.config.js
 _files
 _solution
 nx-dev/tutorial/**/templates
+
+# Generated by napi-rs (outputs of build-native)
+packages/nx/src/native/index.d.ts
```
FrozenPandaz: Yes, these are generated but these are also important for hashing.

If we add it to `.nxignore`, then it will yield false positive cache hits.

Let's not make any changes to this file.

FrozenPandaz (approved):
The other files are fine. Please remove the changes to `.nxignore`.

PR #34424, opened 2026-02-12
fix(core): skip stale recomputations and prevent lost file changes in daemon

leosvelperez:
## Current Behavior

When file changes arrive rapidly, the daemon triggers multiple concurrent project graph recomputations that all run to completion — wasting CPU/memory on redundant work and returning stale results.

Additionally, after processing file changes, the daemon clears all tracked files indiscriminately. Files that changed mid-recomputation are silently lost and never reflected in the project graph until another unrelated file change arrives.

## Expected Behavior

Stale recomputations detect when a newer one has started and exit early, chaining to the newer promise so callers always get the freshest result.

File change tracking now uses versioned maps. Each batch of file watcher events gets a unique version, and only files matching the snapshotted version are cleared after processing. Files that changed mid-recomputation are preserved and picked up by the next cycle.

### DIFF
--- a/packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
+++ b/packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
@@ -66,8 +66,10 @@ export let fileMapWithFiles:
 export let currentProjectFileMapCache: FileMapCache | undefined;
 export let currentProjectGraph: ProjectGraph | undefined;
 
-const collectedUpdatedFiles = new Set<string>();
-const collectedDeletedFiles = new Set<string>();
+// Maps file path to a version counter that increments on each modification.
+// This lets us detect mid-flight re-modifications when clearing processed files.
+const collectedUpdatedFiles = new Map<string, number>();
+const collectedDeletedFiles = new Map<string, number>();
 const projectGraphRecomputationListeners = new Set<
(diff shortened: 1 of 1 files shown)


AgentEnder (approved):

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -334,11 +340,23 @@ async function processFilesAndCreateAndSerializeProjectGraph(
         throw e;
       }
     }
+
+    // Early exit if a newer recomputation has started - chain to the newer one
+    if (isStale()) {
+      return cachedSerializedProjectGraphPromise;
+    }
+
     await processCollectedUpdatedAndDeletedFiles(
```
FrozenPandaz, line 352: Double, triple check that this will indeed take the stale updated/deleted files into account.

leosvelperez: So, the changes didn't introduce new issues, but while reviewing, I did find a couple of pre-existing race conditions. So, I sent another commit to fix them too.

Hardanish-Singh (approved):

FrozenPandaz (approved):

PR #34425, opened 2026-02-12
fix(core): remove redundant `allWorkspaceFiles` from the project graph pipeline

leosvelperez:
## Current Behavior

`allWorkspaceFiles` (a flat `FileData[]` of every file in the workspace) is built, stored, copied, and passed through the project graph pipeline — even though no consumer uses it. The data is fully redundant with `fileMap.projectFileMap` + `fileMap.nonProjectFiles`. In the daemon, it's deep-copied on every graph serialization via `copyFileData()` and rebuilt on every incremental update via `buildAllWorkspaceFiles()`.

## Expected Behavior

`allWorkspaceFiles` is removed from:

- `retrieveWorkspaceFiles` return value
- `buildProjectGraphUsingProjectFileMap` parameters
- `hydrateFileMap` / `getFileMap` / `storedAllWorkspaceFiles`
- `fileMapWithFiles` daemon state and `SerializedProjectGraph` interface
- `WorkspaceFileMap` interface and `updateFileMap` return value

This eliminates redundant allocations, copies, and retained references.

Native/Rust code is unaffected — it uses `rustReferences.allWorkspaceFiles` (`ExternalObject`), which is a separate code path.

### DIFF
--- a/packages/nx/src/daemon/server/handle-hash-tasks.ts
+++ b/packages/nx/src/daemon/server/handle-hash-tasks.ts
@@ -18,7 +18,7 @@ export async function handleHashTasks(payload: {
   cwd: string;
   collectInputs?: boolean;
 }) {
-  const { error, projectGraph, allWorkspaceFiles, fileMap, rustReferences } =
+  const { error, projectGraph, rustReferences } =
     await getCachedSerializedProjectGraphPromise();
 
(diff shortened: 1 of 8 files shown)


AgentEnder (changes requested):
Discussed offline

PR #34427, opened 2026-02-12
cleanup(core): reduce terminal output duplication and allocations in task runner

leosvelperez:
## Current Behavior

Terminal output in the task runner is accumulated via repeated string concatenation (`terminalOutput += chunk`). Each `+=` on a growing string causes V8 to allocate a new, larger string and copy the old contents, resulting in O(n²) allocation behavior for tasks with large output. Additionally, `PseudoTtyProcess.onExit` didn't pass `terminalOutput` to its callbacks, forcing callers like `TaskOrchestrator` to duplicate output accumulation logic with a separate `onOutput` listener.

## Expected Behavior

- Terminal output is collected in `string[]` arrays and joined once at the end, reducing intermediate allocations from O(n²) to O(n)
- `PseudoTtyProcess.onExit` now passes `terminalOutput` as a second argument, matching the signature of other `RunningTask` implementations
- `TaskOrchestrator` no longer needs a special code path for `PseudoTtyProcess` — unified `onExit` handling for all task types
- `tui-summary-life-cycle` accumulates output in chunks during execution and stores the finalized string on task completion, allowing chunk arrays to be GC'd
- `SeriallyRunningTasks` and `RunningNodeProcess` similarly switched to chunk-based accumulation
- `BatchProcess` and `NodeChildProcessWithNonDirectOutput` lazily join and cache their terminal output

### DIFF
--- a/packages/nx/src/tasks-runner/life-cycles/tui-summary-life-cycle.ts
+++ b/packages/nx/src/tasks-runner/life-cycles/tui-summary-life-cycle.ts
@@ -51,21 +51,27 @@ export function getTuiTerminalSummaryLifeCycle({
   const inProgressTasks = new Set<string>();
   const stoppedTasks = new Set<string>();
 
+  // Chunks accumulated progressively during task execution
+  const taskOutputChunks: Record<string, string[]> = {};
+  // Finalized output strings set on task completion — read by summary print functions
   const tasksToTerminalOutputs: Record<string, string> = {};
   const tasksToTaskStatus: Record<string, TaskStatus> = {};
 
   const taskIdsInTheOrderTheyStart: string[] = [];
 
+  const getTerminalOutput = (taskId: string): string =>
+    tasksToTerminalOutputs[taskId] ?? taskOutputChunks[taskId]?.join('') ?? '';
+
   lifeCycle.startTasks = (tasks) => {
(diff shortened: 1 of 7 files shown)


packages/nx/src/tasks-runner/life-cycles/tui-summary-life-cycle.ts
```diff
@@ -51,21 +51,27 @@ export function getTuiTerminalSummaryLifeCycle({
   const inProgressTasks = new Set<string>();
   const stoppedTasks = new Set<string>();
 
+  // Chunks accumulated progressively during task execution
+  const taskOutputChunks: Record<string, string[]> = {};
```
FrozenPandaz, line 55: This is still a duplicate array of the one in the running task...

With the results now having the full terminal output.. can we not store this duplicate?

FrozenPandaz (changes requested):

packages/nx/src/tasks-runner/life-cycles/tui-summary-life-cycle.ts
FrozenPandaz: Nevermind, I see that we need the chunks for continuous tasks.

FrozenPandaz (approved):

PR #34430, opened 2026-02-12
feat(core): support q to quit in graph

AgentEnder:
## Current Behavior
TUI is used for 90% of real usage commands, and q exits it. No other command is quittable with q, but I find myself hitting Q to exit them anyways due to the tui muscle memory

## Expected Behavior
Most commands can be quit with q.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/graph/graph.ts
+++ b/packages/nx/src/command-line/graph/graph.ts
@@ -55,6 +55,7 @@ import { readFileMapCache } from '../../project-graph/nx-deps-cache';
 import { filterUsingGlobPatterns } from '../../hasher/task-hasher';
 import { ConfigurationSourceMaps } from '../../project-graph/utils/project-configuration-utils';
 import { findMatchingProjects } from '../../utils/find-matching-projects';
+import { handleQToQuit } from '../../utils/handle-q-to-quit';
 
 import { createTaskHasher } from '../../hasher/create-task-hasher';
 import { ProjectGraphError } from '../../project-graph/error-types';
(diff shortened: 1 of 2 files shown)


packages/nx/src/tasks-runner/life-cycles/dynamic-run-many-terminal-output-life-cycle.ts
```diff
@@ -327,6 +332,7 @@ export async function createRunManyDynamicOutputRenderer({
 
   lifeCycle.endCommand = () => {
     clearRenderInterval();
+    cleanupQToQuit();
```
FrozenPandaz: I'm hesitant about adding it here... so I'd drop it.

AgentEnder:
@copilot lets strip this down to just the graph cli command for now

packages/nx/src/tasks-runner/life-cycles/dynamic-run-one-terminal-output-life-cycle.ts
```diff
@@ -101,6 +101,7 @@ export async function createRunOneDynamicOutputRenderer({
   let renderDependentTargetsIntervalId: NodeJS.Timeout | undefined;
 
   const moveCursorToStartOfDependentTargetLines = () => {
+    readline.cursorTo(process.stdout, 0);
```
FrozenPandaz, line 104: This seems unrelated?

FrozenPandaz (changes requested):

PR #34443, opened 2026-02-13
PARKED FOR LATER feat(core): replace picocolors with native styleText

meeroslav:
This PR replaces the `picocolors` with native Node's `styleText` and removes the leftover references to `chalk`

## Related Issue(s)

Fixes #


AgentEnder:
Style text wasn't marked stable until node 22 - I think we should hold off until next major? I'll leave it open for Jason / Jack to weigh in, it may be fine but in node 20 it was still marked as active development.

image

image

PR #34450, opened 2026-02-13
docs(nx-dev): tech intro page structure improvements

barbados-clemens:
Work on making a tech intro pages more consistent with each other and focus on "answering the 80%" for the given technology.

Focusing on 
- Angular
- Maven/Gradle
- TS
- Vite
- Vitest/Jest

The changes are based around answering the following, where each "category" of page might have a different set of depth for the answer.

1. Why do I want to use this plugin?
  - Plugins are considered fully optional and are aimed at providing better DX for a technology, such as inferred setup, generators, migrations.
  - some plugins (like TSC) might have special call outs in some of this, but generally the same for all plugins.
2. How do I use this plugin in my workspace?
  - also pretty commonly the "same" for all plugins in terms of "setup"
  - where they differ is mostly for frameworks, e.g. Angular, React. You're looking at setting up a project to use these tools
  - For Build/Test tools you're looking at adding to an existing project, or converting from one to another. 
    - Build/Test tools are "means to an end", so should callout if the goal is tool + framework in a "new" context point to the framework based plugin page. Otherwise, show adding to an existing project like React project. 
3. What do I need to know about using this plugin?
  - understanding finer details of a plugin options, e.g. buildable & publishable
  - extra generators for the plugin. e.g. "convert-to-swc"
  - generally I like the idea of having a "CI considerations" where we talk about CI setups that can help, e.g. options or batch mode etc.

closes DOC-407


astro-docs/src/content/docs/technologies/angular/introduction.mdoc
```diff
@@ -1,123 +1,142 @@
...
-- Executors providing extra capabilities on top of the Angular Devkit builders:
-  - Provide ESBuild plugins
-  - Provide custom webpack configurations
-- Utilities for automatic workspace refactoring
-
-{% aside type="note" title="Currently using the Angular CLI?" %}
-You can easily and mostly **automatically migrate from an Angular CLI** project to Nx! Learn
-more [here](/docs/technologies/angular/migration/angular).
-{% /aside %}
+[Angular](https://angular.dev) is a framework for building web applications. Nx has first party support for Angular projects. If you're familar with the Angular CLI, `ng`, then Nx CLI will feel very familiar.
```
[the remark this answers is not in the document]
barbados-clemens: Fixed in 002f2227bc — corrected 'familar' → 'familiar'.

astro-docs/src/content/docs/technologies/typescript/introduction.mdoc
```diff
@@ -1,35 +1,61 @@
...
+description: Manage TypeScript and JavaScript libraries in your Nx workspace with automatic typechecking, builds via TSC or SWC, and support for project references and package manager workspaces.
+keywords: [typescript, javascript, js, ts]
 sidebar:
   label: Introduction
   order: 1
 filter: 'type:References'
 ---
 
-The JS plugin contains executors and generators that are useful for JavaScript/TypeScript projects in an Nx workspace.
+The TypeScript plugin for Nx, `@nx/js`, provides [generators for creating TypeScript and JavaScript projects](#generate-and-manage-typescript-libraries), and the `@nx/js/typescript` plugin automatically infers [`typecheck` and `build` tasks](#how-nxjstypescript-infers-tasks) for your projects using package manager workspaces and TypeScript project references. Maintaining TypeScript project references in a monorepo can be cumbersome, the `@nx/js` plugin comes with a [sync generator](/docs/concepts/sync-generators) to automatically upkeep these references.
```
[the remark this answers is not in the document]
barbados-clemens: Fixed in 002f2227bc — updated anchor to `#how-the-typescript-plugin-infers-tasks` (both occurrences).

astro-docs/src/content/docs/technologies/test-tools/jest/introduction.mdoc
```diff
@@ -6,389 +6,242 @@ sidebar:
...
 
-{% /tabitem %}
-{% /tabs %}
+Typically, in CI it's recommended to use `nx affected -t test --parallel=[# CPUs] --runInBand` for the best performance.
 
-```typescript
-// global-setup-swc.ts
-import { registerTsProject } from '@nx/js/src/internal';
-const cleanupRegisteredPaths = registerTsProject('./tsconfig.base.json');
+This is because each jest process creates a workers based on system resources, running multiple projects via nx and using jest workers will create too many process overall causing the system to run slower than desired. Using the `--runInBand` flag tells jest to run in a single process. You can then leverage Nx parallelism to run multiple jest projects at once.
```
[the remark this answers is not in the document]
barbados-clemens: Fixed in 002f2227bc — corrected 'creates a workers' → 'creates workers' and 'too many process' → 'too many processes'.

astro-docs/src/content/docs/technologies/java/maven/introduction.mdoc
```diff
@@ -124,14 +146,63 @@ The `@nx/maven` plugin is configured in the `plugins` array in `nx.json`.
...
+nx connect
+```
+
+### Batch Mode
+
+Batch mode runs multiple Gradle tasks in a single invocation. Enable it by setting `NX_BATCH_MODE=true`:
+
+```shell
+NX_BATCH_MODE=true nx run-many -t build test
+```
```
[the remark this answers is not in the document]
barbados-clemens: Fixed in 002f2227bc — changed 'Gradle' → 'Maven'.

astro-docs/src/content/docs/technologies/angular/introduction.mdoc
```diff
@@ -1,123 +1,142 @@
...
+By default, Nx configures ESLint for linting, Jest for unit tests. Read more of the options available for the application generator in the [angular library generators reference](/docs/technologies/angular/generators#library).
 
-- ESLint as the linter.
-- Jest as the unit test runner.
-- Cypress as the E2E test runner.
+There are also generators for setting up common items inside your projects, such as components and services:
 
-We can then serve, build, test, lint, and run e2e tests on the application with the following commands:
+```shell
+nx g @nx/angular:component libs/my-lib/src/lib/my-component --project=my-lib
```
jaysoo: `--project` is not an option, you can just pass the path without any options.

```
nx g @nx/angular:component libs/my-lib/src/lib/my-component
```

astro-docs/src/content/docs/technologies/angular/introduction.mdoc
```diff
@@ -1,123 +1,142 @@
 ---
-title: Overview of the Nx Angular Plugin
-description: The Nx Plugin for Angular contains executors, generators, and utilities for managing Angular applications and libraries within an Nx workspace.
+title: Angular Plugin for Nx
```
jaysoo, line 2: I think for Angular, we should also mention that with Nx and `@nx/angular`, when you generate projects it is more scalable since we don't have one giant `angular.json` that contains all projects.

This is the best way to build a scalable Angular monorepo.

astro-docs/src/content/docs/technologies/build-tools/vite/introduction.mdoc
```diff
@@ -1,49 +1,91 @@
...
 sidebar:
   label: Introduction
 filter: 'type:References'
 ---
 
-The Nx plugin for [Vite](https://vite.dev/) and [Vitest](https://vitest.dev/).
+[Vite](https://vite.dev/) is a fast build tool and dev server for modern web apps.
 
-[Vite](https://vite.dev/) is a build tool that aims to provide a faster and leaner development experience for modern web projects.
+The `@nx/vite` plugin adds [inferred targets](#configuration), [project configuration generators](#setup), and optional Vitest test splitting for CI.
```
jaysoo: Let's not mention Vitest here, but we can link tot he Vitest docs instead if needed. We split `@nx/vitest` from `@nx/vite` so people not using Vite for bundling can still use Vitest for unit tests.

Maybe a callout for Nx v21 and prior versions that `@nx/vite` contained what's now in `@nx/vitest`.  We'll clean this up once those versions are no longer support.

astro-docs/src/content/docs/technologies/java/maven/introduction.mdoc
```diff
@@ -40,9 +31,11 @@ You can install Nx globally. Depending on your package manager, use one of the f
 {% tabitem label="npm" %}
 
 ```shell
-npm add --global nx@latest
+npm install --global nx@latest
```
jaysoo: Maybe npm should be last for non-JS pages? Just in case people don't want to install Node packages... even though it technically is required.

I don't feel to strongly about this, so we could keep it as is for consistency across all pages.

astro-docs/src/content/docs/technologies/react/introduction.mdoc
```diff
@@ -1,144 +1,172 @@
...
+nx g @nx/react:lib libs/my-lib --bundler=rollup
+nx g @nx/react:lib libs/my-lib --publishable --importPath=@myorg/my-lib
 ```
 
-Replace `my-new-app` and `my-new-lib` with the name of your projects.
+The same [bundler options](#choose-a-bundler) apply when creating buildable libraries. Read more about the options available for the [library generator](/docs/technologies/react/generators#library).
+You can also read more about [buildable and publishable libraries](/docs/concepts/buildable-and-publishable-libraries).
 
-### Creating Hooks
+### Generate Components and Hooks
```
jaysoo: Let's leave the component/hooks generators out. They are kinda useful for Angular due boilerplate and some standard format for how Angular components are structured.

It's pretty easy to create React components from scratch, and our generators don't provide much value IMO. We may look at deprecating them in the future.

astro-docs/src/content/docs/technologies/react/introduction.mdoc
```diff
@@ -1,144 +1,172 @@
...
-Read more about [building and publishing libraries here](/docs/concepts/buildable-and-publishable-libraries).
+To start the application in development mode, run `nx serve my-app`. Read more about the options available for the [application generator](/docs/technologies/react/generators#application).
+
+#### Choose a Bundler
 
-### Creating Components
+The `--bundler` option selects the build tool for your application or buildable library. You can learn more about the specific bundler you scaffoled your project with on the bundler's plugin page:
 
-Adding a component to an existing project can be done with:
+- [Vite](/docs/technologies/build-tools/vite/introduction)
```
jaysoo: Add `(Recommended)` to Vite since it's the default we use.

astro-docs/src/content/docs/technologies/react/next/introduction.mdoc
```diff
@@ -1,365 +1,240 @@
...
+Read more about the options available for the [library generator](/docs/technologies/react/next/generators#library).
 
-{% /tabitem %}
-{% tabitem label="Using the @nx/next:server executor" %}
+## Develop Next.js Applications
 
-You can run `nx serve my-new-app` to serve a Next.js application called `my-new-app` for development. This will start the dev server at <http://localhost:4200>.
-
-To serve a Next.js application for production, add the `--prod` flag to the serve command:
+### Generate Pages and Components
```
jaysoo: I'd skip the components and pages generators. They aren't that useful compared to Angular, and we've always been missing the app router page generator.

In general we'll be deprecating these small generators in the future since AI is great at creating these.

astro-docs/src/content/docs/technologies/test-tools/vitest/introduction.mdoc
```diff
@@ -99,143 +193,49 @@ While Vitest is most often used for unit tests, there are cases where it can be
...
-### Vitest
-
-Primary configurations for Vitest will be via the `vitest.config.ts` or `vite.config.ts` file that is generated for your project. Learn more about [Vitest configurations](https://vitest.dev/config/).
-
-### Nx
-
-The Nx task options can be configured via the [project config file](/docs/reference/project-configuration) or via the command line flags.
-
-If you're using [inferred tasks](/docs/concepts/inferred-tasks), you can [provide the Vitest args](/docs/guides/tasks--caching/pass-args-to-commands) for the command you're running.
+### Task splitting with Atomizer
```
jaysoo: I don't like that this section is different from Jest. It's not just for e2e, but that would be one example of a slow test. I'd structure both Jest and Vitest the same way, where CI considerations for slow tests (e.g. e2e tests) can be solved by splitting them.

jaysoo (approved):

PR #34462, opened 2026-02-16
fix(core): prevent staggered and duplicate lines in dynamic output

leosvelperez:
## Current Behavior

- In some `nx run-many` executions, terminal output can appear staggered or visually misaligned instead of updating cleanly in place.
- For `run-many` cases that end up running a single task (especially when TUI is not active), the spinner/status line can be rendered twice.

## Expected Behavior

- Dynamic terminal output updates should remain stable and aligned, with clean in-place refreshes.
- Single-task `run-many` should display a single spinner/status line with no duplicate rendering.

### DIFF
--- a/packages/nx/src/tasks-runner/run-command.ts
+++ b/packages/nx/src/tasks-runner/run-command.ts
@@ -913,7 +913,14 @@ export function setEnvVarsBasedOnArgs(
     process.env.NX_STREAM_OUTPUT = 'true';
     process.env.NX_PREFIX_OUTPUT = 'false';
   }
-  if (nxArgs.outputStyle === 'dynamic' || nxArgs.outputStyle === 'tui') {
+  // Dynamic output styles only require forced streaming when the full TUI is
+  // active. If the TUI is disabled (e.g. single-task run-many fallback), the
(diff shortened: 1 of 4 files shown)


packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -913,7 +913,14 @@ export function setEnvVarsBasedOnArgs(
...
   }
-  if (nxArgs.outputStyle === 'dynamic' || nxArgs.outputStyle === 'tui') {
+  // Dynamic output styles only require forced streaming when the full TUI is
+  // active. If the TUI is disabled (e.g. single-task run-many fallback), the
+  // dynamic-legacy renderer should own the terminal output without interleaved
+  // task streaming.
+  if (
+    (nxArgs.outputStyle === 'dynamic' || nxArgs.outputStyle === 'tui') &&
+    isTuiEnabled()
+  ) {
```
AgentEnder: isTuiEnabled is set when running yargs middlewares, and is always false if output style is set to dynamic

```suggestion
  if (
    isTuiEnabled()
  ) {
```

AgentEnder (approved):

PR #34469, opened 2026-02-16
feat(core): automatically set up ai agents in cnw/init when run from within an ai agent

MaxKless:
We want to minimize prompts for people and agents. But we also want to help them by setting up nx config for them so their agents can work optimally.
If they're executing `nx init` or `create-nx-workspace` from within an agent, it's a reasonable assumption that they'll want the best AI config for that specific agent - so we set it up for them.


AgentEnder (approved):
Looks good

PR #34479, opened 2026-02-17
fix(core): prevent silent ts-node fallback breaking ESM plugins on No…

### LINKED ISSUE #34472: fix(core): pinned @swc-node/register peerDep (1.11.1) causes silent ts-node fallback and ERR_REQUIRE_ESM on Node 24
## Description

After upgrading from nx 22.3.3 to 22.5.1, all workspace plugins with `"type": "module"` in their `package.json` fail to load with `ERR_REQUIRE_ESM` in the isolated plugin worker process. Node version is 24.11.0 and did not change — this is purely an nx version regression.

## Root Cause

**PR #34215** (`feat(js): bump swc to latest versions`) changed the `@swc-node/register` peer dependency from `"^1.8.0"` to `"1.11.1"` (exact pin). This was a routine dependency bump, not driven by any functional requirement.

Projects that have `@swc-node/register` installed at a version satisfying `^1.8.0` but below `1.11.1` (e.g., `1.9.2`) silently lose the SWC transpiler. Because `@swc-node/register` is an **optional** peer dependency, pnpm drops it without warning when the version doesn't match. This causes `packageIsInstalled('@swc-node/register')` in `register.js` to return `false`, and `getTranspiler()` falls back to ts-node.

On Node 24 with `"type": "module"` plugins, ts-node's CJS hook intercepts `require()` of `.ts` files and throws `ERR_REQUIRE_ESM`.

### Version timeline

| Version | `@swc-node/register` peerDep | Behavior |
|---|---|---|
| ≤ 22.4.5 | `^1.8.0` (range) | ✅ Matches `@swc-node/register@1.9.2`, SWC transpiler used |
| ≥ 22.5.0 | `1.11.1` (exact pin) | ❌ Doesn't match `1.9.2`, silently dropped, falls back to ts-node |

### How the version got pinned

1. PR #34215 (commit `2bd8ef3f72`) set the pnpm catalog to `'^1.11.1'` (range with caret)
2. Commit `fdabc14892` ("chore(repo): update nx to 22.5.0-beta.2") resolved the catalog entry to the exact `'1.11.1'` (no caret), likely via release automation running `pnpm install`
3. pnpm `catalog:` on publish replaces with the literal catalog value, so the published package got `"1.11.1"` exact

### Contributing factor: `registerPluginTSTranspiler` ignores Node's native TS support

Even with the `@swc-node/register` mismatch, this wouldn't break on Node 24 if `registerPluginTSTranspiler()` in `transpiler.ts` checked `process.features?.typescript` (Node's native TS support). PR #34202 added `NX_PREFER_NODE_STRIP_TYPES` / `preferNodeStripTypes` to `registerTsProject()` in `register.ts`, but this check was not applied to `registerPluginTSTranspiler()`. This means the plugin worker always registers a transpiler (ts-node or SWC) even when Node can handle TypeScript natively.

## Error

```
ERR_REQUIRE_ESM: Must use import to load ES Module

NX  Failed to load 6 Nx plugin(s):
  - @example/my-plugin: [object Object]
  ...
```

The `[object Object]` in the error message is because `get-plugins.ts` wraps serialized error objects with `new Error(String(result.reason))`.

## Steps to Reproduce

1. Have a workspace with `@swc-node/register` at any version `>=1.8.0 <1.11.1` (e.g., `1.9.2`) installed as a transitive dependency
2. Have workspace plugins with `"type": "module"` in `package.json` that point `main`/`exports` to `.ts` source files
3. Use Node 24 (which has `process.features.typescript` truthy)
4. Upgrade nx from any version ≤ 22.4.5 to any version ≥ 22.5.0
5. Run `nx show projects` — all `"type": "module"` plugins fail with `ERR_REQUIRE_ESM`

## Suggested Fixes

1. **Restore a wider peer dependency range** for `@swc-node/register` — changing back to `"^1.8.0"` (or at minimum keeping the caret: `"^1.11.1"`) prevents silent fallback for projects with older versions installed
2. **Apply `preferNodeStripTypes` to `registerPluginTSTranspiler`** — `transpiler.ts` should check `process.features?.typescript` and skip transpiler registration on Node 22.6+, matching the logic already in `registerTsProject()` from PR #34202

## Workaround

Add `@swc-node/register@^1.11.1` and `@swc/core@^1.15.8` as explicit devDependencies to satisfy the new peer dep range.

## Environment

- **nx**: 22.5.1 (works on 22.4.5, broken on 22.5.0+)
- **Node**: 24.11.0
- **pnpm**: 10.18.3
- **OS**: macOS (arm64)

Nitin75408:
…de 24

## Current Behavior
After upgrading Nx from ≤ 22.4.5 to ≥ 22.5.0, ESM workspace plugins written in TypeScript fail to load on Node 24 with `ERR_REQUIRE_ESM`. This is caused by three compounding issues:

1. Commit 2bd8ef3f72 changed the `@swc-node/register` peer dependency from `^1.8.0` to an exact pin (`1.11.1` via `catalog:swc`). Since it's an optional peer, pnpm silently drops it when users have any other compatible version (e.g. `1.9.x`), causing `packageIsInstalled('@swc-node/register')` to return false.

2. When SWC is not detected, Nx falls back to ts-node, which registers CommonJS require hooks. On Node 24, loading `.ts` files from ESM packages via `require()` throws `ERR_REQUIRE_ESM`.

3. `registerPluginTSTranspiler()` does not check for Node's native TypeScript support (`process.features.typescript`), even though `registerTsProject()` already does. This means a transpiler is always registered even when unnecessary.

## Expected Behavior
- ESM workspace plugins load correctly on Node 24
- `@swc-node/register` is resolved with a compatible range (`^1.8.0`), not an exact pin
- When Node natively supports TypeScript (Node 22.6+), the plugin transpiler registration is skipped, avoiding the broken ts-node fallback
- When SWC is missing and Node has native TS support, `getTranspiler()` skips ts-node instead of registering CJS hooks that break ESM
- Existing setups that worked on ≤ 22.4.5 continue to work

## Related Issue(s)

Fixes #34472

### DIFF
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -75,8 +75,8 @@
     "yargs-parser": "catalog:"
   },
   "peerDependencies": {
-    "@swc-node/register": "catalog:swc",
-    "@swc/core": "catalog:swc"
+    "@swc-node/register": "^1.8.0",
+    "@swc/core": "^1.3.85"
   },
   "peerDependenciesMeta": {
     "@swc-node/register": {
--- a/packages/nx/src/plugins/js/utils/register.spec.ts
+++ b/packages/nx/src/plugins/js/utils/register.spec.ts
@@ -34,3 +34,67 @@ describe('getTsNodeCompilerOptions', () => {
     ).toEqual(['es2022']);
   });
 });
+
+describe('getTranspiler - native TypeScript support', () => {
+  const originalDescriptor = Object.getOwnPropertyDescriptor(
+    process,
+    'features'
+  );
+
+  function setProcessFeatures(value: any) {
+    Object.defineProperty(process, 'features', {
+      value,
+      writable: true,
+      configurable: true,
+    });
+  }
+
+  afterEach(() => {
+    if (originalDescriptor) {
+      Object.defineProperty(process, 'features', originalDescriptor);
+    } else {
+      delete (process as any).features;
+    }
+  });
+
+  it('should still use SWC when available even with native TS support', () => {
+    // Even when Node natively supports TS, SWC should be preferred when available
+    // because it's a faster, more complete transpiler
+    setProcessFeatures({ typescript: 'strip' });
+
+    const { getTranspiler } = require('./register');
+
+    // With SWC installed (as in this dev environment), getTranspiler should
+    // return a transpiler factory function regardless of native TS support
+    const result = getTranspiler({ module: 1 });
+    expect(result).toBeDefined();
+  });
+
+  it('should work correctly without native TS support', () => {
+    // Verify baseline behavior: getTranspiler should not throw
+    // when process.features.typescript is not available
+    setProcessFeatures(undefined);
+
+    const { getTranspiler } = require('./register');
+    expect(() => getTranspiler({ module: 1 })).not.toThrow();
+  });
+
+  it('should check process.features.typescript at call time not import time', () => {
+    // The native TS check uses process.features?.typescript which is
+    // evaluated at call time. This ensures the check is dynamic and
+    // doesn't use a cached module-level value.
+    const { getTranspiler } = require('./register');
+
+    // First call without native TS
+    setProcessFeatures(undefined);
+    const result1 = getTranspiler({ module: 1 });
+
+    // Second call with native TS - should still work (SWC takes priority)
+    setProcessFeatures({ typescript: 'strip' });
+    const result2 = getTranspiler({ module: 1 });
+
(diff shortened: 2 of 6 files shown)


Coly010:
Hey :) I'm going to close this PR because there are multiple things incorrect with it.
We want to use the catalog for instance - we can update it to a compatible range from `1.11` 
We also do not want to rely on native type stripping, especially not as a higher priority than ts-node.

PR #34485, opened 2026-02-17
fix(core): prevent changelog from resolving git tags for dependent pr…

### LINKED ISSUE #34438: `releaseChangelog()` should not attempt to get tags & generate changelogs for dependent projects when `projectsRelationship: 'independent'`
### Current Behavior

When running `releaseChangelog()` (or `nx release changelog`) on a project whose release is configured with `projectsRelationship: 'independent'`, the code attempts to get git history & the latest release tag for every dependent project, even though those dependents won't actually have changelog entries generated.

### Expected Behavior

`releaseChangelog()` should not attempt to get the latest release tag for dependent projects. That info does not get used, the git operations are heavy, and `releaseChangelog()` should not error if, say, a dependent project does not have a release yet.

### GitHub Repo

https://github.com/w-biggs/nx-changelog-issue/tree/main

### Steps to Reproduce

Say I have a project `utils` with two dependents: `icons` and `components`, and I run `nx release 1.1.0 -p @my-org/utils` to release only `utils`. You can reproduce this with:

```bash
git clone https://github.com/w-biggs/nx-changelog-issue.git
cd nx-changelog-issue
npm install
npm run demo:with-logging
```

Internally, the following code creates a `projectNodes` array with all three projects' nodes and iterates through them. For each project, it tries to find the previous git tag release, errors if it can't, then runs `generateChangelogForProjects()`.

https://github.com/nrwl/nx/blob/51420790c35e3217caa2142234c58edbd567dd36/packages/nx/src/command-line/release/changelog.ts#L453-L564

In the demo, you'll see it successfully processes `utils` (which has a tag `@my-org/utils@0.0.1`), but then errors on `@my-org/components` when trying to find its tag - even though `generateChangelogForProjects()` will never actually do anything on the dependent projects' iterations, because `projectsVersionData = utils` and `projects = [{ name: icons or components }]`, so `projectsVersionData[project.name]` is always falsy.

https://github.com/nrwl/nx/blob/51420790c35e3217caa2142234c58edbd567dd36/packages/nx/src/command-line/release/changelog.ts#L1092-L1112

You can see a rough workaround working with `npm run demo:with-fix`, which does not add dependents to the `projectNodes` array for independent projects.

### Nx Report

```shell
Node           : 24.13.0
OS             : darwin-arm64
npm            : 11.6.2
daemon         : Available

nx             : 22.5.0
@nx/js         : 22.5.0
@nx/workspace  : 22.5.0
@nx/devkit     : 22.5.0
typescript     : 5.9.3
Registered Plugins:
@nx/js/typescript
Cache Usage: 0.00 B / 46.04 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Nitin75408:
…ojects in independent releases

## Current Behavior

When running `nx release changelog` (or `releaseChangelog()`) for a project in an independent release group, Nx incorrectly includes dependent projects in the changelog processing list. It attempts to resolve git tags and fetch git history for these dependents — even though they will never have changelog entries generated. If a dependent has never been released (no git tag), Nx throws a fatal error: "Unable to determine the previous git tag."

## Expected Behavior

For independent release groups, only the explicitly targeted projects should have git tags resolved and changelog entries generated. Dependent projects should be excluded from changelog processing entirely, since they are versioned independently and may never have been released.

## Related Issue(s)

Fixes #34438

### DIFF
--- a/packages/nx/src/command-line/release/changelog.ts
+++ b/packages/nx/src/command-line/release/changelog.ts
@@ -370,7 +370,8 @@ export function createAPI(
       preid: string | undefined,
       checkAllBranchesWhen: CheckAllBranchesWhen,
       requireSemver: boolean,
-      strictPreid: boolean
+      strictPreid: boolean,
+      projectName?: string
     ): Promise<string | null> => {
       if (fromSHACache.has(cacheKey)) {
         return fromSHACache.get(cacheKey);
@@ -386,6 +387,8 @@ export function createAPI(
         requireSemver,
         strictPreid,
         useAutomaticFromRef,
+        projectName,
+        verbose: args.verbose,
       });
       fromSHACache.set(cacheKey, sha);
       return sha;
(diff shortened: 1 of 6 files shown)


jaysoo (approved):
Good fix. `releaseChangelog()` with independent projects shouldn't error trying to find git tags for dependents that were never independently released. The `getProjectsForChangelog()` approach correctly excludes dependents for independent groups, and the defense-in-depth guard is a nice touch.

Tests are thorough. One cleanup: please remove the unrelated pnpm-lock.yaml changes to keep the diff focused.

LGTM otherwise.

jaysoo:
Close this as I do not believe this is the right fix, and the description does not contain how this fix works.

PR #34487, opened 2026-02-17
feat(misc): use caret range for swc dependencies in pnpm catalog

### LINKED ISSUE #34472: fix(core): pinned @swc-node/register peerDep (1.11.1) causes silent ts-node fallback and ERR_REQUIRE_ESM on Node 24

Coly010:
Use a range for the swc dependencies

Fixes #34472

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -218,14 +218,14 @@ catalogs:
       version: 1.1.4
   swc:
     '@swc-node/register':
-      specifier: 1.11.1
+      specifier: ^1.11.1
       version: 1.11.1
     '@swc/cli':
       specifier: 0.8.0
       version: 0.8.0
     '@swc/core':
-      specifier: 1.15.8
-      version: 1.15.8
+      specifier: ^1.15.8
+      version: 1.15.10
     '@swc/helpers':
       specifier: 0.5.18
       version: 0.5.18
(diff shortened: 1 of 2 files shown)


comp615:
Does the version bump still need to happen? In theory for peer deps it'd be nice to be the lowest functional requirement (you can still install a higher one in the repo itself)

Coly010:
Yes :)

FrozenPandaz (approved):

PR #34488, opened 2026-02-17
feat(core): improve codex support for configure-ai-agents

MaxKless:
Codex has only basic MCP/AGENTS.md support right now. Also, because it used to have only global-level config files, we had some extra logic around configuring that.

We want Codex to get the latest skills too (they don't support custom subagents yet, though) and use project-level config files that they now support.


AgentEnder (approved):

FrozenPandaz:
@MaxKless Please add a PR description here

PR #34491, opened 2026-02-17
fix(core): properly quote shell metacharacters in CLI args passed to tasks

### LINKED ISSUE #26682: Playwright --grep flag doesn't work with multiple tags when run via nx command
### Current Behavior

Playwright accepts multiple tags in grep if they are separated by a pipe and wrapped in quotes. When command is run with nx the quotes are stripped and the command fails.

Both commands below fail:

`nx affected -t e2e --grep="@visual-regression|@smoke" `
`nx affected -t e2e -- --grep="@visual-regression|@smoke" `

```
 /bin/sh: @smoke: command not found
      node:events:496
            throw er; // Unhandled 'error' event
            ^
...
Node.js v20.12.2
      Warning: command "npx playwright test --grep=@visual-regression|@smoke" exited with non-zero status code
```

When run directly with playwright from working directory it succeeds:

`npx playwright test --grep "@visual-regression|@smoke"`

### Expected Behavior

Nx playwright executor should support multiple grep arguments and not strip the quotes

### GitHub Repo

_No response_

### Steps to Reproduce

1. Add a playwright e2e project
2. Create two tests with different tags
3. Run the command `npx nx affected e2e --grep "@tag1|@tag2"`
4. Notice failure

### Nx Report

```shell
Node   : 20.12.2
OS     : darwin-arm64
npm    : 10.5.0

nx (global)        : 18.3.1
nx                 : 18.3.1
@nx/js             : 18.3.1
@nx/jest           : 18.3.1
@nx/linter         : 18.3.1
@nx/eslint         : 18.3.1
@nx/workspace      : 18.3.1
@nx/cypress        : 18.3.1
@nx/devkit         : 18.3.1
@nx/eslint-plugin  : 18.3.1
@nx/next           : 18.3.1
@nx/playwright     : 18.3.1
@nx/plugin         : 18.3.1
@nx/react          : 18.3.1
@nx/storybook      : 18.3.1
@nrwl/tao          : 18.3.1
@nx/vite           : 18.3.1
@nx/web            : 18.3.1
@nx/webpack        : 18.3.1
typescript         : 5.3.3
```

### Failure Logs

_No response_

### Package Manager Version

_No response_

### Operating System

- [X] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

### LINKED ISSUE #32305: Quoted CLI flags are not passed through to tasks correctly
### Current Behavior

Duplicate of https://github.com/nrwl/nx/issues/26682

Playwright accepts a `--grep="pattern"` flag. When passed via nx it doesn't work.

e.g. `nx playwright-test app --grep="should show button"`

```
Error: No tests found.
Make sure that arguments are regular expressions matching test files.
You may need to escape symbols like "$" or "*" and quote the arguments.
```

if we double quote the parameter it works as expected, e.g.

`nx playwright-test app --grep='"should show button"'`

```
  1 passed (3.1s)
```

### Expected Behavior

The arguments should be passed through without needing double quoting

### GitHub Repo

_No response_

### Steps to Reproduce

1. set up nx repo with an app and playwright
2. add a script to package.json to run playwright
3. run the script via nx and pass in the `--grep` flag with a string that should match a test name

### Nx Report

```shell
Node           : 24.4.1
OS             : darwin-arm64
yarn           : 4.1.1

nx                     : 21.3.10
@nx/js                 : 21.3.10
@nx/jest               : 21.3.10
@nx/eslint             : 21.3.10
@nx/workspace          : 21.3.10
@nx/cypress            : 21.3.10
@nx/devkit             : 21.3.10
@nx/module-federation  : 21.3.10
@nx/plugin             : 21.3.10
@nx/react              : 21.3.10
@nx/storybook          : 21.3.10
@nx/vite               : 21.3.10
@nx/web                : 21.3.10
typescript             : 5.4.2
Registered Plugins:
@nx/js/typescript
@nx/jest/plugin
@nx/react/router-plugin
@nx/storybook/plugin
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

baer:
## Current Behavior

When CLI arguments containing shell metacharacters (like `|`, `&`, `$`, `;`, `*`, etc.) are passed through Nx to underlying tasks, they are not properly quoted, causing shell interpretation errors.

For example, running:
```bash
nx test app --grep="@tag1|@tag2"
```

Would fail with `/bin/sh: @smoke: command not found` because the pipe character `|` was interpreted by the shell as a pipe operator instead of being passed as a literal string to the underlying command.

Users had to use awkward double-quoting workarounds like `--grep='"@tag1|@tag2"'` to get the expected behavior.

## Expected Behavior

CLI arguments containing shell metacharacters should be automatically quoted before being passed to underlying commands, so that:
```bash
nx test app --grep="@tag1|@tag2"
```

Works correctly and passes `--grep="@tag1|@tag2"` to the underlying test runner without shell interpretation.

## Related Issue(s)

Fixes #32305
Fixes #26682

## Implementation Details

- Created a shared `needsShellQuoting()` utility in `packages/nx/src/utils/shell-quoting.ts` that detects shell metacharacters
- Updated `wrapArgIntoQuotesIfNeeded()` in `run-commands.impl.ts` to use the shared utility
- Updated `stringShouldBeWrappedIntoQuotes()` in `serialize-overrides-into-command-line.ts` to use the shared utility
- Fixed a bug where `arg.split('=')` would incorrectly split values containing `=` (e.g., `--define=FOO=bar|baz`)
- Added proper escaping of embedded double quotes when wrapping values
- Added comprehensive test coverage

### Shell metacharacters now handled:
`|` `&` `;` `<` `>` `(` `)` `$` `` ` `` `\` `"` `'` `*` `?` `[` `]` `{` `}` `~` `#` `!` and whitespace

### DIFF
--- a/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
@@ -657,6 +657,30 @@ describe('Run Commands', () => {
         )
       ).toEqual('echo --define="FOO=bar|baz"');
     });
+
+    it('should not re-quote word-split fragments of a single-quoted JSON value', () => {
+      // Simulates what happens when the shell word-splits:
+      //   --config \'{"env":{"cliArg":"i am from the cli args"}}\'
+      // The shell produces these separate argv entries because \" makes "
+      // literal but doesn't prevent word splitting:
+      expect(
+        interpolateArgsIntoCommand(
+          'echo',
+          {
+            __unparsed__: [
+              '--config',
+              '\'{"env":{"cliArg":"i',
+              'am',
+              'from',
+              'the',
+              'cli',
+              'args"}}\'',
+            ],
+          } as any,
+          true
+        )
+      ).toEqual(`echo --config '{"env":{"cliArg":"i am from the cli args"}}'`);
+    });
   });
 
   describe('--color', () => {
--- a/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
+++ b/packages/nx/src/executors/run-commands/run-commands.impl.spec.ts
@@ -584,6 +584,79 @@ describe('Run Commands', () => {
         )
       ).toEqual(`echo --hello="test 123" "hello world" "random config" 456`); // should wrap aroound __unparsed__ args with key value
     });
+
+    it.each([
+      ['pipe (|)', '--grep=@tag1|@tag2', 'echo --grep="@tag1|@tag2"'],
+      ['ampersand (&)', '--flag=a&b', 'echo --flag="a&b"'],
+      ['dollar sign ($)', '--path=$HOME/dir', 'echo --path="$HOME/dir"'],
+      ['semicolon (;)', '--cmd=echo;ls', 'echo --cmd="echo;ls"'],
+      ['parentheses', '--expr=(a+b)', 'echo --expr="(a+b)"'],
+      ['asterisk (*)', '--pattern=*.txt', 'echo --pattern="*.txt"'],
+      ['backtick (`)', '--cmd=`pwd`', 'echo --cmd="`pwd`"'],
+      ['angle brackets (>)', '--compare=a>b', 'echo --compare="a>b"'],
+      [
+        'question mark (?)',
+        '--pattern=file?.txt',
+        'echo --pattern="file?.txt"',
+      ],
+      [
+        'square brackets ([])',
+        '--pattern=[abc].txt',
+        'echo --pattern="[abc].txt"',
+      ],
+      ['hash (#)', '--tag=#important', 'echo --tag="#important"'],
+      ['tilde (~)', '--path=~/documents', 'echo --path="~/documents"'],
+      ['newline', '--msg=hello\nworld', 'echo --msg="hello\nworld"'],
+      ['tab', '--msg=hello\tworld', 'echo --msg="hello\tworld"'],
+    ])('should wrap shell metacharacter %s in quotes', (_, input, expected) => {
+      expect(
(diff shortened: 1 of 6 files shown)


jaysoo (approved):
Solid fix with excellent test coverage. The `needsShellQuoting()` function with comprehensive POSIX metacharacter regex is a big improvement over the old check (which only handled spaces, `{`, and `"`). The fix for splitting on first `=` only and escaping existing double quotes before wrapping is well thought through.

Minor note for future consideration: double-quoting doesn't prevent `$` and backtick expansion in POSIX shells. This is a pre-existing limitation (not introduced by this PR), but a brief comment in the code noting this would help future maintainers.

LGTM.

leosvelperez:
@baer thanks for contributing!

I fixed an issue caught by our e2e test suite and added additional tests to prevent regressions. Could you re-verify on your end that everything still works for you?

baer:
Thanks for the fixes, @leosvelperez\! I verified locally, and everything still works. The `rejoinQuotedFragments` handling for word-split quoted values is a great catch. I hadn't considered the case where shell word-splitting would fragment a single-quoted JSON value before the quoting logic could process it. The `isAlreadyQuoted` guard checking both ends is also a good improvement over my first-char-only check. Appreciate the additional test coverage. Glad for the E2E rn.

leosvelperez (approved):
Thanks for the contribution!

PR #34500, opened 2026-02-18
fix(nx-dev): update breadcrumb links to match sidebar

barbados-clemens:
navigation of the breadcrumbs could lead to confusing state since they were based around the folder structure.

Breadcrumbs are now based around the sidebar structure so they match the hierarchy of content. 

Note: I left the existing index file based route pages in place in case there are any links people have booked marked/linked to in other locations. these will get cleaned up when we finally rewrite all the URLs to their new content locations

### DIFF
--- a/astro-docs/markdoc.config.mjs
+++ b/astro-docs/markdoc.config.mjs
@@ -244,6 +244,15 @@ export default defineMarkdocConfig({
         },
       },
     },
+    sidebar_group_cards: {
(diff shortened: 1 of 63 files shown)


astro-docs/markdoc.config.mjs
barbados-clemens, line 1: added new component to handle instead of making the index card component dual purposed atm. felt easier to maintain for the different impl needs.

jaysoo (approved):

PR #34503, opened 2026-02-18
feat(core): add safe plugin cache write utilities with LRU eviction

AgentEnder:
## Current Behavior

 Plugin cache writes (`writeJsonFile`, `writeFileSync`) across the codebase have inconsistent error handling:
- Some throw on failure, aborting project graph calculation entirely
- Some silently swallow errors, leaving corrupted cache files on disk
- No mechanism exists to recover from oversized or corrupted caches
- `nx-deps-cache.ts` retries 5 times then throws, crashing the graph

## Expected Behavior

- Plugin cache write failures **never** abort graph calculation — they warn and continue
- A centralized `safeWritePluginCache` utility handles all hash-map plugin caches with a 3-step strategy:
1. Attempt full write
2. On failure: evict oldest 50% of entries (LRU) and retry
3. On second failure: wipe cache file, log warning, return without throwing
- `PluginCache<T>` class wraps cache data with a Proxy that transparently tracks access order, enabling true LRU
eviction (not just insertion order)
- Access order is stored as a simple `string[]` array — front is oldest, back is most recent
- Backward-compatible with 3 on-disk formats: legacy plain `Record`, timestamp-based `{ entries, accessedAt }`, and
current `{ entries, accessOrder }`
- All plugin cache consumers migrated: package-json, js/lockfile, dotnet, cypress, playwright, gradle, maven
- `nx-deps-cache.ts` changed from throw-after-retries to warn + cleanup
- New utilities exported via `@nx/devkit` internals for plugin authors

## Related Issue(s)

Fixes NXC-3833"

### DIFF
--- a/packages/cypress/src/plugins/plugin.ts
+++ b/packages/cypress/src/plugins/plugin.ts
@@ -8,9 +8,7 @@ import {
   normalizePath,
   type NxJsonConfiguration,
   type ProjectConfiguration,
-  readJsonFile,
   type TargetConfiguration,
-  writeJsonFile,
 } from '@nx/devkit';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
 import { loadConfigFile } from '@nx/devkit/src/utils/config-utils';
@@ -19,6 +17,7 @@ import { getLockFileName } from '@nx/js';
 import { readdirSync } from 'fs';
 import { hashObject } from 'nx/src/devkit-internals';
 import { workspaceDataDirectory } from 'nx/src/utils/cache-directory';
+import { PluginCache, readPluginCache } from 'nx/src/utils/plugin-cache-utils';
 import { globWithWorkspaceContext } from 'nx/src/utils/workspace-context';
 import { dirname, join, relative } from 'path';
 import { NX_PLUGIN_OPTIONS } from '../utils/constants';
@@ -29,20 +28,6 @@ export interface CypressPluginOptions {
   openTargetName?: string;
   componentTestingTargetName?: string;
   ciComponentTestingTargetName?: string;
-}
-
-function readTargetsCache(cachePath: string): Record<string, CypressTargets> {
-  try {
-    return process.env.NX_CACHE_PROJECT_GRAPH !== 'false'
-      ? readJsonFile(cachePath)
-      : {};
-  } catch {
-    return {};
-  }
-}
-
-function writeTargetsToCache(cachePath: string, results: CypressTargets) {
-  writeJsonFile(cachePath, results);
 }
 
 const cypressConfigGlob = '**/cypress.config.{js,ts,mjs,cjs}';
@@ -67,17 +52,17 @@ export const createNodes: CreateNodesV2<CypressPluginOptions> = [
       workspaceDataDirectory,
       `cypress-${optionsHash}.hash`
     );
-    const targetsCache = readTargetsCache(cachePath);
+    const pluginCache = readPluginCache<CypressTargets>(cachePath);
     try {
       return await createNodesFromFiles(
         (configFile, options, context) =>
-          createNodesInternal(configFile, options, context, targetsCache),
+          createNodesInternal(configFile, options, context, pluginCache),
         configFiles,
         options,
         context
       );
     } finally {
-      writeTargetsToCache(cachePath, targetsCache);
+      pluginCache.writeToDisk(cachePath);
     }
   },
 ];
@@ -88,7 +73,7 @@ async function createNodesInternal(
   configFilePath: string,
   options: CypressPluginOptions,
   context: CreateNodesContextV2,
-  targetsCache: CypressTargets
+  pluginCache: PluginCache<CypressTargets>
 ) {
   options = normalizeOptions(options);
   const projectRoot = dirname(configFilePath);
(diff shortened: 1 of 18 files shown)


packages/gradle/src/plugin-v1/nodes.ts
```diff
@@ -77,7 +69,9 @@ export const createNodesV2: CreateNodesV2<GradlePluginOptions> = [
       workspaceDataDirectory,
       `gradle-${optionsHash}.hash`
     );
-    const targetsCache = readTargetsCache(cachePath);
+    const pluginCache =
+      readPluginCache<Partial<ProjectConfiguration>>(cachePath);
+    const targetsCache = pluginCache.data;
```
AgentEnder: Lets derive the cache path from within the read fn

```suggestion
    const pluginCache =
      readPluginCache<Partial<ProjectConfiguration>>(prefix: string, hashContext: any);
    const targetsCache = pluginCache.data;
```

packages/nx/src/utils/plugin-cache-utils.ts
```diff
@@ -0,0 +1,275 @@
...
+
+    const toDelete = this.accessOrder.splice(
+      0,
+      Math.ceil(this.accessOrder.length / 2)
+    );
+    for (const key of toDelete) {
+      delete this.entries[key];
+    }
+    return new PluginCache<T>(this.entries, this.accessOrder);
+  }
```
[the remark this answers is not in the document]
AgentEnder: It mutates, should be fine though?

packages/cypress/src/plugins/plugin.ts
```diff
@@ -67,17 +52,17 @@ export const createNodes: CreateNodesV2<CypressPluginOptions> = [
       workspaceDataDirectory,
       `cypress-${optionsHash}.hash`
     );
-    const targetsCache = readTargetsCache(cachePath);
+    const pluginCache = readPluginCache<CypressTargets>(cachePath);
```
FrozenPandaz: Can we change this to `new PluginCache(cachePath)`?

packages/nx/src/utils/plugin-cache-utils.ts
```diff
@@ -0,0 +1,251 @@
...
+ */
+export function readPluginCache<T>(cachePath: string): PluginCache<T> {
+  try {
+    if (
+      process.env.NX_CACHE_PROJECT_GRAPH === 'false' ||
+      !existsSync(cachePath)
+    ) {
+      return new PluginCache<T>();
+    }
+    const raw = JSON.parse(readFileSync(cachePath, 'utf-8'));
```
FrozenPandaz: use readJsonFile

packages/nx/src/utils/plugin-cache-utils.ts
```diff
@@ -0,0 +1,251 @@
...
+      raw &&
+      typeof raw === 'object' &&
+      'entries' in raw &&
+      'accessOrder' in raw &&
+      Array.isArray(raw.accessOrder)
+    ) {
+      return new PluginCache<T>(raw.entries, raw.accessOrder);
+    }
+
+    // Previous timestamp format: { entries, accessedAt }
```
FrozenPandaz: we don't need this

packages/nx/src/utils/plugin-cache-utils.ts
```diff
@@ -0,0 +1,251 @@
...
+      'accessedAt' in raw
+    ) {
+      const accessOrder = Object.keys(raw.entries).sort(
+        (a, b) => (raw.accessedAt[a] ?? 0) - (raw.accessedAt[b] ?? 0)
+      );
+      return new PluginCache<T>(raw.entries, accessOrder);
+    }
+
+    // Legacy format: plain Record<string, T>
+    if (raw && typeof raw === 'object' && !Array.isArray(raw)) {
```
FrozenPandaz: else make a new one

packages/nx/src/utils/plugin-cache-utils.ts
```diff
@@ -0,0 +1,251 @@
...
+        entries: this.entries,
+        accessOrder: [...this.accessOrder],
+      });
+    } catch (e) {
+      // RangeError → string too large, recoverable via eviction
+      if (e instanceof RangeError) {
+        const reduced = this.evictOldestHalf();
+        try {
+          content = JSON.stringify(reduced);
+          logger.warn(
```
FrozenPandaz: No warning

FrozenPandaz (changes requested):

packages/nx/src/utils/plugin-cache-utils.ts
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Evicts the oldest 50% of entries (front of the access-order queue)
+   * and returns the remaining entries + accessOrder as a plain object.
+   */
+  private evictOldestHalf(): PluginCacheData<T> {
+    const accessOrderArr = [...this.accessOrder];
+    if (accessOrderArr.length === 0) return { entries: {}, accessOrder: [] };
+
+    const cutoff = Math.ceil(accessOrderArr.length / 2);
+    const remaining = accessOrderArr.slice(cutoff);
```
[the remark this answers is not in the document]
AgentEnder: Nope, this is correct - if the thing is too long with only 1 project we have to just dump it

FrozenPandaz (approved):

PR #34517, opened 2026-02-19
fix(core): prevent WASI fallback from hanging in CI environments

Shadowgandor:
## Current Behavior

When the native binary for a platform is not available, Nx falls back to loading WebAssembly (WASI). This fallback can hang indefinitely in CI environments with no error message or timeout, causing CI pipelines to time out after 20+ minutes with zero output.

This issue affects all Nx v21+ versions and has been reported by multiple users across different CI environments (Bamboo, GitLab CI, GitHub Actions, Azure Pipelines).

## Expected Behavior

Nx should fail fast with a clear error message when the native binary is missing, providing actionable instructions to fix the issue instead of hanging silently.

## Solution

This PR adds a check in `packages/nx/src/native/index.js` that detects when the native binary is missing **before** attempting to load it. Instead of silently falling back to WASI (which hangs), Nx now:

1. **Fails fast** with a clear, actionable error message
2. Tells users exactly which package to install (e.g., `@nx/nx-linux-x64-gnu`)
3. Provides instructions for npm/yarn/pnpm
4. Links to the troubleshooting documentation

Example error message:
```
Nx could not find the native binary for your platform (linux-x64).

The expected package "@nx/nx-linux-x64-gnu" is not installed.

This commonly happens in CI environments when optional dependencies are not installed.

To fix this issue, try one of the following:

1. Explicitly install the platform-specific package:
   npm install -D @nx/nx-linux-x64-gnu

2. If using npm, ensure optional dependencies are installed:
   npm install --include=optional

3. If you want to allow Nx to fall back to WebAssembly (may cause hangs in CI):
   Set the environment variable: NX_NATIVE_REQUIRE_ALLOW_FALLBACK=true

For more information, see: https://nx.dev/troubleshooting/troubleshoot-nx-install-issues
```

### New Environment Variable

`NX_NATIVE_REQUIRE_ALLOW_FALLBACK=true` - Users who need WASI fallback can opt-in (with a warning that it may hang).

## Related Issue(s)

Fixes #32750

🤖 Generated with [Claude Code](https://claude.ai/code)


jaysoo (commented):
I understand the motivation — silent WASI fallback that hangs in CI is a bad experience. However, there are several concerns:

1. **No tests** for the new detection logic.
2. **Duplicates platform/arch mapping** that already exists in napi-rs generated code.
3. **Modifies quasi-generated `index.js`** — changes here may be overwritten by napi-rs.
4. **Musl detection** (`ldd --version`) is fragile across environments.
5. **Treats the symptom, not the cause** — the real question is why WASI hangs, not how to avoid falling back to it.

@AgentEnder is assigned to the underlying issue, and the team may want a different approach (e.g., fixing the WASI runtime itself, or making the native binary available for more platforms). I'd suggest coordinating with the team on the desired direction before investing more effort here.

AgentEnder:
Hey, like jack said the proper direction here is to fix the hanging wasi build, not skip it. We are looking into why its hanging, but do not have a concrete answer yet.

PR #34523, opened 2026-02-19
fix(core): use recursive FSEvents on macOS instead of non-recursive kqueue

### LINKED ISSUE #34522: Native file watcher broken on macOS in large monorepos since 22.5.0 (kqueue scaling issue from #34329)
## Current Behavior

After upgrading from Nx 22.4.5 to 22.5.0, the daemon's native file watcher no longer detects file changes on macOS in large monorepos. The watcher subscribes successfully but never fires change callbacks. This breaks `nx watch` and `watch-deps` completely.

**This is a scale-dependent bug**: it works fine in small workspaces (~30 directories) but breaks silently in large ones (~5,250+ directories).

## Expected Behavior

The native file watcher should detect file modifications and deliver change events, as it did in 22.4.x.

## Root Cause

PR #34329 switched all watched paths from recursive to non-recursive via `WatchedPath::non_recursive()`. On macOS, the `notify` crate uses **FSEvents** for recursive watches but **kqueue** for non-recursive watches. kqueue silently fails at scale due to vnode table pressure.

Apple's own documentation explicitly recommends FSEvents over kqueue for large hierarchies: *"If you are monitoring a large hierarchy of content, you should use file system events instead."* ([Kernel Queues docs](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html))

### Why kqueue fails silently

Each non-recursive kqueue watch requires `open(path, O_EVTONLY)` which holds a vnode reference. On the tested system, `kern.num_vnodes == kern.maxvnodes` (263,168), meaning the vnode table is saturated. Under vnode pressure, the kernel recycles vnodes with kqueue watches attached without notifying the watcher. No error is returned, no partial delivery occurs.

### Confirmed: event filter is NOT the problem

We verified that the `watch_filterer.rs` macOS `FileEventKind` allowlist is correct. With recursive FSEvents watches, macOS emits specific event kinds (`Create(File)`, `Modify(Data(Content))`, `Remove(File)`, `Modify(Name(Any))`) that the current filter already handles. Zero events are rejected. The `Modify(Any)` / `Create(Any)` variants seen with kqueue are not needed with FSEvents.

## Confirmed: Size-Dependent

| | **Nx 22.4.5** | **Nx 22.5.1** |
|---|---|---|
| **Small repo (6 projects, ~30 dirs)** | Works (recursive FSEvents) | **Works** (~30 non-recursive kqueue watches) |
| **Large repo (354 projects, ~5,250 dirs)** | Works (recursive FSEvents) | **Broken** (~5,250 non-recursive kqueue watches, zero events) |

## Bisect Results

| Version | Watcher Works? |
|---------|---------------|
| 22.4.5  | Yes |
| **22.5.0** | **No** |
| 22.5.1  | No |

## Fix

**PR #34523** uses platform-conditional watch modes:
- **macOS:** Single recursive watch on the workspace root (uses FSEvents natively)
- **Linux/Windows:** Non-recursive per-directory watches (preserves #33781 inotify fix)

This also improves daemon startup on macOS from ~10 minutes to <1 second by skipping directory enumeration (FSEvents watches the entire tree from one root path).

## Steps to Reproduce

1. Use a pnpm monorepo on macOS (Apple Silicon) with 100+ workspace projects
2. Upgrade to Nx 22.5.0+
3. `nx reset && nx daemon --start`
4. Run: `nx watch --all -- echo "$NX_FILE_CHANGES"`
5. Edit any source file
6. Observe: no output, daemon log shows no file change events

## Nx Report

```
Node           : 24.11.0
OS             : darwin-arm64 (macOS 26.3 Tahoe)
pnpm           : 10.18.3
nx             : 22.5.1
```

## Workaround

Pin Nx to 22.4.5 until #34523 is merged.

## Related

- #33781 - Daemon file watcher slow to start (Linux inotify, the issue #34329 was fixing)
- #34329 - PR that introduced the regression
- [notify-rs/notify#596](https://github.com/notify-rs/notify/issues/596) - kqueue "Too many open files" at scale

comp615:
## Current Behavior

Since Nx 22.5.0, the daemon's native file watcher silently drops all file change events on macOS in large monorepos (~5,250+ watched directories). `nx watch`, `nx serve`, and any daemon-dependent file watching is broken.

The root cause is that #34329 switched all watched paths to `WatchedPath::non_recursive()`. On macOS, the `notify` crate uses **kqueue** for non-recursive watches instead of **FSEvents**. kqueue silently fails at scale due to vnode table pressure (`kern.num_vnodes == kern.maxvnodes`), causing the daemon to never detect file changes.

This is a **scale-dependent** bug: it works fine in small workspaces (~30 directories) but breaks silently in large ones.

| | **Nx 22.4.5** | **Nx 22.5.0+** |
|---|---|---|
| **Small repo (~30 dirs)** | Works (FSEvents) | Works (~30 kqueue watches) |
| **Large repo (~5,250+ dirs)** | Works (FSEvents) | **Broken** (kqueue silently drops all events) |

## Expected Behavior

The macOS file watcher should detect file creates, modifications, and deletions at any scale, matching the behavior of Nx 22.4.x.

## Fix

Use platform-conditional watch modes:
- **macOS:** Single recursive watch on the workspace root (uses FSEvents natively)
- **Linux/Windows:** Non-recursive per-directory watches (preserves the #33781 inotify fix)

On macOS, FSEvents handles recursive watching from a single root path, so directory enumeration and dynamic registration are skipped entirely. This also improves daemon startup time on macOS from ~10 minutes to <1 second in a 354-project monorepo.

### What changed in `watcher.rs`

1. **Initial pathset:** On macOS, watch only the root directory recursively via FSEvents instead of enumerating all directories for non-recursive kqueue watches.
2. **Dynamic directory registration (`on_action`):** Wrapped in `#[cfg(not(target_os = "macos"))]` since FSEvents already watches the full tree.

Linux/Windows behavior is completely unchanged.

### Why the event filter is fine as-is

We verified that with recursive FSEvents watches, macOS emits specific `FileEventKind` variants (`Create(File)`, `Modify(Data(Content))`, `Remove(File)`, `Modify(Name(Any))`) that the current `watch_filterer.rs` already handles correctly. Zero events were rejected by the catch-all. The `Modify(Any)` / `Create(Any)` variants are kqueue artifacts that are not needed with FSEvents.

### Why kqueue fails silently

Apple's [File System Events Programming Guide](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html) explicitly recommends FSEvents over kqueue for large hierarchies: *"If you are monitoring a large hierarchy of content, you should use file system events instead."* kqueue requires `open(path, O_EVTONLY)` per watched directory. Under vnode table pressure, the kernel recycles vnodes with kqueue watches attached without notifying the watcher. There is no error, no partial delivery, and no diagnostic signal.

## Tested on

- macOS 26.3 (Tahoe), Apple Silicon (arm64), APFS
- 354-project pnpm monorepo (~19,865 non-ignored directories)
- Verified: file modifications, file creates, and file deletes all detected
- Daemon init time: ~10 min (with enumeration) -> <1s (with root-only FSEvents watch)

## Related Issue(s)

Fixes #34522

### DIFF
--- a/packages/nx/src/native/watch/watcher.rs
+++ b/packages/nx/src/native/watch/watcher.rs
@@ -201,20 +201,26 @@ impl Watcher {
                 return action;
             }
 
-            // Check for new directory creation events and dynamically register watches.
-            // Without this, non-recursive watches would miss changes in newly created dirs.
-            let new_directories = extract_new_directories(&action.events, &ignore_globs_for_action);
+            // On macOS, FSEvents watches the entire tree recursively from the root,
+            // so we don't need to dynamically register new directories.
+            #[cfg(not(target_os = "macos"))]
+            {
+                // Check for new directory creation events and dynamically register watches.
+                // Without this, non-recursive watches would miss changes in newly created dirs.
+                let new_directories =
+                    extract_new_directories(&action.events, &ignore_globs_for_action);
 
-            if !new_directories.is_empty() {
-                let mut path_set = watched_path_set_for_action.lock();
-                path_set.extend(new_directories);
-                trace!(
-                    count = path_set.len(),
-                    "updating pathset with new directories"
-                );
-                watch_exec_for_action
-                    .config
-                    .pathset(path_set.iter().map(|p| WatchedPath::non_recursive(p)));
+                if !new_directories.is_empty() {
+                    let mut path_set = watched_path_set_for_action.lock();
+                    path_set.extend(new_directories);
+                    trace!(
+                        count = path_set.len(),
+                        "updating pathset with new directories"
+                    );
+                    watch_exec_for_action
+                        .config
+                        .pathset(path_set.iter().map(|p| WatchedPath::non_recursive(p)));
+                }
             }
 
             let mut origin_path = origin.clone();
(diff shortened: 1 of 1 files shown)


comp615 (commented):
Note: I feel like there might be a better refactor / solution here (e.g. pull them all up since they are common? Should `FileEventKind::Modify(ModifyKind::Data(_)) => continue,` even exist now?)...but I didn't have a full understanding and this is a narrow change. Feel free to build upon!

FrozenPandaz (approved):
Thank you! It makes sense to stick with kqueue recursive watcher for macos.

I'm going to follow up with some unit tests to prevent this from happening in the future.

Thank you for your contribution! 👍

PR #34530, opened 2026-02-20
fix(nextjs): use out/ as default output dir when output is export

asumaran:
## Current Behavior

The `getOutputs()` function in the `@nx/next` plugin only reads `distDir` from the resolved Next.js config to determine the build output directory. When a project uses `output: 'export'` (for [static site generation](https://nextjs.org/docs/app/building-your-application/deploying/static-exports)), Next.js generates output in `out/` by default, but the plugin ignores this and always defaults to `.next/`.

This causes Nx to cache/restore `.next/` instead of `out/`, which breaks deployment pipelines that depend on the `out/` directory for static export builds.

## Expected Behavior

The plugin should resolve the output directory the way `next build` does. Next.js decides it in `hasCustomExportOutput`:

```js
return config.output === 'export' && config.distDir !== '.next';
```

When that holds, `distDir` is the export directory and `.next` is reused internally for intermediate artifacts. Otherwise the export lands in `out`.

Using that same condition, rather than just checking whether `distDir` is set, covers one case worth spelling out. `withNx` returns `{ distDir: '.next', ...userConfig }` while `global.NX_GRAPH_CREATION` is set, which is exactly when this plugin creates its nodes. So a `withNx` project always arrives with a `distDir` even when the user never wrote one, and treating any present `distDir` as custom would leave those projects pointing at `.next/`.

### Edge cases

| Config | Expected output dir | Before this PR | After this PR |
|--------|-------------------|----------------|---------------|
| `{}` (default) | `.next` | `.next` | `.next` |
| `{ distDir: 'build' }` | `build` | `build` | `build` |
| `{ output: 'export' }` | `out` | `.next` | `out` |
| `{ output: 'export', distDir: 'custom' }` | `custom` | `custom` | `custom` |
| `{ output: 'export', distDir: '.next' }` | `out` | `.next` | `out` |
| `withNx({ output: 'export' })` | `out` | `.next` | `out` |
| `{ output: 'standalone' }` | `.next` | `.next` | `.next` |

## Changes

- `getOutputs()` in `packages/next/src/plugins/plugin.ts` now extracts a unified `resolvedConfig` and branches on `output: 'export'` first, applying the same `distDir !== '.next'` condition Next.js uses.
- Added 7 test cases in `packages/next/src/plugins/plugin.spec.ts` covering the resolution: `output: 'export'` on its own, with a custom `distDir`, and with `distDir: '.next'` (the shape `withNx` produces); `distDir` without `output: 'export'`; sync and async config functions; and a root project. They assert on the resolved `outputs` array directly rather than snapshotting the whole node, so a failure points at the output dir instead of at unrelated target config.

## Related Issue(s)

N/A - discovered via broken deployment pipeline when using `output: 'export'` in Next.js projects within an Nx monorepo.

### DIFF
--- a/packages/next/src/plugins/plugin.spec.ts
+++ b/packages/next/src/plugins/plugin.spec.ts
@@ -48,6 +48,7 @@ describe('@nx/next/plugin', () => {
       expect(nodes).toMatchSnapshot();
     });
   });
+
   describe('integrated projects', () => {
     const tempFs = new TempFs('test');
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


asumaran:
@jaysoo please let me know if I should address anything else before review.

asumaran:
bump

AgentEnder:
Thanks for the PR, and welcome. Could you rebase this on `master`? It conflicts as it stands, and no Actions run exists for the head commit either, so a rebased push is what gets it both mergeable and built.

<details>
<summary>

</summary>

> Assignee is the `scope: nextjs` rotation pick, confirmed against git history. jaysoo has the deepest history in `packages/next`.
>
>
>
> No GitHub closing reference and no Linear key, so this is unlinked. Kept rather than closed, since `getOutputs` only reads `distDir`, so `output: 'export'` builds cache the wrong directory. A narrow bug fix with tests.
>
> The description carries an edge-case table for the precedence between `distDir` and `output: 'export'`, which is what makes the change checkable.

</details>

asumaran:
@AgentEnder done

jaysoo (approved):

jaysoo:
Thanks @asumaran !

PR #34536, opened 2026-02-20
fix(core) not finding workspace project from deep imports on enforce module boundaries

Ryntak94:
## Current Behavior
`@nx/enforce-module-boundaries` was not identifying the importing of an app if it was installed as a dependency in the `package.json` and was not listed in the `paths` of the tsconfig.

Also `targetProjectLocator.findImportInWorkspaceProjects` wasn't finding projects in the workspace if they weren't namespaced and were using subpaths.

## Expected Behavior
`targetProjectLocator.findImportInWorkspaceProjects` should find projects even if they aren't namespaced and are using subpaths.
`@nx/enforce-module-boundaries` should error on importing apps even if they are not namespaced, installed as project dependencies, and don't exist in the `paths` of the relevant tsconfig.

## Related Issue(s)
none (as far as I can tell)

Fixes #

### DIFF
--- a/packages/eslint-plugin/src/rules/enforce-module-boundaries.spec.ts
+++ b/packages/eslint-plugin/src/rules/enforce-module-boundaries.spec.ts
@@ -1515,15 +1515,13 @@ Violation detected in:
     expect(failures.length).toEqual(1);
   });
 
-  it('should error on importing an app', () => {
+  it('should error on importing an app (in tsconig paths)', () => {
     const failures = runRule(
       {},
       `${process.cwd()}/proj/libs/mylib/src/main.ts`,
       `
         import '@mycompany/myapp';
         import('@mycompany/myapp');
-        import('myapp');
-        import('myapp/subpath');
       `,
       {
         nodes: {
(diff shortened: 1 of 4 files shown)


Ryntak94:
@FrozenPandaz Hey, I just wanted to know if this was something we could get looked at. This behavior (I believe it's a bug) is affecting our organization and I'd like to know if I should look for another solution for us (at least for now) or if we might be able to get this merged in in a reasonable amount of time.

Thanks Jason!

Ryntak94:
@FrozenPandaz wanted to follow up on this

AgentEnder:
Thanks for the PR. Assigned to @leosvelperez now, and he will review as time allows.

<details>
<summary>

</summary>

> Assignee is a subsystem override on `scope: linter`. The owners table puts the enforce-module-boundaries rule with leosvelperez.
>
>
> No GitHub closing reference and no Linear key. The body says outright "Related Issue(s): none (as far as I can tell)". Kept rather than closed, since it is a targeted change to `targetProjectLocator.findImportInWorkspaceProjects` for non-namespaced projects imported via subpaths, not a speculative feature.
>
> It changes when the rule fires. A workspace with non-namespaced projects imported via subpaths passes lint today and would start failing, so that behaviour change is the call to make before the implementation matters.

</details>

PR #34544, opened 2026-02-22
fix(core): send minimal task graph over IPC to forked processes

Nitin75408:
## Current Behavior

The `TaskOrchestrator` sends the entire `TaskGraph` (140MB+ in large monorepos with 2000+ projects) via IPC to every forked child process. This causes 20+ second delays before tasks can start due to JSON serialization overhead, and this cost is paid per-worker (21+ parallel workers).

For batch executors, both `batchTaskGraph` (the relevant subset) and `fullTaskGraph` (the entire graph) are sent redundantly.

## Expected Behavior

For non-batch tasks, only a minimal task graph containing the executing task and its direct dependencies is sent over IPC. For batch tasks, the `batchTaskGraph` (already scoped to the batch) is used for the executor context, and the redundant `fullTaskGraph` is no longer transmitted.

This reduces IPC payload size from 140MB+ to a few KB per task, eliminating the serialization bottleneck.

## Related Issue(s)

Fixes #33995

### DIFF
--- a/packages/nx/src/tasks-runner/batch/batch-messages.ts
+++ b/packages/nx/src/tasks-runner/batch/batch-messages.ts
@@ -24,7 +24,6 @@ export interface RunTasksMessage {
   executorName: string;
   projectGraph: ProjectGraph;
   batchTaskGraph: TaskGraph;
-  fullTaskGraph: TaskGraph;
 }
 
 export interface CompleteTaskMessage {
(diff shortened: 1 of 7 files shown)


jaysoo (commented):
Good optimization — sending 140MB+ of task graph over IPC per task in large monorepos is clearly wasteful. Sending only the target task + direct dependencies is the right approach.

Two items to address before this can merge:

1. **Breaking change for batch executors**: Removing `fullTaskGraph` from batch executors would break custom batch executor implementations. This needs to be documented as a breaking change, or the full graph should be preserved for batch mode.

2. **Unrelated pnpm-lock.yaml changes**: Please remove these from the PR to keep the diff focused.

Note: PR is still in draft status. The core approach is sound — just clean these up when you're ready to mark it for review.

PR #34545, opened 2026-02-22
fix(core): prevent race condition between cache cleanup and remote ca…

### LINKED ISSUE #34032: maxCacheSize file outputs cleanup race condition with @nx/shared-fs-cache
### Current Behavior

In CI using the @nx/shared-fs-cache (can reproduce locally with CI=true env) when setting a max cache size, if running some command creating outputs exceeding the max cache size, cache entries will be deleted by cleanup operation, however the "shared" db will become out of sync causing NX to think it has some operations cached, leading it to use those "cached" operations, but no file outputs are restored because the folder containing the file outputs was deleted.

### Expected Behavior

When deleting cache entries, the shared DB should stay in sync with what has been deleted, so NX doesn't think it has something cached that it doesn't. 
I think it would also be very useful with a column in the cache_ouputs table that indicates whether we expect there to be some file output - in that case we could fail with an explicit error if the output is missing, rather than just silently failing as it is now.

### GitHub Repo

_No response_

### Steps to Reproduce

1. Install @nx/shared-fs-cache 
2. Set max cache size to a low value e.g. 10MB
3. Set env variable CI=true
4. Run a command that will cause the max cache size to be exceeded e.g. `nx run-many --targets=build --parallel=4` (seems to happen both with/without parallel)
5. Delete output of one of the build commands and run the build command for that package specifically. Nx will use the "remote cache", but will not restore any files as the folder containing the output files has been deleted because of the max cache size limitation, but NX thinks that it has it in the cache because there is an entry in the shared.db database found in the cache folder.

The local DB workspace-data stays in sync probably because it's handled right next to the deletion of the files in the cache.rs file, whereas there's probably some sort of async race condition going on with the remote cache?

### Nx Report

```shell

Node           : 22.10.0
OS             : darwin-arm64
pnpm           : 10.6.2

nx          : 22.1.3
@nx/devkit  : 22.1.3
typescript  : 5.9.2

Nx key licensed packages
Licensed to Lenus eHealth ApS.

@nx/shared-fs-cache  : 5.0.0

Community plugins:
@nx/shared-fs-cache : 5.0.0
Cache Usage: 67.54 MB / 100.00 MB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Nitin75408:
## Current Behavior

When using `@nx/shared-fs-cache` (or any remote cache) with a `maxCacheSize` setting, cache entries evicted by the local cleanup are not reflected in the remote cache's database. This leads to:

1. **Stale remote cache entries**: The remote DB reports cache hits for entries whose files no longer exist
2. **Silent failure on retrieval**: Nx reports a successful cache hit but no output files are restored (the workspace outputs are actually **deleted** during the failed restore)
3. **Lost build outputs**: Users must manually rebuild affected packages with no clear error message

## Expected Behavior

- When local cache cleanup evicts entries before they can be stored remotely, the remote store is skipped with a warning (preventing stale DB entries)
- When retrieving from remote cache, the outputs path is validated before applying results — missing or empty paths are treated as cache misses with an explicit warning
- The task orchestrator provides defense-in-depth validation for remote cache results with expected file outputs
- The `RemoteCacheV2` abstract class now includes an optional `remove()` method so remote cache implementations can stay in sync with local evictions

## Related Issue(s)

Fixes #34032

### DIFF
--- a/packages/nx/src/tasks-runner/cache-db.spec.ts
+++ b/packages/nx/src/tasks-runner/cache-db.spec.ts
@@ -1,0 +1,327 @@
+import {
+  mkdirSync,
+  mkdtempSync,
+  realpathSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+
+// Create a stable temp dir that can be used by module-level mocks
+const stableTmpDir = realpathSync(
+  mkdtempSync(join(tmpdir(), 'nx-cache-test-'))
+);
+
+// --- Mock setup (must come before imports of code under test) ---
+
+const mockNxCacheInstance = {
+  cacheDirectory: stableTmpDir,
+  get: jest.fn(),
+  put: jest.fn(),
+  applyRemoteCacheResults: jest.fn(),
+  getTaskOutputsPath: jest.fn(),
+  getCacheSize: jest.fn().mockReturnValue(0),
+  copyFilesFromCache: jest.fn().mockReturnValue(0),
+  removeOldCacheRecords: jest.fn(),
+  checkCacheFsInSync: jest.fn().mockReturnValue(true),
+};
+
+jest.mock('../native', () => ({
+  NxCache: jest.fn().mockImplementation(() => mockNxCacheInstance),
+  IS_WASM: false,
+  getDefaultMaxCacheSize: jest.fn().mockReturnValue(100 * 1024 * 1024),
+  HttpRemoteCache: jest.fn(),
+}));
+
+jest.mock('../config/nx-json', () => ({
+  readNxJson: jest.fn().mockReturnValue({}),
+  NxJsonConfiguration: {},
+}));
+
+jest.mock('../utils/db-connection', () => ({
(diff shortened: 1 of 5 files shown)


jaysoo (changes requested):
The race condition between cache cleanup (`maxCacheSize` eviction) and remote cache retrieval is a real issue. However, this PR has some structural problems:

1. **Dead code**: `remove()` is added to the `RemoteCacheV2` interface but never called anywhere. If it's intended for future use, it shouldn't be in this PR.

2. **Workaround, not fix**: The `existsSync` checks happen after `copyFilesFromCache` already ran, so the stale `shared.db` entries remain. The real fix belongs in the Rust `cache.rs` or `@nx/shared-fs-cache` layer where eviction should atomically update the database.

3. **Timing**: The checks detect the issue but can't prevent it — there's still a race window between checking existence and using the cache entry.

Please also remove the unrelated pnpm-lock.yaml changes.

PR #34551, opened 2026-02-23
feat(core): add --json flag for better AX to nx list

MaxKless:
### Current Behavior

  nx list <plugin> shows generator/executor names and descriptions in text format only. It does not show where the plugin or its generators/executors are located on disk, and there is no machine-readable output option.

  ### Expected Behavior

  - nx list --json outputs all local and installed plugins with their paths and capability types
  - nx list <plugin> --json outputs detailed structured JSON including resolved paths to each generator/executor implementation and schema
  - nx list <plugin> (text mode) now also shows the plugin's root path

### DIFF
--- a/e2e/nx/src/misc.test.ts
+++ b/e2e/nx/src/misc.test.ts
@@ -567,21 +567,23 @@ describe('Nx Commands', () => {
       const parsed = JSON.parse(jsonOutput);
 
       expect(parsed.name).toBe('@nx/js');
-      expect(parsed.path).toBeDefined();
-      expect(typeof parsed.generators).toBe('object');
-      expect(typeof parsed.executors).toBe('object');
+      expect(parsed.path).toContain('node_modules/@nx/js');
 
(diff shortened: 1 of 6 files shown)


e2e/nx/src/misc.test.ts
```diff
@@ -543,6 +543,46 @@ describe('Nx Commands', () => {
...
+    }, 120000);
+
+    it('should list plugin capabilities as JSON with --json flag', () => {
+      const jsonOutput = runCLI('list @nx/js --json');
+      const parsed = JSON.parse(jsonOutput);
+
+      expect(parsed.name).toBe('@nx/js');
+      expect(parsed.path).toBeDefined();
+      expect(typeof parsed.generators).toBe('object');
+      expect(typeof parsed.executors).toBe('object');
```
AgentEnder: Can we check actual values to some extent here

AgentEnder (approved):
Sm nit on the e2e

PR #34634, opened 2026-02-27
feat(core): add safe plugin cache write utilities with LRU eviction

FrozenPandaz:
## Current Behavior

Plugin cache writes (`writeJsonFile`, `writeFileSync`) across the codebase have inconsistent error handling:
- Some throw on failure, aborting project graph calculation entirely
- Some silently swallow errors, leaving corrupted cache files on disk
- `nx-deps-cache.ts` retries 5 times then throws, crashing the graph

## Expected Behavior

- Plugin cache write failures **never** abort graph calculation — they warn and continue
- A centralized `PluginCache<T>` class with explicit `get()`/`set()` API that tracks access order for LRU eviction
- Access tracking is append-only during the session (just array pushes — no O(n) scanning per access)
- Dedup and capping happen once at write time via `toSerializable(maxEntries)`
- `safeWritePluginCache` wraps writes in try/catch, warns + cleans up on failure (never throws)
- Backward compatible: reads old plain `Record<string, T>` cache format and migrates transparently
- All plugin cache consumers migrated: cypress, playwright, dotnet, gradle (v1+v2), maven, package-json, js/lockfile
- `nx-deps-cache.ts` changed from throw-after-retries to warn + cleanup
- New utilities exported via `@nx/devkit` internals for plugin authors

### Key design difference from #34503

This takes a simpler approach than #34503:
- **No Proxy** — explicit `get()`/`set()` methods instead of intercepting bracket access
- **No per-access O(n) indexOf** — access log is append-only, deduped once at write time
- **Simpler write strategy** — single try/catch + warn instead of 3-step evict-and-retry (the retry-after-eviction in #34503 assumes write failures are caused by cache size, but common causes are disk full, permissions, or corrupted files — evicting entries won't help)

## Related Issue(s)

Fixes NXC-3833

### DIFF
--- a/packages/cypress/src/plugins/plugin.ts
+++ b/packages/cypress/src/plugins/plugin.ts
@@ -8,9 +8,7 @@ import {
   normalizePath,
   type NxJsonConfiguration,
   type ProjectConfiguration,
-  readJsonFile,
   type TargetConfiguration,
-  writeJsonFile,
 } from '@nx/devkit';
 import { calculateHashForCreateNodes } from '@nx/devkit/src/utils/calculate-hash-for-create-nodes';
 import { loadConfigFile } from '@nx/devkit/src/utils/config-utils';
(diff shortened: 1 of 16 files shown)


AgentEnder (changes requested):
This requires knowing a concrete max entry count. It won't work for our case where the json serialization blows up, as the max count depends on the value sizes too, not just the num entries.

We should detect the serialization failure and go on that... like in the original pr... ill close this one back out. Theres some good ideas here that are net nice simplification (like the access deduce pass and move away from the proxy), but let's follow that into the original pr instead of completely scrapping it?

PR #34635, opened 2026-02-27
chore(repo): remove `is-ci` dependency

roli-lpci:
## Current Behavior

The repo uses the [`is-ci`](https://www.npmjs.com/package/is-ci) npm package (which wraps `ci-info`) in e2e test utilities, the issues-scraper script, and the root `prepare` script. Meanwhile, the core `packages/nx` already has its own native CI detection at `packages/nx/src/utils/is-ci.ts` — so `is-ci` is redundant.

## Expected Behavior

CI detection uses `!!process.env.CI` (consistent with the core package's approach) and the `is-ci` + `@types/is-ci` devDependencies are removed.

## Related Issue(s)

None — this is a dependency cleanup.

## Changes

- **e2e/utils/get-env-info.ts**: Replace `import * as isCI from 'is-ci'` with `export const isCI = !!process.env.CI`
- **e2e/utils/create-project-utils.ts**: Import `isCI` from `./get-env-info` instead of `is-ci`
- **e2e/utils/command-utils.ts**: Import `isCI` from `./get-env-info` instead of `is-ci`
- **e2e/utils/global-setup.ts**: Inline `const isCI = !!process.env.CI` (file doesn't statically import from get-env-info)
- **scripts/issues-scraper/index.ts**: Inline `const isCI = !!process.env.CI`
- **package.json**: Simplify `prepare` from `is-ci || husky` to `husky` (Husky v9 handles CI gracefully), remove `is-ci` and `@types/is-ci` from devDependencies


AgentEnder:
IsCI isnt used in any of our published deps and isnt hurting anything, no reason to make this change right now.

PR #34645, opened 2026-02-27
cleanup(linter): reuse ESLint instance across child projects in plugin

omer-za:
## Current Behavior

In `internalCreateNodesV2`, the ESLint plugin creates a **new `ESLint` instance per child project root** to check whether the project has non-ignored lintable files via `isPathIgnored`:

```ts
const eslint = new ESLint({
  cwd: join(context.workspaceRoot, projectRoot),
});
for (const file of lintableFilesPerProjectRoot.get(projectRoot) ?? []) {
  if (!(await eslint.isPathIgnored(...))) { ... }
}
```

In large monorepos, this means hundreds or thousands of ESLint instantiations during graph calculation. Each instantiation involves loading and resolving the ESLint configuration, which is the dominant cost in the plugin.

## Expected Behavior

A single ESLint instance should be shared across all child projects under the same ESLint config directory. Since `isPathIgnored` receives **absolute paths**, the instance's `cwd` does not affect the result — it only needs to resolve the correct config, which is the same for all children of a given config root.

## Changes

* In `internalCreateNodesV2`: create one lazily-initialized shared ESLint instance per config directory (the `configDir`), reused across all child project roots
* Projects that have their own `.eslintignore` file fall back to a per-project ESLint instance, preserving correct behavior for ESLint v8 (which resolves `.eslintignore` relative to `cwd`)

## Benchmark

Measured on a real monorepo with **1,609 projects** and a single root ESLint config.

| | Before | After | Change |
|---|---|---|---|
| Cold run (avg of 3) | **~18,700ms** | **~7,300ms** | **-61%** |
| ESLint instances created | 1,457 | 8 | **-99.5%** |

### Breakdown

| Version | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| Baseline (upstream) | 23,482ms | 16,238ms | 16,389ms |
| With shared instance | 7,473ms | 7,012ms | 7,515ms |

### Methodology

* `NX_DAEMON=false` to prevent daemon caching
* ESLint plugin hash cache cleared between each cold run
* `performance.now()` instrumentation around `internalCreateNodesV2`
* Each measurement repeated 3 times
* **Verified same number of lint targets** (1,473) before and after — no behavior change

### Why `configDir === '.'` is not excluded

The earlier revision excluded root-level ESLint configs (`configDir === '.'`) from the optimization as a conservative safety measure. However, benchmarking showed this **completely disables the optimization** for the most common monorepo setup (single root ESLint config), dropping 1,456 of 1,457 instances back to per-project instantiation with zero measurable improvement.

The per-project `.eslintignore` check already handles the ESLint v8 concern: projects with their own `.eslintignore` still get dedicated instances, while all others share one instance whose `cwd` correctly resolves the root config and root `.eslintignore`.

## Related

This is the `createNodesV2` (Nx plugin) code path. The change does not affect ESLint execution itself — only the graph calculation phase.

### DIFF
--- a/packages/eslint/src/plugins/plugin.ts
+++ b/packages/eslint/src/plugins/plugin.ts
@@ -80,6 +80,17 @@ const internalCreateNodesV2 = async (
   const configDir = dirname(configFilePath);
   const eslintVersion = ESLint.version;
 
+  let sharedEslint: ESLintType;
+  const getEslint = (projectRoot: string) => {
+    if (existsSync(join(context.workspaceRoot, projectRoot, '.eslintignore'))) {
+      return new ESLint({ cwd: join(context.workspaceRoot, projectRoot) });
+    }
+    sharedEslint ??= new ESLint({
+      cwd: join(context.workspaceRoot, configDir),
+    });
+    return sharedEslint;
+  };
+
   const projects: CreateNodesResult['projects'] = {};
   await Promise.all(
     projectRootsByEslintRoots.get(configDir).map(async (projectRoot) => {
@@ -93,9 +104,7 @@ const internalCreateNodesV2 = async (
 
       let hasNonIgnoredLintableFiles = false;
       if (configDir !== projectRoot || projectRoot === '.') {
-        const eslint = new ESLint({
-          cwd: join(context.workspaceRoot, projectRoot),
-        });
+        const eslint = getEslint(projectRoot);
         for (const file of lintableFilesPerProjectRoot.get(projectRoot) ?? []) {
           if (
             !(await eslint.isPathIgnored(join(context.workspaceRoot, file)))


packages/eslint/src/plugins/plugin.ts
```diff
@@ -80,6 +80,11 @@ const internalCreateNodesV2 = async (
   const configDir = dirname(configFilePath);
   const eslintVersion = ESLint.version;
 
+  const sharedEslint =
+    configDir !== '.'
+      ? new ESLint({ cwd: join(context.workspaceRoot, configDir) })
+      : undefined;
```
leosvelperez: Let's lazy-load the instantiation so we don't do it when all projects for the config dir are cached, and account for ESLint v8:
```suggestion
  let sharedEslint: ESLintType;
  const getEslint = (projectRoot: string) => {
    if (
      configDir === '.' ||
      existsSync(join(context.workspaceRoot, projectRoot, '.eslintignore'))
    ) {
      return new ESLint({ cwd: join(context.workspaceRoot, projectRoot) });
    }
    sharedEslint ??= new ESLint({
      cwd: join(context.workspaceRoot, configDir),
    });
    return sharedEslint;
  };
```

packages/eslint/src/plugins/plugin.ts
```diff
@@ -93,9 +98,9 @@ const internalCreateNodesV2 = async (
 
       let hasNonIgnoredLintableFiles = false;
       if (configDir !== projectRoot || projectRoot === '.') {
-        const eslint = new ESLint({
-          cwd: join(context.workspaceRoot, projectRoot),
-        });
+        const eslint =
+          sharedEslint ??
+          new ESLint({ cwd: join(context.workspaceRoot, projectRoot) });
```
leosvelperez:
```suggestion
        const eslint = getEslint(projectRoot);
```

leosvelperez (changes requested):
Thanks for the contribution!

This is on the right track, but Nx still officially supports ESLint v8, which does discover potential `.eslintignore` relative to the `cwd`. So, let's tweak your approach a bit to account for this.

leosvelperez:
@omer-za did you re-verify the perf and correctness after applying the suggested changes?

omer-za:
> @omer-za did you re-verify the perf and correctness after applying the suggested changes?

Verified, pushed a small fix and also updated the description

leosvelperez (approved):
Thank you!

PR #34682, opened 2026-03-03
fix(core): correct TUI sidebar viewport height off-by-one

comp615:
_This PR description was generated by Amp._

## Current Behavior

The TUI task sidebar appears scrolled down by one entry on render, hiding the first continuous task from view.

Screenshot 2026-03-03 at 10 56 09 AM

Here you can see that `watch-deps` is running in addition to `dev`.

## Expected Behavior

All continuous tasks should be visible in the sidebar without any initial scroll offset.

## Related Issue(s)

N/A (discovered via visual inspection)

## Details

The viewport height calculation in `tasks_list.rs` subtracted 4 rows for header overhead, but the actual overhead is only 3 rows:

1. `top_margin` (1 row)
2. Header content (1 row)
3. Spacing row (1 row)

This off-by-one caused the viewport to be 1 row too small, making the task list appear scrolled down by one entry on initial render.

### Changes

1. **Unified overhead constant**: Introduced `TABLE_HEADER_OVERHEAD_ROWS = 3` and `SCROLLBAR_Y_OFFSET = 2`, replacing scattered hardcoded values. The scrollbar y-offset is 2 (not 3) because the scrollbar visually spans from the spacing row downward, while the viewport only counts content rows.

2. **Stale scroll correction**: Added logic in `set_viewport_height` to reset `scroll_offset` to 0 when the viewport grows from a small placeholder size (≤ 5) to the real terminal size, provided the selected item is still visible from the top. This prevents stale scroll positions set during initialization from hiding top items.

3. **Regression tests**: Added `test_viewport_growth_resets_stale_scroll_offset` and `test_viewport_growth_preserves_scroll_when_selection_not_visible` to cover the stale scroll correction.

4. **Snapshot updates**: 14 snapshot files updated to reflect one additional visible row.

### History

The `4` originated in commit `6541751a` (James Henry, Apr 2025) with the comment "Reserve space for pagination and borders." However, the table had no borders (`Block::default()` without `Borders`), and pagination lived in its own layout chunk. The `4` was a rough estimate that was never precisely derived from the actual header structure. When scrolling replaced pagination in commit `1782e8c7` (Leosvel Perez Espinola, Sep 2025), the value was carried forward unchanged. The same commit also introduced a separate `header_and_spacing_rows = 2` for the scrollbar y-offset, confirming these values were set independently without unified accounting.

Also documents the `copy-built-package` script in `AGENTS.md` for AI-assisted development workflows.


FrozenPandaz (approved):

PR #34693, opened 2026-03-04
fix(core): add .claude/worktrees to gitignore

juristr:
## Current Behavior

When Claude Code creates worktrees under `.claude/worktrees/`, Nx picks them up as workspace projects. This causes duplicate project errors that break the project graph.

## Expected Behavior

`.claude/worktrees` is gitignored so worktree copies of the repo don't interfere with Nx's project detection.

**Changes:**
- Add `.claude/worktrees` to all three gitignore templates for new workspaces
- Add migration (22.6.1) for existing workspaces using `addEntryToGitIgnore`


MaxKless (approved):

packages/nx/migrations.json
```diff
@@ -137,6 +137,12 @@
       "description": "Updates the nx wrapper.",
       "implementation": "./src/migrations/update-22-1-0/update-nx-wrapper"
     },
+    "22-6-1-add-claude-worktrees-to-git-ignore": {
+      "cli": "nx",
+      "version": "22.6.1-beta.0",
```
FrozenPandaz: Should be `22.6.0-beta.10`

FrozenPandaz (approved):

Issue #32788, opened 2025-09-19
CI Hanging at `Creating project graph nodes with 7 plugins`


AaronLayton:
### Current Behavior

When running `npx nx release --yes --verbose` in our Gitlab pipelines, the job is hanging at 

`Creating project graph nodes with 7 plugins`

```bash
$ npx nx release --yes --verbose
Creating project graph nodes with 7 plugins
 NX   Failed to process project graph.
     - Error: @nx/next/plugin timed out after 10 minutes during createNodes. As a last resort, you can set NX_PLUGIN_NO_TIMEOUTS=true to bypass this timeout.
      at Timeout.<anonymous> (/builds/remarkablenet/packages/npm/rcm-platform/node_modules/nx/src/project-graph/plugins/isolation/plugin-pool.js:229:17)
      at listOnTimeout (node:internal/timers:588:17)
      at processTimers (node:internal/timers:523:7)
      Error: @nx/next/plugin timed out after 10 minutes during createNodes. As a last resort, you can set NX_PLUGIN_NO_TIMEOUTS=true to bypass this timeout.
          at Timeout.<anonymous> (/builds/remarkablenet/packages/npm/rcm-platform/node_modules/nx/src/project-graph/plugins/isolation/plugin-pool.js:229:17)
          at listOnTimeout (node:internal/timers:588:17)
          at processTimers (node:internal/timers:523:7)
Cleaning up project directory and file based variables
00:01
ERROR: Job failed: exit code 1
```

### Expected Behavior

A graph should be created or at least some more verbose logging to tell me what is happening. Releases should be processed and packages created.

### Steps to Reproduce

Our project was previously setup and running correctly, we added a new project and whilst we can get it to --dry-run locally (after restarting the daemon a few times) but it just hangs now in the CI

### Nx Report

```shell
This is local machine, not the CI 

Node           : 22.19.0
OS             : win32-x64
Native Target  : x86_64-windows
npm            : 10.9.3

nx                     : 21.3.11
@nx/js                 : 21.3.11
@nx/eslint             : 21.3.11
@nx/workspace          : 21.3.11
@nx/devkit             : 21.3.11
@nx/esbuild            : 21.3.8
@nx/eslint-plugin      : 21.3.11
@nx/module-federation  : 21.3.8
@nx/next               : 21.3.11
@nx/react              : 21.3.11
@nx/vite               : 21.3.11
@nx/web                : 21.3.11
@nx/webpack            : 21.3.11
[...]

Issue #31835, opened 2025-07-04
@nx/gradle is extremely slow when compared to running gradle directly


refactorthis:
### Current Behavior

I have found that adding only a single gradle build task for each project to Nx is much faster than using `@nx/gradle` and building all subtasks in Nx itself.

Looking at @nx/quarkus, @nx/spring, etc.. they have chosen to go with this approach - use Nx as the entry point to the project's build, but let gradle do the building.

The downside to these libraries is that they don't understand the build dependencies, so you need to create implicit dependencies for each project. This isn't scalable across the organisation for me.

What is the recommended practice here?

### Expected Behavior

As above.

### Steps to Reproduce

1. Multiple isolated projects in a monorepo with many gradle tasks.

### Nx Report

```shell
I am not at my machine, will provide more detail if required.
```

### Failure Logs

```shell

```

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

Lutonite:
I was facing the same issue, while the time to run is bearable locally thanks to the Gradle Daemon, on CI it would take ages because it had to fork the JVM for the Daemon for each task execution.

I discovered that Nx has a `--batch` argument and that an implementation for it was added in `@nx/gradle` (though I think the documentation on this could be improved a bit), adding this in our CI has allowed all Gradle-related tasks to run at once and it has greatly improved run times, though that seems to hide all possibilities of having detailed logs about what fails when it fails 🤔

MaxKless:
Hey @refactorthis, thanks for the report! 
As @Lutonite mentioned, we recommend using the `--batch` argument in this case as it groups multiple tasks to be run together. There are some issues with the TUI & `--batch` still, but we're working on them as I type this out :) 

I'll close this for now but let me know if your issues persist even in batch mode

Issue #27494, opened 2024-08-17
Nx operations hangs indefinitely


JbIPS:
### Current Behavior

I'm trying to add Nx to my TS monorepo.

On first try, I installed it (pnpm i nx) but any run would hang on indefinitely. So I removed it, tried to reinstall it but now it hangs at postinstall
```
node_modules/.pnpm/nx@19.5.0_@swc+core@1.7.10/node_modules/nx: Running postinstall script
```
I tried other approach like pnpm dlx nx@latest init or even npx nx@latest init but it hangs at
```
 NX   🧐 Checking dependencies 
```
Also tried pnpm dlx create-nx-workspace from scratch but it hangs on the second step
```
Creating your workspace in test-nx
```
### Expected Behavior

Inits operations should complete gracefully

### Steps to Reproduce

1. `pnpm dlx create-nx-workspace`

### Nx Report

```shell
❯ pnpm dlx nx report                      
Packages: +115
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 123, reused 111, downloaded 4, added 115, done
.cache/pnpm/dlx/ogenweua4lihpne4pbq6bewrt4/1915f46eda6-1d714/node_modules/.pnpm/nx@19.6.0/node_modules/nx: Running postinstall script, done in 373ms

(...and nothing for at least 20min)
```

### Package Manager Version

PNPM v9.7.0 or NPM v10.5.0

### Operating System

- [ ] macOS
- [X] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

A common logs I see with `NODE_DEBUG=*` on all the nx command is
```
MODULE 121282: RELATIVE: requested: ./error-types from parent.id /home/******/node_modules/.pnpm/nx@19.5.7_@swc+core@1.7.11/nod
e_modules/nx/src/project-graph/build-project-graph.js
MODULE 121282: looking for ["/home/*******/node_modules/.pnpm/nx@19.5.7_@swc+core@1.7.11/node_modules/nx/src/project-graph"]
NET 121243: pipe true /tmp/plugin121243-0.sock
NET 121243: connect: attempting to connect to /tmp/plugin121243-0.sock:NaN (addressType: NaN)
MODULE 121282: Module._load REQUEST ./utils/project-configuration-utils parent: /home/*********/node_modules/.pnpm/nx@19.5.7_@swc+
[...]

PR #34705, opened 2026-03-04
fix(gradle): destroy streams on exit to prevent daemon pipe hang

nimo71:
## Current Behavior

`execGradleAsync()` in `exec-gradle.ts` can hang indefinitely during project graph creation. This manifests as Nx freezing on "Creating project graph nodes" — particularly in CI environments, but also reproducible locally.

## Expected Behavior

`execGradleAsync()` should resolve/reject promptly after the Gradle task completes, regardless of whether the Gradle daemon continues running in the background.

## Root Cause

`execGradleAsync()` uses `execFile()` with `shell: true`, which spawns a shell wrapper process. The Gradle daemon, started by `gradlew`, inherits the shell's stdout/stderr pipe file descriptors. After the `nxProjectGraph` task completes:

1. The `gradlew` wrapper script exits
2. The Gradle daemon continues running, holding the inherited pipe FDs open
3. The shell wrapper process cannot exit because its child (the daemon) still holds its pipes
4. Node.js never fires the `exit` event on the child process
5. The Promise in `execGradleAsync()` never resolves

This is a well-documented interaction between:
- [nodejs/node#5637](https://github.com/nodejs/node/issues/5637) — `execFile` hangs when child exits but disowned children hold pipes
- [gradle/gradle#3987](https://github.com/gradle/gradle/issues/3987) — Gradle daemon inherits parent file descriptors

## Fix

Destroy stdout/stderr streams immediately after the `exit` event fires. By this point, all output data has already been buffered into the `stdout` variable via the `data` event handlers, so no output is lost. Destroying the streams releases the pipe FDs from Node's perspective, allowing the shell wrapper to exit and the Promise to resolve — regardless of daemon state.

```typescript
cp.on('exit', (code, signal) => {
    if (code === null) code = signalToCode(signal);
    // Forcibly destroy streams to prevent Gradle daemon pipe hang.
    cp.stdout?.destroy();
    cp.stderr?.destroy();
    if (code === 0) {
        res(stdout);
    } else {
        rej(stdout);
    }
});
```

## Workarounds Users Currently Need

Without this fix, users must resort to:
- `org.gradle.daemon.idletimeout=5000` in `gradle.properties` (adds 5s latency per invocation)
- CI retry loops with `timeout` + `pkill -f GradleDaemon` (unreliable, wastes CI minutes)
- `--no-daemon` flag (doesn't work reliably with Gradle 8.1+/9.x — a "single-use daemon" is still forked when `org.gradle.jvmargs` is set)

## Related Issues

- #29618 — NX project graph calculation hanging indefinitely
- #32788 — CI hanging at "Creating project graph nodes"
- #31835 — @nx/gradle extremely slow compared to running Gradle directly
- #27494 — Nx operations hang indefinitely

### DIFF
--- a/packages/gradle/src/utils/exec-gradle.ts
+++ b/packages/gradle/src/utils/exec-gradle.ts
@@ -63,6 +63,17 @@ export function execGradleAsync(
 
     cp.on('exit', (code, signal) => {
       if (code === null) code = signalToCode(signal);
+      // Forcibly destroy streams to prevent Gradle daemon pipe hang.
+      // When shell: true is used, the Gradle daemon inherits the shell's
+      // stdout/stderr pipes and holds them open after the task completes.
+      // This prevents the shell wrapper from exiting, which in turn
+      // prevents the 'exit' event from firing on the ChildProcess.
+      // By this point all output has already been buffered, so destroying
+      // the streams is safe and releases the pipe FDs from Node's
+      // perspective.
+      // See: nodejs/node#5637, gradle/gradle#3987
+      cp.stdout?.destroy();
+      cp.stderr?.destroy();
       if (code === 0) {
         res(stdout);
       } else {


AgentEnder:
Thanks for the fix, and welcome. Could you rebase this on `master`? It conflicts as it stands, and no Actions run exists for the head commit either, so a rebased push is what gets it both mergeable and built.

<details>
<summary>

</summary>

> Assignee comes from the `scope: java` guidance rather than the rotation draw, since lourw owns Gradle day to day.
>
>
>
> No GitHub closing reference and no Linear key, so this is unlinked. Kept rather than closed, since it is +11/-0 against a hang, with the mechanism written out in the body.
>
> The body writes the mechanism out: the daemon inherits the shell wrapper's stdout and stderr pipe file descriptors and holds them open after `gradlew` exits.

</details>

PR #34720, opened 2026-03-05
fix(vitest)!: resolve reportsDirectory against workspace root

llwt:
## Current Behavior

When `reportsDirectory` is configured with `{workspaceRoot}` token in `nx.json` targetDefaults:

```json
"@nx/vitest:test": {
  "options": {
    "reportsDirectory": "{workspaceRoot}/coverage/{projectRoot}"
  }
}
```

Coverage output lands in the wrong location. For example, with a project at `apps/my-app`, coverage goes to `apps/my-app/coverage/apps/my-app/` instead of the intended `coverage/apps/my-app/`.

## Expected Behavior

Coverage output should be written to `<workspaceRoot>/coverage/apps/my-app/`.

## Root Cause

Nx's `resolveNxTokensInOptions` strips `{workspaceRoot}/` from option values and replaces `{projectRoot}`, producing a workspace-root-relative path (e.g. `coverage/apps/my-app`). The vitest executor then passed this directly to vitest, which resolved it relative to the **project root** — not the workspace root.

## Fix

Resolve non-absolute `reportsDirectory` paths against the workspace root before passing them to vitest, so vitest writes coverage to the correct location.

## Test Plan

- Added unit tests for the new `resolveReportsDirectory` helper
- Verified with a reproduction workspace that coverage now lands at the correct path


packages/vitest/migrations.json
```diff
@@ -17,6 +17,11 @@
       },
       "description": "Create AI Instructions to help migrate users workspaces past breaking changes for Vitest 4.",
       "implementation": "./src/migrations/update-22-1-0/create-ai-instructions-for-vitest-4"
+    },
+    "update-22-6-0-prefix-reports-directory": {
```
jaysoo, line 21: Not a huge deal, but targeting the next beta (22.6.0-beta.10) would help anyone who already updated to beta.0. Mostly it's just us.

jaysoo (approved):

packages/vitest/migrations.json
llwt: Done!

packages/vitest/migrations.json
```diff
@@ -17,6 +17,11 @@
...
+      "version": "22.6.0-beta.10",
```
llwt:
```suggestion
      "version": "22.6.0-beta.11",
```

PR #34722, opened 2026-03-05
feat(core): add polygraph command to initiaze cross-repo sessions

vsavkin:
## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/nx-cloud/utilities/axios.ts
+++ b/packages/nx/src/nx-cloud/utilities/axios.ts
@@ -1,27 +1,16 @@
 import type { AxiosRequestConfig } from 'axios';
 import { join } from 'path';
-import {
-  ACCESS_TOKEN,
-  NX_CLOUD_NO_TIMEOUTS,
-  UNLIMITED_TIMEOUT,
-} from './environment';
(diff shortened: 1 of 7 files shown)


packages/nx/src/nx-cloud/utilities/axios.ts
```diff
@@ -32,7 +21,6 @@ export function createApiAxiosInstance(options: CloudTaskRunnerOptions) {
     axiosConfigBuilder({
       baseURL: baseUrl,
       timeout: NX_CLOUD_NO_TIMEOUTS ? UNLIMITED_TIMEOUT : 10000,
-      headers: { authorization: accessToken },
```
AgentEnder, line 35: Will this cause issues for people still using an access token? My gut feeling is that the axios client that gets built would not be authenticated?

If this axios client is only used for downloading the light client and the light client uses a separate client thats auth'd though this may be fine?

vsavkin: it should not. The only place where we call this is from updateManager and it doesn't care for the token.

AgentEnder (approved):

PR #34735, opened 2026-03-06
fix(gradle): exclude non-JS gradle sub-projects from eslint plugin

leosvelperez:
## Current Behavior

The `@nx/eslint/plugin` infers a `lint` target for `gradle-project-graph` because it contains a single `.ts` file (`publish-maven.ts`), even though it's a Kotlin/Gradle project. Similarly, the parent `gradle` project's `eslint .` scans into non-JS sub-project directories unnecessarily.

## Expected Behavior

Non-JS Gradle sub-projects (`project-graph`, `batch-runner`) should not have lint targets inferred by the ESLint plugin, and the parent `gradle` project's lint should not scan into those directories.

## Changes

- Add `project-graph` and `batch-runner` to ESLint `ignorePatterns` in `packages/gradle/.eslintrc.json`

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -245,7 +245,11 @@
     },
     {
       "plugin": "@nx/eslint/plugin",
-      "exclude": ["packages/**/__fixtures__/**/*"],
+      "exclude": [
+        "packages/**/__fixtures__/**/*",
+        "packages/gradle/project-graph/**/*",
+        "packages/gradle/batch-runner/**/*"
+      ],
       "options": {
         "targetName": "lint"
       }
--- a/nx.json
+++ b/nx.json
@@ -245,11 +245,7 @@
     },
     {
       "plugin": "@nx/eslint/plugin",
-      "exclude": [
-        "packages/**/__fixtures__/**/*",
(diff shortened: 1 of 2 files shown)


nx.json
```diff
@@ -245,7 +245,11 @@
     },
     {
       "plugin": "@nx/eslint/plugin",
-      "exclude": ["packages/**/__fixtures__/**/*"],
+      "exclude": [
+        "packages/**/__fixtures__/**/*",
+        "packages/gradle/project-graph/**/*",
```
FrozenPandaz: We do not need these exclusions because we've eslintignored the typescript files in there right? So there's no target even without these exclusions.. seems to work on my machine.

FrozenPandaz (approved):

nx.json
leosvelperez: It's not "needed", but the goal was to save some processing. This eliminates some project roots from processing (they are still matched by the plugin glob and discarded when the plugin checks they don't have non-ignored lintable files), but it shouldn't make too big a difference, so we can simplify the config. I removed them.

PR #34736, opened 2026-03-06
fix(core): improve nx wrapper error message for malformed nx.json

MaxKless:
## Current Behavior

When `nx.json` exists but contains a syntax error (e.g. invalid JSON), the nx wrapper shows:

```
[NX]: The "nx.json" file is required when running the nx wrapper.
```

This is misleading because the file exists — it's just malformed.

## Expected Behavior

The nx wrapper now distinguishes between a missing `nx.json` and a malformed one:

- **Missing file**: `[NX]: The "nx.json" file is required when running the nx wrapper.`
- **Parse error**: `[NX]: Failed to parse "nx.json": <actual error>. See ...`

The existence check is done upfront before attempting to parse, so the `catch` block only handles parse errors.

### DIFF
--- a/packages/nx/src/command-line/init/implementation/dot-nx/nxw.ts
+++ b/packages/nx/src/command-line/init/implementation/dot-nx/nxw.ts
@@ -106,6 +106,13 @@ function ensureUpToDateInstallation() {
   const nxJsonPath = path.join(__dirname, '..', 'nx.json');
   let nxJson: NxJsonConfiguration;
 
+  if (!fs.existsSync(nxJsonPath)) {
+    console.error(
+      '[NX]: The "nx.json" file is required when running the nx wrapper. See https://nx.dev/recipes/installation/install-non-javascript'
+    );
+    process.exit(1);
+  }
+
   try {
     nxJson = require(nxJsonPath);
     if (!nxJson.installation) {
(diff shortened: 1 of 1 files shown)


packages/nx/src/command-line/init/implementation/dot-nx/nxw.ts
```diff
@@ -114,9 +121,9 @@ function ensureUpToDateInstallation() {
       );
       process.exit(1);
     }
-  } catch {
+  } catch (e: unknown) {
     console.error(
-      '[NX]: The "nx.json" file is required when running the nx wrapper. See https://nx.dev/recipes/installation/install-non-javascript'
+      `[NX]: Failed to parse "nx.json": ${e instanceof Error ? e.message : e}. See https://nx.dev/recipes/installation/install-non-javascript`
```
FrozenPandaz: Check the type of the error to switch between the 2 errors.

FrozenPandaz (changes requested):

packages/nx/src/command-line/init/implementation/dot-nx/nxw.ts
MaxKless: done

FrozenPandaz (approved):

PR #34740, opened 2026-03-06
fix(testing): infer task inputs from jest config file references

leosvelperez:
## Current Behavior

The Jest plugin infers `test` target inputs from the preset file path but ignores other config properties that reference files — transforms, setup files, module name mappers, reporters, watch plugins, etc. Changes to those files don't invalidate the test cache.

## Expected Behavior

All Jest config properties that reference files are resolved and included as task inputs, matching Jest's merge semantics:

- **Replaced** (config wins over preset): `resolver`, `globalSetup`, `globalTeardown`, `snapshotResolver`, `snapshotSerializers`, `testResultsProcessor`, `runner`, `reporters`, `watchPlugins`
- **Concatenated** (preset + config): `setupFiles`, `setupFilesAfterEnv`
- **Deep merged** (config keys win): `moduleNameMapper`, `transform`

Also handles `jest-runner-`/`jest-watch-` prefix resolution, `<rootDir>` in preset values, and Windows path normalization.

Adds a `useJestResolver` option to control whether jest-resolve is used for input resolution, decoupled from `disableJestRuntime`. By default, inputs are resolved using path-based classification (fast, no filesystem calls beyond what's already done). When `useJestResolver` is enabled, jest-resolve is used instead, which follows symlinks and honors custom `moduleDirectories`/`modulePaths` — more accurate for workspace-linked packages but slower due to filesystem probing per resolved path.

### DIFF
--- a/astro-docs/src/content/docs/technologies/test-tools/jest/introduction.mdoc
+++ b/astro-docs/src/content/docs/technologies/test-tools/jest/introduction.mdoc
@@ -165,12 +165,13 @@ Configure the plugin in `nx.json`:
 }
 ```
 
-| Option               | Type    | Default | Description                                                                                                                    |
-| -------------------- | ------- | ------- | ------------------------------------------------------------------------------------------------------------------------------ |
-| `targetName`         | string  | `test`  | Name of the inferred Jest target.                                                                                              |
-| `ciTargetName`       | string  | none    | Creates a CI-only target for atomized test tasks.                                                                              |
-| `ciGroupName`        | string  | none    | Custom group name for atomized tasks in Nx Cloud/UI.                                                                           |
-| `disableJestRuntime` | boolean | `true`  | When `true`, Nx skips creating the Jest runtime and computes inputs/outputs itself. Set to `false` to enable the Jest runtime. |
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -122,7 +122,14 @@
     },
     "test": {
       "dependsOn": ["test-native", "build-native", "^build-native"],
-      "inputs": ["default", "^production", "{workspaceRoot}/jest.preset.js"],
+      "inputs": [
```
FrozenPandaz: I imagined we would do this in the jest plugin.

If a resolver is specified.. the resolver can be added as an input?

I guess it goes deeper though? if the resolver imports another file... we would not have it added? 😕 

I mean, we also wouldn't have it added with this change...

Same with the preset.. global setup, and global teardown... setupFiles.. and module name mapper? 😵‍💫 

Is this feasible?

FrozenPandaz (changes requested):

nx.json
leosvelperez: Fair. Let me explore and see what we can do without a major perf hit.

FrozenPandaz (approved):

PR #34741, opened 2026-03-06
fix(testing): handle undefined options in playwright e2e preset

yjaaidi:
Calling Playwright's `nxE2EPreset(__filename)`(without options) fails.

P.S.: I just stumbled upon https://github.com/nrwl/nx/pull/34302 & https://github.com/nrwl/nx/pull/34301
I guess it's better if we let Playwright handle the default value for `openHtmlReport`


leosvelperez:
Closing as fixed by https://github.com/nrwl/nx/pull/34750.

PR #34775, opened 2026-03-10
chore(core): replace handleImport with native import()

FrozenPandaz:
## Current Behavior

`handleImport()` is used across ~65 files in `packages/nx/src/` as a CJS→ESM bridge. It tries `require()` first and falls back to `import()`. All call sites are in async functions, making the `require()` attempt unnecessary.

## Expected Behavior

Use native `await import()` directly, removing the unnecessary indirection. This simplifies the codebase by:
- Removing 67 files worth of `handleImport` wrapper calls
- Deleting the `handle-import.ts` utility and its tests
- Removing the `handleImport` export from `devkit-internals.ts`

## Related Issue(s)

Follow-up cleanup for PR #34111

### DIFF
--- a/packages/nx/src/adapter/ngcli-adapter.ts
+++ b/packages/nx/src/adapter/ngcli-adapter.ts
@@ -75,6 +75,7 @@ import {
   resolveSchema,
 } from '../config/schema-utils';
 import { resolveNxTokensInOptions } from '../project-graph/utils/project-configuration-utils';
+import { handleImport } from '../utils/handle-import';
 
 export async function createBuilderContext(
   builderInfo: {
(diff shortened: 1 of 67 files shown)


FrozenPandaz:
## Update: Reverted

After investigating CI failures, I discovered that `handleImport` **cannot** be replaced with plain `import()`.

**Root cause**: When CJS code calls `import()` on a CJS module, Node.js wraps `module.exports` under a `default` property. This breaks all destructuring patterns like:

```js
const { AppLifeCycle } = await import('../native/index.js')
// Returns undefined — actual value is at .default.AppLifeCycle
```

`handleImport` uses `require()` first, which returns `module.exports` directly. It only falls back to `import()` for ESM-only packages (catching `ERR_REQUIRE_ESM`). This is necessary as long as `nx` ships as `"type": "commonjs"`.

Pushed a revert commit — this PR now has no net changes. Will close it.

PR #34772, opened 2026-03-10
fix(nx-dev): adding missing legacy route redirects

barbados-clemens:
closes: DOC-439
closes: DOC-442

PR #34787, opened 2026-03-11
fix(nx-dev): update missing redirects in framer proxy

barbados-clemens:


barbados-clemens:
fixed here instead: https://github.com/nrwl/nx/pull/34772

PR #34789, opened 2026-03-11
fix(core): skip analytics prompt for cloud commands

MaxKless:
## Current Behavior

The analytics prompt (`ensureAnalyticsPreferenceSet()`) and `startAnalytics()` run for all commands, including cloud commands like `download-cloud-client`. When a cloud command runs outside an Nx workspace, `saveAnalyticsPreference()` creates an almost-empty `nx.json` (`{ "analytics": true }`) in whatever directory you happen to be in.

## Expected Behavior

Cloud commands skip the analytics prompt and `startAnalytics()` entirely, since they may run without a workspace and there is no appropriate `nx.json` to write to.

## Related Issue(s)

Fixes the issue where running `nx download-cloud-client` outside a workspace creates a spurious `nx.json`.

### DIFF
--- a/packages/nx/bin/nx.ts
+++ b/packages/nx/bin/nx.ts
@@ -104,27 +104,30 @@ async function main() {
       handleMissingLocalInstallation(workspace ? workspace.dir : null);
     }
 
-    // Prompt for analytics preference if not set
-    try {
-      await ensureAnalyticsPreferenceSet();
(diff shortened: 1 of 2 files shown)


Coly010 (dismissed):
Msged internally

FrozenPandaz (approved):

PR #34821, opened 2026-03-13
feat(core): add task and project count telemetry via performance lifecycle

FrozenPandaz:
## Current Behavior

Telemetry only reports event duration via the `measureAndTrack` helper, which uses a `[track] ` prefix convention. No task-level metrics (count, cache hits, project count) are reported.

## Expected Behavior

- New event dimensions: `taskCount`, `projectCount`, `cachedTaskCount` available for telemetry events
- `TaskTelemetryLifeCycle` reports task execution metrics (duration, task count, project count, cached task count) — only runs on the main CLI process, not on DTE agents
- `performance.measure()` with `detail: { track: true, ... }` replaces the `measureAndTrack` / `[track] ` prefix pattern
- Perf observer automatically forwards detail entries matching known GA4 dimension keys
- `reportEvent` simplified to a pass-through — callers use `customDimensions` keys directly
- `customDimensions` and `EventParameters` exported for use by callers

### DIFF
--- a/packages/nx/src/utils/perf-logging.ts
+++ b/packages/nx/src/utils/perf-logging.ts
@@ -1,21 +1,13 @@
 import { PerformanceObserver } from 'perf_hooks';
 
-import { reportPerfEvent } from '../analytics';
+import { customDimensions, reportEvent } from '../analytics';
+import type { EventParameters } from '../analytics';
 import { isOnDaemon } from '../daemon/is-on-daemon';
 import { serverLogger } from '../daemon/logger';
 
-const TRACK_PREFIX = '[track] ';
-
-/**
- * Like `performance.measure()` but automatically reports to telemetry.
- */
-export function measureAndTrack(
-  name: string,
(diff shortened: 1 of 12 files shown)


packages/nx/src/utils/perf-logging.ts
```diff
@@ -24,22 +16,29 @@ if (!initialized) {
 
   const obs = new PerformanceObserver((list) => {
     for (const entry of list.getEntries()) {
-      const shouldTrack = entry.name.startsWith(TRACK_PREFIX);
-      const displayName = shouldTrack
-        ? entry.name.slice(TRACK_PREFIX.length)
-        : entry.name;
+      const detail = (entry as any).detail;
```
AgentEnder: Is the `as any` here needed? I feel like we should be able to narrow it based on the entry type?

AgentEnder (approved):

AgentEnder (approved):

PR #34823, opened 2026-03-13
fix(core): prevent phantom connections and dead polling in plugin workers

### LINKED ISSUE #34388: The plugin worker for \node_modules\nx\src\plugins\js is exiting as it was not connected to within 5 seconds.
### Current Behavior

The plugin worker for *\node_modules\nx\src\plugins\js is exiting as it was not connected to within 5 seconds. Plugin workers expect to receive a socket connection from the parent process shortly after being started. If you are seeing this issue, please report it to the Nx team at https://github.com/nrwl/nx/issues.

### Expected Behavior

Build and start the angular project 

### GitHub Repo

_No response_

### Steps to Reproduce

1.

### Nx Report

```shell
The plugin worker for *\node_modules\nx\src\plugins\project-json\build-nodes\project-json is exiting as it was not connected to within 5 seconds. Plugin workers expect to receive a socket connection from the parent process shortly after being started. If you are seeing this issue, please report it to the Nx team at https://github.com/nrwl/nx/issues.
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

AgentEnder:
The setInterval(async, 10) polling loop used to connect to plugin workers created overlapping connection attempts because setInterval does not await its callback. This caused phantom socket connections, event loop saturation, and cascading worker deaths.

Changes:
- Replace setInterval with recursive setTimeout so only one connection attempt is ever in flight at a time
- Close the worker's server on first connection to reject phantom connections that would create duplicate load timeouts
- Detect worker exit during polling and reject immediately instead of burning through 10,000 attempts against a dead socket
- Clear _connectPromise on failure so ensureAlive() retries instead of re-awaiting a permanently rejected promise

Fixes: #34388

### DIFF
--- a/packages/nx/src/project-graph/plugins/isolation/isolated-plugin.ts
+++ b/packages/nx/src/project-graph/plugins/isolation/isolated-plugin.ts
@@ -196,7 +196,12 @@ export class IsolatedPlugin implements LoadedNxPlugin {
 
     if (!this._connectPromise) {
       logger.verbose(`[plugin-client] restarting worker for "${this.name}"`);
-      this._connectPromise = this.spawnAndConnect();
+      this._connectPromise = this.spawnAndConnect().catch((err) => {
+        // Clear the cached promise so subsequent calls can retry
+        // instead of re-awaiting a permanently-rejected promise.
+        this._connectPromise = null;
+        throw err;
+      });
     }
 
     await this._connectPromise;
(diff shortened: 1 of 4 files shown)


packages/nx/src/project-graph/plugins/isolation/plugin-worker.ts
```diff
@@ -37,14 +37,22 @@ const socketPath = process.argv[2];
 const expectedPluginName = process.argv[3];
 
 let connectErrorTimeout = setErrorTimeout(
-  5000,
+  30_000,
   `The plugin worker for ${expectedPluginName} is exiting as it was not connected to within 5 seconds. ` +
```
FrozenPandaz: this should also say 30 seconds... probably make it a variable

packages/nx/src/project-graph/plugins/isolation/plugin-worker.ts
```diff
@@ -37,14 +37,22 @@ const socketPath = process.argv[2];
...
 
 const server = createServer((socket) => {
   connectErrorTimeout?.clear();
+
+  // Stop accepting new connections immediately. There should only
+  // ever be one host → worker connection since the worker is spawned
+  // per host process. Closing the server here prevents phantom
+  // connections (e.g., from overlapping connection attempts) from
+  // creating duplicate load timeouts that would kill the worker.
+  server.close();
```
FrozenPandaz: this would cause the isoalted plugin to try 9999 more times * 10 ms getting an error for server is closed

packages/nx/src/project-graph/plugins/isolation/isolated-plugin.ts
```diff
@@ -564,22 +569,44 @@ async function startPluginWorker(name: string) {
...
+            new Error(
+              `Plugin worker for "${name}" exited with code ${code} before the connection was established.`
+            )
+          );
+        }
+      });
+
+      const poll = async () => {
+        if (stopped) return;
         const socket = await isServerAvailable(ipcPath);
```
FrozenPandaz: check out the daemon is server available logic and move towards a while async loop

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #34837, opened 2026-03-14
feat(devkit): expose FsTree, flushChanges, and printChanges as public API

Fasosnql:
Adds FsTree class, flushChanges, printChanges, and TreeWriteOptions to the  @nx/devkit public API so generators can be invoked programmatically without  exec/spawn. Enables LLM-based and agentic workflows where changes must be  inspectable before being written to disk.

BREAKING CHANGE: No

closed #34836

## Current Behavior
`FsTree`, `flushChanges`, and `printChanges` are internal to the `nx` package and not accessible via the `@nx/devkit` public API. The only public tree-related exports from `@nx/devkit` today are the `Tree` and `FileChange` interfaces (both type-only).

This forces consumers who want to run generators programmatically — without shelling out to `nx generate` — to either:
1. Use `exec`/`spawn`, which immediately writes all changes to disk with no opportunity for in-memory inspection or validation
2. Import from the unsupported internal path `nx/src/generators/tree`, which is not a public API and subject to breaking changes

## Expected Behavior
`FsTree`, `flushChanges`, `printChanges`, and `TreeWriteOptions` are exported from `@nx/devkit`, enabling the following programmatic workflow:

```typescript
import { FsTree, flushChanges, printChanges } from '@nx/devkit';
import { libraryGenerator } from '@nx/js';

// Create a tree rooted at the real workspace
const tree = new FsTree(process.cwd(), true);

// Run a generator function directly — nothing touches disk yet
await libraryGenerator(tree, { name: 'my-lib', bundler: 'tsc' });

// Optionally add or modify files in the virtual tree
tree.write('libs/my-lib/src/custom.ts', myContent);

// Inspect changes before writing
const changes = tree.listChanges();
printChanges(changes);

// Validate, then flush to disk — or discard entirely
if (valid) {
  flushChanges(tree.root, changes);
}
```

This is particularly valuable for LLM-based and agentic code generation workflows, where changes must be inspectable and validatable before being committed to disk.

## Related Issue(s)

Fixes #34836

### DIFF
--- a/packages/nx/src/devkit-exports.ts
+++ b/packages/nx/src/devkit-exports.ts
@@ -6,7 +6,8 @@
 /**
  * @category Tree
  */
-export type { Tree, FileChange } from './generators/tree';
+export type { Tree, FileChange, TreeWriteOptions } from './generators/tree';
+export { FsTree, flushChanges, printChanges } from './generators/tree';
 
 /**
  * @category Workspace


FrozenPandaz:
Let's design this properly in https://github.com/nrwl/nx/discussions/34836 before we commit to anything.

PR #34842, opened 2026-03-15
fix(maven): filter out external-to-external dependencies in createDependencies

llwt:
## Current Behavior

When `nx-maven-plugin` v0.0.15 emits external-to-external Maven dependency edges (e.g., `maven:org.springframework:spring-core` → `maven:org.springframework:spring-jcl`), the `createDependencies()` function in `@nx/maven` attempts to resolve both source and target through `rootToProjectMap.get()`. Since external Maven identifiers are not workspace project roots, this returns `undefined`, which then crashes Nx core's `fileDataDepTarget()` with:

```
TypeError: Cannot read properties of undefined (reading '1')
```

This makes all Nx commands that build the project graph (`nx show projects`, `nx graph`, etc.) fail in any Maven workspace that has transitive external dependencies.

## Expected Behavior

External-to-external Maven dependency edges should be silently filtered out in `createDependencies()`, since they cannot be resolved to workspace projects. Only workspace-to-workspace and workspace-to-external dependencies should be returned.

After this fix, `nx show projects`, `nx graph`, and all other Nx commands work correctly in Maven workspaces with transitive external dependencies.

## Related Issue(s)

Fixes #34840

## Changes

- **`packages/maven/src/plugins/dependencies.ts`**: Added validation to filter out dependencies where the source or target cannot be resolved to a workspace project. This matches the pattern used in the Gradle plugin fix (#29943).
- **`packages/maven/src/plugins/dependencies.spec.ts`**: Added comprehensive test coverage for `createDependencies()` with 6 test cases covering:
  - No maven data available
  - Workspace-to-workspace dependencies
  - Workspace-to-external dependencies
  - External-to-external dependencies (filtered out)
  - Unknown source roots (filtered out)
  - Unknown target roots (filtered out)

### DIFF
--- a/packages/maven/src/plugins/dependencies.ts
+++ b/packages/maven/src/plugins/dependencies.ts
@@ -32,18 +32,22 @@ export const createDependencies: CreateDependencies = async (
   );
 
   // Extract and transform dependencies from the mavenData
-  const transformedDependencies = mavenData.createDependenciesResults.map(
-    (dep) => ({
-      ...dep,
-      source: dep.source.startsWith('maven:')
-        ? dep.source
-        : rootToProjectMap.get(dep.source),
-      // External deps use maven: prefix — pass through as-is
-      target: dep.target.startsWith('maven:')
-        ? dep.target
-        : rootToProjectMap.get(dep.target),
-    })
-  );
+  const transformedDependencies = [];
+  for (const dep of mavenData.createDependenciesResults) {
+    // External-to-external edges (e.g., maven:org.foo:bar -> maven:org.baz:qux)
+    // are emitted by the Kotlin plugin but cannot be dependency sources
+    if (dep.source.startsWith('maven:')) {
+      continue;
+    }
+    const source = rootToProjectMap.get(dep.source);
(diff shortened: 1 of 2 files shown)


packages/maven/src/plugins/dependencies.ts
```diff
@@ -32,18 +32,22 @@ export const createDependencies: CreateDependencies = async (
...
+      continue;
+    }
+    const source = rootToProjectMap.get(dep.source);
+    // Pass through external targets (maven: prefix) as-is
+    const target = dep.target.startsWith('maven:')
+      ? dep.target
+      : rootToProjectMap.get(dep.target);
+    if (source && target) {
+      transformedDependencies.push({ ...dep, source, target });
+    }
```
llwt, line 49: @FrozenPandaz Is the `else` case here a valid/common situation? Should we log if we're unable to resolve source/dest for non-external targets?

FrozenPandaz:
THis was fixed in the original PR #34368 but you have to use a compatible version so `22.6.0-beta.13`

PR #34861, opened 2026-03-16
fix(core): prevent DB corruption from concurrent initialization

leosvelperez:
## Current Behavior

When multiple processes call `connectToNxDb()` concurrently (plugin workers via `startAnalytics()`, daemon, main CLI), two bugs can corrupt the workspace database:

**Bug 1: Lock file inode race.** `unlock_file()` deletes the lock file after unlocking, allowing a subsequent `File::create()` to produce a new file with a different inode. Two processes can hold "the lock" simultaneously on different file objects, breaking mutual exclusion.

**Bug 2: Partial file cleanup on version mismatch/connection failure.** The `reason` arm and `Err` arm in `initialize_db` call `remove_file(db_path)` which only deletes `.db`, leaving stale `.db-wal` and `.db-shm` on disk. The recursive `initialize_db` creates a fresh `.db`, but SQLite detects the stale WAL (different inode salt) and deletes it — destroying all data that existed only in the WAL.

Both bugs lead to:
```
Database file exists but has no metadata table.
```

## Expected Behavior

1. Lock file persists across lock/unlock cycles — all processes serialize through the same inode
2. When DB recreation is needed, all auxiliary files (`.db`, `.db-wal`, `.db-shm`) are cleaned up together via `remove_all_database_files`

### DIFF
--- a/packages/nx/src/native/db/initialize.rs
+++ b/packages/nx/src/native/db/initialize.rs
@@ -125,15 +125,10 @@ fn create_db_error(operation: &str, error: anyhow::Error) -> anyhow::Error {
 
 pub(super) struct LockFile {
     file: File,
-    path: PathBuf,
 }
 
 pub(super) fn unlock_file(lock_file: &LockFile) {
-    if lock_file.path.exists() {
-        fs4::fs_std::FileExt::unlock(&lock_file.file)
-            .and_then(|_| remove_file(&lock_file.path))
-            .ok();
-    }
+    fs4::fs_std::FileExt::unlock(&lock_file.file).ok();
 }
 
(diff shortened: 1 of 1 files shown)


packages/nx/src/native/db/initialize.rs
```diff
@@ -125,15 +125,10 @@ fn create_db_error(operation: &str, error: anyhow::Error) -> anyhow::Error {
 
 pub(super) struct LockFile {
```
FrozenPandaz: Seems like we can remove this now because it's just a File

FrozenPandaz (approved):

PR #34863, opened 2026-03-16
fix(core): show pid and cwd when waiting for task in another nx process

juristr:
## Current Behavior

QoL improvement. When a task is already running in another Nx process, the message only says:
```
Waiting for doc-designer:serve in another nx process
```
No indication of where that other process is running, making it hard to identify the source (especially with multiple agents/terminals).

image

## Expected Behavior

The message now includes the pid and working directory of the other process:
```
Waiting for doc-designer:serve in another nx process (pid: 12345, cwd: /path/to/workspace)
```

Screenshot 2026-03-16 at 16 49 02

## Related Issue(s)

N/A

### DIFF
--- a/packages/nx/src/native/tasks/running_tasks_service.rs
+++ b/packages/nx/src/native/tasks/running_tasks_service.rs
@@ -15,6 +15,13 @@ pub const SCHEMA: &str = "CREATE TABLE IF NOT EXISTS running_tasks (
     cwd TEXT NOT NULL
 );";
 
+#[napi(object)]
+pub struct RunningTaskContext {
+    pub task_id: String,
+    pub pid: u32,
+    pub cwd: String,
+}
+
 #[napi]
 struct RunningTasksService {
     db: Arc<Mutex<NxDbConnection>>,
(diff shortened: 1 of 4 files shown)


packages/nx/src/native/tasks/running_tasks_service.rs
```diff
@@ -116,6 +122,27 @@ impl RunningTasksService {
         Ok(())
     }
 
+    #[napi]
+    pub fn get_running_task_context(
```
FrozenPandaz: Return this information out of `get_running_tasks` itself

juristr: yeah I figured that might be the cleaner solution 🤔. But I wanted to first understand whether we have the appetite for this feature in the first place 😆.

packages/nx/src/native/tasks/running_tasks_service.rs
```diff
@@ -116,6 +122,27 @@ impl RunningTasksService {
...
+                let command: String = row.get(1)?;
+                let cwd: String = row.get(2)?;
+                Ok((pid, command, cwd))
+            },
+        )? {
+            Ok(Some(RunningTaskContext { pid, cwd }))
+        } else {
+            Ok(None)
+        }
+    }
```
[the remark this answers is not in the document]
juristr: Not applicable here — `NxDbConnection::query_row` already calls `.optional()` on the raw rusqlite result, converting `QueryReturnedNoRows` into `Ok(None)`. The `?` operator is safe.

PR #34871, opened 2026-03-16
fix(gradle): remove annotations from atomizer

lourw:
## Current Behavior

The Gradle plugin's test class atomizer was incorrectly treating Kotlin annotation class declarations as test targets, generating invalid test tasks for them. This caused issues when projects defined custom annotations alongside their test classes—the atomizer would attempt to run annotation classes as tests.                                        

## Expected Behavior
                                                                                                                                                
              
Both the AST-based parser and the regex fallback parser now skip annotation classes, matching the existing behavior for data classes, enum classes, sealed classes, and abstract classes. A new test suite covers all annotation class exclusion scenarios for both parsers, including files with multiple annotation classes and mixed annotation/test class files.

## Related Issue(s)

Fixes #


barbados-clemens (approved):

FrozenPandaz (approved):
Please update maven

PR #34880, opened 2026-03-17
fix(storybook): update configuration generator nx.json

Roozenboom:
update namedImports and targetDefaults to not include `tsconfig.storybook.json` in case the framework is angular

## Current Behavior

When generating storybook configuration for a new project, the nx.json is updated with an entry in the namedImports and targetDefaults for `tsconfig.storybook.json. This file does not exists when angular is used as uiFramework

## Expected Behavior

In case the uiFramework is angular the file `tsconfig.storybook.json` should not be added to the namedInputs or targetDefaults in nx.json

## Related Issue(s)

Fixes #34879


leosvelperez (approved):
Thanks!

PR #34889, opened 2026-03-17
docs(core): show supported version range for  subcommand

AgentEnder:
## Current Behavior
There's not a great way to attach metadata to CLI commands to influence rendering the markdown docs for them

## Expected Behavior
There's a metadata system currently used to express minimum support version for the show target command

## Related Issue(s)

Fixes #

### DIFF
--- a/astro-docs/src/content/docs/features/CI Features/sandboxing.mdoc
+++ b/astro-docs/src/content/docs/features/CI Features/sandboxing.mdoc
@@ -149,6 +149,10 @@ the JSON report.
 Check what your tasks currently declare before enabling sandboxing.
 Use `nx show target` to inspect a specific target's resolved inputs and outputs:
 
+{% aside type="caution" title="Nx 22.6+ Required" %}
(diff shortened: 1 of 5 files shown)


astro-docs/src/content/docs/features/CI Features/sandboxing.mdoc
```diff
@@ -149,6 +149,10 @@ the JSON report.
 Check what your tasks currently declare before enabling sandboxing.
 Use `nx show target` to inspect a specific target's resolved inputs and outputs:
 
+{% aside type="caution" title="Nx 22.6+ Required" %}
+The `--inputs` and `--outputs` flags for `nx show target` require Nx 22.6 or later.
+{% /aside %}
```
barbados-clemens, line 154: this is already mentioned min nx version. but it states 22.5+ at the top of this article. we should just change that to 22.6 if that's actually what is needed

barbados-clemens (approved):

PR #34890, opened 2026-03-17
feat(vite): add configurable ts paths build/test targets and stabilize build coordination

shairez:
## Current Behavior

When using the Vite Nx plugins:
- `nxViteTsPaths` assumes fixed target names (`build`/`test`) and does not support custom target naming for dependency builds.
- Path alias resolution may fail when a tsconfig path key has multiple candidate values because only the first path is tried.
- Build coordination in watch/serve can fail silently or behave inconsistently:
  - non-zero build exits are not surfaced clearly
  - watcher callbacks may receive undefined payloads
  - vitest browser workflows can trigger duplicate builds in some scenarios

## Expected Behavior

With these changes:
- `nxViteTsPaths` supports configurable `buildTarget` and `testTarget` options, with backward-compatible defaults.
- Tsconfig path resolution correctly iterates through multiple mapped paths and resolves the first valid file.
- Build coordination behavior is more robust and predictable:
  - failed dependency builds are surfaced and forcing the process to stop instead of silently failing and letting vitest browser mode open the browser (for example)
  - undefined watcher payloads are safely handled
  - build coordination is only attached in the right execution path, preventing duplicate builds in vitest browser mode

### DIFF
--- a/e2e/vite/src/vite.test.ts
+++ b/e2e/vite/src/vite.test.ts
@@ -192,12 +192,22 @@ describe('@nx/vite/plugin', () => {
         `generate @nx/react:library ${myBuildableLib} --directory=libs/${myBuildableLib} --bundler=vite --unitTestRunner=vitest --buildable`
       );
 
+      // Note: target names must not collide with the inferred targets or the
+      // atomized targets from ciTargetName (e.g. `test-ci`).
       updateJson(`apps/${myApp}/project.json`, (json) => {
-        json.targets['test-ci'] = { ...json.targets.test };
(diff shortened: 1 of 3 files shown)


jaysoo:
Hey @shairez, thanks for this PR. Can you add a test to `e2e-vite` as well to cover this case? We can help you with that as well if needed.

shairez:
Thanks @jaysoo !
Added a test, please let me know if it's OK or if anything missing :pray:

jaysoo (approved):

PR #34901, opened 2026-03-18
fix(testing): gracefully handle broken jest configs in alias migration

juristr:
## Current Behavior

The `replace-removed-matcher-aliases` migration crashes the entire migration runner when a project has a broken or misconfigured `jest.config.ts`. This was discovered while upgrading nx-labs from Nx 21 to 22 via `nx migrate --run-migrations`.

The migration uses Jest's `readConfig()` and `Runtime.createContext()` to resolve jest configs, but has no error handling around these calls. If any project has issues like:

- A missing file referenced in the config (e.g. `.lib.swcrc` that was renamed to `.swcrc`)
- A missing transform module (e.g. `@swc/jest` not installed)
- A missing preset file (e.g. `jest.preset.ts` instead of `jest.preset.js`)

...the entire migration fails with an opaque error:

```
Error: Command failed: /var/folders/.../node_modules/.bin/nx _migrate --run-migrations
    at checkExecSyncError (node:child_process:925:11)
  status: 1,
  stdout: null,
  stderr: null
```

The actual errors are swallowed by the nested `execSync` call, making it very hard to debug.

## Expected Behavior

The migration should skip projects with broken jest configs and continue processing the rest of the workspace.

## Fix

Wrapped the `readConfig` / `Runtime.createContext` / `SearchSource.getTestPaths` block in a try-catch that skips the failing project. This matches the defensive pattern already used in `packages/jest/src/plugins/plugin.ts` for similar jest config resolution.

## Test Plan

- Added 3 tests covering: missing file reference, missing preset, and verifying valid projects are still processed when a sibling project has a broken config
- All existing tests continue to pass

### DIFF
--- a/packages/jest/src/migrations/update-21-3-0/replace-removed-matcher-aliases.ts
+++ b/packages/jest/src/migrations/update-21-3-0/replace-removed-matcher-aliases.ts
@@ -146,10 +146,19 @@ async function resolveTestPaths(
     return null;
   }
 
-  const source = new SearchSource(jestContext);
-  const specs = await source.getTestPaths(
-    config.globalConfig,
-    config.projectConfig
-  );
+  let specs: Awaited<ReturnType<SearchSource['getTestPaths']>>;
+  try {
+    const source = new SearchSource(jestContext);
+    specs = await source.getTestPaths(
+      config.globalConfig,
+      config.projectConfig
+    );
+  } catch (e) {
+    const message = e instanceof Error ? e.message : String(e);
+    logger.warn(
+      `Could not resolve test paths for "${jestConfigFile}": ${message}. Skipping this project for matcher alias replacement.`
+    );
+    return null;
+  }
   return specs.tests.map((t) => posix.normalize(relative(tree.root, t.path)));
 }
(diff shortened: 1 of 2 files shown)


packages/jest/src/migrations/update-21-3-0/replace-removed-matcher-aliases.ts
```diff
@@ -98,23 +98,58 @@ async function getTestFilePaths(tree: Tree): Promise<string[]> {
...
-    }
+    return null;
   }
 
-  return Array.from(testFilePaths);
+  const source = new SearchSource(jestContext);
+  const specs = await source.getTestPaths(
+    config.globalConfig,
+    config.projectConfig
+  );
```
leosvelperez: Broken/invalid regex patterns can also throw here. We need to guard this as well.

leosvelperez (changes requested):

packages/jest/src/migrations/update-21-3-0/replace-removed-matcher-aliases.ts
juristr: good catch

leosvelperez (approved):

PR #34941, opened 2026-03-20
fix(core): batch processes should be tracked by sandboxing

AgentEnder:
## Related Issue(s)

Fixes #


AgentEnder:
This ended up w/ not enough value as we can't guarantee no violations here.

PR #34946, opened 2026-03-21
chore(devkit): build devkit to local dist and use nodenext

FrozenPandaz:
## Current Behavior

`@nx/devkit` builds to `<workspaceRoot>/dist/packages/devkit/` — outside its own package directory. Other packages reach into that shared workspace `dist/` to consume devkit, and consumers using `moduleResolution: nodenext` need a custom Module._resolveFilename hack to find it.

This is inconsistent with `nx` itself, which already builds to `packages/nx/dist/` (#34111).

## Expected Behavior

`@nx/devkit` builds to `packages/devkit/dist/` and is consumed via a standard `exports` map. Workspace consumers resolve through normal pnpm symlinks; the `@nx/nx-source` condition still lets in-repo code resolve to `.ts` source during dev.

This unblocks the rest of the workspace from migrating in the same direction (a `dist-build-migration` Claude skill is included as a per-package playbook).

## How to review

The PR has **307 files but only 3 categories of work**. Most of the diff is mechanical.

### 1. The actual migration — review carefully (~10 files)

Devkit package config and the `@nx/nx-source` condition wiring:

- `packages/devkit/package.json` — `exports` map, `typesVersions`, `files`, `type: commonjs`, `main`/`types` repointed to `dist/`
- `packages/devkit/tsconfig.lib.json` — `outDir: dist`, `nodenext` module/resolution
- `packages/devkit/project.json` — `build-base` outputs, `nx-release-publish.packageRoot`, `manifestRootsToUpdate`
- `packages/devkit/internal.ts` — re-exports `src/utils/*` and `src/generators/*` symbols (replaces the `@nx/devkit/src/*` deep-import pattern)
- `packages/devkit/eslint.config.mjs` — ignore `dist`
- `packages/devkit/{README.md → readme-template.md}` — README is now generated, gitignored
- `.gitignore` — ignore the generated `packages/devkit/README.md`
- `scripts/nx-release.ts` — devkit's `package.json` now lives at `packages/devkit/package.json`, not `dist/packages/devkit/package.json`
- `scripts/patched-jest-resolver.js` — adds `@nx/nx-source` condition

### 2. Mass import rewrites — already marked viewed (~205 files)

Every internal import of `@nx/devkit/src/utils/<x>` or `@nx/devkit/src/generators/<x>` was rewritten to `@nx/devkit/internal`:

```diff
-import { foo } from '@nx/devkit/src/utils/bar';
+import { foo } from '@nx/devkit/internal';
```

I marked these files as **viewed** in the GitHub review UI to clear them from the unread queue. They're across nearly every plugin (angular, react, next, vite, webpack, rspack, expo, jest, etc.).

### 3. Follow-on cleanups (~10 files)

These appear in their own commits and are easy to review individually:

- **`cleanup(angular-rspack)`** — removes the `patchDevkitRequestPath` runtime hack from 14 example configs; devkit now resolves through standard node_modules (the MF patch stays — module-federation isn't migrated yet).
- **`cleanup(devkit)` typedoc** — drops dead path mappings + redundant include manipulation in `astro-docs/src/plugins/utils/typedoc/typedoc.ts` (verified docs build is byte-identical with/without the removed config).
- **`cleanup(devkit)` eslint ignores** — drops `'**/*.d.ts'` from `packages/devkit/eslint.config.mjs` since `.d.ts` files only emit to `dist/` (already ignored).
- **`fix(angular)` eslint quote** — quote-agnostic regex in `e2e/angular/src/projects-linting.test.ts` so the test still disables `prefer-standalone` after the angular-eslint generator switched to double quotes (latent bug surfaced by CI).
- **`fix(testing)` jest migration** — removes a stray unused `@nx/devkit/internal` import.
- **e2e test fallout** — `e2e/{angular,next}/src/*.test.ts` lose access to `@nx/devkit/src/utils/string-utils` (e2e tests can't use `/internal`); inline equivalents using public `names()` API. `e2e/nx-build/src/nx-build.test.ts` updates the expected output path.

## Local verification

- `pnpm nx run-many -t test,build,lint -p devkit` ✓
- `pnpm nx run astro-docs:build` ✓ — devkit reference pages still generate (148 markdown files; identical to a build with the path mappings re-added as a control)
- `pnpm nx run-many -t build -p examples-angular-rspack-csr-css,examples-angular-rspack-ssr-css,examples-angular-rspack-zoneless-csr-css,examples-angular-rspack-mf-host,examples-angular-rspack-mf-remote --skip-nx-cache` ✓ — examples build without the devkit patch

## Known follow-up (not in this PR)

- `scripts/nx-release.ts` still has `hackFixForDevkitPeerDependencies()` (a band-aid from #32406 that re-adds `<=` to devkit's `nx` peer-dep range after `nx release version` strips it). The proper fix is to set `preserveMatchingDependencyRanges: true` in `packages/devkit/project.json` and delete the hack — that needs a release dry-run to verify, so it's queued separately.

## Related Issue(s)

Follow-up to #34111.


packages/devkit/package.json
```diff
@@ -42,12 +63,41 @@
...
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./package.json": "./package.json",
+    "./migrations.json": "./migrations.json",
+    "./internal": {
+      "@nx/nx-source": "./internal.ts",
+      "types": "./dist/internal.d.ts",
+      "default": "./dist/internal.js"
+    },
```
AgentEnder, line 86: We are gonna break a ton of community plugins if we dont leave the export path for the project name and root utils (and maybe the create nodes hash utils) as exported just as an FYI

AgentEnder (approved):

PR #34951, opened 2026-03-22
feat(core): add shell tab-completion (bash, zsh, fish, powershell)

FrozenPandaz:
## Current Behavior

Nx has no shell tab-completion. Users have to type full command, project, target, and generator names from memory.

## Expected Behavior

After installing the generated completion script, users get instant tab-completion for:

- **Commands**: `nx <TAB>` lists every command (`run`, `generate`, `add`, ...) **plus** every infix target name found in the project graph (`build`, `serve`, `compile`, `bundle-rsc`, ...).
- **Subcommands**: `nx show <TAB>` → `project`, `projects`, `target`
- **Projects**: `nx show project <TAB>` shows all workspace projects
- **`project:target` pairs**: `nx show target <TAB>` and `nx run <TAB>` first complete project names with a trailing `:`, then a second TAB lists that project's targets — no backspace + manual `:` needed
- **Target names**: `nx run-many -t <TAB>` / `nx affected -t <TAB>`
- **Infix target invocations**: `nx build <TAB>` shows projects that actually have the target. Works for **any** target name in the graph, not a hardcoded set.
- **`nx show target <project>:<target>` keywords**: `nx show target i<TAB>` → `inputs`, `outputs`. Also `nx show target inputs <TAB>` / `nx show target outputs <TAB>` complete project:target.
- **Generators**: `nx g <TAB>` lists plugins (with trailing `:` for two-stage drilling) **and** the bare generator names so `nx g app<TAB>` → `application`. `nx g @nx/react:<TAB>` lists that plugin's generators. Workspace-local plugin projects (a `libs/my-plugin` with its own `generators.json`) are discovered alongside installed npm plugins.
- **`nx add`**: `nx add @nx/r<TAB>` lists the first-party `@nx/*` set (`@nx/react`, `@nx/rspack`, `@nx/remix`, ...).
- **Descriptions**: shells that render them (zsh, fish) get `value\tdescription` pairs in the menu. bash and powershell get bare names — bash has no description protocol in `compgen -W`; powershell uses the single-arg `CompletionResult` ctor.

### Installation

```sh
nx completion bash >> ~/.bashrc
nx completion zsh  >> ~/.zshrc
mkdir -p ~/.config/fish/completions && nx completion fish > ~/.config/fish/completions/nx.fish
nx completion powershell | Out-File -Append $PROFILE
```

Zsh also requires `autoload -U compinit && compinit` above the nx block; the generated script prints a friendly warning pointing to this if `compdef` isn't loaded. Fish's `>` redirect doesn't auto-create parent directories, hence the `mkdir -p`.

### Debugging

If completion produces no suggestions, set `NX_VERBOSE_LOGGING=1` and press TAB again. The wrappers stop discarding completion `stderr` when the flag is set, and the binary's `catch` surfaces the underlying error there. Reuses Nx's existing verbose-logging knob — no completion-specific env var.

## Implementation notes

- **Pure JS, no native binary.** The earlier Rust prototype is dropped in favor of a JS completion handler that reads `.nx/workspace-data/project-graph.json` directly.
- **Bin-level short-circuit.** `bin/nx.ts` intercepts at the top of `main()`:
  - When `NX_COMPLETE=<shell>` is set (per-TAB request from the wrapper), it skips workspace-root detection, dotenv, daemon, native module load, and the yargs command tree, then dispatches to the value / command-surface completion paths.
  - When `nx completion <shell>` (the install command) is invoked, the same hoisted short-circuit prints the wrapper script and exits — no further bootstrap.

  This is required because `parserConfiguration({ 'strip-dashed': true })` defeats yargs' own completion detection, otherwise causing the `$0` infix handler to fire and build the full project graph (~3s cold, spawning ~25 plugin workers). The hoisting also means no scattered `argv[2] === 'completion'` special cases downstream.
- **Wrappers as plain files.** The four shell wrappers (`bash.sh`, `zsh.zsh`, `fish.fish`, `powershell.ps1`) live as real files under `packages/nx/src/command-line/completion/scripts/`. `scripts.ts` is a thin loader. Editors give real shell syntax highlighting; `shellcheck` / `fish_indent` work; no `\$` escape forest from TS template literals.
- **Trailing-colon UX for `project:target`.** Completions for `show target` / `run` contexts return `project:` (with colon). Bash and zsh scripts detect trailing colons and suppress the inserted space (`compopt -o nospace`, `compadd -S ''`). Fish has no per-completion nospace primitive (works in practice — the user deletes the space or types `:`); powershell same limitation.
- **zsh uses `compadd -d`, never `_describe`.** `_describe` splits on `:` and would mangle `my-app:build` into value `my-app` with description `build`. The wrapper parses `value\tdescription` from each completion line and feeds parallel value/display arrays to `compadd -d`. `formatDescription` collapses any literal TAB in untrusted plugin descriptions to a space so the separator can't be forged.
- **fish descriptions are native.** Fish's `complete -a` reads `value\tdescription` candidates the same way zsh's `compadd -d` does — no wrapper change needed, just the description-emitting branch broadened from "zsh only" to "shells that render descriptions".
- **Stale graph tolerated.** Completion reads `.nx/workspace-data/project-graph.json` directly without triggering a recompute. A slightly stale graph is fine; recomputing it would blow the latency budget. There's an explicit `// do not fix this by triggering a recompute` comment guarding the choice.
- **Workspace-root walk-up.** The completion fast path skips the normal bootstrap that sets `NX_WORKSPACE_ROOT_PATH`, so a local `resolveWorkspaceRoot()` walks from `process.cwd()` up to the nearest `nx.json`. The POSIX wrappers do a parallel walk-up at TAB time to find a workspace-local `nx` binary (falling back to PATH only outside a workspace).
- **Generator listing avoids one file read in stage 1.** Stage 1 (`nx g <TAB>`) only needs *names* of plugins that declare a generator collection, not their generators — `collectPluginDirs()` reads each plugin's `package.json` for the `generators` field, skips reading `generators.json` until stage 2.
- **INFIX target completion is graph-driven.** `run/completion.ts` walks the cached project graph at module load and registers a completion path for every unique target name. Falls back to a conventional set (`build`, `serve`, `test`, ...) for cold workspaces with no graph yet.

End-to-end completion latency: **~140ms typical**, scales with project-graph size.

## How to test locally

1. `pnpm nx build nx`
2. `pnpm link ./packages/nx`
3. Install the wrapper for your shell (see Installation above).
4. Open a new shell and tab away.

Verify:

- `nx <TAB>` lists commands **and** infix targets (`build`, `serve`, plus any custom ones in your workspace)
- `nx show target <TAB>` completes to `project:` (no trailing space) — second TAB lists targets
- `nx run-many -t <TAB>` lists target names
- `nx build <TAB>` lists projects that have a `build` target
- `nx g <TAB>` lists plugins **and** bare generator names; `nx g app<TAB>` → `application`
- `nx g @nx/react:<TAB>` lists that plugin's generators (workspace-local plugins also discoverable)
- `nx add @nx/r<TAB>` lists first-party plugins
- `nx show target i<TAB>` → `inputs`, `outputs`
- `NX_VERBOSE_LOGGING=1 nx <TAB>` surfaces any swallowed errors to stderr
- Completion feels instant (~140ms typical)


AgentEnder:
~~doesn't seem to be working in zsh?~~

We figured this out, there are two limitations:

- This only works for users that have a global Nx installation, otherwise we'd be putting entries into ~/.zshrc that point at a repo-specific path
- This doesn't work at all for users that are invoking nx via `npx nx` or similar, or have an alias for the `nx` command

So on my machine, I had `alias nx="npx nx"` in my shell config. This meant the prompt provider wasn't hit.

packages/nx/src/command-line/completion/scripts/powershell.ps1
AgentEnder, line 1: This one really needs tested

packages/nx/src/command-line/completion/scripts/bash.sh
AgentEnder, line 1: Lets validate this one too

packages/nx/bin/nx.ts
```diff
@@ -37,6 +37,50 @@ const isTsExt = extname(__filename).endsWith('.ts');
...
+        console.error(e);
+      }
+    }
+    return;
+  }
+
+  // `nx completion <shell>` install path — print the wrapper, skip the
+  // rest of the bootstrap. Unknown shell falls through to yargs.
+  if (process.argv[2] === 'completion') {
+    const shell = process.argv[3];
```
AgentEnder: Can we detect current shell and use it by default? Having the argv[3] as optional would be neat

packages/nx/src/command-line/add/completion.ts
```diff
@@ -0,0 +1,47 @@
+import { registerCompletion } from '../completion/metadata';
+
+// `nx add <plugin>` accepts any npm package; suggest the first-party set.
+// Keep alphabetical.
+const FIRST_PARTY_PLUGINS = [
```
AgentEnder: Use what report/list use

packages/nx/src/command-line/completion/scripts/powershell.ps1
AgentEnder, line 1: Could we do `cmd` too? I don't know if it supports things

packages/nx/src/command-line/completion/argv-layout.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface ParsedCompletionArgs {
+  tokens: string[];
+  current: string;
+  previousToken: string;
+}
+
+export function parseCompletionArgs(): ParsedCompletionArgs | null {
+  const tail = process.argv.slice(2);
+  const tokens = tail[0] === 'nx' ? tail.slice(1) : tail;
```
AgentEnder, line 19: Why do we have this as a ternary? When does it not start with `nx`?

packages/nx/src/command-line/completion/command-completions.ts
```diff
@@ -0,0 +1,154 @@
...
+  const matched = getCommandCompletions(parsed.current, parsed.tokens);
+  const completions =
+    matched !== null
+      ? matched
+      : getTopLevelCommands(parsed.current, shellRendersDescriptions());
+
+  if (completions === null || completions.length === 0) return false;
+
+  for (const line of completions) {
+    process.stdout.write(line + '\n');
```
AgentEnder: Why not console.log?

packages/nx/src/command-line/completion/command-completions.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (handler?.description === false) continue; // hidden
+    seen.add(name);
+    const desc = withDesc ? formatDescription(handler?.description) : '';
+    completions.push(desc ? `${name}${DESC_SEPARATOR}${desc}` : name);
+  }
+
+  // Infix targets + any other top-level completion-only paths.
+  const { getRegisteredTopLevelPaths } = require('./metadata') as {
+    getRegisteredTopLevelPaths: () => string[];
+  };
```
AgentEnder: typeof import(....)?

packages/nx/src/command-line/completion/command-completions.ts
```diff
@@ -0,0 +1,154 @@
...
+  for (const name of getRegisteredTopLevelPaths()) {
+    if (seen.has(name)) continue;
+    if (current && !name.startsWith(current)) continue;
+    completions.push(name);
```
AgentEnder, line 55: Maybe add seen here just for consistency

packages/nx/src/command-line/completion/command-completions.ts
```diff
@@ -0,0 +1,154 @@
...
+  args: string[]
+): string[] | null {
+  // Lazy require to avoid circular dependency with nx-commands.
+  const { commandsObject } = require('../nx-commands') as {
+    commandsObject: any;
+  };
+  const handlers = commandsObject
+    .getInternalMethods()
+    .getCommandInstance()
+    .getCommandHandlers();
```
AgentEnder: Make this a util or smth?

packages/nx/src/command-line/completion/command-completions.ts
```diff
@@ -0,0 +1,154 @@
...
+      const subName = String(usagePattern).split(/\s+/)[0];
+      if (subName === '$0') continue;
+      const formatted = withDesc ? formatDescription(desc as string) : '';
+      completions.push(
+        formatted ? `${subName}${DESC_SEPARATOR}${formatted}` : subName
+      );
+    }
+  }
+
+  const opts = temp.getOptions();
```
AgentEnder: Should be similar util, ideally reusing some stuff

packages/nx/src/command-line/completion/command-completions.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/** value\tdescription separator. TAB because completion values can contain
+ *  colons (`my-app:build`); descriptions get TABs collapsed. */
+export const DESC_SEPARATOR = '\t';
+
+/** Strip yargs' i18n marker and collapse stray TABs so they can't forge
+ *  the value/description separator. */
+export function formatDescription(raw: string | undefined): string {
+  if (!raw) return '';
+  return raw.replace(/^__yargsString__:/, '').replace(/\t/g, ' ');
```
AgentEnder, line 126: Does this ever actually even show up? We don't have i18n for the CLI

packages/nx/src/command-line/completion/command-object.ts
```diff
@@ -0,0 +1,79 @@
...
+export const yargsCompletionCommand: CommandModule = {
+  command: 'completion',
+  describe:
+    'Output shell completion script for bash, zsh, fish, or powershell. Run `nx completion --help` for installation instructions.',
+  builder: (yargs) =>
+    yargs
+      .command(bashCompletionCommand)
+      .command(zshCompletionCommand)
+      .command(fishCompletionCommand)
+      .command(powershellCompletionCommand)
```
AgentEnder: Use positional arg instead, so it can be optional and swap based on current shell

packages/nx/bin/nx.ts
AgentEnder: Actually, if no shell provided - multiselect. We can look for presence of ~/.zshrc etc to see if available.

packages/nx/bin/nx.ts
```diff
@@ -37,6 +37,50 @@ const isTsExt = extname(__filename).endsWith('.ts');
...
+      shell === 'powershell'
+    ) {
+      const { printCompletionScript } = await import(
+        'nx/src/command-line/completion/scripts'
+      );
+      await printCompletionScript(shell, {
+        force: process.argv.includes('--force'),
+      });
+      return;
+    }
```
AgentEnder: This should probably go through yargs, its not the hot path being exercised when running with NX_COMPLETE

packages/nx/src/command-line/completion/completion-providers.ts
```diff
@@ -0,0 +1,246 @@
...
+  let dir = process.cwd();
+  while (true) {
+    if (existsSync(join(dir, 'nx.json'))) {
+      return dir;
+    }
+    const parent = dirname(dir);
+    if (parent === dir) return process.cwd(); // fs root reached
+    dir = parent;
+  }
+}
```
AgentEnder: We already have this

packages/nx/src/command-line/completion/completion-providers.ts
```diff
@@ -0,0 +1,246 @@
...
+      join(workspaceRoot, '.nx', 'workspace-data');
+    const graphPath = join(dataDir, 'project-graph.json');
+    if (!existsSync(graphPath)) {
+      return null;
+    }
+    return JSON.parse(readFileSync(graphPath, 'utf-8'));
+  } catch {
+    return null;
+  }
+}
```
AgentEnder: We also have this?

packages/nx/src/command-line/completion/completion-providers.ts
```diff
@@ -0,0 +1,246 @@
...
+export function getProjectNamesWithTarget(
+  current: string,
+  targetName: string
+): string[] {
+  const graph = getCachedProjectGraph();
+  if (!graph?.nodes) {
+    return [];
+  }
+  const matches: string[] = [];
+  for (const [name, node] of Object.entries<any>(graph.nodes)) {
```
AgentEnder: No cast

packages/nx/src/command-line/completion/completion-providers.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (node?.data?.targets && targetName in node.data.targets) {
```
AgentEnder: node.data.targets[target] is probably safer here

packages/nx/src/command-line/completion/completion-providers.ts
```diff
@@ -0,0 +1,246 @@
...
+  const targetPrefix = current.slice(colonIdx + 1);
+  return getTargetNameCompletions(targetPrefix, projectName).map(
+    (t) => `${projectName}:${t}`
+  );
+}
+
+/** Generator completion. Stage 1 (`nx g <TAB>`) emits plugin names (with `:`)
+ *  and bare generator names (for `nx g application`); stage 2 emits
+ *  `plugin:generator`. */
+export function completeGenerator(current: string): string[] {
```
AgentEnder, line 68: There are a lot of things here that we have existing utils (and types) for... listing plugins that have generators, reading the generators.json, etc

packages/nx/src/command-line/completion/completion-providers.ts
```diff
@@ -0,0 +1,246 @@
...
+  const generatorsJson = readJsonSafe(join(pluginDir, field));
+  if (!generatorsJson) {
+    return [];
+  }
+  const collection =
+    generatorsJson.generators ?? generatorsJson.schematics ?? {};
+  return Object.keys(collection).filter((k) => !collection[k]?.hidden);
+}
+
+function readJsonSafe(path: string): any | null {
```
AgentEnder, line 174: Should probably use `readJsonFile`?

packages/nx/src/command-line/completion/registrations.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+// disables completion for that form — these assertions catch it.
+
+describe('completion/registrations', () => {
+  it('registers `run` with a positional completer', () => {
+    const match = findCompletionMetadata(['run']);
+    expect(match?.metadata.positionals?.[0]?.complete).toBeInstanceOf(Function);
+  });
+
+  it('registers `generate` and its `g` alias', () => {
+    expect(findCompletionMetadata(['generate'])).not.toBeNull();
```
AgentEnder: For each command, for each arg, if arg has alias alias returns same object

packages/nx/src/command-line/add/completion.ts
```diff
@@ -0,0 +1,21 @@
+import { registerCompletion } from '../completion/metadata';
+
+// Derived from nx's own packageGroup — same list `nx report` uses.
```
AgentEnder: Use a util for this, share it with report.... think readNxPackageGroup or smth

packages/nx/src/command-line/completion/argv-layout.ts
AgentEnder: We think this is because of `nxw`./ `.nx` setup

packages/nx/src/command-line/completion/command-object.ts
```diff
@@ -0,0 +1,78 @@
...
+      await scripts.printCompletionScript(args.shell, args);
+      process.exit(0);
+    }
+    const chosen = await pickShellsInteractively();
+    if (chosen.length === 0) {
+      process.stderr.write('nx: no shells selected — nothing installed.\n');
+      process.exit(0);
+    }
+    for (const shell of chosen) {
+      await scripts.printCompletionScript(shell, args);
```
AgentEnder: Split into printCompletionScript and installCompletionScript, move `--stdout` flag branching to this file. Should never print multiple completion scripts to stdout, no practical reason for that.

packages/nx/src/command-line/completion/registrations.ts
AgentEnder, line 1: This is imported in `value-completions`, maybe it should be in `metadata`?

packages/nx/src/command-line/completion/command-object.ts
```diff
@@ -0,0 +1,81 @@
...
+      process.exit(0);
+    }
+    const chosen = await pickShellsInteractively();
+    if (chosen.length === 0) {
+      process.stderr.write('nx: no shells selected — nothing installed.\n');
+      process.exit(0);
+    }
+    for (const shell of chosen) {
+      emit(shell, args);
+    }
```
AgentEnder: Throw if `--stdout` and chosen > 1

packages/nx/src/command-line/completion/scripts.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Write the wrapper to the shell's default rc location, replacing any prior
+ * nx-completion block (idempotent via the begin/end markers). Logs the
+ * resolved path and a one-line "open a new shell" hint.
+ */
+export function installCompletionScript(
+  shell: Shell,
+  args: CompletionScriptArgs = {}
+): void {
+  if (!args.force && !isNxOnPath()) warnNxNotOnPath(shell);
```
AgentEnder: This warning would be emitted multiple times when a user picks more than one shell

packages/nx/src/command-line/completion/scripts.ts
```diff
@@ -0,0 +1,127 @@
...
+  writeScriptToRcFile(shell);
+}
+
+function writeScriptToRcFile(shell: Shell): void {
+  const script = generateScript(shell);
+  const path = installPathFor(shell);
+  if (!path) {
+    process.stderr.write(
```
AgentEnder: console.warn :)

packages/nx/src/command-line/completion/scripts.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+  mkdirSync(dirname(path), { recursive: true });
+
+  // Fish keeps each completion in its own file; bash/zsh/powershell append
+  // to a shared rc file so we strip any existing nx block first.
+  if (shell === 'fish') {
+    writeFileSync(path, script);
+  } else {
+    const existing = existsSync(path) ? readFileSync(path, 'utf8') : '';
+    const stripped = existing.replace(
```
AgentEnder: Why are we stripping the headers out here? We want those in the updated file s.t. its easy to remove

packages/nx/src/command-line/completion/scripts.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (!dir) continue;
+    for (const name of names) {
+      if (existsSync(join(dir, name))) return true;
+    }
+  }
+  return false;
+}
+
+function warnNxNotOnPath(shell: Shell): void {
+  process.stderr.write(
```
AgentEnder: logger.warn or console.warn?

packages/nx/src/command-line/completion/scripts.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+const wrappers: Record<Shell, string> = {
+  bash: readFileSync(join(__dirname, 'scripts', WRAPPER_FILES.bash), 'utf-8'),
+  zsh: readFileSync(join(__dirname, 'scripts', WRAPPER_FILES.zsh), 'utf-8'),
+  fish: readFileSync(join(__dirname, 'scripts', WRAPPER_FILES.fish), 'utf-8'),
+  powershell: readFileSync(
+    join(__dirname, 'scripts', WRAPPER_FILES.powershell),
+    'utf-8'
+  ),
+};
```
AgentEnder: Inline these into generateScript, only read the ones you need

packages/nx/src/command-line/completion/trigger.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (originalNxComplete === undefined) {
+      delete process.env.NX_COMPLETE;
+    } else {
+      process.env.NX_COMPLETE = originalNxComplete;
+    }
+  });
+
+  describe('isCompletionRequest', () => {
+    it('is false when NX_COMPLETE is unset', () => {
+      delete process.env.NX_COMPLETE;
```
AgentEnder: We have a helper for dealing with env vars in tests to prevent cross test contamination

AgentEnder: withEnvVars or smth

packages/nx/src/command-line/completion/trigger.ts
```diff
@@ -0,0 +1,22 @@
...
+export function isCompletionRequest(): boolean {
+  return Boolean(process.env.NX_COMPLETE);
+}
+
+export function getCompletionShell(): CompletionShell | null {
+  const raw = process.env.NX_COMPLETE;
+  if (raw && KNOWN_SHELLS.has(raw)) {
+    return raw as CompletionShell;
+  }
+  return null;
```
AgentEnder, line 21: Maybe throw or warn

AgentEnder: This is fine, falls back to bash-like behavior

packages/nx/src/command-line/completion/value-completions.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    afterEach(() => {
+      logSpy.mockRestore();
+    });
+
+    // argv layout at TAB time: [node, nx-bin, ...shellTokens, currentPartial]
+    // where shellTokens[0] is the command name as the shell sees it ('nx').
+    function setArgv(...userTokens: string[]) {
+      process.argv = ['node', 'nx-bin', 'nx', ...userTokens];
+    }
```
AgentEnder: Can we add a default parameter to tryValueCompletion or smth so we aren't mutating global state in these

packages/nx/src/command-line/generate/completion.ts
```diff
@@ -0,0 +1,8 @@
+import { registerCompletion } from '../completion/metadata';
+import { completeGenerator } from '../completion/completion-providers';
+
+const generateCompletion = {
+  positionals: [{ complete: completeGenerator }],
+};
+registerCompletion('generate', generateCompletion);
+registerCompletion('g', generateCompletion); // alias
```
AgentEnder, line 8: FUTURE: flags, and flag types can be read from generator if already known... think things like root knowing to use directories, project name selectors, etc... things with enum, bool, and so on.

packages/nx/src/command-line/generate/completion.ts
AgentEnder, line 1: Generate should really still allow hitting `tab` to autocomplete path segments

AgentEnder (changes requested):

packages/nx/src/command-line/completion/scripts/bash.sh
FrozenPandaz: validated

packages/nx/src/command-line/completion/scripts/powershell.ps1
FrozenPandaz: cmd.exe has no real completion API. Unlike bash/zsh/fish/powershell, cmd's built-in tab completion is hardcoded to
  filenames/dirs. There's no way to register a custom completer without third-party tooling like [Clink](https://chrisant996.github.io/clink/).

Powershell is the choice for Windows

packages/nx/src/command-line/generate/completion.ts
FrozenPandaz: It does now

packages/nx/src/command-line/completion/command-completions.ts
FrozenPandaz:
```
> require('./packages/nx/dist/src/command-line/nx-commands.js').commandsObject.getInternalMethods().getUsageInstance().getDescriptions()
{ version: '__yargsString__:Show version number' }
```

AgentEnder (approved):

PR #34960, opened 2026-03-23
fix(module-federation): strip version suffix from npm dependency names for bun compatibility

theescodes:
## Current Behavior

When using Bun as the package manager, the Nx project graph includes version numbers in external node dependency targets (e.g., `npm:@ngrx/store@21.0.1`), whereas npm uses `npm:@ngrx/store` (without version).

The `collectDependencies` function in `packages/module-federation/src/utils/dependencies.ts` strips the `npm:` prefix but not the version suffix, resulting in package names like `@ngrx/store@21.0.1` being passed to `sharePackages()`. These fail to match entries in `package.json` (which uses `@ngrx/store`), so **no npm packages are shared** between Module Federation host and remotes, causing runtime errors such as:

```
NG0201: No provider found for InjectionToken @ngrx/store Root Store Provider
```

## Expected Behavior

The version suffix is stripped from npm dependency names before they are added to the shared packages set. For example:
- `@ngrx/store@21.0.1` -> `@ngrx/store`
- `rxjs@7.8.1` -> `rxjs`
- `lodash` (no version) -> `lodash` (unchanged)

This ensures shared packages are correctly resolved regardless of the package manager (npm, pnpm, yarn, or bun).

## Related Issue(s)

https://github.com/nrwl/nx/issues/35135


leosvelperez (approved):
Thanks!

PR #34961, opened 2026-03-23
fix(angular): update duplicate migration keys

JackWBaker:
Fix duplicate migrations keys as only the last migration was getting picked up.

## Current Behavior
When running nx migrate only the module federation migration is picked up and not updating the core angular packages.

## Expected Behavior
Both migrations should be added to migrations.json.


leosvelperez (approved):
Thanks!

PR #34965, opened 2026-03-23
fix(js): stop generating baseUrl in tsconfig, use ./ prefix for path mappings

### LINKED ISSUE #32958: NX-generated tsconfig uses deprecated baseUrl causing incompatibility with modern bundlers
# NX-generated tsconfig uses deprecated `baseUrl` causing incompatibility with modern bundlers

## Current Behavior

NX generates `tsconfig.base.json` with `baseUrl: "."` and non-relative paths in the `paths` mapping:

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@myorg/lib1": ["packages/lib1/src"],
      "@myorg/lib2": ["packages/lib2/src"]
    }
  }
}
```

This configuration causes failures with modern TypeScript tooling that follows newer TypeScript standards, specifically:
- **tsdown** with `tsgo` option (TypeScript isolated declarations)
- Potentially other bundlers using TypeScript's isolated declarations feature

## Error Messages

When using modern bundlers with this configuration:

```
TS5090: Non-relative paths are not allowed. Did you forget a leading './'?

TS5102: Option 'baseUrl' has been removed. Please remove it from your configuration.
  Use '"paths": {"*": ["../../*"]}' instead.
```

## Expected Behavior

NX should generate tsconfig with relative paths that are compatible with modern TypeScript tooling:

```json
{
  "compilerOptions": {
    "paths": {
      "@myorg/lib1": ["./packages/lib1/src"],
      "@myorg/lib2": ["./packages/lib2/src"]
    }
  }
}
```

## Steps to Reproduce

1. Create a new NX workspace with TypeScript libraries
2. Install and configure `tsdown` for building a library
3. Enable DTS generation (auto-detected or explicit)
4. Run build

**Example tsdown.config.ts:**
```typescript
import { defineConfig } from 'tsdown';

export default defineConfig({
  sourcemap: true,
  tsconfig: 'tsconfig.lib.json',
  skipNodeModulesBundle: true
  // dts is auto-detected from package.json
});
```

**Build fails with:**
```
ERROR: Build failed with errors:
[UNRESOLVED_IMPORT] Error: Could not resolve './types.js' in ../types/src/index.d.ts
```

## Workaround

Manually update `tsconfig.base.json`:

```diff
{
  "compilerOptions": {
-   "baseUrl": ".",
    "paths": {
-     "@myorg/lib1": ["packages/lib1/src"],
-     "@myorg/lib2": ["packages/lib2/src"]
+     "@myorg/lib1": ["./packages/lib1/src"],
+     "@myorg/lib2": ["./packages/lib2/src"]
    }
  }
}
```

After this change, builds work correctly with modern bundlers.

### Why This Matters

1. **TypeScript Evolution**: The TypeScript team has deprecated `baseUrl` in favor of relative paths
2. **Modern Tooling**: Tools using TypeScript's isolated declarations (like tsdown's tsgo) require the newer format
3. **Future Compatibility**: As more tools adopt isolated declarations, this will become a broader issue

### Related Issues

- [tsdown issue #523](https://github.com/rolldown/tsdown/issues/523) - Initially thought to be a tsdown bug, but root cause is NX's tsconfig format
- [NX issue #11374](https://github.com/nrwl/nx/issues/11374) - Earlier discussion about baseUrl/paths usage

## Proposed Solution

1. **Update NX generators** to create tsconfig files with relative paths (using `./` prefix) and without `baseUrl`
2. **Provide migration** for existing workspaces (perhaps an `nx migrate` command)
3. **Update documentation** to reflect this change and explain the reasoning
4. **Add warning** when detecting the old format to guide users to update

## Environment

- **NX Version**: (affects all versions using current tsconfig template)
- **Node Version**: v20+
- **Package Manager**: npm/pnpm/yarn
- **Operating System**: All platforms
- **Affected Bundlers**: tsdown (confirmed), potentially others using isolated declarations

## Additional Information

This is not just a tsdown-specific issue. As TypeScript evolves and more tools adopt isolated declarations and modern TypeScript features, the current NX-generated tsconfig format will cause increasing compatibility problems.

The fix is straightforward (add `./` prefix to paths, remove `baseUrl`) and maintains backward compatibility with existing TypeScript compilation while enabling compatibility with modern tooling.

leosvelperez:
## Current Behavior

Nx generators set `compilerOptions.baseUrl: "."` in generated tsconfig files and write path mappings as bare relative paths (e.g., `my-lib/src/index.ts`). `baseUrl` is deprecated in TS 6 and removed in TS 7.

## Expected Behavior

Nx no longer generates `baseUrl` in any tsconfig. Path mappings use `./` prefix (e.g., `./my-lib/src/index.ts`), making them relative to the tsconfig file without needing `baseUrl`. Existing user tsconfigs with `baseUrl` continue to work correctly.

### Generator and template changes
- Remove `baseUrl` from all templates and generator code
- `addTsConfigPath` normalizes lookup paths with `./` prefix
- Move and remove generators handle `./`-prefixed paths correctly
- Angular secondary entry points and Remix server entry paths use `./` prefix

### `resolvePathsBaseUrl` helper
New function (in `ts-config.ts`, duplicated in `register.ts`) that walks the tsconfig `extends` chain to determine the correct directory for resolving `paths` values. Finds where `paths` is defined, then looks for the applicable `baseUrl` from that point toward the root — ignoring child overrides that don't apply to the paths-defining tsconfig. When no `baseUrl` applies, returns the directory of the tsconfig that defines `paths`. All path resolver plugins and buildable-libs-utils use this helper.

### Runtime and bundler fixes for baseUrl-less tsconfigs
- **Rollup**: resolve path mappings to absolute in compiler options override using `resolvePathsBaseUrl`; use original tsconfig path (not tmp) for resolution base
- **Module Federation**: add `workspaceRoot` to `resolve.modules` in all 8 MF plugin variants (Angular/React webpack, Angular/React rspack, webpack SSR, rspack SSR, Angular rspack plugin, rspack plugin) so workspace-relative expose paths resolve without `baseUrl`
- **Next.js/Jest**: null out SWC `resolvedBaseUrl` in generated jest configs to prevent SWC from doing incorrect path alias resolution; Nx jest resolver handles this via `resolvePathsBaseUrl`
- **register.ts**: use `resolvePathsBaseUrl` for correct path alias registration
- **Path resolver plugins** (webpack, rspack, vite, expo, react-native, jest, react component testing): use `resolvePathsBaseUrl` for correct path resolution
- **buildable-libs-utils**: resolve tmp tsconfig paths to absolute so they work without `baseUrl`
- **eslint-plugin**: handle `./`-prefixed paths in AST utils

## Related Issue(s)

Fixes #32958

### DIFF
--- a/packages/js/src/utils/typescript/ts-config.ts
+++ b/packages/js/src/utils/typescript/ts-config.ts
@@ -1,6 +1,6 @@
 import { offsetFromRoot, Tree, updateJson, workspaceRoot } from '@nx/devkit';
-import { existsSync } from 'fs';
-import { dirname, join } from 'path';
+import { existsSync, readFileSync } from 'fs';
+import { dirname, isAbsolute, join, resolve } from 'path';
 import type * as ts from 'typescript';
 import { ensureTypescript } from './ensure-typescript';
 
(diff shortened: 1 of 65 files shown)


njsokol:
Thanks for working on this @leosvelperez! Newer tools like `oxc` linting with typeaware rely on this `baseUrl` being removed so appreciate the effort in removing the dependency in the generators

packages/js/src/utils/typescript/ts-config.ts
```diff
@@ -93,12 +93,75 @@ export function addTsConfigPath(
...
+export function resolvePathsBaseUrl(tsconfigPath: string): string {
+  const chain: { dir: string; raw: any }[] = [];
+  let current: string | null = tsconfigPath;
+  while (current) {
+    const absolute = resolve(current);
+    const dir = dirname(absolute);
+    try {
+      const raw = JSON.parse(readFileSync(absolute, 'utf-8'));
+      chain.push({ dir, raw });
+      current = raw.extends ? resolve(dir, raw.extends) : null;
```
jaysoo: This does not handle arrays like `extends: ["./tsconfig.base.json"]` or `extends: ["./tsconfig.base.json", "./tsconfig.extra.json"]`. Even though we don't generate this way it is valid and users may have done this. https://github.com/microsoft/TypeScript/pull/50403

packages/nx/src/plugins/js/utils/register.ts
```diff
@@ -551,3 +552,41 @@ export function getTsNodeCompilerOptions(compilerOptions: CompilerOptions) {
...
+function resolvePathsBaseUrl(tsconfigPath: string): string {
+  const chain: { dir: string; raw: any }[] = [];
+  let current: string | null = tsconfigPath;
+  while (current) {
+    const absolute = resolve(current);
+    const dir = dirname(absolute);
+    try {
+      const raw = JSON.parse(readFileSync(absolute, 'utf-8'));
+      chain.push({ dir, raw });
+      current = raw.extends ? resolve(dir, raw.extends) : null;
```
jaysoo: Same thing for extends here.

jaysoo (approved):

PR #34976, opened 2026-03-24
cleanup(angular): remove postcss-url dependency

JackWBaker:
Remove the postcss-url from the angular generators as it does not seem to be used by ng-packagr or tailwind anymore.
https://github.com/ng-packagr/ng-packagr/pull/2736
https://v3.tailwindcss.com/docs/using-with-preprocessors#using-post-css-as-your-preprocessor

## Current Behavior
When generating an angular library postcss-url is automatically added as dependency.

## Expected Behavior
postcss-url should not be added to be dependencies as it is not used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -276,7 +276,6 @@
     "postcss": "8.4.38",
     "postcss-import": "~14.1.0",
     "postcss-preset-env": "~7.5.0",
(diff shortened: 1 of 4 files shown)


leosvelperez (approved):
Thanks!

Issue #34936, opened 2026-03-20
angular-rspack: poor local dev incremental build performance


Timebutt:
### Current Behavior

I'd like to post a follow-up issue of this [GitHub issue](https://github.com/nrwl/nx/issues/32644). The discussion in that issue is closed, but there are more performance issues lingering in the current `rspack-angular` implementation that I feel should be addressed.

I'm currently in the process of switching a big Nx project from the legacy Angular webpack builder to the new Rspack based approach. Migration proved easy, despite having multiple custom plugins and an intricate module federation setup. One important open issue caused issues however, the  `patch-package` fix in this open [GitHub PR](https://github.com/nrwl/nx/pull/34197) addressed the problem.

My problems started to appear whenever I started testing out the local development flow with the @nx/rspack-angular plugin. I noticed terrible rebuild performance and started investigating. I was very happy to see the original [GitHub issue](https://github.com/nrwl/nx/issues/32644) mentioned earlier, thinking the latest Nx version (22.4+) would address my issues but that does not seem to be the case. The performance improvements recently released might be very helpful, but I think there is still something very off in the incremental build flow that seriously hampers performance.

When running the local dev server for a few large Angular applications, I noticed Rspack was at times 2x/3x slower than Webpack when making changes! This, while the initial build for the local server can be 5x/6x faster than using webpack. The original issue also mentions people noticing the rebuild times get much longer as the application scales, which is one of the things I've noticed as well.

I can't share the codebase I'm experiencing this in, but I modified the  `ng-bundler-benchmar` repo @Coly010 created to benchmark the build times of different bundlers to include a setup that allows one to serve the application using `webpack`, `esbuild` and `rspack`. The branch `add-local-dev-configurations` of [my fork of that repo](https://github.com/Timebutt/ng-bundler-benchmark/tree/add-local-dev-configurations) is just that.

The example does not entirely reproduce the major slowdown in rebuild times I'm seeing in my enterprise grade application, but does show a few things:

- Incremental builds in `webpack` are faster than when using `rspack` (`3s` for `webpack` vs `4-6s` on my MacBook pro)
- Incremental builds for `rspack` seem to trigger recompiling twice. The original issue had [an MR](https://github.com/nrwl/nx/pull/34154) to fix this issue, but it seems the debounce time is not sufficient for this repo, or something else is going on
[...]

PR #34979, opened 2026-03-24
fix(angular-rspack): ensure rebuild chunks emitted summary accurate

Coly010:
## Current Behavior
Currently, rebuilds produce a summary table that is not accurate.
A key difference in how Rspack sets `chunks.rendered` compared to Webpack resulted in the logic for determining rebuilt chunks to be incorrect.

## Expected Behavior
Ensure the summary table for emitted chunks is accurate on rebuild

## Related Issues

It does not _fully_ solve #34936, however it should be a good first step to finding out if too many chunks are being emitted


jaysoo (approved):

jaysoo:
Thanks @Coly010 !

PR #34983, opened 2026-03-24
fix(js): generate bundled type definitions for entry points in sub-directories

lherman-vd:
When using a custom configuration for rollup (or vite) to add secondary entry points in a sub-directory of the resulting package (in cases where having all of them in root would be undesirable), the dts-bundle plugin generates invalid export statements referencing relative to the root when it should be relative to the entry file's location.

This resolves the path of the source declaration file relative to the location of the generated bundle entry file.

## Current Behavior

If I configure a bundler to generate a secondary entrypoint in a sub-directory as suggested by [rollup](https://rollupjs.org/configuration-options/#input) to have my entry file not located in the generated package's root directory, then the generated typings by the `dts-bundle` plugin point to an unknown file (`src` dir relative to a sub-directory of the generated package)

## Expected Behavior

When I use said configuration, the generated typings file refers to the source dts by pointing to the directory relative to the entry file's position. (i.e. for `entry/entry-file.d.ts` it exports types from `../src/entry/entry-file`).

Note that the goal of this fix is only to make Nx's well thought-out toolchain work with customized entry points. The goal of this fix is not to allow subdirectory exports through Nx's `additionalEntryPoints`, which would incur more changes in the toolchain.

## Related Issue(s)

None, this is a spontaneous fix from an issue I've encountered requiring me to patch the `dts-bundle` plugin. As this is due to custom configuration, I could well understand this is not meant to be supported as-is (though it makes it difficult to use with pre-configured Nx settings).

### DIFF
--- a/packages/rollup/src/plugins/with-nx/type-definitions.spec.ts
+++ b/packages/rollup/src/plugins/with-nx/type-definitions.spec.ts
@@ -68,4 +68,76 @@ describe('typeDefinitions', () => {
       });
     })();
   });
+
+  it('should emit .d.ts files with relative references to source files', () => {
+    const mockBundle = {
+      'index.js': {
+        type: 'chunk',
+        isEntry: true,
+        fileName: 'index.js',
+        facadeModuleId: '/project/src/index.ts',
+        exports: ['default'],
+      },
+      'entry-point.js': {
+        type: 'chunk',
+        isEntry: true,
+        fileName: 'entry-point.js',
+        facadeModuleId: '/project/src/entry-point1/entry-point.ts',
+        exports: [],
+      },
+      'entry-point2/entry-point.js': {
+        type: 'chunk',
+        isEntry: true,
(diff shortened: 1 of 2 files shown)


AgentEnder:
Thanks for the fix, and welcome. @leosvelperez has it. Could you rebase this on `master`? It conflicts as it stands, and the CI approval on the head commit expired before anyone ran it, so there is nothing left to approve and a rebased push is what gets it both mergeable and built.

<details>
<summary>

</summary>

> The `SUCCESS` rollup on the head commit is `graphite-app` and `netlify` alone, a bot review and a deploy preview, so nothing about the branch itself has been built.
>
> Unlinked, in both GitHub's closing references and Linear. Kept rather than closed, since the plugin emits export paths relative to the package root instead of the generated entry file, which is a narrow correctness fix.

</details>

lherman-vd:
@AgentEnder thanks, branch has been rebased.

PR #34994, opened 2026-03-24
fix(repo): add copy-assets plugin and migrate all packages from legacy-post-build

FrozenPandaz:
## Current Behavior

Each package defines a `legacy-post-build` target in `project.json` with inline asset copy configuration. Inputs are not accurately declared, leading to sandbox violations in CI. The asset globs, ignores, and outputs must be manually kept in sync across 37 packages.

## Expected Behavior

A `copy-assets` createNodesV2 plugin reads `assets.json` from each package and automatically generates the target with:
- Inputs derived from asset globs (positive patterns first, then negations)
- Outputs derived using the same dest logic as `CopyAssetsHandler`
- `dependentTasksOutputFiles` for gitignored build artifacts (jars, native binaries)
- Automatic `outDir` exclusion from asset copies

## Changes

**New infrastructure:**
- Add `copy-assets` createNodesV2 plugin in `tools/workspace-plugin`
- Add `copy-assets` executor (simplified from `legacy-post-build` — just copies assets, no package.json field manipulation)
- Extract `normalizeAssets` and `getAssetOutputPath` into reusable module in `packages/js`
- Add `copyReadme` namedInput for copy-readme build targets

**Migration (all 37 packages):**
- Create `assets.json` for every package defining what to copy
- Remove all `legacy-post-build` targets from project.json files
- Remove `legacy-post-build` target defaults from nx.json
- Remove redundant `copy-local-native` target (replaced by `.node`/`.wasm` in asset glob)

**Cleanup:**
- Remove dead config: `creator-files` globs, non-existent template files, typo'd directory names
- Use root-level `tsconfig*.json` ignore instead of recursive (so template tsconfigs in `files/` dirs are still copied)
- Replace `!(*.ts)` extglob patterns with explicit globs (extglobs don't work correctly in Nx inputs)
- Fix gradle lint: add `buildTargets: ["build-base"]` and `tslib` dependency
- Add jar outputs to maven `_package` target for correct `dependentTasksOutputFiles` resolution

**Other fixes:**
- Exclude `.swc` directories from sandbox write checks
- Enable typecheck for `angular-rspack` packages (remove `addTypecheckTarget: false`)
- Pin workspace-plugin deps to explicit versions, add `@nx/plugin` to root package.json


tools/workspace-plugin/src/executors/copy-assets/executor.ts
```diff
@@ -0,0 +1,66 @@
...
+    projectDir,
+    rootDir: workspaceRoot,
+    outputDir: outputPath,
+    assets: options.assets,
+  });
+
+  try {
+    await assetHandler.processAllAssetsOnce();
+  } catch (error) {
+    logger.error(`Error processing assets: ${error.message || error}`);
```
jaysoo, line 58: Nit: if `error: unknown` then you'd to check `instanceof Error` or that `message` is a property because using. We don't have strict mode on so it's passing typechecks.

jaysoo (approved):

PR #34996, opened 2026-03-24
fix(devkit): add timing guidance to TaskResult and remove unintentional spread

llwt:
## Current Behavior

The `TaskResult` interface has no JSDoc guidance about where to find task timing information. Users looking at the [TaskResult reference docs](https://nx.dev/docs/reference/devkit/TaskResult) don't see `startTime`/`endTime` and have no indication that timing is available on the nested `Task` object.

Additionally, the batch execution path in [`task-orchestrator.ts`](https://github.com/nrwl/nx/blob/master/packages/nx/src/tasks-runner/task-orchestrator.ts#L625) uses `...result` to spread the batch `TaskResult` (from `misc-interfaces.ts`) onto the lifecycle `TaskResult`, which unintentionally leaks `startTime`, `endTime`, and `success` properties onto the object. This only happens for batch tasks, not for regular tasks, making the behavior inconsistent.

## Expected Behavior

- The `TaskResult` JSDoc now points users to `Task.startTime` and `Task.endTime` for timing information, which will surface on the [generated reference docs](https://nx.dev/docs/reference/devkit/TaskResult)
- The `...result` spread is removed from the batch path so that only the intended `TaskResult` properties (`code`, `task`, `status`, `terminalOutput`) are set — timing stays on the `Task` object where it belongs (already copied via `task.startTime = result.startTime` / `task.endTime = result.endTime` on the lines above)

## Related Issue(s)

N/A

### DIFF
--- a/packages/nx/src/tasks-runner/life-cycle.ts
+++ b/packages/nx/src/tasks-runner/life-cycle.ts
@@ -9,6 +9,10 @@ import { TaskStatus } from './tasks-runner';
 
 /**
  * The result of a completed {@link Task}
+ *
+ * Task timing information (start and end timestamps) is available
+ * on the {@link Task} object itself via {@link Task.startTime} and
+ * {@link Task.endTime}.
  */
 export interface TaskResult {
   task: Task;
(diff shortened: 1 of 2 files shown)


AgentEnder (approved):

barbados-clemens (approved):

FrozenPandaz (changes requested):
That's because it's on `Task` so the numbers are there 👀 

https://nx.dev/docs/reference/devkit/Task#endtime

FrozenPandaz (approved):

PR #34066, opened 2026-01-09
fix(linter): allow for wildcards paths in enforce-module-boundaries rule

JesseZomer:
closed #32190

## Current Behavior

eslint crashes when tsconfig.base.json path includes a * and you have an import going to that project
## Expected Behavior
The plugin shouldn't crash and it should auto fix to a working import

## Related Issue(s)

https://github.com/nrwl/nx/issues/32190

Fixes #32190

Issue #32190, opened 2025-08-01
@nx/enforce-module-boundaries crashes while resolving relative imports with wildcards


JesseZomer:
### Current Behavior

Recently we switched from barrel files to using path wildcards (mainly because of vitest taking a long time to load everything in the barrel files) like so:
```
    "paths": {
      "@module-boundaries-bug/libA/*": ["libs/libA/src/lib/*"],
      "@module-boundaries-bug/libB/*": ["libs/libB/src/lib/*"]
    }
```
however before vscode would do a relative import of code in our libraries and the `eslint enforce-module-boundaries` rule would auto-fix this to the correct path import. But this doesn't happen anymore and I'm getting an error in the console (see failure logs)

It looks like [this line](https://github.com/nrwl/nx/blob/master/packages/eslint-plugin/src/rules/enforce-module-boundaries.ts#L341) is giving `getRelativeImportPath` the path including the wildcard, which crashes.

### Expected Behavior

Ideally it would resolve the wildcard before giving the path to `getRelativeImportPath` so that the auto-fixer can still do it's job.

### GitHub Repo

https://github.com/JesseZomer/module-boundaries-bug

### Steps to Reproduce

1. Open `libA.ts`
2. see Eslint output

### Nx Report

```shell
nx report

 NX   Report complete - copy this into the issue template

Node           : 22.16.0
OS             : darwin-arm64
Native Target  : aarch64-macos
npm            : 10.9.2

nx (global)            : 21.1.3
nx                     : 21.3.10
@nx/js                 : 21.3.10
@nx/jest               : 21.3.10
@nx/eslint             : 21.3.10
@nx/workspace          : 21.3.10
@nx/angular            : 21.3.10
@nx/devkit             : 21.3.10
@nx/eslint-plugin      : 21.3.10
@nx/module-federation  : 21.3.10
@nx/rspack             : 21.3.10
@nx/web                : 21.3.10
@nx/webpack            : 21.3.10
typescript             : 5.8.3
---------------------------------------
Community plugins:
angular-eslint : 20.1.1
---------------------------------------
Cache Usage: 0.00 B / 92.64 GB
```

### Failure Logs

```shell
Occurred while linting /Users/jesse/dev/module-boundaries-bug/libs/libA/src/lib/componentA/LibA.ts:3
Rule: "@nx/enforce-module-boundaries"
    at readFileSync (node:fs:443:20)
    at t.readFileSync (node:electron/js2c/node_init:2:10960)
[...]

Issue #30491, opened 2025-03-24
enforce-module-boundaries error when path has two sources


robertIsaac:
### Current Behavior

when there is an import with two parts e.g.

```"@fundamental-ngx/docs/core/*": ["libs/docs/core/*/index.ts", "libs/docs/core/*/src/index.ts"],```

trying to lint this file error

### Expected Behavior

the lint to work

### GitHub Repo

https://github.com/SAP/fundamental-ngx

### Steps to Reproduce

1. clone the project
2. run yarn
3. run `nx lint docs-platform`

### Nx Report

```shell
Node           : 22.14.0
OS             : win32-x64
Native Target  : x86_64-windows
yarn           : 4.7.0

nx                     : 20.6.2
lerna                  : 8.2.1
@nx/js                 : 20.6.2
@nx/jest               : 20.6.2
@nx/eslint             : 20.6.2
@nx/workspace          : 20.6.2
@nx/angular            : 20.6.2
@nx/devkit             : 20.6.2
@nx/eslint-plugin      : 20.6.2
@nx/module-federation  : 20.6.2
@nx/plugin             : 20.6.2
@nx/rspack             : 20.6.2
@nx/web                : 20.6.2
@nx/webpack            : 20.6.2
typescript             : 5.8.2
---------------------------------------
Registered Plugins:
@nx/eslint/plugin
---------------------------------------
Community plugins:
angular-eslint : 19.2.0
---------------------------------------
Local workspace plugins:
         @fundamental-ngx/nx-plugin
         @fundamental-ngx/platform
         @fundamental-ngx/core
         @fundamental-ngx/cx
```

### Failure Logs

```shell
> nx run docs-platform:lint

> eslint .


Oops! Something went wrong! :(

ESLint: 9.22.0

Error: ENOENT: no such file or directory, open 'F:\fundamental-ngx\libs\docs\platform\schema\src\index.ts'
Occurred while linting F:\fundamental-ngx\libs\docs\platform\docs-routes.ts:4
Rule: "@nx/enforce-module-boundaries"
    at readFileSync (node:fs:442:20)
    at getRelativeImportPath (F:\fundamental-ngx\node_modules\@nx\eslint-plugin\src\utils\ast-utils.js:101:47)
    at Object.fix (F:\fundamental-ngx\node_modules\@nx\eslint-plugin\src\rules\enforce-module-boundaries.js:277:98)
[...]

Issue #21889, opened 2024-02-20
@nx/enforce-module-boundaries throws no such file or directory, lstat ...


Noderoid64:
### Current Behavior

Prerequisites:

Configured "@nx/enforce-module-boundaries": "error" in the root .eslintrc.json file,

workspace structure
project
- apps
-- web
--- project.json
- nx.json

running: npx nx run web:lint

Receiving: NX   ENOENT: no such file or directory, lstat '/Work/repos/project/apps/web/src/environments/environment'

### Expected Behavior

No errors

### Steps to Reproduce

1. run npx nx run web:lint

### Nx Report

```shell
Node   : 18.12.1
   OS     : win32-x64
   npm    : 8.19.2

   nx                 : 16.9.1
   @nx/js             : 16.9.1
   @nx/jest           : 16.9.1
   @nx/linter         : 16.9.1
   @nx/workspace      : 16.9.1
   @nx/angular        : 16.9.1
   @nx/cypress        : 16.9.1
   @nx/eslint-plugin  : 16.10.0
   @nrwl/tao          : 16.9.1
   @nx/webpack        : 16.9.1
   typescript         : 5.1.6
   ---------------------------------------
   Community plugins:
   @ngneat/spectator        : 15.0.1
   @ngrx/component          : 16.0.1
   @ngrx/component-store    : 16.0.1
   @ngrx/effects            : 16.0.1
   @ngrx/router-store       : 16.0.1
   @ngrx/store              : 16.0.1
   @ngrx/store-devtools     : 16.0.1
   angular-auth-oidc-client : 16.0.1
   ng-mocks                 : 14.11.0
```

### Failure Logs

```shell
>  NX   ENOENT: no such file or directory, lstat '/Work/repos/aproject/apps/web/src/environments/environment'

   Occurred while linting C:\Work\repos\aproject\apps\web\src\app\app.component.ts:5
   Rule: "@nx/enforce-module-boundaries"

Error: ENOENT: no such file or directory, lstat '/Work/repos/aproject/apps/web/src/environments/environment'
Occurred while linting C:\Work\repos\aproject\apps\web\src\app\app.component.ts:5
Rule: "@nx/enforce-module-boundaries"
    at lstatSync (node:fs:1574:3)
    at getRelativeImportPath (C:\Work\repos\aproject\node_modules\@nx\eslint-plugin\src\utils\ast-utils.js:82:28)
[...]

PR #35007, opened 2026-03-25
fix(linter): prevent ENOENT crash in getRelativeImportPath for unresolvable paths

SharonLougheed:
I saw a bug on a bad import line, but when I fixed it, more lint errors/warnings popped up in the same file visually. But fixing that file wasn't enough, because I kept finding even more lint problems in files without that import error. Ran the lint command and saw an ENOENT crash. After a small fix, _many_ lint issues emerged. Looks like eslint doesn't catch errors you throw at it. (I wrote this paragraph myself, but I worked with _Claude Opus 4.6 Thinking_ for fixing and detailing the rest of this PR.)

## Current Behavior

When a project uses wildcard tsconfig path aliases (e.g. `@myorg/mylib/*` → `libs/mylib/src/*`), the `enforce-module-boundaries` rule's auto-fixer calls `getRelativeImportPath` with the unresolved glob path (e.g. `libs/mylib/src/*`). The function's `lstatSync` returns `null` for this path, no file extension resolves it either, and execution falls through to `readFileSync` — which throws `ENOENT: no such file or directory`.

Because ESLint does not catch errors thrown inside `fix()` functions (see [eslint/eslint#13872](https://github.com/eslint/eslint/issues/13872)), the impact depends on the context:

- **CLI** (`nx lint` / `eslint .`): the error propagates to `eslint-helpers.js` where `controller.abort()` kills the **entire lint run**, suppressing every diagnostic across all files. The ENOENT error itself is printed, but there is no indication that all other diagnostics were lost — the user sees an error about one file and reasonably assumes everything else was checked.
- **IDE extension**: the ESLint language server lints open files individually with no shared `AbortController`, so only the **crashing file's** diagnostics are lost. The crash is effectively invisible — no squiggles, no Problems panel entry, no notification (the error is only logged to the ESLint Output channel). Every other open file still shows its lint errors normally, so the linter appears to be working fine.

### Reproduction

Clone [SharonLougheed/module-boundaries-bug](https://github.com/SharonLougheed/module-boundaries-bug/) and run `nx lint libA`. The linter crashes with:

```
ENOENT: no such file or directory, open '.../libs/libB/src/lib/*'
Rule: "@nx/enforce-module-boundaries"
```

**0 lint errors are reported**, even though there are 7 real violations in the library.

## Expected Behavior

`getRelativeImportPath` should return `undefined` when the file path cannot be resolved, instead of falling through to `readFileSync`. The callers already handle `undefined` — they skip the auto-fix suggestion but still report the lint error.

After this fix, the same `nx lint libA` run reports:

```
LibA.ts:3:1  error  Projects cannot be imported by a relative or absolute path  @nx/enforce-module-boundaries
utils.ts     ...4 errors, 3 warnings (no-var, prefer-const, no-explicit-any, no-unused-vars, no-debugger)

✖ 7 problems (4 errors, 3 warnings)
```

### Relationship to #34066

#34066 (merged in v22.5.3 by @JesseZomer) resolves wildcard paths at the call site in `enforce-module-boundaries.ts`, which fixes the specific wildcard scenario. This PR adds a defensive guard inside `getRelativeImportPath` itself, so that _any_ unresolvable path — not just wildcards — returns `undefined` instead of crashing. Issues #30491 and #16716 describe non-wildcard variants of the same crash that are not addressed by #34066.

## Related Issue(s)

Fixes #35006
Related: #30491, #16716, #21889, #32190 (closed by #34066)


leosvelperez (approved):
Thanks for the contribution!

PR #35023, opened 2026-03-26
feat(release): support multiple version plans for Docker releases

aqeelat:
## Current Behavior

When using Docker releases with version plans, if multiple version plans exist for a single project (e.g., both a patch and a minor bump), the version derivation logic would not correctly handle the comparison between different bump types. Additionally, there was no defensive type checking to ensure `versionActionsVersion` was a valid string before being passed to Docker version processing, which could lead to unexpected behavior when the version data contained non-string values.

## Expected Behavior

- Docker releases correctly support multiple version plans, selecting the highest version bump when multiple plans exist
- Only string values are passed to Docker version processing, with clear errors thrown for invalid types
- The `deriveSpecifierFromVersionPlan` function correctly handles `null` current versions by using the first available plan without comparison

## Summary of Changes

1. **Added `resolveVersionActionsVersionForDocker` helper function** (`packages/nx/src/command-line/release/version/release-group-processor.ts:43-57`)
   - Validates that `newVersion` is a string before passing to Docker versioning
   - Throws a clear error message for non-string values

2. **Updated `handleDockerVersion` to accept `unknown` type** (`packages/docker/src/release/version-utils.ts:27-30`)
   - Added defensive type checking to resolve only string values
   - Non-string values are treated as `undefined`

3. **Improved `deriveSpecifierFromVersionPlan` null handling** (`packages/nx/src/command-line/release/version/deriver-specifier-from-version-plans.ts`)
   - Added early return when `currentVersion` is null (first version plan is used without comparison)
   - Replaced `gt()` with `compare() > 0` for version comparison
   - Added `resolveDerivedVersion` helper to handle semver `inc()` return types

4. **Added comprehensive tests**
   - Unit test for multiple version plans with Docker releases
   - Unit test for non-string `versionActionsVersion` handling
   - E2E test for Docker releases with multiple version plans

## Test Plan

- [x] Unit tests pass: `nx run-many -t test -p nx docker`
- [x] E2E test added for multiple version plans scenario
- [x] Prepush validation passes

### Validation (tests fail without the fixes)

Reverted the fixes from `release-group-processor.ts` and `version-utils.ts`, then ran the test suites to confirm the new tests catch the bugs:

| Test | File | Failure without fix |
|------|------|---------------------|
| `should throw a clear error when docker processing receives a non-string project version` | `release-group-processor.spec.ts` | Passes `{invalid: true}` object through to Docker instead of throwing |
| `ignores non-string versionActionsVersion values` | `docker/version-utils.spec.ts` | Returns `"[object Object]"` instead of `null` |

All tests pass again after restoring the fixes. No other tests were affected.

-----
The code was written with help from:
* GPT-5.2 Extra High via Codex
* GLM 5.0 via OpenCode

### DIFF
--- a/e2e/docker/src/docker.test.ts
+++ b/e2e/docker/src/docker.test.ts
@@ -70,6 +70,60 @@ describe('Docker E2Es', () => {
     });
 
     it(
+      'should allow releasing docker images with multiple version plans',
+      async () => {
+        const myapp = 'api';
+        createDockerApp(myapp);
+        addDockerReleaseConfiguration(myapp, {
+          skipVersionActions: false,
+          versionPlans: true,
+          versionSchemes: {
+            production: '{versionActionsVersion}',
+          },
+        });
+        addDockerPluginIfNotExists();
+
+        runCLI(`reset`);
+
+        const result = runCLI(`run ${myapp}:docker:build`);
+        expect(result.includes('Successfully ran target')).toBeTruthy();
+
+        createFile(
+          '.nx/version-plans/bump-patch.md',
+          `---
+${myapp}: patch
+---
+
+Patch docker app
+`
+        );
+        createFile(
+          '.nx/version-plans/bump-minor.md',
+          `---
+${myapp}: minor
+---
+
+Minor docker app
+`
+        );
+
+        await runCommandAsync(`git add .nx/version-plans`);
+        await runCommandAsync(
+          `git commit -m "chore: add docker version plans"`
+        );
+
+        const releaseResult = runCLI(
+          'release --dockerVersionScheme=production --yes --verbose'
+        );
+        expect(releaseResult.includes('Successfully ran target')).toBeTruthy();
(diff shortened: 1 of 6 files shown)


AgentEnder:
Thanks for the PR. It conflicts with `master` at the moment, so could you rebase it? The CI approval on the head commit expired before anyone ran it too, so there is nothing left to approve and a rebased push is what creates a fresh run. @JamesHenry gets to it as time allows once that is done.

<details>
<summary>

</summary>

> The `PENDING` rollup on the head commit is a third-party check rather than our CI. That is not a red run, so no draft is proposed.
>
> Unlinked, in both GitHub's closing references and Linear. Kept rather than closed: it is labelled a feature but behaves as a defect report, since a project with both a patch and a minor version plan gets the wrong bump today.
>
> How the version-plan specifier is derived is the part to agree on before a line by line review is worth the effort.

</details>

PR #35028, opened 2026-03-26
fix(webpack): bump postcss-loader to ^8.2.1 to eliminate transitive yaml@1.x CVE

robert-sidzinka:
## Current Behavior

`@nx/webpack` depends on `postcss-loader@^6.1.1`, which pulls in `cosmiconfig@7` → `yaml@1.x`. The `yaml@1.x` package has a known stack overflow vulnerability ([GHSA-48c2-rrv3-qjmp](https://github.com/advisories/GHSA-48c2-rrv3-qjmp)).

## Expected Behavior

By bumping `postcss-loader` to `^8.2.1`, the transitive dependency chain is eliminated entirely — `postcss-loader@8` uses `cosmiconfig@9`, which no longer depends on `yaml` at all. This is a cleaner fix than applying a `pnpm.overrides` workaround.

The upgrade is safe because:
- `postcss-loader@8` peer deps (`postcss ^7||^8`, `webpack ^5`) are unchanged
- The `implementation` option and function-based `postcssOptions` API used by `@nx/webpack` are fully supported in v8
- Nx already requires Node 18+, matching postcss-loader@8's engine requirement

## Related Issue(s)

Fixes #35025

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -50,7 +50,7 @@
     "parse5": "4.0.0",
     "postcss": "^8.4.38",
     "postcss-import": "~14.1.0",
-    "postcss-loader": "^6.1.1",
+    "postcss-loader": "^8.2.1",
     "rxjs": "^7.8.0",
     "sass": "^1.85.0",
     "sass-embedded": "^1.83.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4252,8 +4252,8 @@ importers:
         specifier: ~14.1.0
         version: 14.1.0(postcss@8.5.3)
       postcss-loader:
-        specifier: ^6.1.1
-        version: 6.2.1(postcss@8.5.3)(webpack@5.101.3)
+        specifier: ^8.2.1
+        version: 8.2.1(@rspack/core@1.6.8(@swc/helpers@0.5.18))(postcss@8.5.3)(typescript@5.9.2)(webpack@5.101.3(@swc/core@1.15.10(@swc/helpers@0.5.18))(esbuild@0.25.0)(webpack-cli@5.1.4))
       rxjs:
         specifier: ^7.8.0
         version: 7.8.2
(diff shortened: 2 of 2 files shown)


FrozenPandaz:
Thanks for flagging this! However, there are a couple of issues with this approach:

### `pnpm.overrides` don't help end users

`pnpm.overrides` only take effect at the root of the project that declares them. They don't propagate to consumers who install `@nx/js`, `@nx/webpack`, etc. So this override would only fix the vulnerability in the Nx repo's own development environment — anyone installing Nx packages in their project would still pull in `yaml@1.10.2` through the same transitive chain (`babel-plugin-macros` → `cosmiconfig@7` → `yaml@1.x`).

For users who are seeing this flagged in their own audits, they'd need to add the override in their own `package.json` themselves.

There are two transitive paths that pull in `yaml@1.x`:

### `babel-plugin-macros`

`@nx/js` → `babel-plugin-macros` → `cosmiconfig@7` → `yaml@1.x`. The package hasn't been updated in 5+ years and still pins `cosmiconfig@^7`. With no upstream fix likely, we may consider dropping `babel-plugin-macros` from our Babel preset — though that's a separate discussion.

### `postcss-loader`

`@nx/webpack` → `postcss-loader@^6.1.1` → `cosmiconfig@7` → `yaml@1.x`. This one we can fix. Bumping `postcss-loader` to `^8.2.1` would eliminate this path entirely — postcss-loader 8 uses `cosmiconfig@9` which doesn't depend on yaml at all. The peer deps are unchanged (`postcss ^7||^8`, `webpack ^5`) and Nx already requires Node 18+, so it should be a safe bump. Would you be interested in updating this PR to take that approach instead?

FrozenPandaz (changes requested):
See comment above

robert-sidzinka:
@FrozenPandaz thanks, I updated PR by upgrading `postcss-loader` package.  
It's indeed a safe bump as no code changes are required.

FrozenPandaz (approved):
Everything looks good! Thank you so much for your contribution!

Issue #35036, opened 2026-03-27
Windows: transient cmd.exe /c "net use" windows appear on file save when running Nx watchers


AntoineFra:
### Current Behavior

On Windows 10/11, transient cmd.exe windows briefly appear whenever a file is edited or saved while an Nx watcher is running (e.g., nx serve). The command observed is:

`cmd.exe /d /s /c "net use"`

The window flashes for a fraction of a second and then disappears. This occurs repeatedly during normal development and creates a visually disruptive experience.

The behavior stops immediately after terminating all Node processes and resumes once an Nx watcher is started again.

This does not appear to be related to VSCode extensions (the issue persists with extensions disabled), and the project is located on a local disk (not a network share).

**Observed Behavior**
- cmd.exe windows flash briefly on file save
- The spawned command is cmd.exe /d /s /c "net use"
- The parent process is a Node.js process started by Nx
- Killing all Node processes stops the behavior
- Restarting nx serve reproduces it
- The behavior occurs even when VSCode extensions are disabled
- The project directory is local

### Expected Behavior

Nx watchers and development tooling should run silently in the background on Windows without spawning visible console windows during normal file watch events.

### Steps to Reproduce

1. Create a new Nx workspace on Windows
2. Run a watcher
3. Open the project in VSCode
4. Modify and save a file
5. Observe transient cmd.exe windows flashing
6. Stop all Node processes -> behavior stops
7. Start nx serve again ->behavior resumes

### Nx Report

```shell
NX   Report complete - copy this into the issue template

Node           : 20.20.2
OS             : win32-x64
Native Target  : x86_64-windows
npm            : 10.8.2
daemon         : Available

nx (global)                  : 22.6.1
nx                           : 22.6.1
@nx/js                       : 22.6.1
@nx/eslint                   : 22.6.1
@nx/workspace                : 22.6.1
@nx/angular                  : 22.6.1
@nx/angular-rspack           : 22.6.1
@nx/angular-rspack-compiler  : 22.6.1
@nx/devkit                   : 22.6.1
@nx/eslint-plugin            : 22.6.1
@nx/module-federation        : 22.6.1
@nx/playwright               : 22.6.1
@nx/rspack                   : 22.6.1
@nx/vite                     : 22.6.1
@nx/vitest                   : 22.6.1
@nx/web                      : 22.6.1
@nx/webpack                  : 22.6.1
typescript                   : 5.9.3
---------------------------------------
Registered Plugins:
@nx/angular/plugin
[...]

PR #35059, opened 2026-03-28
fix(core): hide forked windows child processes

raashish1601:
## Current behavior
On Windows, several Nx watch/dev code paths still use bare `child_process.fork()` calls. Those child processes can surface visible console windows during background restarts or dependent rebuilds, which matches the flashing-window regressions reported in #35036 and the earlier watcher issue.

## This PR
- adds `windowsHide: true` to the fork-based task runner paths used for background executor work
- applies the same Windows console suppression to the fork-based dev/server helpers that restart child processes during local development
- keeps the change narrowly behavioral on Windows without changing the command arguments or process wiring

## Validation
- `git diff --check`
- attempted `npx prettier --check` on the touched files, but the repo-local config requires `prettier-plugin-tailwindcss` which is not installed in this sparse worker clone

### DIFF
--- a/packages/angular-rspack/src/lib/plugins/angular-ssr-dev-server.ts
+++ b/packages/angular-rspack/src/lib/plugins/angular-ssr-dev-server.ts
@@ -52,7 +52,9 @@ export class AngularSsrDevServer implements RspackPluginInstance {
             if (!existsSync(serverPath)) {
               await new Promise<void>((res) => setTimeout(res, 50));
             }
-            this.#devServerProcess = fork(serverPath);
+            this.#devServerProcess = fork(serverPath, [], {
+              windowsHide: true,
+            });
             this.#devServerProcess.on('spawn', () => {
               this.#wsServer.sendReload();
             });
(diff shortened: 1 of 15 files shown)


AgentEnder:
Thanks for the PR. Two separate things need a push from you: the branch conflicts with `master`, so it needs a rebase, and a real task in the affected run is failing, so this goes to draft for now. @AgentEnder has the review once that run passes. While you are in there, the `#35036` in the body is only a plain mention, so nothing closes on merge. If this is the fix for that report, make it `Fixes #35036`, and if it covers only part of it, say which part.

<details>
<summary>

</summary>

> Failing checks on the head commit:
>
> ```
> main-linux
> linux / affected --targets=lint,test,build,e2e,e2e-ci,format-native,lint-native,gradle:build-ci,vale
> Self-Healing
> ```
>
> Not a formatting check. The author notes in the body that they could not run the repo's Prettier config locally, which is worth knowing if the failure turns out to be formatting after all.
>
> Added `os: windows` because the behavior only shows up there. Confirming the fix needs a Windows machine, which is the practical constraint on whoever reviews it.

</details>

PR #35062, opened 2026-03-28
fix(vite): update vitest and plugin-react-swc versions for vite 8 compat

jaysoo:
## Current Behavior

When creating a workspace with `create-nx-workspace` using presets that install vitest (react-monorepo with vite bundler, nuxt with vitest, vue, etc.) and **npm** as the package manager, `npm install` fails with `ERESOLVE unable to resolve dependency tree`.

**Root cause:** vitest `~4.0.x` depends on `@vitest/mocker@4.0.x` which has a peer dependency on `vite: "^6.0.0 || ^7.0.0"` — it does **not** support vite 8. Since Nx now defaults to installing `vite@^8.0.0`, npm cannot satisfy `@vitest/mocker`'s peer dependency.

Additionally, `@vitejs/plugin-react-swc@^3.5.0` has a peer dependency on `vite: "^4 || ^5 || ^6 || ^7"` (no vite 8 support), and the SWC compiler path in `ensure-dependencies.ts` lacks the vite-version detection logic that the babel path already has.

## Expected Behavior

- Workspaces using vite 8 + vitest install successfully with npm
- `@vitejs/plugin-react-swc` version is selected based on the installed vite version (v4.3+ for vite 8, v3.x for older vite)

## Changes

1. **Bump `vitestV4Version`** from `~4.0.0`/`~4.0.8` to `~4.1.0` in both `@nx/vite` and `@nx/vitest` — vitest 4.1.x's dependencies support vite 8
2. **Bump `vitestV4CoverageV8Version` and `vitestV4CoverageIstanbulVersion`** to `~4.1.0` to match
3. **Update `vitePluginReactSwcVersion`** from `^3.5.0` to `^4.3.0` (first version with vite 8 peer dep support), add `vitePluginReactSwcV3Version = '^3.5.0'` for backward compat
4. **Add vite-version detection** to the SWC path in `ensure-dependencies.ts` (both `@nx/vite` and `@nx/vitest`), mirroring the existing babel path logic

## Related Issue(s)

Fixes NXC-4164


FrozenPandaz (approved):
We should update vitest in the root package.json too right?

PR #30426, opened 2025-03-19
fix(devkit): formatFiles should check for root prettier config before using prettier

JamesHenry:
## Current Behavior

`formatFiles` and `formatFilesWithPrettierIfAvailable` naively only check if `prettier` is resolvable in `node_modules`, not if it is actually configured to be used in the workspace in question.

## Expected Behavior

`formatFiles` and `formatFilesWithPrettierIfAvailable` will be safe noops if the user does not have `prettier` installed OR they have not got a root prettier config of some kind.

This PR also cleaned up two seemingly outdated/unused formatting related utils.

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/30403

PR #35089, opened 2026-03-31
feat(core): add built-in support for oxfmt formatter alongside prettier

### LINKED ISSUE #30403: Nx Release forces Prettier and has no way to disable, breaking biome builds

vsavkin:
## Current Behavior

Nx has Prettier hardwired as its only formatter across:

- CLI commands (`nx format:write` / `nx format:check`)
- Generator-level formatting (`formatFiles(tree)` in `@nx/devkit`)
- Workspace creation (prompts "Would you like to use Prettier?")

`nx init` formats almost nothing. The only exception is `nx.json` in the Turborepo flow, and that one
hardcodes Prettier. Every other flow writes through `writeJsonFile` — `JSON.stringify(data, null, 2)`
with no trailing newline — so `package.json`, `nx.json`, `.vscode/extensions.json` and the
Nest/Angular flows' generated files ignore whatever the repo is actually configured for.

## Expected Behavior

Nx supports both **oxfmt** and **Prettier**:

- Auto-detects which formatter is configured, first from config files (`.oxfmtrc.json`,
  `.prettierrc`, etc.), then from a formatter declared in the root `package.json`
- If both are configured, oxfmt wins and Nx warns once so the choice is not silent
- Workspace creation asks "Which code formatter would you like to use?" — prettier (default), oxfmt
  (experimental), none. `workspace:new` and `workspace:preset` run directly still default to `none`
- oxfmt setup writes `.oxfmtrc.json` with `singleQuote` only; line width is left at oxfmt's own
  default of 100 rather than being pinned to Prettier's 80
- No formatter configured is no longer an error: `nx format` warns and exits 0 (fixes #30403)

### Changes

**Formatter detection and dispatch** (`packages/nx/src/utils/formatters/`)

- `index.ts` — `detectFormatter()` / `detectFormatterInTree()` and the `FormatterType` union
- `oxfmt.ts`, `prettier.ts` — per-formatter implementations; `shared.ts` — the shared stdout buffer size
- The oxfmt config names are the four oxfmt itself *discovers* (`.oxfmtrc.json`, `.oxfmtrc.jsonc`,
  `oxfmt.config.ts`, `oxfmt.config.mts`). `-c` accepts more extensions, but discovery never looks for
  them, so treating one as configuration would format on options oxfmt ignores
- Two config files in one directory are reported rather than resolved: oxfmt defines no precedence
  between them and refuses to run, so picking one would format against a config the next
  `nx format:write` rejects
- A formatter that is merely *resolvable in `node_modules`* is still not treated as intent (#30426).
  A formatter **declared in the root `package.json`** is, for either formatter — both run on their
  defaults with no config file, so a declared dependency is the only signal an unconfigured
  workspace gives

**CLI** (`packages/nx/src/command-line/format/format.ts`)

- Branches on the detected formatter; oxfmt is invoked with `execFile` and an args array rather than
  through a shell
- Errors if the workspace is configured for a formatter that is not installed, instead of surfacing
  a raw `MODULE_NOT_FOUND`

**Generator formatting** (`packages/devkit/src/generators/format-files.ts`, `packages/nx/src/generators/internal-utils/format-changed-files.ts`)

- Formats in memory through oxfmt's programmatic `format()` — nothing is staged inside the
  workspace, which would race the daemon's file watcher mid-generator
- That API accepts only `FormatConfig`, so `overrides` and `ignorePatterns` (config-file-only keys it
  silently ignores) are applied here instead, along with `.editorconfig` mapping and
  `.gitignore`/`.prettierignore` filtering
- Config is resolved per file from its own directory upwards, as the CLI does, so nested
  `.oxfmtrc.json` files apply to generated code too
- A root config that exists only in the Tree is honoured for the JSON forms. A staged
  `oxfmt.config.ts`/`.mts` cannot be evaluated before it is written, so it is reported rather than
  formatted on defaults — files formatted on the wrong config would fail the next `nx format:check`
- Failures are reported per file with the path, since oxfmt's own message is context-free
  ("Unexpected token")
- The two internal helpers were renamed off "prettier" now that they route to either formatter

**`nx init` and `nx import`** (`packages/nx/src/command-line/init/implementation/format.ts`)

- Each init flow records the files it creates or modifies; init then runs the repo's own formatter
  over exactly those, and nothing else — an existing repo's other files are not init's to reformat
- Covers every flow (monorepo, npm, Turborepo, Angular CLI, Nest, `.nx` install, plugin
  configuration), so the Turborepo prettier special-case is gone
- `.gitignore` and the vendored `.nx` wrapper scripts are deliberately excluded — two of them have no
  formatter at all, and Prettier exits non-zero on a file it cannot parse
- Failure is non-fatal: the repo is already initialised, so a formatter that cannot run costs a
  warning, not the setup. `NX_SKIP_FORMAT` is honoured, matching generators
- `nx import` drains the same recorded set before each of its commit amends, so the formatting lands
  inside the commit

**Workspace creation** (`packages/create-nx-workspace/`, `packages/workspace/`)

- Prompt offers prettier / oxfmt / none, defaulting to prettier while oxfmt is pre-1.0
- Per-generator formatting is skipped in favour of one `nx format --all` after install, and a failure
  there warns rather than passing silently. That pass is skipped entirely when the preset set up no
  formatter, so `--formatter=none` no longer ends a successful creation on "No formatter configured"
- `formatter` is declared in every app generator schema and forwarded by every preset, including the
  `apps` and `npm` presets, which generate no project and previously dropped it
- The generated `.vscode/extensions.json` recommends the Prettier extension only when Prettier is the
  chosen formatter, and the oxc extension when oxfmt is

**Docs**

- New `astro-docs/src/content/docs/reference/code-formatting.mdoc` — detection order, precedence,
  switching an existing workspace, and where generator formatting still differs from
  `nx format:write`
- Formatter-specific wording removed from the KB and environment-variable pages

**Testing**

- `oxfmt` is ESM-only, so `jest.preset.js` maps it to a CommonJS wrapper that shells out to the real
  binary — formatting assertions still exercise real oxfmt
- New specs for the formatter batch, `checkWithOxfmt`/`checkWithPrettier` exit-code handling,
  formatter detection, `nx format` itself, and the init formatting pass

## Behavior changes worth calling out

Intentional, but they are changes rather than pure additions:

1. **A workspace that declares a formatter without configuring it now gets formatted by it.**
   Previously `formatFiles` and `nx format` required a config file, so a repo with `prettier` in its
   root `package.json` and no `.prettierrc` was skipped entirely. #30426 ruled out treating a
   formatter as configured because it is *resolvable in `node_modules`*, which a transitive
   dependency satisfies; a declared root dependency is a choice, and is now honoured for oxfmt and
   Prettier alike.
2. **`nx init` formats the whole of a file it modified**, not just its own additions. Running init on
   a repo that has a `.prettierrc` it never applied will reformat that `package.json` in full.
   Bounded to files init actually wrote.
3. **`--preset=npm` now honours `--formatter`.** It previously skipped the preset generator entirely
   and silently dropped the flag. When a formatter is requested it now runs the preset generator, and
   the install that follows is what puts the formatter in the workspace; a plain `--preset=npm` is
   unchanged.
4. **A generated workspace no longer always recommends the Prettier VS Code extension.** The
   template hardcoded `esbenp.prettier-vscode`, which was right when Prettier was the only option
   and wrong for an oxfmt workspace.
5. **On an oxfmt workspace, `package.json` keys get sorted.** oxfmt's `sortPackageJson` defaults to
   true and is deliberately left at its default.

## Known limitations

- `.editorconfig` is resolved once from the workspace root rather than per file, matching what the
  oxfmt CLI does. The walk continues above the workspace root until a file declares `root = true`,
  so a repo nested inside a larger checkout keeps that checkout's shared settings.
- Because the CLI resolves `.editorconfig` from the directory it runs in, running `nx format` from
  inside a subdirectory picks up that directory's file instead.
- A root `.editorconfig` that sets `indent_size` without `indent_style` is applied by generator
  formatting and ignored by `nx format:write`, so generated files can fail your own
  `nx format:check`. Set `indent_style` alongside it.

## Related Issue(s)

Fixes #30403


benpsnyder:
thanks @vsavkin -- used cc to do some analysis on Analog v3 `alpha` branch vs `beta` and you can review that here. https://github.com/nrwl/nx/issues/35111
some suggestions in there pertaining to `oxfmt`
I can do follow-up PRs to this merging if you like

pumano:
can't wait when this PR landed

pumano:
@vsavkin @FrozenPandaz feature almost done. Did you have plans to merge it to v 23?

astro-docs/src/content/docs/reference/code-formatting.mdoc
```diff
@@ -0,0 +1,109 @@
...
+description: Learn how Nx detects which code formatter your workspace uses, how oxfmt and Prettier differ, and how to switch between them.
+sidebar:
+  order: 6
+  label: Code formatting
+filter: 'type:References'
+---
+
+Nx formats code when you run `nx format:write` or `nx format:check`, at the end of most generators,
+over the files `nx init` writes, and once over a newly created workspace.
+All of them use the same formatter, which Nx detects from your workspace.
```
FrozenPandaz:
```suggestion
Nx formats code when you run `nx format:write` or `nx format:check` and at the end of most generators and migrations.
```

astro-docs/src/content/docs/reference/code-formatting.mdoc
```diff
@@ -0,0 +1,109 @@
...
+
+## How Nx picks a formatter
+
+Detection runs in this order, and the first match wins:
+
+1. An oxfmt config file exists at the workspace root.
+2. A Prettier config file exists at the workspace root, or `package.json` has a `prettier` key.
+3. `package.json` declares an `oxfmt` dependency.
+
+An installed Prettier is not enough on its own, because Prettier is a common transitive dependency.
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/reference/code-formatting.mdoc
```diff
@@ -0,0 +1,109 @@
...
+
+If nothing matches, `nx format:write` and `nx format:check` warn and exit `0`, and generators skip
+formatting.
+That is what keeps Nx from reformatting a workspace that uses Biome, dprint, or anything else.
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/reference/code-formatting.mdoc
```diff
@@ -0,0 +1,109 @@
...
+
+The oxfmt config files Nx looks for:
+
+- `.oxfmtrc.json`
+- `.oxfmtrc.jsonc`
+- `oxfmt.config.ts`
+- `oxfmt.config.mts`
+
+Keep one per directory.
+Given two, oxfmt fails with `Both '.oxfmtrc.json' and '.oxfmtrc.jsonc' found` rather than choosing.
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/reference/code-formatting.mdoc
```diff
@@ -0,0 +1,109 @@
...
+```
+
+```json {% title=".oxfmtrc.json" %}
+{
+  "singleQuote": true,
+  "printWidth": 80
+}
+```
+
+oxfmt's defaults differ from Prettier's, so set these two explicitly if you want your existing
```
FrozenPandaz:
```suggestion
```

packages/create-nx-workspace/src/create-workspace.ts
```diff
@@ -242,6 +249,65 @@ export async function createWorkspace<T extends CreateWorkspaceOptions>(
     await setupCI(directory, ciProvider, packageManager);
   }
 
+  // Format once now that dependencies are on disk. `@nx/workspace:new` formats
```
FrozenPandaz: There's way too many comments here.. please trim these.

packages/js/src/generators/library/library.ts
```diff
@@ -102,8 +102,10 @@ export async function libraryGeneratorInternal(
       tsConfigName: schema.rootProject ? 'tsconfig.json' : 'tsconfig.base.json',
       addTsConfigBase: true,
       addTsPlugin,
-      // In the new setup, Prettier is prompted for and installed during `create-nx-workspace`.
-      formatter: isUsingTsSolutionSetup(tree) ? 'none' : 'prettier',
+      // `formatter` rides along on the spread above. It used to be pinned to
+      // prettier here, which silently overrode the formatter the caller chose -
+      // `create-nx-workspace --preset=ts-standalone --formatter=oxfmt` came out
+      // configured for prettier. When it is absent, js:init picks the default.
```
FrozenPandaz:
```suggestion
```

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #35097, opened 2026-03-31
fix(repo): narrow copy-assets outputs to prevent overlap with build-base

leosvelperez:
## Current Behavior

Broad globs in `assets.json` (`**/*.json`, `**/*.js`, `**/*.d.ts`) cause `copy-assets` targets to claim cache ownership over files also produced by `build-base` (tsc). Even though a recent change reordered `copy-assets` to run before `build-base` (reducing the likelihood of the race condition), the underlying task ownership model is still broken — both targets claim overlapping files in their output patterns.

## Expected Behavior

Each target exclusively owns its output files. `copy-assets` only claims non-tsc assets (templates, type declarations, native artifacts), and `build-base` owns all compiler outputs.

## Changes

**37 `assets.json` files** — replaced broad extension globs with narrow, destination-safe patterns: template dirs (`**/files/**`), schema type declarations (`src/**/schema.d.ts`), non-tsc extensions (`.jar`, `.node`, `.wasm`, `.md`), and explicit file paths for package-specific assets.

**30 `tsconfig.lib.json` files** — removed `**/*.json` from `include` so tsc only compiles TypeScript. JSON files are now handled by copy-assets with explicit entries (`@(package|executors|generators|migrations).json`, `src/**/schema.json`).

**Exception:** jest and vite use `import('./schema.json')` which requires JSON in tsconfig scope with `composite: true`. These keep `src/**/schema.json` in tsconfig.

**vite** — excludes `test-utils.ts` from lib build (only used by specs) and includes it in `tsconfig.spec.json`.

### DIFF
--- a/packages/jest/tsconfig.lib.json
+++ b/packages/jest/tsconfig.lib.json
@@ -7,7 +7,7 @@
     "tsBuildInfoFile": "../../dist/packages/jest/tsconfig.tsbuildinfo"
   },
   "exclude": ["**/*.spec.ts", "**/*.test.ts", "jest.config.ts"],
-  "include": ["**/*.ts", "**/*.json"],
+  "include": ["**/*.ts", "src/**/schema.json"],
   "references": [
     {
       "path": "../js/tsconfig.lib.json"
--- a/packages/angular/assets.json
+++ b/packages/angular/assets.json
@@ -8,15 +8,7 @@
       "glob": "**/files/**/.gitkeep"
     },
     {
-      "glob": "**/*.json",
-      "ignore": ["tsconfig*.json", "project.json", ".eslintrc.json"]
-    },
-    {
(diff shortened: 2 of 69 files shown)


packages/jest/tsconfig.lib.json
```diff
@@ -7,7 +7,7 @@
     "tsBuildInfoFile": "../../dist/packages/jest/tsconfig.tsbuildinfo"
   },
   "exclude": ["**/*.spec.ts", "**/*.test.ts", "jest.config.ts"],
-  "include": ["**/*.ts", "**/*.json"],
+  "include": ["**/*.ts", "src/**/schema.json"],
```
FrozenPandaz, line 10: Why is this one different?

FrozenPandaz (approved):

PR #35102, opened 2026-03-31
chore(repo): make build-base depend on copy-assets instead of reverse

FrozenPandaz:
## Current Behavior

`copy-assets` depends on `build-base`, which means it has to wait for the entire `build-native` → `build-base` chain to finish before it can start — even though asset copying doesn't need compiled output.

## Expected Behavior

`build-base` depends on `copy-assets` instead. This lets `copy-assets` start immediately (in parallel with `build-native` and `^build-base`) rather than waiting for them to complete first.

image

### Changes

- Removed `dependsOn: ['build-base']` from the `copy-assets` target generated by `copy-assets-plugin.ts`
- Added `copy-assets` to `build-base.dependsOn` in `nx.json` and project-level overrides (`gradle`)
- Removed now-transitive `copy-assets` references from `build.dependsOn` in `nx.json`, `packages/nx`, `packages/angular`, and `packages/gradle`
- Removed `build-base` from `copy-assets.dependsOn` in `packages/nx` and `packages/dotnet` project overrides

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -77,7 +77,6 @@
       "dependsOn": [
         "^build",
         "typecheck",
-        "copy-assets",
         "build-base",
         "build-native"
       ],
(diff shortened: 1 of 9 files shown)


leosvelperez (approved):

FrozenPandaz:
These failing checks are unrelated

PR #35109, opened 2026-03-31
fix(core): reduce published nx package size with files allowlist

FrozenPandaz:
## Current Behavior

The `nx` npm package includes unnecessary files in the published artifact:
- ~137 Rust source files (`.rs`) from `src/native/`
- Test fixtures, snapshots, and other dev-only files
- Duplicate source directories (`bin/`, `plugins/`, `schemas/`, etc.) alongside their compiled `dist/` equivalents

The `.npmignore` blocklist approach missed several file types, and the CI `.node` file cleanup (`find ./dist`) no longer works because the `nx` package now builds to `packages/nx/dist/` instead of `dist/packages/nx/`.

## Expected Behavior

- Only `dist/` and essential root JSON files (`migrations.json`, `executors.json`, `generators.json`) are published
- No Rust source, test fixtures, snapshots, or duplicate source dirs in the published package
- Native type declarations (`src/native/index.d.ts`) are copied to `dist/` via `assets.json` so all exports reference `dist/` consistently
- CI correctly removes `.node` files from `packages/nx/dist/` before publishing to npm

## Related Issue(s)

N/A — discovered during package audit.


scripts/nx-release.ts
```diff
@@ -42,7 +42,7 @@ const VALID_AUTHORS_FOR_LATEST = [
   if (!options.local && process.env.GITHUB_ACTIONS) {
     // Delete all .node files that were built during the previous steps
     // Always run before the artifacts step because we still need the .node files for native-packages
-    execSync('find ./dist -name "*.node" -delete', {
+    execSync('find ./dist ./packages/nx/dist -name "*.node" -delete', {
```
leosvelperez, line 45: There are a couple more things we should either delete or ignore in the `package.json`'s `files`:

- `dist/native-packages/`
- `dist/tsconfig.tsbuildinfo`

packages/nx/package.json
```diff
@@ -9,6 +9,13 @@
     "url": "https://github.com/nrwl/nx.git",
     "directory": "packages/nx"
   },
+  "files": [
+    "dist",
+    "migrations.json",
+    "executors.json",
+    "generators.json",
+    "README.md"
```
leosvelperez: nit: `README.md` is not needed. npm auto-includes README.md (and package.json, LICENSE, CHANGELOG) regardless of the `files` field.
```suggestion
    "generators.json"
```

scripts/nx-release.ts
FrozenPandaz: I made `files` more specific.

leosvelperez (approved):

PR #35118, opened 2026-04-01
feat(webpack): update webpack-cli args --config-node-env

thomasgosse:
## Current Behavior
The @nx/webpack plugin uses webpack-cli v5, which provides the `--node-env` flag to set `process.env.NODE_ENV` during build and serve. It is not compatible with [the new version of webpack-cli](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%407.0.0), which changed the `--node-env` into `--config-node-env`.

## Expected Behavior
The `@nx/webpack` plugin now targets webpack-cli v7 for newly generated projects. In this version, the `--node-env` flag has been renamed to `--config-node-env`. All usages in the webpack plugin have been updated to use the new flag name.

The root `package.json` still pins webpack-cli v5 because the graph-client app relies on the remote `@nx/webpack` package, which is not compatible with webpack-cli v7.

## Changes
- Updated `webpackCliVersion` from ^5.1.4 to ^7.0.2 in `packages/webpack/src/utils/versions.ts`
- Replaced all `--node-env` occurrences with `--config-node-env` in `packages/webpack/src/plugins/plugin.ts`

## BREAKING CHANGES

I think the next version of `@nx/webpack` should be breaking because it won't work with webpack-cli < v7.
`webpack-cli` v7 requires Node 20.9.0 or later, which seems compatible with the minimum Node version supported by Nx.


packages/webpack/src/plugins/plugin.ts
```diff
@@ -164,7 +164,7 @@ async function createWebpackTargets(
 
   targets[options.buildTargetName] = {
     command: `webpack-cli build`,
-    options: { cwd: projectRoot, args: ['--node-env=production'] },
+    options: { cwd: projectRoot, args: ['--config-node-env=production'] },
```
jaysoo: This would break existing webpack-cli@5 users, we should detect the version and infer the right flag.

packages/webpack/src/utils/versions.ts
```diff
@@ -3,7 +3,7 @@ export const nxVersion = require('../../package.json').version;
 export const swcLoaderVersion = '0.1.15';
 export const tsLibVersion = '^2.3.0';
 
-export const webpackCliVersion = '^5.1.4';
+export const webpackCliVersion = '^7.0.2';
```
jaysoo, line 6: A migration to update existing workspace would be nice as long as there aren't breaking changes -- if there are we can provide a migration.

jaysoo (commented):
I would look into webpack-cli@7 changes, if there aren't any breaking changes, updating existing workpsace through a migration would be great. Otherwise, we still need to support older CLI versions, so the plugin needs to check version and fork the flag logic.

I appreciate the PR!

packages/webpack/src/plugins/plugin.spec.ts
```diff
@@ -114,7 +123,173 @@ describe('@nx/webpack/plugin', () => {
...
+                          },
+                        },
+                      },
+                      "technologies": [
+                        "webpack",
+                      ],
+                    },
+                    "options": {
+                      "args": [
+                        "--config-node-env=production",
```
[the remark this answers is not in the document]
thomasgosse: Good catch, the test was not working. I changed where the webpack-cli version is used and how I mock it in test, it should now be good!

thomasgosse:
Thanks for the review @jaysoo!
I've pushed some updates to address your feedback:
- Modified the flag behavior to handle different webpack-cli versions
- Added test coverage for these changes

Let me know if there's anything else that needs adjustment.

> A migration to update existing workspace would be nice as long as there aren't breaking changes -- if there are we can provide a migration.
I've verified compatibility, and everything looks good:
- webpack-dev-server v4+ (the flag was dropped in [webpack-cli@v6](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%406.0.0))
- Node version 20.0.9+

No breaking changes detected for this repo. However, I'm not familiar with the migration system in Nx, so I'd appreciate any guidance on whether a migration is needed here 😁

PR #35141, opened 2026-04-01
fix(js): support auto mode for non-pnpm lock files in affected detection

### LINKED ISSUE #35173: `@nx/js` plugin `projectsAffectedByDependencyUpdates` does not account for `"resolutions"` in package.json
### Current Behavior

When using **yarn** as the package manager + the `@nx/js` plugin with `"projectsAffectedByDependencyUpdates": "auto"`, adding or modifying a package entry in the root `package.json`'s `resolutions` field — and running `yarn install` which updates `yarn.lock` — results in **zero projects being marked as affected** by `nx affected`.

This seems to be caused by two independent gaps in the affected calculation pipeline for this plugin, both in `node_modules/nx/src/plugins/js/project-graph/affected/`:

**Gap 1 — `npm-packages.js` (`getTouchedNpmPackages`):**
The function that detects `package.json` changes hardcodes a check for only `dependencies` and `devDependencies`:

```js
if (
  isJsonChange(c) &&
  (c.path[0] === 'dependencies' || c.path[0] === 'devDependencies') &&
  c.path.length === 2
)
```

The `resolutions` field is never checked. Changes to it are silently ignored regardless of package manager.

**Gap 2 — `lock-file-changes.js` (`getTouchedProjectsFromLockFile`) with `"auto"` mode:**
The `auto` mode's logic for scoping affected projects to specific packages is pnpm-only, which is not documented. It reads the `importers` section of the pnpm lock file format. For yarn, the inner block is skipped entirely and an empty array is returned — even though `yarn.lock` has changed:

```js
const getProjectPathsAffectedByDependencyUpdates = (changedLockFile) => {
    const changedProjectPaths = new Set();
    if (PNPM_LOCK_FILES.includes(changedLockFile.file)) { // ← never true for yarn.lock
        // ...pnpm-specific importers parsing...
    }
    return Array.from(changedProjectPaths); // always [] for yarn
};
```

Because `auto` returns early with `[]`, the fallback `"all"` behavior (which would correctly mark all projects affected on any lock file change) is **never reached**.

### Expected Behavior

When a package is added or modified under `resolutions` in the root `package.json` and `yarn.lock` is subsequently updated:

1. `getTouchedNpmPackages` should inspect the `resolutions` field the same way it inspects `dependencies` and `devDependencies`, marking the changed npm node(s) as touched.
2. The existing recursive reversed-graph traversal in `affected-project-graph.js` (`addAffectedNodes`) should then correctly propagate through npm-to-npm edges (already built from the yarn lock file by `yarn-parser.js`) to identify all workspace projects transitively affected — even if the resolved package is not a **direct** dependency of those projects.
3. When `projectsAffectedByDependencyUpdates` is set to `"auto"`, a `yarn.lock` change with no corresponding `package.json` `resolutions`/`dependencies` change should fall back to marking all projects affected (matching the `"all"` behavior), rather than silently returning `[]`.

### GitHub Repo

https://github.com/dallas-scott/nx-js-resolutions-affected-issue

### Steps to Reproduce

> **Baseline:** the `main` branch of the [linked repo](https://github.com/dallas-scott/nx-js-resolutions-affected-issue) serves as the comparison base — no separate baseline commit needed.

### 1. Add a `resolutions` entry to the root `package.json`

```json
{
  "resolutions": {
    "semver": "7.6.3"
  }
}
```

### 2. *(Optional)* Update the lock file

```sh
yarn install
```

> Updating the lock file is **not required** to trigger the bug — the `package.json` change alone
> is sufficient, since most of the yarn-related affected detection reads from `package.json`
> rather than `yarn.lock`. Including this step only demonstrates that even a lock file change
> does not rescue the `"auto"` path.

### Nx Report

```shell
Node           : 22.14.0
OS             : darwin-arm64
yarn           : 3.6.1

nx (global)            : 21.6.10
nx                     : 21.3.10
@nx/js                 : 21.3.10
@nx/jest               : 21.3.10
@nx/linter             : 19.8.6
@nx/eslint             : 21.3.10
@nx/workspace          : 21.3.10
@nx/cypress            : 21.3.10
@nx/devkit             : 21.3.10
@nx/eslint-plugin      : 21.3.10
@nx/module-federation  : 21.3.10
@nx/node               : 21.3.10
@nx/plugin             : 21.3.10
@nx/react              : 21.3.10
@nx/remix              : 19.8.6
@nx/storybook          : 21.3.10
@nx/vite               : 21.3.10
@nx/web                : 21.3.10
nx-cloud               : 19.1.0
@nrwl/nx-cloud         : 19.1.0
Registered Plugins:
@bestowinc/nx-playwright
@bestowinc/nx-playwright
@bestowinc/nx-vitest-sharder
@bestowinc/nx-vitest-sharder
@nx/storybook/plugin
Community plugins:
@bestowinc/eslint-plugin : 0.0.1
@jscutlery/semver        : 5.7.1
@nrwl/remix              : 19.8.6
Local workspace plugins:
@bestowinc/nx-vitest-sharder
@bestowinc/microsite-plugin
@bestowinc/tinkertoy-utils
@bestowinc/library-plugin
@bestowinc/eslint-plugin
@bestowinc/chromatic
Cache Usage: 22.67 KB / 92.64 GB
  - @nx/linter@19.8.6
  - @nx/remix@19.8.6
```

### Failure Logs

```shell
N/A
```

### Package Manager Version

yarn@3.6.1

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

- **`resolutions` is the idiomatic yarn mechanism** for overriding transitive dependency versions (e.g., forcing a patched version of a nested package for a security fix). It is not standard practice to duplicate such entries into `dependencies`/`devDependencies`, as those fields imply direct usage by the project's own source code.
- The transitive graph traversal infrastructure already works correctly. `yarn-parser.js` (`getYarnLockfileDependencies`) builds npm-to-npm dependency edges from `yarn.lock`, and `addAffectedNodes` in `affected-project-graph.js` recursively traverses the reversed graph. The only missing piece is the initial "touch" signal for the changed npm node.
- **Workarounds** (until fixed):
  - Set `"projectsAffectedByDependencyUpdates": "all"` in `nx.json` — any `yarn.lock` change marks all projects affected.
  - Add `"{workspaceRoot}/package.json"` to the `sharedGlobals` named input — any root `package.json` change invalidates all project caches.
- **Affected versions:** Reproduced on Nx `v21.x` with yarn@3.6.1. The pnpm `"auto"` behavior is unaffected by this issue.

llwt:
## Current Behavior

The `projectsAffectedByDependencyUpdates: "auto"` setting only works for pnpm lock files. For npm (`package-lock.json`), yarn (`yarn.lock`), and bun (`bun.lock`/`bun.lockb`), auto mode silently returns an empty array -- meaning lockfile-only changes (e.g. `npm audit fix`, transitive dependency updates) produce zero affected projects.

## Expected Behavior

Auto mode detects affected projects from lock file changes for all supported package managers:

- **pnpm** (`pnpm-lock.yaml`): Inspects the `importers` section to determine exactly which workspace projects had dependency changes (unchanged).
- **npm** (`package-lock.json`): Inspects the `packages` entries to determine which workspace projects had dependency changes.
- **bun** (`bun.lock`): Inspects the `workspaces` section to determine which workspace projects had dependency changes.
- **yarn** (`yarn.lock`): The lock file is a flat list with no per-project structure, so all projects are marked as affected when any dependency changes.
- **Binary lock files** (`bun.lockb`) / **WholeFileChange**: Cannot be parsed for granular changes, so all projects are marked as affected.

The implementation uses a `LOCK_FILE_RESOLVERS` map with a `SupportedLockFile` type guard so that adding a new lock file format requires adding exactly one entry -- no separate lists to keep in sync.

## Related Issue(s)

Follow-up to #34937 which documented the pnpm-only limitation.

Fixes NXC-4185
Fixes https://github.com/nrwl/nx/issues/35173


packages/nx/src/plugins/js/project-graph/affected/lock-file-changes.ts
```diff
@@ -51,27 +69,93 @@ export const getTouchedProjectsFromLockFile: TouchedProjectLocator<
...
+
+  // If any change is a WholeFileChange we cannot determine which
+  // specific projects are affected, so signal "all projects".
+  if (changes.some((c) => !isJsonChange(c))) {
+    return null;
   }
-  return Array.from(changedProjectPaths);
+
+  if (!isSupportedLockFile(changedLockFile.file)) {
+    // TODO: should we log a warning here?
```
llwt: In theory, we should never hit this branch.

@AgentEnder What's idiomatic in the nx codebase for handling "should never hit this path but theoretically possible" cases?

Log warning? Throw?

AgentEnder: In this case id log a warning and return an empty array i think. Theres not a set pattern.

llwt: Will do.

Empty array, or `null` (signifying "all")?

AgentEnder (approved):

PR #35163, opened 2026-04-02
fix(core): replace LGPL-licensed @ltd/j-toml with BSD-3-Clause smol-toml

FrozenPandaz:
## Current Behavior

Nx 22.6.0 introduced `@ltd/j-toml` as a dependency, which uses the LGPL-3.0 license. This is a copyleft license that causes compliance issues for enterprise users whose policies prohibit LGPL dependencies.

## Expected Behavior

Replace `@ltd/j-toml` with `smol-toml` (BSD-3-Clause license), which is permissive and enterprise-friendly. `smol-toml` is actively maintained, supports the full TOML v1.0 spec, and provides both CJS and ESM exports.

### Changes
- Swapped `@ltd/j-toml` for `smol-toml` v1.6.1 in root and `packages/nx/package.json`
- Updated imports in `set-up-ai-agents.ts` and `test-utils.ts`
- Removed `@ltd/j-toml`-specific API calls (`TOML.Section()`, `TOML.inline()`, stringify options)
- Updated test assertions and snapshots to match `smol-toml` output format (double quotes, section-style tables)

## Related Issue(s)

Fixes #35156

### DIFF
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -50,7 +50,7 @@
   },
   "homepage": "https://nx.dev",
   "dependencies": {
-    "@ltd/j-toml": "^1.38.0",
+    "smol-toml": "1.6.1",
     "@napi-rs/wasm-runtime": "0.2.4",
     "@yarnpkg/lockfile": "^1.1.0",
     "@yarnpkg/parsers": "3.0.2",
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "@jest/reporters": "catalog:jest",
     "@jest/test-result": "catalog:jest",
     "@jest/types": "catalog:jest",
-    "@ltd/j-toml": "^1.38.0",
+    "smol-toml": "1.6.1",
     "@module-federation/enhanced": "2.1.0",
     "@module-federation/sdk": "^2.1.0",
     "@monodon/rust": "2.3.0",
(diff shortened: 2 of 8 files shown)


AgentEnder (approved):

packages/nx/package.json
```diff
@@ -50,7 +50,7 @@
   },
   "homepage": "https://nx.dev",
   "dependencies": {
-    "@ltd/j-toml": "^1.38.0",
+    "@iarna/toml": "2.2.5",
```
pmaieref: Feel free to ignore this comment, as I am not really familiar with the "toml" format. But I checked out this alternative library and it seems to be (especially v.2.2.5) based on a pretty old version of the toml-spec and not really maintained (last release 6 years ago).

Google gave me https://www.npmjs.com/package/toml as an alternative and it was recently updated to support v1.1.0 of the toml-spec, but the new version seems to only explicitly support spec 1.1.0, so YMMV. There are some more libraries available but I did not check them.

I do not know your constraints in detail or what toml-spec you target, so "@iarna/toml" might as well be a good fit. Just wanted to share my observation.

JiggyJinjo: [toml](https://www.npmjs.com/package/toml) package doesn't have stringify() so it would not be a direct replacement. 

I think  [smol-toml ](https://www.npmjs.com/package/smol-toml) would be better. It's the most downloaded (10+M weekly) and is actively maintained (last release 15 days ago).

[@iarna/toml](https://www.npmjs.com/package/@iarna/toml) is compliant only up to  v1.0.0-rc.1 and hasn't seen an update since 2020.

FrozenPandaz: Good point :) Switched to `smol-toml` I see you've opened another PR so let's work together on that one ;)

FrozenPandaz:
Closing in favor of #35188 which takes the same approach. Left a review comment there about pinning the version.

PR #35172, opened 2026-04-03
fix(core): optimize warm cache performance for task execution

### LINKED ISSUE #31067: .env variables are not considered for inputs caching
### Current Behavior

I'd like to reopen the discussion around this issue: [https://github.com/nrwl/nx/issues/20216](https://github.com/nrwl/nx/issues/20216)

It appears that environment variables defined in `.env` files are not respected by the cache input configuration.

**Example setup:**

**apps/my-app/.env**

```env
VERSION=0.1.0
```

**nx.json**

```json
{
  "namedInputs": {
    "default": [
      "{projectRoot}/**/*",
      "{projectRoot}/.env",          // <-- Ignored, likely due to being in `.gitignore`
      {
        "env": "VERSION"             // <-- Also ignored — why?
      },
      "sharedGlobals"
    ],
    "production": ["default"],
    "sharedGlobals": []
  }
}
```

Despite explicitly including the `VERSION` variable in `namedInputs`, they appear to be ignored by the caching system.

### Expected Behavior

Modifying a variable in the `.env` file should invalidate the cache if the task declares that environment variable as an input.

### GitHub Repo

https://github.com/jogelin/issue-dotenv-vars-not-available-for-inputs

### Steps to Reproduce

1. Clone Repo & npm install
2. Create a `.env` file at `apps/my-app` containing
  ```
  VERSION=0.1.0
  ```
3. Run
  ```
  npx nx build my-app (First run, NOT CACHED)
  ```
4. Run Again
  ```
  npx nx build my-app (Second run, CACHED)
  ```
5. Edit `apps/my-app/.env`
  ```
  VERSION=0.2.0
  ```
6. Run Again
  ```
  npx nx build my-app (Still CACHED even is VERSION is modified)
  ```

### Nx Report

```shell
~/d/tm/issue-dotenv-vars-not-available-for-inputs  main  npx nx report                      1 ✘  system 

Node           : 22.15.0
OS             : darwin-arm64
npm            : 10.9.2

nx                     : 20.8.1
@nx/js                 : 20.8.1
@nx/eslint             : 20.8.1
@nx/workspace          : 20.8.1
@nx/devkit             : 20.8.1
@nx/module-federation  : 20.8.1
@nx/next               : 20.8.1
@nx/react              : 20.8.1
@nx/web                : 20.8.1
@nx/webpack            : 20.8.1
Registered Plugins:
@nx/js/typescript
@nx/next/plugin
Cache Usage: 647.01 MB / 92.64 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

FrozenPandaz:
## Current Behavior

Running cached tasks in a large workspace takes much longer than the work actually warrants. The task orchestrator hashes, cache-checks, schedules, and reports each task individually:

- Per-task JS→Rust→SQLite cache lookups (~N boundary crossings).
- Per-task daemon IPC calls for recording/matching output hashes (~2N sequential round-trips).
- Per-task filesystem scans for output expansion.
- `TasksSchedule` re-sorts the full schedule array on every insert (O(n²·log n)).
- The coordinator awaits each worker one at a time via a single-task dispatch path.

On a 1,110-project benchmark workspace, this accumulates into multi-second overhead even when every task is a cache hit with nothing to do.

## Expected Behavior

Warm cache runs should resolve quickly. Every hot-path operation that can be batched — cache lookups, daemon calls, scheduling, output-hash tracking, filesystem scans — is batched, and the coordinator dispatches discrete workers in parallel rather than sequentially.

### Rust native

- **`NxCache.get_batch`** (`cache.rs`) — single `UPDATE … WHERE hash IN (…) RETURNING` with Rayon-parallel terminal-output reads replaces N individual round-trips. Uses `rarray` for the `IN` clause, groups the query / build stages as separate helpers, and collects rows straight into a `HashMap`.
- **`get_files_for_outputs_batch`** — Rayon-parallel filesystem scanning for cached-output expansion. Drop the old singular `get_files_for_outputs` napi export now that every caller goes through the batch path.

### Daemon output tracking

- Replace the non-batch `recordOutputsHash` / `outputsHashesMatch` chain (client → server handlers → outputs-tracking helpers) with `*Batch` equivalents. The single-entry path was dead after the orchestrator routed everything through the batch methods.
- Short-circuit `outputsHashesMatchBatch` when the daemon has no recorded hashes — avoids an unnecessary Rayon filesystem scan right after `nx reset`.
- Skip recording for `local-cache-kept-existing`: the daemon already has the right hash.

### Task orchestrator

- **Coordinator loop** rewrite: bulk-resolve all cache hits up-front, batch-hash remaining unhashed tasks before any per-task lifecycle fires, then dispatch cache-miss workers concurrently up to `parallelism`. Tracks in-flight workers as a `Set<Promise<void>>` instead of a counter, and the dispatch is extracted to `dispatchDiscreteWorker` + `handleDiscreteWorkerFailure` instead of an inline fire-and-forget IIFE.
- **Split cache check from task execution**: `resolveCachedTasksBulk` reports hits/misses without touching the execution path, so the coordinator can batch lifecycle calls for the hit set and only dispatch workers for misses.
- **Route `applyCachedResults` through `DbCache.getBatch`** — one daemon call per cycle instead of one per task.
- **Group slots**: `closeGroup` can overflow without throwing when all slots are claimed (parallelism gating is enforced elsewhere), and every task keeps a single `groupId` through `runDiscreteTasks`.
- **Drop the dead single-entry fallback paths** in the orchestrator (`shouldCopyOutputsFromCache`, `recordOutputsHash`, and the `processTask will hash individually` try/catch around `hashTasks`).
- **`getExecutorForTask` takes a `projects` record directly** instead of re-deriving it from the project graph per call via a module-level `WeakMap`. Callers compute the record once.

### Task scheduler

- **Parallelism gating + sort stability** — batch scheduling collects all schedulable roots, pushes them in one pass, and sorts the array once per cycle instead of after every insert.

### Repo / infra

- `benchmarks/package.json`: run benches through `pnpm exec nx` so hyperfine sub-shells always hit the workspace-local CLI.
- `fix(core): honor NX_NO_CLOUD / neverConnectToCloud in runner selection` (`run-command.ts`) — surfaced while unblocking the benchmark CI: with an ambient `NX_CLOUD_AUTH_TOKEN` on the CI agent, `getTasksRunnerPath` still routed through the cloud shell even when `NX_NO_CLOUD=true` was set. The cloud client's light-client require bridge then loaded the default tasks runner from the parent workspace's `node_modules/nx` (a published version), creating a cross-version API mismatch. The guard now short-circuits runner selection to the default path when cloud is explicitly opted out of.
- Test updates for the new `TasksSchedule(projectGraph, projects, taskGraph, options)` signature.

## Benchmark Results

Measured locally against the `benchmarks/` workspace (1,110 projects) with the `bench:*` scripts. Baseline is committed in `benchmarks/baseline.json`.

| Benchmark | Goal | Baseline | Current | vs Goal | vs Baseline |
  |---|---|---|---|---|---|
  | version | 50ms | 348ms | 333ms | +566% | -4% |
  | show-projects | 100ms | 710ms | 647ms | +547% | -9% |
  | cat-warm | 300ms | 2.44s | 1.17s | +290% | -52% |
  | copy-warm | 1.00s | 6.85s | 1.35s | +35% | -80% |
  | build-warm | 5.00s | 17.75s | 1.44s | -71% | -92% |

## Related Issue(s)

Fixes #31067

No specific issue — performance-focused, driven by benchmark wall time on large workspaces.

### DIFF
--- a/packages/nx/src/hasher/hash-task.ts
+++ b/packages/nx/src/hasher/hash-task.ts
@@ -220,16 +220,21 @@ export async function hashTasks(
     // the built-in batch hasher always produces a hash, but user-written
     // custom hashers are untrusted and an empty/undefined hash would
     // violate the task_details schema downstream.
-    const hashedTasks = tasks.filter((task) => task.hash);
+    const hashedTasks = [];
+    for (const t of tasks) {
+      if (!t.hash) {
+        continue;
+      }
+
+      hashedTasks.push({
+        hash: t.hash,
+        project: t.target.project,
+        target: t.target.target,
+        configuration: t.target.configuration,
+      });
+    }
     if (hashedTasks.length > 0) {
-      taskDetails.recordTaskDetails(
-        hashedTasks.map((task) => ({
-          hash: task.hash,
-          project: task.target.project,
-          target: task.target.target,
-          configuration: task.target.configuration,
-        }))
-      );
+      taskDetails.recordTaskDetails(hashedTasks);
     }
   }
 
--- a/packages/nx/src/hasher/hash-task.ts
+++ b/packages/nx/src/hasher/hash-task.ts
@@ -214,14 +214,17 @@ export async function hashTasks(
   await Promise.all([...customHasherPromises, batchHashPromise]);
 
   if (taskDetails?.recordTaskDetails) {
-    taskDetails.recordTaskDetails(
-      tasks.map((task) => ({
-        hash: task.hash,
-        project: task.target.project,
-        target: task.target.target,
-        configuration: task.target.configuration,
-      }))
-    );
+    const hashedTasks = tasks.filter((task) => task.hash);
+    if (hashedTasks.length > 0) {
+      taskDetails.recordTaskDetails(
+        hashedTasks.map((task) => ({
+          hash: task.hash,
+          project: task.target.project,
+          target: task.target.target,
+          configuration: task.target.configuration,
+        }))
+      );
+    }
   }
 
   performance.mark('hashMultipleTasks:end');
--- a/packages/nx/src/hasher/hash-task.ts
+++ b/packages/nx/src/hasher/hash-task.ts
@@ -54,7 +54,11 @@ export async function hashTasksThatDoNotDependOnOutputsOfOtherTasks(
     })
     .map((t) => t.task);
 
-  const hashes = await hasher.hashTasks(tasksToHash, taskGraph, process.env);
+  const perTaskEnvs: Record<string, NodeJS.ProcessEnv> = {};
+  for (const task of tasksToHash) {
+    perTaskEnvs[task.id] = process.env;
+  }
+  const hashes = await hasher.hashTasks(tasksToHash, taskGraph, perTaskEnvs);
   const ioService = getTaskIOService();
   const hasInputSubscribers = ioService.hasTaskInputSubscribers();
   for (let i = 0; i < tasksToHash.length; i++) {
@@ -138,11 +142,19 @@ export async function hashTask(
   );
 }
 
+/**
+ * Batch-hash `tasks`. `perTaskEnvs` must contain an entry keyed by
+ * `task.id` for every task — the per-task env is what each task's
+ * custom hasher sees and what the built-in hasher reads
+ * `HashInstruction::Environment` inputs against. Callers that
+ * genuinely want to hash against a single shared env should build
+ * `{ [task.id]: env }` for every task.
+ */
 export async function hashTasks(
   hasher: TaskHasher,
   projectGraph: ProjectGraph,
   taskGraph: TaskGraph,
-  env: NodeJS.ProcessEnv,
+  perTaskEnvs: Record<string, NodeJS.ProcessEnv>,
   taskDetails: TaskDetails | null,
   tasksToHashOverride?: Task[]
 ) {
@@ -181,7 +193,7 @@ export async function hashTasks(
       workspaceConfig: projectsConfigurations,
       projectsConfigurations,
       nxJsonConfiguration: nxJson,
-      env,
+      env: perTaskEnvs[task.id],
     } as any);
     task.hash = value;
     task.hashDetails = details;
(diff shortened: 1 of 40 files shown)


packages/nx/src/hasher/hash-task.ts
```diff
@@ -214,14 +216,21 @@ export async function hashTasks(
...
-        project: task.target.project,
-        target: task.target.target,
-        configuration: task.target.configuration,
-      }))
-    );
+    // Guard against a custom hasher resolving with a falsy value —
+    // the built-in batch hasher always produces a hash, but user-written
+    // custom hashers are untrusted and an empty/undefined hash would
+    // violate the task_details schema downstream.
+    const hashedTasks = tasks.filter((task) => task.hash);
```
AgentEnder: follow up there's probably more perf to be gained here by doing this in a for loop instead of .filter.map

packages/nx/src/tasks-runner/task-orchestrator.ts
```diff
@@ -656,7 +808,118 @@ export class TaskOrchestrator {
...
+    doNotSkipCache: boolean,
+    task: Task,
+    groupId: number
+  ): void {
+    const worker: Promise<void> = this.runTaskDirectly(
+      doNotSkipCache,
+      task,
+      groupId
+    )
+      .then(() => {})
```
AgentEnder: rm

AgentEnder (approved):

AgentEnder (commented):
- [P2] Hash scheduled tasks with task-specific env — https://github.com/nrwl/nx/blob/215d1209404fc8a08b362605b146e7f1f86cab47/packages/nx/src/tasks-runner/task-orchestrator.ts#L269-269
    When two or more unhashed tasks are ready, this path sets task.hash using this.batchEnv, so processTask() later skips the
    per-task getTaskSpecificEnv() hash path. For tasks whose hash depends on target/project .env files or a custom hasher that
    reads env, the cache key is computed without those task-specific values; the same task will be hashed differently when it is
    the only ready task, and can incorrectly reuse stale cache after env changes. Hash these tasks with each task's task-
    specific env, or leave them for processTask().
  - [P2] Preserve busy retries for batch cache lookups — [src](https://github.com/nrwl/nx/blob/215d1209404fc8a08b362605b146e7f1f86cab47/benchmarks)/packages/nx/src/native/cache/cache.rs#L185-191
    This new batch lookup executes the UPDATE ... RETURNING via stmt.query_map() directly, bypassing the
    retry_db_operation_when_busy! retry wrapper used by NxDbConnection::query_row(). When another Nx process briefly holds the
    SQLite write lock, batch cache reads for multiple tasks will now surface DatabaseBusy and can fail the whole run instead of
    retrying like the single-task cache path.

AgentEnder (changes requested):

AgentEnder (approved):

PR #35175, opened 2026-04-05
fix(core): kill discrete tasks and use tree-kill for batch cleanup on SIGINT

FrozenPandaz:
## Current Behavior

When Nx receives SIGINT (Ctrl+C), `performCleanup()` in the task orchestrator kills continuous tasks and run-commands tasks, but **not discrete tasks** (executor-based builds like `@nx/js:tsc`, `nx:run-script`, etc.). In TUI mode, fork workers run in separate PTY process groups and don't receive SIGINT directly, so cleanup hangs waiting for tasks that are never terminated.

Additionally, `BatchProcess.kill()` only sends a signal to the immediate child process, leaving grandchild processes (e.g. Java JVM, Gradle daemon, Gradle workers) alive.

The Gradle batch executor also uses `execSync`, which blocks the Node event loop and prevents the worker from responding to signals.

## Expected Behavior

All task types — discrete, continuous, and run-commands — should be explicitly killed during SIGINT cleanup. Batch processes should use `tree-kill` to terminate the entire process tree. The Gradle batch executor should use async `spawn` instead of blocking `execSync`.

## Related Issue(s)

Closes NXC-4194

### DIFF
--- a/e2e/nx/src/run.test.ts
+++ b/e2e/nx/src/run.test.ts
@@ -2,6 +2,7 @@ import {
   checkFilesExist,
   cleanupProject,
   fileExists,
+  getStrippedEnvironmentVariables,
   isWindows,
   newProject,
   readJson,
@@ -14,6 +15,7 @@ import {
   updateFile,
   updateJson,
 } from '@nx/e2e-utils';
+import { execSync } from 'child_process';
 import { PackageJson } from 'nx/src/utils/package-json';
 import * as path from 'path';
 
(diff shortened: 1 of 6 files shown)


leosvelperez (approved):

lourw (approved):

Stanzilla:
@FrozenPandaz might need something like this as well? https://github.com/chromaui/chromatic-cli/pull/1261 `ps` is not included in the `:slim` docker images for example that many people use.

Disclaimer: Did not read the full PR, just remembered seeing this other PR in Chromatic the other day and figured I could mention it.

PR #35183, opened 2026-04-06
fix(nextjs): respect src option when setting sourceRoot in application generator

### LINKED ISSUE #35181: @nx/next:application generator ignores --src flag when setting sourceRoot
## Current Behavior

When generating a Next.js application with `--src=true` (the default), the generator creates the `src/` directory structure correctly but sets `sourceRoot` in the project configuration to the project root instead of the `src/` subdirectory:

```json
{
  "root": "apps/my-app",
  "sourceRoot": "apps/my-app",
  "projectType": "application"
}
```

## Expected Behavior

`sourceRoot` should respect the `--src` flag:
- `--src=true` (default): `"sourceRoot": "apps/my-app/src"`
- `--src=false`: `"sourceRoot": "apps/my-app"`

This is consistent with how other Nx generators handle `sourceRoot` (e.g. `@nx/react:application` sets `sourceRoot` to `${options.appProjectRoot}/src`).

## Steps to Reproduce

1. Create a new Nx workspace
2. Run `nx g @nx/next:application --directory=apps/my-app` (uses `--src=true` by default)
3. Check the generated `project.json` or `package.json` nx config
4. Observe that `sourceRoot` is set to `apps/my-app` instead of `apps/my-app/src`

## Nx Report

```
Node           : 24.11.0
OS             : macOS
pnpm           : 10.28.2

nx             : 22.6.4
@nx/next       : 22.6.4
```

## Operating System

- [x] macOS

## Additional Information

**Root Cause:** In `packages/next/src/generators/application/lib/add-project.ts` (line 69), `sourceRoot` is hardcoded to `options.appProjectRoot` without checking `options.src`:

```ts
const project: ProjectConfiguration = {
  root: options.appProjectRoot,
  sourceRoot: options.appProjectRoot,  // ignores options.src
  // ...
};
```

The `src` option is defined in `schema.json` with `default: true` and is correctly used in `create-application-files.ts` to determine file paths, but `add-project.ts` does not use it for `sourceRoot`.

tuffz:
## Current Behavior

The `@nx/next:application` generator ignores the `--src` flag when setting `sourceRoot` in `project.json`. Even when `src` is `true` (the default), `sourceRoot` is set to the project root instead of the `src/` subdirectory:

```json
{
  "root": "apps/my-app",
  "sourceRoot": "apps/my-app",
  "projectType": "application"
}
```

This is inconsistent with the actual file structure created by the generator, which correctly places app files under `src/`.

## Expected Behavior

`sourceRoot` should respect the `--src` flag:
- `--src=true` (default): `"sourceRoot": "apps/my-app/src"`
- `--src=false`: `"sourceRoot": "apps/my-app"`

This is consistent with how other Nx generators handle `sourceRoot` (e.g. `@nx/react:application`).

## Related Issue(s)

Fixes #35181


packages/next/src/generators/application/lib/add-project.ts
```diff
@@ -68,7 +68,9 @@ export function addProject(host: Tree, options: NormalizedSchema) {
 
   const project: ProjectConfiguration = {
     root: options.appProjectRoot,
-    sourceRoot: options.appProjectRoot,
+    sourceRoot: options.src
+      ? joinPathFragments(options.appProjectRoot, 'src')
+      : options.appProjectRoot,
```
leosvelperez, line 73: We must also persist this value to `packageJson.nx.sourceRoot` when `useProjectJson` is `false`. TypeScript-solution workspaces use `package.json`-based project configuration, and this project object is not written in that path, so `sourceRoot` remains unset.

Please cover both `package.json` and `project.json` configurations when applying the `src` option.

leosvelperez (changes requested):

tuffz:
Fixed in https://github.com/nrwl/nx/pull/36833

PR #35188, opened 2026-04-07
fix(core): replace LGPL-licensed @ltd/j-toml with BSD-3-Clause smol-toml

JiggyJinjo:
## Current Behavior

`nx` depends on `@ltd/j-toml` which is licensed under **LGPL-3.0**. This creates licensing concerns for projects that bundle or distribute nx, as LGPL requires downstream users to allow relinking/modification of the LGPL component.

A previous attempt ([3446dd2](https://github.com/nrwl/nx/commit/3446dd2f77a1b182f9b64a83586ab68a2f0c063f)) replaced it with `@iarna/toml`, but that library:
- Only supports TOML 1.0.0-rc.1 (not even the final 1.0.0 spec)
- Has been effectively unmaintained since ~2020
- Is significantly slower than alternatives

## Expected Behavior

Use a permissively-licensed, actively maintained, fast TOML parser.

## Solution

Replace `@ltd/j-toml` with [`smol-toml`](https://github.com/squirrelchat/smol-toml) (BSD-3-Clause):

| | @ltd/j-toml (current) | @iarna/toml (other PR) | **smol-toml (this PR)** |
|---|---|---|---|
| License | LGPL-3.0 | ISC | **BSD-3-Clause** |
| TOML spec | 1.0.0 | 1.0.0-rc.1 | **1.1.0** |
| Weekly downloads | — | 2.7M | **6.8M** |
| Maintained | Yes | Dormant (~2020) | **Active (2026)** |
| Performance | Baseline | ~same | **2-4x faster** |
| CJS support | Yes | Yes | **Yes** |

### Changes

- Replace `@ltd/j-toml` imports with `smol-toml` in `set-up-ai-agents.ts` and `test-utils.ts`
- Remove j-toml-specific APIs (`TOML.Section()`, `TOML.inline()`, `newlineAround` option) in favor of smol-toml's simpler `parse()`/`stringify()`
- Update inline test snapshots (single quotes → double quotes, minor formatting differences)

### Note on inline snapshots

Some inline snapshots may need a final update via `--updateSnapshot` once CI runs the full test suite. The snapshot changes included here follow the pattern observed from smol-toml's output format (double-quoted strings, no leading newline before first section).

## Test plan

- [ ] CI passes with updated snapshots
- [ ] `nx configure-ai-agents` generates valid Codex config.toml
- [ ] Release versioning for Rust/Cargo projects still produces correct Cargo.toml output

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@jest/reporters": "catalog:jest",
     "@jest/test-result": "catalog:jest",
     "@jest/types": "catalog:jest",
-    "@ltd/j-toml": "^1.38.0",
+    "smol-toml": "^1.6.1",
     "@module-federation/enhanced": "2.1.0",
     "@module-federation/sdk": "^2.1.0",
     "@monodon/rust": "2.3.0",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@jest/reporters": "catalog:jest",
     "@jest/test-result": "catalog:jest",
     "@jest/types": "catalog:jest",
-    "smol-toml": "^1.6.1",
+    "smol-toml": "1.6.1",
     "@module-federation/enhanced": "2.1.0",
     "@module-federation/sdk": "^2.1.0",
     "@monodon/rust": "2.3.0",
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -50,7 +50,7 @@
   },
   "homepage": "https://nx.dev",
   "dependencies": {
-    "@ltd/j-toml": "^1.38.0",
+    "smol-toml": "^1.6.1",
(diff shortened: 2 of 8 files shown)


FrozenPandaz (changes requested):
The version should be pinned to an exact version rather than using a caret range.

In both `package.json` and `packages/nx/package.json`, change:
```diff
- "smol-toml": "^1.6.1",
+ "smol-toml": "1.6.1",
```

This ensures consistent behavior across all installs and avoids surprise breakage from a minor/patch update.

I am going to update this PR to do so.

FrozenPandaz (approved):

FrozenPandaz:

PR #35193, opened 2026-04-07
fix(core): create process report on fatal error in .nx/workspace-data

AgentEnder:
## Current Behavior
On process segfault there is no way at all to diagnose where it is coming from

## Expected Behavior
On segfault we write a report file that has information in it and log that info on the next nx command if one happens to be ran. 

Because segfaults exit the node process immediately, we do not have another way of handling them so we do have to rely on a follow up command to surface them. Unfortunately the errors are pretty rare so the likelihood of this helping isn't awesome, but it could give us some info.

## AI summary

This pull request introduces a new utility for surfacing Node.js fatal error diagnostic reports and refactors how project graph data is accessed in the format command. The most significant changes are the addition of the fatal error reporting mechanism and the simplification of how project graph data is loaded, which should improve maintainability and reliability.

**Fatal error reporting utility:**

* Added a new `surfaceFatalErrorReports` function in `report-on-fatal-error.ts` that scans for Node.js fatal error reports in the workspace data directory, surfaces them in the console, and emits GitHub Actions workflow annotations and step summaries if running in CI. Reports are renamed after processing to prevent duplication.

**Format command refactoring:**

* Refactored `format.ts` to remove unused imports (`FileData`, `allFileData`) and to only load the project graph when necessary, which reduces unnecessary file system and computation overhead. [[1]](diffhunk://#diff-ff26ab1495d5bd9854f0377c1aad51b1a749fa0ea82d2c66d363c06d417ebc67L6-R6) [[2]](diffhunk://#diff-ff26ab1495d5bd9854f0377c1aad51b1a749fa0ea82d2c66d363c06d417ebc67L23)
* Updated the logic in `getPatterns` and related functions to lazily load the project graph and to remove the need for passing all workspace files, simplifying function signatures and improving performance. [[1]](diffhunk://#diff-ff26ab1495d5bd9854f0377c1aad51b1a749fa0ea82d2c66d363c06d417ebc67L104) [[2]](diffhunk://#diff-ff26ab1495d5bd9854f0377c1aad51b1a749fa0ea82d2c66d363c06d417ebc67R111) [[3]](diffhunk://#diff-ff26ab1495d5bd9854f0377c1aad51b1a749fa0ea82d2c66d363c06d417ebc67L135-R138) [[4]](diffhunk://#diff-ff26ab1495d5bd9854f0377c1aad51b1a749fa0ea82d2c66d363c06d417ebc67L155)

## Related Issue(s)

Fixes #


packages/nx/bin/nx.ts
```diff
@@ -51,6 +53,11 @@ async function main() {
   const workspace = findWorkspaceRoot(process.cwd());
 
   if (workspace) {
+    const workspaceDataDir = workspaceDataDirectoryForWorkspace(workspace.dir);
+    process.report.reportOnFatalError = true;
+    process.report.directory = workspaceDataDir;
+    surfaceFatalErrorReports(workspaceDataDir);
```
FrozenPandaz: Let's not do this because.. it seems like extra code that isn't really useful most of the time unless 2 Nx commands are run...

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #35204, opened 2026-04-08
fix(core): native watcher rewrite + daemon hardening for daemon-on e2e

### LINKED ISSUE #12: Convert-to-workspace Schematic doesn't update protractor.conf
Config specs and project url needs to be updated.

### LINKED ISSUE #13: Workspace schematic should include code formatting
It is very annoying to deal with formatting code (especially via tslint). Instead we should use tools like clang-format or prettier to do that.

We should change the workspace (and covert-to-workspace) schematic in the following way:

- add the scripts dir
- put a file called format.sh in the scripts dir. The script should format all tsfiles in libs and apps. See the nx repo on how to do it. 
- update package.json to include format command (it will run the script)
- update tslint.json to remove all checks related to formatting code

Finally, we should update the schematics to generate code that is formatted properly. So running `yarn format` will produce 0 changes. The last part might be tricky in some situations, but things like quotes, spaces, etc can be handled in the same PR.

FrozenPandaz:
## Current Behavior

This branch started as a fix for a flaky `e2e/nx/src/watch.test.ts`. Once we flipped `NX_DAEMON='true'` in the e2e harness, several long-latent daemon issues surfaced; they're fixed in the same PR. The most user-visible regression — `nx release version` silently dropping projects from a fixed release group after `git checkout` — is the last item below and is what `e2e/release/src/multiple-release-branches.test.ts` exercises.

**Native watcher.** `nx watch` dropped and duplicated events on Linux/Windows:
1. `watchexec.config.pathset()` registered new directories asynchronously — files written before the real inotify registration were lost.
2. Every raw `notify` event fired a callback (no debounce) so one logical write produced multiple `nx watch` invocations.
3. The daemon called `notifyFileWatcherSockets` twice per event (eagerly for updates/deletes, again post-recompute for creates), doubling invocations.
4. `getProjectsAndGlobalChanges` looked changed files up in `fileMap.projectFileMap`, so brand-new files fell through to `globalFiles` and `--projects=foo` watchers never matched.
5. `nx:test-native` didn't compile on master (walker test used `nix::unistd::mkfifo` behind a feature that wasn't enabled).
6. macOS regressed after the notify migration: `FsEventWatcher` dropped immediately because the closure had no cfg-active references on darwin.

**Daemon (surfaced once it ran end-to-end in CI):**
7. `getPluginsSeparated` reused `pendingPluginsPromise` via `??=` after the plugins-config hash changed, serving stale plugins forever after `nx add`.
8. The auto-recompute path updated the in-memory graph but never persisted it; subprocesses reading the disk cache saw stale data (flaky `eslint dependency-checks`).
9. `startInBackground` had an fd race: a concurrent `reset()` could null `this._out`/`this._err` while we awaited `open()`, crashing the spawn with `Cannot read properties of null (reading 'fd')`.
10. The daemon's env-reflection was additive only — `NX_*` vars set by one CLI invocation persisted forever, leaking (e.g., `NX_PREFER_NODE_STRIP_TYPES=true` from a single `nx report` poisoned every later plugin load). Plugin worker's `setWorkerEnv` had the same bug.
11. Several e2e cache-hit assertions matched `[local cache]`, but the daemon-on path emits `[existing outputs match the cache, left as is]` (output kept rather than restored).

**Daemon graph cache (more flake hunting after the watcher rewrite landed):**
12. `resetInternalStateIfNxDepsMissing` ran before every `getCachedSerializedProjectGraphPromise` and reset state whenever `project-graph.json` was missing AND a cached promise existed — but on cold start, *every* request before the first persist matches that condition. It was firing constantly, tearing down in-flight first computes and forcing a redundant recompute. Worse, its `catch` branch unconditionally reset on any `fileExists` error.
13. When `resetInternalState` fires from inside `processCollectedUpdatedAndDeletedFiles`'s catch (e.g., `retrieveWorkspaceFiles` throws on partial disk state), `cachedSerializedProjectGraphPromise` is set to `undefined`. A concurrent stale compute that hits `chainToLatest` after the reset would read back `undefined`; the `if (stale) return stale` guard treats `undefined` as falsy, so the stale compute falls through and commits its (now-stale) data to module state.

**Force-flush concurrency (flagged by Graphite review):**
14. The processing thread dropped any extra `ForceFlush` messages it found while draining the channel (the `ProcMsg::ForceFlush(_) => { /* drop the extras */ }` arm). Their reply senders were silently discarded; those callers waited the full 500 ms `recv_timeout` and returned an empty Vec. Concurrent CLI requests would each see that latency hit.

**Watcher event misclassification (the `nx release version` failure on `multiple-release-branches.test.ts`):**
15. `git checkout` does `unlink + write` on some tracked files. inotify fires `IN_DELETE` then `IN_CREATE` for the same path; both landed in the watcher's accumulator within one burst.
16. `merge_event`'s priority rule was *Delete > Create > Update* — meaning a `Create` arriving after a `Delete` for the same path was silently dropped. The accumulator emitted only the Delete.
17. Downstream `updateFilesInContext` on the JS side then *removed* the (still-existing) file from the Rust workspace context's index. `multiGlobWithWorkspaceContext` no longer returned it. Plugins didn't process it. The project was missing from the resulting graph, with no error.
18. `release version` then evaluated `isProjectPublic`, which checks for the project's `package.json` in the file map. With the file gone from the index, the check returned `false`, the project was excluded as not-public, and the release group ran with one fewer project than expected.
19. Two related classification bugs surfaced during the diagnosis: (a) the early-return for `!meta_exists` emitted Delete *unconditionally*, even for `Modify`/`Create` events whose stat happened to fail during a transient atomic-rename window; (b) notify-rs delivers a third coalesced rename event (`Modify(Name(Both))`) that fell through to the generic `Modify(_) → Update` arm, overriding the per-side `From` Delete on the source path of an `fs::rename`.

**Maven (separate but bundled):**
20. Tests asserted on `BUILD SUCCESS` in batch mode, but the resident batch runner's `BatchExecutionListener.sessionEnded` is a no-op — Maven's footer is replaced by `Nx Maven Summary`. Those assertions could never match.

#### Architecture

```
Before: Watcher → watchexec (async config loop) → notify → inotify/FSEvents
After:  Watcher → notify → inotify/FSEvents (direct, synchronous)
```

The watcher is one Rust struct (`packages/nx/src/native/watch/watcher.rs`) that owns:
- a `notify::RecommendedWatcher` (the OS-level subscription),
- a single `mpsc::Sender<ProcMsg>` shared by the notify event handler and the napi force-flush method,
- a long-running **processing thread** that reads from the corresponding `Receiver`.

#### How events are streamed

```
notify (OS) ─► NotifyForwarder ─┐
                                ├─► mpsc<ProcMsg> ─► processing thread ─► napi tsfn ─► JS callback
JS force_flush_pending() ───────┘                                       (or sync_channel reply)
```

1. **Notify side.** `notify::recommended_watcher(NotifyForwarder { tx })` hands every fs event to `NotifyForwarder::handle_event`, which wraps it as `ProcMsg::Notify(event)` and pushes it onto the channel. No work is done on the notify thread beyond the send.

2. **Processing thread.** A dedicated thread runs an infinite loop with three branches:
   - `ProcMsg::Notify(event)` — filter through gitignore/nxignore (`watch_filterer.rs`), run the new-directory fast path on Linux/Windows if the event is a `Create(Folder)`, transform notify's kinds into our `EventType` (Create/Update/Delete), and merge each resulting `WatchEventInternal` into a per-path `HashMap` accumulator. The merge rules are:
     - `(_, Delete)` → Delete (file is gone, final state wins)
     - `(_, Create)` → Create (file is in its initial state from our perspective; overrides earlier Update or Delete — the regression case 15-18 above)
     - `(Delete, Update)` → Update (file came back with new content)
     - `(Create, Update)` / `(Update, Update)` → keep existing (Create wins on Linux's IN_CREATE+IN_MODIFY pair)
   - `ProcMsg::ForceFlush(reply)` — drain whatever notify has buffered (`while let Ok(msg) = rx.try_recv()` so anything in flight at the moment of the request is included), **collect every queued ForceFlush reply channel encountered during the drain**, snapshot the accumulator, and send the same snapshot to all collected callers (closes #14 above). Used by the daemon to absorb pending events before serving a cached project graph (closes the IDLE_WINDOW race where a 99 ms-old event would otherwise miss the next read).
   - `Err(RecvTimeoutError::Timeout)` — the wake-up deadline elapsed; emit the accumulator to JS via the napi `ThreadsafeFunction` (`callback_tsfn.call(Ok(events), NonBlocking)`), clear it, and reset.

3. **Trailing-edge debounce.** Each `Notify` ingest updates a single `flush_deadline = min(now + IDLE_WINDOW, burst_start + MAX_WAIT)`:
   - `IDLE_WINDOW = 100 ms` — flush when the channel goes quiet for this long. Resets on every arriving event, so any burst with gaps <100 ms coalesces into one flush.
   - `MAX_WAIT = 500 ms` — starvation cap from the first event of the burst.
   - The loop's `rx.recv_timeout(wait)` either picks up the next message or returns `Timeout` exactly at `flush_deadline`. No separate timers, no cross-flush state, no `recent_paths` book-keeping.
   - When idle (`flush_deadline = None`) the loop polls on `SHUTDOWN_POLL` so `stop_flag` is observed promptly.

4. **napi tsfn handoff.** The processing thread invokes `callback_tsfn.call(Ok(Vec<WatchEvent>), NonBlocking)` to schedule the JS callback on Node's main thread. The notify thread itself never crosses into JS — only the processing thread does, and only at flush time. `Watcher::watch` is now a thin wrapper around an internal `watch_inner` that takes a generic callback, so unit tests can exercise the same loop without a JS runtime.

5. **`force_flush_pending` (synchronous drain).** The daemon calls `watcher.force_flush_pending()` from JS before reading the cached project graph. JS-side napi method creates a `sync_channel::<Vec<WatchEvent>>(1)` reply pair and sends `ProcMsg::ForceFlush(reply_tx)` on the same channel notify events flow through. The processing thread sees the request in order with any buffered notify events, drains, takes the accumulator, and replies. Concurrent callers all receive the same snapshot (closes #14).

#### Event classification (`types.rs::transform_event_to_watch_events`)

- A failed stat (`!meta_exists(metadata)`) only short-circuits to `EventType::Delete` when the notify event_kind is actually `Remove(_)`. For `Modify`/`Create` events whose stat happened to fail during a transient atomic-rename window, we fall through to the platform-specific path which derives the type from `event_kind` alone (closes #19a).
- `Modify(Name(RenameMode::Both | Any))` — the coalesced rename event notify-rs delivers in addition to per-side `From`/`To` events — is now skipped, so it can't override the `From` Delete on the source path of a rename (closes #19b).

#### New-directory fast path (Linux/Windows)

When notify reports a `Create(Folder)` on inotify/ReadDirectoryChangesW (which only watch the dirs they were given), the processing thread:
1. Calls `watcher.watch(new_dir, NonRecursive)` **synchronously** — the OS watch is active the moment this returns.
2. Re-walks the new directory and merges every existing entry into the accumulator with `EventType::Create`.
3. Recursively registers any subdirectories it found (so a `mkdir -p a/b/c/d` still hooks every level).

Because (1) is synchronous, files written between `mkdir` and the `watch()` call are caught by the re-walk in (2). Notify events for those same files arriving later just merge into the same accumulator entry — no duplication. macOS doesn't need this path: `FsEventWatcher` is recursive.

`notify_watcher` is held as `Option<Arc<Mutex<RecommendedWatcher>>>` on the struct so the processing-thread closure can clone the `Arc`; otherwise the macOS `move`-closure (which has no cfg-active references) wouldn't capture the watcher and `FsEventWatcher` would drop the moment `watch()` returned.

### Daemon

- **Plugin reload on config change.** `getPluginsSeparated` clears `pendingPluginsPromise` and tears down workers when `nx.json#plugins`'s hash changes.
- **Persist project graph after auto-recompute.** New `persistProjectGraphToDisk` helper called from `kickOffRecompute` and `getCachedSerializedProjectGraphPromise`.
- **fd race fix.** Open log handles into locals first, assign to `this._out`/`this._err` after both opens resolve, pass the local `fd`s to `spawn`.
- **Two-way `NX_*` env reflection.** New `applyDaemonEnvFromClient(newEnv)` helper writes new keys AND deletes any `NX_`-prefixed key the daemon has that the client doesn't (skipping daemon-side exclusions and required settings).
- **`scheduleTimeoutId` → in-flight promise** for self-documenting recompute scheduling.
- **Single `notifyFileWatcherSockets` registration** at daemon startup; one notification per batch.
- **Map new files by project root.** `getProjectsAndGlobalChanges` uses `createProjectRootMappings` + `findProjectForPath`, cached by reference identity on `currentProjectGraph`.
- **Cache invalidation gated on first persist** (closes #12). New `cacheHasBeenPersisted` flag set in `persistProjectGraphToDisk`. `resetInternalStateIfNxDepsMissing` returns early if the flag is false — before the first successful write, a missing `project-graph.json` is the *expected* state. Its `catch` branch no longer auto-resets on transient stat errors.
- **`chainToLatest` defensive kickOff** (closes #13). When a stale compute hits `chainToLatest` and finds `cachedSerializedProjectGraphPromise === undefined` (because `resetInternalState` ran), it now kicks off a successor synchronously and returns that promise instead of `undefined`. Prevents the "stale compute commits stale data" path that the falsy guard let through.

### Test suite

- `NX_DAEMON='true'` is the default in `e2e/utils/command-utils.ts` so CI exercises the daemon path.
- Cache-hit assertions updated to `[existing outputs match the cache, left as is]` for the daemon-on no-op restore path. One assertion in `cache.test.ts:216` reverted to `[local cache]` because that test adds an extra file to `dist/`, invalidating the outputs hash and forcing a real restore.
- Maven batch tests assert on `Successfully ran target X` instead of `BUILD SUCCESS`. `verbose: true` dropped where it only added Maven `-X` debug noise.
- Watch e2e harness uses `tree-kill` and waits for `close` before reading output. Default wait shortened from 6s/8s → 1s/2s now that debounce is deterministic.
- Walker test uses `std::os::unix::net::UnixListener::bind` instead of `nix::unistd::mkfifo` — same `is_hashable_file` rejection, no `nix` feature flag needed.
- **10 Rust watcher tests** (`packages/nx/src/native/watch/watcher.rs`) drive `Watcher::watch_inner` end-to-end with real fs ops on a tempdir — inotify/FSEvents → notify-rs → ProcMsg channel → EventIngestor → accumulator → callback. Cover: git-style unlink+write, vim-style atomic rename, plain in-place update, fresh create, rm, create+rm, cross-name rename, multi-file burst coalescing, hardcoded-ignored paths never reach callback, and 8-way concurrent `force_flush_pending` (regression for #14). Tests use canonicalized tempdir paths so the synthetic-gitignore filter scopes correctly on macOS where `/tmp` symlinks to `/private/tmp`.

## Testing

- `pnpm nx run e2e-nx:e2e-ci--src/watch.test.ts --skip-nx-cache` — 8/8 passed.
- `pnpm nx run e2e-js:e2e-ci--src/js-strip-types.test.ts` — 3/3 (was failing on test 3 before the env-reflection fix).
- `pnpm nx run e2e-maven:e2e-ci--src/maven-batch.test.ts` — 3/3 (was failing on test 1 + 3 before the assertion fix).
- `pnpm nx run e2e-release:e2e-ci--src/multiple-release-branches.test.ts` — 2/2 (was failing both before the merge-event fix; reproduced locally and confirmed via the daemon log dump that `git checkout` was emitting a stray Delete for one of the package.json files).
- `pnpm nx run nx:test-native` — 349 tests pass, including the 10 new watcher tests on Linux and macOS.
- macOS path validated end-to-end with the Arc fix.

## Related Issue(s)

N/A — flaky-test investigation that grew into a daemon-stability hardening pass.

### DIFF
--- a/packages/nx/src/daemon/client/daemon-environment.ts
+++ b/packages/nx/src/daemon/client/daemon-environment.ts
@@ -142,16 +142,18 @@ export function getDaemonEnv() {
  */
 export function applyDaemonEnvFromClient(newEnv: NodeJS.ProcessEnv): boolean {
   let changed = false;
-  for (const key in newEnv) {
-    if (process.env[key] !== newEnv[key]) {
-      process.env[key] = newEnv[key];
-      changed = true;
-    }
-  }
-  for (const key of Object.keys(process.env)) {
-    if (
+  const allKeys = new Set([
+    ...Object.keys(process.env),
+    ...Object.keys(newEnv),
+  ]);
+  for (const key of allKeys) {
+    if (key in newEnv) {
+      if (process.env[key] !== newEnv[key]) {
+        process.env[key] = newEnv[key];
+        changed = true;
+      }
+    } else if (
       key.startsWith('NX_') &&
-      !(key in newEnv) &&
       !isExcludedEnvVar(key) &&
       !Object.hasOwn(DAEMON_ENV_REQUIRED_SETTINGS, key)
     ) {
--- a/packages/nx/src/daemon/client/daemon-environment.ts
+++ b/packages/nx/src/daemon/client/daemon-environment.ts
@@ -134,11 +134,11 @@ export function getDaemonEnv() {
 }
 
 /**
- * Without the deletion step, an `NX_` var set by one client (e.g.
- * `NX_PREFER_NODE_STRIP_TYPES=true` for a single command) would persist in
- * the daemon and leak into every subsequent client's project-graph
- * computation. Deletion is scoped to `NX_` so unrelated environment the
- * daemon process needs to operate (PATH, HOME, etc.) is left alone.
+ * Without the deletion step, a var set by one client (e.g.
+ * `NX_PREFER_NODE_STRIP_TYPES=true` or `JAVA_TOOL_OPTIONS=...` for a single
+ * command) would persist in the daemon and leak into every subsequent
+ * client's project-graph computation. Deletion skips excluded vars and
+ * required settings, which the daemon owns and clients should not control.
  */
 export function applyDaemonEnvFromClient(newEnv: NodeJS.ProcessEnv): boolean {
   let changed = false;
@@ -153,7 +153,6 @@ export function applyDaemonEnvFromClient(newEnv: NodeJS.ProcessEnv): boolean {
         changed = true;
       }
     } else if (
-      key.startsWith('NX_') &&
       !isExcludedEnvVar(key) &&
       !Object.hasOwn(DAEMON_ENV_REQUIRED_SETTINGS, key)
     ) {
--- a/packages/nx/src/daemon/client/daemon-environment.ts
+++ b/packages/nx/src/daemon/client/daemon-environment.ts
@@ -132,3 +132,32 @@ export function getDaemonEnv() {
   }
   return Object.assign(env, DAEMON_ENV_REQUIRED_SETTINGS);
 }
+
+/**
+ * Without the deletion step, an `NX_` var set by one client (e.g.
+ * `NX_PREFER_NODE_STRIP_TYPES=true` for a single command) would persist in
+ * the daemon and leak into every subsequent client's project-graph
+ * computation. Deletion is scoped to `NX_` so unrelated environment the
+ * daemon process needs to operate (PATH, HOME, etc.) is left alone.
+ */
+export function applyDaemonEnvFromClient(newEnv: NodeJS.ProcessEnv): boolean {
+  let changed = false;
+  for (const key in newEnv) {
+    if (process.env[key] !== newEnv[key]) {
+      process.env[key] = newEnv[key];
+      changed = true;
+    }
+  }
+  for (const key of Object.keys(process.env)) {
+    if (
+      key.startsWith('NX_') &&
+      !(key in newEnv) &&
+      !isExcludedEnvVar(key) &&
+      !Object.hasOwn(DAEMON_ENV_REQUIRED_SETTINGS, key)
+    ) {
+      delete process.env[key];
+      changed = true;
+    }
+  }
+  return changed;
+}
--- a/packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
+++ b/packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
@@ -42,7 +42,6 @@ import {
   updateFilesInContext,
 } from '../../utils/workspace-context';
 import { workspaceRoot } from '../../utils/workspace-root';
-import { serverLogger } from '../logger';
 import { ProgressTopics } from '../../utils/progress-topics';
 import {
   subscribeClientToTopic,
@@ -51,6 +50,7 @@ import {
 import { notifyFileChangeListeners } from './file-watching/file-change-events';
 import { notifyFileWatcherSockets } from './file-watching/file-watcher-sockets';
 import { notifyProjectGraphListenerSockets } from './project-graph-listener-sockets';
+import { serverLogger } from '../logger';
 
 interface SerializedProjectGraph {
   error: Error | null;
@@ -77,6 +77,7 @@ export let currentSourceMaps: ConfigurationSourceMaps | undefined;
 // This lets us detect mid-flight re-modifications when clearing processed files.
 const collectedUpdatedFiles = new Map<string, number>();
 const collectedDeletedFiles = new Map<string, number>();
+
 const projectGraphRecomputationListeners = new Set<
   (
     projectGraph: ProjectGraph,
@@ -201,7 +202,7 @@ export async function getCachedSerializedProjectGraphPromise(
   }
 }
 
-export function addUpdatedAndDeletedFiles(
+export function scheduleProjectGraphRecomputation(
   createdFiles: string[],
   updatedFiles: string[],
   deletedFiles: string[]
@@ -224,10 +225,7 @@ export function addUpdatedAndDeletedFiles(
     deletedFiles.length > 0
   ) {
     notifyFileChangeListeners({ createdFiles, updatedFiles, deletedFiles });
-  }
-
-  if (updatedFiles.length > 0 || deletedFiles.length > 0) {
-    notifyFileWatcherSockets(null, updatedFiles, deletedFiles);
+    notifyFileWatcherSockets(createdFiles, updatedFiles, deletedFiles);
   }
 
   if (createdFiles.length > 0) {
@@ -247,10 +245,6 @@ export function addUpdatedAndDeletedFiles(
         );
       const { projectGraph, sourceMaps, error } =
         await cachedSerializedProjectGraphPromise;
-
-      if (createdFiles.length > 0) {
-        notifyFileWatcherSockets(createdFiles, null, null);
-      }
 
       notifyProjectGraphRecomputationListeners(projectGraph, sourceMaps, error);
     }, waitPeriod);
--- a/packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
+++ b/packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
@@ -352,8 +352,15 @@ async function processFilesAndCreateAndSerializeProjectGraph(
   ];
   const myGeneration = ++recomputationGeneration;
 
-  // Helper to check if this recomputation is stale (a newer one has started)
-  const isStale = () => myGeneration !== recomputationGeneration;
+  // A newer kickOffRecompute has already replaced
+  // cachedSerializedProjectGraphPromise. Returning it lets the async
+  // unwrap chain our caller onto the newer compute's result; pass
+  // notifyAbort=true when the graph-phase counters still need balancing.
+  const chainToLatest = (notifyAbort: boolean) => {
+    if (myGeneration === recomputationGeneration) return null;
+    if (notifyAbort) notifyPluginsGraphAborted(plugins);
+    return cachedSerializedProjectGraphPromise;
+  };
 
   try {
     performance.mark('hash-watched-changes-start');
@@ -398,14 +405,8 @@ async function processFilesAndCreateAndSerializeProjectGraph(
       }
     }
 
-    // Stale: a newer kickOffRecompute already replaced
-    // cachedSerializedProjectGraphPromise. Chain to it via the cached
-    // pointer; the async unwrap means our caller just sees the newer
-    // compute's result.
-    if (isStale()) {
-      notifyPluginsGraphAborted(plugins);
-      return cachedSerializedProjectGraphPromise;
-    }
+    const stalePostCreateNodes = chainToLatest(true);
+    if (stalePostCreateNodes) return stalePostCreateNodes;
 
     await processCollectedUpdatedAndDeletedFiles(
       projectConfigurationsResult,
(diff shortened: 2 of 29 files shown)


packages/nx/src/daemon/client/daemon-environment.ts
```diff
@@ -132,3 +132,33 @@ export function getDaemonEnv() {
...
+        changed = true;
+      }
+    } else if (
+      !isExcludedEnvVar(key) &&
+      !Object.hasOwn(DAEMON_ENV_REQUIRED_SETTINGS, key)
+    ) {
+      delete process.env[key];
+      changed = true;
+    }
+  }
```
AgentEnder, line 162: Make sure this doesn't drop DAEMON_ENV_OVERRIDABLE_SETTINGS, they should be able to be overridden, but not able to be deleted outright

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -102,16 +106,20 @@ export async function getCachedSerializedProjectGraphPromise(
...
+    // flushed past IDLE_WINDOW yet would be invisible to the staleness
+    // check below — the daemon would return a stale graph.
+    await flushPendingWorkspaceChanges();
+
+    // If an auto-recompute is in flight, let it finish. It already writes
+    // `cachedSerializedProjectGraphPromise` and notifies listeners, so after
+    // awaiting we just fall through to serve its result.
+    if (autoRecomputePromise) {
+      await autoRecomputePromise;
     }
```
AgentEnder: note: check that mid-graph events still cancel later graph hooks (e.g. file change during createNodes cancels createDependencies, instead of running it and then throwing out results

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -217,44 +213,54 @@ export function addUpdatedAndDeletedFiles(
...
-        notifyFileWatcherSockets(createdFiles, null, null);
-      }
+async function runAutoRecomputeLoop() {
+  do {
+    cachedSerializedProjectGraphPromise =
+      processFilesAndCreateAndSerializeProjectGraph(
+        await getPluginsSeparated()
+      );
+    const result = await cachedSerializedProjectGraphPromise;
+    notifyProjectGraphRecomputationListeners(
```
AgentEnder: open question: should we do this now or after settling

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
```diff
@@ -217,44 +213,54 @@ export function addUpdatedAndDeletedFiles(
...
+    );
 
-      notifyProjectGraphRecomputationListeners(projectGraph, sourceMaps, error);
-    }, waitPeriod);
-  }
+    // Subprocesses that read the cache directly (e.g. eslint rules
+    // calling readCachedProjectGraph) bypass the daemon socket, so
+    // they only see updates that hit disk.
+    persistProjectGraphToDisk(result);
+  } while (collectedUpdatedFiles.size > 0 || collectedDeletedFiles.size > 0);
```
AgentEnder: maybe follow up, it'd be nice if we could emit a note for the spinner here that would be displayed for at least say 500ms, as is if we emitted a progress event to say "recomputing bc files changed", it'd immediately be cleared by the createNodes starting up

packages/nx/src/daemon/server/server.ts
```diff
@@ -255,20 +256,14 @@ async function handleMessage(socket: Socket, data: string) {
...
-    if (shouldRecomputeGraph) {
-      serverLogger.log('Graph recompute necessary due to env variable refresh');
-      forwardEnvToPluginWorkers(payload.env);
-      invalidateGraphCache();
-    }
+  if (
+    isDaemonMessage(payload) &&
+    payload.env &&
+    applyDaemonEnvFromClient(payload.env)
+  ) {
```
AgentEnder:
```suggestion
  if (
    isDaemonMessage(payload) &&
    payload.env
  ) {
    const shouldRecomputeGraph =  applyDaemonEnvFromClient(payload.env)
```

packages/nx/src/native/watch/watcher.rs
```diff
@@ -1,80 +1,129 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
-use std::collections::hash_map::Entry;
 use std::path::MAIN_SEPARATOR;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc::{RecvTimeoutError, Sender, SyncSender};
```
AgentEnder: Use the other message channel thing if possible

packages/nx/src/native/watch/watcher.rs
```diff
@@ -161,203 +333,196 @@ impl Watcher {
...
-            let ignore_globs_for_action = ignore_globs.clone();
-            watch_exec.config.on_action(move |mut action| {
-                let signals: Vec<Signal> = action.signals().collect();
-
-                if signals.contains(&Signal::Terminate) {
-                    trace!("terminate - ending watch");
-                    action.quit();
-                    return action;
-                }
+            let ignore_globs = build_ignore_glob_set();
```
AgentEnder: Which globs are these, and why only not macos?

packages/nx/src/native/watch/watcher.rs
```diff
@@ -161,203 +333,196 @@ impl Watcher {
...
-                        // If its empty, insert
-                        Entry::Vacant(e) => {
-                            e.insert(g);
+                        let watch_events: Vec<WatchEvent> =
+                            accumulator.values().map(|e| e.into()).collect();
+                        trace!(count = watch_events.len(), "force-flushing events");
+                        accumulator.clear();
+                        burst_start = None;
+                        flush_deadline = None;
+                        let _ = reply.send(watch_events);
```
AgentEnder: Maybe trace failed sends?

packages/nx/src/native/watch/watcher.rs
```diff
@@ -161,203 +333,196 @@ impl Watcher {
...
+                        if !accumulator.is_empty() {
+                            let watch_events: Vec<WatchEvent> =
+                                accumulator.values().map(|e| e.into()).collect();
+                            trace!(count = watch_events.len(), "flushing accumulated events");
+                            callback_tsfn
+                                .call(Ok(watch_events), ThreadsafeFunctionCallMode::NonBlocking);
+                            accumulator.clear();
                         }
+                        burst_start = None;
+                        flush_deadline = None;
```
AgentEnder: Maybe function this?

AgentEnder: sendAccumulatedEventsAndResetState

packages/nx/src/native/watch/watcher.rs
```diff
@@ -161,203 +333,196 @@ impl Watcher {
...
+                    }
+                    Err(RecvTimeoutError::Disconnected) => {
+                        trace!("notify channel disconnected, exiting");
```
AgentEnder: Should we shut down daemon, panic, etc? We don't want live daemon that receives no watch events

packages/nx/src/native/walker.rs
```diff
@@ -453,27 +453,29 @@ nested/child-two/
...
     #[cfg(all(unix, not(target_arch = "wasm32")))]
     #[test]
-    fn skips_named_pipes() {
-        use nix::sys::stat::Mode;
-        use nix::unistd::mkfifo;
+    fn skips_non_regular_files() {
+        use std::os::unix::net::UnixListener;
 
         let temp_dir = setup_fs();
-        let fifo_path = temp_dir.path().join("a-named-pipe");
```
AgentEnder, line 464: Lets do both, not replace fifo stuff... there was an issue opened about fifos

packages/nx/src/native/watch/watch_filterer.rs
```diff
@@ -20,135 +19,102 @@ pub struct WatchFilterer {
...
+
+        // .nxignore takes precedence over .gitignore. Only consult it for
+        // paths under the origin — references outside the workspace are
+        // ignored.
+        let nx_match = self
+            .nx_ignore
+            .as_ref()
+            .filter(|_| path.starts_with(&self.origin))
+            .map(|ig| ig.matched_path_or_any_parents(path, is_dir))
+            .unwrap_or(Match::None);
```
AgentEnder: Can we restucture this so its not a long method chain, the filter being unrelated to the object its applied to makes it hard to read this

AgentEnder (changes requested):

packages/nx/src/daemon/server/project-graph-incremental-recomputation.ts
AgentEnder: Fine

AgentEnder (approved):

PR #35216, opened 2026-04-08
fix(maven): honor settings.xml in Maven 3 batch runner

jeff-miller-cfa:
## Current Behavior

When `@nx/maven` runs Maven targets in Nx batch mode, the Maven 3 batch runner builds a `MavenExecutionRequest` and only calls `MavenExecutionRequestPopulator.populateDefaults()`. That path injects default remote repositories but does not merge user and global `settings.xml` (mirrors, servers, proxies, profile repositories, etc.). Resolution can ignore corporate mirrors and behave differently from the `mvn` CLI and from non-batch execution.

## Expected Behavior

Batch mode should apply the same effective settings as the `mvn` CLI: build settings with `SettingsBuilder`, then `populateFromSettings()` before `populateDefaults()`. Global `settings.xml` should resolve the same way as the launcher (`${maven.conf}/settings.xml`), so `maven.conf` is set from `maven.home` when unset.

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/34948


FrozenPandaz (approved):

PR #35264, opened 2026-04-10
chore(misc): add versioned docs snapshot script

jaysoo:
## Current Behavior

No automated way to create static snapshots of the docs site for versioned branches (e.g., v22.nx.dev). The existing `release-docs.ts` force-pushes the full source branch, requiring Netlify to build from source.

## Expected Behavior

`node ./scripts/create-versioned-docs.mts v22` creates a deployable orphan branch with the pre-built static site:

- Fetches latest stable git tag for the major version (e.g., `22.6.4`)
- Builds `astro-docs` (v21+) or `nx-dev` with static export (v18-v20)
- Creates orphan branch with pre-built files at `nx-dev/nx-dev/.next/`
- Includes `netlify.toml` that skips `@netlify/plugin-nextjs` for pure static serving
- Resolves `GITHUB_TOKEN` from 1Password or env var
- Server-side redirects for `/docs` → `/docs/getting-started/intro`
- `--force` flag to overwrite existing branches

Deployed via Netlify branch deploys at `v{major}.nx.dev`.

### Usage

```bash
node ./scripts/create-versioned-docs.mts v22
node ./scripts/create-versioned-docs.mts v21 --force
git push -f origin v22
```

### Tested

- v21 https://v21.nx.dev/docs
- v20 https://v20.nx.dev/docs
- v19 https://v19.nx.dev/docs 

## Related Issue(s)

Fixes DOC-69

### DIFF
--- a/nx-dev/nx-dev/netlify.toml
+++ b/nx-dev/nx-dev/netlify.toml
@@ -23,3 +23,45 @@ included_files = [
   "!node_modules/webpack/**",
   "!node_modules/sass/**",
 ]
+
+[[redirects]]
+  from = "https://16.nx.dev/*"
+  to = "https://nx.dev/:splat"
+  status = 308
+  force = true
+
+[[redirects]]
+  from = "https://17.nx.dev/*"
+  to = "https://nx.dev/:splat"
+  status = 308
+  force = true
+
+[[redirects]]
+  from = "https://18.nx.dev/*"
+  to = "https://nx.dev/:splat"
+  status = 308
+  force = true
+
+[[redirects]]
+  from = "https://19.nx.dev/*"
+  to = "https://v19.nx.dev/:splat"
+  status = 308
+  force = true
+
+[[redirects]]
+  from = "https://20.nx.dev/*"
+  to = "https://v20.nx.dev/:splat"
+  status = 308
+  force = true
+
+[[redirects]]
+  from = "https://21.nx.dev/*"
+  to = "https://v21.nx.dev/:splat"
+  status = 308
+  force = true
+
+[[redirects]]
+  from = "https://22.nx.dev/*"
+  to = "https://v22.nx.dev/:splat"
+  status = 308
+  force = true
--- a/scripts/create-versioned-docs.mts
+++ b/scripts/create-versioned-docs.mts
@@ -358,9 +358,12 @@ if (latestTag) {
 // --- Resolve GITHUB_TOKEN (needed for GitHub stars in docs build) ---
 if (!process.env.GITHUB_TOKEN) {
   try {
-    const token = execSync("op read 'op://Employee/API Keys/github_token'", {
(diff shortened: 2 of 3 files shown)


nx-dev/nx-dev/netlify.toml
```diff
@@ -23,3 +23,39 @@ included_files = [
   "!node_modules/webpack/**",
   "!node_modules/sass/**",
 ]
+
+[[redirects]]
+  from = "https://16.nx.dev/*"
```
jaysoo, line 28: These will need to land in prod to be verified. We can point the DNS entry to Netlify instead of Vercel to test if it works or not. Unfortunately there's no UI in Netlify to set these.

Should be low risk since the versioned docs aren't that important, and we can easily revert DNS changes if things don't work properly.

scripts/create-versioned-docs.mts
```diff
@@ -0,0 +1,443 @@
...
+    .filter((t) => t && !t.includes('-'));
+
+  if (tags.length === 0) return null;
+
+  tags.sort((a, b) => {
+    const pa = a.split('.').map(Number);
+    const pb = b.split('.').map(Number);
+    for (let i = 0; i < Math.max(pa.length, pb.length); i++) {
+      const diff = (pa[i] ?? 0) - (pb[i] ?? 0);
+      if (diff !== 0) return diff;
```
barbados-clemens, line 91: will this work with the beta.x versions? or do we need to handle that at all here?

e.g. 
```
20.1.0-beta.0
20.1.0-beta.1
20.1.0-beta.2
20.1.0-beta.3
20.1.0-beta.4
20.1.0-beta.5
```
git tags?

scripts/create-versioned-docs.mts
```diff
@@ -0,0 +1,443 @@
...
+    `No stable tags found for major version ${majorVersion}. Using current branch: ${currentRef}`
+  );
+}
+
+// --- Resolve GITHUB_TOKEN (needed for GitHub stars in docs build) ---
+if (!process.env.GITHUB_TOKEN) {
+  try {
+    const token = execSync("op read 'op://Employee/API Keys/github_token'", {
+      cwd: ROOT,
+    })
```
barbados-clemens: nit: might want to add the account so it doesn't error on peoples who have personal 1p accounts as defaults. 
ref PR where we did this in Ocean: https://github.com/nrwl/ocean/pull/10638

scripts/create-versioned-docs.mts
jaysoo: No, this will not work... it's based on stable release using git tags. I think that's fine? We'll never intentionally version a beta.

scripts/create-versioned-docs.mts
jaysoo: yeah that's fair

PR #35268, opened 2026-04-12
chore(core): remove `identity-obj-proxy` dependency

gameroman:
The `identity-obj-proxy` dependency can be removed as it is a simple 3 lines of code


AgentEnder:
I'm going to go ahead and just close this one out - its 3 lines and taking a bit too long to turn green - the latest CI failures look like the inlined version isn't working, and I don't think theres actually a reason to spend time on this right now

Thanks

Issue #35270, opened 2026-04-12
Support pnpm 11 multi-document lockfiles when using Bun as runtime


benpsnyder:
## Title

`nx/js/dependencies-and-lockfile` fails on pnpm 11 multi-document `pnpm-lock.yaml` files produced by `managePackageManagerVersions`

## Summary

We are adopting this split model in our Nx monorepo:

- `pnpm 11` is the single source of truth for installs and lockfile updates
- `Bun` is the default runtime for workspace commands like `bun x nx ...` and `bun run ...`

That model works, except for one Nx incompatibility:

When `pnpm-workspace.yaml` enables `managePackageManagerVersions: true`, pnpm 11 writes a multi-document `pnpm-lock.yaml`. Nx then fails while building the project graph because the `nx/js/dependencies-and-lockfile` plugin appears to assume the lockfile is a single YAML document.

## Observed failure

Command:

```bash
bun x nx sync
```

Error:

```text
 NX   Failed to process project graph.

An error occurred while processing files for the nx/js/dependencies-and-lockfile plugin.
  - pnpm-lock.yaml: expected a single document in the stream, but found more
```

## Why this matters

This blocks a practical setup that seems increasingly reasonable:

- pnpm handles package management and lockfile policy
- Bun handles execution/runtime speed
- Nx is invoked via `bun x nx ...`

Without this fix, we had to disable a pnpm 11 feature purely to keep Nx working:

```yaml
managePackageManagerVersions: false
```

That is a workaround, not a real fix.

## Repro

Repo config:

`package.json`

```json
{
  "packageManager": "pnpm@11.0.0-rc.0",
  "engines": {
    "pnpm": ">=11.0.0-rc.0 <12",
    "bun": ">=1.3.12"
  }
}
```

`nx.json`

```json
{
  "cli": {
    "packageManager": "pnpm"
  }
}
```

`pnpm-workspace.yaml`

```yaml
managePackageManagerVersions: true
packageManagerStrictVersion: true
sharedWorkspaceLockfile: true
```

Steps:

```bash
pnpm install
bun x nx sync
```

## What pnpm 11 writes

The generated lockfile becomes multi-document YAML. The first document contains package-manager metadata and the second contains the workspace lock data.

Example shape:

```yaml
---
lockfileVersion: '9.0'

importers:
  .:
    configDependencies: {}
    packageManagerDependencies:
      '@pnpm/exe':
        specifier: 11.0.0-rc.0
        version: 11.0.0-rc.0
[...]

PR #35271, opened 2026-04-12
fix(core): support pnpm multi-document lockfiles

benpsnyder:
## Current Behavior

Before this change, `nx/js/dependencies-and-lockfile` assumed `pnpm-lock.yaml` contained a single YAML document.

When `pnpm-workspace.yaml` enables `managePackageManagerVersions: true`, pnpm 11 writes a package-manager metadata document before the workspace lock document. Nx then fails while building the project graph with:

```text
An error occurred while processing files for the nx/js/dependencies-and-lockfile plugin.
- pnpm-lock.yaml: expected a single document in the stream, but found more
```

## Expected Behavior

After this change, Nx reads multi-document pnpm lockfiles, selects the workspace lockfile document, and continues parsing dependencies normally.

This allows monorepos to keep pnpm 11 package-manager metadata enabled while still running Nx commands such as `bun x nx sync` without disabling `managePackageManagerVersions`.

## Related Issue(s)

Fixes Issue https://github.com/nrwl/nx/issues/35270

## Validation

- `pnpm exec jest packages/nx/src/plugins/js/lock-file/pnpm-parser.spec.ts --config packages/nx/jest.config.cts --runInBand`
- Verified `getPnpmLockfileNodes` against the repository multi-document `pnpm-lock.yaml` via direct source invocation


benpsnyder:
Downstream validation report for this PR: https://gist.github.com/benpsnyder/4056f4634af8fb8e2fe40006e0d9cfe9

This patch fixed the failure in a real pnpm 11 workspace that is using pnpm-managed runtimes.

Downstream setup that reproduced the issue:

- `package.json`: `"packageManager": "pnpm@11.0.0-rc.0"`
- `package.json`: `devEngines.runtime` for both `bun@1.3.12` and `node@^24.10.0`
- `pnpm-workspace.yaml`: `managePackageManagerVersions: true`
- `pnpm-workspace.yaml`: `packageManagerStrictVersion: true`
- `pnpm-workspace.yaml`: `sharedWorkspaceLockfile: true`
- `pnpm-workspace.yaml`: `allowBuilds.bun: true`
- `pnpm-workspace.yaml`: downstream `patchedDependencies` entry pointing at the patch in this gist

Observed lockfile shape after `pnpm install`:

- first YAML document contains `configDependencies` and `packageManagerDependencies`
- second YAML document contains the actual workspace lockfile
- root importer contains `bun: runtime:1.3.12` and `node: runtime:24.14.1`
- packages contain `bun@runtime:1.3.12` and `node@runtime:24.14.1`

Verification results from the downstream workspace:

- `pnpm install` succeeded
- `pnpm exec bun --version` => `1.3.12`
- `pnpm exec node --version` => `v24.14.1`
- `bun x nx show projects --json | jq 'length'` => `207`

One pnpm 11/Bun nuance I hit while validating this: if `bun` is also kept as a normal root `devDependency`, pnpm did not materialize `bun@runtime:*` in the lockfile and `pnpm exec bun` resolved to the dependency install instead of the managed runtime. Removing the root `bun` package dependency and relying on `devEngines.runtime` fixed that. `allowBuilds.bun: true` still had to remain because the Bun runtime uses the `bun` package name for build approval.

The gist includes:

- the exact downstream patch file used against `nx@22.7.0-pr.35019.f500fd7`
- an evidence pack with commands and outcomes
- workspace and GitHub Actions snippets showing how this was wired in a pnpm workspace
- full environment details for the machine and toolchain used during validation

pnpm-workspace.yaml
```diff
@@ -10,6 +10,7 @@ packages:
   - 'examples/angular-rspack/module-federation/host'
   - 'examples/angular-rspack/module-federation/remote'
   - 'packages/nx/native-packages/*'
+managePackageManagerVersions: false
```
leosvelperez: This is not applicable to the Nx repo yet, isn't it? It's a pnpm v11 setting, and this repo still uses v10.

packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
```diff
@@ -268,6 +268,76 @@ type LockfileFileV9 = Omit<InlineSpecifiersLockfile, 'importers' | 'packages'> &
...
+
+function isPnpmLockfileDocument(
+  value: unknown
+): value is LockfileFile | LockfileFileV9 {
+  return (
+    typeof value === 'object' && value !== null && 'lockfileVersion' in value
+  );
+}
+
+function scorePnpmLockfileDocument(
```
leosvelperez: Pnpm seems to have a trivial positional solution based on a two-document format: https://github.com/pnpm/pnpm/blob/main/lockfile/fs/src/yamlDocuments.ts. We should follow the same approach instead of using a custom heuristic. I understand a position-less heuristic could account for future reorders or changes to the pnpm format, but future changes could also break the heuristic's assumptions.
Let's keep it simple and follow pnpm's own parsing. Let's add a reference to the relevant permalink(s).

leosvelperez (changes requested):
Thanks for the contribution! Let's simplify some things, and we should be good to go.

pnpm-workspace.yaml
benpsnyder: Reverted

packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
benpsnyder: Agreed, switched to pnpm's own positional approach. `parseAndNormalizePnpmLockfile` now mirrors `extractMainDocument` from `lockfile/fs/src/yamlDocuments.ts`: if the file starts with `---\n`, read the content after the next `\n---\n` separator; otherwise parse as a single document. Dropped `selectPnpmLockfileDocument`, `isPnpmLockfileDocument`, and `scorePnpmLockfileDocument` (~60 lines), replaced with a ~13-line `extractMainLockfileDocument` helper. Added permalink references to the pnpm source in comments above the helper.

All 34 `pnpm-parser.spec.ts` cases still pass, including the `pnpm 11 multi-document lockfiles` test.

leosvelperez (approved):
Thanks!

PR #35282, opened 2026-04-14
fix(angular-rspack): speed up watch-mode rebuilds

### LINKED ISSUE #34936: angular-rspack: poor local dev incremental build performance

leosvelperez:
## Current Behavior

Watch-mode rebuilds in `@nx/angular-rspack` are significantly slower than expected — often several times slower than the Angular CLI's esbuild-based dev server on the same project. On large apps, a single-file change can take 10+ seconds to rebuild even though only one file changed.

Three root causes:

1. `buildAndAnalyze` eagerly runs `JavaScriptTransformer.transformData()` on **every** file returned by `emitAffectedFiles()`. Due to Angular's `.ngtypecheck.ts` shim version mismatch between the first build (placeholder content) and subsequent rebuilds (real analysis content), TypeScript's incremental builder reports all ~7800 files as affected — so all ~7800 files go through the expensive transform on every rebuild.
2. `ComponentStylesheetBundler` is constructed with `incremental: false`, which disables result caching, selective invalidation, and esbuild's incremental rebuild mode for component styles.
3. `JavaScriptTransformer.close()` is called in the `emit` hook, which fires on every compilation — tearing down and re-spawning the worker thread pool on each rebuild.

## Expected Behavior

Watch-mode rebuilds perform on par with the Angular CLI's esbuild pipeline. A single-file change should only transform the changed file and reuse cached output for everything else.

Changes:

- **Lazy transform** — `buildAndAnalyze` now stores raw emitted content in a `rawEmitCache` and invalidates matching transformed cache entries only when the raw content actually changed. The Angular transform loader runs `transformData()` lazily when Rspack requests a module, mirroring esbuild's `onLoad` pattern. Files Rspack doesn't re-process are never re-transformed.
- **Incremental stylesheet bundler** — `ComponentStylesheetBundler` is now constructed with `incremental: !!options.watch` and `invalidate(modifiedFiles)` is called on rebuilds, matching the Angular CLI's `setup-bundling.js` behavior.
- **Persistent worker pool** — `JavaScriptTransformer.close()` moved from the per-build `emit` hook to the `shutdown` hook so workers live across rebuilds (mirrors esbuild plugin's `onDispose`).

Measured on a benchmark app with ~7800 files (Angular 21.2):

| Pipeline                | Warm rebuild |
| ----------------------- | ------------ |
| rspack (before)         | ~12s         |
| rspack (after)          | ~3.6s        |
| esbuild (Angular CLI)   | ~3.8s        |

## Related Issue(s)

Fixes #34936

### DIFF
--- a/packages/angular-rspack-compiler/src/compilation/build-and-analyze.ts
+++ b/packages/angular-rspack-compiler/src/compilation/build-and-analyze.ts
@@ -4,6 +4,13 @@ import { AngularCompilation } from '../models';
 
 const JS_TS_FILE_PATTERN = /\.[cm]?[jt]sx?$/;
 
+/**
+ * Raw JS/TS content emitted by the Angular compilation, keyed by normalized
+ * filename. The Rspack loader transforms entries on demand so only files the
+ * bundler actually requests pay the JavaScript transformation cost.
+ */
+export const rawEmitCache = new Map<string, string>();
+
 export async function buildAndAnalyze(
   angularCompilation: AngularCompilation,
   typescriptFileCache: Map<string, string | Uint8Array>,
@@ -15,24 +22,23 @@ export async function buildAndAnalyze(
   } of await angularCompilation.emitAffectedFiles()) {
     const normalizedFilename = normalize(filename.replace(/^[A-Z]:/, ''));
 
-    // Skip JavaScript transformation for non-JS/TS files (JSON, CSS, etc.)
-    // Let Rspack handle these through its native module rules
+    const text =
+      typeof contents === 'string'
+        ? contents
+        : Buffer.from(contents).toString();
+
     if (!JS_TS_FILE_PATTERN.test(normalizedFilename)) {
-      const text =
-        typeof contents === 'string'
-          ? contents
-          : Buffer.from(contents).toString();
       typescriptFileCache.set(normalizedFilename, text);
       continue;
     }
 
-    await javascriptTransformer
-      .transformData(normalizedFilename, contents, true, false)
-      .then((contents) => {
-        typescriptFileCache.set(
-          normalizedFilename,
-          Buffer.from(contents).toString()
-        );
-      });
+    // Drop the previous transformed output when the raw emit changes so the
+    // loader produces a fresh transform on next access.
+    const previousRaw = rawEmitCache.get(normalizedFilename);
+    rawEmitCache.set(normalizedFilename, text);
+
+    if (previousRaw !== text) {
+      typescriptFileCache.delete(normalizedFilename);
+    }
   }
 }
--- a/packages/angular-rspack-compiler/src/compilation/setup-compilation.ts
+++ b/packages/angular-rspack-compiler/src/compilation/setup-compilation.ts
@@ -27,6 +27,7 @@ export interface SetupCompilationOptions {
   hasServer?: boolean;
   includePaths?: string[];
   sass?: Sass;
+  watch?: boolean;
 }
 
 export const DEFAULT_NG_COMPILER_OPTIONS: ts.CompilerOptions = {
@@ -113,7 +114,7 @@ export async function setupCompilation(
       tailwindConfiguration,
     },
     options.inlineStyleLanguage,
-    false
+    !!options.watch
   );
 
   return {
--- a/packages/angular-rspack-compiler/src/compilation/setup-with-angular-compilation.ts
+++ b/packages/angular-rspack-compiler/src/compilation/setup-with-angular-compilation.ts
@@ -25,6 +25,14 @@ export async function setupCompilationWithAngularCompilation(
     options.hasServer,
     false
   );
+
+  // Drop the bundler's cached results for files the watcher reported as
+  // modified so dependent stylesheets get rebuilt; skipped on the initial
+  // build when there's nothing to invalidate.
+  if (modifiedFiles) {
+    componentStylesheetBundler.invalidate(modifiedFiles);
+  }
+
   modifiedFiles ??= new Set(rootNames);
 
   const fileReplacements: Record<string, string> =
--- a/packages/angular-rspack/src/lib/models/augmented-compilation.ts
+++ b/packages/angular-rspack/src/lib/models/augmented-compilation.ts
@@ -10,6 +10,7 @@ export const NG_RSPACK_SYMBOL_NAME = 'NG_RSPACK_BUILD';
 export type NG_RSPACK_COMPILATION_STATE = {
   javascriptTransformer: JavaScriptTransformer;
   typescriptFileCache: SourceFileCache['typeScriptFileCache'];
+  rawEmitCache: Map<string, string>;
   i18n?: I18nOptions;
 };
 export type NgRspackCompilation = Compilation & {
--- a/packages/angular-rspack/src/lib/plugins/angular-rspack-plugin.ts
+++ b/packages/angular-rspack/src/lib/plugins/angular-rspack-plugin.ts
@@ -6,6 +6,7 @@ import {
 } from '@angular/build/private';
 import {
   buildAndAnalyze,
+  rawEmitCache,
   DiagnosticModes,
   JavaScriptTransformer,
   SourceFileCache,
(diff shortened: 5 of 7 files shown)


Timebutt:
Hey @leosvelperez! I was very excited about your changes and when reading them I thought 'this is it!' so wanted to try things out as fast as possible. I couldn't find any build artifacts in the Nx CI pipeline so built everything from scratch and dropped the updated versions of `@nx/angular-rspack` and `@nx/angular-rspack-compiler` into the `node_modules` of the big project I'm working on. I added some `console.log()` tracing statements to ensure what I'm seeing is using the new implementation.

Unfortunately I'm not seeing the performance difference you're mentioning :/ I'll do some averaging for rebuild times with the new implementation and then roll-back and compare with what it was before, but if there is a change it's definitely not close to the expected `3x` or `4x` faster. I'm hoping I made a mistake in setting up my local dev with the new changes but the tracing I added seems to disprove that.

I also still observe a long delay between saving a file and actually seeing the CLI pick up the file change, maybe that's where a lot of time is lost? I have an app that takes `40s` for a rebuild, and most of it is just waiting for the CLI to react in any way (as in: showing the progress bar for instance).

Sometimes I'm also seeing a situation in which even when I change and save a file, nothing is picked up by the CLI. I'm not sure if after a very (!) long time the change gets picked up, or if that was after I hit Control+C but it did pick it up eventually. Very long time is 2min+, longer than the initial build.

Is there a chance something else (an earlier relevant change) is missing in your branch? If not, is there a published pre-release version I can use instead of what I built, to exclude the fact I may have made a mistake?

Apart from that: I'll gladly provide you with any information you might need from my setup to be able to debug this issue further.

dan-winters:
I second that this PR is fairly important for the plugin. Is there any concern on addtional memory usage with the rawEmitCache ? Coming from a rather large repository space, this was my first thought; perhaps it's unfounded since this in only for local development?

leosvelperez:
@Timebutt I tested it against the repro you provided (https://github.com/Timebutt/ng-bundler-benchmark/tree/add-local-dev-configurations), but there was a bug in the benchmark script I used. I've rerun everything manually with more care, and you're right: this isn't improving anything. I'm closing it and will keep investigating.

dan-winters:
@leosvelperez unsure if this helps, but from my experience the first watch mode rebuild is slow but it's significantly faster on additional updates

PR #35283, opened 2026-04-14
fix(release): surface swallowed publish errors when stdout is not valid JSON

ekkolon:
## Current Behavior

When `npm publish` or `pnpm publish` fails, the executor assumes `err.stdout` (always) contains valid JSON. If a lifecycle script (e.g. `prepublishOnly`) fails, it writes plaintext to stdout instead. This causes `JSON.parse` to throw and the actual error to be swallowed by the outer catch, printing only a generic "something unexpected went wrong" message.

## Expected Behavior

Wrap `JSON.parse(err.stdout...)` in a try/catch block. If parsing fails, fall back to logging raw stderr/stdout directly and return early, so lifecycle script failures and other non-JSON errors are always visible to the user.

## Related Issue(s)

Fixes #34497

### DIFF
--- a/packages/js/src/executors/release-publish/release-publish.impl.ts
+++ b/packages/js/src/executors/release-publish/release-publish.impl.ts
@@ -456,7 +456,18 @@ Please update the local dependency on "${depName}" to be a valid semantic versio
         };
       }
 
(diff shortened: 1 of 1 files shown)


leosvelperez (approved):
Thank you!

PR #35285, opened 2026-04-14
fix(linter): properly support bulk suppressions in lint executor

### LINKED ISSUE #35284: feat(@nx/eslint): bulk suppression options (suppressAll, suppressRule) do not work with the lint executor
## Issue

The `@nx/eslint:lint` executor accepts `suppressAll`, `suppressRule`, and `suppressionsLocation` options (added in schema.json), but **these options have no effect** at runtime. The executor passes them to the ESLint constructor, but ESLint's programmatic API silently ignores them — suppression handling only exists in the ESLint CLI.

This means:
- Running `nx run <project>:lint` with `suppressAll: true` does NOT create an `eslint-suppressions.json` file
- Existing `eslint-suppressions.json` files are NOT applied when linting through Nx
- Users must fall back to running `eslint` directly to use bulk suppressions

### Root Cause

1. **ESLint v9.x**: The `SuppressionsService` class in `eslint/lib/cli.js` handles writing/applying suppressions, but it is only used by the CLI. The `ESLint` constructor silently ignores `suppressAll`, `suppressRule`, and `suppressionsLocation` options.

2. **ESLint v10.x**: The programmatic API gained `applySuppressions: true` as a proper option that makes `lintFiles()` apply suppressions from the suppressions file. However, `suppressAll`/`suppressRule` (for *writing* suppressions) remain CLI-only.

### Expected Behavior

- `suppressAll: true` → writes all errors to `eslint-suppressions.json` 
- `suppressRule: ['no-console']` → writes violations for specified rules to `eslint-suppressions.json`
- `pruneSuppressions: true` → removes stale entries from `eslint-suppressions.json`
- When `eslint-suppressions.json` exists → suppressed errors should be filtered from results

### Current Behavior

All suppression options are silently ignored. Lint results are identical whether or not suppression options are set.

### Fix

I have a PR ready that:

1. **ESLint v10+**: Sets `applySuppressions: true` on the ESLint constructor (enables programmatic suppression support)
2. **ESLint v9.x**: Manually integrates `SuppressionsService` post-lint, mirroring what the ESLint CLI does
3. **All versions**: Handles `suppressAll`/`suppressRule` (writing) and `pruneSuppressions` (pruning) post-lint
4. Adds `pruneSuppressions` and `passOnUnprunedSuppressions` executor options (matching ESLint CLI flags)

### Related

- ESLint suppressions docs: https://eslint.org/docs/latest/use/suppressions
- ESLint v10 `applySuppressions` option: https://eslint.org/docs/latest/integration/nodejs-api#-new-eslintoptions

aqeelat:
## Summary

Fixes #35284

The `@nx/eslint:lint` executor accepts `suppressAll`, `suppressRule`, and `suppressionsLocation` options but they have no effect at runtime because ESLint's programmatic API silently ignores them.

### `eslint-utils.ts`
- For ESLint v10+: Set `applySuppressions: true` on the ESLint constructor, enabling programmatic suppression support
- Add `applySuppressions()` helper that integrates `SuppressionsService` post-lint (for writing, pruning, and applying suppressions)
- Add `getSuppressionsFilePath()` helper for resolving the suppressions file path
- Return `isEslintV10` flag from `resolveAndInstantiateESLint()`
- **Extract `validateSuppressionOptions()`** and **`resolveSuppressionsService()`** into named functions for testability and reuse
- **Restore deleted JSDoc comments**: `errorOnUnmatchedPattern`, `noEslintrc`/`useEslintrc`, quiet filter, and type cleanup comments — these explain *why* the code exists and prevent future regressions
- **Fix `ESLint?.version` null access crash** when the ESLint class cannot be resolved (e.g., in certain test environments)
- **Fix `getSuppressionsFilePath` bug**: custom `suppressionsLocation` was silently ignored when `SuppressionsService` couldn't be required, falling back to the default filename instead
- **Fix `getSuppressionsFilePath` return type**: changed from `string | null` to `string` (it never actually returned null)
- **Fix `suppress()` empty array bug**: `SuppressionsService.suppress(results, [])` suppresses nothing; normalize empty `suppressRule` to `undefined` when `suppressAll` is true
- **Add JSDoc architectural comments** on `resolveAndInstantiateESLint`, `getSuppressionsFilePath`, and `applySuppressions` explaining the v9/v10 suppression handling split and why ESLint internals are used

### `lint.impl.ts`
- Add post-lint suppression handling after `eslint.lintFiles()`:
  - **ESLint v9.x**: Always apply suppressions from the file if it exists (since v9's programmatic API doesn't support it)
  - **All versions**: Handle `suppressAll`/`suppressRule` (writing) and `pruneSuppressions` (pruning) post-lint
  - Support `passOnUnprunedSuppressions` to fail on unused suppressions
- **Add early validation** of mutually exclusive suppression options before any ESLint work, providing clear error messages at the executor level instead of deep in the utility layer
- **Guard post-lint suppressions behind ESLint >=9.24.0 check** (review feedback: `!isEslintV10` was true for v8 too, which would crash)

### `schema.json` / `schema.d.ts`
- Add `pruneSuppressions` option (equivalent to `--prune-suppressions` CLI flag)
- Add `passOnUnprunedSuppressions` option (equivalent to `--pass-on-unpruned-suppressions` CLI flag)

### Tests
- Update existing test mocks to include new schema fields and exports
- Add 7 new tests in `lint.impl.spec.ts` covering suppression behavior for v9.x and v10
- Add 8 new unit tests in `eslint-utils.spec.ts`:
  - `validateSuppressionOptions`: 5 tests for all mutual exclusivity combinations
  - `getSuppressionsFilePath`: 3 tests for custom location, fallback, and null-safety

## Version handling design

The suppression handling differs between ESLint versions:

- **ESLint v9.24.0+**: The programmatic API (`new ESLint({ suppressAll: true })`) silently ignores suppression options. We must call `SuppressionsService` directly post-lint to write suppressions, prune unused entries, and apply existing suppressions to results.
- **ESLint v10.0.0+**: Supports `applySuppressions: true` on the constructor, which causes `lintFiles()` to automatically apply suppressions during linting. Writing (suppressAll/suppressRule) and pruning still happen post-lint.

**Note**: Both paths use ESLint's internal `SuppressionsService` (`eslint/lib/services/suppressions-service`). The methods `suppress()`, `prune()`, `load()`, and `applySuppressions()` are not part of ESLint's public API and could change between versions.

## Manual Test Results

Tested with ESLint v9.39.4 and v10.3.0:

| # | Scenario | v9 (9.39.4) | v10 (10.3.0) |
|---|----------|-------------|--------------|
| 1 | Baseline lint (1 error) | PASS | PASS |
| 2 | `suppressAll` creates file, exits 0 | PASS | PASS |
| 3 | Re-run with existing suppressions (auto-apply) | PASS | PASS |
| 5 | `suppressRule` for specific rule | PASS | PASS |
| 6 | Custom `suppressionsLocation` | PASS | PASS |
| 7 | `pruneSuppressions` removes stale entries | PASS | PASS |
| 8 | `passOnUnprunedSuppressions` fails on unused | PASS | PASS |
| 9 | `suppressAll` + `suppressRule` mutual exclusivity error | PASS | PASS |
| 10 | `suppressAll` + `pruneSuppressions` mutual exclusivity error | PASS | PASS |
| 11 | `suppressRule` + `pruneSuppressions` mutual exclusivity error | PASS | PASS |

### Bug found during testing

`SuppressionsService.suppress(results, [])` with an empty `suppressRule` array suppresses nothing — it must be `undefined` for "suppress all". Fixed by normalizing empty arrays to `undefined`.

---

*This PR description was generated and updated with the assistance of AI (GLM-5.1 via z.ai Coding Plan).*

### DIFF
--- a/packages/eslint/src/executors/lint/lint.impl.spec.ts
+++ b/packages/eslint/src/executors/lint/lint.impl.spec.ts
@@ -65,10 +65,10 @@ jest.mock('./utility/eslint-utils', () => {
   return {
     resolveAndInstantiateESLint: (...args: any[]) =>
       mockResolveAndInstantiateESLint(...args),
-    applySuppressions: (...args: any[]) =>
-      mockApplySuppressions(...args),
+    applySuppressions: (...args: any[]) => mockApplySuppressions(...args),
     getSuppressionsFilePath: (...args: any[]) =>
       mockGetSuppressionsFilePath(...args),
+    validateSuppressionOptions: jest.fn(),
   };
 });
 import lintExecutor from './lint.impl';
@@ -1082,7 +1082,9 @@ Please see https://nx.dev/recipes/tips-n-tricks/eslint for full guidance on how 
     it('should call applySuppressions for ESLint v9.x when suppressions file exists', async () => {
       setupMocks();
       MockESLint.version = '9.24.0';
-      mockGetSuppressionsFilePath.mockReturnValue('/tmp/eslint-suppressions.json');
+      mockGetSuppressionsFilePath.mockReturnValue(
+        '/tmp/eslint-suppressions.json'
+      );
       (fs.existsSync as jest.Mock).mockImplementation((p: string) => {
         if (p === '/tmp/eslint-suppressions.json') return true;
         return realFs.existsSync(p);
@@ -1120,7 +1122,9 @@ Please see https://nx.dev/recipes/tips-n-tricks/eslint for full guidance on how 
     it('should call applySuppressions with suppressAll when suppressAll is true', async () => {
       setupMocks();
       MockESLint.version = '9.24.0';
-      mockGetSuppressionsFilePath.mockReturnValue('/tmp/eslint-suppressions.json');
+      mockGetSuppressionsFilePath.mockReturnValue(
+        '/tmp/eslint-suppressions.json'
+      );
       (fs.existsSync as jest.Mock).mockImplementation((p: string) => {
         if (p === '/tmp/eslint-suppressions.json') return true;
         return realFs.existsSync(p);
@@ -1150,7 +1154,9 @@ Please see https://nx.dev/recipes/tips-n-tricks/eslint for full guidance on how 
     it('should call applySuppressions with pruneSuppressions when pruneSuppressions is true', async () => {
       setupMocks();
       MockESLint.version = '9.24.0';
-      mockGetSuppressionsFilePath.mockReturnValue('/tmp/eslint-suppressions.json');
+      mockGetSuppressionsFilePath.mockReturnValue(
+        '/tmp/eslint-suppressions.json'
+      );
       (fs.existsSync as jest.Mock).mockImplementation((p: string) => {
         if (p === '/tmp/eslint-suppressions.json') return true;
         return realFs.existsSync(p);
@@ -1180,7 +1186,9 @@ Please see https://nx.dev/recipes/tips-n-tricks/eslint for full guidance on how 
     it('should fail when passOnUnprunedSuppressions is true and there are unused suppressions', async () => {
       setupMocks();
       MockESLint.version = '9.24.0';
-      mockGetSuppressionsFilePath.mockReturnValue('/tmp/eslint-suppressions.json');
+      mockGetSuppressionsFilePath.mockReturnValue(
+        '/tmp/eslint-suppressions.json'
+      );
       (fs.existsSync as jest.Mock).mockImplementation((p: string) => {
         if (p === '/tmp/eslint-suppressions.json') return true;
         return realFs.existsSync(p);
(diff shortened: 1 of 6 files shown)


wilcoschoneveld:
What is the status on this PR? Support for bulk suppressions would be very useful for us! @aqeelat @FrozenPandaz @leosvelperez

leosvelperez:
@aqeelat sorry for the delay getting to this one. Could you please rebase and resolve the existing conflicts so we can review with the correct final changeset?

While updating it, please also check/address the following:

- Align `passOnUnprunedSuppressions` with ESLint. It should allow unused suppressions, but the current implementation reverses that behavior and requires pruning.
- Rework the ESLint 10 pruning path. It currently receives already-suppressed results, which can cause active suppression entries to be removed as unused.
- Make the default suppression-file behavior match the schema. The implementation currently uses the workspace root while the schema documents the project root.
- Update the existing “Bulk Suppression” examples in `packages/eslint/docs/eslint-examples.md`. They persist `suppressAll` and `suppressRule` in the lint target, so later violations are also suppressed. These should be shown as one-time command overrides.
- Add broader behavioral tests for the supported suppression workflows.

PR #35288, opened 2026-04-14
fix(js): resolve build output dir from globbed outputs in node executor

### LINKED ISSUE #35301: @nx/js:node serve crashes with Cannot find module '<lib>/dist/**/*.{js,cjs,mjs,...}{,.map}' when serving an app that depends on a buildable TS lib (TS-solution preset)
### Current Behavior

  In a TS-solution workspace (`create-nx-workspace --preset=ts`):

  - `apps/api` — NestJS app generated with `@nx/nest:app` (webpack + tsc, served via `@nx/js:node`)
  - `packages/my-lib` — buildable TS library generated with `@nx/js:lib --bundler=tsc`
  - `apps/api/src/app/app.service.ts` imports a function from `@<scope>/my-lib`

  `nx build api` succeeds. `nx serve api` compiles, forks the child node process, and crashes on startup:

  ```
  Error: Cannot find module '/abs/path/packages/my-lib/dist/**/*.{js,cjs,mjs,jsx,d.ts,d.cts,d.mts}{,.map}'
  Require stack:
  - /abs/path/apps/api/dist/main.js
      at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1207:15)
      at Function.Module._load (node:internal/modules/cjs/loader:1038:27)
      at Function.Module._load (.../node_modules/@nx/js/src/executors/node/node-with-require-overrides.js:15:31)
      at Module.require (node:internal/modules/cjs/loader:1289:19)
      at require (node:internal/modules/helpers:182:18)
      at Array.__webpack_modules__ (.../apps/api/dist/external commonjs "@<scope>/my-lib":1:1)
  ```

  The literal glob pattern ends up as the argument to `require()`.

### Expected Behavior

  `nx serve api` should boot. The library should resolve either via its compiled output directory or via the symlinked workspace package in `node_modules`. The runtime require-override
  should never point `require()` at a glob.

  ## Root Cause

I'm fairly new to Nx and don't have deep tooling experience, so take this with a grain of salt, as it came out of an AI-assisted investigation, not my own expertise. Posting it in case it's a useful starting point.

  The chain as I understand it:

  1. `@nx/js/typescript`'s Crystal plugin infers a `build` target for the lib whose first output is a glob (`{projectRoot}/dist/**/*.{js,cjs,mjs,jsx,d.ts,d.cts,d.mts}{,.map}`), designed
  for the cache hasher — not a real filesystem path.
  2. Having a `build` target makes `@nx/webpack` classify the lib as buildable and externalize it via `webpack-node-externals`, so the bundle ends up with `module.exports =
  require("@<scope>/my-lib")`.
  3. `@nx/js:node`'s `calculateResolveMappings` (in `@nx/js/src/executors/node/node.impl.js`) takes `c.outputs[0]` — the glob from step 1 — joins it with the workspace root, and puts it in
   `NX_MAPPINGS`.
  4. `node-with-require-overrides.js` rewrites the runtime `require('@<scope>/my-lib')` to that glob string, and Node's CJS resolver throws `MODULE_NOT_FOUND`.

  My guess: `calculateResolveMappings` was written back when `outputs[0]` was always a concrete directory, and Crystal-style glob outputs quietly break that assumption.

### GitHub Repo

https://github.com/gg-martins091/nx-bug-repro

### Steps to Reproduce

This script reproduces the crash from scratch (uses `@latest` for `create-nx-workspace` and matches `@nx/nest`/`@nx/webpack`/`@nx/js` to whatever `nx` version the preset installs):

  ```bash
  #!/usr/bin/env bash
  set -euo pipefail

  REPRO_DIR="${REPRO_DIR:-$HOME/nx-bug-repro}"
  [ -e "$REPRO_DIR" ] && { echo "$REPRO_DIR exists"; exit 1; }

  PARENT_DIR="$(dirname "$REPRO_DIR")"
  NAME="$(basename "$REPRO_DIR")"

  cd "$PARENT_DIR"
  npx --yes create-nx-workspace@latest "$NAME" --preset=ts --pm=npm --nxCloud=skip
  cd "$REPRO_DIR"

  NX_VERSION="$(node -p "require('./package.json').devDependencies.nx")"
  npm i -D "@nx/nest@$NX_VERSION" "@nx/webpack@$NX_VERSION" "@nx/js@$NX_VERSION"

  npx nx g @nx/nest:app apps/api --linter=none --unitTestRunner=none --no-interactive
  npx nx g @nx/js:lib packages/my-lib --bundler=tsc --linter=none --unitTestRunner=none --no-interactive

  cat > apps/api/src/app/app.service.ts <<EOF
  import { Injectable } from '@nestjs/common';
  import { myLib } from '@$NAME/my-lib';

  @Injectable()
  export class AppService {
    getData() {
      return { message: myLib() };
    }
  }
  EOF

  npx nx serve api
  ```

  Inspect the offending shape directly without running the full repro:

  ```sh
  npx nx show project @<scope>/my-lib --json | jq '.targets.build.outputs'
  # first entry is a glob, not a directory or file path
  ```

  Minimal reproduction of the crash with *no Nx in the loop at all* — just Node plus the 20-line require-override — after `nx build api`:

  ```sh
  NX_FILE_TO_RUN="$PWD/apps/api/dist/main.js" \
  NX_MAPPINGS='{"@<scope>/my-lib":"'"$PWD"'/packages/my-lib/dist/**/*.{js,cjs,mjs,jsx,d.ts,d.cts,d.mts}{,.map}"}' \
    node node_modules/@nx/js/src/executors/node/node-with-require-overrides.js
  ```

  Same `MODULE_NOT_FOUND` on the glob path. Proves the entire chain collapses to: the executor hands a glob to `require()`.

  ## Workarounds tried

  - **Make the lib source-only** (`main`/`types`/every `exports` condition → `./src/index.ts`, mirroring how `@nx/js:lib` generates a lib *without* `--bundler=tsc`).
  `isValidPackageJsonBuildConfig` then returns `false`, no `build` target is inferred, the lib is allowlisted, webpack bundles its sources inline, and the runtime require-override never fires. Works, but requires changing the lib's package.json shape and dropping `rootDir` from the consuming app's `tsconfig.app.json` so `ts-loader` will transpile cross-project sources. (This seems hacky and misses the point of a buildable lib?!?)

### Nx Report

```shell
Node           : 22.14.0
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.9.2
daemon         : Available

nx                     : 22.6.4
@nx/js                 : 22.6.4
@nx/eslint             : 22.6.4
@nx/workspace          : 22.6.4
@nx/jest               : 22.6.4
@nx/devkit             : 22.6.4
@nx/eslint-plugin      : 22.6.4
@nx/module-federation  : 22.6.4
@nx/nest               : 22.6.4
@nx/node               : 22.6.4
@nx/playwright         : 22.6.4
@nx/react              : 22.6.4
@nx/rollup             : 22.6.4
@nx/vite               : 22.6.4
@nx/vitest             : 22.6.4
@nx/web                : 22.6.4
@nx/webpack            : 22.6.4
@nx/docker             : 22.6.4
Registered Plugins:
@nx/js/typescript
@nxlv/python
@nx/vite/plugin
@nx/webpack/plugin
@nx/eslint/plugin
@nx/jest/plugin
@nx/playwright/plugin
Community plugins:
@nxlv/python : 22.1.1
Cache Usage: 323.44 KB / 25.09 GB
```

### Failure Logs

```shell

```

### Package Manager Version

npm 10.9.2

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

leosvelperez:
## Current Behavior

The `@nx/js:node` executor fails when combined with the inferred `@nx/js/typescript` build target — in two distinct ways:

1. **Script-to-run resolution**: fails with `Could not find <project>/dist/**/*.{js,cjs,mjs,jsx,d.ts,d.cts,d.mts}{,.map}/main.js. Make sure your build succeeded.` — the executor's `getFileToRun` was appending `main.js` onto the glob output pattern.
2. **Buildable-dep import resolution**: when the node app imports a buildable lib that also uses the inferred build target, `require('@scope/my-lib')` throws `MODULE_NOT_FOUND` because `calculateResolveMappings` passes the literal glob into `NX_MAPPINGS`, which the CJS require override / ESM loader then tries to resolve.

Both regressed after #35041 narrowed the inferred build target's `outputs[0]` from `{projectRoot}/dist` to `{projectRoot}/dist/**/*.{js,...}{,.map}` (to prevent cross-OS cache pollution).

## Expected Behavior

`outputs` entries are cache patterns and may legitimately contain globs. The node executor now strips the glob portion back to the last path separator before using the value as a directory, in both `getFileToRun` and `calculateResolveMappings`. Handles `**`, `*`, `?`, character classes, brace expansion, extglob, and Windows/POSIX separators.

## Related Issue(s)

Fixes #35198
Fixes #35301

### DIFF
--- a/packages/js/src/executors/node/node.impl.ts
+++ b/packages/js/src/executors/node/node.impl.ts
@@ -453,7 +453,13 @@ function getFileToRun(
         projectRoot: project.data.root,
         workspaceRoot: context.root,
       });
-      return path.join(outputFilePath, 'main.js');
+      // `outputs` are cache patterns and may contain globs (e.g. the inferred
+      // `@nx/js/typescript` build target scopes its output to
+      // `{projectRoot}/dist/**/*.{js,...}` to avoid caching non-tsc files).
+      // Strip the glob portion back to the last path separator before it to
+      // recover the base output directory.
+      const outputDir = stripGlobToBaseDir(outputFilePath);
+      return path.join(outputDir, 'main.js');
     }
     const fallbackFile = path.join('dist', project.data.root, 'main.js');
 
@@ -483,6 +489,16 @@ function getFileToRun(
   }
 
   return join(context.root, buildOptions.outputPath, outputFileName);
+}
+
+function stripGlobToBaseDir(pathWithGlob: string): string {
+  const globIdx = pathWithGlob.search(/[*?[{(]/);
+  if (globIdx === -1) {
+    return pathWithGlob.replace(/[\\/]+$/, '');
+  }
+  const prefix = pathWithGlob.slice(0, globIdx);
+  const lastSep = Math.max(prefix.lastIndexOf('/'), prefix.lastIndexOf('\\'));
+  return lastSep === -1 ? '' : prefix.slice(0, lastSep);
 }
 
 function fileToRunCorrectPath(fileToRun: string): string {
--- a/packages/js/src/executors/node/node.impl.ts
+++ b/packages/js/src/executors/node/node.impl.ts
@@ -405,7 +405,11 @@ function calculateResolveMappings(
   );
   return dependencies.reduce((m, c) => {
     if (c.node.type !== 'npm' && c.outputs[0] != null) {
-      m[c.name] = joinPathFragments(context.root, c.outputs[0]);
+      // `outputs` are cache patterns and may contain globs (e.g. from the
+      // inferred `@nx/js/typescript` build target). Strip the glob portion
+      // so the runtime require overrides resolve to the actual output dir.
+      const outputDir = stripGlobToBaseDir(c.outputs[0]);
+      m[c.name] = joinPathFragments(context.root, outputDir);
     }
     return m;
   }, {});


gg-martins091:
Hey @leosvelperez, thanks for this fix.

I just wanted to confirm something, as you closed the [issue ](https://github.com/nrwl/nx/issues/35301) I opened as duplicate of [35198](https://github.com/nrwl/nx/issues/35198), however, taking a closer look they seem to be different issues.

My issue is only when the app imports a buildable lib. Running the nestjs api with `serve` works fine if there is no imports of the buildable lib.

The issue you mentioned is affected directly when running the app and not specifically when it imports a buildable lib.

The stack traces also differ, [mine](https://github.com/nrwl/nx/issues/35301) is related to how imports are resolved and the one this PR fixes is related to how scripts to run are resolved.

Just wanted to make sure that I understand this correctly, since I tested these changes locally and still see the same error.

gg-martins091:
> The stack traces also differ, [mine](https://github.com/nrwl/nx/issues/35301) is related to how imports are resolved and the one this PR fixes is related to how scripts to run are resolved.

Using your `stripGlobToBaseDir` in calculateResolveMappings fixed my issue locally, I just dont know what else it could break... Cloning the repo right now to run the tests.

```diff
   function calculateResolveMappings(
     context: ExecutorContext,
     options: NodeExecutorOptions
   ): Record<string, string> {
     const parsed = parseTargetString(options.buildTarget, context);
     const { dependencies } = calculateProjectBuildableDependencies(/* … */);
     return dependencies.reduce((m, c) => {
       if (c.node.type !== 'npm' && c.outputs[0] != null) {
  -      m[c.name] = joinPathFragments(context.root, c.outputs[0]);
  +      // Same treatment as getFileToRun above — `outputs[0]` may be a glob
  +      // from the inferred `@nx/js/typescript` build target. Without stripping
  +      // it, the runtime `require('@scope/my-lib')` is rewritten to the literal
  +      // glob string via NX_MAPPINGS and Node throws MODULE_NOT_FOUND.
  +      const outputDir = stripGlobToBaseDir(c.outputs[0]);
  +      m[c.name] = joinPathFragments(context.root, outputDir);
       }
       return m;
     }, {} as Record<string, string>);
   }
  ```

leosvelperez:
@gg-martins091 good catch! You're right. The issue is caused by the same underlying root cause but it requires an additional fix. I'll update this PR with the fix for what you reported. Thanks for the investigation!

FrozenPandaz (approved):

PR #35289, opened 2026-04-14
fix(core): don't hang when workspace contains a named pipe

leosvelperez:
## Current Behavior

If a workspace contains a named pipe (FIFO) — e.g. created via `mkfifo` — any `nx` command that walks the workspace (e.g. `nx show projects`) hangs forever with no output. The native file hasher calls `std::fs::read` on the FIFO, which blocks indefinitely waiting for a writer.

## Expected Behavior

Special files that cannot be meaningfully hashed as workspace content (FIFOs, sockets, block/char devices) are skipped by the native walker, so `nx` commands complete normally.

## Implementation Notes

- Filter added in `nx_walker` (parallel and wasm variants) so only regular files and symlinks are emitted as `NxFile` entries. Uses the metadata already fetched for `mod_time`, so there is no extra syscall cost.
- Symlinks are still emitted to preserve existing symlink-follow behavior via `std::fs::read`.
- Added a unix-only regression test (`skips_named_pipes`) that creates a FIFO with `nix::unistd::mkfifo` and asserts the walker skips it.

## Related Issue(s)

Fixes #35228

### DIFF
--- a/packages/nx/src/native/walker.rs
+++ b/packages/nx/src/native/walker.rs
@@ -86,6 +86,10 @@ where
             return None;
         };
 
+        if !is_hashable_file(&metadata.file_type()) {
+            return None;
+        }
+
         Some(NxFile {
             full_path: String::from(dir_entry.path().to_string_lossy()),
             normalized_path: file_path.to_normalized_string(),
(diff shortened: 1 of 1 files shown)


AgentEnder (approved):

huonw:
Thanks for fixing! 😄 

> Symlinks are still emitted to preserve existing symlink-follow behavior via std::fs::read.

I think a symlink could theoretically point to a non-hashable file, like a fifo; for example:

```shell
mkfifo test
ln -s test test2
```

I think `test` (FIFO) will be skipped, but I'm not sure about `test2` (symlink to FIFO).

Is that handled?

(Other edge cases include symlinks to symlinks to symlinks to FIFO, etc.)

PR #35293, opened 2026-04-14
fix(core): preserve FORCE_COLOR=0 intent for forked child tasks

### LINKED ISSUE #35292: FORCE_COLOR=0 is deleted then re-introduced as 'true' for forked child tasks
## Current Behavior

When `FORCE_COLOR=0` is set in the environment (common in CI to suppress ANSI codes), Nx deletes it in `bin/nx.ts` and then re-introduces `FORCE_COLOR=true` for every forked child task. This causes all child processes (ESLint, TypeScript, Vitest, etc.) to emit ANSI escape codes even though the user explicitly requested no colors.

### Root Cause

The workaround for [picocolors#100](https://github.com/alexeyraspopov/picocolors/issues/100) introduced in PR #34520 deletes `FORCE_COLOR` from the parent process, but `task-env.ts` defaults `undefined` to `'true'`:

**Step 1 — `bin/nx.ts` (lines 4-8):**
```typescript
if (process.env.FORCE_COLOR === '0') {
  process.env.NO_COLOR = '1';
  delete process.env.FORCE_COLOR; // FORCE_COLOR is now undefined
}
```

**Step 2 — `task-env.ts` (line 19) and `task-orchestrator.ts` (lines 681, 1000):**
```typescript
process.env.FORCE_COLOR === undefined ? 'true' : process.env.FORCE_COLOR
// Since FORCE_COLOR was deleted → evaluates to 'true'
```

**Step 3 — `task-env.ts` (line 76):**
```typescript
FORCE_COLOR: forceColor, // 'true' is injected into every child's env
```

The child process receives both `FORCE_COLOR=true` and `NO_COLOR=1`, but per the [force-color spec](https://force-color.org/) and [Node.js docs](https://nodejs.org/api/cli.html), `FORCE_COLOR` takes precedence over `NO_COLOR`.

## Expected Behavior

`FORCE_COLOR=0` should propagate to child tasks, resulting in no ANSI escape codes in their output.

## Impact

This affects any CI environment that sets `FORCE_COLOR=0` to get parseable plain-text output. In our case, ANSI escape codes in ESLint output broke downstream log parsers that use regex to extract lint errors (the `✖ N problems (M errors, O warnings)` summary line contains invisible ANSI bytes that prevent matching).

## Suggested Fix

Preserve the user's original `FORCE_COLOR` intent before deleting it, and consult the saved value when building child environments:

```typescript
// bin/nx.ts — save intent before deleting
if (process.env.FORCE_COLOR === '0') {
  process.env.NX_ORIGINAL_FORCE_COLOR = '0';
  process.env.NO_COLOR = '1';
  delete process.env.FORCE_COLOR;
}
```

```typescript
// task-env.ts / task-orchestrator.ts — check saved value
process.env.FORCE_COLOR === undefined
  ? (process.env.NX_ORIGINAL_FORCE_COLOR === '0' ? '0' : 'true')
  : process.env.FORCE_COLOR
```

## Workaround

Setting `FORCE_COLOR=false` (string) instead of `FORCE_COLOR=0` bypasses the `=== '0'` check in `bin/nx.ts`, so the value passes through to children unchanged. However, `'false'` is not a standard `FORCE_COLOR` value per either [force-color.org](https://force-color.org/) or the [chalk/supports-color convention](https://github.com/chalk/supports-color).

## Environment

- Nx version: 21.x (any version after PR #34520, merged 2025-02-25)
- Node.js: 22+
- OS: Linux (CI), macOS (local)

🤖 _This issue was drafted with the help of AI_

comp615:
> **Note:** An alternative to this PR is to wait for the upstream [picocolors#100](https://github.com/alexeyraspopov/picocolors/issues/100) fix and then remove the `FORCE_COLOR=0` workaround in `bin/nx.ts` entirely. However, picocolors has not had a release or meaningful activity in over two years and appears to be unmaintained, so this fix ensures correct behavior in the meantime.

## Current Behavior

When `FORCE_COLOR=0` is set (common in CI to suppress ANSI codes), `bin/nx.ts` deletes it from `process.env` as a workaround for [picocolors#100](https://github.com/alexeyraspopov/picocolors/issues/100). However, `task-env.ts` and `task-orchestrator.ts` then default the undefined `FORCE_COLOR` to `'true'` for all forked child tasks, re-enabling ANSI colors in every child process.

```
User sets FORCE_COLOR=0
  → bin/nx.ts deletes FORCE_COLOR, sets NO_COLOR=1
  → task-env.ts sees FORCE_COLOR === undefined → defaults to 'true'
  → child processes receive FORCE_COLOR=true (overrides NO_COLOR)
  → ANSI escape codes in ESLint, TypeScript, Vitest output
```

## Expected Behavior

`FORCE_COLOR=0` should propagate to child tasks, resulting in no ANSI escape codes.

## Solution

1. **`bin/nx.ts`**: Save the original value in `NX_ORIGINAL_FORCE_COLOR` before deleting
2. **`task-env.ts`**: Extract a `getForceColorForChild()` helper that checks `NX_ORIGINAL_FORCE_COLOR` when `FORCE_COLOR` is undefined
3. **`task-orchestrator.ts`**: Use the shared helper at both call sites

## Impact

This affects any CI environment that sets `FORCE_COLOR=0`. In our case, ANSI escape codes in ESLint output broke downstream log parsers that use regex to extract lint errors (the `✖ N problems` summary line contains invisible ANSI bytes that prevent matching).

## Tests

Added 4 unit tests for `getForceColorForChild()` covering:
- Explicit `FORCE_COLOR` passthrough
- `FORCE_COLOR=0` → deleted → restored via `NX_ORIGINAL_FORCE_COLOR`
- Default to `'true'` when neither is set
- `FORCE_COLOR` takes precedence over `NX_ORIGINAL_FORCE_COLOR`

Fixes #35292

🤖 _This PR was drafted with the help of AI_

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -537,7 +537,8 @@
             "NX_PREFIX_OUTPUT",
             "NX_INFER_ALL_PACKAGE_JSONS",
             "NX_AI_FILES_USE_LOCAL",
-            "NX_WINDOWS_PTY_SUPPORT"
+            "NX_WINDOWS_PTY_SUPPORT",
+            "NX_ORIGINAL_FORCE_COLOR"
           ]
         }
       }
--- a/packages/nx/bin/nx.ts
+++ b/packages/nx/bin/nx.ts
@@ -4,6 +4,7 @@
 // See: https://github.com/alexeyraspopov/picocolors/issues/100
 
 if (process.env.FORCE_COLOR === '0') {
+  process.env.NX_ORIGINAL_FORCE_COLOR = '0';
   process.env.NO_COLOR = '1';
   delete process.env.FORCE_COLOR;
 }
--- a/packages/nx/src/tasks-runner/task-env.spec.ts
+++ b/packages/nx/src/tasks-runner/task-env.spec.ts
@@ -6,6 +6,7 @@ import { Task } from '../config/task-graph';
 import {
   getEnvFilesForTask,
   getEnvVariablesForTask,
+  getForceColorForChild,
   loadAndExpandDotEnvFile,
 } from './task-env';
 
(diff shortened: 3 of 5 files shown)


comp615:
@FrozenPandaz FYI

Also note that I think it might be OK if force-color true wasn't set in CI

AgentEnder (approved):

AgentEnder (approved):

PR #35294, opened 2026-04-14
fix(angular-rspack): fixes issues with angular-rspack hmr

dan-winters:
HMR is freezing on compilation failures but should support recompilation when updating files after a previous compilation error

## Current Behavior
When running dev server with HMR, compilation errors are freezing the dev server

## Expected Behavior
HMR should not abort or freeze and should allow recompilation when saving updated files.

Fixes #35040


AgentEnder (approved):

leosvelperez (approved):
Thanks!

PR #35248, opened 2026-04-10
feat(core): add json input type for selective JSON field hashing

FrozenPandaz:
## Current Behavior

When configuring task inputs for cache hashing, Nx hashes entire files. For JSON config files like `tsconfig.json` or `package.json`, changing any field invalidates the cache — even fields irrelevant to the task (e.g., changing `description` in package.json invalidates a build task).

The only special case is tsconfig, which has hardcoded selective hashing in the native hasher.

## Expected Behavior

A new `json` input type allows users and plugins to specify exactly which fields from a JSON file should be included in the hash. This enables more granular cache invalidation.

```jsonc
// Only hash the "engines" field from package.json
{ "json": "{projectRoot}/package.json", "fields": ["engines"] }

// Hash all of compilerOptions except paths
{ "json": "{workspaceRoot}/tsconfig.json", "fields": ["compilerOptions"], "excludeFields": ["compilerOptions.paths"] }
```

### Features
- **`{workspaceRoot}` and `{projectRoot}` tokens** — same syntax as `fileset` inputs
- **Glob support** — e.g. `{projectRoot}/tsconfig*.json`
- **Dot notation** — nested field paths like `compilerOptions.target`
- **Allowlist (`fields`) and denylist (`excludeFields`)** — can be used together
- **Deterministic hashing** — canonical JSON serialization with sorted keys

### Files changed
- **TypeScript**: `InputDefinition` type + JSON schemas for IDE support
- **Rust NAPI bridge**: `JsonInput` struct, `Either8` → `Either9`, `Input::Json` variant
- **Rust hash planning**: `HashInstruction::JsonFileSet`, planner emits it from `gather_self_inputs`
- **Rust hasher**: new `hash_json.rs` with field filtering, canonical serialization, and 10 unit tests

PR #35298, opened 2026-04-15
fix: sort object keys before JSON.stringify in transformProjectGraphForRust

### LINKED ISSUE #35297: Non-deterministic ProjectConfiguration hash due to JSON.stringify key-order sensitivity in transformProjectGraphForRust
## Current Behavior

When rerunning a failed CI pipeline (Nx Cloud DTE), tasks that previously **succeeded** are re-executed because their `ProjectConfiguration` hash has changed between runs — even though **no code was modified**.

The `nx tasks compare` tool shows that `ProjectConfiguration` hashes differ between runs for affected projects:

```
"management-integration:ProjectConfiguration": "10538082656014434668"
"management-modules-video-recordings:ProjectConfiguration": "10839486168096120338"
```

## Root Cause

The function `transformProjectGraphForRust` in `packages/nx/src/native/transform-objects.ts` serializes target `options` and `configurations` using `JSON.stringify()`:

```ts
targets[targetName] = {
  executor: targetConfig.executor,
  inputs: targetConfig.inputs,
  outputs: targetConfig.outputs,
  options: JSON.stringify(targetConfig.options),
  configurations: JSON.stringify(targetConfig.configurations),
  parallelism: targetConfig.parallelism,
};
```

`JSON.stringify()` preserves **JavaScript object key insertion order**. The Rust native hasher receives these serialized strings and hashes them as-is. If any Nx plugin or the target-defaults merge step produces `options`/`configurations` objects with keys in a different insertion order on different CI runners, the serialized string changes, which causes the `ProjectConfiguration` hash to change.

### Why the Rust hasher only partially compensates

The Rust hasher **does** sort target names alphabetically before hashing, making the hash independent of target insertion order. However, it does **not** normalize the content of the stringified `options`/`configurations` — these are opaque strings by the time they reach Rust.

## Proof

Tested locally using the native `HashPlanner` + `TaskHasher` APIs:

**Same options, different key order → different `ProjectConfiguration` hash:**

| Options string | ProjectConfiguration hash |
|---|---|
| `{"cwd":"...","env":{...},"command":"jest","passWithNoTests":true}` | `10839486168096120338` |
| `{"passWithNoTests":true,"command":"jest","env":{...},"cwd":"..."}` | `14689511571626992510` |

**With a `sortObjectKeys` fix applied before `JSON.stringify`:**

Reversed ALL target orders AND ALL options/configurations key orders for ALL projects across the entire project graph → **530 `ProjectConfiguration` entries, 0 mismatches, identical task hashes.**

## Expected Behavior

The `ProjectConfiguration` hash should be deterministic regardless of JS object key insertion order in target `options`/`configurations`.

## Suggested Fix

Recursively sort object keys before stringifying in `transformProjectGraphForRust`:

```ts
function sortObjectKeys(obj: unknown): unknown {
  if (obj === null || obj === undefined || typeof obj !== 'object') {
    return obj;
  }
  if (Array.isArray(obj)) {
    return obj.map(sortObjectKeys);
  }
  const sorted: Record<string, unknown> = {};
  for (const key of Object.keys(obj).sort()) {
    sorted[key] = sortObjectKeys((obj as Record<string, unknown>)[key]);
  }
  return sorted;
}

// In transformProjectGraphForRust:
targets[targetName] = {
  executor: targetConfig.executor,
  inputs: targetConfig.inputs,
  outputs: targetConfig.outputs,
  options: JSON.stringify(sortObjectKeys(targetConfig.options)),
  configurations: JSON.stringify(sortObjectKeys(targetConfig.configurations)),
  parallelism: targetConfig.parallelism,
};
```

Alternatively, the normalization could be done on the Rust side before hashing.

## Environment

- **Nx version**: 22.6.3
- **Node version**: 22.x
- **CI**: GitHub Actions + Nx Cloud DTE
- **OS**: Linux (CI runners)

## Related

This is analogous to the fix in commit `db31f30` which sorted keys before hashing for `AllExternalDependencies` via `hashObject`. The same principle applies here — any data fed to a hasher must be order-normalized.

omer-za:
## Summary

Fixes #35297

`JSON.stringify()` preserves JavaScript object key insertion order. In `transformProjectGraphForRust`, target `options` and `configurations` are serialized with `JSON.stringify()` and passed to the Rust native hasher, which hashes them as-is for `ProjectConfiguration`.

If any Nx plugin or the target-defaults merge step produces these objects with keys in a different insertion order on different CI runners, the serialized string changes, causing the `ProjectConfiguration` hash to become **non-deterministic**. This leads to:

- Nx Cloud cache invalidation on CI reruns
- Previously succeeded tasks being re-executed unnecessarily
- Wasted CI compute time

## The Fix

Adds a recursive `sortObjectKeys` helper that normalizes key order before serialization, making the `ProjectConfiguration` hash independent of JS object construction order.

Note: The Rust hasher already sorts **target names** before hashing (making the hash independent of target insertion order), but it does not normalize the **content** of the stringified `options`/`configurations`. This fix closes that gap.

## Evidence

Tested locally using the native `HashPlanner` + `TaskHasher` APIs:

### Without fix
| Options string | ProjectConfiguration hash |
|---|---|
| `{"cwd":"...","env":{...},"command":"jest","passWithNoTests":true}` | `10839486168096120338` |
| `{"passWithNoTests":true,"command":"jest","env":{...},"cwd":"..."}` | `14689511571626992510` |

### With fix
Reversed ALL target orders AND ALL options/configurations key orders for ALL projects across the entire project graph:
- **530 `ProjectConfiguration` entries, 0 mismatches**
- **Identical task hashes**

## Related

This is analogous to the fix in commit `db31f30` which sorted keys before hashing for `AllExternalDependencies` via `hashObject`.

### DIFF
--- a/packages/nx/src/native/tests/transform-objects.spec.ts
+++ b/packages/nx/src/native/tests/transform-objects.spec.ts
@@ -1,0 +1,117 @@
+import { transformProjectGraphForRust } from '../transform-objects';
+import { ProjectGraph } from '../../config/project-graph';
+
+/**
+ * `transformProjectGraphForRust` feeds the Rust hasher. Any difference in the
+ * shape it returns — including key insertion order inside JSON-stringified
+ * fields — produces a different ProjectConfiguration hash, which causes
+ * spurious cache misses across CI runs that built the same logical graph in
+ * different key orders (e.g. plugins iterating Map vs. Object, or async
+ * resolution races).
+ *
+ * These tests assert that the transform is invariant to key insertion order
+ * for every field that ends up in the hash input.
+ */
+describe('transformProjectGraphForRust', () => {
+  function reverseKeys<T>(value: T): T {
+    if (Array.isArray(value)) {
+      return value.map(reverseKeys) as unknown as T;
+    }
+    if (value && typeof value === 'object') {
+      const out: Record<string, unknown> = {};
+      for (const key of Object.keys(value).reverse()) {
+        out[key] = reverseKeys((value as Record<string, unknown>)[key]);
+      }
+      return out as unknown as T;
+    }
+    return value;
+  }
+
+  function makeGraph(transform: <T>(v: T) => T = (v) => v): ProjectGraph {
+    const inputs = [
+      'default',
+      { runtime: 'node -v' },
+      { env: 'CI' },
+      { externalDependencies: ['typescript', 'react'] },
+    ];
+    const outputs = ['{workspaceRoot}/dist', '{projectRoot}/coverage'];
+    const options = {
+      foo: 1,
+      bar: 2,
+      baz: { x: 1, y: 2, z: 3 },
+    };
+    const configurations = {
+      development: { mode: 'dev', flag: true },
+      production: { mode: 'prod', flag: false },
+    };
+    const namedInputs = {
+      default: ['{projectRoot}/**/*'],
+      production: [
+        'default',
+        '!{projectRoot}/**/*.spec.ts',
+        { externalDependencies: ['react', 'typescript'] },
+      ],
+    };
+
+    return {
+      nodes: {
+        foo: {
+          name: 'foo',
+          type: 'lib',
+          data: {
+            root: 'libs/foo',
+            namedInputs: transform(namedInputs),
+            tags: ['scope:shared'],
+            targets: {
+              build: {
+                executor: 'nx:run-commands',
+                inputs: transform(inputs),
+                outputs: transform(outputs),
+                options: transform(options),
+                configurations: transform(configurations),
+                parallelism: false,
+              },
+            },
+          },
+        },
+      },
+      dependencies: { foo: [] },
+      externalNodes: {},
+    } as unknown as ProjectGraph;
+  }
+
+  it('produces an identical Rust input for two graphs that only differ in key insertion order', () => {
+    const a = transformProjectGraphForRust(makeGraph());
+    const b = transformProjectGraphForRust(makeGraph(reverseKeys));
+
+    expect(JSON.stringify(b)).toEqual(JSON.stringify(a));
+  });
+
+  it('normalises target.options key order (covers the original PR case)', () => {
+    const fwd = transformProjectGraphForRust(makeGraph());
+    const rev = transformProjectGraphForRust(makeGraph(reverseKeys));
+
+    expect(rev.nodes.foo.targets.build.options).toEqual(
+      fwd.nodes.foo.targets.build.options
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


omer-za:
Update — extended the fix to cover `inputs`, `outputs`, and `namedInputs`.

We rolled out the original two-field fix (`options`, `configurations`) as a `patch-package` patch in our monorepo and re-ran the same GitHub Actions workflow twice with identical commit SHAs. We were still seeing `ProjectConfiguration` cache misses on `typecheck` (and a handful of other targets) on the second run.

Root cause: three more fields that flow into the same `transformProjectGraphForRust` codepath were still preserving JS key insertion order:

- `target.inputs` — entries can be objects: `{ env }`, `{ runtime }`, `{ externalDependencies }`, `{ dependentTasksOutputFiles }`, `{ fileset }`
- `target.outputs` — rare, but objects are valid
- `projectNode.data.namedInputs` — top-level keys plus nested objects (same shape as `target.inputs`)

These get reordered between CI runs whenever a plugin iterates a `Map` vs. a plain object, async resolution races, or a `tsconfig.json` with `extends` is walked in a different order. The 5-field normalisation closes the remaining gap.

Latest commit:

- Applies `sortObjectKeys` to all five fields
- Adds `packages/nx/src/native/tests/transform-objects.spec.ts` with per-field determinism tests plus a full-graph reverse-key case

Local verification: applying the patched JS to our 1500-project workspace and running `nx graph` twice now produces a byte-identical `ProjectConfiguration` hash.

omer-za:
For anyone who wants to reproduce this from a clean slate, I built a small Nx 22 monorepo that triggers it deterministically:

→ https://github.com/omer-za/nx-cache-miss-repro

\`\`\`
git clone https://github.com/omer-za/nx-cache-miss-repro
cd nx-cache-miss-repro
npm install
npm run demo
\`\`\`

It's a real workspace (5 \`@nx/js\` libraries + the standard \`@nx/js/typescript\` plugin + a custom inference plugin in \`tools/plugins/\`). The custom plugin assembles its target's \`options\` and \`configurations\` objects in I/O completion order — the same way real plugins like \`@nx/jest\`, \`@nx/eslint\`, \`@nx/cypress\` build their target shapes when their internals fan out work via \`Promise.all\` / \`fs.readdir\` / async resolution.

The demo runs four scenarios and asserts:

| scenario | hashes match across runs | second \`nx run :deploy\` is a cache hit |
| --- | --- | --- |
| stock nx@22.6.3 | ❌ | ❌ |
| with this PR applied | ✅ | ✅ |

It exits non-zero if reality disagrees with that table, so it doubles as a regression test.

meeroslav:
This was resolved by https://github.com/nrwl/nx/pull/35248

PR #35304, opened 2026-04-15
feat(linter): default to ESLint v10 and migrate existing workspaces

leosvelperez:
## Current Behavior

Nx generators scaffold ESLint projects with ESLint v9 and its ecosystem. When run in an existing workspace, the same generators overwrite ecosystem packages with hardcoded v9 pins instead of respecting what the user already has installed.

Existing workspaces also have no path onto ESLint v10. `nx migrate` leaves them on v9, and upgrading by hand breaks them: v10 removed the eslintrc format along with the `ESLINT_USE_FLAT_CONFIG` escape hatch, so `.eslintrc.*` files stop being read at all, and several widely used plugins have no v10 release.

## Expected Behavior

- **New workspaces and projects** scaffold with ESLint v10 by default, along with a peer-compatible ecosystem (`typescript-eslint`, `@eslint/js`, `@eslint/eslintrc`, `@eslint/compat`, `eslint-plugin-vue`, `@vue/eslint-config-typescript`, etc.).
- **Existing workspaces on ESLint v9** move to v10 on `nx migrate`. A new `23.2.0` `packageJsonUpdates` group bumps `eslint` to `^10.0.0` and the `typescript-eslint` packages to `^8.58.0`, and a paired migration prepares the workspace for the new major.
- **Existing workspaces on ESLint v10** get ecosystem packages pinned to ranges compatible with ESLint v10.
- **React projects on ESLint v10** swap `eslint-plugin-import` for `eslint-plugin-import-x` and move `eslint-plugin-react-hooks` to v7. `eslint-plugin-react` and `eslint-plugin-jsx-a11y` are dropped because neither has an ESLint v10 release, so v10 React projects lint with a smaller rule set. ESLint v9 React projects keep the existing plugin set.
- In all cases, packages already installed by the user are **preserved verbatim**. Generators no longer silently upgrade installed versions.

Projects that get their React rules through `@nx/eslint-plugin` (`flat/react`, `flat/react-base`, `flat/react-jsx`) need no config change: the presets resolve `eslint-plugin-import-x` and pin the classic `rules-of-hooks` and `exhaustive-deps` rules on v10. A hand-written config that pulls `eslint-plugin-react-hooks`'s own `recommended` preset gets the v7 set instead, which grows from 2 rules to 16 and adds the React Compiler rules with 12 of them at error severity.

## Migration

`@nx/eslint` gains a hybrid `update-23-2-0-migrate-to-eslint-v10` migration, gated on the workspace landing on ESLint v10. Its deterministic half converts the eslintrc configs that are left and replaces the plugins that cannot run on v10. Everything that needs judgment goes to the paired prompt, and surfaces as next steps when the migration runs outside the agentic flow. Source files are never edited, and no rule is turned off to make lint pass.

> [!NOTE]
> Two framework integrations still pin ESLint v9 through their own peer ranges, so the update group carries `incompatibleWith` for `angular-eslint` below 21.3 and `eslint-config-next` below 16. Those workspaces stay on ESLint v9, which remains supported, and move to v10 once the integration does. `nx migrate` skips a held group silently, so such a workspace gets no output explaining why ESLint stayed where it was.

## Migration coverage

| Change in ESLint v10 | Treatment |
| --- | --- |
| eslintrc format and `ESLINT_USE_FLAT_CONFIG` removed | Converted through `@nx/eslint:convert-to-flat-config`. `.eslintrc.js` and `.eslintrc.cjs` are read statically rather than executed, so a config built from literals converts like any other. |
| A JavaScript eslintrc config that computes its values | Prompt. Those values only exist once the file runs, and executing a user config to read them is not something the migration can do safely. |
| Plugins with no v10 release (`eslint-plugin-import`, `eslint-plugin-jsx-a11y`, `eslint-plugin-react`, and `eslint-plugin-react-hooks` below 7) | Removed from `package.json`, with `eslint-plugin-import-x` installed in place of `eslint-plugin-import` and react-hooks moved to 7.1.1. |
| Configs and source files that reference a removed plugin | Prompt. Only the `import` rules have an equivalent to move to; a `react/` or `jsx-a11y/` rule has to be dropped or replaced case by case, including the ones named in directive comments. |
| `/* eslint-env */` comments now reported as errors | Prompt. The globals have to move into the flat config or into a `/* global */` comment, and which ones belong where depends on the file. |
| Node.js floor raised to `^20.19.0 \|\| ^22.13.0 \|\| >=24` | Reported. Widening `engines.node` is the workspace's call. |
| Installed plugins whose `peerDependencies.eslint` excludes v10 | Reported, read from `node_modules`. |
| Rule set otherwise unchanged | Nothing. v10 ships the same rules as v9. The two sets that do grow, `eslint:recommended` through `@eslint/js` v10 and react-hooks `recommended` through v7, are reported. |

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/nxc-4017-3807ed77)


gperdomor:
Please merge this 🙏🏼

dotrakoun-clearstreet:
https://github.com/es-tooling/eslint-plugin-jsx-a11y-x for eslint 10

PR #35320, opened 2026-04-17
fix(core): hash tasks before cache lookup to prevent null-hash crash on Cloud agents

FrozenPandaz:
## Current Behavior

Nx Cloud agents crash on warm-cache runs with:

```
Error: Failed to convert JavaScript value `Null` into rust type `String`
    at DbCache.getBatch (.../cache.js:115:41)
    at TaskOrchestrator.fetchCacheHits (.../task-orchestrator.js:270:47)
    at TaskOrchestrator.resolveCachedTasks (.../task-orchestrator.js:564:38)
    at runDiscreteTasks (.../init-tasks-runner.js:133:42)
  code: 'StringExpected'
```

Regression introduced by #35172 (warm-cache perf optimization). `init-tasks-runner.ts::createOrchestrator` calls `orchestrator.processAllScheduledTasks()` **without awaiting** — that method only sets up hashing promises in a private `processedTasks` map. `runDiscreteTasks` then immediately calls `resolveCachedTasks`, which hits `cache.getBatch(tasks.map(t => t.hash))`. Hashes are still `null` because the queued `processTask` → `hashTask` promises haven't resolved.

## Expected Behavior

Warm-cache agent runs resolve cached tasks without crashing.

Rather than papering over with another `await`, the underlying state machine is simplified. The `processedTasks: Map<string, Promise<NodeJS.ProcessEnv>>` field bundled three concerns into one shared promise (env memo, hash dedupe, `scheduleTask` lifecycle) and created the timing trap. Investigation showed it was now vestigial:

- `getTaskSpecificEnv` already memoizes internally via `taskSpecificEnvCache` (`task-env.ts:35-50`).
- No `LifeCycle` implementation in the codebase implements `scheduleTask`.
- `task.hash` itself is the natural gate for hash dedupe.

So `processedTasks`, `processTask`, and `processAllScheduledTasks` are deleted. Two single-purpose helpers replace them:

- `ensureHashes(tasks)` — batch-hashes any tasks lacking a hash; idempotent because `hashTasks` filters by `!task.hash` internally.
- `fireScheduleLifecycle(tasks)` — fires `scheduleTask` once per task, deduped by a `scheduledIds: Set<string>`.

Both run in coordinator step 1 (before any cache check or dispatch — preserves the pre-refactor "queued = scheduled" timing) and defensively in `resolveCachedTasks`, `runTaskDirectly`, `startContinuousTask`, and `applyFromCacheOrRunBatch`. Idempotent dedupe means safe-to-call from any path.

Bonus: coordinator step 1's hash guard `unhashed.length > 1` → `> 0`. The `> 1` was a micro-optimization that silently skipped cache lookup for single-task cycles (because `resolveCachedTasksBulk` filters by `task.hash &&`). Cost more in lost cache hits than it saved in batch setup.

Net: **−29 lines** in `task-orchestrator.ts` despite adding two helper docstrings.

## Related Issue(s)

No GitHub issue — surfaced internally on a CI run after upgrading Nx.

### DIFF
--- a/packages/nx/src/tasks-runner/init-tasks-runner.ts
+++ b/packages/nx/src/tasks-runner/init-tasks-runner.ts
@@ -175,6 +175,14 @@ export async function runDiscreteTasks(
     lifeCycle
   );
 
+  // Kick off hashing + scheduleTask lifecycle for every task immediately so
+  // they overlap with the batch dispatch loop below. Both helpers are
+  // idempotent (hashTasks filters !task.hash; lifecycle dedupes via Set), so
+  // the per-task defensive calls in runTaskDirectly stay cheap no-ops.
+  const prepared = orchestrator
+    .ensureHashes(tasks)
+    .then(() => orchestrator.fireScheduleLifecycle(tasks));
+
   let groupId = 0;
   let nextBatch = orchestrator.nextBatch();
   const batchResults: Array<Promise<TaskResult[]>> = [];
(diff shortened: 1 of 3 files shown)


FrozenPandaz:
Superseded by #35322 — minimal version of this fix. Closing in favor of the smaller change.

PR #35323, opened 2026-04-17
fix(release): scope fallback to project history for new packages

### LINKED ISSUE #35327: nx release scans entire git history for new packages without release tags
## Description

When using `nx release` with `conventionalCommits: true` and `automaticFromRef: true` in a monorepo, adding a new package causes the release process to scan the **entire repository git history** instead of only the commits relevant to the new package. This results in severe performance degradation, excessive warnings, and potential CI timeouts.

## Root Cause

When a project has no release tags (i.e., it was just added to the monorepo), two code paths fall back to `getFirstGitCommit()` — the very first commit in the repository:

1. **Version resolution** ([`derive-specifier-from-conventional-commits.ts#L43`](https://github.com/nrwl/nx/blob/master/packages/nx/src/command-line/release/version/derive-specifier-from-conventional-commits.ts#L43)):
   ```ts
   const previousVersionRef = latestMatchingGitTag
     ? latestMatchingGitTag.tag
     : fallbackCurrentVersionResolver === 'disk'
       ? await getFirstGitCommit()  // ← scans from the dawn of time
       : undefined;
   ```

2. **Changelog resolution** ([`version-plan-filtering.ts#L168`](https://github.com/nrwl/nx/blob/master/packages/nx/src/command-line/release/changelog/version-plan-filtering.ts#L168)):
   ```ts
   if (useAutomaticFromRef) {
     return await getFirstGitCommit();  // ← same issue
   }
   ```

This causes `git log <first-ever-commit>..HEAD --name-status` to process every commit in the repository, even for a package that was added yesterday.

## Reproduction

**`nx.json`:**
```json
{
  "release": {
    "projects": ["packages/*"],
    "releaseTagPattern": "release/{projectName}/{version}",
    "changelog": { "automaticFromRef": true },
    "version": {
      "conventionalCommits": true,
      "fallbackCurrentVersionResolver": "disk",
      "generatorOptions": {
        "currentVersionResolver": "disk",
        "specifierSource": "conventional-commits"
      }
    }
  }
}
```

**Steps:**
1. Have a monorepo with a long git history (1000+ commits)
2. Add a new package (e.g., `packages/my-new-lib/package.json` with `"version": "1.0.0"`)
3. Run `nx release --skip-publish`

**Observed behavior:**
- Release step takes **5+ minutes** scanning the entire history
- Dozens of warnings for every old/renamed/removed package referenced in ancient commits:
  ```
  The affected projects might have not been identified properly.
  The package(s) <old-package-name> were not found.
  ```
- In CI environments, this often exceeds pipeline timeout limits

**Expected behavior:**
- `nx release` should only scan commits since the project was first introduced
- No warnings for packages that predate the new project

## Real-World Impact

In our monorepo (~1200 commits, ~30 packages, 4 newly added packages):

| Metric | Current | Expected |
|--------|---------|----------|
| Commits scanned per new package | ~1200 | Only since package was added |
| "Package not found" warnings | ~30 per run | 0 |
| Release step time | 5m 30s | Seconds |
| CI pipeline | Intermittent timeouts | Stable |

## Proposed Fix

PR #35323 introduces `getFirstProjectCommit(projectRoot)` which uses:

```
git rev-list --reverse HEAD --first-parent -- <projectRoot>/package.json
```

This finds the first commit that touched the project's `package.json` and returns its parent (so the creation commit is included in `from..to` ranges). It's used as the fallback instead of `getFirstGitCommit()` for individual project version resolution and changelog generation.

Workspace-level and fixed release group changelogs continue using `getFirstGitCommit()` since they have no single project root to scope to.

The function gracefully falls back to `getFirstGitCommit()` if the project-scoped lookup fails, maintaining full backward compatibility.

## Workaround

Until this is resolved, users can create baseline release tags for new packages before running `nx release`:

```bash
# For each release project without a tag, create a baseline tag at HEAD
for project in $(npx nx show projects); do
  if [ -z "$(git tag -l "release/${project}/*")" ]; then
    version=$(node -p "require('./<project-root>/package.json').version")
    git tag "release/${project}/${version}"
  fi
done
```

## Environment

- **Nx version:** 22.5.3 (issue exists in all versions with `automaticFromRef`)
- **Node:** 24.x
- **Package manager:** pnpm 10.x
- **CI:** Amazon Linux arm64 — 15 CPUs, 26 GB RAM (c8g.4xlarge)
- **Local:** macOS Tahoe 26.4.1 — Apple M3 Pro, 18 GB RAM

zanlucathiago:
## Current Behavior

Closes #35327

When a project has no release tags (e.g., newly added to a monorepo), `nx release` falls back to `getFirstGitCommit()` — the very first commit in the **entire repository**. This means `git log <first-commit>..HEAD` scans every commit ever made, regardless of whether they are relevant to the new project.

In large monorepos this causes:

1. **Severe performance degradation** — the release step can take 5+ minutes just scanning irrelevant history
2. **Noisy warnings** — `"The affected projects might have not been identified properly. The package(s) X were not found"` for every old/renamed/removed package referenced in ancient commits
3. **CI timeouts** — the cumulative overhead can push pipelines past their timeout limits

### Reproduction

1. Create a monorepo with `release.changelog.automaticFromRef: true` and `release.version.conventionalCommits: true`
2. Have a long git history (1000+ commits) with commits referencing packages that no longer exist
3. Add a new package to the workspace
4. Run `nx release --skip-publish`
5. Observe it scanning the entire git history and logging warnings for every unresolvable package reference

## Expected Behavior

For new packages without release tags, `nx release` should only scan commits since the project was first introduced — not the entire repository history.

## Implementation

Adds `getFirstProjectCommit(projectRoot)` in `utils/git.ts`:

```
git rev-list --reverse HEAD --first-parent -- <projectRoot>/package.json
```

This finds the first commit that touched the project's `package.json` and returns its **parent**, so that `from..HEAD` ranges include the creation commit itself.

The function is used as the fallback in two places:

| Location | Before | After |
|----------|--------|-------|
| `deriveSpecifierFromConventionalCommits` | `getFirstGitCommit()` | `getFirstProjectCommit(projectRoot)` |
| `resolveChangelogFromSHA` | `getFirstGitCommit()` | `getFirstProjectCommit(projectRoot)` when `projectRoot` is provided |

For **workspace-level** and **fixed release group** changelogs (where there's no single project root), the existing `getFirstGitCommit()` behavior is preserved.

### Fallback safety

If `getFirstProjectCommit` fails for any reason (e.g., the project root doesn't contain a `package.json`, or the git command fails), it gracefully falls back to `getFirstGitCommit()` — maintaining backward compatibility.

## Impact

For a real-world monorepo with ~1200 commits and 4 new packages:

| Metric | Before | After |
|--------|--------|-------|
| Commits scanned per new package | ~1200 | Only commits since package was added |
| "Package not found" warnings | ~30 per run | 0 |
| Release step time | 5+ minutes | Seconds |

### DIFF
--- a/packages/nx/src/command-line/release/utils/git.ts
+++ b/packages/nx/src/command-line/release/utils/git.ts
@@ -748,7 +748,7 @@ export async function getFirstProjectCommit(
         'HEAD',
         '--first-parent',
         '--',
-        `${projectRoot}/package.json`,
+        projectRoot,
       ])
     ).trim();
     const firstCommit = result.split('\n')[0];
--- a/packages/nx/src/command-line/release/utils/git.ts
+++ b/packages/nx/src/command-line/release/utils/git.ts
@@ -732,6 +732,44 @@ export async function getFirstGitCommit() {
   }
 }
 
+/**
+ * Returns the parent of the first commit that touched the given project root,
+ * so that `from..HEAD` ranges include the project's creation commit.
+ * Falls back to getFirstGitCommit() if the project history cannot be determined.
+ */
+export async function getFirstProjectCommit(
+  projectRoot: string
+): Promise<string> {
+  try {
+    const result = (
+      await execCommand('git', [
+        'rev-list',
+        '--reverse',
+        'HEAD',
+        '--first-parent',
+        '--',
+        `${projectRoot}/package.json`,
+      ])
+    ).trim();
+    const firstCommit = result.split('\n')[0];
+
+    if (firstCommit) {
+      // Return the parent so the creation commit is included in from..to ranges
+      try {
+        return (
+          await execCommand('git', ['rev-parse', `${firstCommit}~1`])
+        ).trim();
+      } catch {
+        // No parent (project was added in the repo's very first commit)
+        return firstCommit;
+      }
+    }
+  } catch {
+    // fall through to fallback
+  }
+  return getFirstGitCommit();
+}
+
 async function getGitRoot() {
   try {
     return (await execCommand('git', ['rev-parse', '--show-toplevel'])).trim();
--- a/packages/nx/src/command-line/release/changelog.ts
+++ b/packages/nx/src/command-line/release/changelog.ts
@@ -370,7 +370,8 @@ export function createAPI(
       preid: string | undefined,
       checkAllBranchesWhen: CheckAllBranchesWhen,
       requireSemver: boolean,
-      strictPreid: boolean
+      strictPreid: boolean,
+      projectRoot?: string
     ): Promise<string | null> => {
       if (fromSHACache.has(cacheKey)) {
         return fromSHACache.get(cacheKey);
@@ -386,6 +387,7 @@ export function createAPI(
         requireSemver,
         strictPreid,
         useAutomaticFromRef,
+        projectRoot,
       });
       fromSHACache.set(cacheKey, sha);
       return sha;
@@ -491,7 +493,8 @@ export function createAPI(
               projectsPreid[project.name],
               releaseGroup.releaseTag.checkAllBranchesWhen,
               releaseGroup.releaseTag.requireSemver,
-              releaseGroup.releaseTag.strictPreid
+              releaseGroup.releaseTag.strictPreid,
+              project.data.root
             );
 
             let commits: GitCommit[];
--- a/packages/nx/src/command-line/release/changelog/version-plan-filtering.ts
+++ b/packages/nx/src/command-line/release/changelog/version-plan-filtering.ts
@@ -8,6 +8,7 @@ import { execCommand } from '../utils/exec-command';
 import {
   getCommitHash,
   getFirstGitCommit,
+  getFirstProjectCommit,
   getLatestGitTagForPattern,
 } from '../utils/git';
 import type { VersionData } from '../utils/shared';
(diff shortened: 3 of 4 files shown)


packages/nx/src/command-line/release/utils/git.ts
```diff
@@ -732,6 +732,44 @@ export async function getFirstGitCommit() {
...
+): Promise<string> {
+  try {
+    const result = (
+      await execCommand('git', [
+        'rev-list',
+        '--reverse',
+        'HEAD',
+        '--first-parent',
+        '--',
+        `${projectRoot}/package.json`,
```
JamesHenry: We don't want a JS specific marker file as the source of truth here, `nx release` and nx more broadly supports all languages and ecosystems. We should be able to use the directory itself here.

```suggestion
        projectRoot,
```

JamesHenry (changes requested):
Thanks a lot for this @zanlucathiago and I'm sorry for the delay in reviewing, just the one change needed please

packages/nx/src/command-line/release/utils/git.ts
zanlucathiago: Makes sense! 

`package.json` would've silently broken the fallback for non-JS projects. 

Pushed the change, thanks!

JamesHenry (approved):

PR #35336, opened 2026-04-17
fix(core): prevent daemon client wedge on failed reconnect

AgentEnder:
## Current Behavior

`DaemonClient` mutates `_daemonStatus` to `CONNECTING` before doing any async work, but several exit paths leave it stuck there without ever settling `_waitForDaemonReady`. Concurrent callers that land in the `CONNECTING` branch then `await` a promise that can never resolve or reject.

Two triggers in the wild:

1. **`startDaemonIfNecessary` DISCONNECTED branch** — if `isServerAvailable()` throws something other than `VersionMismatchError`, or `startInBackground()` throws (e.g. the `global.NX_PLUGIN_WORKER` guard), status stays `CONNECTING` and `_daemonReady` is never called. The first caller correctly sees the error; everyone else parks forever.
2. **`handleConnectionError`** — on `VersionMismatchError` or a failed reconnect, only the current in-flight message is rejected. The freshly-created `_waitForDaemonReady` is never settled and status stays `CONNECTING`.

In plugin workers (where `startInBackground` always throws) this manifested as `createNodes` hanging until the isolated-plugin 10-minute timeout fired. The end-user symptom was a daemon that reported "Reusing in-memory cached project graph because no files changed" and then never responded, because the cached promise itself was wedged on plugin-worker helper calls that would never return.

## Expected Behavior

- **State-machine hygiene:** every transition into `CONNECTING` is paired with a fresh `_waitForDaemonReady`, and every failure path rejects it and returns status to `DISCONNECTED`. Concurrent callers get the error; the next caller can retry.
- **Plugin-worker fast-fail:** `handleConnectionError` short-circuits when `global.NX_PLUGIN_WORKER` is set. A plugin worker can't coherently reconnect to a replacement daemon (graph state, cached context, version skew), so every queued and future message fails immediately with a clear *"plugin worker lost its daemon connection and cannot reconnect"* error instead of hanging until timeout.

## Implementation

* Added `_daemonReadyFailed` to pair with `_daemonReady`, and helpers `createReadyPromise()` / `failDaemonReady()` so every failure path can unblock awaiters in one place.
* Wrapped `startDaemonIfNecessary` DISCONNECTED body in try/catch → `failDaemonReady` on any error.
* Rejected `_waitForDaemonReady` from every failure branch of `handleConnectionError` (VersionMismatchError, non-VME throw, `!serverAvailable`).
* Short-circuited `handleConnectionError` for `NX_PLUGIN_WORKER`.
* Added `client.spec.ts` with five tests covering both wedge modes and the plugin-worker fast-fail. All five fail against `master` and pass after the fix.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/daemon/client/client.spec.ts
+++ b/packages/nx/src/daemon/client/client.spec.ts
@@ -1,0 +1,162 @@
+import { DaemonClient } from './client';
+import { VersionMismatchError } from './daemon-socket-messenger';
+
+declare global {
+  // eslint-disable-next-line no-var
+  var NX_PLUGIN_WORKER: boolean | undefined;
+}
+
+// Mirrors the private enum in client.ts for test assertions.
+const enum DaemonStatus {
+  CONNECTING = 0,
+  DISCONNECTED = 1,
+  CONNECTED = 2,
+}
+
+type TestClient = {
+  _daemonStatus: DaemonStatus;
+  _waitForDaemonReady: Promise<void> | null;
+  _daemonReady: (() => void) | null;
+  currentMessage: unknown;
+  currentResolve: ((v: unknown) => void) | null;
+  currentReject: ((e: unknown) => void) | null;
+  startDaemonIfNecessary: () => Promise<void>;
+  handleConnectionError: (err: Error) => Promise<void>;
+  isServerAvailable: () => Promise<boolean>;
+  startInBackground: () => Promise<number>;
+  setUpConnection: () => void;
+  waitForServerToBeAvailable: (opts: {
+    ignoreVersionMismatch: boolean;
+  }) => Promise<boolean>;
+  establishConnection: () => void;
+  registerDaemonProcessWithMetricsService: (
+    pid: number | null
+  ) => Promise<void>;
+};
+
+function asTest(client: DaemonClient): TestClient {
+  return client as unknown as TestClient;
+}
+
+describe('DaemonClient state machine', () => {
+  let client: TestClient;
+
+  beforeEach(() => {
+    client = asTest(new DaemonClient());
+    // Never actually register with metrics during tests.
+    jest
+      .spyOn(
+        client as unknown as { registerDaemonProcessWithMetricsService: any },
+        'registerDaemonProcessWithMetricsService'
+      )
+      .mockResolvedValue(undefined);
+  });
+
+  afterEach(() => {
+    delete global.NX_PLUGIN_WORKER;
+    jest.restoreAllMocks();
+  });
+
+  describe('startDaemonIfNecessary()', () => {
+    it('does not wedge concurrent callers when the initial connect throws', async () => {
+      const err = new Error('spawn failed');
+      jest.spyOn(client, 'isServerAvailable').mockResolvedValue(false);
+      jest.spyOn(client, 'startInBackground').mockRejectedValue(err);
+
+      // Kick off caller A synchronously; it flips status DISCONNECTED -> CONNECTING.
+      const first = client.startDaemonIfNecessary();
+      // Caller B now enters the CONNECTING branch and awaits _waitForDaemonReady.
+      const second = client.startDaemonIfNecessary();
+
+      await expect(first).rejects.toThrow('spawn failed');
+      await expect(second).rejects.toThrow('spawn failed');
+
+      // A failed attempt must leave the client DISCONNECTED so the next
+      // caller can retry instead of parking on a pending promise forever.
+      expect(client._daemonStatus).toBe(DaemonStatus.DISCONNECTED);
+    });
+
+    it('lets a future caller retry after a failed connect attempt', async () => {
+      const err = new Error('spawn failed');
+      const isServerAvailable = jest
+        .spyOn(client, 'isServerAvailable')
+        .mockResolvedValue(false);
+      const startInBackground = jest
+        .spyOn(client, 'startInBackground')
+        .mockRejectedValueOnce(err)
+        .mockResolvedValueOnce(1234);
+      jest.spyOn(client, 'setUpConnection').mockImplementation(() => {
+        /* no-op */
+      });
+
+      await expect(client.startDaemonIfNecessary()).rejects.toThrow(
+        'spawn failed'
+      );
+      expect(client._daemonStatus).toBe(DaemonStatus.DISCONNECTED);
+
+      await expect(client.startDaemonIfNecessary()).resolves.toBeUndefined();
+      expect(client._daemonStatus).toBe(DaemonStatus.CONNECTED);
+      expect(isServerAvailable).toHaveBeenCalledTimes(2);
+      expect(startInBackground).toHaveBeenCalledTimes(2);
+    });
+  });
+
+  describe('handleConnectionError() inside a plugin worker', () => {
+    beforeEach(() => {
+      global.NX_PLUGIN_WORKER = true;
+    });
+
+    it('does not attempt reconnection', async () => {
+      const waitSpy = jest
+        .spyOn(client, 'waitForServerToBeAvailable')
+        .mockResolvedValue(true);
+      const currentReject = jest.fn();
+      client.currentReject = currentReject;
+
+      await client.handleConnectionError(new Error('daemon gone'));
+
+      expect(waitSpy).not.toHaveBeenCalled();
+      expect(currentReject).toHaveBeenCalledTimes(1);
+      const rejectedError = currentReject.mock.calls[0][0] as Error;
+      expect(rejectedError.message).toMatch(/plugin worker/i);
+    });
+
+    it('unblocks concurrent callers parked on _waitForDaemonReady', async () => {
+      // Capture the ready promise that was created in reset().
+      const pendingReady = client._waitForDaemonReady!;
+      client.currentReject = jest.fn();
+
+      await client.handleConnectionError(new Error('daemon gone'));
+
+      // The promise concurrent callers were awaiting must not be left
+      // hanging — it must surface the plugin-worker error.
+      await expect(pendingReady).rejects.toThrow(/plugin worker/i);
+      expect(client._daemonStatus).toBe(DaemonStatus.DISCONNECTED);
+    });
+  });
+
+  describe('handleConnectionError() with a version mismatch on reconnect', () => {
+    it('surfaces the error to concurrent callers awaiting _waitForDaemonReady', async () => {
+      jest
+        .spyOn(client, 'waitForServerToBeAvailable')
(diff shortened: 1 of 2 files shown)


AgentEnder (commented):
Deep review at \`b6b1c8d\` — four independent lanes (implementation, verification, approach, security), each required to prove it opened the diff. Reviewed as a draft.

The wedge is real, the fix genuinely closes it, and the tests are unusually strong — two of them fail on base by \*timing out\*, which is the most direct possible demonstration of a hang. \*\*Two Important findings\*\*, both about the plugin-worker half rather than the state-machine half.

\## Important

\*\*1. \`client.ts:1156-1172\` — the plugin-worker guard does not implement the policy its comment states.\*\* Three lanes converged on this independently.

The comment says reconnecting "can't produce a correct result" and that the branch will "Fail every queued \*\*and future\*\* message with a clear error." The code fails only the in-flight message plus whoever is parked on the current \`\_waitForDaemonReady\`, then leaves \`\_daemonStatus = DISCONNECTED\` — which is exactly the state \`startDaemonIfNecessary\` treats as "go connect."

Measured on unmodified HEAD, with \`global.NX_PLUGIN_WORKER = true\`, after the short-circuit fires:

\- \*\*Daemon reachable\*\* (routine — the parent Nx process restarts one): the worker silently reconnects. \`probeServer\` 1, \`setUpConnection\` 1, \`startInBackground\` 0, final status \`CONNECTED\`. That is precisely the outcome the comment declares impossible.

\- \*\*No daemon reachable\*\*: the next queued message dies with \`startInBackground\`'s guard text verbatim — \`Fatal Error: Something unexpected has occurred. Plugin Workers should not start a new daemon process. Please report this issue.\` Only the single in-flight message ever sees \`Plugin worker lost its daemon connection and cannot reconnect: …\`.

So neither half of the comment holds: it does reconnect, and the "clear error" reaches exactly one message. This repo treats a comment contradicting the code as a defect, and this one also mis-sells the PR's central product decision.

Two ways out, and the choice is yours: make the condition \*\*terminal\*\* (a \`\_pluginWorkerConnectionLost\` flag checked at the top of \`startDaemonIfNecessary\`, throwing \`pluginWorkerError\`), or reword the comment to what the code does — it suppresses this event's 60-second \`waitForServerToBeAvailable\` poll, nothing more. Not a regression either way: base also reconnects to a replacement daemon, so this is an incomplete policy rather than a step backwards.

\*\*2. \`client.spec.ts:540-554\` — \`unblocks concurrent callers parked on \_waitForDaemonReady\` certifies behavior the guard does not provide, because it asserts on a state production cannot reach.\*\*

It captures the promise \`reset()\` created on a brand-new client, then calls \`handleConnectionError\` directly. In production \`handleConnectionError\` is reachable only from the close callback installed by \`setUpConnection\` — and \`setUpConnection\` is immediately followed by \`\_daemonStatus = CONNECTED; this.\_daemonReady()\`, with the close handler having already set DISCONNECTED at \`client.ts:1117\`.

Measured on a realistic connect-then-die sequence: \`same promise object still installed: true\`, \`ready promise after guard: resolved\`, \`status after guard: 1\`. So \*\*both\*\* effects of \`failDaemonReady\` at \`client.ts:1170\` are no-ops in production, and the actual plugin-worker unwedging comes from the new \`startDaemonIfNecessary\` catch — which this test does not exercise. Fix: drive a successful \`startDaemonIfNecessary\` first, then park a caller, then fire the connection error — or drop the \`failDaemonReady(pluginWorkerError)\` call and this test together.

\## Strengths

\- \*\*The body's "all five fail against master" claim is TRUE, verified by executing it.\*\* With the base \`client.ts\` copied over HEAD (delta confirmed byte-for-byte to be exactly this PR's hunks) and the HEAD spec kept: \*\*5 failed, 18 passed, 2 skipped\*\*. Restored and re-confirmed 23 passed.

\- \*\*The five tests fail for four different reasons across three different hunks\*\* — not one shared cause:

  | test | base failure | hunk pinned |

  |---|---|---|

  | does not wedge concurrent callers | jest timeout 5003ms (caller B parks forever) | \`startDaemonIfNecessary\` catch |

  | lets a future caller retry | assertion: status CONNECTING, expected DISCONNECTED | same hunk, different consequence |

  | does not attempt reconnection | throws in \`getSocketPath\` — base really does take the reconnect path | plugin-worker short-circuit |

  | unblocks concurrent callers | jest timeout 4861ms | \`failDaemonReady\` in that short-circuit |

  | surfaces the error | jest timeout 5001ms | \`failDaemonReady\` at :1191 |

  Two of the five failing by \*\*timeout\*\* is the strongest available demonstration of the wedge.

\- \*\*\`createReadyPromise\`'s unhandled-rejection idiom is exactly right\*\*: \`.catch(() => {})\` is attached to the \*original\* promise and the original is returned, so a rejection nobody observes cannot crash the process while real awaiters still see it. Every assignment to \`\_waitForDaemonReady\` (:287, :1045, :1175) goes through it, so no settler can belong to an unguarded promise.

\- \*\*Both transitions into CONNECTING are now paired with a fresh ready promise, and there are no others\*\* — enumerated, not assumed. Callers in the CONNECTING branch read \`\_waitForDaemonReady\` synchronously with the status check (no \`await\` between :1079 and :1080), so there is no TOCTOU on which promise they park on. \`failDaemonReady\` runs before \`throw err\`, so a parked caller's rejection is queued as a microtask ahead of any re-entering caller.

\- \*\*The body's triggers are supported by the code\*\*, which matters because there is no linked issue: \`client.ts:1415-1419\` guards \`startInBackground\` on \`global.NX_PLUGIN_WORKER\` unconditionally and before any other work, and \`plugin-worker.ts:41\` sets that flag in production. So "always throws in a plugin worker" is accurate.

\- \*\*Security is clean and was checked rather than waved through\*\*: \`global.NX_PLUGIN_WORKER\` has exactly one writer in the repo (\`plugin-worker.ts:41\`) with no env/config/message path to it, so the gate is not attacker-influencable; and an adversarial harness confirmed queued messages \*\*reject rather than hang\*\* — no caller can mistake a non-delivered write (\`FLUSH_SYNC_GENERATOR_CHANGES_TO_DISK\`, \`RECORD_TASK_RUNS\`) for a delivered one. The change strictly replaces hangs with rejections.

\## Maintainer calls

\- \*\*Fast-fail rather than reconnect for plugin workers — endorsed, but the stated rationale is not the real one.\*\* The comment's reasoning (graph state, cached context, plugin module state make reconnection incoherent) does not hold up: the RPCs plugin workers actually issue (\`glob\`, \`multiGlob\`, \`hashGlob\`, \`getFilesInDirectory\`, \`getWorkspaceContextFileData\`) are stateless filesystem queries a replacement daemon would answer correctly, and genuine version skew is already caught by \`VersionMismatchError\`. The sound reasons are structural: \`startInBackground\` throws in a worker so nobody in-process can bring a daemon back, and the worker's host is usually the daemon that just died. What it buys is skipping a 60s poll (\`delayMs: 10, maxAttempts: 6000\`) that almost always ends in the same failure. The case it costs is a daemon \*restarting\* while a CLI-hosted worker runs — base would reconnect there, this PR fails the command. Worth accepting, and worth rewriting the comment to the real justification.

\- \*\*Whether the policy should be "in-flight only" or "the whole worker" is your call.\*\* Finding 1 is not that one is right, only that the comment must match whichever you choose.

\## Approach

Sound. Every mutation of \`\_daemonStatus\`/\`\_waitForDaemonReady\` lives in this one class (10 sites, all private), and nothing outside \`client.ts\` reads them, so the fix is contained. The alternatives were weighed:

\- \*\*Single-flight connect promise\*\* (\`this.connecting ??= this.doConnect().finally(…)\`) is structurally immune to the whole bug class — the cached promise \*is\* the operation rather than a hand-settled twin — and would delete three fields and all five \`failDaemonReady\` calls. It loses on risk, not elegance: \`handleConnectionError\` doubles as the in-flight-message resend path and \`establishConnection()\` resolves readiness from a third place, so it is a rewrite of the client's concurrency core with no pre-existing regression harness beyond the spec this PR adds. Good follow-up.

\- \*\*Explicit discriminated state object\*\* would make status/promise desync unrepresentable, which is the real residual weakness: after \`failDaemonReady\` the client is DISCONNECTED while \`\_waitForDaemonReady\` holds a \*rejected\* promise, and the "every CONNECTING transition pairs with a fresh promise" invariant is upheld by convention at three call sites rather than by construction.

\- \*\*Reusing the existing \`Deferred<T>\`\*\* was checked and correctly rejected — it lives in \`adapter/rxjs-for-await.ts\`, which imports \`rxjs\` at top level, so it would drag rxjs into the daemon client's module graph.

\## Suggestions

\- Collapse the three scattered \`failDaemonReady\` calls in \`handleConnectionError\` into one try/catch after \`createReadyPromise()\` — the PR's own better idiom, already used in \`startDaemonIfNecessary\`.

\- \`client.spec.ts:65\` mirrors \`DaemonStatus\` as a local \`const enum\`; a reorder in \`client.ts:145\` would silently make every state assertion test the wrong value. Export the enum instead.

\- \`reset()\`'s \`createReadyPromise()\` result is now unreachable (both CONNECTING transitions create a fresh promise first), so it can be \`null\` there.

\- The \`try\` opened at \`client.ts:1048\` also wraps \`\_daemonReady()\`, \`getDaemonProcessIdSync()\` and \`registerDaemonProcessWithMetricsService()\`; a throw there would flip an established CONNECTED connection back to DISCONNECTED. Latent only — end the \`try\` at \`this.\_daemonReady()\`.

\- \`client.ts:184\` \`(err: Error) => void | null\` parses as a function \*returning\* \`void | null\`, not a nullable function. It replicates the pre-existing sibling on line 183 and \`?.\` makes it runtime-safe, so it is a nit — but both lines are cheap to fix while here.

\- The body says "Added \`client.spec.ts\`"; the file pre-existed and was modified. Body wording only.

\- \`client.ts:1164\` nests a log dump inside a log dump — with a token seeded into \`daemon.log\`, the resulting message contained that line 3 times and the absolute log path 3 times (the third because \`clientLogger.log\` writes the incoming error into the same file \`daemonProcessException\` then re-reads). Same consumer, so noise rather than disclosure.

\---

\*Verdict: \*\*lgtm\*\* by the pipeline's ordering — two deduplicated Important findings, no Critical. Recorded as such rather than re-tiered by taste, but both are worth closing before this leaves draft, especially finding 1 since it misdescribes the headline product decision.\*

AgentEnder:
Superseded by #36820 — same branch, same commits, now on `nrwl/nx` instead of my fork so CI runs with a write token. The review I posted on 2026-08-13 stays readable here.

PR #35346, opened 2026-04-20
fix(testing): include config file path in plugin hash calculation

mungodewar:
## Current Behavior

The createNodesV2 function in the nx/cypress plugin calculates the hash based only on the output of the standard `calculateHashForCreateNodes`. This means that when multiple Cypress configuration files exist within a single application hash. As a result, only one set of inferred targets is generated, and the additional configuration variants are effectively ignored.

## Expected Behavior
The createNodesV2 function now takes the configuration file name into the hash calculation, removing the limitation of a single config within an application.


meeroslav (approved):

packages/cypress/src/plugins/plugin.ts
```diff
@@ -89,12 +89,10 @@ async function createNodesInternal(
...
-  const hash = await calculateHashForCreateNodes(
-    projectRoot,
-    options,
-    context,
-    [getLockFileName(detectPackageManager(context.workspaceRoot))]
-  );
+  const hash =
+    (await calculateHashForCreateNodes(projectRoot, options, context, [
+      getLockFileName(detectPackageManager(context.workspaceRoot)),
+    ])) + configFilePath;
```
AgentEnder:
```suggestion
    (await calculateHashForCreateNodes(projectRoot, options, context, [
      getLockFileName(detectPackageManager(context.workspaceRoot)),
      configFilePath
    ]));
```

mungodewar: I did have a look if this were possible, however, I saw an example of the pattern within one of the other plugins: https://github.com/nrwl/nx/blob/7f3155c9df775953651bbd65ffad6cd056d4bc4b/packages/remix/src/plugins/plugin.ts#L129 Couldn't go deeper into the native hashing to see how this was used and ended up going for this pattern. This looks a lot cleaner!

PR #35350, opened 2026-04-20
fix(webpack): set projectReferences to true for tsc compiler

alexandremaz:
## Current Behavior
ts-loader ^9.5.5 breaks webpack tsc build for an app using a workspace lib,
with typescript TS6059 error : lib index.ts is not under rootDir of app
because rootDir is no longer overriden to undefined : https://github.com/TypeStrong/ts-loader/pull/1679, thus highlighting the fact that projectReferences (ts loader option) was not set to true

https://github.com/TypeStrong/ts-loader/blob/972a4289d4ed4bbe7763980a932e16bc91766922/README.md?plain=1#L720-L728

Otherwise, source files in referenced projects will be treated as if they’re part of the root project, thus leading to 

> ERROR TS6059: File 'libs/lib-1/index.ts' is not under 'rootDir'

when building app that reference a workspace lib

## Expected Behavior
nx should be able to build app with webpack and tsc, without any errors

## Related Issue(s)

Fixes #35017


leosvelperez:
Thanks for digging into this, @alexandremaz! We went with a different fix in #36188. The default `@nx/webpack` tsc path runs `ts-loader` in `transpileOnly` mode (`transpileModule`), which `projectReferences` doesn't affect, so we widen the loader's `rootDir` to the workspace root instead. Closing in favor of that one.

PR #35077, opened 2026-03-30
feat(release): Bump patch with preid if preid is passed

mpsanchis:
## Current Behavior

Transitive bumps across release groups do not respect `preid`.

## Expected Behavior

If in a PR, where we expect snapshots/prereleases (passing `--preid=foo`), a release group is modified, its dependents should be bumped with a preid as well.

## Related Issue(s)

Fixes https://github.com/nrwl/nx/issues/35080

PR #35352, opened 2026-04-20
fix(release): respect preid for cross-releaseGroup bumps

sessa:
## Summary

- When `--preid` is passed to `nx release`, transitive version bumps that previously always used `patch` now use `prepatch` so dependent release groups stay on the prerelease line (e.g. `1.0.1-alpha.0` instead of `1.0.1`).
- Covers `DEPENDENCY_WAS_BUMPED`, `OTHER_PROJECT_IN_FIXED_GROUP_WAS_BUMPED_DUE_TO_DEPENDENCY`, and `determineSideEffectBump` (`DEPENDENCY_ACROSS_GROUPS_WAS_BUMPED`).

## Context

Fixes https://github.com/nrwl/nx/issues/35080 (repro: only the downstream release group gets a prerelease bump today; dependents get a plain patch).

This matches the approach in https://github.com/nrwl/nx/pull/35077 from @mpsanchis; this PR was opened from a separate fork for visibility while that PR is pending review.

## Test plan

- [ ] `nx test nx` (or targeted `release-group-processor` specs) — not run here: local `nx` project graph fails in this environment because `@nx/gradle` cannot download Gradle (SSL PKIX).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/nx/src/command-line/release/version/release-group-processor.spec.ts
+++ b/packages/nx/src/command-line/release/version/release-group-processor.spec.ts
@@ -1424,6 +1424,73 @@ describe('ReleaseGroupProcessor', () => {
       `);
     });
 
+    it('should use prepatch instead of patch when preid is set and a dependent project is bumped across release groups', async () => {
+      const { nxReleaseConfig, projectGraph, filters } =
+        await createNxReleaseConfigAndPopulateWorkspace(
+          tree,
+          `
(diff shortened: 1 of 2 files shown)


sessa:
Closing as duplicate of https://github.com/nrwl/nx/pull/35077 which addresses the same fix for https://github.com/nrwl/nx/issues/35080.

PR #35353, opened 2026-04-21
docs(repo): add conformance rule for NX_* env var documentation

barbados-clemens:
## Current Behavior

`NX_*` env vars can be added to Nx source without being documented in `astro-docs/src/content/docs/reference/environment-variables.mdoc`. Nothing catches the drift.

## Expected Behavior

Adds a conformance rule (`env-vars-documented`) that fails when an `NX_*` var is read in source but missing from the docs. Covers TS/JS `process.env.NX_*` and Rust `env::var` / `env!` patterns, skipping tests and fixtures. An `ignore` list in `nx.json` handles internal markers not meant to be documented.

Also documents the user-facing vars the first run surfaced (self-hosted cache, provenance, plugin isolation, Nx Cloud timeouts, etc.) and marks `NX_CLOUD_AUTH_TOKEN` and `NX_CLOUD_DISTRIBUTED_EXECUTION_AGENT_COUNT` as deprecated.

## Related Issue(s)

N/A — internal tooling improvement.


terminalOutput
AgentEnder, line 1: @barbados-clemens this shouldn't be in the PR

barbados-clemens: wtf? where did this come from 🤔

nx.json
```diff
@@ -376,6 +376,46 @@
...
         "rule": "./dist/workspace-plugin/src/conformance-rules/types-versions-exports-sync"
+      },
+      {
+        "rule": "./dist/workspace-plugin/src/conformance-rules/env-vars-documented",
+        "options": {
+          "ignore": [
+            "NX_ANALYTICS_SESSION_ID",
+            "NX_FORKED_TASK_EXECUTOR",
+            "NX_RUN_COMMANDS_DIRECTLY",
+            "NX_NATIVE_COMMAND_RUNNER",
```
AgentEnder: This toggles behavior, you could argue it should be documented

nx.json
```diff
@@ -376,6 +376,46 @@
...
+            "NX_DAEMON_PROCESS",
+            "NX_IPC_CHANNEL_ID",
+            "NX_PSEUDO_TERMINAL_EXEC_ARGV",
+            "NX_GENERATE_DOCS_PROCESS",
+            "NX_RELEASE_INTERNAL_SUPPRESS_FILTER_LOG",
+            "NX_WORKSPACE_ROOT_PATH",
+            "NX_IMPORT_SOURCE",
+            "NX_IMPORT_DESTINATION",
+            "NX_PROJECT_GRAPH_CACHE_DIRECTORY",
+            "NX_DAEMON_SOCKET_DIR",
```
AgentEnder: This one changes where the daemon socket is created, some users have had to do so if their tmpdir path was too long, maybe should be documented

nx.json
```diff
@@ -376,6 +376,46 @@
...
+            "NX_CONSOLE",
+            "NX_WRAPPER_SKIP_INSTALL",
```
AgentEnder: I'd argue we should doc this one, skips the `npm i` during the wrapper bootup

nx.json
```diff
@@ -376,6 +376,46 @@
...
+            "NX_DEBUG_TELEMETRY",
+            "NX_TUI_INLINE_MODE",
+            "NX_TUI_SKIP_CAPABILITY_CHECK",
+            "NX_SKIP_NATIVE_FILE_CACHE",
```
AgentEnder: I'm not sold either way but you could argue this is an escape hatch that is worth mentioning in the table...

I almost wish we had either more sections or a way to toggle common vs advanced settings

nx.json
```diff
@@ -376,6 +376,46 @@
...
+            "NX_SKIP_CHECK_REMOTE",
+            "NX_CLI_SET",
+            "NX_WORKSPACE_ROOT",
+            "NX_TERMINAL_OUTPUT_PATH",
+            "NX_TERMINAL_CAPTURE_STDERR",
+            "NX_VERSION",
+            "NX_STREAM_OUTPUT",
+            "NX_PREFIX_OUTPUT",
+            "NX_CACHE_FAILURES",
```
AgentEnder: This one is questionable too, but maybe should be documented

tools/workspace-plugin/src/conformance-rules/env-vars-documented/index.ts
```diff
@@ -0,0 +1,119 @@
...
+const NX_CORE_PROJECT = 'nx';
+const MAX_EXAMPLE_FILES = 3;
+
+const SCANNABLE_EXT = /\.(ts|tsx|js|mjs|cjs|rs)$/;
+const TEST_FILE = /\.(spec|test)\.(ts|tsx|js)$/;
+const EXCLUDED_SEGMENT = /(__fixtures__|__snapshots__|\/files\/|\/dist\/)/;
+
+const TS_JS_USAGE = /process\.env(?:\.|\[['"`])(NX_[A-Z0-9_]+)/g;
+const RUST_ENV_VAR = /(?:std::)?env::var(?:_os)?\s*\(\s*"(NX_[A-Z0-9_]+)"/g;
+const RUST_ENV_MACRO = /env!\s*\(\s*"(NX_[A-Z0-9_]+)"/g;
```
AgentEnder: The rust env macro is compile time, not execution time, so I don't think it should ever point at something that should be documented

AgentEnder (changes requested):
The rust bit is the only real change I'm requesting, the others are "we may want to doc this one" comments to open a discussion

AgentEnder (approved):

PR #35365, opened 2026-04-21
fix(react): support Vite 8 for React Router apps

jaysoo:
## Current Behavior

`@react-router/dev` peer dep tops out at Vite 7, so the React app generator forces Vite 7 when `--use-react-router` is passed, and the react-router typecheck e2e test is skipped in CI.

## Expected Behavior

`@react-router/dev` 7.14.2 expands its peer dep to include Vite 8 (`^5.1.0 || ^6.0.0 || ^7.0.0 || ^8.0.0`), so:

- `useViteV7: true` is no longer forced when generating a React Router app
- The `useViteV7` field is removed from `ViteConfigurationGeneratorSchema` (was only added for this workaround and was never wired into the configuration generator body)
- The react-router typecheck e2e test is un-skipped
- `reactRouterVersion` is bumped to `^7.14.2`
- A new `22.7.0` packageJsonUpdate migrates `@react-router/*` packages to `7.14.2` in existing workspaces

## Related Issue(s)

Fixes NXC-4182

### DIFF
--- a/packages/react/migrations.json
+++ b/packages/react/migrations.json
@@ -195,6 +195,47 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "22.7.0": {
+      "version": "22.7.0-beta.18",
+      "packages": {
+        "react-router": {
+          "version": "7.14.2",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-router/dev": {
+          "version": "7.14.2",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-router/node": {
+          "version": "7.14.2",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-router/serve": {
+          "version": "7.14.2",
+          "alwaysAddToPackageJson": false
+        },
+        "@react-router/express": {
+          "version": "7.14.2",
+          "alwaysAddToPackageJson": false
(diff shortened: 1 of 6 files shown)


packages/react/src/generators/application/lib/bundlers/add-vite.ts
```diff
@@ -43,8 +43,6 @@ export async function setupViteConfiguration(
     addPlugin: options.addPlugin,
     projectType: 'application',
     port: options.port,
-    // React Router does not yet support Vite 8, so force Vite 7.
-    ...(options.useReactRouter ? { useViteV7: true } : {}),
```
FrozenPandaz, line 47: Should we keep this based on what version of `react-router` is used?

FrozenPandaz (approved):

packages/react/migrations.json
```diff
@@ -195,6 +195,47 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "22.7.0": {
+      "version": "22.7.0-beta.16",
```
FrozenPandaz: Also while you're here, you should update this to 22.7.0-beta.18.

FrozenPandaz (approved):

PR #35367, opened 2026-04-21
feat(core): show target uses task graph + filter broken dependsOn during normalization

AgentEnder:
## Current Behavior

- During project graph normalization, dependsOn entries whose target doesn't exist are left in place, which means downstream consumers have to re-validate against the real target set every time.
- `nx show target` resolves the `Depends On` list using a heuristic on raw `dependsOn` configs. That heuristic doesn't fully match what `createTaskGraph` would schedule — e.g. a `{ projects: ['lib-a'] }` entry is listed even if `lib-a` doesn't actually have the referenced target, and there's no indication of what transitively runs when the target is scheduled.

### 1. Normalize-project-nodes drops broken same-project dependsOn entries

A new `normalizeTargetsDependsOn` pass runs inside `normalizeProjectNodes` after target defaults have been merged. It filters:

- Bare-string entries (e.g. `"prebuild"`) when the referenced target doesn't exist on the same project.
- Object entries with no `projects`, `projects: 'self'`, or `projects: '{self}'` when the referenced target doesn't exist on the same project.

Cross-project entries are intentionally left alone — `"^target"`, `{ target: 'X', dependencies: true }`, and `{ projects: [...] }` need the real project graph (and in some cases the dep edges) to validate, so they're resolved later during task graph creation.

### 2. `nx show target` is task-graph-driven

`showTargetInfoHandler` now builds a task graph rooted at the requested target via `createTaskGraph`. The `Depends On` list reads direct task dependencies off the graph, giving the same filtering and resolution behavior as `nx run`:

- Non-existent targets disappear from the list.
- `^target` expands to real dependency-project task IDs.
- `{ projects: [...] }` entries drop projects that don't actually have the target.

A new `transientTasks` field on the JSON output surfaces everything that runs transitively. The text renderer shows a summary line beneath `Depends On`:

```
Depends On:
  lib-a:build
  and 5 build, compile transient tasks
```

With >3 unique target names, the summary collapses to a bare count (`and 12 transient tasks`) to stay scannable. Source-map hints for `--verbose` are preserved — each direct dep task ID is matched back to its originating `depConfig` entry.

## Related Issue(s)

None — proactive correctness + observability improvement.

### DIFF
--- a/packages/nx/src/command-line/show/show-target/info.ts
+++ b/packages/nx/src/command-line/show/show-target/info.ts
@@ -3,7 +3,11 @@ import type { ProjectGraph } from '../../../config/project-graph';
 import type { InputDefinition } from '../../../config/workspace-json-project-json';
 import type { ConfigurationSourceMaps } from '../../../project-graph/utils/project-configuration/source-maps';
 import { getNamedInputs } from '../../../hasher/task-hasher';
-import { getDependencyConfigs } from '../../../tasks-runner/utils';
+import { createTaskGraph } from '../../../tasks-runner/create-task-graph';
+import {
+  createTaskId,
+  getDependencyConfigs,
+} from '../../../tasks-runner/utils';
 import type { ShowTargetBaseOptions } from '../command-object';
 import {
   resolveTarget,
@@ -57,12 +61,13 @@ function resolveTargetInfoData(t: ResolvedTarget) {
       .map(([name, config]) => [name, config.dependsOn])
   );
 
-  const depConfigs = getDependencyConfigs(
-    { project: projectName, target: targetName },
-    extraTargetDeps,
-    graph,
-    [...allTargetNames]
-  );
+  const depConfigs =
+    getDependencyConfigs(
+      { project: projectName, target: targetName },
+      extraTargetDeps,
+      graph,
+      [...allTargetNames]
+    ) ?? [];
 
   // Determine the hoisted command value and which option key it came from
   let command: string | undefined;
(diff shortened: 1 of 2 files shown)


packages/nx/src/project-graph/utils/normalize-project-nodes.ts
```diff
@@ -125,6 +131,53 @@ export function normalizeImplicitDependencies(
...
+  if (!dep || !dep.target) {
+    return true;
+  }
+
+  // Anything that references other projects (via `projects` or
+  // `dependencies: true`) is validated later during task graph creation.
+  const selfRef =
+    (dep.projects === undefined && !dep.dependencies) ||
+    dep.projects === 'self' ||
+    dep.projects === '{self}';
```
FrozenPandaz: Mm this is very old.. I don't think we still recognize this anywhere else in the configuration?

packages/nx/src/project-graph/utils/normalize-project-nodes.ts
```diff
@@ -125,6 +131,53 @@ export function normalizeImplicitDependencies(
...
+}
+
+function isValidDependsOnEntry(
+  dep: string | TargetDependencyConfig,
+  ownTargets: Record<string, TargetConfiguration>
+): boolean {
+  if (typeof dep === 'string') {
+    // "^target" references dependency projects, which aren't known yet at
+    // this stage - leave it for task graph creation.
+    if (dep.startsWith('^')) {
```
FrozenPandaz: Mm so we can't actually filter this all the way down here...

packages/nx/src/command-line/show/show-target/info.ts
```diff
@@ -166,6 +167,114 @@ function resolveDependencyProjects(
...
+    taskGraph = createTaskGraph(
+      graph,
+      extraTargetDeps,
+      [projectName],
+      [targetName],
+      configuration,
+      {}
+    );
+  } catch {
+    return { dependsOn: [], depSourceIndices: [], transientTasks: [] };
```
FrozenPandaz: Hm so if we throw an error during task graph creation.. such as... circular deps... we should have like.. the full list probably.. not an empty list?

packages/nx/src/command-line/show/show-target/info.ts
```diff
@@ -166,6 +167,114 @@ function resolveDependencyProjects(
...
+  }
+
+  const rootId = Object.keys(taskGraph.tasks).find((id) => {
+    const t = taskGraph.tasks[id].target;
+    return t.project === projectName && t.target === targetName;
+  });
```
FrozenPandaz: This seems complex.. I think we can do something like `taskGraph.dependencies[createTaskId(target)]` right?

packages/nx/src/command-line/show/show-target/info.ts
```diff
@@ -166,6 +167,114 @@ function resolveDependencyProjects(
...
+    if (!task) continue;
+    dependsOn.push(depTaskId);
+    depSourceIndices.push(
+      findDepConfigIndex(task.target, depConfigs, projectName, graph)
+    );
+  }
+
+  const transientTasks = Object.keys(taskGraph.tasks).filter(
+    (id) => id !== rootId && !directDepSet.has(id)
+  );
```
FrozenPandaz: Transient? Is that the right word? It should be transitive right?

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #35382, opened 2026-04-22
docs(nx-cloud): document `no-output-timeout` launch template syntax

barbados-clemens:
allow setting a timeout for nx agents based on the last time output was
emitted.

By default this is 10m. this is set per launch template but can be
overridden per step via env var: `NX_NO_OUTPUT_TIMEOUT`

fixes DOC-488


jaysoo (approved):

astro-docs/src/content/docs/reference/environment-variables.mdoc
```diff
@@ -87,21 +87,22 @@ The following environment variables are specific to Nx Cloud and can be used to
...
+| `NX_CLOUD_API`                        | string  | The URL of the Nx Cloud instance to connect to. Overrides `nxCloudUrl` in `nx.json`.                                                                                                                                     |
+| `NX_CLOUD_ENCRYPTION_KEY`             | string  | Enable end-to-end encryption of artifacts. Artifacts will be encrypted/decrypted on your machine. Can also be set via the `encryptionKey` property in `nx.json`.                                                         |
+| `NX_CLOUD_NO_TIMEOUTS`                | boolean | Disables the default 10-second timeout for Nx Cloud requests.                                                                                                                                                            |
+| `NX_NO_OUTPUT_TIMEOUT`                | string  | Overrides the launch template's [`no-output-timeout`](/docs/reference/nx-cloud/launch-templates#launch-templatestemplate-nameno-output-timeout) for an Nx Agent step. Duration string (e.g., `10m`, `1h`); `0` disables. |
```
barbados-clemens: this was the only env var added, but format caused whole table to get diffed

PR #35387, opened 2026-04-22
fix(core): fall back to workspace project lookup when require resolves a hoisted copy

FrozenPandaz:
## Current Behavior

`TargetProjectLocator.findProjectFromImport` returns `undefined` whenever `resolveImportWithRequire` succeeds but the resolved path lives under `node_modules/` — it never reaches the `findImportInWorkspaceProjects` fallback.

In pnpm workspaces, this drops edges between workspace projects whenever the consumer imports a workspace package it doesn't declare as a dependency. `require.resolve` walks up through `node_modules` and lands on the hoisted external copy in `.pnpm/` (e.g. `node_modules/.pnpm/@nx+jest@.../node_modules/@nx/jest/index.js`) instead of the workspace symlink. `findProjectOfResolvedModule` sees `node_modules/` in the path, returns `undefined`, and the edge is silently lost.

Concrete example in this repo: `packages/eslint/src/generators/workspace-rules-project/workspace-rules-project.ts` calls `ensurePackage<typeof import('@nx/jest')>('@nx/jest', nxVersion)`. The Rust import locator correctly captures `@nx/jest` as a static import, but the edge `@nx/eslint -> @nx/jest` never materializes in the project graph. #35377 worked around this by declaring `@nx/jest` as an optional peer dependency so the edge would be picked up via `explicit-package-json-dependencies` instead.

## Expected Behavior

When `findProjectOfResolvedModule` can't map the resolved path to a workspace project, fall through to `findImportInWorkspaceProjects` — the workspace-package-metadata lookup that already handles "locally linked workspace project where the output might not exist yet." It correctly maps the import expression (e.g. `@nx/jest`) to the corresponding workspace project via `entryPointsToProjectMap`.

After this fix, on this repo:
- `eslint -> jest (static)` materializes without needing the peer-dep workaround
- `@nx/nx-source -> eslint-rules (static)` is recovered (root `eslint.config.mjs` imports from `./tools/eslint-rules`)
- No edges removed

Added a regression test that stubs `resolveImportWithRequire` to return a `node_modules/.pnpm/...` path and asserts the workspace fallback resolves to the correct project. Verified it fails against the pre-fix code.

### DIFF
--- a/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.spec.ts
+++ b/packages/nx/src/plugins/js/project-graph/build-dependencies/target-project-locator.spec.ts
@@ -1232,6 +1232,56 @@ describe('TargetProjectLocator', () => {
     });
   });
 
+  describe('findProjectFromImport workspace fallback', () => {
+    it('should fall back to findImportInWorkspaceProjects when require resolves into node_modules', () => {
+      // Scenario: workspace package `@org/pkg1` is imported from a file in
+      // another workspace project that does NOT declare `@org/pkg1` as a
+      // dependency. In pnpm layouts, Node's resolver can walk up and find a
+      // hoisted copy of the package in `node_modules/.pnpm/...`, causing
+      // `resolveImportWithRequire` to return a path under `node_modules/`.
+      // The edge to `pkg1` would previously be dropped; it should now be
+      // recovered via the workspace-metadata fallback.
+      const projects: Record<string, ProjectGraphProjectNode> = {
+        pkg1: {
+          name: 'pkg1',
+          type: 'lib',
+          data: {
+            root: 'pkg1',
+            metadata: {
+              js: {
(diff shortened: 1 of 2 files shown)


FrozenPandaz:
Closing — investigated further and found this fix doesn't actually solve the motivating case.

`findNpmProjectFromImport` runs before the workspace fallback and short-circuits when the imported package exists as an external node (which is the case for every published `@nx/*` package in this workspace). My fallback is never reached for those imports.

The `eslint → jest` edge I thought my fix produced was actually from PR #35377's peerDep declaration, via `explicit-package-json-dependencies`. The storybook → web case confirmed it: trace shows `findNpmProjectFromImport → npm:@nx/web` returning first.

A real fix for the `ensurePackage<typeof import(...)>` case would need to prefer workspace projects over external nodes when both exist — a bigger behavior change worth more design discussion.

PR #35388, opened 2026-04-22
fix(core): keep continuous children alive when nx:noop orchestrator completes

AgentEnder:
## Current Behavior

A target with `executor: nx:noop` and `dependsOn` pointing at continuous tasks is unusable as a top-level orchestrator.

Given a config like:

```jsonc
"dev": {
  "executor": "nx:noop",
  "dependsOn": ["serve", "watch-css"]
}
```

where `serve` and `watch-css` are continuous, running `nx run app:dev`:

1. `serve` and `watch-css` start and become ready.
2. The `nx:noop` fast-path (`task-orchestrator.ts:1128`) exits `dev` instantly with success.
3. `cleanUpUnneededContinuousTasks` (`task-orchestrator.ts:1785-1809`) runs on every completion. Its predicate keeps a continuous task alive only if it's in `initializingTaskIds` *and still incomplete*, OR it's a `continuousDependencies` entry of some incomplete task.
4. The only initiating task (`dev`) just completed, and the continuous children aren't directly in `initializingTaskIds` — they're only deps of the now-complete noop.
5. Both continuous children get `SIGKILL`'d immediately.

## Expected Behavior

Running a top-level `nx:noop` orchestrator should keep its continuous `dependsOn` children alive. The orchestrator process stays running until the continuous children exit (e.g. user Ctrl+C), matching the ergonomics of `nx run-many -t serve,watch-css`.

## Fix

Expand `initializingTaskIds` inside the orchestrator through `nx:noop` tasks when seeding the keep-alive anchor set. The real tasks beneath a noop become the anchors so they survive the noop's instant completion.

- New helper `expandInitiatingTasksThroughNoop` in `packages/nx/src/tasks-runner/utils.ts`: BFS that drops any `nx:noop` from the anchor set and enqueues its `dependencies` + `continuousDependencies`, recursing through nested noops; cycle-safe via a `visited` set.
- Wire it into `TaskOrchestrator.initializingTaskIds` at `task-orchestrator.ts:101`.
- The public `initiatingTasks` array is left untouched, so TUI output (`tui-summary-life-cycle.ts`), atomized `e2e-ci` coordinator semantics, hashing, and caching are unchanged. For atomized chunk patterns the expansion has no visible effect because children aren't continuous.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/tasks-runner/utils.ts
+++ b/packages/nx/src/tasks-runner/utils.ts
@@ -436,6 +436,48 @@ export function getExecutorNameForTask(task: Task, projectGraph: ProjectGraph) {
   return getTargetConfigurationForTask(task, projectGraph)?.executor;
 }
 
+/**
+ * Expand a set of initiating task IDs by walking through any `nx:noop` tasks
+ * and replacing them with their direct dependencies + continuous dependencies.
+ * Non-noop tasks are kept as-is; cycles are safe.
+ *
+ * An `nx:noop` executor returns immediately, so if it is the only thing
+ * anchoring a continuous child, the child gets killed by
+ * `cleanUpUnneededContinuousTasks` the moment the noop completes. Treating the
+ * noop's dependencies as the real anchors preserves the intended orchestration.
+ */
+export function expandInitiatingTasksThroughNoop(
+  initiatingTasks: Task[],
+  taskGraph: TaskGraph,
+  projectGraph: ProjectGraph
+): Set<string> {
+  const expanded = new Set<string>();
+  const visited = new Set<string>();
+  const queue: string[] = initiatingTasks.map((t) => t.id);
+
+  while (queue.length > 0) {
+    const taskId = queue.shift()!;
+    if (visited.has(taskId)) continue;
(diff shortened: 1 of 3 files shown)


packages/nx/src/tasks-runner/utils.ts
```diff
@@ -436,6 +436,48 @@ export function getExecutorNameForTask(task: Task, projectGraph: ProjectGraph) {
...
+ * anchoring a continuous child, the child gets killed by
+ * `cleanUpUnneededContinuousTasks` the moment the noop completes. Treating the
+ * noop's dependencies as the real anchors preserves the intended orchestration.
+ */
+export function expandInitiatingTasksThroughNoop(
+  initiatingTasks: Task[],
+  taskGraph: TaskGraph,
+  projectGraph: ProjectGraph
+): Set<string> {
+  const expanded = new Set<string>();
```
FrozenPandaz, line 454: I believe we have a util for this somewhere.

FrozenPandaz (approved):

PR #35389, opened 2026-04-22
chore(misc): richer telemetry for init and connect commands

jaysoo:
## Current Behavior

`nx init` error telemetry is largely opaque: ~22% of starts land in a bare `Command failed: npm install` bucket, and ~5% record an empty `errorMessage`. We can't tell what's actually going wrong. `nx connect` has no start/error events at all — failures (missing remote, auth, network) go untracked.

## Expected Behavior

Telemetry-only change. Child-process calls in init pipe stderr so the captured output reaches the error payload; error events now include `errorName` (from Node `e.code` or an extracted `E…`/`ERR_…` token like `E404`, `ERESOLVE`, `EINTEGRITY`, `ERR_PNPM_*`) and the same env context (`nodeVersion`, `os`, `packageManager`, `isCI`, `aiAgent`) as start events. `toErrorString` fixes the empty-message bucket. `nx connect` gains proper start/complete/error events.

No behavioral fixes — once the enriched data comes in we'll prioritize real fixes by actual failure distribution.

## Related Issue(s)

Fixes NXC-4262

### DIFF
--- a/packages/nx/src/command-line/nx-cloud/connect/command-object.ts
+++ b/packages/nx/src/command-line/nx-cloud/connect/command-object.ts
@@ -1,7 +1,6 @@
 import { Argv, CommandModule } from 'yargs';
 import { handleImport } from '../../../utils/handle-import';
 import { linkToNxDevAndExamples } from '../../yargs-utils/documentation';
-import { nxVersion } from '../../../utils/versions';
 import { withVerbose } from '../../yargs-utils/shared-options';
 
 export const yargsConnectCommand: CommandModule = {
@@ -15,13 +14,6 @@ export const yargsConnectCommand: CommandModule = {
     await (
       await handleImport('./connect-to-nx-cloud.js', __dirname)
     ).connectToNxCloudCommand({ ...args, checkRemote });
-    await (
-      await handleImport('../../../utils/ab-testing.js', __dirname)
-    ).recordStat({
-      command: 'connect',
-      nxVersion,
-      useCloud: true,
-    });
     process.exit(0);
   },
 };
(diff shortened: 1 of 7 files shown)


packages/nx/src/command-line/nx-cloud/connect/command-object.ts
```diff
@@ -11,18 +20,54 @@ export const yargsConnectCommand: CommandModule = {
...
       useCloud: true,
+      meta: { type: 'start', ...baseMeta },
     });
-    process.exit(0);
+
+    try {
+      const checkRemote = process.env.NX_SKIP_CHECK_REMOTE !== 'true';
+      await (
+        await handleImport('./connect-to-nx-cloud.js', __dirname)
+      ).connectToNxCloudCommand({ ...args, checkRemote });
```
FrozenPandaz: Can you move this logic into `connectToNxCloudCommand` itself and keep this file as is?

packages/nx/src/command-line/nx-cloud/connect/command-object.ts
```diff
@@ -3,6 +3,15 @@ import { handleImport } from '../../../utils/handle-import';
 import { linkToNxDevAndExamples } from '../../yargs-utils/documentation';
 import { nxVersion } from '../../../utils/versions';
 import { withVerbose } from '../../yargs-utils/shared-options';
+import { recordStat } from '../../../utils/ab-testing';
```
FrozenPandaz: we don't want to add these imports here because then it'll slow down every comand even if it's not connect

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #35392, opened 2026-04-22
chore(repo): declare lazy-loaded packages as optional peers

AgentEnder:
## Current Behavior

Many Nx plugin packages lazy-load other plugins at runtime via `ensurePackage()` or the `require('@nx' + '/...')` pattern. These dependencies weren't declared in `package.json` at all, which meant:

- `pnpm install` in the monorepo happened to find them only via hoisting from unrelated `devDependencies`.
- Published packages gave no install-time signal about what optional peers a consumer might want.
- The dependencies were invisible to dependency-graph tooling, supply-chain audits, and any future semantic-versioning logic.

## Expected Behavior

Every lazy-loaded plugin or tool is declared as `peerDependencies` + `peerDependenciesMeta: { X: { optional: true } }`, matching the existing convention in `@nx/angular`, `@nx/angular-rspack`, `@nx/eslint`, etc. This gives consumers correct install/publish semantics without requiring them to install peers they don't use.

For two packages — `@nx/workspace` and `@nx/js` — several of their newly-declared peers transitively reverse-depend on them. Raw package.json edges would cause `@nx/js:typescript-sync` to produce circular TypeScript project references (TS6202). Those two packages use `implicitDependencies: ["!name", …]` in `project.json` to drop the cyclic graph edges, keeping the task graph and tsc builds cycle-free without modifying the sync generator itself.

Commits:
1. **workspace + js**: 14 optional peers on `@nx/workspace`, 5 on `@nx/js`, plus `implicitDependencies` negations in each `project.json`.
2. **Plugin packages**: `@nx/angular`, `@nx/expo`, `@nx/next`, `@nx/nuxt`, `@nx/react-native`, `@nx/storybook`, `@nx/vite`, `@nx/vue`, `@nx/web`. Cycles don't form for any of these, so no `implicitDependencies` negations were needed. `@nx/js:typescript-sync` populated the corresponding `tsconfig.lib.json` project references, which are committed alongside the `package.json` changes.

## Related Issue(s)

Fixes #

## Test plan

see: https://staging.nx.app/runs/m3Otv2Xl7m?sandboxViolations=true&query=%3Atest


AgentEnder:
NOTE: Blocked by #35401, currently no way to skip the sync generator adding the refs

packages/rspack/project.json
```diff
@@ -5,9 +5,34 @@
   "projectType": "library",
   "targets": {
     "build": {
+      "dependsOn": ["build-base"],
       "outputs": ["{workspaceRoot}/dist/packages/rspack/README.md"],
       "command": "node ./scripts/copy-readme.js rspack",
       "inputs": ["copyReadme"]
+    },
+    "build-base": {
```
FrozenPandaz: this stuff changed even though no implicits were added?

FrozenPandaz (approved):

PR #35398, opened 2026-04-23
fix(dotnet): correct output paths for Web SDK and centralized dist setups

AgentEnder:
## Current Behavior

The `@nx/dotnet` analyzer emits incorrect `outputs` for two project layouts:

1. **`Microsoft.NET.Sdk.Web` projects** (`webapi` / `webapp` templates): the `obj` output is emitted as `{projectRoot}/<ProjectName>/obj` — the project directory is double-counted.

   ```json
   "outputs": [
     "{projectRoot}/bin",
     "{projectRoot}/MyWebApp/obj"
   ]
   ```

   The Web SDK resolves `BaseIntermediateOutputPath` through `[MSBuild]::NormalizePath`, producing an absolute path rooted at the project directory. The analyzer made that path workspace-relative, then prefixed it with `{projectRoot}` — effectively counting the project directory twice.

2. **Centralized `dist/` via `Directory.Build.props`** (a pattern used by nx-dotnet and similar repos): paths that escape the project directory are still prefixed with `{projectRoot}`:

   ```json
   "outputs": [
     "{projectRoot}/dist/apps/foo",
     "{projectRoot}/dist/intermediates/apps/foo/obj"
   ]
   ```

## Expected Behavior

Each output is prefixed according to where it actually lives:

- Paths under the project directory → `{projectRoot}/…`
- Paths under the workspace root but outside the project → `{workspaceRoot}/…`
- Artifacts-output layout (SDK-managed `artifacts/`) → `{workspaceRoot}/artifacts/…`

`TargetBuilder` path helpers now return fully-qualified, Nx-prefixed strings; callers no longer compute a per-target prefix.

A new xUnit test project at `packages/dotnet/analyzer.Tests` pins the behaviour with seven tests covering both reported scenarios plus the artifacts layout. It's wired as an Nx project (`MsbuildAnalyzer.Tests`) so its `test` target piggybacks on `targetDefaults.test.dependsOn: ["test-native", ...]` — meaning `nx affected -t test` on CI runs the suite without any workflow change.

This PR also registers `@nx/dotnet` in `nx.json` (scoped to the analyzer projects) so the workspace uses the plugin it ships.

## Related Issue(s)

Fixes #33971

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Restore .NET analyzer projects
-        run: dotnet restore packages/dotnet/analyzer.Tests/MsbuildAnalyzer.Tests.csproj
+        run: dotnet restore nx.sln
 
       - name: Nx Report
         run:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,9 @@ jobs:
       - name: Install project dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Restore .NET analyzer projects
+        run: dotnet restore packages/dotnet/analyzer.Tests/MsbuildAnalyzer.Tests.csproj
+
       - name: Nx Report
         run:
           pnpm nx report
--- a/nx.json
+++ b/nx.json
@@ -251,6 +251,16 @@
       "exclude": ["packages/react-native/src/generators/application/files/**/*"]
     },
     {
+      "plugin": "@nx/dotnet",
+      "include": [
+        "packages/dotnet/analyzer/**",
+        "packages/dotnet/analyzer.Tests/**"
+      ],
+      "options": {
+        "test": { "targetName": "test-native" }
+      }
+    },
+    {
       "plugin": "@nx/storybook/plugin",
       "options": {
         "serveStorybookTargetName": "serve:storybook",
--- a/nx.json
+++ b/nx.json
@@ -258,7 +258,7 @@
       ],
       "options": {
         "build": { "targetName": "build:dotnet" },
-        "test": { "targetName": "test-native" }
+        "test": { "targetName": "test:dotnet" }
       }
     },
     {
--- a/nx.json
+++ b/nx.json
@@ -257,6 +257,7 @@
         "packages/dotnet/analyzer.Tests/**"
       ],
       "options": {
+        "build": { "targetName": "build:dotnet" },
         "test": { "targetName": "test-native" }
       }
     },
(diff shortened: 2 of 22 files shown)


nx.json
```diff
@@ -250,6 +250,17 @@
       },
       "exclude": ["packages/react-native/src/generators/application/files/**/*"]
     },
+    {
+      "plugin": "@nx/dotnet",
+      "include": [
```
FrozenPandaz: rm

.github/workflows/ci.yml
```diff
@@ -81,6 +81,9 @@ jobs:
       - name: Install project dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Restore .NET analyzer projects
+        run: dotnet restore packages/dotnet/analyzer.Tests/MsbuildAnalyzer.Tests.csproj
```
FrozenPandaz: Let's go with a solution file at the root

packages/dotnet/analyzer/Utilities/TargetBuilder.PathHelpers.cs
```diff
@@ -17,39 +24,95 @@ private static bool UsesArtifactsOutput(Dictionary<string, string> properties)
...
+        {
+            var relative = Path.GetRelativePath(normalizedWorkspace, normalizedPath)
+                .Replace('\\', '/').TrimEnd('/');
+            return relative == "." || string.IsNullOrEmpty(relative)
+                ? "{workspaceRoot}"
+                : $"{{workspaceRoot}}/{relative}";
+        }
+
+        // Absolute path outside the workspace - return as-is for Nx to treat as absolute.
+        return normalizedPath.Replace('\\', '/').TrimEnd('/');
```
FrozenPandaz: This isn't really allowed as an output so returning this isn't correct.

packages/dotnet/src/plugins/create-nodes.spec.ts
```diff
@@ -143,7 +143,7 @@ describe('@nx/dotnet - createNodes', () => {
 
       const result = mergeUserTargetConfigurations(node, options);
 
-      expect(result.targets?.build).toEqual({
+      expect(result.targets?.['build']).toEqual({
```
FrozenPandaz: Why did this change?

packages/nx/src/native/db/initialize.rs
```diff
@@ -10,7 +10,7 @@ use std::path::{Path, PathBuf};
 use tracing::{debug, trace};
 
 /// Bump this ONLY when the database schema changes.
-pub const DB_VERSION: &str = "2";
+pub const DB_VERSION: &str = "3";
```
FrozenPandaz: rm

FrozenPandaz (approved):

PR #35405, opened 2026-04-23
fix(release): skip expensive changelog operations when changelogs are disabled

### LINKED ISSUE #35366: release: releaseChangelog() performs expensive operations even when changelogs are disabled
## Current Behavior

When running `nx release` with changelogs effectively disabled (both `workspaceChangelog: false` and per-group `changelog: { projectChangelogs: false, file: false }`), the `releaseChangelog()` step still performs expensive operations including:

1. A full `createProjectGraphAsync()` call (line 46 of `changelog.js`)
2. Git log operations to gather commit data for all projects
3. Full changelog processing for all release groups

This causes a noticeable delay (30+ seconds in larger monorepos) between the version step completing ("Image tagged with...") and the git tagging step ("Tagging commit with git").

## Expected Behavior

When changelogs are disabled, `releaseChangelog()` should skip all expensive operations, or ideally not be called at all by the top-level `release()` function.

## Root Cause

There are two issues:

### 1. Truthy object passes the "enabled" check

In `changelog.js` (line 107-108), the early return check is:

```js
const changelogGenerationEnabled = !!nxReleaseConfig.changelog.workspaceChangelog ||
    Object.values(nxReleaseConfig.groups).some((g) => g.changelog);
if (!changelogGenerationEnabled) {
    // early return
    return {};
}
```

When a release group has `changelog: { projectChangelogs: false, file: false }`, `g.changelog` evaluates to a truthy object, so the early return is skipped. Users must set `changelog: false` (boolean) to trigger the early return, which is not obvious from the documentation.

### 2. `createProjectGraphAsync()` runs before the "enabled" check

Even if the early return is fixed, `createProjectGraphAsync()` at line 46 runs unconditionally before the changelog-enabled check at line 107. This means a full project graph rebuild happens regardless of whether changelogs are needed.

## Steps to Reproduce

1. Create an Nx workspace with multiple release groups
2. Configure `nx.json` with changelogs disabled using objects instead of boolean `false`:
   ```json
   {
     "release": {
       "groups": {
         "my-group": {
           "changelog": {
             "projectChangelogs": false,
             "file": false
           }
         }
       },
       "changelog": {
         "workspaceChangelog": false
       }
     }
   }
   ```
3. Run `nx release --yes`
4. Observe a significant delay between the version step output and the "Tagging commit with git" message

## Nx Report

```
Node           : 24.13.0
OS             : darwin-arm64
pnpm           : 10.30.3

nx                     : 22.6.0
@nx/js                 : 22.6.0
@nx/docker             : 22.6.0
@nx/devkit             : 22.6.0
```

## Suggested Fix

1. **Normalize the changelog config check** to treat `{ projectChangelogs: false, file: false }` as disabled, or document that users must use `changelog: false` (boolean) to fully disable it.

2. **Move `createProjectGraphAsync()` after the "enabled" check** in `changelog.js`, so the expensive project graph creation is skipped when changelogs are disabled.

3. **Alternatively**, have the top-level `release()` function in `release.js` check if changelogs are truly enabled before calling `releaseChangelog()` at all.

### Workaround

Set per-group changelog to boolean `false` instead of an object:

```json
{
  "release": {
    "groups": {
      "my-group": {
        "changelog": false
      }
    }
  }
}
```

This triggers the early return at line 109 of `changelog.js`, though `createProjectGraphAsync()` at line 46 still runs unnecessarily.

thdk:
## Current Behavior                                                                                                                                          
                  
  When running nx release with changelogs disabled via config objects (e.g., changelog: `{ projectChangelogs: false, file: false })`, the `releaseChangelog()` step still performs expensive operations including a full `createProjectGraphAsync()` call and git log operations. This causes a noticeable delay (30+  seconds in larger monorepos) between the version step and git tagging.
                                                                                                                                                            
  Two root causes:                                                                                                                                        
  1. The changelog "enabled" check treats any truthy object as enabled — { file: false, createRelease: false } passes the check because it's a truthy object, even though it produces no output.                                                                                                                
  2. The top-level release() function calls releaseChangelog() unconditionally, which redundantly recreates the project graph and resolves config even when changelogs are disabled.                                                                                                                                  
                                                                                                                                                            
  Additionally, the "Changelogs are disabled" warning is shown even when changelogs are intentionally disabled via config, which is noisy when called from the top-level release() command.                                                                                                                          
                                                                                                                                                          
##  Expected Behavior                                                                                                                                         
                                                                                                                                                          
  When changelogs are effectively disabled (no file output and no remote release), releaseChangelog() should be skipped entirely by the top-level release() function, avoiding all expensive operations. The "changelogs are disabled" warning should only appear when explicitly running the standalone nx release changelog subcommand.                                                                                                                                     
                                                                                                                                                          
##  Related Issue(s)

  Fixes #35366


thdk:
Anything I can do from my side to resolve the build failure?

JamesHenry (approved):
Thanks @thdk!

PR #35413, opened 2026-04-23
chore(gradle): use task graph when traversing task excludes

lourw:
## Current Behavior

When we calculate the excludes for the gradle executors, we currently traverse the surface level dependencies and also via the project graph. We need to cover all transitive dependencies instead. 

## Expected Behavior

Use the task graph to derive excludes instead. This also is easier to traverse multiple levels of dependencies than using the project graph.

## Related Issue(s)

Fixes #


packages/gradle/src/executors/gradle/gradle-batch.impl.ts
FrozenPandaz: This seems like the same as `getExcludeTasksFromTaskGraph`.. right?

packages/gradle/src/executors/gradle/get-exclude-task.ts
```diff
@@ -1,118 +1,74 @@
...
-    const projectList = Array.isArray(dep.projects)
-      ? dep.projects
-      : [dep.projects];
-    return {
-      project: projectList[0] !== 'self' ? projectList[0] : owningProject,
-      target,
-    };
-  }
-  return { project: owningProject, target };
+export function getGradleTaskName(
```
FrozenPandaz: If the below comment is right.. this might be able to be private?

packages/gradle/src/executors/gradle/get-exclude-task.ts
```diff
@@ -1,118 +1,74 @@
...
   return excludes;
 }
 
-export function getGradleTaskName(
-  pt: Target,
-  nodes: Record<string, ProjectGraphProjectNode>
-): string | null {
-  return nodes[pt.project]?.data?.targets?.[pt.target]?.options?.taskName;
-}
+export function getAllDependsOnFromTaskGraph(
```
FrozenPandaz, line 41: If the below comment is right.. this might be able to be private?

packages/gradle/src/executors/gradle/get-exclude-task.spec.ts
```diff
@@ -1,336 +1,321 @@
-import { getExcludeTasks, getAllDependsOn } from './get-exclude-task';
-import { Target } from '@nx/devkit';
+import {
+  getExcludeTasksFromTaskGraph,
+  getAllDependsOnFromTaskGraph,
```
FrozenPandaz: Do we have to test this separately? It can be tested through getExcludeTasksFromTaskGraph

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #35416, opened 2026-04-23
fix(js): include npm alias targets in pruned lockfile

### LINKED ISSUE #35205: generatePackageJson omits dependencies with package aliases
### Current Behavior

https://github.com/nrwl/nx/issues/27098 

This issue not fixed.

when use 

```
    "zod": "^3.24.2",
    "zod4": "npm:zod@^4.1.13",
```

generatePackageJson just it make zod: "zod": "4.1.13"

### Expected Behavior

```
    "zod": "^3.24.2",
    "zod4": "npm:zod@^4.1.13",
```

### GitHub Repo

_No response_

### Steps to Reproduce

i will make pr.

### Nx Report

```shell
i will make pr.
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

thdk:
## Current Behavior                                                                                                                        
                                                                                                                                            
  When a pnpm lockfile contains npm aliases (e.g. auth0-legacy: auth0@4.37.0 as a dependency of auth0@5.5.0), the @nx/js:prune-lockfile  executor produces a pruned lockfile that is missing the real package snapshot for the alias target.                                       
   
  This causes pnpm install --frozen-lockfile to fail with:                                                                                  
                                                                                                                                          
  ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for 'auth0@4.37.0' in pnpm-lock.yaml                                      
                                                                                                                                            
  The root cause is in two places:                                                                                                          
                                                                                                                                            
  1. getDependencies in pnpm-parser.ts: when building the project graph dependency edges from the lockfile, the alias name (e.g. auth0-legacy) is used to look up the target node (npm:auth0-legacy@... or npm:auth0-legacy). Since no node exists with the alias name, the dependency edge is silently dropped, leaving auth0@4.37.0 unreachable in the graph.                                                      
  2. traverseNode in project-graph-pruning.ts: even if an alias node existed in the graph, the traversal never followed through to the real underlying package node to include it and its transitive dependencies.                                                                    
  
##  Expected Behavior                                                                                                                         
                                                                                                                                          
  The pruned lockfile should include all npm alias targets and their transitive dependencies. pnpm install --frozen-lockfile should succeed without errors when the project uses packages that declare npm alias dependencies.
                                                                                                                                            
##  Related Issue(s)                                                                                                                        
                                                   
  Fixes #35205

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
+++ b/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
@@ -539,9 +539,18 @@ function getDependencies(
                 findVersion(versionRange, name, isV5),
                 isV5
               );
-              const target =
+              let target =
                 ctx.externalNodes[`npm:${name}@${version}`] ||
                 ctx.externalNodes[`npm:${name}`];
+              // For npm aliases (e.g. "auth0-legacy": "auth0@4.37.0"),
+              // the alias name won't match any node. Fall back to
+              // looking up the raw versionRange as a package reference.
+              if (!target && name !== versionRange) {
+                const rawVersion = parseBaseVersion(versionRange, isV5);
+                target =
+                  ctx.externalNodes[`npm:${rawVersion}`] ||
+                  ctx.externalNodes[`npm:${versionRange}`];
+              }
               if (target) {
                 const dep: RawProjectGraphDependency = {
                   source: node.name,
--- a/packages/nx/src/plugins/js/lock-file/project-graph-pruning.spec.ts
+++ b/packages/nx/src/plugins/js/lock-file/project-graph-pruning.spec.ts
@@ -271,6 +271,145 @@ describe('project-graph-pruning', () => {
       expect(result.externalNodes?.['npm:blurhash@2.0.5']).toBeDefined();
     });
 
+    it('should traverse real package node behind npm alias', () => {
+      // Setup: auth0@5.5.0 depends on auth0-legacy (alias for auth0@4.37.0)
+      // auth0@4.37.0 depends on jose. The pruned graph must include all three.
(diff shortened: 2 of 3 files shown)


AgentEnder:
Thanks for the PR. @leosvelperez has it, and it will get a review as time allows.

<details>
<summary>

</summary>

> The body splits the cause in two, a dropped edge in `pnpm-parser` and a traversal that never follows through to the aliased package. Either half on its own still leaves the pruned lockfile broken, so both parts need checking.
>
> Nothing has built this. The head commit's first Actions run needed maintainer approval, never got it, and passed GitHub's 30 day cutoff, at which point the pending run is deleted, so there is nothing left to approve and a push is what creates a fresh one. The `SUCCESS` rollup is the `netlify` deploy preview alone.
>
> #35205 is linked.

</details>

DuncanWilder:
In case it helps get this PR merged, I wanted to highlight that we're experiencing this issue and **this PR fixes the issue**.

From my AI, for further details

> We’re also affected with Nx **23.2.1**, pnpm **12.4.1**, and `@mastra/core` **1.67.0**, using `@nx/esbuild:esbuild` with `generatePackageJson: true`.
>
> The root lockfile contains the alias targets, but Nx’s pruned lockfile omits them. In our case, the frozen production install succeeds, then importing Mastra fails with `Cannot find module '@ai-sdk/provider-utils-v6'`.
>
> We tested the logic from #35416 against our setup: it preserves the missing dependencies, and isolated production installs and Mastra imports pass for both affected apps.

PR #35423, opened 2026-04-24
fix(linter): only rewrite workspace-package peer deps to workspace:*

### LINKED ISSUE #35318: peerDepsVersionStrategy 'workspace' applies workspace:* to all peer deps, not just workspace packages
## Current Behavior

When `peerDepsVersionStrategy: 'workspace'` is configured in `@nx/dependency-checks`, the `--fix` behavior applies `workspace:*` to **all** peer dependencies, including external npm packages like `react`, `axios`, `@reduxjs/toolkit`.

This causes `pnpm install` to fail:

```
ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In libs/my-lib: "react@workspace:*" is in the
dependencies but no package named "react" is present in the workspace
```

The version mismatch check also flags existing peer dependencies with normal semver ranges (e.g. `"react": "^18.3.1"`) and auto-fixes them to `workspace:*`.

## Expected Behavior

`peerDepsVersionStrategy: 'workspace'` should only emit `workspace:*` for packages that are actual workspace members (exist in `projectGraph.nodes`). External npm packages in `peerDependencies` should retain their normal version resolution (installed version or root package.json version).

## Steps to Reproduce

1. Create a pnpm workspace with two packages: `pkg-a` and `pkg-b`
2. `pkg-a` has a build target and depends on both `pkg-b` (workspace) and `react` (npm)
3. In `pkg-a/package.json`, list both in `peerDependencies`: `"pkg-b": "workspace:*", "react": "^18.3.1"`
4. Configure `@nx/dependency-checks` with `peerDepsVersionStrategy: 'workspace'`
5. Run `nx lint pkg-a` -- observe `react` is flagged as version mismatch
6. Run `nx lint pkg-a --fix` -- observe `react` is changed to `workspace:*`
7. Run `pnpm install` -- fails with ERR_PNPM_WORKSPACE_PKG_NOT_FOUND

## Nx Report

```
Node           : 24.14.1
OS             : darwin-arm64
pnpm           : 10.28.2
nx             : 22.6.5
@nx/eslint-plugin : 22.6.5
@nx/js         : 22.6.5
```

## Package Manager Version

pnpm 10.28.2

## Operating System

macOS

## Additional Information

The fix is in `packages/eslint-plugin/src/rules/dependency-checks.ts` at the two code paths that check `peerDepsVersionStrategy === 'workspace'` (~line 214 and ~line 283). They should additionally verify the package is a workspace member (e.g. check `projectGraph.nodes` or confirm it is not in `projectGraph.externalNodes`) before applying `workspace:*`.

Related: PR #33417 that added peerDepsVersionStrategy

FrozenPandaz:
## Current Behavior

The `@nx/eslint-plugin/dependency-checks` rule's `peerDepsVersionStrategy: 'workspace'` option unconditionally rewrote **every** peer dependency to `workspace:*`, including external npm packages like `react`, `axios`, etc. Running `pnpm install` (or any equivalent) on the resulting `package.json` failed with `ERR_PNPM_WORKSPACE_PKG_NOT_FOUND` because those packages do not exist in the workspace. `eslint --fix` was therefore actively breaking working workspaces that enabled the strategy.

Two fix branches in `packages/eslint-plugin/src/rules/dependency-checks.ts` had the bug:

- The `missingDependencies` fix (around L299), which inserted new peer entries at `workspace:*`.
- The version-mismatch fix (around L370-382), which rewrote any non-`workspace:*` range to `workspace:*`.

## Expected Behavior

`peerDepsVersionStrategy: 'workspace'` should only produce `workspace:*` ranges for peer dependencies that are **workspace-published packages**. External npm packages must retain the normal installed-version logic so the resulting `package.json` is installable.

This PR:

- Builds a `workspacePackageNames` `Set<string>` once from `projectGraph.nodes[*].data.metadata.js.packageName`.
- Gates both `peerDepsVersionStrategy === 'workspace'` branches on `workspacePackageNames.has(packageName)`. Workspace packages continue to get `workspace:*`; external packages fall through to the existing installed-version / catalog / root-package-json resolution.

Three existing tests in `dependency-checks.spec.ts` that encoded the buggy behavior (asserting `workspace:*` for external packages) were updated to reflect the correct behavior. Two new focused tests were added to pin down each case explicitly (workspace package → `workspace:*`; external package → installed range preserved).

## Related Issue(s)

Fixes #35318

Follow-up to #33417, which introduced `peerDepsVersionStrategy`.

### DIFF
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -170,14 +170,16 @@ export default ESLintUtils.RuleCreator(
     );
     const expectedDependencyNames = Object.keys(npmDependencies);
 
-    // Package names published by workspace projects. Used to decide whether
-    // `peerDepsVersionStrategy: 'workspace'` should rewrite a peer dep range
-    // to `workspace:*` — only workspace packages should get that treatment;
-    // external npm packages (react, axios, …) must keep their real ranges.
+    // Packages eligible for `workspace:*` rewrites under
+    // `peerDepsVersionStrategy: 'workspace'`. Must be both a workspace project
+    // and registered in the package manager's workspaces — otherwise
+    // `workspace:*` won't resolve at install time.
     const workspacePackageNames = new Set<string>();
     for (const node of Object.values(projectGraph.nodes)) {
-      const name = node.data?.metadata?.js?.packageName;
-      if (name) workspacePackageNames.add(name);
+      const js = node.data?.metadata?.js;
+      if (js?.packageName && js.isInPackageManagerWorkspaces) {
+        workspacePackageNames.add(js.packageName);
+      }
     }
 
     const packageJson = JSON.parse(context.sourceCode.getText());
--- a/packages/eslint-plugin/src/rules/dependency-checks.ts
+++ b/packages/eslint-plugin/src/rules/dependency-checks.ts
@@ -170,6 +170,16 @@ export default ESLintUtils.RuleCreator(
     );
     const expectedDependencyNames = Object.keys(npmDependencies);
 
+    // Package names published by workspace projects. Used to decide whether
+    // `peerDepsVersionStrategy: 'workspace'` should rewrite a peer dep range
+    // to `workspace:*` — only workspace packages should get that treatment;
+    // external npm packages (react, axios, …) must keep their real ranges.
+    const workspacePackageNames = new Set<string>();
+    for (const node of Object.values(projectGraph.nodes)) {
+      const name = node.data?.metadata?.js?.packageName;
+      if (name) workspacePackageNames.add(name);
+    }
+
     const packageJson = JSON.parse(context.sourceCode.getText());
     const projPackageJsonDeps = getProductionDependencies(packageJson);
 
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/dependency-checks.ts
```diff
@@ -170,6 +170,16 @@ export default ESLintUtils.RuleCreator(
...
     const expectedDependencyNames = Object.keys(npmDependencies);
 
+    // Package names published by workspace projects. Used to decide whether
+    // `peerDepsVersionStrategy: 'workspace'` should rewrite a peer dep range
+    // to `workspace:*` — only workspace packages should get that treatment;
+    // external npm packages (react, axios, …) must keep their real ranges.
+    const workspacePackageNames = new Set<string>();
+    for (const node of Object.values(projectGraph.nodes)) {
+      const name = node.data?.metadata?.js?.packageName;
+      if (name) workspacePackageNames.add(name);
```
leosvelperez: Should we also take `metadata.js.isInPackageManagerWorkspaces` into account? If the package is not included in the workspaces, it might not be resolved unless something else (e.g. TS path mappings) is used for resolution.

leosvelperez (approved):

PR #35426, opened 2026-04-24
fix(js): rewrite workspace refs in all dep types during lockfile pruning

### LINKED ISSUE #35425: fix(js): workspace devDependencies ignored by prune-lockfile executor
## Current Behavior

When a pnpm workspace app has a workspace package listed in `devDependencies` (e.g. `"@scope/lib-a": "workspace:*"`), the `@nx/js:prune-lockfile` executor produces output that fails `pnpm install --frozen-lockfile` with:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because

Failure reason:
  specifiers in the lockfile don't match specifiers in package.json
  * 1 dependencies were added: @scope/lib-a@workspace:*
```

This happens because of two bugs that cause workspace packages in `devDependencies` (and `optionalDependencies`/`peerDependencies`) to be silently skipped during lockfile pruning.

## Expected Behavior

The pruned lockfile and rewritten `package.json` should correctly handle workspace packages listed in any dependency section (`dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`), and `pnpm install --frozen-lockfile` should succeed.

Even with `--prod` (which skips devDependency installation), pnpm still validates that the lockfile matches the full `package.json` — including `devDependencies` — before proceeding. So the pruned output must be consistent across all dependency sections.

## Steps to Reproduce

1. Create a pnpm workspace monorepo with Nx
2. Create a buildable library (e.g. `lib-a`)
3. Create an app that lists that library in `devDependencies`:
   ```json
   {
     "devDependencies": {
       "@scope/lib-a": "workspace:*"
     }
   }
   ```
4. Run `nx prune-lockfile <app>`
5. In the pruned output directory, run `pnpm install --frozen-lockfile`
6. Observe `ERR_PNPM_OUTDATED_LOCKFILE` error

### Bug 1: `createPrunedLockfile` only rewrites `dependencies`

The [`createPrunedLockfile`](https://github.com/nrwl/nx/blob/master/packages/js/src/executors/prune-lockfile/prune-lockfile.ts#L66-L77) function rewrites `workspace:*` references to `file:./workspace_modules/<pkg>` in the output `package.json`, but only iterates over `packageJson.dependencies`:

```typescript
for (const [pkgName, pkgVersion] of Object.entries(
  packageJson.dependencies ?? {}
)) {
  if (
    pkgVersion.startsWith('workspace:') ||
    pkgVersion.startsWith('file:') ||
    pkgVersion.startsWith('link:') ||
    workspacePackages.has(pkgName)
  ) {
    packageJson.dependencies[pkgName] = `file:./workspace_modules/${pkgName}`;
  }
}
```

`devDependencies`, `optionalDependencies`, and `peerDependencies` are left with the original `workspace:*` specifier. When pnpm reads the output `package.json`, it sees `workspace:*` but the lockfile has no matching entry, causing the mismatch error.

**Fix**: iterate all four dependency types.

### Bug 2: `mapRootSnapshot` only searches `importerSnapshot.dependencies`

The [`mapRootSnapshot`](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts#L788-L806) function builds the pruned lockfile's root importer snapshot. It iterates all dep types from `package.json`, but when looking up a workspace package in the root lockfile's importer snapshot, it only searches `importerSnapshot.dependencies`:

```typescript
const workspaceDep =
  importerSnapshot.dependencies &&
  importerSnapshot.dependencies[packageName];
```

If the workspace package is in `devDependencies` in the lockfile importer (which it will be, matching `package.json`), the lookup returns `undefined`, and no entry is written to the pruned lockfile.

Additionally, when a workspace dep is found, it always writes to `snapshot.dependencies` regardless of the source section:

```typescript
snapshot.dependencies = snapshot.dependencies || {};
snapshot.dependencies[packageName] = `link:./workspace_modules/${packageName}`;
```

This can cause pnpm to report specifiers "moved" between sections.

**Fix**: search all dep sections in the importer snapshot, and write to the matching section in the output.

## Nx Report

```
Node           : 24.11.0
OS             : darwin-arm64
pnpm           : 10.33.0

nx                     : 22.6.3
@nx/js                 : 22.6.3
@nx/workspace          : 22.6.3
@nx/devkit             : 22.6.3
```

## Failure Logs

See failure run on sample repo: https://github.com/thdk/nx-monorepo-demo/actions/runs/24610859310/job/71964886968

Error from `pnpm install --frozen-lockfile` in the pruned output directory:

```
 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/package.json

Failure reason:
  specifiers in the lockfile don't match specifiers in package.json
  * 1 dependencies were added: @scope/lib-a@workspace:*
```

## Additional Information

- Related to #34655 (transitive workspace deps), but this is a separate bug specifically about non-`dependencies` sections being ignored
- The `workspace:*` protocol mismatch also affects `optionalDependencies` and `peerDependencies`, not just `devDependencies`

thdk:
## Current Behavior

The `@nx/js:prune-lockfile` executor only rewrites `workspace:*` references to `file:./workspace_modules/<pkg>` in `packageJson.dependencies`. Workspace packages listed in `devDependencies`, `optionalDependencies`, or `peerDependencies` are silently skipped.

Similarly, the pnpm lockfile pruner (`mapRootSnapshot`) only searches `importerSnapshot.dependencies` when looking up workspace packages, so workspace deps in other sections produce no entry in the pruned lockfile.

This causes `pnpm install --frozen-lockfile` to fail with `ERR_PNPM_OUTDATED_LOCKFILE` because the lockfile specifiers don't match the `package.json`.

## Expected Behavior

The pruned lockfile and rewritten `package.json` correctly handle workspace packages in all dependency sections (`dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`), and `pnpm install --frozen-lockfile` succeeds.

### Changes

**Bug 1 fix** (`packages/js/src/executors/prune-lockfile/prune-lockfile.ts`):
- Iterate all four dependency types (`dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`) instead of only `dependencies` when rewriting workspace/file/link references in the output `package.json`.

**Bug 2 fix** (`packages/nx/src/plugins/js/lock-file/pnpm-parser.ts`):
- Search `importerSnapshot.dependencies`, `devDependencies`, and `optionalDependencies` when looking up workspace packages in the root lockfile importer snapshot.
- Write the pruned entry to the matching dep section (instead of always writing to `dependencies`), so pnpm doesn't report specifiers "moved" between sections.

## Related Issue(s)

Fixes #35425

### DIFF
--- a/packages/js/src/executors/prune-lockfile/prune-lockfile.ts
+++ b/packages/js/src/executors/prune-lockfile/prune-lockfile.ts
@@ -63,16 +63,24 @@ function createPrunedLockfile(packageJson: PackageJson, graph: ProjectGraph) {
 
   const workspacePackages = getWorkspacePackagesFromGraph(graph);
 
-  for (const [pkgName, pkgVersion] of Object.entries(
-    packageJson.dependencies ?? {}
-  )) {
-    if (
-      pkgVersion.startsWith('workspace:') ||
-      pkgVersion.startsWith('file:') ||
-      pkgVersion.startsWith('link:') ||
-      workspacePackages.has(pkgName)
-    ) {
-      packageJson.dependencies[pkgName] = `file:./workspace_modules/${pkgName}`;
+  for (const depType of [
+    'dependencies',
+    'devDependencies',
+    'optionalDependencies',
+    'peerDependencies',
+  ] as const) {
+    for (const [pkgName, pkgVersion] of Object.entries(
+      packageJson[depType] ?? {}
+    )) {
+      if (
+        pkgVersion.startsWith('workspace:') ||
+        pkgVersion.startsWith('file:') ||
+        pkgVersion.startsWith('link:') ||
+        workspacePackages.has(pkgName)
+      ) {
+        packageJson[depType][pkgName] =
+          `file:./workspace_modules/${pkgName}`;
+      }
     }
   }
 
--- a/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
+++ b/packages/nx/src/plugins/js/lock-file/pnpm-parser.ts
@@ -785,9 +785,11 @@ function mapRootSnapshot(
           for (const [importerPath, importerSnapshot] of Object.entries(
             rootImporters
           )) {
+            // Search all dep sections in the importer, not just dependencies
             const workspaceDep =
-              importerSnapshot.dependencies &&
-              importerSnapshot.dependencies[packageName];
+              importerSnapshot.dependencies?.[packageName] ||
+              importerSnapshot.devDependencies?.[packageName] ||
+              importerSnapshot.optionalDependencies?.[packageName];
             if (workspaceDep) {
               const workspaceDepImporterPath = workspaceDep.replace(
                 'link:',
@@ -800,8 +802,11 @@ function mapRootSnapshot(
               importers[packageName] = importerKeyForPackage;
               snapshot.specifiers[packageName] =
                 `file:./workspace_modules/${packageName}`;
-              snapshot.dependencies = snapshot.dependencies || {};
-              snapshot.dependencies[packageName] =
+              // Write to the same section as in the original package.json
+              const section =
+                depType === 'peerDependencies' ? 'dependencies' : depType;
+              snapshot[section] = snapshot[section] || {};
+              snapshot[section][packageName] =
                 `link:./workspace_modules/${packageName}`;
               break;
             }


leosvelperez:
Superseded by #36139, which includes this fix as part of a more comprehensive solution for pruned pnpm deployments. Thank you for the contribution and investigation.

PR #35429, opened 2026-04-24
feat(vite): add compiler option to vite plugin for tsgo support

estevaolucas:
## Description

Adds a `compiler` option to the `@nx/vite` plugin, mirroring the same option already in `@nx/js` (added in #33821).

This lets users specify an alternative TypeScript compiler for the inferred `typecheck` target — specifically `tsgo` from `@typescript/native-preview` (TypeScript 7 Go compiler).

## Problem

`@nx/vite` hardcodes `'tsc'` for typecheck while `@nx/js` already supports a configurable `compiler` option. Users who want `tsgo` have to patch `@nx/vite` manually.

## Solution

3 lines in `packages/vite/src/plugins/plugin.ts`:

1. Add `compiler?: string` to `VitePluginOptions` (with JSDoc)
2. `options.compiler ?? 'tsc'` instead of hardcoded `'tsc'` (Vue projects still use `vue-tsc`)
3. `options.compiler ??= 'tsc'` in `normalizeOptions`

## Usage

```json
// nx.json
{
  "plugins": [
    {
      "plugin": "@nx/vite/plugin",
      "options": {
        "compiler": "tsgo"
      }
    }
  ]
}
```

## Benchmarks (large monorepo, ~400 projects)

| Project | `tsc` | `tsgo` | Speedup |
|---|---|---|---|
| `dashboard-web` | 2.33s | 0.43s | **5.4×** |
| `market-react` | 11.57s | 3.64s | **3.2×** |

On CI: total typecheck CPU dropped **2.7×**, allowing us to eliminate worker sharding entirely.

Currently working around this with a pnpm patch on `@nx/vite` — happy to remove it once this lands.

## Prior art

- #33821 — same `compiler` option added to `@nx/js`
- #35047 / #35167 — Nx team experimented with tsgo internally

### DIFF
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -423,7 +423,9 @@ async function buildViteTargets(
     const hasVuePlugin = viteBuildConfig.plugins?.some(
       (p) => p.name === 'vite:vue'
     );
-    const typeCheckCommand = hasVuePlugin ? 'vue-tsc' : (options.compiler ?? 'tsc');
+    const typeCheckCommand = hasVuePlugin
+      ? 'vue-tsc'
+      : (options.compiler ?? 'tsc');
 
     targets[options.typecheckTargetName] = {
       cache: true,
--- a/packages/vite/src/plugins/plugin.ts
+++ b/packages/vite/src/plugins/plugin.ts
@@ -42,10 +42,12 @@ export interface VitePluginOptions {
   serveStaticTargetName?: string;
   typecheckTargetName?: string;
   /**
-   * The compiler to use for type-checking. Defaults to `tsc`.
-   * Set to `tsgo` to use the TypeScript Go compiler (`@typescript/native-preview`).
+   * The compiler to use for type-checking. When unset, defaults to `vue-tsc`
+   * for Vue projects (detected via the `vite:vue` plugin) and `tsc` otherwise.
+   * Set to `tsgo` to use the TypeScript Go compiler (`@typescript/native-preview`),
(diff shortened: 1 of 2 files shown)


barbados-clemens (approved):

FrozenPandaz (approved):

packages/vite/src/plugins/plugin.ts
```diff
@@ -418,7 +423,9 @@ async function buildViteTargets(
     const hasVuePlugin = viteBuildConfig.plugins?.some(
       (p) => p.name === 'vite:vue'
     );
-    const typeCheckCommand = hasVuePlugin ? 'vue-tsc' : 'tsc';
+    const typeCheckCommand = hasVuePlugin
+      ? 'vue-tsc'
+      : (options.compiler ?? 'tsc');
```
comp615: Imo, the explicit compiler should take precedent over `hasVuePlugin` (e.g. what if we have vue-tsgo). So:
```
const typeCheckCommand = options.compiler ?? hasVuePlugin
      ? 'vue-tsc'
      : 'tsc';
```

But the NX folks can probably be the arbiters

FrozenPandaz: Yeah it's a good consideration.

Ultimately.. I'm not sure we can anticipate that in this version. The gap is that we also need the right external dependencies.. which sometimes aren't exactly the same as the compiler... so we're going to explicitly handle these 3 valid compiler settings.

When there is a vue-tsgo alternative we can look separately into providing support.

FrozenPandaz:
Thank you for the contribution!

PR #35441, opened 2026-04-24
chore(repo): prevent unit tests from walking the real workspace

AgentEnder:
## Current Behavior

Running `nx:test` (jest) from `packages/nx` produces ~4500 cross-project sandbox violations. Reads span every plugin's `src/generators/**`, `src/migrations/**`, `src/executors/**`, schemas, docs, spec files and `.gitignore`s — essentially the entire monorepo.

Root cause is that tests end up computing the **real** project graph:

- `packages/nx/src/utils/workspace-root.ts` freezes `workspaceRoot` on first import by walking up from `process.cwd()`. With jest's cwd set to `packages/nx` and no `NX_WORKSPACE_ROOT_PATH` env var, it resolves to the real repo root.
- `scripts/patched-jest-resolver.js` did set `NX_WORKSPACE_ROOT_PATH`, but only when `process.argv[1]` contained `jest-worker` or `argv[3]` had `:test`. Under `jest --passWithNoTests --detectOpenHandles --forceExit` with `maxWorkers: 1` (what the Nx test runs use), neither branch matches, so the env var was never set.
- The existing `@nx/devkit.createProjectGraphAsync` mock in `scripts/unit-test-setup.js` is specifier-keyed, so relative imports inside `packages/nx` (e.g. `'../../project-graph/project-graph'`) bypass it and call the real graph builder.

The smoking gun in the process tree: hundreds of `plugin-worker.ts` subprocesses and `git remote-https` calls to `nrwl/nx-ai-agents-config` — both only happen when the real, isolated project graph is computed.

## Expected Behavior

Unit tests never touch the real workspace filesystem or spawn plugin workers. Three layered fixes:

1. **`scripts/patched-jest-resolver.js`** — always set `NX_WORKSPACE_ROOT_PATH` to a throwaway `tmp/unit` dir. Runs at resolver module-load, which is before any `require('nx/...')`, so `workspace-root.ts` is guaranteed to freeze to the sandbox dir.
2. **`scripts/unit-test-setup.js`** — add `jest.doMock('nx/src/project-graph/project-graph', …)` that returns an empty graph for `createProjectGraphAsync`, `createProjectGraphAndSourceMapsAsync`, and `buildProjectGraphAndSourceMapsWithoutDaemon`. Jest keys mocks by resolved absolute path, so the relative imports inside `packages/nx` hit the same mock.
3. **`scripts/unit-test-setup.js`** — guard mock on `loadIsolatedNxPlugin`. If any test slips past the graph mocks and tries to spawn a plugin worker, it throws with an actionable error instead of silently scanning the monorepo.

`packages/nx/src/utils/workspace-root.spec.ts` directly exercises the walk-up logic in `workspaceRootInner`, which now short-circuits on the always-set env var. The suite scopes the var away in `beforeAll`/`afterAll`.

## Related Issue(s)

This is opened as a draft to validate the hypothesis by re-running the CI sandbox report and confirming the cross-project violations drop.

Local smoke test: `packages/nx` suites across `workspace-root`, `run-many`, `affected`, `release/config`, `migrate`, `task-hasher`, `native-task-hasher-impl`, `package-json/create-nodes`, `project-json/build-nodes`, `explicit-*-dependencies`, `normalize-project-nodes`, `show/projects`, and both `isolation/*` specs — 11 suites, 177 tests, all passing.


AgentEnder:
Force-corrected the previous attempt — the first push was too aggressive on two axes:

- Setting `NX_WORKSPACE_ROOT_PATH` unconditionally in `patched-jest-resolver.js` broke a lot of unrelated suites that asserted on the real `workspaceRoot`. **Reverted** the conditional gate in the resolver.
- Mocking `loadIsolatedNxPlugin` to throw on every call blocked legitimate uses against `TempFs` roots (e.g. `getOnlyDefaultPlugins(tempFs.tempDir)` in `explicit-project-dependencies.spec.ts`). **Narrowed** the guard to inspect the `root` argument and only throw when it equals the real workspace root; passes through otherwise.

Also added `jest.unmock('./project-graph')` to `project-graph.spec.ts`, the one suite that exercises the real `buildProjectGraphAndSourceMapsWithoutDaemon` and would otherwise hit the global graph mock.

Verified the full `packages/nx` jest suite now matches the master baseline (5 pre-existing suite failures, all unrelated to this change — Node 24 / `--experimental-vm-modules` issues on master).

FrozenPandaz (approved):

PR #35464, opened 2026-04-27
fix(core): avoid workspace-wide walk in expand_outputs

marco2216:
## Current Behavior

When a cached target's `outputs` array contains any entry with a glob (`*`, `**`), the native `expand_outputs(workspaceRoot, outputs[])` (called from `NxCache.put` / `copyFilesFromCache` via `_expandOutputs` in `packages/nx/src/tasks-runner/cache.ts`) walks the **entire workspace** once per call - regardless of the glob's anchor depth or how few files it matches. Even a glob anchored to a leaf directory containing one file pays the full cost.

On a 50k-file workspace (warm OS cache, M-series Mac SSD) this is ~55 ms per call; on production workspaces with several hundred thousand tracked files / build artifacts it scales to multiple seconds per cached task. Concrete file paths and pure directory paths take a fast path inside the same Rust function (<1 ms) - the cost is specifically tied to the workspace-rooted walk triggered by glob entries.

The sibling `get_files_for_outputs` (in the same module, exposed via `getFilesForOutputsBatch` and already used by `outputs-tracking.ts`) already does this correctly: it partitions globs by their anchored root using `partition_glob` and walks only those subtrees. So the underlying primitive is already fast on the same inputs.

## Expected Behavior

`expand_outputs` should resolve glob entries without walking the entire workspace, matching the behavior of `get_files_for_outputs`. On the same inputs the two functions return the same flat list of resolved files; routing the glob branch through `get_files_for_outputs` makes `expand_outputs` ~27x faster on globs and constant-time independent of workspace size.

## Patch

`packages/nx/src/native/cache/expand_outputs.rs`: in the glob branch of `_expand_outputs`, when there are no negated globs (the common case), delegate to `get_files_for_outputs` instead of building a workspace-rooted `nx_walker_sync`. Negated globs keep the existing workspace-rooted walk so that ignore semantics (`!path/to/exclude`) are preserved exactly. No JS or napi-rs surface change - all native callers and the JS `expandOutputs` binding benefit transparently.

## Reproducer & numbers

Standalone reproducer: https://github.com/marco2216/nx-expand-outputs-glob-perf-repro — `pnpm install && bash bench.sh`. Default seed is 50k synthetic files, generated locally on first run; scale up via `SEED_FILES=200000 bash bench.sh` to amplify the bug. nx is pinned to current latest.

```
$ bash bench.sh   # before
concrete (1 entry):    mean=0.00 ms
glob (1 entry, *):     mean=54.80 ms
glob (1 entry, **):    mean=55.44 ms
glob (5 entries, *):   mean=53.77 ms   # single walk per call, batched across entries
directory (1 entry):   mean=0.00 ms

$ bash bench.sh   # after this PR (verified against the equivalent JS-side patch
                  # routing _expandOutputs through getFilesForOutputsBatch)
concrete (1 entry):    mean=0.00 ms
glob (1 entry, *):     mean=2.02 ms
glob (1 entry, **):    mean=1.88 ms
glob (5 entries, *):   mean=1.82 ms
directory (1 entry):   mean=1.76 ms
```

In a large monorepo we maintain, applying the equivalent change as a workspace-side workaround (removing globs from `outputs`) drops our prebuild from 22 s to 9 s serial, and 15 s to 3.7 s parallel (~4x speedup). The reproducer is a synthetic minimum; the real-world win on production codebases is substantially larger.

## Correctness verification

Ran cache miss -> cache hit cycle for targets with both concrete and glob `outputs` configs in the reproducer; both restore output files byte-correct on cache hit. The existing Rust test suite for this module (`should_expand_outputs`, `should_handle_multiple_extensions`, `should_handle_multiple_outputs_with_negation`, `should_expand_outputs_with_symlinks_and_globs`, `should_handle_cache_put_and_restore_with_symlinks`) was reviewed against the new code path:
- Non-glob entries: unchanged fast path.
- Glob entries without negation: now flow through `get_files_for_outputs`, which produces the same result set on these tests.
- Glob entries with negation: unchanged; still walks workspace root with `nx_walker_sync` so `!apps/web/.next/cache` semantics are preserved exactly.

## New test

Added `should_not_walk_outside_glob_anchored_root` in `expand_outputs.rs` that:
- builds a temp workspace where the glob's anchored root contains 2 matching files,
- adds decoy files outside the anchored root (`src/`, `dist/other-app/`, `noise/file-{0..49}.txt`) that the previous workspace-wide walker would visit,
- calls `expand_outputs("dist/app/**/*.js")` and asserts the result matches `get_files_for_outputs` on the same inputs and contains exactly the two anchored-root files.

This catches both functional drift between the two functions and any future regression that reintroduces a workspace-rooted walk in the glob fast path.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/native/cache/expand_outputs.rs
+++ b/packages/nx/src/native/cache/expand_outputs.rs
@@ -66,7 +66,7 @@ where
     // `get_files_for_outputs`, which partitions globs by their anchored root
     // and walks only those subtrees (instead of the entire workspace). On
     // workspaces with many tracked files this avoids a costly full-tree walk
-    // per `NxCache.put` / `copyFilesFromCache` call (see #<issue>).
+    // per `NxCache.put` / `copyFilesFromCache` call (see https://github.com/nrwl/nx/pull/35464).
     //
     // Negated globs still take the slower workspace-rooted walk so that
     // ignore semantics (`!path/to/exclude`) are preserved exactly.
--- a/packages/nx/src/native/cache/expand_outputs.rs
+++ b/packages/nx/src/native/cache/expand_outputs.rs
@@ -62,6 +62,22 @@ where
         .map(|s| s[1..].to_string())
         .collect::<Vec<_>>();
 
+    // Fast path: when there are no negated globs, delegate to
+    // `get_files_for_outputs`, which partitions globs by their anchored root
+    // and walks only those subtrees (instead of the entire workspace). On
+    // workspaces with many tracked files this avoids a costly full-tree walk
+    // per `NxCache.put` / `copyFilesFromCache` call (see #<issue>).
+    //
+    // Negated globs still take the slower workspace-rooted walk so that
+    // ignore semantics (`!path/to/exclude`) are preserved exactly.
+    if negated_globs.is_empty() {
+        trace!(
+            "No negated globs; delegating to get_files_for_outputs for {} entries",
+            regular_globs.len()
+        );
+        return get_files_for_outputs(directory.as_path(), regular_globs);
+    }
+
     let regular_globs = regular_globs
         .into_iter()
         .map(|s| {
@@ -322,6 +338,55 @@ mod test {
                 "packages/nx/src/native/nx.darwin-arm64.node",
                 "test.txt"
             ]
+        );
+    }
+
+    #[test]
+    fn should_not_walk_outside_glob_anchored_root() {
+        // Regression test: when a glob is anchored to a deep subdirectory
+        // (e.g. `dist/app/**/*`) the previous implementation walked the entire
+        // workspace, paying O(workspace size) per call. After delegating the
+        // non-negated path to `get_files_for_outputs`, we should only walk
+        // the anchored subtree, so files that live outside that subtree must
+        // never be returned and the result must match what
+        // `get_files_for_outputs` produces for the same inputs.
+        let temp = TempDir::new().unwrap();
+
+        // Files inside the glob's anchored root
+        temp.child("dist/app/index.js").write_str("a").unwrap();
+        temp.child("dist/app/nested/chunk.js")
+            .write_str("b")
+            .unwrap();
+
+        // Decoy files outside the glob's anchored root that the old
+        // workspace-wide walker would still visit (cost) but that must
+        // never appear in the result.
+        temp.child("src/main.ts").write_str("src").unwrap();
+        temp.child("dist/other-app/index.js")
+            .write_str("other")
+            .unwrap();
+        for i in 0..50 {
+            temp.child(format!("noise/file-{i}.txt"))
+                .write_str("noise")
+                .unwrap();
+        }
+
+        let entries = vec!["dist/app/**/*.js".to_string()];
+
+        let mut expanded = expand_outputs(temp.display().to_string(), entries.clone()).unwrap();
+        expanded.sort();
+
+        let mut via_get_files = get_files_for_outputs(temp.path(), entries).unwrap();
+        via_get_files.sort();
+
+        // expand_outputs and get_files_for_outputs must agree on this input.
+        assert_eq!(expanded, via_get_files);
+
+        // Sanity: the resolved set is exactly the files under the anchored
+        // root, with nothing from `src/`, `dist/other-app/`, or `noise/`.
+        assert_eq!(
+            expanded,
+            vec!["dist/app/index.js", "dist/app/nested/chunk.js"]
         );
     }
 
--- a/packages/nx/src/native/cache/expand_outputs.rs
+++ b/packages/nx/src/native/cache/expand_outputs.rs
@@ -62,6 +62,22 @@ where
         .map(|s| s[1..].to_string())
         .collect::<Vec<_>>();
 
+    // Fast path: when there are no negated globs, delegate to
+    // `get_files_for_outputs`, which partitions globs by their anchored root
+    // and walks only those subtrees (instead of the entire workspace). On
+    // workspaces with many tracked files this avoids a costly full-tree walk
+    // per `NxCache.put` / `copyFilesFromCache` call (see #<issue>).
+    //
+    // Negated globs still take the slower workspace-rooted walk so that
+    // ignore semantics (`!path/to/exclude`) are preserved exactly.
+    if negated_globs.is_empty() {
+        trace!(
+            "No negated globs; delegating to get_files_for_outputs for {} entries",
+            regular_globs.len()
+        );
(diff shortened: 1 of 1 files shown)


marco2216:
Hey @FrozenPandaz would you mind having a look at this and giving me some feedback on whether this issue and solution approach makes sense? Ofc. it's not optimal that this issue remains if negated globs are there, so maybe there's a better way.  Thank you!

marco2216:
Hello there, do you all need anything to be able to review this? Been waiting for a review for a few months now, if you don't think this makes sense please le me know! Sorry for tagging but not sure what else I can do.
@jaysoo @leosvelperez

amitkastel:
Any updates regarding the review? Could be super handful.

AgentEnder (changes requested):
Thanks for this, and for the benchmarks and the standalone reproducer. The diagnosis is right. `_expand_outputs` really does walk the whole workspace once per cached task as soon as any `outputs` entry contains a glob, and anchoring the walk with `partition_glob` is the right way to fix it.

The problem is the delegation target. `get_files_for_outputs` is documented as returning "a list of existing files" and walks with `nx_walker`, which drops directories. `_expand_outputs` is documented as returning "directories and files", and `cache.rs` copies exactly what it returns. So the claim that the two return the same list holds for files, but not for the contract.

I ran both versions over the same fixtures. Files match everywhere. Directories do not:

| entries | base | this PR |
| --- | --- | --- |
| `dist/app/**/*.js` | 3 files | same |
| `dist/app/**` | 6, incl. `dist/app/emptydir`, `dist/app/nested` | 4, both directories gone |
| `dist/app` + `dist/lib/**/*.js` | 7, incl. both directories | 5, both gone |
| `dist/app` + `!dist/app/nested` | 4 | same (negated branch untouched) |

`dist/app/**` is the row that worries me. It is one of the most common `outputs` values, and it is a pure glob, so this is not limited to mixed entry lists.

Four things follow, all measured through the shipped put and restore path:

1. Empty output directories are never stored and never restored. Nothing else in the result set implies them.
2. Restore stops clearing stale files. The directory entry is what triggered `remove_dir_all`, so a file left behind by another branch now survives a cache hit, and the restored tree stops matching a real build.
3. Overlapping entries produce duplicates. `get_files_for_outputs` accumulates per root with no dedupe, so `["{options.outputPath}", "{options.outputPath}/**/*.map"]` copies every map file twice and records twice its size. `maxCacheSize` eviction reads that number.
4. A symlinked directory now gets followed. `nx_walker` hands the anchored root to `ignore::WalkBuilder`, which follows the walk root when the root is itself a symlink. `nx_walker_sync` never did. This matters most on the restore side, where the root is the cache artifact directory rather than the workspace.

All four have one cause and one fix, and it is close to what you already wrote. Keep `nx_walker_sync` and the existing `glob_set.is_match` filter, but take the walk roots from `partition_globs_into_map` (already in this file at :170) instead of starting at `directory`, then re-prefix each result with its root the way `get_files_for_outputs` does. Same anchoring and the same speedup, without the contract change: `nx_walker_sync` yields directories and does not follow a symlinked root. It would also let you extend the optimization to the negated branch later, since its ignores are matched as `**/{pattern}` and so are anchor-independent.

One note on the test. `should_not_walk_outside_glob_anchored_root` passes unchanged against `master`; I copied it onto the base file to check. Its first assertion compares `expand_outputs` against `get_files_for_outputs`, and after this change those are the same call for that input. The literal assertion below it uses a shape that does not differ between base and master. A case built on `dist/app/**` over a fixture containing an empty directory would fail without the fix.

Thanks again. This is a real problem, and the measurements you did made it easy to check.

marco2216:
Thanks — the contract critique is right, and I'd rather agree on the shape before pushing anything, because the part I got stuck on is the root source you pointed at.

I prototyped the change locally: `nx_walker_sync` and the `glob_set.is_match` filter kept as they are, only the walk roots moved. That does hold the contract — directories, empty directories, the `remove_dir_all` on restore and the no-duplicates property all behave as they do on `master` (I diffed the expansions against the current implementation over 15 entry shapes: directories, globs, negations, overlaps, scoped names, symlinks, ignored dirs), and `["dist/app/**"]` on a 20k-file fixture goes from 19.3ms to 0.14ms per call.

The roots can't come from `partition_globs_into_map`, though. `partition_glob` runs each segment through `parse_glob`, which consumes extglob characters even when no group follows them:

| entry | `partition_glob` root |
| --- | --- |
| `dist/app/**` | `dist/app` ✅ |
| `dist/libs/@scope/pkg/` | `dist/libs/scope/pkg` ❌ |
| `dist/libs/a+b/` | `dist/libs/a` ❌ |

A walk anchored at a path that doesn't exist finds nothing, so that version fails the repo's own `should_match_output_paths_consistently_with_expand_outputs` — "fixture expanded to nothing for entries `["dist/libs/@scope/pkg"]`". It's the trap `match_output_paths` already documents: `@`, `+` and `,` are ordinary in directory names.

Two more things the workspace-wide walk did for free on the way down, which any anchored version has to reproduce:

- `nx_walker_sync` **does** follow a symlinked walk root — `walkdir` descends when the root itself is the link. It never showed before because the walk always started at the workspace.
- Ignored directories: anchoring inside `node_modules`, or inside a negated entry, skips the pruning `filter_entry` used to do on the way down.

So there are two ways to go, and I'd like your call on which:

1. **Anchor on a conservative literal prefix, in this PR.** Take the leading segments `contains_glob_pattern` leaves alone (`dist/libs` for `dist/libs/@scope/pkg`), drop roots nested inside another root, and handle the two cases above explicitly. A shorter root only widens the walk, while a wrong one drops files, so erring short is safe. Self-contained, and what I have working locally.
2. **Fix the root cause first, in its own PR.** Make `parse_glob` treat `@ + ? ! *` as extglob only when followed by `(`, so literal segments survive partitioning. Then `partition_glob` gives exact roots and this PR becomes what you described.

   The same mangling is also live today, outside this PR: for `outputs: ["dist/libs/@scope/pkg/**/*.js"]`, `get_files_for_outputs` walks `dist/libs/scope/pkg`, finds nothing, and returns `[]`, so `hash_task_output` - the `dependentTasksOutputFiles` input - hashes over an empty set. I changed the output file's contents between two calls and got the same hash (`3244421341483603138`) both times; the identical test against `dist/libs/plain/pkg` resolves the file and the hash changes. It needs a glob entry - a plain directory entry takes the `path.exists()` branch and is fine - which is probably why it has not surfaced. I have both failing tests ready.

   `parse_glob` feeds `convert_glob`, every glob set and the project-graph globbing, so it wants its own change with those tests in front of it.

Happy to do either, or 2 then 1.

PR #35489, opened 2026-04-29
fix(core): preserve user-supplied nx.metadata in package.json projects

wilcoxmd:
## Current Behavior

User-supplied `nx.metadata` keys in a project's `package.json` are discarded by the package-json plugin. Given:

```jsonc
// apps/my-app/package.json
{
  "name": "my-app",
  "description": "my app",
  "nx": {
    "metadata": { "foo": "bar" }
  }
}
```

`nx show project my-app --json | jq .metadata` returns only `{ "targetGroups": {...}, "description": ... }` — the `foo` key is gone.

The cause is in [`getMetadataFromPackageJson`](https://github.com/nrwl/nx/blob/master/packages/nx/src/utils/package-json.ts#L149-L160): the function destructures `nx` from `packageJson` but does not spread `nx?.metadata` into the returned object. That object then replaces (rather than merges with) the user's metadata when the plugin spreads it onto the project configuration in [`buildProjectConfigurationFromPackageJson`](https://github.com/nrwl/nx/blob/master/packages/nx/src/plugins/package-json/create-nodes.ts#L182).

This makes `nx.metadata` unusable as an extension point for user tooling that wants to attach annotations to projects via package.json. 

## Expected Behavior

User-supplied keys on `nx.metadata` are preserved alongside the auto-generated `targetGroups` and `description`. When a key collides, the user-supplied value wins.

After this change:

```
$ nx show project my-app --json | jq .metadata
{
  "targetGroups": { "NPM Scripts": [...] },
  "description": "...",
  "foo": "bar"
}
```

## Related Issue(s)

N/A

### DIFF
--- a/packages/nx/src/utils/package-json.ts
+++ b/packages/nx/src/utils/package-json.ts
@@ -13,7 +13,10 @@ import {
 } from '../config/workspace-json-project-json';
 import type { Tree } from '../generators/tree';
 import { readJson } from '../generators/utils/json';
-import { mergeTargetConfigurations } from '../project-graph/utils/project-configuration/target-merging';
+import {
+  mergeMetadata,
+  mergeTargetConfigurations,
+} from '../project-graph/utils/project-configuration/target-merging';
 import { getCatalogManager } from './catalog';
 import { readJsonFile } from './fileutils';
 import { getNxRequirePaths } from './installation-directory';
@@ -195,7 +198,7 @@ export function getMetadataFromPackageJson(
   const { scripts, nx, description, name, exports, main, version } =
     packageJson;
   const includedScripts = nx?.includedScripts || Object.keys(scripts ?? {});
-  return {
+  const inferredMetadata: ProjectMetadata = {
     targetGroups: {
       ...(includedScripts.length ? { 'NPM Scripts': includedScripts } : {}),
     },
@@ -207,8 +210,8 @@ export function getMetadataFromPackageJson(
       packageMain: main,
       isInPackageManagerWorkspaces,
     },
-    ...nx?.metadata,
   };
+  return mergeMetadata(null, null, null, nx?.metadata ?? {}, inferredMetadata);
 }
 
 export function getTagsFromPackageJson(packageJson: PackageJson): string[] {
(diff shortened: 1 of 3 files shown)


packages/nx/src/utils/package-json.ts
```diff
@@ -207,6 +207,7 @@ export function getMetadataFromPackageJson(
       packageMain: main,
       isInPackageManagerWorkspaces,
     },
+    ...nx?.metadata,
```
AgentEnder: We should probably use this util here: 

https://github.com/nrwl/nx/blob/5805c5f3fb235014c3b6e3538fee0d34edd5242e/packages/nx/src/project-graph/utils/project-configuration/target-merging.ts#L49

Its not gonna be super ergonomic, but invocation would be smth like

```
mergeMetadata(null,null,null,inferredMetadata,packageJson.nx?.metadata ?? {})
```

AgentEnder (changes requested):

packages/nx/src/utils/package-json.ts
wilcoxmd: Oh, thanks for the pointer! Will update the change to use this!

wilcoxmd: @AgentEnder I updated the PR to use that util! I have the arguments ordered so that user-supplied keys will still take priority, which is a slightly different argument order than you shared. But if you ordered your example intentionally because you _don't_ want user-specified keys to have priority then let me know and we can switch the args around.

AgentEnder (approved):

AgentEnder:
@copilot resolve the merge conflicts in this pull request

AgentEnder:
Thanks for the PR, and for turning the review round. @AgentEnder has already approved this and CI is green, so what is left is landing it rather than reviewing it. The one thing in the way is that it now conflicts with `master`, so could you rebase? He takes it from there.

<details>
<summary>

</summary>

> The `SUCCESS` rollup on the head commit comes off four concluded `github-actions` suites, so this one genuinely ran.
>
> Not linked: no GitHub closing reference and no Linear key. That normally leans an unlinked community PR towards closing, but the request-changes-then-approve cycle here is the agreement a linked issue would have supplied.

</details>

PR #35493, opened 2026-04-29
fix(angular): disable vitest watch by default

leosvelperez:
## Current Behavior

Generated Angular projects using `vitest-angular` create a test target without an explicit `watch` value. The Angular unit-test builder defaults watch mode to `true` in TTY environments, so running generated projects through monorepo workflows such as `nx run-many` can leave test tasks running instead of exiting.

## Expected Behavior

Generated Angular `vitest-angular` test targets explicitly set `watch: false`, matching Nx Vitest's default non-watch behavior and preserving terminating test tasks for `run-many` and affected workflows. Users can still opt into watch mode with `--watch` or a watch configuration.

### DIFF
--- a/packages/angular/src/generators/utils/add-vitest.ts
+++ b/packages/angular/src/generators/utils/add-vitest.ts
@@ -50,7 +50,7 @@ export async function addVitestAngular(
     : '@angular/build:unit-test';
   const project = readProjectConfiguration(tree, options.name);
   project.targets ??= {};
-  project.targets.test = { executor, options: {} };
+  project.targets.test = { executor, options: { watch: false } };
   updateProjectConfiguration(tree, options.name, project);
 
   const nxJson = readNxJson(tree);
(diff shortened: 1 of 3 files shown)


packages/angular/src/generators/utils/add-vitest.ts
```diff
@@ -50,7 +50,7 @@ export async function addVitestAngular(
     : '@angular/build:unit-test';
   const project = readProjectConfiguration(tree, options.name);
   project.targets ??= {};
-  project.targets.test = { executor, options: {} };
+  project.targets.test = { executor, options: { watch: false } };
```
juristr, line 53: should we generate this for new setups or implicitly have a `watch: false` and allow ppl to override in the `options` if they want to?

leosvelperez: The Angular CLI builder wouldn't have the implicit value, so we'd need to have different behavior/generation depending on the executor. This generator supports using one or the other.

I prefer keeping the executor aligned with the upstream Angular CLI builder and with an explicit option. It also prevents changing the behavior of existing workspaces, which would require us to provide a migration to opt them out anyway.

AgentEnder (approved):

Issue #35444, opened 2026-04-24
NX is downloading 22.7 without asking


ssilva-soap:
### Current Behavior

When I start any task using NX, the Daemon is starting, downloading the new 22.7 version, and I get the "Server stopped because: "NX_VERSION_CHANGED""

### Expected Behavior

I don't wanna update to the new version now. I wanna keep using the 22.6.5

### Steps to Reproduce

1.
Start any task that use NX

### Nx Report

```shell
I can't run, it get stuck
```

### Failure Logs

```shell
DEBUG nx::native::ide::install: Nx Console extension is not installed
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:39.323Z - [NX-CONSOLE]: Console status check completed, shouldPrompt: false
DEBUG nx::native::ide::install: Nx Console extension is not installed
DEBUG nx::native::ide::install: Cursor extension installation not supported on this platform
DEBUG nx::native::ide::install: No command available for editor: Cursor
DEBUG nx::native::ide::install: Cursor extension installation not supported on this platform
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.616Z - [Server] Daemon outdated: NX_VERSION_CHANGED
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.616Z - [Server] Shutting down daemon (no restart)...
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.619Z - [WATCHER]: Processing file changes in outputs
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.623Z - [WATCHER]: Stopping the watcher for *** (sources)
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.623Z - Closed a connection. Number of open connections: 0
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.623Z - [WATCHER]: Stopping the watcher for *** (outputs)
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.624Z - [LATEST-NX]: Cleaning up latest Nx installation from /tmp/tmp-149751-2DKo79V25NZ9
[NX v22.6.5 Daemon Server] - 2026-04-24T21:36:40.625Z - Server stopped because: "NX_VERSION_CHANGED"
[NX v22.6.5 Daemon Client] - 2026-04-24T21:36:40.976Z - [Messenger] Sending message type: RECORD_OUTPUTS_HASH
[...]

PR #35497, opened 2026-04-29
feat(core): add --mode and --multi-major-mode flags to nx migrate

leosvelperez:
## Current Behavior

`nx migrate <target>` processes every entry in the resolved `packageJsonUpdates`, including third-party packages outside the target's `nx.packageGroup`. There is no way to scope a run to Nx itself, no safety rail when jumping multiple major versions, and the cascade can propose downgrades for deps that have already been bumped past the historical pin.

## Expected Behavior

Adds two new flags to `nx migrate`, plus a downgrade-prevention safeguard and a bare-invocation default.

### `--mode={first-party|third-party|all}`

Scope which packages get migrated. Only valid when the target is the canonical Nx package (`nx`, or `@nx/workspace` as an era-aware alias).

- **`first-party`** — bumps Nx and the packages in its `nx.packageGroup`; third-party deps are left untouched.
- **`third-party`** — anchors at the installed Nx version and walks `--from=nx@0.0.0 --exclude-applied-migrations` internally, surfacing third-party catch-up that earlier first-party-only runs left behind.
- **`all`** (default) — everything; matches existing behavior.

Defaults to `all` outside a TTY; prompts in an interactive terminal when the target is canonical Nx. Rejects `--from`, `--exclude-applied-migrations`, and out-of-bounds `--to nx@…` (or a higher `nx@…` positional) when `mode=third-party` — those already imply going past the installed version.

### `--multi-major-mode={direct|gradual}`

Handles the case where the target jumps two or more major versions from installed.

- **Interactive (TTY):** prompts with the smallest-step recommendations — `latest in current major` (when at least a minor ahead of installed), `next major`, and `migrate directly to <target>`. The smallest available step is tagged `[recommended]`.
- **Non-interactive:** warns and proceeds with the requested target.
- **`--multi-major-mode=direct`** (or `NX_MULTI_MAJOR_MODE=direct`) — skip the prompt/warn and migrate directly to the requested target.
- **`--multi-major-mode=gradual`** (or `NX_MULTI_MAJOR_MODE=gradual`) — skip the prompt and pick the smallest recommended step automatically; re-run `nx migrate` to continue toward the originally requested target. Falls back silently to the requested target when no stepwise option is available.

### Downgrade prevention

`packageJsonUpdates` entries that would move a workspace dep backwards from its current pin are now dropped from the proposed updates. Common case: a workspace that manually bumped a dep past the version a historical `packageJsonUpdates` entry pins.

### Bare invocation

`nx migrate` with no positional now defaults to `nx@latest` instead of erroring, then runs through the mode + multi-major flow.

### DIFF
--- a/packages/nx/src/command-line/migrate/command-object.ts
+++ b/packages/nx/src/command-line/migrate/command-object.ts
@@ -84,6 +84,12 @@ function withMigrationOptions(yargs: Argv) {
       type: 'boolean',
       default: false,
     })
+    .option('mode', {
+      describe:
+        "Restrict which packages to migrate. 'first-party' processes only the target package and the packages in its nx.packageGroup; 'all' processes everything.",
+      type: 'string',
+      choices: ['first-party', 'all'],
+    })
     .check(
       ({ createCommits, commitPrefix, from, excludeAppliedMigrations }) => {
         if (!createCommits && commitPrefix !== defaultCommitPrefix) {
--- a/packages/nx/src/command-line/migrate/command-object.ts
+++ b/packages/nx/src/command-line/migrate/command-object.ts
@@ -90,6 +90,12 @@ function withMigrationOptions(yargs: Argv) {
       type: 'string',
       choices: ['first-party', 'third-party', 'all'],
     })
+    .option('acceptMultiMajorUpdate', {
+      describe:
+        'Skip the multi-major migration prompt/warning and migrate directly to the target version even when it crosses more than one major boundary. The recommended process is to update one major version at a time. Equivalent env var: NX_ACCEPT_MULTI_MAJOR_UPDATE=true.',
+      type: 'boolean',
+      default: false,
+    })
     .check(
       ({
         createCommits,
--- a/packages/nx/src/command-line/migrate/command-object.ts
+++ b/packages/nx/src/command-line/migrate/command-object.ts
@@ -90,11 +90,11 @@ function withMigrationOptions(yargs: Argv) {
       type: 'string',
       choices: ['first-party', 'third-party', 'all'],
     })
-    .option('acceptMultiMajorUpdate', {
+    .option('multiMajorMode', {
       describe:
-        'Skip the multi-major migration prompt/warning and migrate directly to the target version even when it crosses more than one major boundary. The recommended process is to update one major version at a time. Equivalent env var: NX_ACCEPT_MULTI_MAJOR_UPDATE=true.',
-      type: 'boolean',
-      default: false,
+        "Skip the multi-major migration prompt/warning and pick how to handle the jump. 'direct' migrates straight to the requested target. 'gradual' migrates to the smallest recommended step (re-run `nx migrate` to continue toward the original target). Equivalent env var: NX_MULTI_MAJOR_MODE=direct|gradual.",
+      type: 'string',
+      choices: ['direct', 'gradual'],
     })
     .check(
       ({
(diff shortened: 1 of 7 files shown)


FrozenPandaz (approved):
Approving — this is solid. A few items I'd suggest as follow-ups (none worth blocking on):

**Quick wins (one-line fixes):**
- `command-object.ts:88` — `--mode` `describe` says it defaults to `'all'`, but in an interactive Nx terminal it prompts with no default. Reword for `--help` accuracy.
- `migrate.ts:1036` — `--mode` + `--run-migrations` throws, but `--multi-major-mode` + `--run-migrations` doesn't. Add the symmetric validation.

**Tighten the safety rails (small but real):**
- `multi-major.ts:26-35, 147-156, 189-198` — three spots where a registry failure silently disables the `--multi-major-mode=gradual` safety rail (gradual quietly behaves like direct). Probably rare in the wild, but the failure mode is "the safeguard the user explicitly asked for doesn't fire and we don't tell them." A single `output.warn` on each path closes it.
- `update-filters.ts` + `migrate.ts:569-571` — downgrade filter and `--mode=first-party` drop entries with no count surfaced. A one-line summary ("Skipped N third-party packages") would make the modes legible.

**Naming (worth deciding before too many users see `--mode`):**
- Consider renaming `first-party` / `third-party` → `direct` / `external`. The `first-party` / `third-party` pair comes from legal/business vocabulary and reads slightly awkwardly in a CLI ("first party to whom?"), and the `--help` description ends up doing all the work to land the meaning. `direct` / `external` is more harmonious — both halves come from the same positional vocabulary, so the pair sits together naturally in `--help`, error messages, and prose. The conceptual mapping stays identical (`direct` = the migration target and its `packageGroup`; `external` = packages the target's migrations touch but doesn't author). Cheaper to rename now than after blog posts and docs lock the current names in.

**Cleanup (no rush):**
- Extract `export type MigrateMode` once and reuse — currently re-typed inline in 8+ sites; drift risk if a 4th mode lands.
- `version-utils.ts:60` / `migrate.ts:884, 998` — three copies of the `< 14.0.0-beta.0` era check. Two of those could route through `resolveCanonicalNxPackage`.
- Multi-major prompt options modeled as a `{name, message}[]` + `recommendedMarked` boolean rather than a discriminated union over `{ both | currentOnly | nextOnly }`. The "≥1 of current/next" precondition currently lives in a comment.
- `getInstalledLegacyNrwlWorkspaceVersion` (`installed-nx-version.ts:74-83`) bypasses the cache-pollution protection from #35444 — inconsistent with the other readers in the same file.

**Test gaps worth backfilling:**
- Reject `--to @nx/workspace@<higher>` (mirrors the existing `nx@<higher>` test at `migrate.spec.ts:2406`).
- `filterDowngradedUpdates` with `~1.2.3` ranges and peer-deps-only packages.
- Bogus `--multi-major-mode=Gradul` / capitalization fallback (no test pins this contract).
- `--mode=third-party --multi-major-mode=gradual` doesn't bypass the third-party walk.

**Architectural thought (for if you revisit):** the silent-drops issue in `--mode=first-party`/`third-party` would dissolve naturally if the cascade emitted a `{ firstParty, thirdParty }` partition rather than filtering via `shouldExcludePackage`. `--mode` then becomes "which bucket(s) you write," and the summary count falls out for free. Not a blocker, just a cleaner shape.

Strengths worth calling out: the module split (`multi-major.ts` / `update-filters.ts` / `version-utils.ts`) is clean, the era-aware `@nx/workspace` alias rationale at `migrate.ts:1889` is exactly the kind of *why* that belongs inline, `filterDowngradedUpdates` is pure + dependency-injected and very testable, and the test suite mocks TTY/CI/registry/env-var dimensions cleanly with proper save/restore.

---

<sub>Generated by `review-pr-deep` (drafted, not auto-posted). To post: edit `~/.claude/triage/prs/35497.md` then run `/review-pending-pr-reviews`.</sub>

PR #35516, opened 2026-04-30
feat(bundling)!: drop legacy typescript plugin and align rollup buildLibsFromSource default

jaysoo:
## Current Behavior

`useLegacyTypescriptPlugin` opt-in keeps the `rollup-plugin-typescript2` code path alive in `@nx/rollup` with a deprecation warning. The `withNx` plugin defaults `buildLibsFromSource` to `false` while the executor schema defaults to `true` — same option, two defaults.

## Expected Behavior

- `useLegacyTypescriptPlugin` option, schema entry, and `rollup-plugin-typescript2` dep removed; only `@rollup/plugin-typescript` remains.
- `withNx` `buildLibsFromSource` default flipped to `true` so it matches the executor.
- `update-23-0-0-remove-use-legacy-typescript-plugin` migration strips the deprecated option from existing project.json `options`/`configurations` and from `rollup.config.{cjs,mjs,js,ts}` `withNx({...})` calls so users see no behavior change after upgrade.

## Breaking Change Note

`@rollup/plugin-typescript` enforces that the TS `outDir` is inside the rollup `output.dir`. Users whose custom `rollup.config.{cjs,mjs}` overrides `output.dir` to a path outside the executor's `outputPath` will fail with `Path of Typescript compiler option 'outDir' must be located inside Rollup 'dir' option`. The legacy `rollup-plugin-typescript2` was permissive here. Affected users should align their custom config's `output.dir` with the executor's `outputPath` (or use the function form `(config) => ({ ...config, output: { ...config.output[0], ...overrides } })` to preserve `dir`). Two e2e cases that exercised this legacy pattern were dropped from `rollup-legacy.test.ts`.

## Related Issue(s)

Fixes NXC-4157


packages/rollup/src/migrations/update-23-0-0/remove-use-legacy-typescript-plugin.ts
```diff
@@ -0,0 +1,109 @@
...
+              delete config.useLegacyTypescriptPlugin;
+              projectJsonChanged = true;
+            }
+          }
+        }
+      }
+    }
+
+    if (projectJsonChanged) {
+      updateProjectConfiguration(tree, projectName, projectConfig);
```
leosvelperez, line 65: We also need to migrate potential target defaults in `nx.json`, which can be keyed by target name or executor name.

packages/rollup/src/migrations/update-23-0-0/remove-use-legacy-typescript-plugin.ts
```diff
@@ -0,0 +1,109 @@
...
+function stripFromRollupConfig(tree: Tree, filePath: string) {
+  const original = tree.read(filePath, 'utf-8');
+  if (!original?.includes(PROPERTY_NAME)) return;
+
+  // Filter by the property NAME (not any descendant Identifier) so we don't
+  // strip `{ alias: useLegacyTypescriptPlugin }` where the identifier appears as a value.
+  const matches = (
+    query(ast(original), 'PropertyAssignment') as PropertyAssignment[]
+  ).filter((p) => p.name && (p.name as any).text === PROPERTY_NAME);
+  if (matches.length === 0) return;
```
leosvelperez, line 131: Should we also handle stripping `ShorthandPropertyAssignment` (e.g. `{ alias }`)? We could at least strip it and maybe not worry about leaving a potentially unused variable declaration behind.

leosvelperez:
We also need to add the companion doc for the migration.

packages/rollup/src/migrations/update-23-0-0/remove-use-legacy-typescript-plugin.ts
jaysoo: Done, added `describe('nx.json targetDefaults', ...)` test

packages/rollup/src/migrations/update-23-0-0/remove-use-legacy-typescript-plugin.ts
jaysoo: Done

leosvelperez (approved):

PR #35517, opened 2026-04-30
feat(vite)!: remove vitest support in favor of @nx/vitest

jaysoo:
## Current Behavior

`@nx/vite` carries a parallel vitest surface (version constants, `:test` executor, `vitest` generator, plugin test inference, init-time install) that duplicates `@nx/vitest`.

## Expected Behavior

`@nx/vite` owns vite. `@nx/vitest` owns vitest. Vitest surface removed from `@nx/vite`; consumers route through `@nx/vitest`.

v23 migration installs `@nx/vitest`, swaps the executor, and registers `@nx/vitest` plugin alongside default-config `@nx/vite/plugin` so test inference is preserved.

## Related Issue(s)

Closes NXC-4158


packages/vite/src/migrations/update-23-0-0/ensure-vitest-package-migration.ts
```diff
@@ -0,0 +1,207 @@
...
+
+function installVitestPackageIfNeeded(tree: Tree): GeneratorCallback {
+  const packageJson = readJson(tree, 'package.json');
+  const hasVitest =
+    packageJson.dependencies?.['@nx/vitest'] ||
+    packageJson.devDependencies?.['@nx/vitest'];
+  if (hasVitest) {
+    return () => {};
+  }
+  return addDependenciesToPackageJson(tree, {}, { '@nx/vitest': nxVersion });
```
leosvelperez: Wouldn't this always install `@nx/vitest`, even when `vitest` might not be used? Probably an edge case, but this is currently always installing the package without knowing if it's needed.

packages/vite/src/migrations/update-23-0-0/ensure-vitest-package-migration.ts
```diff
@@ -0,0 +1,207 @@
...
+      } else {
+        delete updatedVitePlugin.options;
+      }
+      newPlugins.push(updatedVitePlugin);
+    } else {
+      newPlugins.push(plugin);
+    }
+  }
+
+  if (!hasVitestPlugin && vitestPluginsToAdd.length > 0) {
```
leosvelperez: If it has a vitest plugin already registered, should we blindly discard migrating the existing `@nx/vite/plugin` registration options? Both plugins could be registered and there are some scenarios where the options from the `@nx/vite/plugin` registration would still apply and be inferred.

leosvelperez (commented):
- We're missing the companion doc for the migration.
- Should this section in the docs be removed: https://nx.dev/docs/technologies/test-tools/vitest/guides/migrating-from-nx-vite#backward-compatibility?

jaysoo:
> * We're missing the companion doc for the migration.
> 
>     * Should this section in the docs be removed: https://nx.dev/docs/technologies/test-tools/vitest/guides/migrating-from-nx-vite#backward-compatibility?

I think we can keep the migration doc and say that `@nx/vite` vitest features were removed in v23.

packages/vite/src/migrations/update-23-0-0/ensure-vitest-package-migration.md
leosvelperez, line 1: Is this location for migration docs something we pick up or handle? I thought they needed to be side-by-side with the migration generator.

packages/vite/src/migrations/update-23-0-0/ensure-vitest-package-migration.ts
FrozenPandaz: Isn't that what the above line is for? Oh that seems like it was added later

packages/vite/src/migrations/update-23-0-0/ensure-vitest-package-migration.ts
```diff
@@ -0,0 +1,254 @@
...
+  );
+  if (vitePluginRegistrations.length === 0) {
+    return;
+  }
+
+  for (const vitePlugin of vitePluginRegistrations) {
+    if (typeof vitePlugin === 'string') {
+      nxJson.plugins.push({ plugin: '@nx/vitest' });
+      continue;
+    }
```
leosvelperez: This would add a `@nx/vitest` registration even in cases where test tasks are not being inferred (e.g. `vitest` not installed, or not `test` configuration). There's a disconnect between this function and `installVitestPackageIfNeeded`, which checks if `vitest` is installed.

leosvelperez (approved):

FrozenPandaz (approved):

PR #35531, opened 2026-05-01
feat(testing)!: deprecate the @nx/cypress:cypress executor

FrozenPandaz:
## Current Behavior

The `@nx/cypress:cypress` executor runs silently with no signal that it's slated for removal. Users running `nx g @nx/cypress:configuration` or `nx g @nx/cypress:component-configuration` on workspaces that don't have `@nx/cypress/plugin` registered get executor-based `e2e` (or `component-test`) targets scaffolded into their `project.json` with no warning.

The `@nx/cypress:cypress` executor is a thin wrapper over the Cypress CLI. The `@nx/cypress/plugin` inferred plugin produces equivalent `e2e` and `component-test` targets directly from `cypress.config.*`, and `nx g @nx/cypress:convert-to-inferred` is the one-shot migration tool to move existing executor-based targets onto the inferred plugin.

## Expected Behavior

The `@nx/cypress:cypress` executor is marked as deprecated for removal in Nx v24. Users see the deprecation message on three surfaces:

- **Runtime warning** at the top of `cypress.impl.ts` — fires every executor invocation, no throttling.
- **Scaffold-time warning** in `configuration.ts` and `component-configuration.ts` — fires when the generators are about to emit an executor-based target (i.e., when `@nx/cypress/plugin` isn't in `nx.json`).
- **Schema-root `x-deprecated`** on `executors/cypress/schema.json` — surfaces in IDE / docs metadata.

All three messages link to the general convert-to-inferred guide at https://nx.dev/docs/guides/tasks--caching/convert-to-inferred.

The `@nx/cypress` package itself, the `@nx/cypress/plugin` inferred plugin, the `configuration` and `component-configuration` scaffolding generators, and the `convert-to-inferred` migration tool all **remain supported**. Only the `:cypress` executor is deprecated. Removal is targeted for v24; this PR is warnings only, no behavior removal.

The PR title uses `feat!` so the deprecation surfaces in the v23 changelog / release notes; individual commits stay `chore` / `docs` / `fix` for clean per-commit history.

### Internal usage / dogfood note

`graph/client-e2e/project.json` uses `@nx/cypress:cypress` and is **intentionally not migrated** in this PR. Hitting the deprecation warning on every internal CI run is the team's continuous dogfood signal that the migration tool, the wording, and the suggested path actually work. If it's painful for us, it's painful for users; that pressure should drive iteration on `convert-to-inferred`. The internal project will be migrated alongside the v24 removal PR (or earlier if the warning becomes legitimately disruptive).

### Surfaces explicitly NOT touched

- Framework-specific cypress component-testing generators (angular / react / next / remix) — unchanged.
- The Angular `ng-add` e2e migrator — unchanged.
- `migrations.json` — no migration entry. Runtime warnings + docs are enough.
- `npm deprecate` — not run; the package isn't deprecated, only the executor.
- Cypress-specific docs page — no per-plugin convert-to-inferred page exists for any other plugin (Detox, Playwright, etc.). Pointing the warnings at the general guide keeps that consistent and avoids one more page to maintain.

### Test plan

- [x] `pnpm nx build cypress` passes.
- [ ] `pnpm nx test cypress` — verify no regressions from the new warning calls.
- [ ] Verify the runtime warning fires when running `nx e2e <project>` on a project that uses the executor (e.g., `graph/client-e2e`).
- [ ] Verify the scaffold-time warning fires on `nx g @nx/cypress:configuration` / `:component-configuration` in a workspace without `@nx/cypress/plugin`.
- [ ] Verify `convert-to-inferred` still works end-to-end as the migration path.

## Related Issue(s)

Fixes NXC-4264

This PR follows the canonical executor-deprecation pattern documented in Linear NXC-4422, established by the `@nx/detox` executor deprecation (NXC-4272 / nrwl/nx#35529).

Related Linear context:

- **NXC-4422** — Pattern: executor deprecation (canonical reference).
- **NXC-4272** / **nrwl/nx#35529** — `@nx/detox` executor deprecation, the reference implementation.
- **NXC-4420** — Nx core: surface schema-root `x-deprecated` on executors at runtime. Once landed, the explicit `logger.warn` in this PR becomes redundant.


AgentEnder (approved):
Should we be deprecating the `addPlugins` nx json flag too?

PR #35539, opened 2026-05-01
fix(js): use TypeScript readConfigFile instead of JSON.parse in resolvePathsBaseUrl

### LINKED ISSUE #35537: Regression (22.7.0): resolvePathsBaseUrl uses JSON.parse — breaks on tsconfig files with // comments (JSONC)
## Current Behavior

After upgrading from Nx 22.6.5 to 22.7.1, all Angular buildable library builds fail with `TS2307: Cannot find module'@workspace/lib-name'` for workspace dependencies. The TypeScript path mappings are resolved relative to the **project directory** instead of the **workspace root**.

## Expected Behavior

Buildable library builds should resolve workspace dependencies correctly, as they did in 22.6.5 and earlier.

## Root Cause

PR #34965 introduced `resolvePathsBaseUrl()` in `packages/js/src/utils/typescript/ts-config.ts`. This function walks the tsconfig `extends` chain using `JSON.parse(readFileSync(...))` to find where `paths` is defined and what `baseUrl` applies.

**Problem**: `JSON.parse` cannot handle `//` comments (JSONC), which are extremely common in tsconfig files. When parsing fails, the `catch` block silently skips the file, **breaking the extends chain walk**. The function never reaches `tsconfig.base.json` (where `paths` and `baseUrl` are defined), so it falls back to the project's own directory as the resolution base.

This produces incorrect path mappings like:
```
"@workspace/data-layer": ["{workspaceRoot}/libs/my-lib/dist/libs/data-layer"]
```
instead of:
```
"@workspace/data-layer": ["{workspaceRoot}/dist/libs/data-layer"]
```

## Reproduction

Any Nx workspace with:
1. Multiple buildable Angular libraries with cross-dependencies
2. A shared intermediate tsconfig (e.g., `libs/tsconfig.json`) in the `extends` chain that contains a `//` comment

Example `libs/tsconfig.json`:
```json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    // TODO: fix this later
    "strictPropertyInitialization": false
  }
}
```

Each lib's `tsconfig.json` extends `"../tsconfig.json"` → which extends `"../tsconfig.json"` (the shared one with the comment) → which extends `"tsconfig.base.json"` (where `paths` + `baseUrl` live).

The `JSON.parse` failure on the commented file breaks the chain, so `resolvePathsBaseUrl` returns the lib directory instead of the workspace root.

## Suggested Fix

Use TypeScript's `readConfigFile` (which handles JSONC) instead of `JSON.parse`:

```typescript
// packages/js/src/utils/typescript/ts-config.ts — resolvePathsBaseUrl()

// Before:
const raw = JSON.parse(readFileSync(absolute, 'utf-8'));

// After:
const ts = require('typescript');
const { config: raw } = ts.readConfigFile(absolute, ts.sys.readFile);
```

This is the same approach already used by the `readTsConfig` function in the same file. The same fix should be applied to the duplicated version in `register.ts` (mentioned in PR #34965).

## Environment

- Nx: 22.7.1
- @nx/angular: 22.7.1
- Angular: 21+
- Node: 22.x

## Workaround

Remove all `//` comments from tsconfig files in the extends chain between the project and the root `tsconfig.base.json`.

### LINKED ISSUE #35824: Buildable libs fail with TS2307 in 22.7.x: resolvePathsBaseUrl uses strict JSON.parse on JSONC tsconfigs, breaking the extends chain and mis-resolving dependency output paths
## Current Behavior

Since **22.7.0**, building a buildable Angular/TS library (`@nx/angular:ng-packagr-lite`, also affects `@nx/js:tsc`) that depends on another buildable workspace library fails with:

```
error TS2307: Cannot find module '@scope/dep' or its corresponding type declarations.
```

even though:
- the project graph edge to the dependency is correct,
- the dependency builds successfully into `dist/`,
- the `@scope/*` path alias exists in `tsconfig.base.json` and points to a real source file.

This regressed in **22.7.0** and is still present in **22.7.5** (current latest). `22.6.5` works.

## Root Cause

The buildable-libs path remapping in `@nx/js` is producing wrong absolute paths for buildable dependency outputs.

`createTmpTsConfig` → `readTsConfigWithRemappedPaths` (`packages/js/src/utils/buildable-libs-utils.ts`) now resolves each `paths` entry to an absolute path using `resolvePathsBaseUrl(originalTsConfig)`, introduced in 22.7.0 to drop reliance on `baseUrl` ahead of its removal in TS 7:

```ts
const pathsBase = resolvePathsBaseUrl(normalizedTsConfig);
for (const key of Object.keys(paths)) {
  paths[key] = paths[key].map((p) =>
    isAbsolute(p) ? p : resolve(pathsBase, p.replace(/^\.\//, '')).replace(/\\/g, '/')
  );
}
```

`resolvePathsBaseUrl` (`packages/js/src/utils/typescript/ts-config.ts`) walks the tsconfig `extends` chain to find the applicable `baseUrl`, **reading each file with strict `JSON.parse`**:

```ts
const raw = JSON.parse(readFileSync(absolute, 'utf-8'));   // throws on JSONC
// ...
catch { /* skip unreadable files */ }                       // swallows the throw, drops the link
```

Nx's own generated tsconfigs are **JSONC** (comments and trailing commas). A standard generated `tsconfig.lib.json` contains:

```jsonc
"compilerOptions": {
  "types": [],   // <- trailing comma
},
```

`JSON.parse` throws on that trailing comma, the empty `catch {}` swallows it, and that link of the `extends` chain is silently dropped. The walk therefore never reaches `tsconfig.base.json` (where `paths` and `baseUrl: "."` are defined), so no `baseUrl` is found and the function hits its fallback:

```ts
return pathsIndex >= 0 ? chain[pathsIndex].dir : dirname(resolve(tsconfigPath));
```

→ it returns the **project directory** instead of the **workspace root**.

`updatePaths` emits buildable-dependency outputs as workspace-root-relative (`dist/libs/shared/dep`). Resolving those against the project dir produces a path that does not exist:

```
got:      <root>/libs/shared/data-access/consumer/dist/libs/shared/dep   (does not exist)
expected: <root>/dist/libs/shared/dep
```

Confirmed via `NX_VERBOSE_LOGGING_PATH_MAPPINGS=true`, which prints the alias mapped to the wrong project-relative absolute path.

## Expected Behavior

`resolvePathsBaseUrl` should read tsconfig files as JSONC (comments + trailing commas), consistent with how the rest of Nx and the TS compiler read tsconfigs (e.g. `ts.readConfigFile` / `@nx/devkit`'s `parseJson` / `stripJsonComments`). The `extends` chain should resolve to `tsconfig.base.json`, find `baseUrl`, and produce workspace-root-relative absolute paths — matching 22.6.x behavior.

## Suggested Fix

Replace the strict `JSON.parse(readFileSync(...))` in `resolvePathsBaseUrl` with a JSONC-tolerant read (the same mechanism already used elsewhere in Nx to read tsconfigs). The silent `catch {}` is also masking the failure and should not be relied upon for control flow.

## GitHub Repo

No public minimal repo yet — can provide on request. Reproduces in any workspace matching the steps below.

## Steps to Reproduce

1. Nx workspace with two buildable libs where lib A imports lib B via a `@scope/*` path alias (`@nx/angular:ng-packagr-lite` or `@nx/js:tsc`).
2. Keep the default trailing comma / comments in the generated `tsconfig.lib.json` (the Nx default).
3. `nx build A --configuration=production --skip-nx-cache` → `error TS2307: Cannot find module '@scope/b'`.
4. Downgrade `nx` + `@nx/*` to `22.6.5` and rebuild → succeeds.

## Version boundary

| `@nx/js` version | `resolvePathsBaseUrl` present | buildable lib build |
|---|---|---|
| 22.6.5 | no | works |
| 22.7.0 | **yes (introduced)** | TS2307 |
| 22.7.5 (latest) | yes | TS2307 |

## Nx Report

```
Nx: 22.7.0 – 22.7.5 (regression from 22.6.5)
@nx/angular: 22.7.5 (ng-packagr-lite executor)
Angular: 21.1.1, ng-packagr: 21.1.0
TypeScript: 5.9.2
Node: 24.16.0, pnpm: 11.4.0
OS: macOS (darwin arm64)
Package manager: pnpm (non-workspace; @scope/* resolved purely via tsconfig.base.json paths)
```

## Failure Logs

```
error TS2307: Cannot find module '@scope/dep' or its corresponding type declarations.

import { Foo } from '@scope/dep';
                     ~~~~~~~~~~~~
```

aqeelat:
## Current Behavior

`resolvePathsBaseUrl` in `packages/js/src/utils/typescript/ts-config.ts` walks the tsconfig `extends` chain using `JSON.parse(readFileSync(...))` to find where `paths` is defined. `JSON.parse` cannot handle `//` or `/* */` comments (JSONC), which are extremely common in tsconfig files. When parsing fails, the `catch` block silently skips the file, breaking the extends chain walk. The function never reaches `tsconfig.base.json` (where `paths` and `baseUrl` are defined), so it falls back to the project's own directory as the resolution base, producing incorrect path mappings like `{workspaceRoot}/libs/my-lib/dist/libs/data-layer` instead of `{workspaceRoot}/dist/libs/data-layer`.

This is a regression introduced in v22.7.0 by PR #34965.

## Expected Behavior

Buildable library builds should resolve workspace dependencies correctly, even when tsconfig files in the extends chain contain JSONC comments (`//` or `/* */`). The fix uses TypeScript's `readConfigFile` (which handles JSONC) — the same approach already used by `readTsConfig` in the same file.

## Related Issue(s)

Fixes #35537
Fixes #35824

### DIFF
--- a/packages/js/src/utils/typescript/ts-config.ts
+++ b/packages/js/src/utils/typescript/ts-config.ts
@@ -1,5 +1,5 @@
 import { offsetFromRoot, Tree, updateJson, workspaceRoot } from '@nx/devkit';
-import { existsSync, readFileSync } from 'fs';
+import { existsSync } from 'fs';
 import { dirname, isAbsolute, join, resolve } from 'path';
 import type * as ts from 'typescript';
 import { ensureTypescript } from './ensure-typescript';
@@ -120,13 +120,20 @@ function ensureRelativePath(p: string): string {
  * tsconfig that defines `paths`.
  */
 export function resolvePathsBaseUrl(tsconfigPath: string): string {
+  if (!tsModule) {
+    tsModule = require('typescript');
+  }
+
   const chain: { dir: string; raw: any }[] = [];
   const queue: string[] = [tsconfigPath];
   while (queue.length > 0) {
     const absolute = resolve(queue.shift()!);
     const dir = dirname(absolute);
     try {
-      const raw = JSON.parse(readFileSync(absolute, 'utf-8'));
+      const { config: raw } = tsModule.readConfigFile(
+        absolute,
+        tsModule.sys.readFile
+      );
       chain.push({ dir, raw });
       const exts: string[] = raw.extends
         ? Array.isArray(raw.extends)
--- a/packages/js/src/utils/typescript/ts-config.ts
+++ b/packages/js/src/utils/typescript/ts-config.ts
@@ -1,4 +1,10 @@
-import { offsetFromRoot, Tree, updateJson, workspaceRoot } from '@nx/devkit';
+import {
+  offsetFromRoot,
+  readJsonFile,
+  Tree,
+  updateJson,
+  workspaceRoot,
+} from '@nx/devkit';
 import { existsSync } from 'fs';
 import { dirname, isAbsolute, join, resolve } from 'path';
 import type * as ts from 'typescript';
@@ -120,20 +126,13 @@ function ensureRelativePath(p: string): string {
  * tsconfig that defines `paths`.
  */
 export function resolvePathsBaseUrl(tsconfigPath: string): string {
-  if (!tsModule) {
-    tsModule = require('typescript');
-  }
-
   const chain: { dir: string; raw: any }[] = [];
   const queue: string[] = [tsconfigPath];
   while (queue.length > 0) {
     const absolute = resolve(queue.shift()!);
     const dir = dirname(absolute);
     try {
-      const { config: raw } = tsModule.readConfigFile(
-        absolute,
-        tsModule.sys.readFile
-      );
+      const raw = readJsonFile(absolute);
       chain.push({ dir, raw });
       const exts: string[] = raw.extends
         ? Array.isArray(raw.extends)
--- a/packages/js/src/utils/typescript/ts-config.spec.ts
+++ b/packages/js/src/utils/typescript/ts-config.spec.ts
@@ -206,6 +206,38 @@ describe('resolvePathsBaseUrl', () => {
     );
   });
 
+  it('should handle tsconfig files with // comments (JSONC) in extends chain', () => {
+    tempFs.createFileSync(
+      'tsconfig.base.json',
+      JSON.stringify({
+        compilerOptions: {
+          baseUrl: '.',
+          paths: { '@lib/*': ['libs/*/src/index.ts'] },
+        },
+      })
+    );
+    tempFs.createFileSync(
+      'tsconfig.intermediate.json',
+      `{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    // TODO: fix this later
+    "strictPropertyInitialization": false
+  }
+}`
+    );
+    tempFs.createFileSync(
+      'project/tsconfig.json',
+      JSON.stringify({
+        extends: '../tsconfig.intermediate.json',
+      })
+    );
+
+    expect(
+      resolvePathsBaseUrl(join(tempFs.tempDir, 'project', 'tsconfig.json'))
+    ).toBe(tempFs.tempDir);
+  });
+
   it('should return tsconfig directory when no paths or baseUrl exist', () => {
     tempFs.createFileSync(
       'tsconfig.json',


packages/js/src/utils/typescript/ts-config.ts
```diff
@@ -120,13 +120,20 @@ function ensureRelativePath(p: string): string {
...
   const queue: string[] = [tsconfigPath];
   while (queue.length > 0) {
     const absolute = resolve(queue.shift()!);
     const dir = dirname(absolute);
     try {
-      const raw = JSON.parse(readFileSync(absolute, 'utf-8'));
+      const { config: raw } = tsModule.readConfigFile(
+        absolute,
+        tsModule.sys.readFile
+      );
```
leosvelperez: Let's avoid loading `typescript`, which is a heavy operation and what this code was trying to avoid in the first place. Some prior call places might already load it, but let's reduce the risk of doing it unnecessarily. We can use the `readJsonFile` helper from `@nx/devkit` that supports comments: 
```suggestion
      const raw = readJsonFile(absolute);
```

leosvelperez (changes requested):
Thanks for the contribution! Left one request to avoid loading `typescript` where it's not strictly needed for perf reasons.

vietd24-philips:
@aqeelat any updates here?

leosvelperez (approved):
Thank you!

PR #35540, opened 2026-05-01
chore(core): unify Task into a single Rust struct

FrozenPandaz:
## Current Behavior

`Task` is defined twice and the two definitions have drifted:

- TS canonical: `packages/nx/src/config/task-graph.ts` — 13 fields including `overrides: any`, `hash`, `hashDetails`, `cache`, `parallelism`.
- Rust / NAPI: `packages/nx/src/native/tasks/types.rs` — 7 fields, missing all the dynamic / options-shaped ones.

Plus a swarm of variant types (`TaskResult` defined in both Rust and TS, `TaskTarget`, etc.). Anything that wants to consume a Task in Rust either gets the lossy 7-field copy or has to translate at the boundary. The historical blocker for unifying was `Task.overrides` — a freeform POJO of CLI args that had no obvious Rust representation.

## Expected Behavior

One canonical `Task` struct, defined in Rust, exposed to TS via NAPI:

- The Rust struct in `packages/nx/src/native/tasks/types.rs` is now the superset (all 13 fields).
- `overrides` is modeled as `serde_json::Value`, surfaced in TS as `Record<string, unknown>`. The `serde-json` napi feature is enabled to make this round-trip work.
- `TaskResult.status` gets the literal-union TS type via `#[napi(ts_type)]` so we can collapse the duplicate `TaskResult` in `tasks-runner/life-cycle.ts` into a re-export from the native bindings.
- `packages/nx/src/config/task-graph.ts` now re-exports `Task` / `TaskGraph` / `TaskTarget` / `TaskHashDetails` from `../native`. Every existing import keeps working.
- A small ergonomics layer for Rust callers: `Task::new(project, target)` (id auto-derived as `{project}:{target}`, matching the formula `create-task-graph.ts` already uses) plus chained `with_*` builders for the optional fields. `with_configuration` recomputes the id with the configuration suffix.

Behavior is unchanged on every existing TS code path. The legacy `overrides: Object` parameter annotations in `create-task-graph.ts` and `hash-plan-inspector.ts` are tightened to `Record<string, unknown>`. Direct `overrides` reads at `combineOptionsForExecutor` call sites and `__overrides_unparsed__` accesses get tight casts that preserve prior runtime semantics, including the throw-on-missing behavior.

This unblocks porting the task scheduler (and eventually the orchestrator) to Rust — without one shared struct, every port has to redo the boundary translation.

### Verification

- `cargo test --lib` — 352 passed, 0 failed.
- `pnpm nx build nx` — clean.
- `pnpm nx run-many -t build` — all packages build (the two `examples-angular-rspack-mf-*` failures predate this branch).
- `pnpm nx run nx:lint` — clean (only a pre-existing skipped-test warning).
- `pnpm nx run nx:test --testPathPatterns='create-task-graph|tasks-runner|hasher'` — all suites pass, including `should interpolate overrides`.

### Out of scope

- Porting orchestrator / scheduler to Rust (this is the foundation for that next).
- Collapsing `HashedTask` (history DB row) and `TaskRun` (history record) — they are storage shapes, not Tasks.

## Related Issue(s)

Foundation for the eventual Rust port of the task scheduler and orchestrator.


packages/nx/src/native/tui/components/snapshots/nx__native__tui__components__tasks_list__tests__cloud_message_two_line_layout.snap
```diff
@@ -5,9 +5,9 @@ expression: terminal.backend()
 "                                             "
 " NX    Completed Test Tasks  Cache   Duration"
 "                                             "
-"> ✔    task1                     -        ..."
-"  ✔    task2                     -        ..."
-"  ✔    task3                     -        ..."
+"> ✔    app1:build                -        ..."
+"  ✔    app1:test                 -        ..."
+"  ✔    app2:lint                 -        ..."
```
AgentEnder, line 10: Suspicious. Why did these change?

AgentEnder (approved):

PR #35541, opened 2026-05-02
feat(devkit): migrate `@nx/devkit/src/...` deep imports

FrozenPandaz:
## Current Behavior

After #34946, `@nx/devkit` ships a strict `exports` map. Deep imports like `@nx/devkit/src/utils/...` and `@nx/devkit/src/generators/...` are no longer reachable through Node module resolution. Workspaces upgrading to `23.x` that reference those paths break with module-resolution errors at runtime / type-check time.

## Expected Behavior

`nx migrate` runs an automated rewrite that covers the realistic shapes of these deep imports.

The migration walks every `.ts`/`.tsx`/`.cts`/`.mts` file in the workspace and:

1. **Buckets named imports by symbol.** Each `@nx/devkit/src/...` import declaration is parsed via the TypeScript compiler API (lazy-loaded with `ensurePackage`). Specifiers in the `internal.ts` re-export list go to `@nx/devkit/internal`; all others go to `@nx/devkit`. Mixed imports split into two declarations.
2. **Falls back for non-named shapes.** Default imports, namespace imports, side-effect imports, `require(...)` calls, and dynamic `import(...)` get the specifier swapped to `@nx/devkit/internal` (the safe default — it re-exports every previously deep-importable symbol).
3. **Collapses duplicate imports.** A second AST pass groups `import { ... } from '@nx/devkit'` and `import { ... } from '@nx/devkit/internal'` declarations by `(specifier, isTypeOnly)`, merging each 2+ group into one declaration with deduplicated specifiers. This handles both the duplicates the rewrite produced and any pre-existing devkit imports the user already had.

Edits are stacked via `applyChangesToString` (devkit's offset-tracking text-mutation helper), then `formatFiles` normalizes formatting.

### Example

Before:
```ts
import { Tree } from '@nx/devkit';
import { dasherize, names } from '@nx/devkit/src/utils/string-utils';
import { addPlugin } from '@nx/devkit/src/utils/add-plugin';
```

After:
```ts
import { Tree, names } from '@nx/devkit';
import { dasherize, addPlugin } from '@nx/devkit/internal';
```

### Tests

29 unit tests cover: single-bucket and mixed-bucket rewrites, `as` aliases, `import type` and inline `type` modifiers, multi-line imports, side-effect / default / namespace fallbacks, `require()` and dynamic `import()` fallbacks, quote-style preservation, pre-existing-import merge, public/internal independence, value-vs-type segregation, specifier deduplication, and a sanity test that every name in `DEVKIT_INTERNAL_SYMBOLS` is bucketed as internal.

A user-facing `update-deep-imports.md` lives next to the implementation; the astro-docs build picks it up via the existing `packages/*/src/migrations/**/*.md` input glob.

## Related Issue(s)

Follow-up to #34946.

### DIFF
--- a/packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
+++ b/packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
@@ -1,19 +1,18 @@
-import {
-  applyChangesToString,
-  ChangeType,
-  ensurePackage,
-  formatFiles,
-  logger,
-  type StringChange,
-  type Tree,
-  visitNotIgnoredFiles,
-} from '@nx/devkit';
+import { logger, type Tree } from 'nx/src/devkit-exports';
 import type {
   ImportDeclaration,
   ImportSpecifier,
   NamedImports,
   SourceFile,
 } from 'typescript';
+import { formatFiles } from '../../generators/format-files';
+import { visitNotIgnoredFiles } from '../../generators/visit-not-ignored-files';
+import { ensurePackage } from '../../utils/package-json';
+import {
+  applyChangesToString,
+  ChangeType,
+  type StringChange,
+} from '../../utils/string-change';
 import { typescriptVersion } from '../../utils/versions';
 
 const TS_EXTENSIONS = ['.ts', '.tsx', '.cts', '.mts'] as const;
@@ -74,7 +73,9 @@ const FALLBACK_RE = /(['"])@nx\/devkit\/src\/[^'"\n]+?\1/g;
 
 let ts: typeof import('typescript') | undefined;
 
-export default async function updateDevkitDeepImports(tree: Tree) {
+export default async function updateDevkitDeepImports(
+  tree: Tree
+): Promise<void> {
   let touchedCount = 0;
 
   visitNotIgnoredFiles(tree, '.', (filePath) => {
--- a/packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
+++ b/packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
@@ -69,6 +69,8 @@ export const DEVKIT_INTERNAL_SYMBOLS: ReadonlySet<string> = new Set([
   'dasherize',
 ]);
 
+const FALLBACK_RE = /(['"])@nx\/devkit\/src\/[^'"\n]+?\1/g;
+
 let ts: typeof import('typescript') | undefined;
 
 export default async function updateDevkitDeepImports(
@@ -118,7 +120,6 @@ export function rewriteDevkitDeepImports(source: string): string {
     if (!stmt.moduleSpecifier.text.startsWith(DEEP_IMPORT_PREFIX)) continue;
 
     const replacement = buildReplacement(stmt, sourceFile);
-    if (replacement === null) continue;
     changes.push(
       {
         type: ChangeType.Delete,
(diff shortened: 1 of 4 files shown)


packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
```diff
@@ -0,0 +1,343 @@
...
+  }
+
+  let updated =
+    changes.length > 0 ? applyChangesToString(source, changes) : source;
+
+  // Fallback: any remaining `@nx/devkit/src/...` specifiers (default imports,
+  // namespace imports, side-effect imports, dynamic `import(...)`, `require(...)`
+  // calls, etc.) get rewritten to `/internal`. We can't bucket by symbol for
+  // those forms, so `/internal` is the safe default since it re-exports every
+  // symbol that was deep-importable.
```
AgentEnder, line 144: Is it true that `internal` is **_every_** symbol that was previously available?

packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
```diff
@@ -0,0 +1,343 @@
...
+const TS_EXTENSIONS = ['.ts', '.tsx', '.cts', '.mts'] as const;
+const DEEP_IMPORT_PREFIX = '@nx/devkit/src/';
+const PUBLIC_SPECIFIER = '@nx/devkit';
+const INTERNAL_SPECIFIER = '@nx/devkit/internal';
+
+// Names re-exported from `@nx/devkit/internal` (see packages/devkit/internal.ts
+// at the time this migration was authored). Anything imported from a
+// `@nx/devkit/src/...` path whose name is NOT in this set is assumed to be
+// part of the stable public `@nx/devkit` API.
+export const DEVKIT_INTERNAL_SYMBOLS: ReadonlySet<string> = new Set([
```
AgentEnder, line 27: Do you think it may be better to read these imports w/ the ts APIs since we are already bringing them in? I'm back-and-forth on it, as it may be easier to just hardcode like this and it shouldn't shift too much.

AgentEnder (approved):
Seems mostly fine, left a few comments but nothing blocking

packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
FrozenPandaz: Yeah I think hard coding them is fine

packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
FrozenPandaz: Good catch — it's not. `internal.ts` covers every symbol that was *previously deep-importable AND that we wanted to keep around as semi-private*, but a deep-importable name that ended up in the stable public API (or that we removed entirely) wouldn't be there. Pointing default / namespace / `require()` / dynamic-`import()` shapes at `/internal` could silently land them on a broken path.

Dropped the fallback in df70f84587. Now only plain named imports are rewritten (those are the ones we can bucket per-symbol via `DEVKIT_INTERNAL_SYMBOLS`); everything else is left untouched, and the migration's user-facing doc calls them out as a manual step.

FrozenPandaz: Walked this back in 4072ea2e06. You're right that `/internal` isn't literally every symbol — but dropping the fallback meant any non-named deep import (default / namespace / side-effect / `require()` / dynamic `import()`) left a `@nx/devkit/src/...` specifier the new exports map rejects, so workspaces with any of those shapes wouldn't compile post-migration.

Restored the fallback as a best-guess rewrite to `/internal`, since that's where the bulk of previously deep-importable symbols actually live. If a user's symbol turns out to be public API instead, they get a clear "no exported member" error pointing at the new entry point — which is a much better signal than an unresolvable `@nx/devkit/src/...` path. Tightened the migration doc to call it a best guess rather than a guarantee, with instructions for the manual fix.

PR #35550, opened 2026-05-03
fix(core): use workspace root for package manager detection in script targets

FrozenPandaz:
## Current Behavior

`readTargetsFromPackageJson` (in `packages/nx/src/utils/package-json.ts`) receives a `workspaceRoot` argument but never passes it to package manager detection:

```ts
for (const script of includedScripts) {
  packageManagerCommand ??= getPackageManagerCommand(); // ← no workspaceRoot
  res[script] = buildTargetFromScript(script, scripts, packageManagerCommand);
}
```

Two consequences:

1. **Wrong package manager** — `detectPackageManager()` defaults `dir = ''`, so the lockfile probe runs in the CWD, not the workspace. When that finds nothing it falls back to `npm_config_user_agent`, so the inferred `runCommand` (`npm run X` vs `pnpm run X` vs `yarn X`) on script targets ends up depending on whoever invoked the nx process rather than on the workspace's actual lockfile.

2. **Module-level cache** — `let packageManagerCommand` (cleared with `??=`) memoizes the first detection result across all subsequent calls in the process. So even if the first call had the right `workspaceRoot`, every later call inherits that detection regardless of *its* `workspaceRoot`. This is also why `packages/nx/src/plugins/package-json/create-nodes.spec.ts` had four pre-existing snapshot failures locally (`pnpm run …` instead of the expected `npm run …`) — the first test in any process locked detection to the host's PM.

This is a follow-up to #35116, which moved package manager detection into the `createNodes` callback for the inferred plugins but missed this code path.

## Expected Behavior

- Drop the module-level cache.
- Thread `workspaceRoot` into both `detectPackageManager` and `getPackageManagerCommand`, so the lockfile probe runs in the right directory.

```ts
if (includedScripts.length > 0) {
  const packageManagerCommand = getPackageManagerCommand(
    detectPackageManager(workspaceRoot),
    workspaceRoot
  );
  for (const script of includedScripts) {
    res[script] = buildTargetFromScript(script, scripts, packageManagerCommand);
  }
}
```

The `packages/nx/src/plugins/package-json/create-nodes.spec.ts` fixture now seeds `package-lock.json` into memfs in a `beforeEach`, matching the pattern #35116 established for plugin specs. Without the lockfile the detector still falls back to the env var; with it, every test deterministically picks `npm`, matching the existing snapshots.

## Verification

Before this PR: 4 failures in `packages/nx/src/plugins/package-json/create-nodes.spec.ts`:

```
✕ should build projects from package.json files
✕ should store js package metadata
✕ should add a script target if the sibling project.json file does not exist
✕ should add a script target if the sibling project.json exists but does not have a conflicting target

Tests: 4 failed, 7 passed, 11 total
```

After this PR:

```
Tests: 11 passed, 11 total
```

## Related Issue(s)

Follow-up to #35116.

### DIFF
--- a/packages/nx/src/utils/package-json.ts
+++ b/packages/nx/src/utils/package-json.ts
@@ -370,12 +370,12 @@ function preparePackageInstallation(pkg: string, requiredVersion: string) {
   };
 
   console.log(`Fetching ${pkg}...`);
-  const packageManager = detectPackageManager();
+  const packageManager = detectPackageManager(workspaceRoot);
   const isVerbose = process.env.NX_VERBOSE_LOGGING === 'true';
   generatePackageManagerFiles(tempDir, packageManager);
 
-  const preInstallCommand = getPackageManagerCommand(packageManager).preInstall;
   const pmCommands = getPackageManagerCommand(packageManager);
+  const preInstallCommand = pmCommands.preInstall;
   let addCommand = pmCommands.addDev;
   if (packageManager === 'pnpm') {
     addCommand = 'pnpm add -D'; // we need to ensure that we are not using workspace command
--- a/packages/nx/src/utils/package-json.ts
+++ b/packages/nx/src/utils/package-json.ts
@@ -187,8 +187,6 @@ export function buildTargetFromScript(
   };
 }
 
-let packageManagerCommand: PackageManagerCommands | undefined;
-
 export function getMetadataFromPackageJson(
   packageJson: PackageJson,
   isInPackageManagerWorkspaces: boolean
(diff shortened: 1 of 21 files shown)


packages/nx/src/utils/package-json.ts
```diff
@@ -231,9 +229,18 @@ export function readTargetsFromPackageJson(
...
   const res: Record<string, TargetConfiguration> = {};
   const includedScripts = nx?.includedScripts || Object.keys(scripts ?? {});
-  for (const script of includedScripts) {
-    packageManagerCommand ??= getPackageManagerCommand();
-    res[script] = buildTargetFromScript(script, scripts, packageManagerCommand);
+  if (includedScripts.length > 0) {
+    const packageManagerCommand = getPackageManagerCommand(
+      detectPackageManager(workspaceRoot),
+      workspaceRoot
+    );
```
AgentEnder: This is a minor degradation, we'll be detecting the package manager / getting the command set once per project instead of once per module lifetime. I'd prefer to move this outside of here, unless you see a good reason to re-detect when reading the targets per package.json... 

If we don't want to have it at module level, I'd be fine with passing it inot the readTargetsFromPackageJson fn or something

AgentEnder (changes requested):
Feel free to push back if you want, but I think this'd be cleaner if we moved where the package manager command was fetched. Previously it was once per workspace, now it'd be once per project etc and it adds an if around the for loop that wouldn't otherwise be needed.

Its definitely pretty minor, but I'd like it changed.

AgentEnder (approved):

AgentEnder (approved):

Issue #35503, opened 2026-04-30
Exploring a modern Nx-native NestJS v12 plugin aligned with Nest CLI and Project Crystal


FrozenPandaz:
### Discussed in https://github.com/nrwl/nx/discussions/35479

<div type='discussions-op-text'>

<sup>Originally posted by **vrxj81** April 28, 2026</sup>
Hi Nx team and community,

I’m currently exploring the design of a modern Nx-native NestJS plugin aligned with the upcoming NestJS v12 direction, and I’d like to ask for early feedback from the Nx core team and community before investing too much implementation effort.

This is not meant as a complaint about the existing `@nx/nest` plugin. The current plugin has served the ecosystem well. The question is whether the NestJS v12 shift creates an opportunity — or need — for a more modern integration model.

## Context

From what I understand, NestJS v12 is moving more strongly toward:

* ESM-first / modern Node.js conventions
* SWC-based build workflows
* Vitest-friendly testing setups
* stronger alignment with the official Nest CLI
* less reliance on older Angular-era schematic assumptions
* potentially more schema-first / Standard Schema-oriented validation patterns
* continued support for Fastify as a serious platform option

At the same time, Nx has evolved significantly with:

* Project Crystal
* inferred targets
* plugin-based workspace intelligence
* stronger support for polyglot / framework-specific tooling without requiring every target to be manually configured

That raises the question:

> Should a future NestJS v12 integration for Nx continue the current pattern of deeply wrapping/generating Nest projects, or should it delegate framework scaffolding to the official Nest CLI where possible and focus the Nx plugin on workspace integration, target inference, and monorepo ergonomics?

## Current thinking

The design principle I’m exploring is:

> Nest owns framework generation.
> Nx owns workspace orchestration.
> Optional architecture conventions should be explicit and configurable.

In practical terms, this would mean a plugin that:

* delegates as much Nest-specific scaffolding as possible to the official Nest CLI
* avoids copying Nest templates into the Nx plugin
* adds Nx workspace intelligence on top:

  * project registration
  * tags
  * source roots
  * path aliases
  * project graph integration
  * module boundary support
  * target inference
* supports Project Crystal-style inferred targets for:

  * `serve`
  * `build`
  * `test`
  * `lint`
  * possibly `e2e`
* aligns with Nest v12 defaults:

  * ESM-first
  * SWC-first
  * Vitest-first where appropriate
* treats Fastify as a first-class option
* provides a migration path for existing Nx/Nest projects

[...]

PR #35551, opened 2026-05-03
feat(nest): infer Nest projects from nest-cli.json

vrxj81:
## Current Behavior

The `@nx/nest` plugin does not detect NestJS projects from `nest-cli.json`, so standard Nest applications are not integrated into the Nx project graph unless their targets are configured manually.

## Expected Behavior

The `@nx/nest` plugin detects projects from `nest-cli.json` and infers cacheable `build` and continuous `start` targets that delegate to the Nest CLI.

Inferred build outputs follow the effective TypeScript configuration used by Nest CLI: an explicit `compilerOptions.tsConfigPath`, a `tsc` builder `configPath`, or the default `tsconfig.build.json` / `tsconfig.json`. This includes inherited `outDir` values and output directories outside the project root.

The Nest library generator also updates `tsconfig.lib.json` through `projectRoot`, which supports TypeScript-solution workspaces before project registration is available.

## Validation

- `pnpm nx run-many -t test,build,lint -p nest`
- repository pre-push validation

## Related Issue(s)

Refs #35503


vrxj81:
## Manual test steps
- created Nest app using `nest new nest-test`
- added Nx using `nx init`
- installed `@nx/nest` plugin through local path entry in `package.json`, running `npm install` and adding plugin config to `nx.json` ( `{ plugins: [ "@nx/nest/plugin"} `)
- removed default nest NPM build target from `package.json`
- ran `nx show projects`
- ran `nx show project nest-test`
- ran `nx build nest-test` twice (target and cache test)
- ran `nx serve nest-test` 

## Results

### nx show projects
```bash
nest-test
```

### nx show project nest-test
image


### nx build nest-test (first run)
```text

> nx run nest-test:build

> nest build


———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Successfully ran target build for project nest-test (2s)
```
### nx build nest-test (second run)
```text
> nx run nest-test:build  [local cache]

> nest build


———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Successfully ran target build for project nest-test (79ms)
```

### nx serve nest-test
```text
> nx run nest-test:serve

> nest start --watch

[13:30:59] Starting compilation in watch mode...

[13:30:59] Found 0 errors. Watching for file changes.


[Nest] 71857  - 06/05/2026, 13:31:00     LOG [NestFactory] Starting Nest application...
[Nest] 71857  - 06/05/2026, 13:31:00     LOG [InstanceLoader] AppModule dependencies initialized +5ms
[Nest] 71857  - 06/05/2026, 13:31:00     LOG [RoutesResolver] AppController {/}: +1ms
[Nest] 71857  - 06/05/2026, 13:31:00     LOG [RouterExplorer] Mapped {/, GET} route +1ms
[Nest] 71857  - 06/05/2026, 13:31:00     LOG [NestApplication] Nest application successfully started +1ms

———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 NX   Successfully ran target serve for project nest-test
```

FrozenPandaz:
Hi @vrxj81 :wave: 

How is this going? Do you need any help? Is it ready for review?

vrxj81:
> Hi @vrxj81 👋
> 
> How is this going? Do you need any help? Is it ready for review?

hi @FrozenPandaz,
It's like you read my mind.
Any pointers would be helpful.
The code is ready for review, but the CI checks main-linux fail and no clue what to do about it.

FrozenPandaz:
The failures seem unrelated. Please try rebasing.

vrxj81:
> The failures seem unrelated. Please try rebasing.

rebased

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,187 @@
...
+  namedInputs: NxJsonConfiguration['namedInputs']
+): TargetConfiguration['inputs'] {
+  return [
+    ...('production' in namedInputs
+      ? ['production', '^production']
+      : ['default', '^default']),
+    {
+      externalDependencies: ['@nestjs/cli'],
+    },
+  ];
```
FrozenPandaz, line 170: We should add the following here so that the input includes the root tsconfig that is probably read to build the nest app.

```
import { TS_SOLUTION_SETUP_TSCONFIG_INPUT } from '@nx/js/internal';
```


```suggestion
  return [
    ...('production' in namedInputs
      ? ['production', '^production']
      : ['default', '^default']),
    {
      externalDependencies: ['@nestjs/cli'],
    },
    TS_SOLUTION_SETUP_TSCONFIG_INPUT,
  ];
```

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const nestCliConfigGlob = '**/nest-cli.json';
+
+export const createNodes: CreateNodesV2<NestPluginOptions> = [
+  nestCliConfigGlob,
+  async (configFilePaths, options, context) => {
+    const normalizedOptions = normalizeOptions(options);
+    const validConfigFiles = configFilePaths.filter((configFilePath) =>
+      checkIfConfigFileShouldBeProject(
+        normalizeProjectRoot(dirname(configFilePath)),
```
FrozenPandaz: nit: I'd recommend passing in the full `configFilePath`..

1. it matches the function name more accurately...
2. Who knows, it might come in handy.. Currently information is effectively being stripped before it enters the function. Passing the full configFilePath is more information and the function itself can choose to normalizeProjectRoot(dirname()) if it needs to.

It's a nit though

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,187 @@
...
+        context
+      )
+    );
+
+    return createNodesFromFiles(
```
FrozenPandaz: Other plugins utilize a cache so that subsequent project graph calculations are fast.

See an example here:
PluginCache instantiated- https://github.com/nrwl/nx/blob/master/packages/vite/src/plugins/plugin.ts#L73
Hashes calculated once up front - https://github.com/nrwl/nx/blob/master/packages/vite/src/plugins/plugin.ts#L102

Cache read - https://github.com/nrwl/nx/blob/master/packages/vite/src/plugins/plugin.ts#L136
Cache write - https://github.com/nrwl/nx/blob/master/packages/vite/src/plugins/plugin.ts#L177

packages/nest/package.json
```diff
@@ -44,9 +44,16 @@
...
     },
+    "./plugin": {
+      "@nx/nx-source": "./plugin.ts",
+      "types": "./dist/plugin.d.ts",
+      "default": "./dist/plugin.js"
+    },
     "./package.json": "./package.json",
     "./migrations.json": "./migrations.json",
     "./generators.json": "./generators.json",
+    "./src/generators/*/schema.json": "./src/generators/*/schema.json",
```
FrozenPandaz: Sorry, I changed the build so these should now point to `./dist/src...`

packages/nest/PLUGIN.md
FrozenPandaz, line 1: We don't need this file here.. it's used mostly for plugins that have atomized tasks which this one does not. 

General documentation for the plugin can be done after.. and will be in the docs

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,187 @@
...
+): string {
+  if (!outputPath) {
+    return joinPathFragments('{projectRoot}', 'dist');
+  }
+
+  if (isAbsolute(outputPath)) {
+    return `{workspaceRoot}/${relative(
+      workspaceRoot,
+      resolve(workspaceRoot, outputPath)
+    )}`;
```
FrozenPandaz: This should also use `joinPathFragments`

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,187 @@
+import {
+  CreateNodesContextV2,
```
FrozenPandaz: We recently deprecated the "V2" versions of these typings.. please use `CreateNodesContext` instead.

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,187 @@
...
+      dependsOn: [`^${options.buildTargetName}`],
+      inputs: getBuildInputs(namedInputs),
+      outputs: buildOutputs ? [buildOutputs] : [],
+      metadata: {
+        technologies: ['nest'],
+        description: 'Build the Nest project.',
+      },
+    },
+    [options.serveTargetName]: {
+      command: 'nest start --watch',
```
FrozenPandaz: Even though this doesn't match most other tools... We should probably use plain `nest start` to align with what NestJS has as their "start" command

```
nx start nest-app         ~= cd nest-app && nest start
nx start nest-app --watch ~= cd nest-app && nest start --watch
```

FrozenPandaz (changes requested):
This is a great start! Thank you so much for contributing this! I think this only requires a handful of changes so I'm looking forward to  getting it merged soon. 🤞

packages/nest/src/plugins/plugin.ts
```diff
@@ -0,0 +1,231 @@
...
+
+function buildNestTargets(
+  projectRoot: string,
+  nestCliConfig: NestCliConfig,
+  options: NestPluginOptions,
+  context: CreateNodesContext
+): Record<string, TargetConfiguration> {
+  const namedInputs = getNamedInputs(projectRoot, context);
+  const buildOutputs = normalizeOutputPath(
+    nestCliConfig.compilerOptions?.outputPath,
```
FrozenPandaz, line 125: `buildNestTargets` derives the build `outputs` from `nestCliConfig.compilerOptions?.outputPath`, but `nest-cli.json` has no `outputPath` property. I checked both `@nestjs/cli`'s `CompilerOptions` type and the schemastore schema — the field is `tsConfigPath` / `builder`, and the actual output directory comes from the referenced tsconfig's `outDir` (default `tsconfig.build.json`) or the SWC builder's `outDir`.

Because `compilerOptions.outputPath` is always `undefined`, `normalizeOutputPath` always falls back to `{projectRoot}/dist`, and the `isAbsolute` / `..` branches are unreachable. Default projects happen to land on `{projectRoot}/dist` so they're fine — but anyone with a custom `outDir` gets an inferred `outputs` pointing at the wrong directory, and Nx will cache/restore the wrong artifacts silently. This is the one spot where copying the webpack plugin's shape doesn't carry over, since there `outputPath` is a real resolved-config value.

Suggested fix: read `compilerOptions.tsConfigPath` (default `tsconfig.build.json`) from `nest-cli.json`, resolve that tsconfig's `compilerOptions.outDir`, and fall back to `{projectRoot}/dist` only when it's unset. The "conservative outputs" test in `plugin.spec.ts` uses the phantom `outputPath` field too, so it'll want a real `outDir`/`tsConfigPath` fixture once this changes.

FrozenPandaz (changes requested):
Thanks for addressing that feedback.. Got 1 more big one for you

johnnyhuirilef:
@vrxj81 I got to the same corner of `nest-cli.json` from a different direction, so this PR was useful reading.

I have been looking at NestJS 12 support in `@nx/nest` and opened a few things along the way — #36875 / #36876, #36878 / #36879, and #36881, a `bundler` option for the Nest application generator. While I tested those against real Nest workspaces, I hit exactly what @FrozenPandaz flagged in July: the Nest CLI takes its output directory from the tsconfig that `compilerOptions.tsConfigPath` points at, or from the SWC builder's `outDir`. There is no `outputPath` in `nest-cli.json`, so the reads at plugin.ts:125-126 always come back `undefined` and the `isAbsolute` / `..` branches never run.

From the outside it looks like the last thing standing between this and a merge.

Are you still on it? If so I would rather not step on your work, and I am happy to review or test a fix against a real workspace whenever that helps.

If you would rather hand that piece off, I can pick it up: read `tsConfigPath`, resolve the referenced tsconfig's `outDir`, fall back to `{projectRoot}/dist` only when it is unset, and swap the phantom `outputPath` fixture in the "conservative outputs" test. Your call either way — this plugin is your work and it should land under your name.

@FrozenPandaz, happy to help here if it is useful.

AgentEnder:
Thanks for the PR. `nest:test` is the task failing on CI, and there are two open change requests from @FrozenPandaz, so I am moving this to draft rather than leaving it in the review queue. Push the fixes and mark it ready, and Jason will pick the review back up.

One of the two failures is `e2e-angular:e2e-ci--src/ng-add.test.ts`, which a Nest plugin change has no obvious reason to break. Worth a glance, but do not lose time on it before `nest:test`.

<details>
<summary>

</summary>

> Task names come from the Nx Cloud run behind the check, CIPE `6a57eef5b8337dba85f3ee2f`, on head `21889baa`. GitHub only reports the workflow, `main-linux`, which is why the earlier note could not say what actually broke.
>
> Both failed tasks were `cache-miss`, so neither is a restored result. The `affected` run itself shows as `CANCELED` rather than `FAILED`, which is Nx stopping the remaining work once a task fails, and the run group records no critical error, so this is a task failure rather than an agent or workflow problem.
>
> These are the tasks as of that commit. A push changes them.

</details>

johnnyhuirilef:
@vrxj81 I was wrong. I said the `outputPath` read was the last item left. The note from @AgentEnder about `nest:test` showed that it is not. So I ran the test suite on your branch to see what is left. The news is better than the CI check suggests.

**The 9 failed tests do not come from your plugin.** They come from the two generator specs, which are behind master. I copied the `application.spec.ts` and `library.spec.ts` files from master over your production code. The full suite then passed, 122 of 122.

Master changed in four places since June:

- TS-solution tsconfigs now use `nodenext`, with `rootDir` and `tsBuildInfoFile`
- TS-solution projects now carry a `package.json` instead of a `project.json`
- Project names are now scoped with `@proj/`
- The `prune-lockfile` outputs moved

The rebase is clean as text, so git showed no conflict. Your branch quietly restored the older expectations.

The fix is one command after a rebase, instead of nine assertions by hand:

```
git checkout master -- packages/nest/src/generators/application/application.spec.ts packages/nest/src/generators/library/library.spec.ts
```

**The rebase onto master is clean.** I did it on my machine to run the tests. There are no conflicts.

**Your change to `library/lib/update-tsconfig.ts` is necessary.** It looks like a cleanup, but it is not. `normalize-options.ts:58-59` sets `projectName` to the scoped `importPath` under TS solutions. At that point no project is registered, so `readProjectConfiguration` throws. `projectRoot` is always set, and it is the same value on the other path. Keep it. I write this here so that a reviewer does not ask you to revert it.

For the `outputPath` item, I opened vrxj81/nx#1 against your branch, not a second PR here. If you take it, the commit lands under your name. It reads the tsconfig that the Nest CLI resolves: first `tsConfigPath`, then the `configPath` of a `tsc` builder, then `tsconfig.build.json` with `tsconfig.json` as the fallback. It takes the `outDir` from that file. It reads the file with `readTsConfig` from `@nx/js`, so an `outDir` that comes through `extends` also counts. Seven tests cover the branches. Take it, take part of it, or close it.

johnnyhuirilef:
@AgentEnder the `nest:test` failure is gone on the rebased head. I ran the suite locally so the result is not just an assumption.

```
Test Suites: 1 skipped, 20 passed, 20 of 21 total
Tests:       1 skipped, 122 passed, 123 total
Snapshots:   27 passed, 27 total
```

I ran what the inferred target runs, from `packages/nest`:

```bash
NODE_OPTIONS=--experimental-vm-modules npx jest --passWithNoTests --detectOpenHandles --forceExit
```

The skipped suite is `resource.spec.ts`. It carries `xdescribe` and a TODO on master too, so this branch did not disable it.

Two changes fixed it together, and neither one works alone. `application.spec.ts` and `library.spec.ts` are back to their master content, and the branch now sits on top of current master. The diff against master is 5 files, all under `packages/nest`.

This only proves the unit suite. Lint, build and e2e still need CI.

vrxj81:
> @AgentEnder the `nest:test` failure is gone on the rebased head. I ran the suite locally so the result is not just an assumption.
> 
> ```
> Test Suites: 1 skipped, 20 passed, 20 of 21 total
> Tests:       1 skipped, 122 passed, 123 total
> Snapshots:   27 passed, 27 total
> ```
> 
> I ran what the inferred target runs, from `packages/nest`:
> 
> ```shell
> NODE_OPTIONS=--experimental-vm-modules npx jest --passWithNoTests --detectOpenHandles --forceExit
> ```
> 
> The skipped suite is `resource.spec.ts`. It carries `xdescribe` and a TODO on master too, so this branch did not disable it.
> 
> Two changes fixed it together, and neither one works alone. `application.spec.ts` and `library.spec.ts` are back to their master content, and the branch now sits on top of current master. The diff against master is 5 files, all under `packages/nest`.
> 
> This only proves the unit suite. Lint, build and e2e still need CI.

@johnnyhuirilef  yes, I’m still on it, and thank you for digging into this so thoroughly.
I merged your fork PR, so the tsconfig/outDir fix is now on the branch with your authorship preserved. I also rebased the branch onto the current upstream master, restored the application and library generator specs to the current baseline, and retained the update-tsconfig.ts change needed for TypeScript-solution workspaces.
The targeted Nest build, test, and lint validation passed, as did the repository’s pre-push gate. The PR is now marked ready for review and all existing review threads are resolved. The new GitHub workflows are currently waiting for the required maintainer approval because the branch comes from a fork.
Thanks again for the careful investigation, the tests, and the fix—it made getting this back into reviewable shape much easier.

leo29plns:
Hi @FrozenPandaz, any updates on this PR?

PR #35554, opened 2026-05-04
docs(nx-cloud): document --targets flag for start-agent

edbzn:
## Current Behavior

`nx-cloud start-agent` supports a `--targets <csv>` flag (visible via `nx-cloud start-agent --help`) that pins an agent to a comma-separated subset of target names. The flag is not mentioned on https://nx.dev/docs/reference/nx-cloud-cli or in the Manual DTE guide, so users hitting OOM or contention on memory-constrained runners cannot discover it without reading the bundled CLI source.

## Expected Behavior

The flag is documented in:

- the `nx-cloud start-agent` Options table on the Cloud CLI reference, with a short usage example
- the Manual DTE guide, with a small example showing per-pool partitioning to isolate memory-heavy task families (e2e, builds, component-tests)

## Related Issue(s)

N/A — documentation gap discovered while debugging OOM-kill failures on a manual-DTE GitHub Actions setup.


leosvelperez:
This was superseded/addressed by https://github.com/nrwl/nx/pull/36742.

PR #35562, opened 2026-05-04
fix(core): run selected projects with --exclude-task-dependencies

djohnson-aperture:
`nx exec --projects @org/taskWithDeps --excludeTaskDependencies -- {command}` should still run the command for the passed project even if it's dependent tasks are ignored.

## Current Behavior
When `nx exec --projects @org/taskWithDeps --excludeTaskDependencies -- echo 1` 1 is never echo'd.
## Expected Behavior
The command should run for at least the passed in projects, even if their dependencies are ignored.


leosvelperez (approved):
Thanks!

PR #35567, opened 2026-05-05
feat(angular)!: remove deprecated ngrx generator

leosvelperez:
## Current Behavior

The `@nx/angular:ngrx` generator is exposed and functional, but has been deprecated since Nx v18 in favor of `@nx/angular:ngrx-root-store` (root state) and `@nx/angular:ngrx-feature-store` (feature state).

## Expected Behavior

The `@nx/angular:ngrx` generator is removed. Users invoke `@nx/angular:ngrx-root-store` for root state and `@nx/angular:ngrx-feature-store` for feature state instead.

A migration splits any `@nx/angular:ngrx` generator defaults set in `nx.json` across the two replacement generator keys, renames the deprecated `module` option to `parent`, and drops the obsolete `root` toggle (intent is now expressed by which generator is invoked). The migration handles both the flat (`"@nx/angular:ngrx"`) and nested (`"@nx/angular": { "ngrx": ... }`) defaults shapes.

BREAKING CHANGE: The `@nx/angular:ngrx` generator has been removed. Use `@nx/angular:ngrx-root-store` for root state and `@nx/angular:ngrx-feature-store` for feature state instead.


packages/angular/src/migrations/update-23-0-0/migrate-ngrx-generator-defaults.ts
```diff
@@ -0,0 +1,141 @@
...
+  'directory',
+  'facade',
+  'barrels',
+  'skipImport',
+  'skipPackageJson',
+  'skipFormat',
+] as const;
+
+export default async function (tree: Tree) {
+  const nxJson = readNxJson(tree);
```
FrozenPandaz, line 47: We should also migrate generator defaults in `project.json` files.

FrozenPandaz (changes requested):

FrozenPandaz (approved):

Issue #35536, opened 2026-05-01
The built version of nx has vulnerable pinned versions


amrsalama:
### Current Behavior

The built nx package on npm has pinned dependencies' versions, e.g. https://www.npmjs.com/package/nx/v/22.7.1?activeTab=code has

- "follow-redirects": "1.15.11"  https://github.com/advisories/GHSA-r4q5-vmmm-2653 (fixed in >=1.16.0)
- "yaml": "2.8.0"  https://github.com/advisories/GHSA-48c2-rrv3-qjmp (fixed in >=2.8.3)
- "brace-expansion": "5.0.2" https://github.com/advisories/GHSA-f886-m6hf-6m8v (fixed in >=5.0.5)

I'm not sure about the build process in nx, but in https://github.com/nrwl/nx/blob/22.7.1/packages/nx/package.json I can see the yaml version is ^2.6.0 (not pinned), and  "follow-redirects" is not included.

This issue started in nx >=22.7.0

### Expected Behavior

Vulnerable versions of dependencies to be updated as in 22.6.x

### Steps to Reproduce

Checking https://www.npmjs.com/package/nx/v/22.7.1?activeTab=code 

### Nx Report

```shell
NX >= 22.7.0
```

### Failure Logs

```shell

```

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

EinfachHans:
`axios` as well

```
axios  1.0.0 - 1.15.1
Severity: high
Axios: Authentication Bypass via Prototype Pollution Gadget in `validateStatus` Merge Strategy - https://github.com/advisories/GHSA-w9j2-pvgh-6h63
Axios: Incomplete Fix for CVE-2025-62718 — NO_PROXY Protection Bypassed via RFC 1122 Loopback Subnet (127.0.0.0/8) in Axios 1.15.0 - https://github.com/advisories/GHSA-pmwg-cvhr-8vh7
Axios: Invisible JSON Response Tampering via Prototype Pollution Gadget in `parseReviver` - https://github.com/advisories/GHSA-3w6x-2g7m-8v23
Axios has prototype pollution read-side gadgets in HTTP adapter that allow credential injection and request hijacking - https://github.com/advisories/GHSA-q8qp-cvcw-x6jj
Axios: Null Byte Injection via Reverse-Encoding in AxiosURLSearchParams - https://github.com/advisories/GHSA-xhjh-pmcv-23jw
[...]

PR #35568, opened 2026-05-05
fix(core): bump axios to 1.16.0 for all packages

Optischa:
Fix Axios CVE's:
Axios: Authentication Bypass via Prototype Pollution Gadget in `validateStatus` Merge Strategy - https://github.com/advisories/GHSA-w9j2-pvgh-6h63
Axios: Incomplete Fix for CVE-2025-62718 — NO_PROXY Protection Bypassed via RFC 1122 Loopback Subnet (127.0.0.0/8) in Axios 1.15.0 - https://github.com/advisories/GHSA-pmwg-cvhr-8vh7
Axios: Invisible JSON Response Tampering via Prototype Pollution Gadget in `parseReviver` - https://github.com/advisories/GHSA-3w6x-2g7m-8v23
Axios has prototype pollution read-side gadgets in HTTP adapter that allow credential injection and request hijacking - https://github.com/advisories/GHSA-q8qp-cvcw-x6jj
Axios: Null Byte Injection via Reverse-Encoding in AxiosURLSearchParams - https://github.com/advisories/GHSA-xhjh-pmcv-23jw
Axios: CRLF Injection in multipart/form-data body via unsanitized blob.type in formDataToStream - https://github.com/advisories/GHSA-445q-vr5w-6q77
Axios: no_proxy bypass via IP alias allows SSRF - https://github.com/advisories/GHSA-m7pr-hjqh-92cm
Axios: unbounded recursion in toFormData causes DoS via deeply nested request data - https://github.com/advisories/GHSA-62hf-57xw-28j9
Axios' HTTP adapter-streamed uploads bypass maxBodyLength when maxRedirects: 0 - https://github.com/advisories/GHSA-5c9x-8gcm-mpgx
Axios: HTTP adapter streamed responses bypass maxContentLength - https://github.com/advisories/GHSA-vf2m-468p-8v99
Axios: Prototype Pollution Gadgets - Response Tampering, Data Exfiltration, and Request Hijacking - https://github.com/advisories/GHSA-pf86-5x62-jrwf
Axios: Header Injection via Prototype Pollution - https://github.com/advisories/GHSA-6chq-wfr3-2hj9
Axios: XSRF Token Cross-Origin Leakage via Prototype Pollution Gadget in `withXSRFToken` Boolean Coercion - https://github.com/advisories/GHSA-xx6v-rp6x-q39c

PR #35577, opened 2026-05-05
fix(core): update axios to 1.16.0

beeman:
Update the root workspace, nx, and create-nx-workspace axios pins to 1.16.0 and collapse the lockfile follow-redirects resolution to 1.16.0.

This keeps the published package metadata current for the latest axios security fixes and removes the GHSA-r4q5-vmmm-2653 scanner hit from the repo lockfile without unrelated lockfile churn.

Axios has accumulated multiple recent advisories, so keeping this dependency current is urgent even when individual reports are scanner-driven.

Advisory: https://github.com/advisories/GHSA-r4q5-vmmm-2653

Related: https://github.com/nrwl/nx/issues/35536

### DIFF
--- a/package.json
+++ b/package.json
@@ -365,7 +365,7 @@
     "@types/three": "^0.166.0",
     "@yarnpkg/lockfile": "^1.1.0",
     "@yarnpkg/parsers": "3.0.2",
-    "axios": "1.15.0",
+    "axios": "1.16.0",
     "classnames": "^2.5.1",
     "cliui": "^8.0.1",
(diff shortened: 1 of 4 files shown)


Optischa:
Same fix for Axios as [here](https://github.com/nrwl/nx/pull/35568)?

FrozenPandaz:
Hey @beeman Thank you for the contribution as always! We resolved it on https://github.com/nrwl/nx/pull/35568 which was opened first and allowed a first time contributor to land their first contribution. :) Thank you!

Issue #35575, opened 2026-05-05
nx release publish marks every task failed under pnpm 11 ("could not be extracted") because pnpm 11's native publish writes nothing to stdout


comp615:
## Current Behavior

When running `nx release publish` (or invoking the `@nx/js:release-publish` executor) in a workspace using **pnpm 11**, every publish task fails with:

```
The pnpm publish output data could not be extracted. Please report this issue on https://github.com/nrwl/nx
```

even though the package is actually published successfully to the registry.

When `releasePublish` is called programmatically with multiple projects and the default `nx-bail=true`, the first batch of publishes complete on the registry but get marked as failed, then **every remaining project in the same invocation is skipped** with `Tasks not run because their dependencies failed or --nx-bail=true`. The end result is a half-published release: tags / changelog / version bumps land in source control but most tarballs never reach the registry. A retry sees `"No new version updates were found, nothing to publish"` because the tags already exist, so the missing tarballs never get a second chance.

## Expected Behavior

`nx release publish` should recognize a successful `pnpm publish --json` invocation under pnpm 11, mark the task as successful, and not abort the rest of the release.

## GitHub Repo

N/A (private monorepo). Reproduction is fully self-contained below.

## Steps to Reproduce

Empirical reproduction of the underlying mismatch (no Nx required):

```sh
mkdir /tmp/pnpm-json-repro && cd /tmp/pnpm-json-repro
echo '{"name":"@scope/x","version":"0.0.1"}' > package.json
echo "x" > index.js

# pnpm 10.x (delegates to npm CLI) — produces the JSON Nx expects:
npx pnpm@10 publish --json --dry-run --no-git-checks --registry=https://example.invalid
# stdout:
# {
#   "id": "@scope/x@0.0.1",
#   "name": "@scope/x",
#   "version": "0.0.1",
#   "size": 211,
#   "filename": "scope-x-0.0.1.tgz",
#   ...
# }

# pnpm 11.x (native publish) — outputs literally 0 bytes on stdout:
npx pnpm@11 publish --json --dry-run --no-git-checks --registry=https://example.invalid
# stdout: (empty, 0 bytes)
# exit code: 0
```

End-to-end reproduction inside Nx:

1. In any Nx workspace using pnpm, set `packageManager: "pnpm@11.0.4"` (or run via `corepack use pnpm@11`).
2. Run `nx release publish --tag latest --dry-run` (or call `releasePublish({...})` programmatically across multiple projects with `maxParallel > 1`).
[...]

PR #35580, opened 2026-05-05
fix(js): trust exit code for pnpm >= 11 in release-publish executor

comp615:
## Current Behavior

`nx release publish` silently fails when the workspace uses **pnpm 11+**:

- pnpm 11 reimplemented `pnpm publish` natively (pnpm/pnpm#10591). It no longer delegates to the npm CLI.
- The new implementation accepts `--json` but writes **nothing to stdout on success** (the publish handler returns `undefined`; `--json` only suppresses the human-readable reporter).
- `release-publish.impl.ts` runs the publish command and parses stdout via `extractNpmPublishJsonData`. With empty stdout the extractor returns `{ jsonData: null }`, the executor logs "The pnpm publish output data could not be extracted" and returns `success: false`.
- Combined with the default `--nx-bail=true`, this aborts the rest of a release after the first parallel batch: tarballs reach the registry for the packages that were attempted, but Nx marks the task failed and skips the remaining packages. Releases look "passed" in CI but are partial.

Tracked at:
- pnpm/pnpm#11476 — `pnpm publish --json produces empty stdout in pnpm 11 (regression from pnpm 10)`
- nrwl/nx#35575 — Nx-side report

## Expected Behavior

When `pm === 'pnpm'` and the detected pnpm major version is `>= 11`, trust the exit code: `execSync` already throws on non-zero exit, so reaching this branch means the publish succeeded. Print a minimal summary line and return `{ success: true }`, matching the existing `pm === 'bun'` branch precedent (added in #34835).

This intentionally drops the verbose `=== Tarball Details ===` table for pnpm >= 11. Once pnpm restores `--json` stdout output (pnpm/pnpm#11476), the JSON extraction path will succeed again and this short-circuit becomes a no-op that can be removed.

The official pnpm-side workaround for users who need the npm-CLI semantics today is `pnpm pack && npm publish *.tgz` (documented in the changeset for pnpm/pnpm#10591).

## Verification of no duplicate PRs

Before opening, I searched open Nx PRs for `extract-npm-publish-json-data`, `release-publish.impl`, and `#35575` — no open PRs touch this code path.

## Related Issue(s)

Refs nrwl/nx#35575
Refs pnpm/pnpm#11476
Refs pnpm/pnpm#10591

🤖 Authored with assistance from Amp.

### DIFF
--- a/packages/js/src/executors/release-publish/release-publish.impl.spec.ts
+++ b/packages/js/src/executors/release-publish/release-publish.impl.spec.ts
@@ -2,6 +2,7 @@ import {
   ExecutorContext,
   readJsonFile,
   detectPackageManager,
+  getPackageManagerVersion,
 } from '@nx/devkit';
 import { execSync } from 'child_process';
 import { PublishExecutorSchema } from './schema';
@@ -18,6 +19,7 @@ jest.mock('../../utils/npm-config');
 jest.mock('@nx/devkit', () => ({
   ...jest.requireActual('@nx/devkit'),
   detectPackageManager: jest.fn(() => 'npm'),
+  getPackageManagerVersion: jest.fn(() => '10.0.0'),
   readJsonFile: jest.fn(),
 }));
 jest.mock('./extract-npm-publish-json-data');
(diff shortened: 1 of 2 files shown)


comp615:
I think since the pnpm fix got in so quick, we might be able to avoid this and just have some pnpm versions that don't work

PR #35584, opened 2026-05-05
fix(core): isolate cache env vars in splitArgs spec

FrozenPandaz:
## Current Behavior

The `splitArgs` describe block in `packages/nx/src/utils/command-line-utils.spec.ts` saves and clears `NX_BASE`, `NX_HEAD`, and `NX_PARALLEL` so the surrounding shell environment doesn't bleed into assertions. However, the production code in `splitArgsIntoNxArgsAndOverrides` also reads four cache-related env vars as fallbacks for the `skipNxCache` / `skipRemoteCache` defaults:

- `NX_SKIP_NX_CACHE`
- `NX_DISABLE_NX_CACHE`
- `NX_SKIP_REMOTE_CACHE`
- `NX_DISABLE_REMOTE_CACHE`

These are not isolated. When a developer runs the test suite via the outer `nx` invocation with flags like `--skipNxCache`, nx propagates them to spawned child processes as `NX_SKIP_NX_CACHE=true`. That env var leaks into the Jest worker, flips `skipNxCache` to `true`, and breaks every test in the `splitArgs` block that asserts on the defaults — six failures in total (`should split nx specific arguments into nxArgs`, `should default to having a base of main`, `should return configured base branch from nx.json`, `should return a default base branch if not configured in nx.json`, `should split projects when it is a string`, `should set base and head based on environment variables in affected mode`).

## Expected Behavior

The `splitArgs` specs should pass regardless of which cache-related flags or env vars are set in the surrounding environment, just like they already do for `NX_BASE` / `NX_HEAD` / `NX_PARALLEL`.

This PR extends the existing save/clear/restore pattern to cover the four cache env vars and factors the conditional-restore logic (delete when originally unset, otherwise reassign — to avoid Node's `process.env[key] = undefined` coercing to the string `"undefined"`) into a small `restoreEnv` helper.

## Related Issue(s)

N/A

### DIFF
--- a/packages/nx/src/native/watch/watcher.rs
+++ b/packages/nx/src/native/watch/watcher.rs
@@ -495,7 +495,9 @@ mod tests {
             }
         });
         w.watch_inner(callback).expect("start watch");
-        std::thread::sleep(Duration::from_millis(150));
+        // Drop any startup events FSEvents leaks from before the watch began.
+        std::thread::sleep(Duration::from_millis(300));
+        captured.lock().unwrap().clear();
         (w, captured)
     }
 
--- a/packages/nx/src/utils/command-line-utils.spec.ts
+++ b/packages/nx/src/utils/command-line-utils.spec.ts
@@ -4,49 +4,38 @@ import { withEnvironmentVariables as withEnvironment } from '../internal-testing
 jest.mock('../project-graph/file-utils');
 
 describe('splitArgs', () => {
-  let originalBase: string;
-  let originalHead: string;
-  let originalParallel: string;
-  let originalSkipNxCache: string;
-  let originalDisableNxCache: string;
-  let originalSkipRemoteCache: string;
-  let originalDisableRemoteCache: string;
+  const blockedEnvVars = [
+    'NX_BASE',
+    'NX_HEAD',
+    'NX_PARALLEL',
+    'NX_SKIP_NX_CACHE',
+    'NX_DISABLE_NX_CACHE',
+    'NX_SKIP_REMOTE_CACHE',
+    'NX_DISABLE_REMOTE_CACHE',
+  ];
+  let envVarsToRestore: Record<string, string | undefined> = {};
+
+  function blockParentEnvVar(key: string) {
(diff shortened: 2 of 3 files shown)


packages/nx/src/native/watch/watcher.rs
```diff
@@ -495,7 +495,9 @@ mod tests {
             }
         });
         w.watch_inner(callback).expect("start watch");
-        std::thread::sleep(Duration::from_millis(150));
+        // Drop any startup events FSEvents leaks from before the watch began.
+        std::thread::sleep(Duration::from_millis(300));
+        captured.lock().unwrap().clear();
```
AgentEnder, line 500: Does this mean we just don't see any events on the daemon for the first 300ms? Probably fine... but weird?

packages/nx/src/utils/command-line-utils.spec.ts
```diff
@@ -7,26 +7,46 @@ describe('splitArgs', () => {
   let originalBase: string;
   let originalHead: string;
   let originalParallel: string;
+  let originalSkipNxCache: string;
+  let originalDisableNxCache: string;
+  let originalSkipRemoteCache: string;
+  let originalDisableRemoteCache: string;
```
AgentEnder: Lets clean these up a bit... something like

```
let envVarsToRestore = {}
function blockParentEnvVar(key) {
  envVarsToRestore[key] = process.env[key]
  delete process.env[key]
}

beforeEach(() => {
  blockParentEnvVar(...)
})

afterEach(() => {
  for const [key,val] of envVarsToRestore....
})
```

AgentEnder (approved):
Question and nit

packages/nx/src/native/watch/watcher.rs
FrozenPandaz: This is inside `#[cfg(test)]`, so the 300ms sleep only affects test setup — the daemon itself doesn't sleep or drop events. The startup leakage we're filtering out is FSEvents reporting pre-watch activity from the tempdir's parent (`/private/tmp`), which only matters for these isolated tests.

PR #35592, opened 2026-05-06
chore(repo): provision build toolchain via mise in publish workflow

AgentEnder:
## Current Behavior

The `publish` workflow's matrix builds (Linux/macOS/Windows native binaries via N-API) install Java, Node.js, and pnpm manually inside each runner/container. With `@nx/dotnet` now in `nx.json`, these builds also need .NET to be available before the project graph can be loaded — and there's no .NET install on any of the matrix entries today, so the workflow fails at the `pnpm nx run-many --target=build-native` step.

The macOS and `armv7-unknown-linux-gnueabihf` matrix entries already use `mise-action` and `mise.toml`, but the four Linux *docker* entries (Debian + Alpine, x64 + arm64) bypass mise entirely and provision tools through hand-rolled `apt-get` / `apk` / `nodesource` / `npm i -g pnpm` steps.

## Expected Behavior

- All four Linux docker matrix entries now install `mise` from a signed/distro source (apt repo at `https://mise.jdx.dev/deb` for Debian, `apk add mise` from Alpine `community` for Alpine) and provision their entire toolchain — Node.js, Java, .NET, Maven, corepack — from `mise.toml`. This drops ~30 lines of bespoke install logic per entry and keeps versions in lockstep with the non-docker matrix entries, which already use `mise-action`.
- Windows entries gain `choco install dotnet-9.0-sdk -y` alongside the existing OpenJDK install (mise's Windows .NET path is broken upstream — see [jdx/mise#4738](https://github.com/jdx/mise/discussions/4738)).
- The FreeBSD build sets `NX_DOTNET_DISABLE=true` (added to both the `env:` block and the `cross-platform-actions/action` `environment_variables` allowlist so the var actually crosses into the FreeBSD VM) to opt out of the plugin entirely.
- `NODE_VERSION` is now forwarded into `docker run` so containers honor the workflow's pinned Node version through `mise.toml`'s tera template instead of falling back to its `24.11.0` default.
- `mise` itself is installed only via signed repositories — no `curl https://mise.run | sh` — so a hijacked DNS lookup against `mise.run` cannot drop a malicious script into our publish pipeline.

## Related Issue(s)

N/A — workflow fix triggered by `@nx/dotnet` being added to `nx.json`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -130,6 +130,7 @@ jobs:
           - host: windows-latest
             setup: |-
               choco install openjdk --version=21.0.0 -y
+              choco install dotnet-9.0-sdk -y
               rustup target add aarch64-pc-windows-msvc
             build: |
               export JAVA_HOME="C:\Program Files\OpenJDK\jdk-21"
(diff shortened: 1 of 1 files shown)


AgentEnder:
Re-opening from upstream branch — see new PR.

PR #35605, opened 2026-05-07
feat(release): allow customizing dependent bump types via version.dependentBumpType

mpsanchis:
Hi guys,

I did a quick attempt at the following in `release-group-processor.ts`:
```
// TODO: Support influencing the side effect bump in a future version, always patch for now like in legacy versioning
```

It might not be perfect, but I think it'll be easier to iterate from something.

## Current Behavior

Hardcoded `patch` bump.

## Expected Behavior

User can decide.

## Related Issue(s)

Fixes #

### DIFF
--- a/packages/nx/src/command-line/release/version/release-group-processor.spec.ts
+++ b/packages/nx/src/command-line/release/version/release-group-processor.spec.ts
@@ -1485,6 +1485,380 @@ describe('ReleaseGroupProcessor', () => {
     });
   });
 
+  describe('dependentBumpType', () => {
+    let processor: any;
+
+    afterEach(() => {
+      processor = null;
+    });
+
+    it('should default to patch bump for dependents (backward compatible)', async () => {
+      const { nxReleaseConfig, projectGraph, filters } =
+        await createNxReleaseConfigAndPopulateWorkspace(
+          tree,
+          `
+          __default__ ({ "projectsRelationship": "independent" }):
+            - projectA@1.0.0 [js]
+              -> depends on projectB
+            - projectB@2.0.0 [js]
+        `,
+          {
+            version: {
+              conventionalCommits: true,
+            },
+          },
+          mockResolveCurrentVersion
+        );
+
+      processor = await createTestReleaseGroupProcessor(
+        tree,
+        projectGraph,
+        nxReleaseConfig,
+        filters
+      );
+
+      mockDeriveSpecifierFromConventionalCommits.mockImplementation(
+        (_, __, ___, ____, { name: projectName }) => {
+          if (projectName === 'projectB') return 'major';
+          return 'none';
+        }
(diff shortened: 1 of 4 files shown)


packages/nx/src/config/nx-json.ts
```diff
@@ -240,6 +240,16 @@ export interface NxReleaseVersionConfiguration {
...
    */
   applyPreidToDependents?: boolean;
+  /**
+   * Controls the semver bump type applied to dependents when a dependency is versioned.
+   * - 'patch' (default): always bump dependents by a patch version
+   * - 'minor': always bump dependents by a minor version
+   * - 'major': always bump dependents by a major version
+   * - 'match': use the same bump type as the dependency that triggered the bump
+   *
+   * This is 'patch' by default for backward compatibility.
```
JamesHenry, line 250: There's no need to say "for backward compatibility here" - patch remains a logical and reasonable default. A minor change in a dependency absolutely does not imply a minor change in a dependent by default, it is entirely context dependent

packages/nx/src/command-line/release/version/release-group-processor.ts
```diff
@@ -1000,6 +1008,28 @@ export class ReleaseGroupProcessor {
...
+    const finalConfig = this.getCachedFinalConfigForProject(projectName);
+    const configuredBumpType = finalConfig.dependentBumpType ?? 'patch';
+    if (configuredBumpType === 'match') {
+      // Strip any 'pre' prefix from the dependency bump type so that
+      // 'prepatch' becomes 'patch', 'preminor' becomes 'minor', etc.
+      if (dependencyBumpType.startsWith('pre')) {
+        return dependencyBumpType.replace(/^pre/, '') as SemverBumpType;
+      }
+      return dependencyBumpType;
+    }
```
JamesHenry, line 1029: This is now pretty clunky in combination with the recently added `applyPreidToBumpType` - this is being stripped only to be potentially readded again. It feels to me like `resolveDependentBumpType` should be responsible for returning the correct value in the first place instead of then needing `this.applyPreidToBumpType(baseBump, dependent),`?

packages/nx/src/config/nx-json.ts
JamesHenry: Also:

- Needs clarification on how it combines with `applyPreidToDependents`
- Needs corresponding `nx-schema.json` update for the new option

packages/nx/src/command-line/release/version/release-group-processor.spec.ts
JamesHenry, line 1: There's no coverage for different groups having different values for the new config, please add

JamesHenry (changes requested):
Thanks a lot for exploring this @mpsanchis!

PR #35608, opened 2026-05-07
feat(core): enable native Node.js TypeScript stripping by default

jaysoo:
## Current Behavior

Loading `.ts` config files always registered `@swc-node/register` (or `ts-node`) + `tsconfig-paths` up front. New workspaces shipped those deps even though Node 22.6+ strips types natively.

## Expected Behavior

Native Node.js TypeScript stripping is now the default for `.ts` configs and plugin loads. swc/ts-node + `tsconfig-paths` register lazily only when native strip fails (matcher-based, ≤3 attempts; covers `MODULE_NOT_FOUND`, `ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX`, ESM-as-CJS syntax errors,`__dirname` in ESM scope, ESM-with-TLA). `@nx/js` init no longer installs `@swc-node/register` / `@swc/core` / `@swc/helpers`. Opt out via `NX_PREFER_NODE_STRIP_TYPES=false`.

Other changes in this PR are to ensure our generators create Node-compatible config files. For example, not using extensions in imports without `exports` will fail, like `@nx/cypress/plugins/cypress-preset.js` instead of `@nx/cypress/plugins/cypress-preset`. Or mixing in `__dirname`/`__filename` into ESM config files. Or using `import`/`export` in CJS files.

## Related Issue(s)

Fixes NXC-4299


leosvelperez (approved):

FrozenPandaz (approved):
LGTM. The native-strip-by-default rollout is well-structured and the fallback ladder is carefully built.

A few non-blocking suggestions for a follow-up:

- **Centralize the matcher set.** `register.ts` and `handle-import.ts` carry parallel matcher lists, and `handle-import.ts` now open-codes a couple of them as raw `e.code` checks. A single exported `isRecoverableTsLoadError(err, filePath)` called from both sites would stop them drifting.
- **Make the 3-attempt cap explicit.** The `loadTsFile` loop is `for (let attempt = 1; ; attempt++)`, bounded only by the registration flags. An explicit bound would codify the "at most three" the JSDoc promises.
- **Targeted tests for the new orchestration.** `requireWithTsconfigFallback` and the `handleImport` retry/redispatch path lean on transitive e2e coverage today — a few unit tests for the error branches (`disableTsConfigPaths` re-throw, no-tsconfig re-throw, the `NX_PREFER_NODE_STRIP_TYPES=false` legacy path) would lock them in.
- Minor: stale `// (22.6+)` comment at `register.ts:833`; `tryResolveLoader` / `packageIsInstalled` swallow non-`MODULE_NOT_FOUND` errors via bare `catch {}`.

None of these block — nice work.

FrozenPandaz (approved):
Still LGTM after the latest push. `pick CJS vs ESM emission from project module type, not file shape` is a solid fix — driving the decision off `isEsmProject` (TS-solution / package.json `type`) instead of sniffing the file body closes the real `__filename is not defined` crash, and the shared `module-format` helper extraction is clean. Good test coverage on the new paths.

Earlier non-blocking suggestions still stand for a follow-up (matcher centralization, explicit attempt cap, `exports` maps for the remaining `@nx/js`/`@nx/webpack`/`@nx/vue`).

Stanzilla:
Is there some output/log to tell when it fails or if my repo is successfully using native mode?

Issue #1651, opened 2019-07-28
Tutorial e2e test in Step 2 only working with --prod flag


Halt001:
## Expected Behavior
The e2e test in the tutorial should start up the page (and then fail the tests). See tutorial here: https://nx.dev/angular/tutorial/02-add-e2e-test

## Current Behavior
The e2e test fails to load the page in the `beforeEach(() => cy.visit('/'));` step. An exception is thrown and it looks like the Angular App component is not instantiated/loaded.

When running the e2e test with the production configuration it does work as expected.  `ng e2e todos-e2e --watch --prod` 

The build option '"optimization": true' (which is part of --prod) is what makes the difference.

## Failure Information (for bugs)
The exception output is:

```
Uncaught SyntaxError: Invalid or unexpected token

This error originated from your application code, not from Cypress.

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the 'uncaught:exception' event.

https://on.cypress.io/uncaught-exception-from-application

Because this error occurred during a 'before each' hook we are skipping the remaining tests in the current suite: 'TodoApps'
```

### Steps to Reproduce

1. Follow tutorial step 1: https://nx.dev/angular/tutorial/01-create-application
2. Follow tutorial step 2: https://nx.dev/angular/tutorial/02-add-e2e-test
3. See it fail  when running: ng e2e todos-e2e --watch
4. See it succeed when running ng e2e todos-e2e --watch --prod

### Context
- version of Nx used. 
8.4.0
- version of Angular CLI used
8.1.1
- `angular.json` configuration
As  created by the tutorial

**Edit:**
After looking into the problem some more it seem to come from a problem in vendor.js, probably code coming from @angular/core. There seems to be a problem with unicode characters in a function name.

The offending line is: `const interpolatedValue = ��ɵinterpolation5(prefix, v0, i0, v1, i1, v2, i2, v3, i3, v4, suffix);`

The strange thing is that only in the Cypress browser window this is the case for this line. Opening another Chrome Window on the same server and looking at the same line reveals this: 
`const interpolatedValue = ɵɵinterpolation5(prefix, v0, i0, v1, i1, v2, i2, v3, i3, v4, suffix);` And also loads the page without problem.

To make matters even more confusing, I've tried a non Nx Angular project with Cypress of the same versions and then there is no problem.

**Edit2**:
[...]

PR #35622, opened 2026-05-08
fix(core): read pod cgroup limits instead of node limits in resource metrics

leosvelperez:
## Current Behavior

When running inside a Linux container or Kubernetes pod, resource metrics report the host node's CPU and memory totals instead of the pod's limits. A pod with constrained CPU/memory shows the underlying node's resources.

The same gap exists for process-isolated Windows containers running inside a Windows Job Object (e.g. Docker `--cpus` / `--memory`): metrics report host values rather than the Job's limits.

## Expected Behavior

Resource metrics report the effective CPU and memory limits enforced by the kernel for the calling process — derived from the cgroup it belongs to on Linux, or the Job Object on Windows. macOS native processes continue to report host values (no equivalent enforcement primitive exists).

### Linux (`cgroup` module)

Resolves the calling process's actual cgroup directory by parsing `/proc/self/cgroup` + `/proc/self/mountinfo`. Reads `cpu.max` / `memory.max` (cgroup v2) or `cpu.cfs_{quota,period}_us` / `memory.limit_in_bytes` (cgroup v1, including the systemd `cpu,cpuacct` co-mount).

Walks from the leaf up to the mount point and takes the minimum finite limit found at any level — the kernel enforces the tightest ancestor's limit (hierarchical enforcement), so leaf-only reads can overreport when a pod-level cgroup is tighter than the container's (common in K8s VPA in-place resize and Burstable QoS).

Composition with `sched_getaffinity` covers cpuset / taskset restrictions. We deliberately bypass `std::thread::available_parallelism()` because rust-lang/rust's implementation applies the cgroup quota with floor division internally, which would silently underreport fractional quotas (e.g. 1.5 cores → 1). Instead we ceil quota / period, matching HotSpot JVM, Go 1.25, .NET, and the `num_cpus` crate.

mountinfo path fields containing spaces / tabs / newlines / backslashes are kernel-encoded as `\040` / `\011` / `\012` / `\134` per `man 5 proc_pid_mountinfo`; we unescape before joining with the cgroup path. `/proc/self/cgroup` itself emits paths raw (verified across kernels v4.18 → v6.13), so no unescape is needed there.

Replaces the prior leaf-only path lookups (which broke on cgroup v1 co-mount and any non-namespaced container setup). The in-tree module is preferred over `sysinfo`'s `cgroup_limits()` (now parent-aware in 0.39 — see *Additional Changes*) because it also covers CPU quota and the v1 co-mount + bind-mount edge cases in a single place we control. 30 unit tests cover cgroup discovery, parsing, ancestor walk, and the v1 / v2 / co-mount / bind-mount / cgroupns=host cases.

### Windows (`job_object` module)

Detects Job Object resource limits via the Win32 API:

- **CPU**: `ceil(host_cpu_count × CpuRate / 10000)` from `JobObjectCpuRateControlInformation` when `HARD_CAP` is set (Docker `--cpus` translates to HARD_CAP). Plus the popcount of the Job's affinity mask (when `LIMIT_AFFINITY` is set), and `GetProcessAffinityMask` (covers Job + manual + system intersections). Takes the minimum.
- **Memory**: minimum of `ProcessMemoryLimit` / `JobMemoryLimit` / `MaximumWorkingSetSize` from `JobObjectExtendedLimitInformation`, gated on the corresponding `LIMIT_*` flags. Mirrors HotSpot and .NET.
- **Skipped**: `WEIGHT_BASED` rate control (relative priority, not a hard limit) and soft-cap rate control (kernel allows transient bursts) — neither maps to a defensible "available cores" number.

Any Win32 failure is treated as "no information"; the caller falls back to host values. No new crate dependency — the existing `winapi` dep is extended with `jobapi`, `jobapi2`, `processthreadsapi`, `winbase`, and `winnt` features.

#### Known limitation: nested Job hierarchies

[`QueryInformationJobObject(NULL, ...)`](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-queryinformationjobobject) returns the **immediate** Job's settings (per MSDN: *"If the job is nested, the immediate job of the calling process is used."*) and Win32 exposes no documented API to enumerate parent Jobs.

Per [`JOBOBJECT_CPU_RATE_CONTROL_INFORMATION`](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-jobobject_cpu_rate_control_information) Remarks, *"the rates set for the job represent its portion of the CPU rate that is allocated to its parent job"* — nested rates compose multiplicatively. So with parent HARD_CAP 50% × child HARD_CAP 50%, the effective rate is 25% of host but we read the child's 50% and report `ceil(host × 0.5)`. Per-process / per-job memory has the same shape (kernel enforces min across the chain; we see only the immediate Job).

Affinity is unaffected — `GetProcessAffinityMask` returns the kernel-effective mask. HotSpot and .NET CoreCLR exhibit the same memory limitation; HARD_CAP CPU-rate detection goes beyond them for Docker `--cpus` parity in the common single-silo case, accepting the nested-Job overreport as the documented cost.

### Cross-platform

`SystemInfo { cpuCores, totalMemory }` shape is unchanged — consumers do not need to update. macOS native processes continue to report host values (no container-style enforcement primitive exists; container runtimes on macOS run Linux VMs where the Linux path applies).

The module doc-comments cross-reference Go 1.25 `internal/runtime/cgroup`, libuv `src/unix/linux.c`, OpenJDK `cgroupSubsystem_linux.cpp` + `os_windows.cpp`, .NET CoreCLR `gc/unix/cgroup.cpp` + `gc/windows/gcenv.windows.cpp`, and Rust stdlib `library/std/src/sys/thread/unix.rs::cgroups`.

## Additional Changes

Two infrastructure chores are bundled into this PR as separate commits:

### `chore(core): bump sysinfo to 0.39.1`

Bumps `sysinfo` from `0.37.2` → `0.39.1`. No source changes were required — all signatures we use (`System`, `Process`, `Pid`, `Signal`, `Disks`, the `*RefreshKind` types, `UpdateKind`, `MINIMUM_CPU_UPDATE_INTERVAL`) are unchanged. `Cargo.lock` deltas:

- New transitive dep `objc2-open-directory` from sysinfo 0.39's soundness fix for user retrieval on Apple targets.
- `windows` family bumped to `0.62.x` per sysinfo's new constraint, which lets the graph converge on single versions of `windows-core`, `windows-link`, `windows-result`, and `windows-strings` — four duplicate `windows-*` entries are deduplicated as a result.

sysinfo 0.39.0 also added `Process::cgroup_limits()` and parent-cgroup memory walking inside `System::cgroup_limits()` — overlapping conceptually with the in-tree `cgroup` module introduced by this PR. The in-tree module is kept because it also covers CPU quota (sysinfo's helpers are memory-only), the cgroup v1 co-mount, and bind-mount path translation in a single implementation under our control. Future consolidation onto the upstream APIs is possible but explicitly out of scope here.

### `chore(repo): bump mise rust to 1.95.0 to match rust-toolchain.toml`

`rust-toolchain.toml` was bumped to `1.95.0` in #35665 to unblock the sysinfo upgrade, but `mise.toml` was left at `1.90.0`. CI installs Rust via mise (which exports `RUSTUP_TOOLCHAIN`, overriding `rust-toolchain.toml`), so CI continued to run on `1.90.0` and failed the sysinfo 0.39 MSRV check until this commit. The two files now agree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,7 +202,7 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
@@ -381,7 +381,7 @@ dependencies = [
  "js-sys",
  "num-traits",
  "wasm-bindgen",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
@@ -1166,7 +1166,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1bd49230192a3797a9a4d6abe9b3eed6f7fa4c8a8a4947977c6f80025f92cbd8"
 dependencies = [
  "rustix 1.1.3",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
@@ -1432,7 +1432,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.62.2",
+ "windows-core",
 ]
 
 [[package]]
@@ -1952,7 +1952,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "754ca22de805bb5744484a5b151a9e1a8e837d5dc232c2d7d8c2e3492edc8b60"
 dependencies = [
  "cfg-if",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
@@ -2531,6 +2531,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "objc2-open-directory"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bb82bed227edf5201dfedf072bba4015a33d3d4a98519837295a90f0a23f676d"
+dependencies = [
+ "objc2",
+ "objc2-core-foundation",
+ "objc2-foundation",
+]
+
+[[package]]
 name = "object"
 version = "0.37.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2596,7 +2607,7 @@ dependencies = [
  "libc",
  "redox_syscall 0.5.18",
  "smallvec",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
@@ -4048,15 +4059,16 @@ dependencies = [
 
 [[package]]
 name = "sysinfo"
-version = "0.37.2"
+version = "0.39.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "16607d5caffd1c07ce073528f9ed972d88db15dd44023fa57142963be3feb11f"
+checksum = "a4deba334e1190ba7cb498327affa11e5ece10d26a30ab2f27fcf09504b8d8b6"
 dependencies = [
  "libc",
  "memchr",
  "ntapi",
  "objc2-core-foundation",
  "objc2-io-kit",
+ "objc2-open-directory",
  "windows",
 ]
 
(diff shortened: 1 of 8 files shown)


FrozenPandaz (changes requested):
Heads up: sysinfo v0.39.0 (released 2026-05-11) now does the parent-cgroup memory walking this PR implements — see [PR #1651](https://github.com/GuillaumeGomez/sysinfo/pull/1651) and the v0.39.0 CHANGELOG. nx is pinned to sysinfo 0.37.2; I've opened #35665 to bump the Rust toolchain to 1.95 (sysinfo 0.39's MSRV) so we can pull that in.

Suggestion: rebase onto sysinfo 0.39 and drop the memory-side cgroup parsing in `cgroup.rs`, but keep the genuinely-novel parts — the CPU `cpu.max` / `cpu.cfs_*` ancestor walk, cgroup v1 co-mount handling, mountinfo octal-escape parsing, and the Windows JobObject reads (no upstream equivalent for any of those; [rust-lang/rust#143709](https://github.com/rust-lang/rust/issues/143709) is the open tracking issue for the Windows side).

One concrete bug worth fixing regardless: `read_v1_cpu_quota` only treats `q <= 0` as unlimited, but cgroup v1 emits the same large positive `PAGE_COUNTER_MAX`-style sentinel for CPU that `read_v1_memory_limit` already filters at line 346 — feeding that into `cores_from_quota` will overflow.

leosvelperez:
@FrozenPandaz, the PR has been rebased and now includes the `sysinfo` update. As discussed offline, we'll keep the custom cgroup implementation due to `sysinfo` still lacking some functionality.

FrozenPandaz (approved):

PR #35641, opened 2026-05-11
fix(core): napi generated file refresh

noah-dfns:
## Current Behavior

WASM bindings currently fail silently due to an update to napi-rs without refreshing wasm bindings.

## Expected Behavior

Native client using wasm can be used successfuly.

## Related Issue(s)

Fixes #


AgentEnder:
Welcome, and thanks for catching this. @FrozenPandaz has it. The `Fixes #` line in the description is empty, though. Could you either fill in the issue number or clear the line?

<details>
<summary>

</summary>

> Regenerating napi output normally spans more than one committed file, and this PR changes one. Worth confirming during review that the regenerated set is complete rather than partial, since a missing sibling file would leave the wasm bindings failing the same way.
>
> Nothing has built this. The head commit's first Actions run needed maintainer approval, never got it, and passed GitHub's 30 day cutoff, at which point the pending run is deleted, so there is nothing left to approve and a push is what creates a fresh one. The `SUCCESS` rollup is the `socket-security` dependency scan alone. That matters more than usual here, since a real build is exactly what would catch a partial regeneration.
>
> Kept open despite the missing link, since regenerating bindings that went stale after a napi-rs update is maintenance rather than a new surface.

</details>

PR #35646, opened 2026-05-11
fix(core): drain in-flight notify events in daemon force_flush_pending

FrozenPandaz:
Closes #35630 — supersedes the input-drift snapshot approach with a simpler watcher-side fix.

## Current Behavior

The daemon's `flushPendingWorkspaceChanges` → native `force_flush_pending` drains the watcher's notify channel with `try_recv`. This misses events where the notify-crate thread has read the kernel inotify event but hasn't yet finished sending it on `notify_rx` — a window normally of microseconds but unbounded under scheduling pressure (parallel CI jobs, container cgroups, NFS). When the window hits, the daemon serves a project graph computed against the pre-write state of the workspace. The `spread.test.ts` e2e has been flaking on this race.

## Expected Behavior

**Core fix (Rust):** the force-flush handler now waits up to 5ms via `recv_timeout` for an in-flight notify-thread send to land before draining and snapshotting. The change is scoped to the force-flush handler — the steady-state idle-window flush is unchanged, so there's no latency cost on the normal event flow. `RecvTimeoutError::Disconnected` is now surfaced via the same `fatal` path the main select arm uses, instead of being silently masked as an empty snapshot.

**Regression coverage:**
- Rust unit test (`force_flush_pending_captures_in_flight_writes`) — 20 iterations of write-then-immediate-flush with no sleeps. Passes deterministically with the fix; fails / is racy without it.
- TypeScript integration spec (`project-graph-incremental-recomputation.spec.ts`) — real native `Watcher` against a TempFs workspace, real production routing callback, mutate a project then immediately request the graph. Asserts the daemon returns a graph reflecting the new project (not a stale cached graph).

**Supporting refactors (needed for the integration spec to work cleanly):**
- Extracted `routeWorkspaceChanges` from `server.ts`'s `handleWorkspaceChanges` so the spec can exercise the real production event-routing path without the daemon's inactivity-timer / error-tracking bookkeeping leaking into the test.
- Live-bound `workspaceDataDirectory`, `cacheDir`, `nxProjectGraph`, `nxFileMap`, `nxSourceMaps`, `DAEMON_DIR_FOR_CURRENT_WORKSPACE`, `DAEMON_OUTPUT_LOG_FILE`, and `taskHistoryFile`. They were previously `const` values frozen at module load — fine in production (the daemon has one workspace root for its lifetime), but tests that swapped `workspaceRoot` couldn't update them, so the daemon would write its cache into the real workspace under test. Added an `onWorkspaceRootChanged` subscription in `workspace-root.ts` and converted each derived path to a refresh-on-change `let`. Production behaviour is unchanged — listeners only fire when `setWorkspaceRoot` is called, which never happens in normal daemon startup.

## Related Issue(s)

Closes #35630.

### DIFF
--- a/packages/nx/src/daemon/tmp-dir.ts
+++ b/packages/nx/src/daemon/tmp-dir.ts
@@ -8,17 +8,20 @@ import { join } from 'path';
 import { workspaceDataDirectory } from '../utils/cache-directory';
 import { createHash } from 'crypto';
 import { tmpdir } from 'tmp';
-import { workspaceRoot } from '../utils/workspace-root';
+import { onWorkspaceRootChanged, workspaceRoot } from '../utils/workspace-root';
 
-export const DAEMON_DIR_FOR_CURRENT_WORKSPACE = join(
-  workspaceDataDirectory,
-  'd'
-);
-
-export const DAEMON_OUTPUT_LOG_FILE = join(
+// `let` so tests that re-pin the workspace root see refreshed paths;
+// production sets these once at module load.
+export let DAEMON_DIR_FOR_CURRENT_WORKSPACE = join(workspaceDataDirectory, 'd');
+export let DAEMON_OUTPUT_LOG_FILE = join(
   DAEMON_DIR_FOR_CURRENT_WORKSPACE,
   'daemon.log'
 );
+
+onWorkspaceRootChanged(() => {
+  DAEMON_DIR_FOR_CURRENT_WORKSPACE = join(workspaceDataDirectory, 'd');
+  DAEMON_OUTPUT_LOG_FILE = join(DAEMON_DIR_FOR_CURRENT_WORKSPACE, 'daemon.log');
+});
 
 export const getDaemonSocketDir = () =>
   join(
--- a/packages/nx/src/daemon/tmp-dir.ts
+++ b/packages/nx/src/daemon/tmp-dir.ts
@@ -18,17 +18,10 @@ export let DAEMON_OUTPUT_LOG_FILE = join(
   'daemon.log'
 );
 
-// Guard: jest.mock of workspace-root in unrelated specs replaces the
-// whole module and loses this export. See cache-directory.ts.
-if (typeof onWorkspaceRootChanged === 'function') {
-  onWorkspaceRootChanged(() => {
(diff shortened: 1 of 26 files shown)


packages/nx/src/daemon/tmp-dir.ts
```diff
@@ -8,18 +8,21 @@ import { join } from 'path';
...
+export let DAEMON_DIR_FOR_CURRENT_WORKSPACE = join(workspaceDataDirectory, 'd');
+export let DAEMON_OUTPUT_LOG_FILE = join(
   DAEMON_DIR_FOR_CURRENT_WORKSPACE,
   'daemon.log'
 );
 
+onWorkspaceRootChanged(() => {
+  DAEMON_DIR_FOR_CURRENT_WORKSPACE = join(workspaceDataDirectory, 'd');
+  DAEMON_OUTPUT_LOG_FILE = join(DAEMON_DIR_FOR_CURRENT_WORKSPACE, 'daemon.log');
+});
```
AgentEnder: This is kinda weird test-only code but 🤷‍♂️

packages/nx/src/project-graph/nx-deps-cache.ts
```diff
@@ -37,13 +38,17 @@ export interface FileMapCache {
...
+// tests re-pin the workspace root. Production sets them once.
+export let nxProjectGraph = join(workspaceDataDirectory, 'project-graph.json');
+export let nxFileMap = join(workspaceDataDirectory, 'file-map.json');
+export let nxSourceMaps = join(workspaceDataDirectory, 'source-maps.json');
+
+onWorkspaceRootChanged(() => {
+  nxProjectGraph = join(workspaceDataDirectory, 'project-graph.json');
+  nxFileMap = join(workspaceDataDirectory, 'file-map.json');
+  nxSourceMaps = join(workspaceDataDirectory, 'source-maps.json');
+});
```
AgentEnder: Again, kinda weird test-only code.... I'd also prefer if there was a function to setup the listener + initial value or smth, but I do get that would be kinda weird... you'd need like an obejct to be able to replace the value or smth and that's gross... I dunno

AgentEnder (approved):
Left some feedback on the `onWorkspaceRootChanged` bit, I don't love that... but I get it kinda? Up to you.

PR #35659, opened 2026-05-12
feat(misc)!: drop deprecated webpack plugin re-exports + v23 polish

jaysoo:
## Current Behavior

`@nx/react` and `@nx/webpack` keep deprecated re-exports for `NxReactWebpackPlugin` and `NxTsconfigPathsWebpackPlugin`. Four v23 migration polish items remain from end-to-end testing: devkit allowlist drift (2 missing internal symbols), rollup `rollup-plugin-typescript2` orphan devDep, playwright executor schema missing `x-deprecated`, rolldown rename selector misses string-literal keys.

## Expected Behavior

- Deprecated re-exports removed; new `update-23-0-0` migrations rewrite imports to the sub-paths (`@nx/react/webpack-plugin`, `@nx/webpack/tsconfig-paths-plugin`).
- Devkit `DEVKIT_INTERNAL_SYMBOLS` gains `emitPluginWorkerLog`, `throwForUnsupportedVersion`.
- Rollup migration also strips `rollup-plugin-typescript2` from devDeps.
- Playwright executor schema gains canonical `x-deprecated`.
- Rolldown rename selector matches both Identifier and StringLiteral keys.
- Internal vite/react generators emit `rolldownOptions` instead of `rollupOptions` for new v23 projects.

## Related Issue(s)

NXC-4318

### DIFF
--- a/packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
+++ b/packages/devkit/src/migrations/update-23-0-0/update-deep-imports.ts
@@ -69,6 +69,8 @@ export const DEVKIT_INTERNAL_SYMBOLS: ReadonlySet<string> = new Set([
   'capitalize',
   'classify',
   'dasherize',
+  'emitPluginWorkerLog',
+  'throwForUnsupportedVersion',
 ]);
 
 // Methods on `jest` and `vi` that take a module specifier as their first
--- a/packages/playwright/src/executors/playwright/schema.json
+++ b/packages/playwright/src/executors/playwright/schema.json
@@ -3,6 +3,7 @@
   "version": 2,
   "title": "Playwright executor",
   "description": "Run Playwright tests.",
+  "x-deprecated": "The `@nx/playwright:playwright` executor is deprecated and will be removed in Nx v24. Run `nx g @nx/playwright:convert-to-inferred` to migrate to the `@nx/playwright/plugin` inferred targets. See https://nx.dev/docs/guides/tasks--caching/convert-to-inferred for details.",
   "type": "object",
   "properties": {
     "browser": {
--- a/packages/react/index.ts
+++ b/packages/react/index.ts
@@ -1,8 +1,3 @@
-import { NxReactWebpackPlugin as _NxReactWebpackPlugin } from './plugins/nx-react-webpack-plugin/nx-react-webpack-plugin';
-
-/** @deprecated Use '@nx/react/webpack-plugin' instead, which can improve graph creation by 150-200ms per file. */
-export const NxReactWebpackPlugin = _NxReactWebpackPlugin;
-
 export {
   extraEslintDependencies,
   extendReactEslintJson,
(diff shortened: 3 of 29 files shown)


FrozenPandaz (changes requested):
A few things worth tightening before merge. `@nx/vitest` generators still emit `rollupOptions` (`packages/vitest/src/utils/generator-utils.ts:143,354`, `packages/vitest/src/generators/configuration/configuration.ts:202`) even though vitest pins vite v8, so freshly-generated configs will need the new migration on day one — either rename here too or document why vitest stays on the old key. The `rollup-plugin-typescript2` strip at `remove-use-legacy-typescript-plugin.ts:117` runs unconditionally, so a user who installed it for their own rollup config silently loses it on `nx migrate` — gate it behind "we actually stripped a legacy config." And `rename-rollup-options-to-rolldown-options.ts:22` calls `.includes` on the result of `tree.read` (typed `string | null`) without the `?.` guard the sibling rollup migration uses — would throw mid-walk and silently leave later vite configs on the old key (one-line fix: `if (!contents) return;`).

FrozenPandaz:
The codemods look good overall — one small robustness nit on the multi-specifier branch.

`endAfterTrailingComma` finds the trailing comma by walking only over whitespace. If a comment sits between the deprecated symbol and its comma, the comma isn't found and the rewrite leaves a stray leading comma in the original import:

```ts
// before
import { NxReactWebpackPlugin /* note */, withReact } from '@nx/react';
// after — syntactically invalid
import {  /* note */, withReact } from '@nx/react';
```

This doesn't throw — `formatFiles` catches the prettier parse error per file and only `console.warn`s, so the migration still exits 0 and silently leaves the broken file behind.

Narrow trigger, but easy to make airtight: instead of text-walking for the comma, rebuild the brace group from the specifier nodes you're keeping.

```ts
} else {
  // NamedImports node spans the whole `{ ... }` group.
  const namedImports = targetSpec.parent;
  const remaining = specifiers
    .filter((s) => s !== targetSpec)
    .map((s) => s.getText());
  contents =
    `import { ${targetSpec.getText()} } from '${NEW_PACKAGE}';\n` +
    contents.slice(0, namedImports.getStart()) +
    `{ ${remaining.join(', ')} }` +
    contents.slice(namedImports.getEnd());
  changed = true;
  didRewrite = true;
  continue;
}
```

Symmetric change for the CJS branch (`targetBinding.parent` is the `ObjectBindingPattern`). This deletes `endAfterTrailingComma` entirely and removes the whole class of comma-math bugs. `remaining.join(', ')` produces the commas, so interleaved comments/whitespace/trailing commas stop mattering, and `formatFiles` re-normalizes spacing anyway.

One tradeoff: rebuilding from `.getText()` drops comments sitting *inside* the brace group. For a one-time codemod that's an acceptable cost vs. emitting invalid syntax.

FrozenPandaz (approved):
All three blocking concerns from the previous review are addressed and covered by tests. Remaining note is a non-blocking nit (separate comment) — fine to merge.

PR #35664, opened 2026-05-12
feat(vite)!: deprecate the nxViteTsPaths and nxCopyAssetsPlugin helpers

jaysoo:
## Current Behavior

`nxViteTsPaths` and `nxCopyAssetsPlugin` from `@nx/vite/plugins/*` run silently. New TS-solution workspaces never emit them; new non-ts-solution workspaces still do.

## Expected Behavior

Both helpers log a one-time deprecation warning on call and carry `@deprecated` JSDoc tags. Behavior unchanged in v23; removal candidate for v24.

- TS-solution: helpers are unused (package-manager workspaces handle path resolution; the project root is the publishable folder so no `package.json` copy is needed). Generator output is helper-free and a spec test locks that in.
- Non-ts-solution: workspaces still need both helpers, so generator output is unchanged for now. A spec test locks in the current emit with a TODO(v24) for the swap to `vite-tsconfig-paths`.

Configure-Vite docs lead with `vite-tsconfig-paths` and `publicDir` / `vite-plugin-static-copy` and flag the helpers as deprecated.

## Related Issue(s)

NXC-4316

### DIFF
--- a/astro-docs/src/content/docs/technologies/build-tools/vite/Guides/configure-vite.mdoc
+++ b/astro-docs/src/content/docs/technologies/build-tools/vite/Guides/configure-vite.mdoc
@@ -169,7 +169,7 @@ Add a `vite.config.ts` file to the root of your project. If you are not using Re
 /// <reference types='vitest' />
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
-import { nxViteTsPaths } from '@nx/vite/plugins/nx-tsconfig-paths.plugin';
+import tsconfigPaths from 'vite-tsconfig-paths';
 
 export default defineConfig({
   root: __dirname,
@@ -191,7 +191,7 @@ export default defineConfig({
     host: 'localhost',
   },
 
-  plugins: [react(), nxViteTsPaths()],
+  plugins: [react(), tsconfigPaths()],
 
   test: {
     reporters: ['default'],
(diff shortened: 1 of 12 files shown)


FrozenPandaz (changes requested):
### The deprecation docs contradict the example configs

`astro-docs/.../configure-vite.mdoc` adds two `caution` asides flagging the helpers as deprecated:

- L31–32 — "`nxViteTsPaths` is deprecated"
- L133–134 — "`nxCopyAssetsPlugin` is deprecated"

But further down the **same page**, the "What your `vite.config.ts` looks like" section still demonstrates the deprecated helper in both example configs:

- **For applications** — L172 `import { nxViteTsPaths } from '@nx/vite/plugins/nx-tsconfig-paths.plugin'`, L194 `plugins: [react(), nxViteTsPaths()]`
- **For libraries** — L223 `import { nxViteTsPaths } from '@nx/vite/plugins/nx-tsconfig-paths.plugin'`, L230 `nxViteTsPaths()`

So the page tells readers the helper is deprecated, then hands them a recommended `vite.config.ts` that uses it. The caution asides even say "as shown above" — but the canonical example config below them shows the old way. Please update those two example configs to `tsconfigPaths()` from `vite-tsconfig-paths` so the page is self-consistent.

### Other docs pages still teach the deprecated helper

Beyond `configure-vite.mdoc`, several other guides still show `nxViteTsPaths` as the current way to do things. Worth fixing here or as a tracked follow-up so the deprecation messaging is consistent site-wide:

- `technologies/vue/nuxt/Guides/deploy-nuxt-to-vercel.mdoc` — L46 import, L69 `plugins: [nxViteTsPaths()]`
- `technologies/react/Guides/adding-assets-react.mdoc` — L344 import, L360 `nxViteTsPaths()`
- `technologies/test-tools/storybook/Guides/one-storybook-for-all.mdoc`
- `technologies/test-tools/storybook/Guides/one-storybook-with-composition.mdoc`
- `technologies/typescript/Guides/switch-to-workspaces-project-references.mdoc`

(`guides/Tasks & Caching/convert-to-inferred.mdoc` also references it, but that guide is about migrating off old patterns, so it's likely fine as-is.)

FrozenPandaz (approved):
Docs updates look good — the example configs now use `tsconfigPaths()` from `vite-tsconfig-paths` consistently across the vite, react, vue/nuxt, and storybook guides. The earlier inconsistency is resolved.

PR #35666, opened 2026-05-12
fix(core): throw actionable error when pnpm .modules.yaml is missing

### LINKED ISSUE #35635: pnpm parser crashes with unhelpful error when .modules.yaml is missing (npm ci used instead of pnpm install)
## Current Behavior

When a workspace has `pnpm-workspace.yaml` (so Nx detects pnpm as the package manager), but `npm ci` is used in CI instead of `pnpm install`, Nx crashes with:

```
Error: Could not find ".modules.yaml" at "<repo>/node_modules/.modules.yaml"
    at loadPnpmHoistedDepsDefinition (nx/dist/src/plugins/js/lock-file/utils/pnpm-normalizer.js:40:15)
    at getNodes (nx/dist/src/plugins/js/lock-file/pnpm-parser.js:302:77)
    at getPnpmLockfileNodes (nx/dist/src/plugins/js/lock-file/pnpm-parser.js:34:12)
    at getLockFileNodes (nx/dist/src/plugins/js/lock-file/lock-file.js:53:59)
    at internalCreateNodes (nx/dist/src/plugins/js/index.js:53:79)
```

This happens because `pnpm install` generates `node_modules/.modules.yaml`, but `npm ci` does not.

## Expected Behavior

Nx should either:
1. Detect that `.modules.yaml` is missing and emit a clear, actionable error message (e.g. "pnpm-workspace.yaml found but .modules.yaml is missing — did you run pnpm install?"), or
2. Fall back gracefully without crashing the entire project graph resolution

## Reproduction

1. Create an Nx workspace with `pnpm-workspace.yaml`
2. Run `npm ci` (or `npm install`) instead of `pnpm install`
3. Run any `nx` command (e.g. `nx show projects`)

## Environment

- **Nx version**: 22.7.0
- **Node version**: 22
- **Package manager**: pnpm (workspace), but npm used in CI
- **OS**: Ubuntu (GitHub Actions `ubuntu-latest`)

## Additional Context

The secondary error `/bin/sh: 1: pnpm: not found` confirms pnpm is not installed in the environment. The root crash happens before that, inside the pnpm lock-file parser, making it hard to diagnose without reading the stack trace carefully.

SAY-5:
## Current Behavior

When a workspace has `pnpm-workspace.yaml` but `node_modules` was populated by a different package manager (e.g. `npm ci` in CI), `loadPnpmHoistedDepsDefinition` throws `Could not find ".modules.yaml" at ...`, which crashes project graph creation with an opaque stack trace.

## Expected Behavior

Hoisted dependency definitions are an optional optimization, so the lockfile parser falls back to an empty map and the rest of the graph resolves.

## Related Issue(s)

Fixes #35635

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
+++ b/packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
@@ -47,10 +47,10 @@ export function loadPnpmHoistedDepsDefinition() {
     const { load } = require('@zkochan/js-yaml');
     return load(content)?.hoistedDependencies ?? {};
   }
-  // `.modules.yaml` is only written by `pnpm install`. It can be absent when a
-  // different package manager populated `node_modules` (e.g. `npm ci` in CI),
-  // in which case we still want the rest of the graph to resolve.
-  return {};
+
+  throw new Error(
+    `pnpm lockfile detected, but "${fullPath}" is missing. This usually means that "node_modules" was not installed with pnpm. Run "pnpm install" or use a single package manager for this workspace.`
+  );
 }
 
 /**
--- a/packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
+++ b/packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
@@ -46,9 +46,11 @@ export function loadPnpmHoistedDepsDefinition() {
     const content = readFileSync(fullPath, 'utf-8');
     const { load } = require('@zkochan/js-yaml');
     return load(content)?.hoistedDependencies ?? {};
-  } else {
-    throw new Error(`Could not find ".modules.yaml" at "${fullPath}"`);
   }
+  // `.modules.yaml` is only written by `pnpm install`. It can be absent when a
+  // different package manager populated `node_modules` (e.g. `npm ci` in CI),
+  // in which case we still want the rest of the graph to resolve.
+  return {};
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/nx/src/plugins/js/lock-file/utils/pnpm-normalizer.ts
```diff
@@ -46,9 +46,11 @@ export function loadPnpmHoistedDepsDefinition() {
     const content = readFileSync(fullPath, 'utf-8');
     const { load } = require('@zkochan/js-yaml');
     return load(content)?.hoistedDependencies ?? {};
-  } else {
-    throw new Error(`Could not find ".modules.yaml" at "${fullPath}"`);
   }
+  // `.modules.yaml` is only written by `pnpm install`. It can be absent when a
+  // different package manager populated `node_modules` (e.g. `npm ci` in CI),
+  // in which case we still want the rest of the graph to resolve.
+  return {};
```
leosvelperez: We should handle this with a clearer error instead of silently returning `{}`.

If Nx detects pnpm but `node_modules/.modules.yaml` is missing, the workspace was likely installed with a different package manager. In a valid `pnpm install`, that file should exist. Falling back to an empty hoist map hides the real problem and can leave Nx building a lower-fidelity dependency graph.

Let's change this to throw an actionable error instead. Something along the lines of:

```suggestion

  throw new Error(
    `pnpm lockfile detected, but "${fullPath}" is missing. This usually means that "node_modules" was not installed with pnpm. Run "pnpm install" or use a single package manager for this workspace.`
  );
```

leosvelperez (changes requested):

leosvelperez (approved):
Thanks!

PR #35676, opened 2026-05-13
fix(rspack): multi-version support compliance for @nx/rspack and @nx/rsbuild

FrozenPandaz:
## Current Behavior

`@nx/rspack` and `@nx/rsbuild` do not follow the multi-version support compliance patterns used by other Nx plugins:

- `@rspack/core` is a regular dependency (not a peer) pinned via `catalog:rspack`. `@rsbuild/core` is a regular dependency pinned at exact `1.1.8`. No peerDependencies range, no per-major version map.
- The init generators unconditionally overwrite the workspace's installed `@rspack/core` / `@rsbuild/core` with the plugin's pinned version when `keepExistingVersions=false`, which is also the schema default.
- Most generators do not assert a supported floor before mutating the workspace, so sub-floor installs silently land in an unsupported state.
- `@nx/rspack` declares `@module-federation/enhanced` and `@module-federation/node` as peer dependencies while `@nx/module-federation` declares them as regular dependencies — inconsistent across the two plugins.
- `packages/rspack/migrations.json` 21.3.0 bumps `@module-federation/enhanced` cross-minor in the 0.x range without a `requires` clause. The 22.2.0 entry bumps `@module-federation/{enhanced,sdk,runtime}` in a single grouped entry with no per-package gates.
- `packages/rsbuild/src/utils/versions.ts` (`1.1.10`) is out of sync with `packages/rsbuild/package.json` (`1.1.8`).
- Supported-versions docs pages list only a default install version, not the actual support window.

## Expected Behavior

Both plugins follow the multi-version compliance patterns, with the cypress-style assert-at-entry approach:

- `@rspack/core`, `@rspack/dev-server`, `@rspack/plugin-react-refresh` are peerDependencies of `@nx/rspack` with range `^1.0.0`. `@rsbuild/core` is a peerDependency of `@nx/rsbuild` with range `^1.0.0`.
- `@module-federation/enhanced` and `@module-federation/node` are declared as regular dependencies in `@nx/rspack` to match `@nx/module-federation`.
- New Angular-style `backwardCompatibleVersions` maps (`packages/{rspack,rsbuild}/src/utils/versions.ts`) and detection utilities (`version-utils.ts`) gate installed-version pickup, with tree-based and runtime variants for generator-time and executor-time callers.
- New `assertSupported{Rspack,Rsbuild}Version(tree)` wrappers around the shared `assertSupportedPackageVersion` helper from `@nx/devkit/internal`. Called at the top of **every** generator entry — `init`, `configuration`, `convert-webpack`, `convert-to-inferred`, and `convert-config-to-rspack-plugin` for rspack; `init` and `configuration` for rsbuild — so sub-floor workspaces fail fast with a consistent message before any mutation.
- Above-window installs are not thrown on. Following the pattern used by `@nx/angular` and `@nx/playwright`, an unknown future major falls through silently to the latest known install constants. The workspace already has its own pin and we honor it.
- Init generator schemas flip `keepExistingVersions` default from `false` to `true`. Init reads `options.keepExistingVersions ?? true`. All `addDependenciesToPackageJson` callsites pass `keepExistingVersions: true` (or thread the option through where the schema exposes it).
- Migration `requires` gates added to `@nx/rspack`:
  - `21.3.0` → `requires: { "@module-federation/enhanced": ">=0.15.0 <0.17.0" }` for the enhanced bump; `@module-federation/node` bump split out with its own `requires` clause.
  - `22.2.0` split into per-package entries for `enhanced`, `sdk`, `runtime`, and `node`, each with its own `requires` clause.
- rsbuild `versions.ts` / `package.json` drift resolved — both now flow through the new map.
- Exact version pins (`1.6.8`, not `^1.6.8`) for the rspack/rsbuild family entries in the version map. With caret ranges, `@rspack/cli` floats to the latest 1.x (e.g. `1.7.11`) while `@rspack/core` stays at `1.6.8` (pinned by `@nx/module-federation`'s transitive dep via the catalog), causing a cli/core API skew that crashes `rspack serve` with `Cannot read properties of undefined (reading 'web')`. Users who upgrade rspack/rsbuild independently still work because `getInstalled*MajorVersion` reads their pin and init honors it.
- Docs pages updated with explicit `^1.0.0` support window and a `Default Installed` column.

### Tests

A parameterized `all-generators-enforce-floor.spec.ts` in each package uses `assertGeneratorsEnforceVersionFloor` from `@nx/devkit/internal-testing-utils` to verify every generator entry throws on sub-floor installs. Catches future regressions where a new generator forgets to call the assert.

### Scope notes

- **v0 dropped.** Plugins had no v0 support to preserve.
- **v2 in a follow-up PR.** `@rspack/core@2.0.3` and `@rsbuild/core@2.0.6` ship as pure ESM and require additional work for the Nx plugins (CJS) to interop. The version-map and detection scaffolding here is structured so the v2 entries slot in trivially once that work lands.

### Validation

- `nx run-many -t test,build,lint -p rspack,rsbuild` — green
- `nx run module-federation:test` — green
- `nx affected -t lint --base=origin/master` — green
- Floor-enforce specs pass: 6 generators (rspack) + 2 generators (rsbuild) verified to throw on sub-floor `~0.7.0`.

## Related Issue(s)

Fixes #NXC-4404
Fixes #NXC-4405


leosvelperez (approved):

astro-docs/src/content/docs/technologies/build-tools/rsbuild/introduction.mdoc
```diff
@@ -13,11 +13,15 @@ The Nx Plugin for Rsbuild contains executors and generators that support buildin
...
-| `@rsbuild/core` | ^1.1.0             |
+| Package         | Supported Versions | Default Installed |
+| --------------- | ------------------ | ----------------- |
+| `@rsbuild/core` | `^1.0.0`           | `^1.1.10`         |
 
-[Nx generators](/docs/features/generate-code) install the latest supported versions automatically when scaffolding new projects.
+[Nx generators](/docs/features/generate-code) install the latest supported version automatically when scaffolding new projects. When `@rsbuild/core` is already installed in your workspace, generators detect the installed version and keep it in place rather than overwriting it.
+
+{% aside type="caution" title="@rsbuild/core v2 not yet supported" %}
+`@rsbuild/core@2` ships as a pure-ESM package, which `@nx/rsbuild` (CommonJS) cannot currently load synchronously. Support is tracked separately; until then, pin `@rsbuild/core` to `^1`. Detection runs at generator time and will surface a clear error if a v2 install is encountered.
```
leosvelperez: We don't validate versions above the latest supported version:
```suggestion
`@rsbuild/core@2` ships as a pure-ESM package, which `@nx/rsbuild` (CommonJS) cannot currently load synchronously. Support is tracked separately; until then, pin `@rsbuild/core` to `^1`.
```

astro-docs/src/content/docs/technologies/build-tools/rspack/introduction.mdoc
```diff
@@ -13,11 +13,15 @@ The Nx Plugin for Rspack contains executors, generators, and utilities for manag
...
-| `@rspack/core` | ^1.6.0             |
+| Package        | Supported Versions | Default Installed |
+| -------------- | ------------------ | ----------------- |
+| `@rspack/core` | `^1.0.0`           | `^1.6.8`          |
 
-[Nx generators](/docs/features/generate-code) install the latest supported versions automatically when scaffolding new projects.
+[Nx generators](/docs/features/generate-code) install the latest supported version automatically when scaffolding new projects. When `@rspack/core` is already installed in your workspace, generators detect the installed version and keep it in place rather than overwriting it.
+
+{% aside type="caution" title="@rspack/core v2 not yet supported" %}
+`@rspack/core@2` ships as a pure-ESM package, which `@nx/rspack` (CommonJS) cannot currently load synchronously. Support is tracked separately; until then, pin `@rspack/core` to `^1`. Detection runs at generator time and will surface a clear error if a v2 install is encountered.
```
leosvelperez: We don't validate versions above the latest supported version:
```suggestion
`@rspack/core@2` ships as a pure-ESM package, which `@nx/rspack` (CommonJS) cannot currently load synchronously. Support is tracked separately; until then, pin `@rspack/core` to `^1`.
```

PR #35686, opened 2026-05-14
fix(misc): stop inferring `projects: 'self'` in `dependsOn` entries

leosvelperez:
## Current Behavior

The `@nx/cypress`, `@nx/jest`, `@nx/playwright` and `@nx/gradle` plugins infer atomized CI `dependsOn` entries that include `projects: 'self'`. That value is the default for `projects` and is no longer supported as an explicit value.

## Expected Behavior

The plugins infer the same atomized CI `dependsOn` entries without setting `projects: 'self'`.

### DIFF
--- a/packages/cypress/src/plugins/plugin.spec.ts
+++ b/packages/cypress/src/plugins/plugin.spec.ts
@@ -368,7 +368,6 @@ describe('@nx/cypress/plugin', () => {
                       {
                         "options": "forward",
                         "params": "forward",
-                        "projects": "self",
                         "target": "e2e-ci--src/test.cy.ts",
                       },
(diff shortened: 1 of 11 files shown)


FrozenPandaz (approved):

AgentEnder:
> That value is the default for projects and is no longer supported as an explicit value.

Just minor clarification, the value isn't the default for projects, its that `projects` explicitly means "filter to projects matching these patterns", while the lack of `projects` and lack of `dependencies` means "target on same project". Its not a default value, it was just a legacy syntax that we dropped support for and failed to note that some of our own plugins still inferred it.

PR #35694, opened 2026-05-15
feat(release)!: drop deprecated releaseTag* flat properties and update v23 defaults

AI-JamesHenry:
> **Major / breaking change** — this PR is part of the Nx v23 cleanup. The `releaseTag*` flat properties are removed from the public type and config surface, and a long-deprecated default is flipped. The 22.0.0 consolidate migration is re-registered for v23 so any workspaces that still carry the legacy keys are auto-migrated on `nx migrate`.

## Current Behavior

- `release.releaseTagPattern`, `release.releaseTagPatternCheckAllBranchesWhen`, `release.releaseTagPatternRequireSemver`, `release.releaseTagPatternPreferDockerVersion`, and `release.releaseTagPatternStrictPreid` exist on `NxReleaseConfiguration` (top level and per group) with `@deprecated` JSDoc, and `nx release` falls back to them when the nested `releaseTag.*` keys are absent.
- `version.adjustSemverBumpsForZeroMajorVersion` defaults to `false`, so `0.x` releases bump like `1.x` (breaking → major, feature → minor) by default. The JSON schema also documents the default as `false`.
- `releaseTag.strictPreid` is documented (both via JSDoc on `NxReleaseConfiguration` and the nx.json JSON schema) as defaulting to `false`. The runtime default has been `true` since v22.
- `version.preserveMatchingDependencyRanges` carries a stale `TODO(v22)` marker and JSDoc that says it defaults to `false`. The runtime default has been `true` since v22, and the duplicated schema definition still claimed `false`.

## Expected Behavior

- The deprecated `releaseTag*` flat properties are removed from `NxReleaseConfiguration`, `packages/nx/schemas/nx-schema.json`, the release config reconciliation in `packages/nx/src/command-line/release/config/config.ts`, and from the docs. Users on `nx migrate` get the 22.0.0 consolidation re-run via the new `23-0-0-consolidate-release-tag-config` migration to fold any straggling keys into the nested `releaseTag` object.
- `version.adjustSemverBumpsForZeroMajorVersion` now defaults to `true`, including in both JSON schema entries. For `0.x` projects, `major` becomes `minor`, `minor` becomes `patch`, and the prerelease equivalents follow suit, matching SemVer V2 semantics. Set the option to `false` to opt out.
- `releaseTag.strictPreid` is documented as defaulting to `true`. JSDoc, schema description, and the docs reference table are updated to match the actual runtime behaviour. Set to `false` to always resolve the latest matching tag regardless of preid.
- `version.preserveMatchingDependencyRanges` has its stale `TODO(v22)` marker removed, its JSDoc updated to "true by default", and the second JSON schema entry corrected from `false` to `true`. No runtime change — this is documentation alignment.
- The user-facing CLI message for resolved versions now reads "based on releaseTag.pattern" instead of "based on releaseTagPattern".

## Related Issue(s)

N/A — this PR is part of the Nx v23 internal cleanup and is not tied to a tracking issue.

### DIFF
--- a/astro-docs/src/content/docs/reference/nx-json.mdoc
+++ b/astro-docs/src/content/docs/reference/nx-json.mdoc
@@ -455,12 +455,8 @@ The `projectsRelationship` property tells Nx whether to release projects indepen
 
 The `releaseTag` property controls how git tags are parsed and generated. By default tags conform to semantic version and are validated as such.
 
-{% aside type="note" title="Nx 22 Changes" %}
-Release tag configuration now uses a nested `releaseTag` object. Old flat properties (`releaseTagPattern`, etc.) are deprecated but work until Nx 23.
-
-**Nx 22+:** `releaseTag.pattern`, `releaseTag.requireSemver`, `releaseTag.strictPreid`
-
-**Nx < 22:** `releaseTagPattern`, `releaseTagPatternRequireSemver`, `releaseTagPatternStrictPreid`
+{% aside type="note" title="Nx 23 Changes" %}
+The legacy flat properties (`releaseTagPattern`, `releaseTagPatternRequireSemver`, `releaseTagPatternStrictPreid`, `releaseTagPatternPreferDockerVersion`, `releaseTagPatternCheckAllBranchesWhen`) were removed in Nx 23. Use the nested `releaseTag` object below. An automated migration in Nx 22 moved configuration to the new shape.
 {% /aside %}
 
 #### Configuration options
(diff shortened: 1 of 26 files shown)


FrozenPandaz (approved):
Nice turnaround on the e2e fallout from the `adjustSemverBumpsForZeroMajorVersion` flip — all the affected files are opted out now, including `conventional-commits-config.workspaces.test.ts`. The `detectLegacyReleaseTagProperties` error path is a good addition too: it closes the gap where a manual (non-`nx migrate`) upgrader would have had a stale flat `releaseTagPattern` silently ignored, and it reuses the existing `CreateNxReleaseConfigError` infrastructure cleanly. Good that it scans groups as well as the top level and ships with a spec.

Approving — the two items below are non-blocking, would be good to land before merge:

1. **The v23 migration has no spec test.** `update-23-0-0/consolidate-release-tag-config.ts` is a thin re-export of the v22 migration with no `.spec.ts`. A future rename/break of the v22 file would ship silently. A minimal test (workspace with a legacy `releaseTagPattern` → assert it lands in `releaseTag.pattern`) pins the delegation.

2. **Stale `releaseTagPattern` strings in `nx-schema.json` descriptions.** Lines 228, 334, and 348 still reference the removed flat property in their `description` text — line 334 is the description of the new `releaseTag.pattern` field itself. Schema validation is fine; this is just editor-tooltip docs that now disagree with the updated TS JSDoc. Find-and-replace to `releaseTag.pattern`.

Minor: the `Fixes #` line in the PR body is empty — drop it or point it at the tracking issue.

PR #35699, opened 2026-05-15
feat(core): rename nx watch --includeDependentProjects to --includeDependencies

AI-JamesHenry:
## Current Behavior

The `nx watch` command exposes a flag called `--includeDependentProjects`. The name is misleading: despite suggesting *dependents*, the flag actually includes the watched project's *dependencies*. This has been a source of confusion for people and agents.

## Expected Behavior

The flag is renamed to `--includeDependencies` (alias `-d`). The new flag is **functionally identical** — only the name changed.

The legacy `--includeDependentProjects` flag is kept as a hidden, deprecated alias for the Nx 23 deprecation window. When used, it forwards its value through to `--includeDependencies` and prints a deprecation warning that explains the rename and confirms the new flag is functionally identical. `TODO(v24)` comments mark every back-compat code path; the alias will be removed in Nx 24.

Internal call sites that pass the flag programmatically (`@nx/js` tsc batch watcher, `@nx/js` node executor, `@nx/web` file-server) have been updated to the new name, along with the generated `watch-deps` command in the `@nx/js` typescript plugin, the `nx watch` CLI example, the workspace-watching docs page, and every plugin snapshot that asserts on the generated command.

## Related Issue(s)

N/A — internal naming cleanup ahead of Nx 24.

### DIFF
--- a/packages/nx/src/command-line/watch/command-object.ts
+++ b/packages/nx/src/command-line/watch/command-object.ts
@@ -16,62 +16,90 @@ export const yargsWatchCommand: CommandModule = {
 };
 
 function withWatchOptions(yargs: Argv) {
-  return withVerbose(yargs)
-    .parserConfiguration({
-      'strip-dashed': true,
-      'populate--': true,
-    })
-    .option('projects', {
-      type: 'string',
-      alias: 'p',
-      coerce: parseCSV,
-      description: 'Projects to watch (comma/space delimited).',
-    })
-    .option('all', {
-      type: 'boolean',
-      description: 'Watch all projects.',
-    })
-    .option('includeDependentProjects', {
-      type: 'boolean',
-      description:
-        'When watching selected projects, include dependent projects as well.',
-      alias: 'd',
-    })
-    .option('includeGlobalWorkspaceFiles', {
-      type: 'boolean',
-      description:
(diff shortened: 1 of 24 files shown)


packages/nx/src/command-line/watch/command-object.ts
```diff
@@ -16,62 +16,90 @@ export const yargsWatchCommand: CommandModule = {
...
+        // TODO(v24): remove the legacy includeDependentProjects pass-through
+        const a = args as Record<string, unknown>;
+        if (
+          a.includeDependentProjects !== undefined &&
+          a.includeDependencies === undefined
+        ) {
+          a.includeDependencies = a.includeDependentProjects;
+          // eslint-disable-next-line no-console
+          console.warn(
+            "--includeDependentProjects was poorly named (it includes the watched project's dependencies, not its dependents) and has been renamed to --includeDependencies. The new flag is functionally identical. --includeDependentProjects will be removed in Nx 24."
```
jaysoo: Since this warning is user facing, maybe "poorly" is the wrong word? Something like `--includeDependentProjects is deprecated in favor of --includeDependencies and will be removed in Nx 24. This flag name  is misleading since it includes the watched project's dependencies, not its own dependents`.

jaysoo (approved):

Issue #34654, opened 2026-03-01
nx terminal freezes


golan132:
### Current Behavior

when running app with the serve commend the terminal prints few logs then freeze, only when i control c the terminal it prints the rest of the logs.

### Expected Behavior

prints the logs as the program runs in the terminal.

### Steps to Reproduce

1. created a ts next project
2. run the project with the serve commend - "npx nx run app-service:serve"

### Nx Report

```shell
PS C:\Users\morad\Desktop\Projects\telepoly> npx nx report

 NX   Report complete - copy this into the issue template

Node           : 25.5.0
OS             : win32-x64
Native Target  : x86_64-windows
npm            : 11.8.0
daemon         : Available

nx             : 22.5.2
@nx/js         : 22.5.2
@nx/eslint     : 22.5.2
@nx/workspace  : 22.5.2
@nx/jest       : 22.5.2
@nx/devkit     : 22.5.2
@nx/nest       : 22.5.2
@nx/node       : 22.5.2
@nx/web        : 22.5.2
@nx/webpack    : 22.5.2
@nx/docker     : 22.5.2
typescript     : 5.9.3
---------------------------------------
Registered Plugins:
@nx/webpack/plugin
---------------------------------------
Cache Usage: 8.59 MB / 93.03 GB
```

### Failure Logs

```shell
no logs
```

### Package Manager Version

npm            : 11.8.0

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

FrozenPandaz:
Thanks for reporting this! To help us investigate, we need more details:

1. **Reproduction repository** — A minimal repo that demonstrates the freeze (ideally just `create-nx-workspace` + the minimum steps to trigger it)
2. **What logs appear** — What output do you see before the freeze? A screenshot or copy-paste would help
3. **Terminal emulator** — Which terminal are you using? (PowerShell, Windows Terminal, CMD, Git Bash, etc.)
4. **Project type** — You mention "ts next project" but have `@nx/nest` installed — is this a NestJS app or a Next.js app?
5. **Node version** — You're on Node 25.5.0 which is very new. Does the issue also occur on Node 22 LTS?
6. **TUI vs legacy** — Does running with `NX_TUI=false npx nx run app-service:serve` make a difference?

Without a reproduction, it's difficult to diagnose what's causing the freeze. Once we have that, we'll prioritize this accordingly.

michael-golden:
[...]

PR #35716, opened 2026-05-17
fix(core): don't inherit stdin into child tasks when TUI is enabled o…

PieterjanDeClippel:
## Current Behavior

On Windows, when the Nx TUI is enabled (`tui.enabled = true` in `nx.json`), child task processes inherit the parent's console stdin because `PseudoTerminal.isSupported()` returns `false` on Windows by default — the existing gate at `pseudo-terminal.ts:253-258` requires `NX_WINDOWS_PTY_SUPPORT='true'` to opt in, citing #22358.

The Rust TUI reads the same console input handle via `crossterm::EventStream` → `ReadConsoleInputW`. The Windows console driver routes each key event to exactly one reader, so as soon as a child opens its stdin reader (Angular dev-server's `r`/`q`/`u` prompts, `dotnet watch`'s `Ctrl+R`, MCP STDIO servers, etc.) it starts intercepting keystrokes destined for the TUI — observed as the TUI becoming completely unresponsive to `q`, `Ctrl+C`, `Esc`, and arrow keys once the served process actually starts running.

The bug has been independently reported in #33720 (Angular `nx serve`, multiple users) and #34654 (Nest + `@rekog/mcp-nest`, traced to MCP STDIO transport — switching to SSE eliminates the freeze, which is a direct proof that child-stdin-read is the trigger).

## Expected Behavior

Two complementary changes:

1. **Re-enable Windows ConPTY by default** in `pseudo-terminal.ts`. The original blocker #22358 (`^[[12;1R` cursor-position-query echoed) is **closed and stale** — it was filed against Nx 18.x with a completely different terminal stack (older `crossterm`, no `ratatui` 0.30, different `portable-pty` path). The class of bug it described is the kind terminal libraries typically resolve over time. The gate's net cost (a frozen TUI on every Windows machine running `nx serve`) is now larger than the residual risk it guards against. The opt-out is now `NX_WINDOWS_PTY_SUPPORT='false'` instead of the previous opt-in, so users who do hit a regression can revert behavior without downgrading.

2. **Defence-in-depth**: when the TUI is enabled on Windows and we still fall through to the no-PTY fork path (e.g. user explicitly set `NX_WINDOWS_PTY_SUPPORT='false'`, or some future incompatibility), use `stdin: 'ignore'` instead of `stdin: 'inherit'` so the child cannot read the console at all. This is centralised in a new `childStdinMode()` helper and applied to four fork sites (batch worker, prefixed/no-TTY, direct-output, and `RunningNodeProcess` in `run-commands`). Children lose direct stdin prompts in that fallback — acceptable because no TUI-mediated input path exists without ConPTY anyway (no PTY = no `i` interactive mode = no key forwarding).

Verified workaround on Windows 11 Pro 10.0.26200 + PowerShell 7 + an Angular dev-server + `dotnet watch` workspace: `NX_WINDOWS_PTY_SUPPORT=true` on the existing release eliminates the freeze entirely (`q`, `Ctrl+C`, arrows, `Tab` all respond normally). This PR makes that the default.

### Test plan

- [x] New unit spec `child-stdin-mode.spec.ts` covers Linux/macOS/Windows × TUI on/off matrix (4/4 pass).
- [x] `is-tui-enabled.spec.ts` unchanged, still 17/17 passing.
- [x] `tasks-schedule.spec.ts` unchanged, still 34/34 passing.
- [x] `tsc --noEmit` clean on `packages/nx/tsconfig.lib.json`.
- [x] Real-world verification by the reporter via `NX_WINDOWS_PTY_SUPPORT=true` (same code path the new default takes) — confirmed in https://github.com/nrwl/nx/issues/33720#issuecomment-4470247204.

## Related Issue(s)

- Fixes #33720
- Refs #34654

### DIFF
--- a/docs/prd/issue-33720-PRD.md
+++ b/docs/prd/issue-33720-PRD.md
@@ -1,0 +1,353 @@
+# PRD — Nx TUI keyboard input freezes on Windows once continuous tasks start
+
+> **Status:** Revised after the reporter's diagnostics ruled out the initial
+> "stdin not a TTY" hypothesis. New root cause identified by code trace and
+> cross-referenced against open Nx issues #33720 and #34654.
+>
+> **Linked open issues:**
+> - <https://github.com/nrwl/nx/issues/33720> — "Terminal UI freezes when using nx run serve" (Windows, multiple reporters)
+> - <https://github.com/nrwl/nx/issues/34654> — "nx terminal freezes" (Windows, traced to a child process that reads stdin)
+
+## Problem statement
+
+When the Nx Terminal UI is enabled (`nx.json` → `"tui": { "enabled": true }`)
+on Windows and the user runs a continuous task (`nx serve`, `npm start` →
+`nx serve`, `npx nx serve` — all reproduce identically), the TUI starts up
+correctly and keyboard input works **briefly** during the task-start phase
+(arrow keys navigate, `Tab` switches panes). **The moment the served child
+process actually begins running, the TUI becomes unresponsive** to every
+key: `q`, `Esc`, `Ctrl+C`, `Ctrl+Z`, arrows, `Tab`, `?`. The only recovery
+is closing the terminal window — Nx's process-tree teardown then propagates
+shutdown to the dotnet and node children correctly (the reporter's
+`tools/scripts/serve-api.mjs` handles `taskkill /T /F` on
+`SIGINT`/`SIGBREAK`/`SIGHUP`), so port 5000 is released, but the exit path
+is unacceptably blunt.
+
+## Reproduction
+
+| Field | Value |
+|---|---|
+| OS | Windows 11 Pro 10.0.26200 |
+| Shell | PowerShell 7+ |
+| Nx version | 22.7.2 |
+| Workspace | `C:\Repos\mintplayer-ng-bootstrap` |
+| `nx.json` | `"tui": { "enabled": true }`, `"useDaemonProcess": false` |
+| Invocation | `npm start`, `nx serve`, and `npx nx serve` all reproduce |
+| Continuous tasks | `ng-bootstrap-demo:serve` (Angular dev-server executor) and `api:serve` (`nx:run-commands` → `node tools/scripts/serve-api.mjs` → `dotnet watch`) |
+
+Reporter's diagnostics (run 2026-05-17):
+
+```text
+PS> node -e "console.log({stdin: process.stdin.isTTY, stderr: process.stderr.isTTY})"
+{ stdin: true, stderr: true }
+```
+
+Both `stdin.isTTY` and `stderr.isTTY` are `true` in the reporter's PowerShell
+session — ruling out the original capability-check hypothesis.
+
+Additional public reproductions on similar Windows configurations confirm
+the symptom independent of the reporter's workspace:
+- #33720 OP: nx 22.1.3, Windows, Angular serve.
+- #33720 NoNamer777: nx (recent), Windows, `pnpm nx serve <app>`, multiple
+  terminal hosts (Windows Terminal, VS Code integrated, Webstorm), multiple
+  shells (PowerShell, cmd, Git Bash) — **all behave identically**.
+- #34654 michael-golden: nx 22.6.3, Windows, Nest serve with
+  `@rekog/mcp-nest`. Trigger isolated to the child app using **MCP STDIO
+  transport** (`process.stdin` reads). Switching the same project to MCP SSE
+  transport (no stdin reads) **makes the freeze go away**.
+
+The MCP-STDIO datapoint is decisive — it shows the freeze is triggered by
+**the child process reading from stdin**, not by anything Nx renders or any
+keyboard the user presses.
+
+## Root cause (high confidence)
+
+**Two processes are racing for the same Windows console input stream**:
+the Nx parent (running the TUI via Rust + crossterm `EventStream`, ultimately
+`ReadConsoleInputW` against the parent's stdin handle) and the **forked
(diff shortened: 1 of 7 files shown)


PieterjanDeClippel:
## How to test this PR locally on Windows

The fix targets a real Windows-only freeze, so verification means actually consuming patched `nx` from a Windows workspace. The Nx repo has a built-in local Verdaccio + release script that makes this turnkey.

### Prerequisites

1. **Rust toolchain** (Nx builds a native addon). One-time install from <https://rustup.rs> (Windows installer is `rustup-init.exe`). Verify:
   ```powershell
   cargo --version
   ```
2. **pnpm**: `npm i -g pnpm` if not already installed.
3. Check out the PR branch:
   ```powershell
   git clone https://github.com/PieterjanDeClippel/nx.git
   cd nx
   git checkout fix/tui-input-windows-stdin-race
   ```

### Pick a preview version

Use any prerelease string. I'll use `22.8.0-preview.0` below — substitute whatever you like (e.g. `22.8.0-preview-pr-35716.0`).

### Two-terminal workflow

**Terminal 1 — start Verdaccio (leave running):**

```powershell
npm run local-registry
# Verdaccio comes up on http://localhost:4873
# Your npm config registry is temporarily pointed there (user-level)
# Ctrl+C when done — it restores your prior registry automatically
```

**Terminal 2 — build + version + publish in one command:**

```powershell
npm run nx-release -- 22.8.0-preview.0
```

What it does:
1. Builds every public package (slow first time — Rust compile + all TS builds; cold runs ~10–20 minutes, cached runs are minutes).
2. Stamps `22.8.0-preview.0` into every `dist/packages/<name>/package.json` and rewrites inter-package dep ranges.
3. Publishes to `http://localhost:4873` in parallel.

`--local` is the default for this script, so there's no risk of accidentally publishing to npm. Source `package.json` files are kept at `0.0.1` throughout — the version only lands in the `dist/` outputs.

**Verify the publish:**

```powershell
curl http://localhost:4873/nx
# or open http://localhost:4873 in a browser
```

### Consume from a test workspace

In whatever workspace reproduces the freeze (Angular `nx serve`, Nest + MCP-STDIO, etc.):

```powershell
$env:npm_config_registry = 'http://localhost:4873/'

# Pin every nx-* dependency to the preview version
npm install --save-exact `
  nx@22.8.0-preview.0 `
  @nx/angular@22.8.0-preview.0 `
  @nx/eslint@22.8.0-preview.0 `
  @nx/eslint-plugin@22.8.0-preview.0 `
  @nx/js@22.8.0-preview.0 `
  @nx/workspace@22.8.0-preview.0
# ...add any other @nx/* packages your workspace uses

# Critical: make sure the workaround env var is NOT set
Remove-Item Env:NX_WINDOWS_PTY_SUPPORT -ErrorAction SilentlyContinue

# Reproduce the original failure mode
npm start  # or: nx serve, npx nx serve, etc.
```

**Expected:** `q`, `Ctrl+C`, `Esc`, arrow keys, `Tab` all respond cleanly while the served processes are running. This is the same end state as the existing `NX_WINDOWS_PTY_SUPPORT=true` workaround (verified on Windows 11 Pro 10.0.26200 + PowerShell 7 + Angular dev-server + `dotnet watch` in <https://github.com/nrwl/nx/issues/33720#issuecomment-4470247204>), now reached without needing the env var.

### Tear down when done

```powershell
# Terminal 1: Ctrl+C — restores your npm config registry

# Test workspace:
Remove-Item Env:npm_config_registry -ErrorAction SilentlyContinue
git checkout -- package.json package-lock.json
npm ci
```

### Notes / gotchas

- **First build is slow.** Cold Rust + ng-packagr + a couple hundred TS projects. Subsequent runs hit Nx cache.
- **`nx-release` clears the local registry by default.** Each run wipes prior versions in `dist/local-registry/storage`. To keep multiple previews simultaneously, pass `--clearLocalRegistry false`:
  ```powershell
  npm run nx-release -- 22.8.0-preview.1 --clearLocalRegistry false
  ```
- **PowerShell quoting:** backticks above are PowerShell line-continuations. In `cmd.exe` use `^` or put everything on one line.
- **Negative-control check:** to confirm the fix is actually load-bearing, after the test workspace works, you can also try setting `$env:NX_WINDOWS_PTY_SUPPORT='false'` and re-running — the **defence-in-depth** part of this PR (`stdin: 'ignore'` in the no-PTY fork path) should still keep `q` / `Ctrl+C` responsive, whereas the unpatched build with that env var set would freeze.

PieterjanDeClippel:
On your host device:

    npm run local-registry
    $env:npm_config_registry = 'http://localhost:4873/'

In the vscode devcontainer

    export npm_config_registry="http://localhost:4873/"
    npm run nx-release --22.8.0-preview-pr-35716.1

AgentEnder:
Thanks for the write-up. Two files in the diff should not ship: `docs/prd/issue-33720-PRD.md` and `docs/prd/issue-33720-plan.md` are planning notes, 701 lines of them, and `docs/prd/` is not a directory this repo has. Please drop them. I have moved this to draft so it is clear the ball is with you rather than with a reviewer, and marking it ready again is all it takes.

Worth flagging for whoever picks up the review: this is really two changes rather than one, the `stdin: 'ignore'` fallback and re-enabling Windows ConPTY by default, and the second is the larger call of the two.

<details>
<summary>

</summary>

> `git ls-tree -d HEAD docs/prd` returns nothing on `master`, so the PR creates that directory as well as the two files in it.
>
> The CI approval on the head commit expired before anyone ran it, so there is nothing left to approve. A push after dropping those files creates a fresh run.

</details>

PR #35739, opened 2026-05-19
feat(core) remote cache connection pool toggle

dan-winters:
## Current Behavior
NX remote cache leverages connection pooling via its underlying library reqwest. In certain environments (such as AWS), load balancers or proxies can improperly disconnect connections without sending any notification.  When this happens, we'll see an error for remote caching.

## Expected Behavior
Allow an environment variable to disable connection pooling. This allows us to operate in spaces where we don't have the means to fix interactions between network layers.

## Related Issue(s)

Fixes #


AgentEnder:
Thanks for the PR. Proxies and load balancers dropping pooled connections is a real problem, and you've done the docs and e2e work as well as the change itself.

What I'm closing on isn't the diff. It's that this adds a public configuration surface with no issue behind it, so nobody on the team has agreed that a user-facing switch is the right answer. Once an env var like that ships we own the name, the docs, and its interaction with every later change to the cache client, and that's worth deciding deliberately rather than as a side effect of a merge.

Would you open a Discussion with the environment detail: which proxy or load balancer, what the failure looks like from Nx's side, and what happens today without the toggle? https://github.com/nrwl/nx/discussions/new?category=feature-requests. Worth having that conversation first because the answer might not be a toggle at all. If `http_remote_cache.rs` is reusing a connection the peer has already closed, retrying on that specific failure would fix it for everyone without anyone having to know the flag exists. The branch stays here either way.

<details>
<summary>

</summary>

> Checked for a link both ways before proposing the close: no GitHub closing reference and no Linear key. What is being closed on is the absence of agreement, not the quality of the branch.
>
> The alternative named above, retrying when `packages/nx/src/native/cache/http_remote_cache.rs` reuses a connection the peer has already closed, is a suggestion for that discussion. Triage did not verify it against the file.
>
> The branch also conflicts with `master`, and its first Actions run needed maintainer approval, never got it, and passed GitHub's 30 day cutoff, so the pending run was deleted and the `SUCCESS` rollup is the `socket-security` dependency scan alone. Neither matters on a close, and no rebase is being asked for.

</details>

PR #35745, opened 2026-05-20
fix(release): scope ambiguous-scope check to active release group's projects

### LINKED ISSUE #35744: nx release: ambiguous-scope check fires globally, blocks unrelated release groups
### Current Behavior

`nx release --group=<group>` walks the full commit range since the last
matching tag and resolves each commit's conventional-commit scope against
the **entire** project graph (`findMatchingProjects` over
`projectGraph.nodes`). If a scope is ambiguous against *any* projects in
the workspace, Nx throws unconditionally — even when those projects are
not in the release group currently being processed.

In workspaces using slash-style subpath project names (e.g.
`@scope/cui` plus subpaths `@scope/cui/forms`, `@scope/cui/select`, …),
a single commit with a short-form scope such as `fix(cui): …`
permanently blocks every release group whose tag-range includes it.

Root-cause source pointers:

- Throw: `packages/nx/src/command-line/release/utils/shared.ts`, in
  `getCommitsRelevantToProjects` (compiled: `dist/.../shared.js:344`).
  The per-commit ambiguity check uses the global project graph, not
  the `projects` argument representing the active release group.
- Resolver: `packages/nx/src/utils/find-matching-projects.ts`, in
  `addMatchingProjectsByName`. Word-boundary regex
  `(?<![@a-zA-Z0-9-])${pattern}(?![@a-zA-Z0-9-])` — slash is *not* in
  the negative class, so scope `cui` matches both `@scope/cui` and
  `@scope/cui/forms`.

### Expected Behavior

The cui-scoped commit is irrelevant to the `code-style` release group;
the group should release without error. The ambiguity check should not
fire on scopes that only collide with projects outside the active
release group.

Any one of these resolutions would unblock affected workspaces:

1. **Scope the ambiguity check to projects in the active release group.**
   A commit whose ambiguous scope only collides with projects outside
   the current group should not block that group.
2. **Honor `version.conventionalCommits.useCommitScope: false` as a
   bypass.** Today this flag only affects indirect patch bumps
   (`semver.ts`); it does nothing to suppress the throw.
3. **Add a `release.conventionalCommits.ignoreAmbiguousScopes: true`
   opt-out.**
4. **Tighten `addMatchingProjectsByName`** so slash-separated subpath
   names aren't matched by their parent's short name (e.g., treat `/`
   as a name-segment separator rather than a boundary).

### GitHub Repo

https://github.com/jmclellan-crexi/nx-release-scope-ambiguity-repro

### Steps to Reproduce

```bash
git clone https://github.com/jmclellan-crexi/nx-release-scope-ambiguity-repro
cd nx-release-scope-ambiguity-repro
pnpm install
pnpm exec nx release --group=code-style --yes --verbose
```

The workspace contains three projects:

- `@scope/code-style` — its own single-project release group.
- `@scope/cui` — meta package, tagged `domain:cui`.
- `@scope/cui/forms` — slash-style subpath, tagged `domain:cui`.

The `cui` release group is `["@scope/cui", "tag:domain:cui"]`, the
`code-style` group is `["@scope/code-style"]`. Both use
`version.conventionalCommits: true`.

Git history:

1. `chore: initial scaffold` (tagged `@scope/code-style@1.0.0` and
   `@scope/cui@1.0.0`).
2. `fix(cui): unrelated change` — touches only `packages/cui/src/index.ts`.

Releasing the `code-style` group is expected to be a no-op (no relevant
commits) but instead throws.

### Nx Report

```shell
Node           : 24.3.0
OS             : linux-x64
Native Target  : x86_64-linux
pnpm           : 10.29.2
daemon         : Available

nx             : 22.7.2
@nx/js         : 22.7.2
@nx/workspace  : 22.7.2
@nx/devkit     : 22.7.2
typescript     : 5.9.3
```

### Failure Logs

```shell
NX   Running release version for project: @scope/code-style

@scope/code-style 🏷️  Resolved the current version as 1.0.0 from git tag "@scope/code-style@1.0.0", based on releaseTagPattern "{projectName}@{version}"

 NX   Ambiguous scope "cui" in commit "fix(cui): unrelated change". Matches: @scope/cui/forms, @scope/cui

Error: Ambiguous scope "cui" in commit "fix(cui): unrelated change". Matches: @scope/cui/forms, @scope/cui
    at getCommitsRelevantToProjects (node_modules/nx/dist/src/command-line/release/utils/shared.js:344:27)
    at async resolveSemverSpecifierFromConventionalCommits (node_modules/nx/dist/src/command-line/release/utils/resolve-semver-specifier.js:13:29)
    at async deriveSpecifierFromConventionalCommits (node_modules/nx/dist/src/command-line/release/version/derive-specifier-from-conventional-commits.js:28:33)
    at async ReleaseGroupProcessor.determineVersionBumpForProject (node_modules/nx/dist/src/command-line/release/version/release-group-processor.js:385:30)
    at async ReleaseGroupProcessor.bumpFixedVersionGroup (node_modules/nx/dist/src/command-line/release/version/release-group-processor.js:231:88)
    at async ReleaseGroupProcessor.processGroup (node_modules/nx/dist/src/command-line/release/version/release-group-processor.js:210:9)
    at async ReleaseGroupProcessor.processGroups (node_modules/nx/dist/src/command-line/release/version/release-group-processor.js:119:13)
    at async releaseVersion (node_modules/nx/dist/src/command-line/release/version.js:163:13)
    at async release (node_modules/nx/dist/src/command-line/release/release.js:78:74)


In CI this surfaces as `ELIFECYCLE Command failed with exit code 1.`
from pnpm. Locally on Node 24, the rejected promise prints the
stacktrace but the process can still exit 0 — which is itself a
secondary issue (errors thrown inside `releaseVersion` aren't bubbling
to the CLI exit code consistently across Node versions).
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Production blast radius

A real-world workspace (~600 projects) using this naming convention has
been release-blocked for ~3 weeks as a result of this bug. The
`@scope/code-style` group cannot publish a new patch because a single
unrelated `fix(cui): …` commit lives in the tag range. The `cui` group
will hit the same trap the next time it's released.

### Why no workspace-level workaround is durable

- **Splitting `nx release` into `version --specifier=patch` +
  `changelog --from=<sha>` + `publish`** — works once, but the SHA must
  be bumped manually after every new ambiguous-scope commit; not a fix.
- **Moving the git tag forward past the offender** — mutates a
  published tag.
- **Rewriting the offending commit message via force-push** — branch
  protection bypass, breaks every open PR.
- **Renaming the Nx project name `@scope/cui` → `cui` to short-circuit
  exact-match resolution in `parseStringPattern`** — works for scope
  `cui` only; the next ambiguous scope (`forms`, `auth`, …) needs its
  own rename. Not a structural fix.

### Willing to contribute

Happy to open a PR if a maintainer is comfortable with one of the
resolutions above. Rough sketch (option 1, smallest behavior change):

1. In `packages/nx/src/command-line/release/utils/shared.ts`,
   `getCommitsRelevantToProjects` already receives a `projects: string[]`
   argument representing the active release group. Filter the
   per-pattern match set down to that projects set before counting:
   only throw when *that* filtered set has `> 1` matches.
2. When the scope is ambiguous against projects outside the active
   group, treat the commit as having no scope for this group's
   processing (fall through to the file-affectedness path that already
   exists when `commit.scope` is empty).
3. Add a unit test covering: (a) ambiguous within the group → still
   throws, (b) ambiguous only outside the group → no throw, treated as
   non-scoped, (c) unambiguous → behavior unchanged.

Open to whichever of options 1–4 maintainers prefer; happy to switch
direction based on review.

jmclellan-crexi:
## Current Behavior

`nx release --group=<group>` walks the full commit range since the last matching tag and resolves each commit's conventional-commit scope against the **entire** project graph (`findMatchingProjects` over `projectGraph.nodes`). If a scope is ambiguous against *any* projects in the workspace, Nx throws unconditionally — even when those projects aren't in the release group currently being processed.

In workspaces with slash-style subpath project names (e.g. a meta package `@scope/cui` plus subpath packages `@scope/cui/forms`, `@scope/cui/select`, …), a single commit with a short-form scope like `fix(cui): …` permanently blocks every release group whose tag-range includes it — including unrelated groups that don't contain any of the matching projects.

Closes #35744.

## Expected Behavior

The ambiguity check should be scoped to the active release group's projects. A scope that only collides with projects *outside* the group should fall through to the file-affectedness path that's already used when a commit has no scope at all.

## Fix

In `getCommitsRelevantToProjects`, `projects: string[]` (already in the function signature as the active release group's projects, materialized as `projectSet` at the top) is now applied to the per-pattern ambiguity check and to the `scopedProjects` set:

1. **Per-pattern check filters to the group.** `inGroupMatches = perPatternMatches.filter(p => projectSet.has(p))`. Throw only when `inGroupMatches.length > 1` — ambiguity *within* the group is still a real error.
2. **`scopedProjects` is intersected with the group.** Cross-group matches no longer bleed into `isProjectScopedCommit` downstream.

When the filtered set is empty or singleton, the commit is treated identically to a commit with no scope for this group's processing — `isProjectScopedCommit` falls back to `false` for projects whose `affectedGraph` includes them via files, matching the existing scope-less behavior.

## Tests

`packages/nx/src/command-line/release/utils/shared.spec.ts`:

- **Existing test for in-group ambiguity** (`should throw when commit scope matches multiple projects (ambiguous scope)`) — converted from a `try { … } catch (err) { expect(…) }` pattern (which silently passed if no throw happened) to `await expect(…).rejects.toThrow(…)`. Existing behavior preserved: throws when `@foo/graph` + `@bar/graph` are both in the active group with scope `graph`.
- **New test for cross-group ambiguity** — active group `['lib-a']`, scope `graph` matches `@foo/graph` + `@bar/graph` (both outside the group). No throw. Commit is included via file-affectedness, `isProjectScopedCommit: false`.
- **New test for partial cross-group ambiguity** — active group `['@foo/graph']`, scope `graph` matches `@foo/graph` (in group) + `@bar/graph` (out). Filtered to one match. No throw, treated as scoped to `@foo/graph` (`isProjectScopedCommit: true`).

Also extended `createMockCommit` to accept an explicit `scope` argument so tests actually exercise the scope-parsing path. The existing ambiguity test relied on `feat(graph): …` in the commit *message*, but `commit.scope` itself was hardcoded to `''`, so `scopePatterns` was always empty and the throw never fired in the test — the assertion lived in a catch block that was never reached. The conversion to `rejects.toThrow` plus the explicit `scope` argument makes the original test actually verify what its name says.

## Verification

```
pnpm exec nx test nx --testPathPatterns=shared.spec
# Tests: 38 passed, 38 total

pnpm exec nx test nx --testPathPatterns="release"
# Tests: 1 skipped, 503 passed, 504 total

pnpm exec nx lint nx
# 0 errors (2 pre-existing warnings, unrelated)
```

Minimal real-world repro repo: https://github.com/jmclellan-crexi/nx-release-scope-ambiguity-repro

## PR Checklist

- [x] Bug-fix only — no API surface changes
- [x] Tests added for new behavior + existing test converted to assert what it claims
- [x] Backwards compatible — in-group ambiguity still throws; scope-less commits unchanged; unambiguous scopes unchanged
- [x] No documentation changes needed (behavior is now closer to what the existing docs imply)

### DIFF
--- a/packages/nx/src/command-line/release/utils/shared.ts
+++ b/packages/nx/src/command-line/release/utils/shared.ts
@@ -489,22 +489,32 @@ Promise<Map<string, { commit: GitCommit; isProjectScopedCommit: boolean }[]>> {
     if (scopePatterns.length > 0) {
       const matches = findMatchingProjects(scopePatterns, projectGraph.nodes);
 
-      // detect ambiguity
+      // Detect ambiguity, but only within the active release group's
+      // projects. A scope that resolves ambiguously against projects
+      // outside the group is irrelevant to this group — fall through
+      // to the file-affectedness path used when a commit has no scope
+      // at all. See https://github.com/nrwl/nx/issues/35744.
       for (const pattern of scopePatterns) {
         const perPatternMatches = findMatchingProjects(
           [pattern],
           projectGraph.nodes
         );
+        const inGroupMatches = perPatternMatches.filter((p) =>
+          projectSet.has(p)
+        );
 
-        if (perPatternMatches.length > 1) {
+        if (inGroupMatches.length > 1) {
           throw new Error(
             `Ambiguous scope "${pattern}" in commit "${commit.message}". ` +
-              `Matches: ${perPatternMatches.join(', ')}`
+              `Matches: ${inGroupMatches.join(', ')}`
           );
         }
       }
 
-      scopedProjects = new Set(matches);
+      // Restrict the scoped-projects set to projects in the active
+      // release group so cross-group matches don't bleed into the
+      // isProjectScopedCommit determination below.
+      scopedProjects = new Set(matches.filter((p) => projectSet.has(p)));
     }
 
     for (const projectName of Object.keys(affectedGraph.nodes)) {
--- a/packages/nx/src/command-line/release/utils/shared.ts
+++ b/packages/nx/src/command-line/release/utils/shared.ts
@@ -456,10 +456,18 @@ export async function getCommitsRelevantToProjects(
   commits: GitCommit[],
   projects: string[],
   nxReleaseConfig: NxReleaseConfig,
-  releaseGraph: ReleaseGraph
+  releaseGraph: ReleaseGraph,
+  // The full set of projects in the active release group. For release
+  // groups with `projectsRelationship: 'independent'`, `projects` only
+  // contains the single project currently being processed, so the scope
+  // matching/ambiguity check below would not see its sibling projects.
+  // `releaseGroupProjects` lets us match scopes against the whole group
+  // regardless of relationship. Defaults to `projects` for fixed groups.
+  releaseGroupProjects: string[] = projects
 ): // Map of projectName to GitCommit[]
 Promise<Map<string, { commit: GitCommit; isProjectScopedCommit: boolean }[]>> {
   const projectSet = new Set(projects);
+  const releaseGroupProjectSet = new Set(releaseGroupProjects);
   const relevantCommits: Map<
     string,
     { commit: GitCommit; isProjectScopedCommit: boolean }[]
@@ -500,7 +508,7 @@ Promise<Map<string, { commit: GitCommit; isProjectScopedCommit: boolean }[]>> {
           projectGraph.nodes
         );
         const inGroupMatches = perPatternMatches.filter((p) =>
-          projectSet.has(p)
+          releaseGroupProjectSet.has(p)
         );
 
         if (inGroupMatches.length > 1) {
@@ -514,7 +522,9 @@ Promise<Map<string, { commit: GitCommit; isProjectScopedCommit: boolean }[]>> {
       // Restrict the scoped-projects set to projects in the active
       // release group so cross-group matches don't bleed into the
       // isProjectScopedCommit determination below.
-      scopedProjects = new Set(matches.filter((p) => projectSet.has(p)));
+      scopedProjects = new Set(
+        matches.filter((p) => releaseGroupProjectSet.has(p))
+      );
     }
 
     for (const projectName of Object.keys(affectedGraph.nodes)) {
--- a/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts
+++ b/packages/nx/src/command-line/release/utils/resolve-semver-specifier.ts
@@ -13,7 +13,12 @@ export async function resolveSemverSpecifierFromConventionalCommits(
   projectGraph: ProjectGraph,
   projectNames: string[],
   releaseConfig: NxReleaseConfig,
-  releaseGraph: ReleaseGraph
+  releaseGraph: ReleaseGraph,
+  // The full set of projects in the active release group. For independent
+  // release groups, `projectNames` only contains the single project being
+  // processed, so this is forwarded separately to keep scope matching
+  // accurate against the whole group. Defaults to `projectNames`.
+  releaseGroupProjects: string[] = projectNames
 ): // Map of projectName to semver bump type
 Promise<Map<string, SemverSpecifier | null>> {
   const commits = await getGitDiff(from);
@@ -23,7 +28,8 @@ Promise<Map<string, SemverSpecifier | null>> {
     parsedCommits,
     projectNames,
     releaseConfig,
-    releaseGraph
+    releaseGraph,
+    releaseGroupProjects
   );
   return determineSemverChange(
     relevantCommits,
--- a/packages/nx/src/command-line/release/utils/shared.spec.ts
+++ b/packages/nx/src/command-line/release/utils/shared.spec.ts
@@ -1406,6 +1406,67 @@ describe('shared', () => {
       expect(result.get('@foo/graph')?.[0].isProjectScopedCommit).toBe(true);
     });
 
+    it('should detect intra-group ambiguity for independent release groups even when a single project is being processed', async () => {
+      // For independent release groups, only the single project being
+      // processed is passed as `projects`, but the full release group is
+      // forwarded as `releaseGroupProjects`. The scope `graph` matches
+      // both @foo/graph and @bar/graph, which both live in the active
+      // release group, so this is a genuine in-group ambiguity and must
+      // still throw. See https://github.com/nrwl/nx/issues/35744.
+      const commits: GitCommit[] = [
+        createMockCommit(
+          'indep1',
+          ['foo/graph/src/index.ts'],
+          'feat(graph): ambiguous within independent group',
+          'graph'
+        ),
+      ];
+
+      await expect(
+        getCommitsRelevantToProjects(
+          mockProjectGraph,
+          commits,
+          // only the single independent project currently being processed
+          ['@foo/graph'],
+          mockReleaseConfig!,
+          mockReleaseGraph,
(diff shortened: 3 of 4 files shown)


jmclellan-crexi:
Reproduces using the minimal repo at https://github.com/jmclellan-crexi/nx-release-scope-ambiguity-repro — `pnpm exec nx release --group=code-style --yes` throws on `master`, succeeds on this branch.

packages/nx/src/command-line/release/utils/shared.ts
```diff
@@ -489,22 +489,32 @@ Promise<Map<string, { commit: GitCommit; isProjectScopedCommit: boolean }[]>> {
...
+      // outside the group is irrelevant to this group — fall through
+      // to the file-affectedness path used when a commit has no scope
+      // at all. See https://github.com/nrwl/nx/issues/35744.
       for (const pattern of scopePatterns) {
         const perPatternMatches = findMatchingProjects(
           [pattern],
           projectGraph.nodes
         );
+        const inGroupMatches = perPatternMatches.filter((p) =>
+          projectSet.has(p)
```
leosvelperez: For release groups with `projectsRelationship: "independent"`, `projectSet` only contains the single/independent project currently being processed, not all the projects in the release group (see `packages/nx/src/command-line/release/version/derive-specifier-from-conventional-commits.ts:37`). So, we need to forward the projects in the release group to this function so we can account for independently released projects and properly match against the release group here.

Let's fix that and add unit tests to cover that scenario as well.

leosvelperez (changes requested):

The changes are in the right direction. Let's address one issue with `projectsRelationship: 'independent'`.

packages/nx/src/command-line/release/utils/shared.ts
jmclellan-crexi: Good catch — fixed in 1dc83ad6c6.

The full release group is now forwarded down the chain as a separate `releaseGroupProjects` argument and used for scope matching, while `projects` continues to drive which projects get commits assigned:

- **`getCommitsRelevantToProjects`** (`shared.ts`) — added `releaseGroupProjects: string[] = projects`. The per-pattern ambiguity check and the `scopedProjects` filter now intersect against `releaseGroupProjectSet` instead of `projectSet`. The commit-assignment loop still uses `projectSet`. Defaults to `projects`, so fixed-group behavior is unchanged.
- **`resolveSemverSpecifierFromConventionalCommits`** — threaded the param through (defaults to `projectNames`).
- **`deriveSpecifierFromConventionalCommits`** — passes `releaseGroup.projects` regardless of relationship, so even when a single independent project drives `affectedProjects`, scopes are matched against the whole group.

Added two unit tests for the independent scenario:
- intra-group ambiguity (`projects: ['@foo/graph']`, `releaseGroupProjects: ['@foo/graph', '@bar/graph']`, scope `graph`) → still throws.
- collision only with a project outside the group (release group is just `['@foo/graph']`) → no throw, treated as scoped to `@foo/graph`.

`shared.spec` is now 40 passing; the full `release` suite is 505 passing / 1 skipped.

leosvelperez (approved):
Thank you!

PR #35755, opened 2026-05-20
fix(nx-plugin): plugin lint checks should use dependentTasksOutputFiles

AgentEnder:
## Related Issue(s)

Fixes #

### DIFF
--- a/.nx/workflows/agents.yaml
+++ b/.nx/workflows/agents.yaml
@@ -74,7 +74,7 @@ common-init-steps: &common-init-steps
 
   - name: Restore .NET analyzer projects
     script: |
-      dotnet restore packages/dotnet/analyzer.Tests/MsbuildAnalyzer.Tests.csproj
+      dotnet restore nx.sln
 
   - name: Configure git metadata (needed for lerna smoke tests)
     script: |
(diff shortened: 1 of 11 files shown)


.nx/workflows/agents.yaml
```diff
@@ -74,7 +74,7 @@ common-init-steps: &common-init-steps
 
   - name: Restore .NET analyzer projects
     script: |
-      dotnet restore packages/dotnet/analyzer.Tests/MsbuildAnalyzer.Tests.csproj
+      dotnet restore nx.sln
```
AgentEnder, line 77: Unrelated, but noticed and fixed on this branch.... the OG command is "fine", but this is more proper in case more .NET code finds its way in.

packages/eslint-plugin/src/rules/nx-plugin-checks.ts
```diff
@@ -538,30 +489,27 @@ export function validatePromptNode(
       },
       node: promptNode.value as any,
     });
-  } else {
-    const promptPath = path.join(
+    return;
+  }
+
+  try {
+    resolveSchema(
```
FrozenPandaz: Is this really how we should resolve the prompt? 🤔

packages/eslint-plugin/src/rules/nx-plugin-checks.ts
```diff
@@ -1,15 +1,21 @@
...
-  readJsonFile,
   workspaceRoot,
 } from '@nx/devkit';
 import { loadTsFile, requireWithTsconfigFallback } from '@nx/js/internal';
+import {
+  ImplementationResolutionError,
+  resolveImplementation,
+  resolveSchema,
+  SchemaResolutionError,
+} from 'nx/src/config/schema-utils';
```
FrozenPandaz: Please import this from `@nx/devkit/internal`, not `nx`

FrozenPandaz (approved):

PR #35782, opened 2026-05-23
fix(js): handle already-published version errors in release-publish executor

olagokemills:
## Current Behavior

When using `nx release publish` with pnpm (or other package managers), if a package version has already been published to the registry, the executor fails with an error instead of gracefully handling it.

## Expected Behavior

The release-publish executor should detect "already published" errors (E403 with 'previously published versions', EPUBLISHCONFLICT, E409 conflicts) and skip the publish gracefully with a warning, rather than failing the entire release process.

## Related Issue(s)

Closes #35235


FrozenPandaz (approved):

PR #35786, opened 2026-05-24
fix(core): avoid tsconfig path false positives for sibling project roots

cw-alexcroteau:
## Summary
- Fixes a false-positive in affected-project detection from root `tsconfig` path changes.
- Ensures a changed path only matches a project when it is inside that project root (`root/`) or exactly equal to `root`.

## Problem
`getTouchedProjectsFromTsConfig` currently matches with:
- `normalizedPath.startsWith(root)`
- or `normalizedPath === root`

Using `startsWith(root)` without a path-segment boundary causes sibling roots that only share a prefix to be matched incorrectly.

Example:
- project A root: `libs/typescript/cdk`
- project B root: `libs/typescript/cdk-utils`
- changed path: `libs/typescript/cdk-utils/index.ts`

Before this change, project A can be incorrectly marked as touched because `"libs/typescript/cdk-utils/...".startsWith("libs/typescript/cdk")` is `true`.

## Fix
Use a segment boundary check:
- `normalizedPath.startsWith(`${root}/`) || normalizedPath === root`

## Tests
Added a regression test in `tsconfig-json-changes.spec.ts` that reproduces the sibling-prefix case and verifies only the intended projects are matched.

## Validation Context (anonymized)
This logic has been exercised in a large private monorepo (~1,356 projects) where this false-positive caused broad unnecessary affected sets for similarly-prefixed roots.

### DIFF
--- a/packages/nx/src/plugins/js/project-graph/affected/tsconfig-json-changes.spec.ts
+++ b/packages/nx/src/plugins/js/project-graph/affected/tsconfig-json-changes.spec.ts
@@ -44,21 +44,6 @@ describe('getTouchedProjectsFromTsConfig', () => {
           .spyOn(tsUtils, 'getRootTsConfigFileName')
           .mockReturnValue(tsConfig);
         jest.clearAllMocks();
-
-        graph.nodes['proj-cdk'] = {
-          name: 'proj-cdk',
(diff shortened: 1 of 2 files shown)


jaysoo:

PR #35787, opened 2026-05-24
perf(core): precompile minimatch pattern in findMatchingConfigFiles

cw-alexcroteau:
## Summary
Optimizes `findMatchingConfigFiles` by compiling the top-level pattern once per invocation and reusing it across all project files.

## What Changed
- Replaced per-file `minimatch(file, pattern, { dot: true })` calls with a single precompiled matcher:
  - `const patternMatcher = new Minimatch(pattern, { dot: true })`
  - `patternMatcher.match(file)` inside the loop
- Added focused unit tests for `findMatchingConfigFiles` behavior (include/exclude and include-negation paths).

## Why this approach (Minimatch over Picomatch)
A picomatch variant is measurably faster in microbenchmarks, but this PR intentionally uses precompiled minimatch for maximum compatibility and lowest adoption risk:
- No dependency changes
- No matcher-semantic changes
- Keeps exactly the same matching engine Nx already uses in this code path
- Easy to backport/cherry-pick across maintained Nx versions where this function has the same shape

## Validation Context (anonymized)
This has been benchmarked and exercised in a large private monorepo (~1,356 projects) that actively uses Nx in CI and local development.

Observed benchmark (same data set and matcher outputs):
- Original per-call minimatch: median 677.8ms
- Precompiled minimatch (this approach): median 28.1ms
- Picomatch experiment (not shipped here): median 3.6ms

The shipped approach was selected for cross-version stability while still removing almost all repeated pattern-parse overhead.

### DIFF
--- a/packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
+++ b/packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
@@ -9,6 +9,7 @@ import { LoadedNxPlugin } from '../plugins/loaded-nx-plugin';
 import {
   createProjectConfigurationsWithPlugins,
   CreateNodesResultEntry,
+  findMatchingConfigFiles,
   MergeError,
   mergeCreateNodesResults,
 } from './project-configuration-utils';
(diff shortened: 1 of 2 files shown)


jaysoo:

cw-alexcroteau:
@jaysoo I'm sorry, I tried following the contribution guidelines. This was not bulk-generated, we use this as a patch actively in our corporate monorepo, that I can sadly not provide.

What would you expect as repro for a performance improvement? It's not _strictly_ a bug, as the PR keyword precises, it's an actual performance improvement. I'm happy to create a monorepo with 1000s of projects if you want in order to prove that this gives the performance improvements that I highlighted.

Is the reason why you suspect "mass generation" the fact that I created a couple PRs at once? That's only because I had active NX patches in our corporate monorepo and I wanted to provide them upstream. I tried following guidelines as much as I could, but it's not always easy to prove that performance improvements fix something when the issue comes from a large corporate repository that cannot be shared.

Happy to provide any extra information needed, but at the moment and for that specific one PR, I do not understand why it was closed so quickly saying it's "bulk-generated".

cw-alexcroteau:
For reference, re-opened as #35793 with an issue and a reproduction repo, hopefully this is better.

PR #35788, opened 2026-05-24
feat(js): add tree-shaken dependency narrowing to the JS project graph

cw-alexcroteau:
## Summary

Add opt-in tree-shaking-style, sub-file dependency narrowing to Nx's JS project-graph analysis so workspaces can produce more precise dependency edges and more accurate affected results.

Concretely, the core JS dependency pipeline now has an implementation that narrows down dependencies based on conservative tree-shaking primitives. This was initially introduced as an external repository available at https://github.com/caseware/nx-node-tree-shake-graph-plugin.

## What Changes

- moves tree-shaken dependency narrowing into the core JS project-graph dependency path
- adds `pluginsConfig['@nx/js'].dependencyNarrowing` as the opt-in config surface
- updates `@nx/node:init` to populate that core JS config for Node workspaces
- adds focused tests for the core narrowing behavior, core config ownership, and the Node init wiring

## Validation

Focused Node 22 validation in this fork:

- `packages/nx/src/plugins/js/project-graph/narrow-dependencies.spec.ts`: 2 passed, 2 total
- `packages/node/src/generators/init/init.spec.ts`: 3 passed, 3 total

PR #35789, opened 2026-05-24
feat(core): surface graph-extension in core plugin capabilities

cw-alexcroteau:
## Summary

Surface graph extension as a first-class capability in the core plugin metadata and `nx list` output.

## Why

This was initially going to be folded into #35788, but that PR's implementation moved tree-shaken dependency narrowing into the core JS graph pipeline instead of keeping a separate `@nx/node` graph authority.

That design change still leaves value in exposing graph hooks explicitly as a plugin capability. Nx plugins can participate in project-graph behavior through `createNodes`, `createNodesV2`, `createDependencies`, and related graph hooks, and the static core-plugin listing should model that in the same terms as the installed-plugin capability pipeline.

## What Changes

- changes `packages/nx/src/utils/plugins/core-plugins.ts` to model core plugin capabilities as first-class capability arrays instead of pre-joined strings
- represents graph participation with the existing `graph-extension` capability label
- updates the list output to format static core-plugin capabilities through the same capability-list path used elsewhere in plugin output
- marks graph-extending core plugins in the static list, including `@nx/js` and `@nx/gradle`
- adds coverage for the static core-plugin listing output

## Validation

Focused Node 22 validation in this fork:

- `packages/nx/src/utils/plugins/output.spec.ts`: 15 passed, 15 total

### DIFF
--- a/packages/nx/src/utils/plugins/core-plugins.ts
+++ b/packages/nx/src/utils/plugins/core-plugins.ts
@@ -1,116 +1,121 @@
+export type CorePluginCapability =
+  | 'executors'
+  | 'generators'
+  | 'graph-extension';
+
 export interface CorePlugin {
   name: string;
-  capabilities: 'executors' | 'generators' | 'executors,generators' | 'graph';
+  capabilities: CorePluginCapability[];
   link?: string;
 }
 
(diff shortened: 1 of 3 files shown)


jaysoo:

PR #35796, opened 2026-05-25
fix(core): avoid tsconfig path false positives for sibling project roots

### LINKED ISSUE #35795: fix(core): avoid tsconfig path false positives for sibling project roots
## Current Behavior

When a root `tsconfig` path mapping changes, `getTouchedProjectsFromTsConfig` matches projects with `normalizedPath.startsWith(root)`.

That causes false positives for sibling project roots that only share a string prefix.

Example:

- `libs/typescript/cdk`
- `libs/typescript/cdk-utils`

A change that only touches `libs/typescript/cdk-utils/...` also marks `libs/typescript/cdk` as touched.

## Expected Behavior

A changed path should only match a project when it is inside that project's root (`root/`) or exactly equal to `root`.

## Reproduction

Public repro repo:
https://github.com/cw-alexcroteau/nx-tsconfig-path-prefix-false-positive-repro-20260525

Passing repro CI:
https://github.com/cw-alexcroteau/nx-tsconfig-path-prefix-false-positive-repro-20260525/actions/runs/26414457118

The repro demonstrates:

- stock behavior: `ts-cdk` and `ts-cdk-utils` are both reported
- fixed behavior: only `ts-cdk-utils` is reported

## Nx Report

Repro repo uses `nx@22.7.3`.

cw-alexcroteau:
## Current Behavior

Root `tsconfig` path changes can mark sibling project roots as touched when they only share a string prefix.

## Expected Behavior

Only the project whose root contains the changed path, or exactly matches it, should be marked as touched.

## Related Issue(s)

Fixes #35795

## Summary

- avoid false positives in `getTouchedProjectsFromTsConfig` for sibling roots with shared prefixes
- add a regression test covering `libs/typescript/cdk` vs `libs/typescript/cdk-utils`
- include external repro evidence:
  - https://github.com/cw-alexcroteau/nx-tsconfig-path-prefix-false-positive-repro-20260525
  - https://github.com/cw-alexcroteau/nx-tsconfig-path-prefix-false-positive-repro-20260525/actions/runs/26414457118

Supersedes closed PR #35786.


leosvelperez (approved):
Thanks!

PR #35798, opened 2026-05-25
feat(js): add tree-shaking-based dependency narrowing to the JS project graph

### LINKED ISSUE #35797: feat(js): add tree-shaking-based dependency narrowing to the JS project graph
## Summary

Nx should add conservative tree-shaking-based dependency narrowing to the JS project graph.

This feature has two distinct aspects:

1. Tree-shaking projects out of the static graph when symbol-level analysis can prove the imported runtime symbols from a side-effect-free target are unused.
2. Narrowing affected-projects when a change only modifies some exported symbols and downstream projects do not import those symbols.

The second aspect is the higher-value outcome, because it directly reduces noisy affected results for real change sets.

Closed PR [#35788](https://github.com/nrwl/nx/pull/35788) carried the original core port, but it is now closed.

## Why This Feature Matters

Today, Nx still keeps JS graph edges at a broader whole-file / whole-project level than necessary.

That means there are two missed opportunities:
- Static graph precision: downstream project edges stay in the graph even when imported symbols from a side-effect-free target are unused and can be conservatively narrowed away.
- Affected precision: a change that only touches one exported helper can still mark downstream projects as affected even if they do not import that symbol.

The affected side is the highest-value part of the feature.

A representative example is a helper module that exports multiple symbols, where changing one export should not keep unrelated downstream projects affected if those projects do not import that symbol.

## Conservative Contract

This is not aggressive dead-code elimination.

The contract should stay conservative:
- remove static-graph edges only when symbol-level analysis can prove the imported runtime symbols are unused
- keep edges in cases such as side-effect imports, dynamic imports, namespace imports, re-exports, and anything else that cannot be proven safe to narrow
- apply the same conservative symbol-level reasoning during affected calculation when only some exports in a file are modified

## Public Reproduction

Public repro and CI:
- Repro repo PR: https://github.com/caseware/nx-node-tree-shake-graph-plugin/pull/9
- CI run: https://github.com/caseware/nx-node-tree-shake-graph-plugin/actions/runs/26421260021

The compat scenario was updated so the affected-narrowing case now mutates `libs/lib-0/src/loader.*` instead of a trivial sentinel file, and the generated helper exports `fetchEnvFromRemoteStore(...)` to mirror a generic loader-style helper change.

## Related Context

- Closed prior PR: https://github.com/nrwl/nx/pull/35788
- Repro branch on fork: `cw-alexcroteau:feat/skip-source-analysis-metadata`

cw-alexcroteau:
## Summary

This PR adds conservative tree-shaking-based dependency narrowing to the JS project graph.

It covers two related capabilities:

1. Static-graph narrowing: tree-shake projects out of the JS project graph when symbol-level analysis can prove the imported runtime symbols from a side-effect-free target are unused.
2. Affected narrowing: reduce the scope of affected-projects when a change only modifies some exported symbols and downstream projects do not import those symbols.

The affected-project narrowing is the higher-value part of the feature, because it directly improves affected precision for real code changes.

## Current Behavior
1. When a project depends on another project via JS imports (i.e. an application importing a single symbol from a library), this creates a transitive relationship that can cascade into a lot of nested affected projects.
2. When affecting a project by changing some symbols in JS code, the all of the library's dependencies are considered affected even for projects that do not use the modified symbols.

## Expected Behavior
1. Projects dependency graph is narrowed down based on imported symbols, especially for transient dependencies. If a project does not import symbols from other projects, it is excluded from the graph edges computation.
2. When affecting a project by changing some symbols in JS code, the modified library's dependencies should be considered as affected only if they use one or more of the modified symbols.

## What This PR Does

In practice, that means:
- removing edges only when symbol-level analysis can prove the imported runtime symbols are unused
- keeping edges for side-effect imports, dynamic imports, namespace imports, re-exports, and other cases that cannot be narrowed safely
- applying the same conservative reasoning during affected calculation when only some exports in a file are modified

## Public Reproduction

- repro PR: https://github.com/caseware/nx-node-tree-shake-graph-plugin/pull/9
- compat CI: https://github.com/caseware/nx-node-tree-shake-graph-plugin/actions/runs/26421260021

## Related Context

- earlier core-port attempt: https://github.com/nrwl/nx/pull/35788

## Related Issue(s)

Fixes #35797


cw-alexcroteau:
@nrwl/nx-pipelines-reviewers I believe this is a sufficiently complex feature addition that warrants its own minor release for testing purposes, per the PR template. Please let me know if you believe the same.

AgentEnder:
Thanks for the PR. @leosvelperez has it. Could you rebase this on `master` first? It conflicts as it stands, and the CI approval on the head commit expired before anyone ran it, so there is nothing left to approve. A rebased push is what gets it both mergeable and built.

<details>
<summary>

</summary>

> The `SUCCESS` rollup on the head commit is the `socket-security` dependency scan alone, so nothing about the branch itself has been built.
>
> The diff is +2435 across 22 files, and it adds a new analysis pass to the JS project graph rather than adjusting an existing one, so agreeing on the approach probably comes before any line by line read.
>
> #35797 is linked and worth reading alongside it.

</details>

cw-alexcroteau:
@AgentEnder Conflicts are resolved and the branch has been rebased onto `master`. Thanks for your reply.

PR #35825, opened 2026-05-29
feat(module-federation): deprecate old generators and add new consumer/provider generators

jaysoo:
The existing MF generators and setup are not modernized. We can simplify the setup and remove strict dependency on Nx executors.

- Consolidate on official module naming: `host -> consumer`, `remote -> provider`
- Always use dynamic runtime (i.e. use runtime utils from `@module-federation/enhanced/runtime` so we don't have to start up every provider app on dev machine
- No more Nx specific bundler plugins, only standard config as seen on module-federation.io


packages/react/src/generators/consumer/files/rsbuild/rsbuild.config.ts__tmpl__
```diff
@@ -0,0 +1,31 @@
...
+
+export default defineConfig({
+  server: { port: PORT },
+  // assetPrefix:true keeps chunk URLs aligned with the dev-server origin -
+  // mirrors the provider config so a consumer that also exposes federated
+  // modules later doesn't need a separate fix.
+  dev: { assetPrefix: true },
+  html: { template: './index.html' },
+  source: { entry: { index: './src/index.ts' } },
+  plugins: [pluginReact()],
```
pkasarda: I wonder why not `pluginModuleFederation` from `@module-federation/rsbuild-plugin`?

packages/react/src/generators/consumer/files/rsbuild/rsbuild.config.ts__tmpl__
```diff
@@ -0,0 +1,31 @@
...
+  tools: {
+    rspack: (config, { appendPlugins }) => {
+      config.output ??= {};
+      config.output.uniqueName = NAME;
+      appendPlugins([
+        new ModuleFederationPlugin({
+          name: NAME,
+          // No build-time `remotes:` block - registered at runtime in
```
pkasarda: any notes around HMR? will it work correctly also with MFEs?

packages/react/src/generators/consumer/files/rsbuild/rsbuild.config.ts__tmpl__
jaysoo: Good point we can swap that one in as it is cleaner. I just pushed an update.

packages/react/src/generators/consumer/files/rsbuild/rsbuild.config.ts__tmpl__
jaysoo: Yes, HMR does work. If you edit a provider app, it will hot-replace what's changed, state is kept.

leosvelperez (approved):

packages/react/src/generators/consumer/files/rsbuild/rsbuild.config.ts__tmpl__
pkasarda: Sorry @jaysoo, what about consumers the MFEs? It is not working for me even with latest beta version... only provider (host) HMR is working

PR #35835, opened 2026-05-29
feat(core): feed migration docs to agents in nx migrate

leosvelperez:
## Current Behavior

A migration's documentation (the co-located markdown explaining what it does) is not referenced in `migrations.json`, so when `nx migrate --run-migrations --agentic` drives an AI agent, the agent has no access to it.

## Expected Behavior

Migration entries can declare a `docs` markdown path. `nx migrate` carries it into the generated `migrations.json`, and under `--agentic` the resolved path is surfaced to the agent (for prompt, hybrid, and validation steps) so it has context on what the migration is meant to do. Existing first-party migrations with co-located docs are wired up via the new field and published with their packages.

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/nxc-4495-6c420036)

### DIFF
--- a/packages/nx/src/command-line/migrate/migrate.ts
+++ b/packages/nx/src/command-line/migrate/migrate.ts
@@ -2374,7 +2374,7 @@ type ExecutableMigration = {
   implementation?: string;
   factory?: string;
   prompt?: string;
-  docs?: string;
+  documentation?: string;
 };
 
 export { isPromptOnlyMigration, isHybridMigration };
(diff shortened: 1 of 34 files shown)


packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -3451,6 +3458,57 @@ export function getImplementationPath(
...
+ *
+ * Returns `undefined` when the migration declares no `docs` or the file can't
+ * be resolved. Docs are supplementary context, so a missing file is non-fatal
+ * here (unlike a missing implementation, which aborts the migration).
+ */
+export function resolveMigrationDocsPath(
+  root: string,
+  migration: { package: string; docs?: string }
+): string | undefined {
+  if (!migration.docs) {
```
FrozenPandaz: Let's name the field `documentation`

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -3451,6 +3458,57 @@ export function getImplementationPath(
...
+    return undefined;
+  }
+
+  let migrationsFilePath: string | null;
+  try {
+    migrationsFilePath = readPackageMigrationConfig(
+      migration.package,
+      root
+    ).migrations;
```
FrozenPandaz: Feels weird that we have to read the migration config again here? Shouldn't there be a phase earlier where we resolve the migrations that we can do this resolution?

FrozenPandaz (changes requested):
Add a regression test that `documentation` survives into the generated migrations.json — it's the one type-unsafe seam (GeneratedMigrationDetails omits docs + an as any cast), so a future refactor could silently drop it with no test
  failing. An analogous prompt test already exists at migrate.spec.ts:3628.

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -3451,6 +3458,57 @@ export function getImplementationPath(
...
+  let docsPath: string;
+  try {
+    docsPath = require.resolve(migration.docs, { paths: [migrationsDir] });
+  } catch {
+    try {
+      // workaround for a bug in node 12
+      docsPath = require.resolve(`${migrationsDir}/${migration.docs}`);
+    } catch {
+      return undefined;
+    }
```
FrozenPandaz: We don't support Node 12... but we should probably also... error for this? 🤔

FrozenPandaz (changes requested):

FrozenPandaz (approved):

PR #35860, opened 2026-06-02
fix(misc): multi-version support compliance for rollup, webpack, and module-federation

FrozenPandaz:
## Current Behavior

`@nx/rollup`, `@nx/webpack`, and `@nx/module-federation` pin their primary third-party packages as **bundled `dependencies`**:

- `@nx/rollup` → `rollup` (`^4.14.0`)
- `@nx/webpack` → `webpack`, `webpack-dev-server` (v5)
- `@nx/module-federation` → `@module-federation/enhanced`, `@module-federation/node`, `@module-federation/sdk` (v2)

This causes version conflicts with the version a workspace actually has installed, prevents `nx migrate --first-party-only` from upgrading Nx without dragging these along, and lets generators silently overwrite a user's pinned version. Several cross-version `packageJsonUpdates` migrations also bump packages **without `requires` gates**, so they fire on workspaces outside the intended source range (e.g. the `@nx/webpack` `sass-loader` v12 → v16 bump, and all of `@nx/module-federation`'s `@module-federation/*` bumps).

## Expected Behavior

Each plugin declares the packages it invokes at runtime as **optional peer dependencies** matching its real support window, so the workspace controls the version:

- `@nx/rollup` → `rollup` `^3.0.0 || ^4.0.0`
- `@nx/webpack` → `webpack` / `webpack-dev-server` / `webpack-cli` `^5.0.0` (the effective floor — the plugin already calls webpack-5-only APIs such as `ids.HashedModuleIdsPlugin`, `webpack.sources`, and `experiments.cacheUnaffected`)
- `@nx/module-federation` → `@module-federation/enhanced` / `@module-federation/node` `^2.0.0` (`@module-federation/sdk` stays a dependency — it is a type-only import surfaced in the public `.d.ts`)

Generators assert the supported floor (with a parameterized `all-generators-enforce-floor` spec), preserve user-pinned versions (`keepExistingVersions` now defaults to `true`), and install the now-peer build tool for new workspaces. Because every workspace that uses these tools already has the corresponding `@nx/*` plugin as a direct dependency, a `packageJsonUpdates` bridge in each plugin installs the peer into existing workspaces on `nx migrate` (and correctly skips workspaces that only carry the plugin transitively). Cross-version migrations are gated on their source range.

### Changes per plugin

- **`@nx/rollup` (NXC-4403)** — `rollup` → optional peer; floor assert (v3) in all generators; `keepExistingVersions` default `true`; install `rollup` on both the inferred-plugin and executor paths; bridge migration for existing workspaces. No runtime branching/version map is needed — the plugin's Rollup usage is already v2/v3/v4-agnostic.
- **`@nx/webpack` (NXC-4410)** — `webpack`/`webpack-dev-server`/`webpack-cli` → optional peers (`^5`); floor assert (v5) in all generators; `keepExistingVersions` default `true`; gate the `sass-loader` v12 → v16 migration on `^12`; install `webpack`/`webpack-dev-server` for new workspaces; bridge migration for existing ones.
- **`@nx/module-federation` (NXC-4393)** — `@module-federation/enhanced`/`node` → optional peers (`^2`); add `requires` gates (on the `@module-federation/enhanced` source range, mirroring the existing `@nx/angular` MF migrations) to every `packageJsonUpdates` entry, and split the independent `http-proxy-middleware` v2 → v3 bump into its own gated entry. This is a runtime-only library (no generators/executors), so there is no floor assert/spec.

## Related Issue(s)

Tracked in Linear under the "Multi-version supported across plugins" milestone:

- NXC-4403 — `@nx/rollup`
- NXC-4410 — `@nx/webpack`
- NXC-4393 — `@nx/module-federation`

No GitHub issue to auto-close.

### DIFF
--- a/packages/rollup/migrations.json
+++ b/packages/rollup/migrations.json
@@ -8,5 +8,15 @@
       "documentation": "./src/migrations/update-23-0-0/remove-use-legacy-typescript-plugin.md"
     }
   },
-  "packageJsonUpdates": {}
+  "packageJsonUpdates": {
+    "23.0.0": {
+      "version": "23.0.0-beta.24",
+      "packages": {
+        "rollup": {
+          "version": "^4.14.0",
+          "alwaysAddToPackageJson": true
+        }
+      }
+    }
+  }
 }
--- a/packages/webpack/migrations.json
+++ b/packages/webpack/migrations.json
@@ -24,6 +24,9 @@
   "packageJsonUpdates": {
     "20.5.0": {
       "version": "20.5.0-beta.3",
+      "requires": {
+        "sass-loader": "^12.0.0"
+      },
       "packages": {
         "sass-loader": {
           "version": "^16.0.4",
@@ -50,6 +53,23 @@
         "webpack": {
           "version": "5.101.3",
           "alwaysAddToPackageJson": false
+        }
+      }
+    },
+    "23.0.0": {
+      "version": "23.0.0-beta.24",
+      "packages": {
+        "webpack": {
+          "version": "^5.101.3",
+          "alwaysAddToPackageJson": true
+        },
+        "webpack-dev-server": {
+          "version": "^5.2.1",
+          "alwaysAddToPackageJson": true
+        },
+        "webpack-cli": {
+          "version": "^7.0.0",
+          "alwaysAddToPackageJson": true
         }
       }
     }
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -61,12 +61,26 @@
     "ts-loader": "^9.3.1",
     "tsconfig-paths-webpack-plugin": "4.2.0",
     "tslib": "catalog:typescript",
-    "webpack": "catalog:",
-    "webpack-dev-server": "catalog:",
     "webpack-node-externals": "^3.0.0",
     "webpack-subresource-integrity": "^5.1.0",
     "@nx/devkit": "workspace:*",
     "@nx/js": "workspace:*"
+  },
+  "peerDependencies": {
+    "webpack": "^5.0.0",
+    "webpack-cli": "^5.0.0 || ^6.0.0 || ^7.0.0",
+    "webpack-dev-server": "^5.0.0"
+  },
+  "peerDependenciesMeta": {
+    "webpack": {
+      "optional": true
+    },
+    "webpack-cli": {
+      "optional": true
+    },
+    "webpack-dev-server": {
+      "optional": true
+    }
   },
   "devDependencies": {
     "nx": "workspace:*"
--- a/packages/webpack/src/executors/dev-server/dev-server.impl.ts
+++ b/packages/webpack/src/executors/dev-server/dev-server.impl.ts
@@ -1,5 +1,4 @@
 import { eachValueFrom } from '@nx/devkit/internal';
-import webpack from 'webpack';
 import {
   ExecutorContext,
   parseTargetString,
@@ -7,7 +6,6 @@ import {
 } from '@nx/devkit';
 
 import { map, tap } from 'rxjs/operators';
-import WebpackDevServer from 'webpack-dev-server';
 
 import { getDevServerOptions } from './lib/get-dev-server-config';
 import {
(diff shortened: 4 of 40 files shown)


packages/rollup/migrations.json
```diff
@@ -8,5 +8,15 @@
...
     }
   },
-  "packageJsonUpdates": {}
+  "packageJsonUpdates": {
+    "23.0.0": {
+      "version": "23.0.0-beta.24",
+      "packages": {
+        "rollup": {
+          "version": "^4.14.0",
+          "alwaysAddToPackageJson": true
```
leosvelperez, line 17: Why force the installation? The `rollup` peer dep is optional according to the `package.json`

packages/webpack/migrations.json
```diff
@@ -52,6 +55,19 @@
...
+    "23.0.0": {
+      "version": "23.0.0-beta.24",
+      "packages": {
+        "webpack": {
+          "version": "^5.101.3",
+          "alwaysAddToPackageJson": true
+        },
+        "webpack-dev-server": {
+          "version": "^5.2.1",
+          "alwaysAddToPackageJson": true
```
leosvelperez, line 68: Same question about forcing the installation here.

packages/webpack/migrations.json
```diff
@@ -52,6 +55,19 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "23.0.0": {
+      "version": "23.0.0-beta.24",
+      "packages": {
+        "webpack": {
+          "version": "^5.101.3",
```
leosvelperez, line 63: Do we need a migration for this if a repo is in v4? v4->v5 has breaking changes, so just bumping the packages would leave the repo in a bad state.

packages/rollup/migrations.json
```diff
@@ -8,5 +8,15 @@
       "documentation": "./src/migrations/update-23-0-0/remove-use-legacy-typescript-plugin.md"
     }
   },
-  "packageJsonUpdates": {}
+  "packageJsonUpdates": {
+    "23.0.0": {
+      "version": "23.0.0-beta.24",
+      "packages": {
+        "rollup": {
+          "version": "^4.14.0",
```
leosvelperez, line 16: Are any breaking changes that need a migration when bumping to v4?

packages/webpack/src/executors/dev-server/dev-server.impl.ts
```diff
@@ -114,6 +112,12 @@ export async function* devServerExecutor(
     }
   }
 
+  // `webpack` and `webpack-dev-server` are optional peer dependencies. Require
+  // them lazily so that importing this executor (e.g. when the project graph
+  // loads `@nx/webpack`) does not throw before they have been installed.
+  const webpack = require('webpack');
+  const WebpackDevServer = require('webpack-dev-server');
```
leosvelperez: The `@nx/webpack:webpack` executor also has top-level imports.

packages/webpack/package.json
```diff
@@ -61,13 +61,27 @@
...
-    "webpack": "catalog:",
-    "webpack-dev-server": "catalog:",
     "webpack-node-externals": "^3.0.0",
     "webpack-subresource-integrity": "^5.1.0",
     "@nx/devkit": "workspace:*",
     "@nx/js": "workspace:*"
   },
+  "peerDependencies": {
+    "webpack": "^5.0.0",
+    "webpack-cli": "^5.0.0 || ^6.0.0",
```
leosvelperez: Do we not support `webpack-cli@7`?

packages/webpack/src/utils/versions.ts
```diff
@@ -1,11 +1,26 @@
...
 
+// Floor for the generator-level support check. The plugin's runtime code uses
+// webpack 5-only APIs (e.g. `ids.HashedModuleIdsPlugin`, `webpack.sources`,
+// `experiments.cacheUnaffected`), so webpack 5 is the supported floor.
+export const minSupportedWebpackVersion = '5.0.0';
+
+// Fresh-install versions written when the package is not already present.
+export const webpackVersion = '^5.101.3';
+export const webpackDevServerVersion = '^5.2.1';
 export const webpackCliVersion = '^5.1.4';
```
leosvelperez: We state we support `^5.0.0 || ^6.0.0`, but we're installing `^5.1.4` by default, is that intended?

packages/webpack/package.json
```diff
@@ -61,13 +61,27 @@
...
+    "webpack-dev-server": "^5.0.0"
```
leosvelperez, line 72: The webpack docs were not updated.

leosvelperez (changes requested):

packages/webpack/package.json
FrozenPandaz: They've been updated

leosvelperez (approved):

PR #35863, opened 2026-06-03
feat(core): suggest available targets and closest task id on run-one errors

AgentEnder:
## Current Behavior

When `nx run` can't resolve the project or target, the error gives no guidance:

- Unknown target on a real project surfaces deep inside the task graph builder as a bare `Cannot find configuration for task <project>:<target>` — no list of what *is* available, no spelling help.
- Unknown project just prints `Cannot find project '<name>'` with no suggestion of what the user might have meant.
- A typo in a target/configuration that happens to contain a `:` (a legal target name in Nx) gets parsed into separate `target` + `configuration` parts by `splitTarget`, so any naive similarity match against the bare `target` can never find the real colon-containing name.

## Expected Behavior

`run-one` now validates the project and target up front and produces a friendly error before reaching the task-graph stage.

**Unknown target:**
```
 NX   Cannot find target "biuld" for project "my-app"

   Did you mean "build"?

   Available targets:
     - run
     - serve
```
- The suggested target is *not* repeated in the "Available targets" list — it is already surfaced on the `Did you mean` line. When the suggestion was the project's only target, the list is omitted entirely rather than printing a bare header.
- The available-targets list is capped at 5 with a `...and N more` line so output stays scannable on big projects.
- The `Did you mean` suggestion uses Levenshtein distance with a tolerance that scales with the length of the part of the input that actually differs from the candidate, so it stays silent for total nonsense — including the case where only one segment of a `project:target` id is wrong (`nx run zzz:build` suggests nothing rather than `api:build`).

**Unknown project — including a colon target:**
```
 NX   Cannot find project 'nxx'

   Did you mean one of these?
     - nx:zzcustom:variant
```
The attempted `project:target[:configuration]` is matched against every `project:target` task id in the graph, so project typos surface the real invocation (`webpai:build` → `webapi:build`), and project typos against colon-named targets still resolve (`nxx:zzcustom:variant` → `nx:zzcustom:variant`).

**Colon-containing target/configuration names:**
Because `splitTarget` parses the trailing segment off into a configuration, both error builders score the **fully rejoined** specifier *and* every progressively shorter form, then keep whichever produced the closest candidate (ties go to the longer form). So `nx:zzcustom:variantt` suggests `zzcustom:variant`, while genuine `target` + `configuration` typos (`biuld:production` → `build`) still match via the shorter form — and a near-exact shorter match is never demoted behind a worse rejoined one (`app:buld:storybook` suggests `build`, not `build-storybook`).

**Out of scope:** a typo in the *configuration* of a target that does exist (`nx run app:build:prodcution`) still falls through to the existing error — the target resolves, so this code path is never reached.

### Implementation notes
- New `packages/nx/src/utils/string-similarity.ts` with `rankByDistance` / `rankSuggestions` / `findClosestMatches` (Levenshtein + a tolerance scaled to the differing portion of the input).
- `run-one.ts` adds exported, unit-testable helpers `getRunOneTargetError` and `getCannotFindProjectError`, a shared `findClosestSpecifier` that scores every progressive join, and `formatAvailableTargets` for the capped list.
- Unit tests cover the new helpers (including the colon-target scenarios, the genuine target+config-typo fallback, the empty-list case, and the cross-form ranking). The existing `echo:fail` e2e is updated to assert the new message.
- Verified end-to-end against this repo's CLI with a temporary colon target.

### DIFF
--- a/e2e/nx/src/run.test.ts
+++ b/e2e/nx/src/run.test.ts
@@ -195,11 +195,13 @@ describe('Nx Running Tests', () => {
         expect(stdout).toMatch(/ECHOED positional --a=123 --no-b/);
       }
 
-      expect(
-        runCLI(`echo:fail ${mylib}`, {
-          silenceError: true,
-        })
-      ).toContain(`Cannot find configuration for task ${mylib}:echo:fail`);
+      const echoFailOutput = runCLI(`echo:fail ${mylib}`, {
+        silenceError: true,
+      });
+      // The target does not exist, so run-one reports the available targets
+      // (the only included script) instead of a cryptic task graph error.
+      expect(echoFailOutput).toContain(`Cannot find target "echo:fail"`);
+      expect(echoFailOutput).toContain(`echo:dev`);
 
       updateJson(`libs/${mylib}/project.json`, (c) => original);
     }, 1000000);
--- a/e2e/nx/src/run.test.ts
+++ b/e2e/nx/src/run.test.ts
@@ -200,7 +200,9 @@ describe('Nx Running Tests', () => {
       });
       // The target does not exist, so run-one reports the available targets
       // (the only included script) instead of a cryptic task graph error.
-      expect(echoFailOutput).toContain(`Cannot find target "echo:fail"`);
+      expect(echoFailOutput).toContain(
+        `Cannot find target "echo:fail" for project "${mylib}"`
+      );
       expect(echoFailOutput).toContain(`echo:dev`);
 
       updateJson(`libs/${mylib}/project.json`, (c) => original);
--- a/e2e/nx/src/run.test.ts
+++ b/e2e/nx/src/run.test.ts
@@ -206,11 +206,13 @@ describe('Nx Running Tests', () => {
         expect(stdout).toMatch(/ECHOED positional --a=123 --no-b/);
       }
 
-      expect(
-        runCLI(`echo:fail ${mylib}`, {
-          silenceError: true,
-        })
-      ).toContain(`Cannot find configuration for task ${mylib}:echo:fail`);
+      const echoFailOutput = runCLI(`echo:fail ${mylib}`, {
+        silenceError: true,
+      });
+      // The target does not exist, so run-one reports the available targets
+      // (the only included script) instead of a cryptic task graph error.
+      expect(echoFailOutput).toContain(`Cannot find target "echo:fail"`);
+      expect(echoFailOutput).toContain(`echo:dev`);
 
       updateJson(`libs/${mylib}/project.json`, (c) => original);
     }, 1000000);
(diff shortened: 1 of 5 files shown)


packages/nx/src/command-line/run/run-one.ts
```diff
@@ -142,12 +160,144 @@ function getProjects(
...
+  const [closestMatch] = findClosestSpecifier(
+    [target, configuration],
+    availableTargets
+  );
+  if (closestMatch) {
+    bodyLines.push(`Did you mean "${closestMatch}"?`, '');
+  }
+
+  if (availableTargets.length) {
+    bodyLines.push(...formatAvailableTargets(availableTargets));
```
FrozenPandaz: We have already shown closestMatch... so we shouldn't show it again.

In fact... we've calculated all these distances... let's sort these by distance.. and tie break alphabetically.

e2e/nx/src/run.test.ts
```diff
@@ -195,11 +195,13 @@ describe('Nx Running Tests', () => {
...
-        runCLI(`echo:fail ${mylib}`, {
-          silenceError: true,
-        })
-      ).toContain(`Cannot find configuration for task ${mylib}:echo:fail`);
+      const echoFailOutput = runCLI(`echo:fail ${mylib}`, {
+        silenceError: true,
+      });
+      // The target does not exist, so run-one reports the available targets
+      // (the only included script) instead of a cryptic task graph error.
+      expect(echoFailOutput).toContain(`Cannot find target "echo:fail"`);
```
FrozenPandaz:
```suggestion
      expect(echoFailOutput).toContain(`Cannot find target "echo:fail" for project "mylib"`);
```

FrozenPandaz (commented):
Let's add a stress test for this.. to make sure it does not take too long for a massive workspace.

packages/nx/src/utils/string-similarity.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Computes the Levenshtein edit distance between two strings, i.e. the minimum
+ * number of single-character insertions, deletions, or substitutions required
+ * to turn `a` into `b`.
+ */
+export function levenshteinDistance(a: string, b: string): number {
```
FrozenPandaz, line 6: The one theoretical optimization it skips: you could abandon a comparison early once the distance provably exceeds the threshold (Math.max(2, ceil(len × 0.4))), or skip candidates whose length differs from the input by more than the threshold (length difference is a lower bound on edit distance). Both would be justified on a hot path; here they'd be complexity for nothing.

So: cheap algorithmically, cheap at realistic scale, and on a once-per-failure path where even an expensive implementation would be fine. No concern.

AgentEnder:
@FrozenPandaz — I pushed on Jul 30 following your Jul 21 review, and nothing has moved here since. The PR is green and mergeable. Mind taking another look when you get a chance?

PR #35871, opened 2026-06-03
docs(nx-dev): show spread token for extending target defaults

jaysoo:
## Current Behavior

The Configuring Tasks tutorial only covers overriding `targetDefaults`.

## Expected Behavior

Adds a section showing the `"..."` spread token to extend a target default for a project instead of replacing it.

Preview: https://deploy-preview-35871--nx-docs.netlify.app/docs/getting-started/tutorials/configuring-tasks#extending-target-defaults-for-a-project

## Related Issue(s)

DOC-509

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/docs-spread-6df4621c)

### DIFF
--- a/astro-docs/src/content/docs/getting-started/Tutorials/configuring-tasks.mdoc
+++ b/astro-docs/src/content/docs/getting-started/Tutorials/configuring-tasks.mdoc
@@ -113,7 +113,7 @@ The `^` prefix means "the same task on projects this project depends on." So `nx
 
 You can also define dependencies without `^` for tasks within the same project:
 
-{% tabs %}
+{% tabs syncKey="project-config-file" %}
 {% tabitem label="package.json" %}
 
 ```jsonc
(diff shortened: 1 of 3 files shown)


AgentEnder (approved):

astro-docs/src/content/docs/getting-started/Tutorials/configuring-tasks.mdoc
```diff
@@ -228,6 +228,25 @@ When many projects share the same task configuration, defining it in every `proj
...
+{
+  "targets": {
+    "build": {
+      "dependsOn": ["...", "generate-api-types"],
+    },
+  },
+}
+```
+
+The `"..."` expands to the target default's values, so `my-app`'s `build` depends on both `^build` (from the default) and `generate-api-types`. The token also works in objects and configurations. For the full reference, see [spread token](/docs/reference/project-configuration#spread-token).
```
AgentEnder: Minor nit, this isn't expanding target defaults... its expanding whatever config existed pre-override so it could also be referencing config from an inference plugin... I don't necessarily want people to be confused thinking it only works for target default based configs.

PR #35874, opened 2026-06-04
fix(core): disallow nx migrate --interactive for Nx v23+ targets

leosvelperez:
## Current Behavior

`nx migrate --interactive` enables the per-package `x-prompt` confirmations regardless of the target version. This conflicts with the new `--mode` functionality: `--mode=third-party --interactive` still fires the prompts, and the legacy opt-out flow runs for v23+ migrations where `--mode` supersedes it.

## Expected Behavior

`--interactive` is gated behind the same v23+ availability gate as the first-party/third-party modes: passing it for a v23+ Nx-equivalent target throws with a pointer to `--mode`, and combining it with `--mode=third-party` is rejected. Pre-v23 and non-Nx targets keep the legacy interactive behavior.

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/nxc-4412-766d87a4)

### DIFF
--- a/packages/nx/src/command-line/migrate/migrate.ts
+++ b/packages/nx/src/command-line/migrate/migrate.ts
@@ -955,6 +955,7 @@ export async function resolveMode(
     hasExcludeAppliedMigrations: boolean;
     installedMajor?: number | null;
     isV23Plus?: boolean;
+    interactive?: boolean;
   } = {
     hasFrom: false,
     hasExcludeAppliedMigrations: false,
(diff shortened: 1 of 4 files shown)


FrozenPandaz (approved):

packages/nx/src/command-line/migrate/migrate.ts
```diff
@@ -972,6 +973,18 @@ export async function resolveMode(
...
   const thirdPartyAvailable = installedMajor !== null && installedMajor >= 23;
 
+  // `--interactive` x-prompts let users skip optional third-party updates; the
+  // modes supersede that choice, so it is disallowed behind the same v23+ gate.
+  if (
+    context.interactive === true &&
+    context.isV23Plus === true &&
+    isNxEquivalentTarget(targetPackage, targetVersion)
+  ) {
+    throw new Error(
```
AgentEnder, line 983: How do lower versions of the CLI handle --mode? They won't have it in their yargs config, so it won't be in `--help`... we probably need more guidance in this message or a link to docs?

PR #35875, opened 2026-06-04
fix(testing): avoid executing unchanged jest configs during project graph construction

leosvelperez:
## Current Behavior

`@nx/jest/plugin` executes every matched jest config (`loadConfigFile`) on every `createNodes` invocation - including fully warm-cache runs and every daemon recompute - because the cache key embeds external files (preset, ts-jest tsconfig chain) that are only knowable after executing the config. Executing a config is the dominant cost, so project graph computation pays it for all configs every time.

## Expected Behavior

The cache key is computable without executing configs: a preliminary key (project files + lockfile + options) plus a second-level content/existence hash of the external files the inference depends on, including the locations probed during ts-jest tsconfig discovery. On a warm cache no configs are executed; only configs with a genuine miss are loaded. Invalidation semantics are unchanged.

## Benchmarks

Fabricated workspace, invoking the plugin's `createNodesV2` function directly (no daemon).

**Setup**

- 300 ts-jest projects, each with a `jest.config.ts` (relative preset `../../jest.preset.js`, `ts-jest` transform pointing at a `tsconfig.spec.json` that extends the root `tsconfig.base.json`), `project.json`, and source/spec files.
- The workspace `jest.preset.js` requires the real `jest`/`jest-config` packages to mirror the module-graph weight of `@nx/jest/preset`. node_modules modules stay require-cached across invocations, exactly as in production (`clearRequireCache` only evicts workspace files).
- Two fixtures: "canonical" (config is a single object literal, the nx-generated shape) and "with imports" (each config additionally imports a local helper module).
- Config loading mechanism: Node v24.15.0 native TypeScript type stripping - no swc/ts-node loader registration, no tsconfig-paths hooks. That is the fastest config-loading path available today, so the "before" column is measured with the best-case loader; older Node versions or configs requiring full transpilation/path mapping make every "before" load more expensive, while "after" is unaffected since it does not execute configs on warm hits.
- Apple M2 Max. 3 interleaved before/after passes; medians (warm rows aggregate 18 samples per cell, spread below +/-4%). `loads` = config modules actually executed during the call, counted via the require cache.

**Scenarios**

- cold: empty plugin cache (first ever run)
- warm recompute: repeated invocation in the same process with no file changes - what the daemon does on every recomputation
- one project changed: a single project source file modified between invocations
- fresh process, warm disk cache: new process with the on-disk plugin cache populated (non-daemon nx invocations, daemon restarts)

**Canonical configs**

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| cold | 687 ms, 300 loads | 665 ms, 300 loads | 1.0x (parity) |
| warm recompute | 187 ms, 300 loads | 115 ms, 0 loads | 1.6x |
| one project changed | 206 ms, 300 loads | 136 ms, 1 load | 1.5x |
| fresh process, warm disk cache | 328 ms, 300 loads | 186 ms, 0 loads | 1.8x |

**Configs with imports**

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| cold | 1156 ms, 600 loads | 1159 ms, 600 loads | 1.0x (parity) |
| warm recompute | 260 ms, 600 loads | 120 ms, 0 loads | 2.2x |
| one project changed | 288 ms, 600 loads | 144 ms, 2 loads | 2.0x |
| fresh process, warm disk cache | 494 ms, 600 loads | 185 ms, 0 loads | 2.7x |

**Takeaways**

- Before executes every config on every invocation; after executes none on a warm hit and exactly the changed one otherwise.
- After's warm cost is flat (~115-120 ms) regardless of config weight - the remainder is shared workspace hashing + cache serialization. Before's cost grows with config weight (187 ms -> 260 ms warm from a single extra helper import per config), so heavier real-world configs widen the gap further.
- Cold path is at parity: the two-level bookkeeping adds no measurable cost when everything must load anyway.

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/improve-jest-plugin-warm-cache-perf-cfc1bd22)

### DIFF
--- a/e2e/js/src/js-strip-types.test.ts
+++ b/e2e/js/src/js-strip-types.test.ts
@@ -119,6 +119,18 @@ describe('native Node.js TypeScript support', () => {
     // logging, which would make per-test assertions flaky based on
     // `uniq()`-randomized lib ordering. Each test does its own cleanup
     // between graph capture and assertions.
+    //
+    // Each test must also `nx reset` BEFORE writing its broken config:
+    // `runCLI` defaults to NX_DAEMON=true, so the generate call leaves a
+    // daemon watching. If the broken config is written while it's alive, the
+    // daemon recomputes the graph, executes the config in its (detached)
+    // plugin worker, and persists the plugin cache keyed on the broken
+    // content. Resetting after the write doesn't help: reset's daemon stop
+    // doesn't wait for the worker, so its cache write can land after the
+    // wipe. The daemon-less assert run then gets a warm cache hit, never
+    // executes the config, and the fallback log it asserts on is never
+    // emitted. Killing the daemon before the broken config exists makes the
+    // assert run the only process that ever executes it.
 
     it(
       'should fall back to swc/ts-node when a TS config uses an enum',
@@ -127,6 +139,9 @@ describe('native Node.js TypeScript support', () => {
         runCLI(
           `generate @nx/js:lib ${lib} --unitTestRunner=jest --no-interactive`
         );
+
+        // Kill the daemon before writing the broken config (see block comment).
+        runCLI('reset');
 
         // enum is not supported by Node native type stripping - must trigger fallback
         updateFile(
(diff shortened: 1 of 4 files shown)


leosvelperez:
Closing: this optimization regresses correctness. Skipping config execution on warm cache hits means changes that flow through untracked inputs (e.g. an out-of-project import that redirects the preset path or ts-jest tsconfig discovery) no longer invalidate the cache, whereas the current always-execute behavior picks them up. Since a jest config is an arbitrary program, any cache keyed on an enumerable input set is an approximation - module-graph tracking would shrink the gap but cannot close it (e.g. env-conditional imports). Not worth trading correctness for the perf win.

PR #35877, opened 2026-06-04
fix(react-native): avoid executing unchanged app configs during project graph construction

leosvelperez:
## Current Behavior

`@nx/expo/plugin` and `@nx/react-native/plugin` execute every matched app config (`loadConfigFile`) on every `createNodes` invocation - including fully warm-cache runs and every daemon recompute - just to decide whether the config belongs to an Expo / React Native app. The loaded config is only used as an include/exclude predicate, never for target shape. Both plugins also mutate the user-provided options object, which changes the option-derived cache hashes between the daemon's first and subsequent recomputes and forces a full cache miss.

## Expected Behavior

The include/exclude decision is cached per (project hash, config file), with a key computable without executing configs (project files + lockfile + options) - but only when the decision derives from inputs covered by that key:

- `app.json` content (pure data, cannot import or execute anything),
- sibling-file existence (`package.json` / `metro.config.js` checks),
- an `expo` dependency in the project's own `package.json`, which pins the decision regardless of what the config exports (the `@nx/expo` application generator always writes one).

A decision resting on **executing** an `app.config.js`/`app.config.ts` is never cached - executed config content can depend on inputs outside the cache key (imports from other projects, fs reads, env vars) - so those configs are re-evaluated on every run, exactly as today, while their targets (which are static) are still cached. Load errors are not cached and resurface on every run. `normalizeOptions` no longer mutates the daemon-held options object.

## Benchmarks

Fabricated workspaces, invoking the plugins' `createNodesV2` function directly (no daemon).

**Setup**

- 300 apps per fixture, each with an app config, `package.json` (expo or react-native dependency), `metro.config.js`, `project.json`, and source files.
- Four fixtures: expo with the canonical nx-generated `app.json`; expo with an `app.config.ts` that imports a local helper module (dynamic-config shape, expo dep in the app's `package.json` as generated); react-native with `app.json`; and expo with the same `app.config.ts` but no expo dep in the app's `package.json` - the only shape whose decision cannot be soundly cached.
- Node v24.15.0, Apple M2 Max. 8 interleaved before/after passes; medians (warm rows aggregate 48 samples per cell). `loads` = app config modules actually executed during the call, counted via the require cache.

**Scenarios**

- cold: empty plugin cache (first ever run)
- warm recompute: repeated invocation in the same process with no file changes - what the daemon does on every recomputation
- one project changed: a single project source file modified between invocations
- fresh process, warm disk cache: new process with the on-disk plugin cache populated (non-daemon nx invocations, daemon restarts)

**@nx/expo - `app.json` configs**

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| cold | 260 ms, 300 loads | 264 ms, 300 loads | 1.0x (parity) |
| warm recompute | 140 ms, 300 loads | 95 ms, 0 loads | 1.5x |
| one project changed | 150 ms, 300 loads | 121 ms, 1 load | 1.2x |
| fresh process, warm disk cache | 203 ms, 300 loads | 144 ms, 0 loads | 1.4x |

**@nx/expo - `app.config.ts` configs (with an import, expo dep in package.json)**

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| cold | 1249 ms, 600 loads | 1271 ms, 600 loads | 1.0x (parity) |
| warm recompute | 232 ms, 600 loads | 89 ms, 0 loads | 2.6x |
| one project changed | 242 ms, 600 loads | 121 ms, 2 loads | 2.0x |
| fresh process, warm disk cache | 490 ms, 600 loads | 146 ms, 0 loads | 3.4x |

**@nx/react-native - `app.json` configs**

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| cold | 237 ms, 300 loads | 230 ms, 300 loads | 1.0x (parity) |
| warm recompute | 135 ms, 300 loads | 85 ms, 0 loads | 1.6x |
| one project changed | 147 ms, 300 loads | 119 ms, 1 load | 1.2x |
| fresh process, warm disk cache | 194 ms, 300 loads | 145 ms, 0 loads | 1.3x |

**@nx/expo - `app.config.ts` configs without an expo dep (decision not cacheable)**

| Scenario | Before | After | Ratio |
| --- | --- | --- | --- |
| cold | 1388 ms, 600 loads | 1380 ms, 600 loads | 1.0x (parity) |
| warm recompute | 232 ms, 600 loads | 229 ms, 600 loads | 1.0x (parity) |
| one project changed | 241 ms, 600 loads | 260 ms, 600 loads | 0.9x |
| fresh process, warm disk cache | 546 ms, 600 loads | 533 ms, 600 loads | 1.0x (parity) |

**Takeaways**

- Before executes every app config on every invocation; after executes none on a warm hit and exactly the changed project's config otherwise - for every shape whose decision is soundly cacheable (everything the nx generators produce).
- After's warm cost is flat (~85-95 ms) regardless of config weight - the remainder is shared workspace hashing + cache serialization. Before's cost grows with config weight (140 ms for `app.json` -> 232 ms for `app.config.ts` with a single helper import), so heavier real-world dynamic configs widen the gap further.
- Cold path is at parity: the decision bookkeeping adds no measurable cost when everything must load anyway.
- The non-cacheable shape (js/ts config whose project does not declare expo) keeps master behavior and cost across all scenarios.

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/improve-react-native-warm-cache-perf-fc70c5d9)

### DIFF
--- a/packages/expo/plugins/plugin.spec.ts
+++ b/packages/expo/plugins/plugin.spec.ts
@@ -1,0 +1,318 @@
+import { CreateNodesContext } from '@nx/devkit';
+import * as configUtils from '@nx/devkit/internal';
+import { TempFs } from 'nx/src/internal-testing-utils/temp-fs';
+import { setupWorkspaceContext } from 'nx/src/utils/workspace-context';
+import { resolve } from 'path';
+import { createNodesV2, ExpoPluginOptions } from './plugin';
+
+jest.mock('nx/src/utils/cache-directory', () => ({
+  ...jest.requireActual('nx/src/utils/cache-directory'),
+  workspaceDataDirectory: 'tmp/project-graph-cache',
+}));
+
+describe('@nx/expo/plugin', () => {
+  const createNodesFunction = createNodesV2[1];
+  const options: ExpoPluginOptions = {};
+  let context: CreateNodesContext;
+  let tempFs: TempFs;
+  let cwd: string;
+  let loadConfigFileSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    tempFs = new TempFs('expo-plugin');
+    cwd = process.cwd();
+    process.chdir(tempFs.tempDir);
+    context = {
+      nxJsonConfiguration: {
+        namedInputs: {
+          default: ['{projectRoot}/**/*'],
+          production: ['!{projectRoot}/**/*.spec.ts'],
+        },
+      },
+      workspaceRoot: tempFs.tempDir,
+    };
+    // Spy with call-through: configs still load, we just count executions.
+    loadConfigFileSpy = jest.spyOn(configUtils, 'loadConfigFile');
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+    jest.resetModules();
+    tempFs.cleanup();
+    process.chdir(cwd);
+  });
+
+  // Absolute path as the plugin computes it (join(workspaceRoot, file)).
+  const abs = (file: string) => resolve(tempFs.tempDir, file);
+  const loadedConfigPaths = () =>
+    loadConfigFileSpy.mock.calls.map((call) => call[0] as string);
+
+  it('should create nodes for an Expo app config', async () => {
+    await tempFs.createFiles({
+      'package-lock.json': '{}',
+      'apps/mobile/app.json': '{ "expo": { "name": "mobile" } }',
+      'apps/mobile/package.json': '{}',
+      'apps/mobile/metro.config.js': '',
+    });
+    setupWorkspaceContext(tempFs.tempDir);
+
+    const results = await createNodesFunction(
+      ['apps/mobile/app.json'],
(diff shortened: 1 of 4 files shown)


leosvelperez:
Closing for the same reason as #35875: the trade is not worth it. The warm-run win here is modest and only applies when the include/exclude decision is pinned by in-key inputs (app.json data or an expo dependency in package.json); JS/TS configs without a pin still execute on every run, so the general case is not improved. Meanwhile the decision cache adds risk at the edges: the filter reads disk truth (readdirSync, package.json) while the cache key only hashes non-ignored files, so a gitignored sibling can leave a stale decision, and a pinned config that starts failing through an out-of-key input no longer surfaces its error at graph construction time. Little benefit, some risk.

PR #35904, opened 2026-06-08
feat(expo): support Expo SDK 56

### LINKED ISSUE #35714: feat(expo): support Expo SDK 56
Expo SDK 56 has been out for a while and @nx/expo doesn't yet support it. Requesting an upgrade so projects can use SDK 56 with Nx.

itsjithinv:
## Current Behavior

`@nx/expo` supports Expo SDK 53–55 (latest/default 55) after the recent
version-lane refactor. It does not yet support **SDK 56**.

Separately, the plugin's Metro and executor wiring predates Expo's SDK 55
`@expo/metro` repackaging, so generated **SDK 55+** apps fail at runtime even
on the existing lanes:

- `expo start` / web bundling crashes with
  `TypeError: Cannot read properties of undefined (reading 'transformFile')` -
  `withNxMetro` merges via the standalone `metro-config` and forces
  `projectRoot` to the workspace root, which collides with Expo's bundled
  `@expo/metro` and breaks the babel transformer's `.babelrc.js` resolution.
- `nx prebuild` / `start` executors throw `Cannot find module
  '@expo/cli/build/bin/cli.js'` because SDK 55+ ships `@expo/cli` with an
  `exports` map (`"./*": "./*.js"`), so the hardcoded bin subpath no longer
  resolves.

## Expected Behavior

Adds an Expo **SDK 56** install lane (RN 0.85.3, React 19.2, `@expo/cli`
~56.1.14, `@expo/metro-config` ~56.0.13, `jest-expo` ~56.0.4) as the default
for new projects, on top of the existing 53–55 lanes, and fixes the SDK 55+
runtime wiring (benefits 55 and 56):

- `withNxMetro` and the Nx resolver prefer `@expo/metro/metro-config` /
  `@expo/metro/metro-resolver` (fallback to the standalone packages for 53/54),
  and no longer override `projectRoot` to the workspace root on SDK 55+.
- Executors resolve the Expo CLI via the stable `expo/bin/cli` entry instead of
  `@expo/cli/build/bin/cli`.
- New SDK 55+ apps no longer install standalone `metro-config`/`metro-resolver`
  or `@expo/metro-config` directly (the generated `metro.config.js` extends
  `expo/metro-config`); those packages are now optional peer dependencies.
- Adds an AI upgrade-instructions migration for moving workspaces to SDK 56.

Verified by generating a workspace from a locally-published build:
`expo start --web` bundles successfully and `expo-doctor`'s
"`@expo/metro-config` installed directly" check passes.

## Related Issue(s)

Fixes #35714


itsjithinv:
@FrozenPandaz @leosvelperez  Can you guys review this PR and merge it if everything looks good? The latest Expo 56 features are still missing in Nx.

packages/expo/migrations.json
```diff
@@ -1,5 +1,11 @@
 {
   "generators": {
+    "update-23-0-0-create-ai-instructions-for-expo-56": {
+      "version": "23.0.0-beta.25",
+      "cli": "nx",
+      "description": "Create AI migration instructions for Expo SDK 56 upgrade",
+      "prompt": "./src/migrations/update-23-0-0/ai-instructions-for-expo-56.md"
```
FrozenPandaz: This has to change to point to the dist one now since I've recently changed how we build/package our packages.

```suggestion
      "prompt": "./dist/src/migrations/update-23-0-0/ai-instructions-for-expo-56.md"
```

packages/expo/migrations.json
```diff
@@ -1,5 +1,11 @@
 {
   "generators": {
+    "update-23-0-0-create-ai-instructions-for-expo-56": {
+      "version": "23.0.0-beta.25",
```
FrozenPandaz: Please bump this version to `23.1.0-beta.1` or whatever is after the current `next` version.

packages/expo/migrations.json
```diff
@@ -1,5 +1,11 @@
 {
   "generators": {
+    "update-23-0-0-create-ai-instructions-for-expo-56": {
```
FrozenPandaz: Please add new migration definitions to the bottom of the object declaration as we normally do

packages/expo/plugins/metro-resolver.ts
```diff
@@ -11,16 +11,25 @@ import { workspaceRoot } from '@nx/devkit';
...
  */
 function getMetroResolver() {
   if (!metroResolver) {
     try {
-      metroResolver = require('metro-resolver');
-    } catch (error) {
-      throw new Error(
-        'metro-resolver is required but not installed. Please install metro-resolver >= 0.82.0'
-      );
+      metroResolver = require('@expo/metro/metro-resolver');
```
FrozenPandaz, line 24: This would mean that `@nx/expo` needs an optional peer dependency on `@expo/metro`... even though `expo` does have a dependency on `@expo/metro`, some package managers like `pnpm` do not allow requiring into it.

packages/expo/src/generators/application/files/base/metro.config.js.template
```diff
@@ -1,6 +1,9 @@
 const { withNxMetro } = require('@nx/expo');
-const { getDefaultConfig } = require('@expo/metro-config');
-const { mergeConfig } = require('metro-config');
+<% if (usesExpoMetro) { %>// Expo SDK 55+ ships Metro via `@expo/metro`. `getDefaultConfig` and
+// `mergeConfig` must come from the Expo-provided Metro instance.
+const { getDefaultConfig } = require('expo/metro-config');
+const { mergeConfig } = require('@expo/metro/metro-config');<% } else { %>const { getDefaultConfig } = require('@expo/metro-config');
```
FrozenPandaz, line 5: Because we generate an import of `@expo/metro/metro-config`, we should also add `@expo/metro` to the workspace dependencies here so that it works with `pnpm` for the same reason as the above comment.

packages/expo/migrations.json
FrozenPandaz: 1. We should have a `packageJsonUpdates` block to update expo dependencies to 56 as we have before. Note the `requires`
2. This migration itself should also have a `requires` block which will gate it on if the user has updated to 56.

The expected behavior is for users to be able to update Nx (and `@nx/expo`) to 23 without necessarily migrating expo itself which modifies application behavior to 56. That would mean that this migration should not run if they don't do migrate to 56. Users can come back and run the optional migrations when they have more time for testing when `expo` should be updated as well as this migration should be run.

The pattern should be present for other updates in this file for you to follow.

FrozenPandaz (changes requested):
Thank you so much for taking initiative and contributing this update! I left some comments but hopefully they're not too major. Hopefully you can get to these soon and we can release this as part of 23.1.0! 🤞

packages/expo/migrations.json
itsjithinv: Done - bumped to 23.1.0-beta.1.

packages/expo/migrations.json
itsjithinv: Done. moved the migration definition to the bottom of the generators object.

packages/expo/plugins/metro-resolver.ts
itsjithinv: Good catch. Added @expo/metro as an optional peerDependency (with peerDependenciesMeta.optional) on @nx/expo, and made generated SDK 55+ apps install @expo/metro directly so require('@expo/metro/metro-config') resolves under pnpm (no requiring into transitive deps).

packages/expo/src/generators/application/files/base/metro.config.js.template
itsjithinv: Done. SDK 55+ apps now get @expo/metro as a direct dependency (via the app package.json template + ensureDependencies), so the generated metro.config.js import resolves under pnpm. Standalone metro-config/metro-resolver are no longer installed for 55+.

packages/expo/migrations.json
itsjithinv: Added a packageJsonUpdates lane for Expo SDK 56 (chains 55->56) and gated the migration with 
```
requires: { expo: ">=55.0.0 <56.0.0" }
``` 
I also added code migrations (gated requires: expo >= 56.0.0) that rewrite `metro.config.js` to @expo/metro, swap the standalone Metro packages for @expo/metro, and update the Jest test-setup - so existing apps don't regress at bundle time when migrating to 56.

packages/expo/migrations.json
itsjithinv: Done! Updated the prompt to `./dist/src/migrations/update-23-1-0/ai-instructions-for-expo-56.md`. Thanks for the heads-up on the `build/packaging` change.

itsjithinv:
Thanks for the review @FrozenPandaz! Addressed all the feedback. Also merged latest master and added migrations so existing 54/55 workspaces upgrade cleanly. Ready for another look.

FrozenPandaz (approved):

PR #35942, opened 2026-06-10
fix(core): evaluate migration requires across journey

leosvelperez:
## Current Behavior

During migration collection, `requires` gates for dependencies with planned updates are evaluated against the final planned endpoint only. Bounded ranges such as `>=9.0.0 <10.0.0` can drop migrations that are needed when a single `nx migrate` run crosses that intermediate range.

## Expected Behavior

`requires` gates now intersect the required range with the dependency journey from the installed version, or `--from` override, to the planned version. Dependencies without a planned update keep the existing installed version point check. `--exclude-applied-migrations` uses the same journey-aware reasoning for bounded ranges, and plugin-author docs describe the semantics, including the non-boundary-crossing caveat.

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/nxc-4549-33e9fa7f)

### DIFF
--- a/astro-docs/src/content/docs/extending-nx/migration-generators.mdoc
+++ b/astro-docs/src/content/docs/extending-nx/migration-generators.mdoc
@@ -96,3 +96,29 @@ If you just need to change dependency versions, you can add some configuration o
   }
 }
 ```
+
+## Gate migrations and package updates by dependency version
+
+Migration entries and `packageJsonUpdates` entries can use `requires` to apply only when another package version is relevant to the workspace.
+
+```json
+// migrations.json
+{
+  "generators": {
+    "migrate-storybook-v9": {
+      "version": "21.2.0",
+      "description": "Updates Storybook v9 configuration",
+      "implementation": "./src/migrations/migrate-storybook-v9/migrate-storybook-v9",
+      "requires": {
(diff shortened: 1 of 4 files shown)


leosvelperez:
Closing. After investigation the premise doesn't hold - full rationale in NXC-4549.

In short: migration generators are intended to be gated by the plugin's version-transition window plus an unbounded `requires` floor (third-party readiness). That's the prevailing pattern (48/52 generator `requires` are floors), not a workaround for broken capping - e.g. `migrate-to-vitest-3` (`>=3.0.0`) and `migrate-to-vitest-4` (`>=4.0.0`) both correctly run on a vitest 2 -> 4 journey under today's point-check.

The reported storybook breakage is a mis-authored cap: `update-21-2-0-migrate-storybook-v9` uses `>=9.0.0 <10.0.0` while its sibling `migrate-storybook-v10` correctly floors at `>=10.0.0`. The fix is to floor the few mis-capped generator entries (storybook v9 x2, react-native, expo), not to change the engine to tolerate caps - which would alter `requires` semantics for every external plugin's capped entries, weaken migration idempotency, and still not deliver the cohort-gating used to motivate it.

PR #35965, opened 2026-06-11
chore(repo): exclude .NET assets cache from write tracking

AgentEnder:
## Current Behavior

When a .NET task's dependencies are restored from the Nx cache, MSBuild overwrites stale `*.assets.cache` intermediate files, which the io-trace sandboxing reports as unexpected write violations even though this is known-safe behavior.

## Expected Behavior

`**/*.assets.cache` writes are excluded from sandbox write tracking, so these known-safe overwrites no longer surface as violations. This mirrors nrwl/ocean#11768, which added the same pattern to the io-trace daemon's built-in `defaultWritePatterns`.

## Related Issue(s)

Fixes #NXC-4521

Supersedes #35920 (which used a read exclusion with a narrower glob; the violation is a write, so this matches the ocean defaults instead).

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/A-recent-PR-added-write-exclusions-for--.assets.cache---duplicate-this-7e1b8711)

### DIFF
--- a/packages/nx/src/command-line/watch/command-object.ts
+++ b/packages/nx/src/command-line/watch/command-object.ts
@@ -52,6 +52,18 @@ function withWatchOptions(yargs: Argv) {
         alias: 'g',
         hidden: true,
       })
+      .option('include', {
+        type: 'string',
+        coerce: parseCSV,
+        description:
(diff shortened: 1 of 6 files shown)


AgentEnder:
Closing — the exclusion belongs in the io-trace defaults in nrwl/ocean (write side landed in nrwl/ocean#11768; read side is being added there as well), not in this repo's config.

PR #35973, opened 2026-06-12
chore(gradle): bump gradle project graph plugin version to 0.1.22

FrozenPandaz:
## Current Behavior

Workspaces reference the `dev.nx.gradle.project-graph` Gradle plugin at version `0.1.21`.

## Expected Behavior

Bump the plugin to `0.1.22`. This release carries the deterministic project-configuration-hash fix from #35972 (sorting `options.includeDependsOnTasks` so the `ProjectConfiguration` hash no longer drifts between JVM runs).

The bump updates the version constant, the plugin's `build.gradle.kts`, and adds a migration (`change-plugin-version-0-1-22`, gated at `23.0.0-rc.2`) that updates existing workspaces' version catalogs and `build.gradle(.kts)` files on `nx migrate`.

> Note: the published `0.1.22` plugin artifact must include #35972 before this is released.

## Related Issue(s)

Follow-up to #35972 (the source fix).

### DIFF
--- a/packages/gradle/migrations.json
+++ b/packages/gradle/migrations.json
@@ -159,6 +159,13 @@
       "description": "Rename imports of `createNodesV2` from `@nx/gradle` and `@nx/gradle/plugin-v1` to the canonical `createNodes` export.",
       "implementation": "./dist/src/migrations/update-23-0-0/migrate-create-nodes-v2-to-create-nodes",
       "documentation": "./dist/src/migrations/update-23-0-0/migrate-create-nodes-v2-to-create-nodes.md"
+    },
+    "change-plugin-version-0-1-22": {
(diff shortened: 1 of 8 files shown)


lourw (approved):

FrozenPandaz:
Now also includes the `fix(gradle): do not cache publishToMavenLocal tasks` change (was #35976). The publish-to-mavenLocal sub-tasks were cacheable, so under distributed CI the local publish was a cache hit on the e2e agent and `~/.m2` never got the new plugin version — which is why this bump's gradle e2e failed. The `isCacheable` change + `project.json` override make those tasks always run, so 0.1.22 actually lands in the e2e agent's `~/.m2`.

PR #35977, opened 2026-06-12
fix: angular-rspack loader error stats

SyMind:
## Summary

Fixes the Angular Rspack plugin error formatting path for loader errors that prevent sealing. Rspack now defaults many stats fields off, so calling `stats.toJson()` without explicit stats options can omit the error data that `statsErrorsToString` expects.

This change uses a minimal local stats options object based on the fields read by `statsErrorsToString`: top-level and child errors, plus the formatting flags for colors, error details, and error stacks.

## Validation

- `corepack pnpm exec prettier --check packages/angular-rspack/src/lib/plugins/angular-rspack-plugin.ts`
- `NODE_OPTIONS='--conditions=@nx/nx-source' corepack pnpm exec vitest --config vitest.config.mts run` from `packages/angular-rspack`

### DIFF
--- a/packages/angular-rspack/src/lib/plugins/angular-rspack-plugin.ts
+++ b/packages/angular-rspack/src/lib/plugins/angular-rspack-plugin.ts
@@ -154,9 +154,15 @@ export class AngularRspackPlugin implements RspackPluginInstance {
       compilation.hooks.afterSeal.tapAsync(PLUGIN_NAME, (callback) => {
         if (!watchRunInitialized && compilation.errors.length > 0) {
           const stats = compilation.getStats();
+          const statsOptions = {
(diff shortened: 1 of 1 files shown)


leosvelperez:
Thanks for the contribution. We could not reproduce the reported missing error data: `stats.toJson()` includes compilation errors in both the supported Rspack 1.x and current Rspack 2.x behaviors. This change also bypasses our normal `verbose` formatting options, causing non-verbose builds to emit untrimmed, uncolored error output.

Please open an issue with a minimal reproduction, exact Nx and Rspack versions, and expected versus actual output if this problem still occurs. We can reconsider a tested fix from that reproduction.

PR #35983, opened 2026-06-12
fix(docker): skip image tagging when resolved docker version is empty

christian-calabrese:
## Current Behavior
When `release.docker.versionSchemes` uses `{versionActionsVersion}` and a project has no resolved version action value, docker release can compute an empty tag and attempt:

`docker tag <local-image> <repo>:`

This fails with `invalid reference format`.

## Expected Behavior
If no docker version can be resolved and no `NX_DOCKER_IMAGE_REF` override is provided, docker tagging should be skipped instead of attempting an invalid `repo:` tag.

## Changes
- Add guard in `packages/docker/src/release/version-utils.ts` before `updateProjectVersion(...)`
- Return `{ newVersion: null, logs: [...] }` when resolved version is empty
- Add test in `packages/docker/src/release/version-utils.spec.ts`:
  - `skips image tagging when resolved version is empty`

## Why this is safe
- Does not affect normal flows where version is resolved
- Does not affect explicit `dockerVersion`
- Does not affect `NX_DOCKER_IMAGE_REF` override
- Prevents invalid docker CLI invocation in edge case

## Repro
1. Configure docker version scheme with `{versionActionsVersion}`
2. Run release for a docker project where `versionActionsVersion` resolves to empty
3. Current behavior: attempts `docker tag ... <repo>:` and fails
4. With this PR: skips tagging for that project and logs the reason


AgentEnder:
Closing this, because the guard you wrote is now on master. #36542 added the same early return in `version-utils.ts`, so a scheme containing `{versionActionsVersion}` with no resolved version returns `{ newVersion: null, logs: [...] }` instead of reaching `docker tag`. Your repro no longer produces the invalid tag.

Thanks for the writeup, and sorry this sat long enough to be overtaken. It was a real bug and you diagnosed it correctly.

<details>
<summary>

</summary>

> Superseded by #36542, `fix(docker): skip unchanged version action projects`. The guard sits at `packages/docker/src/release/version-utils.ts:50-62`, ahead of the `updateProjectVersion(...)` call on line 72, and master carries an equivalent test at `version-utils.spec.ts:82-104` covering the same scheme and the same assertion. Only the log wording differs from what this PR proposed.
>
> One difference worth recording rather than hiding: master checks `versionActionsVersion == null`, so a literal empty string would still pass through, where this PR checked for emptiness. That gap is not reachable today, because the only caller passes `projectVersionData.newVersion`, which is a semver or null. If a future caller can pass an empty string, this is where it would bite.

</details>

PR #36006, opened 2026-06-15
feat(linter)!: drop eslint v8 support

leosvelperez:
## Current Behavior

Nx supported both ESLint v8 and v9. The `@nx/eslint` runtime, generators, executors, and inferred plugin carried v8-specific branches, version floors allowed v8-only ranges, and `useFlatConfig` chose flat vs eslintrc purely by the installed ESLint version - which could select flat config on an eslintrc workspace and crash generators.

## Expected Behavior

ESLint v8 support is removed and Nx targets ESLint v9+. Flat config is the default for new workspaces, while existing eslintrc workspaces stay supported: `useFlatConfig` now respects a root flat/eslintrc config file and the `ESLINT_USE_FLAT_CONFIG` env var. Version floors, the lockfile, and docs move to v9+. Generator specs across the linting-capable plugins assert flat config by default and each retains at least one eslintrc test.

BREAKING CHANGE: ESLint v8 is no longer supported. Nx requires ESLint v9 or later.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/drop-eslint-v8-ad08cc1c)


packages/eslint/src/plugins/plugin.ts
```diff
@@ -507,13 +501,12 @@ function buildEslintTargets(
...
+  // Supported ESLint versions (v9+) default to flat config, so only set the env
+  // var when the workspace still uses eslintrc, to force the legacy loader.
   const useFlatConfig = eslintConfigs.some((config) => isFlatConfig(config));
-  // Flat config is default for 9.0.0+
-  const defaultSetting = gte(eslintVersion, '9.0.0');
-  if (useFlatConfig !== defaultSetting) {
+  if (!useFlatConfig) {
     targetConfig.options.env = {
-      ESLINT_USE_FLAT_CONFIG: useFlatConfig ? 'true' : 'false',
+      ESLINT_USE_FLAT_CONFIG: 'false',
```
FrozenPandaz, line 509: Do we have to set this to `false`?

packages/eslint/src/utils/config-file.ts
```diff
@@ -3,18 +3,12 @@ import { basename, dirname, join, resolve } from 'path';
...
 
-export const ESLINT_OLD_CONFIG_FILENAMES = [
-  '.eslintrc',
-  '.eslintrc.js',
-  '.eslintrc.cjs',
-  '.eslintrc.yaml',
-  '.eslintrc.yml',
-  '.eslintrc.json',
-];
+export const ESLINT_OLD_CONFIG_FILENAMES = eslintrcFilenames;
```
FrozenPandaz, line 11: Do we still need the old config file names?

FrozenPandaz (changes requested):

packages/eslint/src/plugins/plugin.ts
leosvelperez: We have to when the repo is not using Flat Config.

packages/eslint/src/utils/config-file.ts
leosvelperez: Yes, we still have some tests covering the eslintrc config. The majority of the tests were switched in the PR to cover Flat Config while keeping only a few tests for the eslintrc config.

FrozenPandaz (approved):

PR #36030, opened 2026-06-17
fix(core): use default import for chalk

arturovt:
chalk 4.x is a CJS module that exports the chalk instance via `module.exports`. Using `import * as chalk` caused tslib to wrap it with `__importStar`, which works but differs from how the module is meant to be consumed. Switching to a default import uses `__importDefault` instead, matching the actual shape of the export and eliminating the need for downstream patches.


leosvelperez:
Closing as a duplicate of https://github.com/nrwl/nx/pull/35523, which was already merged.

PR #36033, opened 2026-06-17
docs(nx): clarify forwardAllArgs default for interpolated commands

### LINKED ISSUE #34756: Docs improvedment for Interpolating Args of Run-Commands
### Documentation issue

- [ ] Reporting a typo
- [ ] Reporting a documentation bug
- [ ] Documentation improvement
- [x] Documentation feedback

### Is there a specific documentation page you are reporting?

https://nx.dev/docs/reference/nx/executors

### Description

Currently, the `forwardAllArgs` option for `run-commands` is enabled by default, so that for a target like this:
```
      "example": {
        "options": {
          "commands": [
            "echo {args.type}",
            "echo"
          ],
          "parallel": false
        }
      }
```

with a command `nx run example --type=foobar`, it will output like this:
```
> echo test

test
> echo --type=test

--type=test
```

the `--type=test` was inserted into every command by default.

I think that it will better and intuitive that the docs also shows an example that disable the auto interpolating behaviour like this target:
```
      "example": {
        "options": {
          "commands": [
            "echo {args.type}",
            "echo"
          ],
          "forwardAllArgs": false,
          "parallel": false
        }
      }
```

### Additional context

I noticed the discussion in this issue: https://github.com/nrwl/nx/issues/3335
It seems that the forwardAllArgs was intended to be false by default?
> I would also agree that it's not reasonable to forward when using commands (plural) so introducing a forwardAllArgs option and defaulting it to false sounds good to me 👍 .

titan-ron:
## Description

Addresses the documentation feedback in #34756.

In the **Interpolating Args** example for the `nx:run-commands` executor, it isn't obvious that commands which don't contain `{args.*}` interpolation still receive **all** forwarded arguments by default. This trips people up when a multi-command list contains a mix of interpolated and non-interpolated commands.

This PR adds a short note to the *Interpolating Args* tab that:

- explains the default forwarding behavior for non-interpolated commands in a `commands` list, with a concrete `echo done --name=example` example, and
- shows how to opt out per-command via `forwardAllArgs: false`.

## Current Behavior

The docs only document `forwardAllArgs` in the *Arguments forwarding* tab, so users following the interpolation example don't realize their other commands receive the forwarded args.

## Expected Behavior

The interpolation example calls out the default and shows how to disable it.

Closes #34756

### DIFF
--- a/packages/nx/docs/run-commands-examples.md
+++ b/packages/nx/docs/run-commands-examples.md
@@ -90,6 +90,39 @@ or simply with:
 nx run frontend:create-script --name=example
 ```
 
+Keep in mind that any command in the list that does not include interpolation will still have all passed arguments forwarded to it by default. For example, given:
+
+```json
+"create-script": {
+    "executor": "nx:run-commands",
+    "options": {
+        "commands": [
+          "echo {args.name}",
+          "echo done"
+        ],
+        "parallel": false
+    }
+}
+```
+
+running `nx run frontend:create-script --name=example` also appends `--name=example` to the second command, executing `echo done --name=example`. Set `forwardAllArgs` to `false` on the commands that should not receive the arguments:
+
+```json
+"create-script": {
+    "executor": "nx:run-commands",
+    "options": {
+        "commands": [
+          "echo {args.name}",
+          {
+            "command": "echo done",
+            "forwardAllArgs": false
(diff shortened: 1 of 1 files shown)


titan-ron:
Recreating this PR with a signed commit.

PR #36077, opened 2026-06-22
feat(core): show a performance report at the end of every run

FrozenPandaz:
## Current Behavior

After a run, Nx prints the cached-task summary but gives no insight into how the
run's wall-clock time was spent or how to make it faster — no breakdown of the
critical-path floor versus parallelism contention, and no actionable guidance.

## Expected Behavior

Every run ends with a concise performance report:

- **Headline stats** — run duration, cache result (hit rate, or `Skipped
  (--skip-nx-cache)`), critical path (the dependency floor: how long the run
  would take with unlimited slots), and recoverable time (wall-clock lost to slot
  contention — recoverable with a higher `--parallel` or more machines).
- **Targeted recommendations**, one per lever and ordered cheapest-action-first:
  speed up/split the longest critical-path tasks (listed inline), raise
  `--parallel`, distribute with Nx Agents, enable remote cache, or drop
  `--skip-nx-cache`. Only the levers that actually apply to the run are shown.
- **A docs link** — a clickable OSC 8 hyperlink where supported, a plain
  auto-linked URL in CI, carrying a `utm=performance-report` tag.
- **Where it renders** — in the Terminal UI, inside the exit-countdown popup
  (pressing `q` mid-run still shows the original exit dialog); otherwise (non-TUI,
  or a single task) it's flushed to the terminal after the run summary. The report
  is delivered exactly once.

## Related Issue(s)

N/A

## Notes

- The analysis (`PerformanceAnalysis`) is a **pure function of the timings the
  lifecycle collects** — no coupling to the orchestrator/scheduler. It derives the
  occupancy timeline, the critical path, and the overhead split entirely from task
  start/end timestamps.
- The report is built once at the end and delivered either through the native TUI
  exit popup or a terminal flush, deduped so it never prints twice.


e2e/release/src/conventional-commits-config.test.ts
AgentEnder, line 1: Follow up: Lets move the `expect.addSnapshotSerializer` to a global setup file or something...

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -1,21 +1,178 @@
...
+/// wrapped across rows — a clickable OSC 8 hyperlink to `href`. ratatui's text
+/// path strips the escape framing from cells (ratatui#1028), so instead we find
+/// the rendered text and replace each per-row run with one self-contained OSC 8
+/// cell (sequence + `CellDiffOption::ForcedWidth` = the run's width), blanking the
+/// rest of the run. One self-contained cell per row keeps every link fragment
+/// robust to incremental diffing. Locating is by SCANNING the buffer (not
+/// re-deriving ratatui's wrap/scroll), matching with whitespace collapsed so a
+/// wrap point — its space turned into trailing padding plus a row break — still
+/// matches; if `visible` isn't found (scrolled out of view, or its rec isn't
+/// shown) nothing is injected. Requires ratatui-core >= 0.1.2 for ForcedWidth.
```
AgentEnder: Trim these ups?

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -1,21 +1,178 @@
...
+fn inject_osc8(f: &mut Frame<'_>, inner_area: Rect, visible: &str, href: &str) {
```
AgentEnder, line 40: Lets move away from OSC 8 in the mouse capture work... this is probably fine for now though... I just don't love us having to look for a pre-trimmed string like this.

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -1,21 +1,178 @@
...
+            if let Some(cell) = f.buffer_mut().cell_mut((col, row)) {
+                cell.set_symbol(" ");
+            }
+        }
+    }
+}
+
+/// Format a millisecond duration like the TS `formatDuration` (e.g. "470ms",
+/// "13.4s", "1m 30s"), so the popup matches the terminal report.
+fn format_duration(ms: f64) -> String {
```
AgentEnder: Can we have a shared util here? We do this in the tasks_list too

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -1,21 +1,178 @@
...
+    let seconds = (ms / 100.0).round() / 10.0;
+    if seconds >= 60.0 {
+        let total = (ms / 1000.0).round() as i64;
+        return format!("{}m {}s", total / 60, total % 60);
+    }
+    format!("{:.1}s", seconds)
+}
+
+/// Append "s" unless `count` is 1 (mirrors the TS `pluralize`; regular plurals).
+fn pluralize(count: u32, noun: &str) -> String {
```
AgentEnder: Ditto, we should probably move to a util

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -24,6 +181,11 @@ pub struct CountdownPopup {
     scrollbar_state: ScrollbarState,
     content_height: usize,
     viewport_height: usize,
+    /// The run report shown above the hint text (None until set).
+    summary: Option<PerformanceSummaryPayload>,
+    /// When pinned, the auto-exit countdown is stopped and the popup stays open
+    /// (e.g. the user scrolled the report) until they explicitly quit.
```
AgentEnder:
```suggestion
```

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -36,9 +198,103 @@ impl CountdownPopup {
             scrollbar_state: ScrollbarState::default(),
             content_height: 0,
             viewport_height: 0,
+            summary: None,
+            pinned: false,
         }
     }
 
+    /// Set the run report shown above the hint text. The visual is built from
+    /// these stats in {@link build_report_lines}.
```
AgentEnder:
```suggestion
```

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -36,9 +198,103 @@ impl CountdownPopup {
...
+            return Vec::new();
+        };
+        let mut lines: Vec<Line<'static>> = Vec::new();
+
+        lines.push(stat_line(
+            "Run duration",
+            format_duration(s.run_duration_ms),
+        ));
+        // Cache sits right under run duration (same section): the report is short
+        // enough that a separate cache section just looked stranded.
```
AgentEnder:
```suggestion
```

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -117,18 +405,83 @@ impl CountdownPopup {
...
+        let report = self.build_report_lines();
+        let has_report = !report.is_empty();
+        if has_report {
+            content.extend(report);
+            // The docs link is a generic recommendation, rendered as a "- " item
+            // directly under the same "Recommendations:" header (no blank line).
+            // A non-linked bullet prefix, then the label which the OSC 8 injection
+            // below turns into a clickable hyperlink. If injection is skipped (e.g.
+            // the popup is too narrow for the label), this stays as plain text.
+            url_line_index = Some(content.len());
```
AgentEnder: trim comments

packages/nx/src/native/tui/components/countdown_popup.rs
```diff
@@ -145,71 +498,95 @@ impl CountdownPopup {
...
-                Span::styled(
-                    // NOTE: I tried OSC 8 sequences here but they broke the layout, see: https://github.com/ratatui/ratatui/issues/1028
-                    "https://nx.dev/terminal-ui",
-                    Style::default().fg(THEME.info),
-                ),
-            ]),
+        // Title. With a report it names the dialog ("Performance Report"); while
+        // the auto-exit countdown runs it adds "— exiting in N..." (dropped once
+        // pinned). Without a report (mid-run quit) it's the original countdown
+        // dialog: just "Exiting in N...".
```
AgentEnder:
```suggestion
```

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+    let entry = this.timings.get(taskId);
+    if (!entry) {
+      entry = { continuous: this.taskGraph.tasks[taskId]?.continuous ?? false };
+      this.timings.set(taskId, entry);
+    }
+    return entry;
+  }
+
+  /** Hashing windows used to attribute waits to hashing. Overridable in tests. */
+  protected hashWindows(): Array<[number, number]> {
```
AgentEnder: nit: Why protected? Lets just make this private if theres not a reason so its a bit less distracting :)

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+    return collectHashWindows();
+  }
+
+  /** Whether the cache was bypassed this run. Overridable in tests. */
+  protected cacheSkipped(): boolean {
+    return (
+      this.skipNxCacheOption ||
+      process.env.NX_SKIP_NX_CACHE === 'true' ||
+      process.env.NX_DISABLE_NX_CACHE === 'true'
```
AgentEnder: Is there some kind of helper here? I don't want us to be reading the env vars directly in more than one spot.

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+    );
+  }
+
+  /** Whether a remote (Nx Cloud) cache is active. Overridable in tests. */
+  protected remoteCacheEnabled(): boolean {
+    try {
+      return isNxCloudUsed(readNxJson());
```
AgentEnder: Do we need to re-read nx.json here? I feel like we probably already have it when we create lifecycles

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+    } catch {
+      return false;
+    }
+  }
+
+  /** Whether the run is in CI. Overridable in tests. */
+  protected isCI(): boolean {
+    return !!process.env.CI;
+  }
```
AgentEnder: We have a different function for checking if ci is enabled or not that checks way more env vars than just CI

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+
+  /** Whether the run is already distributing across machines (Agents/DTE). Overridable in tests. */
+  protected distributingTasks(): boolean {
+    return !!process.env.NX_CLOUD_DISTRIBUTED_EXECUTION_AGENT_COUNT;
```
AgentEnder: Future work... it'd be kinda nice if the env var read here came from the light client. Something like isCloudEnabled && lightClient.isDistributedExecution()

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+   * than sampling one instant, so it's robust to when tasks hand off.
+   */
+  private buildSegments(
+    timed: Array<{ id: string; start: number; end: number }>,
+    eligible: Map<string, number>,
+    parallel: number
+  ): Segment[] {
+    const occDelta = new Map<number, number>();
+    const waitDelta = new Map<number, number>();
+    const monopolyDelta = new Map<number, number>();
```
AgentEnder: monopoly => nonParallel or something more descriptive

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+   * tasks eligible-but-not-running (`waiting`), and how many `parallelism: false`
+   * tasks are running (`monopoly`). A `parallelism: false` task occupies the whole
+   * pool. Single source of truth for slot contention; integrates over time rather
+   * than sampling one instant, so it's robust to when tasks hand off.
+   */
+  private buildSegments(
+    timed: Array<{ id: string; start: number; end: number }>,
+    eligible: Map<string, number>,
+    parallel: number
+  ): Segment[] {
```
AgentEnder: This function could be a bit more readable... and likely deserves some unit tests of its own somewhere... Can we extract it and test it nicer?

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+    firstTaskStart: number,
+    hashWindows: Array<[number, number]>
+  ): number {
+    const merged: Array<[number, number]> = [];
+    for (const [s, e] of hashWindows
+      .map(([s, e]): [number, number] => [s, Math.min(e, firstTaskStart)])
+      .filter(([s, e]) => e > s)
+      .sort((a, b) => a[0] - b[0])) {
+      const last = merged[merged.length - 1];
+      if (last && s <= last[1]) {
```
AgentEnder: Should this be last[0]? Again I kinda feel there should be a more readable way to express this function.... The chained map, filter, sort, inside the iterand of the for loop isn't clear, what is merged representing, what are we pusing, etc... Its just not very readable.

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+    const summary = lifeCycle.getSummary();
+    if (!summary) {
+      return;
+    }
+    // Post-TUI the terminal can still be in raw mode (no \n → \r\n translation),
+    // which staircases plain "\n". Use \r\n on a TTY; plain \n when piped.
+    // `formatReport` already ends with a trailing newline, so don't add another
+    // (that left a stray blank line that broke exact-match e2e snapshots).
+    const eol = process.stdout.isTTY ? '\r\n' : '\n';
+    process.stdout.write(formatReport(summary).split('\n').join(eol));
```
AgentEnder: This seems wrong?  `console.log`?

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1237 @@
...
+      console.error(e);
+    }
+  }
+}
+
+/**
+ * Absolute-epoch [start, end] hashing windows, read from nx's existing `hash*` perf
+ * measures. Returns [] if the performance API is unavailable.
+ */
+function collectHashWindows(): Array<[number, number]> {
```
AgentEnder: Do these differ based on daemon enablement?

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1208 @@
...
+    return out;
+  }
+  for (const { visible, href } of REPORT_LINKS) {
+    out = out.split(href).join(terminalLink(visible, href));
+  }
+  for (const { phrase, href } of PHRASE_LINKS) {
+    out = out.split(phrase).join(terminalLink(phrase, href));
+  }
+  return out;
+}
```
AgentEnder: Why are we building the report with one set of links baked in and then replacing with the OSC8 sequences instead of invoking inline?

output.push(link(href, visible)) instead of buildOutputs.split(...).join(...)

A utility function `terminalLink(text: string, href: string)` may make more sense?

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,1208 @@
...
+  if (ms < 1000) {
+    return `${Math.round(ms)}ms`;
+  }
+  const seconds = Math.round(ms / 100) / 10;
+  if (seconds >= 60) {
+    const totalSeconds = Math.round(ms / 1000);
+    return `${Math.floor(totalSeconds / 60)}m ${totalSeconds % 60}s`;
+  }
+  return `${seconds.toFixed(1)}s`;
+}
```
AgentEnder: We almost surely have existing utils for this, if by chance we don't, can we use the rust util?

packages/nx/src/tasks-runner/run-command.ts
```diff
@@ -221,6 +228,20 @@ async function getTerminalOutputLifeCycle(
...
+      if (typeof boundEndCommand === 'function') {
+        // Only multi-task runs feed the popup; a single-task run exits with no popup, leaving its report for the terminal flush below.
+        typedAppLifeCycle.endCommand = () =>
+          boundEndCommand(
+            tasks.length > 1
+              ? (getPerformanceSummaryPayload() ?? undefined)
+              : undefined
+          );
+      }
+
```
AgentEnder: This is weird? I don't think this seems correct.

AgentEnder (changes requested):

packages/nx/src/tasks-runner/life-cycles/performance-life-cycle.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (!entry) {
+      entry = { continuous: this.taskGraph.tasks[taskId]?.continuous ?? false };
+      this.timings.set(taskId, entry);
+    }
+    return entry;
+  }
+
+  /** Analyze the collected timings into a structured summary, or `null` when no discrete task timings were recorded. */
+  getSummary(): PerformanceSummary | null {
+    return new PerformanceAnalysis(
```
AgentEnder, line 93: I question a bit i PerformanceAnalysis actually needs to be a class if we are creating it, and only using the result of a member that is immediately invoked... Feels more function-y to me?

AgentEnder (approved):

PR #36124, opened 2026-06-26
fix(js): add migration to keep node types on ts-node configs under typescript 6

jaysoo:
## Current Behavior

The `add-ignore-deprecations-for-ts6` migration keeps `ts-node.compilerOptions` on `moduleResolution: node10`, but under TypeScript 6 the `node`/`node10`/`classic` resolutions no longer auto-load `@types/node` the way 5.x did. ts-node compiles `jest.config.ts` (which uses `module.exports`) with that block, so the `module` global disappears:

```
jest.config.ts(2,1): error TS2591: Cannot find name 'module'.
```

(Surfaced by nx-console's `nxls-e2e` after migrating to 23.1.)

## Expected Behavior

Adds `23-1-0-add-node-types-to-ts-node-config` - a deterministic generator that adds `"types": ["node"]` to `ts-node.compilerOptions` blocks using a node-style resolution and lacking a `types` field. Only reads `tsconfig*.json` (no source walk), gated `typescript >=6.0.0`, and scoped to the `ts-node` block (pinning `types` on the main `compilerOptions` would wrongly narrow the whole project).

## Related Issue(s)

Fixes NXC-4596

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/nx-23.1.0-beta.0-Migration-24e91166)

### DIFF
--- a/packages/js/migrations.json
+++ b/packages/js/migrations.json
@@ -37,6 +37,14 @@
       },
       "factory": "./dist/src/migrations/update-23-1-0/add-ignore-deprecations-for-ts6",
       "documentation": "./dist/src/migrations/update-23-1-0/add-ignore-deprecations-for-ts6.md"
+    },
+    "23-1-0-add-node-types-to-ts-node-config": {
+      "version": "23.1.0-beta.5",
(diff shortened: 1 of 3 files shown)


jaysoo:
Not a full fix, the `types: ['*']` will be the actual fix.. we'll do another PR

Issue #34655, opened 2026-03-01
`@nx/js:prune-lockfile` does not include transitive workspace dependencies in pruned lockfile (pnpm)


lucasvieirasilva:
### Current Behavior

When an application depends on workspace libraries that themselves depend on other workspace libraries (transitive workspace dependencies), `@nx/js:prune-lockfile` fails to:

1. Generate `importers` blocks for transitive workspace dependencies in the pruned pnpm lockfile.
2. Include the npm dependencies of those transitive workspace modules in the packages section.
3. Rewrite the link: paths to match the flat workspace_modules/ layout used by @nx/js:copy-workspace-modules.

For example, given:
- `api` depends on `@myorg/module-a` (direct, in `package.json`)
- `@myorg/module-a` depends on `@myorg/module-b` (transitive workspace dep)
- `@myorg/module-b` depends on `some-npm-package`

The pruned lockfile will:
- Include a `workspace_modules/@myorg/module-a`: importer block (correct)
- Not include a `workspace_modules/@myorg/module-b`: importer block (bug)
- Not include `some-npm-package` in the packages section (bug)

Reference `@myorg/module-b` from `module-a`'s importer block with the original monorepo-relative `link:` path (e.g., `link:../infra/module-b`) instead of the correct flat path (link:../module-b) (bug)

Meanwhile, `@nx/js:copy-workspace-modules` correctly copies all transitive workspace modules into `dist/workspace_modules/`, so there is a mismatch between what's on disk and what's in the lockfile.

### Expected Behavior

`@nx/js:prune-lockfile` should recursively discover all transitive workspace dependencies and:

1. Generate `importers` blocks for every workspace module present in `dist/workspace_modules/`.
2. Include all npm dependencies of transitive workspace modules in the `packages` section.
3. Rewrite all `link:` paths to reflect the flat `workspace_modules/` layout.

### Steps to Reproduce

1. Create a workspace with three projects:

```
apps/api/package.json          -> depends on @myorg/lib-a (workspace:*)
libs/lib-a/package.json        -> depends on @myorg/lib-b (workspace:*)
libs/lib-b/package.json        -> depends on some-npm-package (e.g., lodash)
```

2. Run `nx run api:prune`.

### Nx Report

```shell
Node           : 24.2.0
OS             : linux-x64
Native Target  : x86_64-linux
pnpm           : 10.14.0
daemon         : Available

nx                     : 22.5.3
@nx/js                 : 22.5.3
[...]

PR #36139, opened 2026-06-29
fix(js): make pruned pnpm output installable and run its workspace modules

### LINKED ISSUE #18402: GeneratePackageJson: package.json with overrides or patches generate incorrect lockfile

### LINKED ISSUE #35425: fix(js): workspace devDependencies ignored by prune-lockfile executor

### LINKED ISSUE #36055: nx/webpack:webpack + generatePackageJson produces invalid pnpm output under pnpm 11 (missing workspace config → ERR_PNPM_LOCKFILE_CONFIG_MISMATCH)
### Current Behavior

When using pnpm 11+ together with `@nx/webpack:webpack` and the option: `generatePackageJson": true`

the build output contains:

-  `package.json`
- `pnpm-lock.yaml`
- …but no `pnpm-workspace.yaml`

 pnpm 11 now stores override configuration in the `pnpm-workspace.yaml` file.

When I run `pnpm install --ingore-workspace --prod --frozen-lockfile` inside the generated `dist/` folder, it fails with:

`ERR_PNPM_LOCKFILE_CONFIG_MISMATCH
The current "overrides" configuration doesn't match the value found in the lockfile
`

### Expected Behavior

The build output should be self‑contained and installable with pnpm 11, meaning:

`pnpm install --ingore-workspace  --prod --frozen-lockfile` should succeed inside` dist/`

override metadata in the lockfile should match the configuration available in the output

NX should therfore:

generate a minimal `pnpm-workspace.yaml` in the output folder,

### GitHub Repo

_No response_

### Steps to Reproduce

see expected behavior

### Nx Report

```shell
-
```

### Failure Logs

```shell
-
```

### Package Manager Version

pnpm@11.7

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

### LINKED ISSUE #36066: `@nx/js:prune-lockfile` is broken with pnpm (no transitive production dependencies)
### Current Behavior

When pruning a pnpm lockfile, the result does not contain transitive dependencies, causing the application to fail at runtime.

### Expected Behavior

The pruned lockfile contains all required production dependencies (internal and external, direct and indirect), and none of the development dependencies (neither in the importers or the packages section).

### GitHub Repo

_No response_

### Steps to Reproduce

```bash
pnpm dlx create-nx-workspace@23 --name=nx-repro-23-pnpm-1781945042541 --preset=ts --formatter=none --ci=skip --skipGit --packageManager=pnpm --nxCloud=skip --analytics=false
cd nx-repro-23-pnpm-1781945042541
pnpm exec nx add @nx/node
pnpm exec nx generate @nx/node:application --directory=apps/backend --useProjectJson --bundler=esbuild --linter=none --unitTestRunner=none --e2eTestRunner=none --framework=express
pnpm exec nx generate @nx/node:library --directory=packages/backend-lib --useProjectJson --compiler=tsc --linter=none --unitTestRunner=none
pnpm add --filter "backend" --workspace "@nx-repro-23-pnpm-1781945042541/backend-lib"
sed -i '1a\import { backendLib } from "@nx-repro-23-pnpm-1781945042541/backend-lib";' 'apps/backend/src/main.ts'
sed -i '2a\console.log(backendLib());' 'apps/backend/src/main.ts'
pnpm exec nx sync
pnpm exec nx run backend:prune
mkdir -p ../nx-repro-23-pnpm-1781945042541-out/backend/dist
cp -r apps/backend/dist/. ../nx-repro-23-pnpm-1781945042541-out/backend/dist
cd ../nx-repro-23-pnpm-1781945042541-out/backend/dist
node main.js
```

### Nx Report

```shell

Node           : 24.14.0
OS             : win32-x64
pnpm           : 10.32.1
daemon         : Disabled

nx (global)  : 22.6.5
nx           : 23.0.0
@nx/js       : 23.0.0
@nx/esbuild  : 23.0.0
@nx/node     : 23.0.0
typescript   : 5.9.3
Registered Plugins:
@nx/js/typescript
Cache Usage: 99.51 KB / 47.63 GB
```

### Failure Logs

```shell
node:internal/modules/cjs/loader:1459
  throw err;
  ^

Error: Cannot find module 'tslib'
Require stack:
- C:\Users\<redacted>\Desktop\nx-repro-2\nx-repro-23-pnpm-1781945042541-out\backend\dist\workspace_modules\@nx-repro-23-pnpm-1781945042541\backend-lib\dist\index.js
- C:\Users\<redacted>\Desktop\nx-repro-2\nx-repro-23-pnpm-1781945042541-out\backend\dist\apps\backend\src\main.js
- C:\Users\<redacted>\Desktop\nx-repro-2\nx-repro-23-pnpm-1781945042541-out\backend\dist\main.js
    at Module.<anonymous> (node:internal/modules/cjs/loader:1456:15)
    at Module._resolveFilename (C:\Users\<redacted>\Desktop\nx-repro-2\nx-repro-23-pnpm-1781945042541-out\backend\dist\main.js:33:36)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1066:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1071:22)
    at Module._load (node:internal/modules/cjs/loader:1242:25)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1556:12)
    at require (node:internal/modules/helpers:152:16)
    at Object.<anonymous> (C:\Users\<redacted>\Desktop\nx-repro-2\nx-repro-23-pnpm-1781945042541-out\backend\dist\workspace_modules\@nx-repro-23-pnpm-1781945042541\backend-lib\dist\index.js:3:17)
    at Module._compile (node:internal/modules/cjs/loader:1812:14) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'C:\\Users\\<redacted>\\Desktop\\nx-repro-2\\nx-repro-23-pnpm-1781945042541-out\\backend\\dist\\workspace_modules\\@nx-repro-23-pnpm-1781945042541\\backend-lib\\dist\\index.js',
    'C:\\Users\\<redacted>\\Desktop\\nx-repro-2\\nx-repro-23-pnpm-1781945042541-out\\backend\\dist\\apps\\backend\\src\\main.js',
    'C:\\Users\\<redacted>\\Desktop\\nx-repro-2\\nx-repro-23-pnpm-1781945042541-out\\backend\\dist\\main.js'
  ]
}

Node.js v24.14.0
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

Please consider removing pnpm from your list of supported package managers until this issue is fixed. Support is currently completely broken by default, at least for backend development with the default options.

Alternatively, please provide the users with a workaround to work with backend projects in pnpm, until this bug is fixed.

I know of the possibility to bundle the backend with esbuild (`bundle` and `thirdParty`) but this is not an option for us. For example, instrumentation libraries like OpenTelemetry require a CJS/ESM module loading process.

### LINKED ISSUE #36140: @nx/js:prune-lockfile output not installable by pnpm for transitive workspace deps (follow-up to #34655)
### Current Behavior

#34655 was closed, and Nx 23 **did** improve `@nx/js:prune-lockfile` — it now generates `importers` blocks for transitive workspace modules and rewrites their `link:` paths to the flat `workspace_modules/` layout. However, the pruned + copied output **still cannot be installed by pnpm**: a transitive-only workspace dependency is never materialized into `node_modules`, so the deployed app crashes at runtime. The only workaround remains the one from #34655 — declaring every transitive workspace dep as a direct dependency of the app — which is exactly what that issue aimed to eliminate.

Given (all under `packages/*`):
- `@myorg/api` depends on `@myorg/lib-a` (direct, `workspace:*`)
- `@myorg/lib-a` depends on `@myorg/lib-b` (transitive workspace dep)
- `@myorg/lib-b` depends on `lodash`

After `nx run api:prune`, the pruned `packages/api/dist/pnpm-lock.yaml` correctly lists the transitive importer and a flat `link:` path:

```yaml
importers:
  .:
    dependencies:
      '@myorg/lib-a':
        specifier: file:./workspace_modules/@myorg/lib-a
        version: link:./workspace_modules/@myorg/lib-a
  workspace_modules/@myorg/lib-a:
    dependencies:
      '@myorg/lib-b':
        specifier: file:../lib-b
        version: link:../lib-b      # importer + flat path correct (fixed since #34655)
  workspace_modules/@myorg/lib-b: # importer present (fixed since #34655)
    dependencies:
      lodash:
        specifier: ^4.17.21
        version: 4.17.21
```

But deploying that output (copying `packages/api/dist` to an isolated dir, as a Docker image does) and installing it:

```bash
CI=true pnpm install --prod --frozen-lockfile --ignore-scripts --node-linker=hoisted
```

only materializes the **root's direct** workspace dep into `node_modules`. `@myorg/lib-b` (transitive-only) is installed **nowhere** — no top-level `node_modules/@myorg/lib-b`, none nested under `lib-a`, no `.pnpm` entry. At runtime:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@myorg/lib-b' imported from .../workspace_modules/@myorg/lib-a/dist/lib/lib-a.js
```

Root cause: `prune-lockfile` expresses the transitive edge as a directory **`link:`** (`@myorg/lib-a → link:../lib-b`). pnpm's contract for a `link:` dependency is "symlink the target as-is; it owns its `node_modules`," so pnpm does **not** install the dependencies of a `link:` target — and `copy-workspace-modules` ships the module without a `node_modules`. The lockfile listing the importer is necessary but not sufficient; pnpm never acts on those importers for a single-project install where the edges are `link:`/`file:`-dir rather than the `workspace:` protocol.

### Expected Behavior

The pruned + copied output should be installable such that **all** transitive workspace dependencies resolve at runtime, without manually adding them as direct deps of the app. Concretely, `prune-lockfile`/`copy-workspace-modules` should emit edges pnpm will recurse — e.g. keep `workspace_modules/*` as real `workspace:` projects, or inject the workspace packages rather than emitting bare `link:` symlinks.

### GitHub Repo

https://github.com/lucasvieirasilva/nx-issue-34655-followup-repro

### Steps to Reproduce

```bash
git clone https://github.com/lucasvieirasilva/nx-issue-34655-followup-repro
cd nx-issue-34655-followup-repro
pnpm install
./scripts/reproduce.sh   # exits non-zero with ERR_MODULE_NOT_FOUND for @myorg/lib-b
```

`scripts/reproduce.sh` does, in full:
1. `nx run api:prune` — builds, then writes `packages/api/dist/{package.json,pnpm-lock.yaml,workspace_modules/}`.
2. Copies `packages/api/dist` to a temp dir **outside the workspace** (so pnpm resolves it as a standalone project, like `COPY dist .` in a Dockerfile — installing in place makes pnpm walk up to the root workspace instead).
3. `CI=true pnpm install --prod --frozen-lockfile --ignore-scripts --node-linker=hoisted`
4. `node index.js` → crashes; `node_modules/@myorg/lib-b` is absent.

Confirmed **not** fixable from the install side (pnpm 11.5.0): dropping `--prod`, adding `packages: ['workspace_modules/*/*']` to the deploy `pnpm-workspace.yaml` (frozen **and** non-frozen), and `inject-workspace-packages=true` all still fail to install the transitive package — the edges are `link:`/`file:`-dir, not the `workspace:` protocol pnpm needs to manage transitive resolution.

### Nx Report

```shell

Node           : 24.15.0
OS             : linux-x64
Native Target  : x86_64-linux
pnpm           : 11.5.0
daemon         : Available

nx             : 23.0.1
@nx/js         : 23.0.1
@nx/workspace  : 23.0.1
@nx/devkit     : 23.0.1
typescript     : 6.0.3
Registered Plugins:
@nx/js/typescript
Cache Usage: 0.00 B / 100.69 GB
```

### Failure Logs

```shell
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@myorg/lib-b' imported from .../workspace_modules/@myorg/lib-a/dist/lib/lib-a.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
    at packageResolve (node:internal/modules/esm/resolve:764:81)
    at moduleResolve (node:internal/modules/esm/resolve:855:18)
    ...
```

### Package Manager Version

pnpm 11.5.0 (also reproduced on 10.x)

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

Follow-up to #34655 (closed). #34655 covered the missing `importers` blocks / npm deps / `link:` path rewrites — addressed in Nx 23. This issue is the remaining end-to-end gap: even with a correct pruned lockfile, the deployed artifact is not installable by pnpm because transitive workspace edges are emitted as `link:`.

**Workaround (unchanged from #34655):** declare every transitive workspace dependency as a direct dependency in the app's `package.json`.

leosvelperez:
## Current Behavior

Several failures break pnpm deployment of pruned build output.

**The install fails.** A production build that emits a `package.json` and a pruned pnpm lockfile (webpack, rspack, vite, and the next/remix/js executors when they generate a lockfile) cannot be installed standalone. `pnpm install --frozen-lockfile` aborts with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`, because the pruned lockfile keeps `overrides`, `ignoredOptionalDependencies`, `settings`, `catalogs`, and the package-extension and pnpmfile checksums, while the output ships no `pnpm-workspace.yaml` and pnpm 11 reads the install settings only from that file.

**Workspace modules crash at runtime.** A node app pruned with `@nx/js:prune-lockfile` and `copy-workspace-modules` throws `Cannot find module` for a copied module's own dependency: pnpm installs a copied module's transitive production dependencies only when the module is declared as a package, and the pruned output declares none. A workspace module the app depends on through `peerDependencies` breaks the install outright, since pnpm rejects a `file:` spec there, and one declared under `devDependencies` is left with a `workspace:*` spec no standalone install can resolve.

**Patches are dropped.** A workspace that uses `pnpm patch` produces an output with no `.patch` files and no `patchedDependencies` config, so the install fails on a config mismatch or the dependency ships unpatched.

**Local-path dependencies do not resolve.** `file:` tarballs, `file:` directories and `link:` directories are written unchanged, so a non-frozen install re-resolves them against workspace paths the artifact does not carry, a copied module's `link:` dependency resolves against the wrong base and leaves a dangling symlink, and a `file:`/`link:` peer a workspace module declares is recorded nowhere in the pruned lockfile, so the install exits 0 and require fails.

**A deploy can rewrite a project's source manifest.** With `generatePackageJson: false`, `generateLockfile: true` and an output path equal to the project root, which a remix build supports, the pnpm build-script approvals and `patchedDependencies` land in the project's own `package.json` and reformat it.

**A bun output is not installable.** It keeps `workspace:*` specifiers that `bun install` cannot resolve outside a workspace, and `@nx/js:prune-lockfile` logs a lockfile path it never writes.

## Expected Behavior

The pruned output installs cleanly under both `pnpm install --frozen-lockfile` and a non-frozen install on every supported pnpm version. Copied workspace modules get their production dependencies installed, including modules the app declares under `devDependencies` or `peerDependencies`, so the app runs. `file:` and `link:` dependencies are relocated to resolve inside the deploy artifact and their closures validated at build time, and `pnpm patch` files and their config ship so patched dependencies stay patched. Nothing outside the output directory is written, and a bun output carries specifiers `bun install` can resolve. Targets that emit the output declare the new artifacts in `outputs` and the pnpm settings sources in `inputs`, so a cache replay restores the full output, and revoking a build-script approval, changing `supportedArchitectures` or the patch set, or switching the pnpm major re-runs the task instead of replaying a stale output.

## Implementation Details

- `generatePrunedDeployOutput` is the entry point every generate-package-json flow calls. It writes the pruned lockfile and, for pnpm, the install-time artifacts that lockfile needs, into either an output directory or a bundler's asset pipeline. It mutates the manifest into the shape the output must ship and keeps it in memory, so a caller writes or emits the `package.json` after the call and nothing reads a manifest back off disk. Bun has no lockfile generation, so it warns and ships nothing, leaving the manifest as authored.
- Pruning drops the resolution-time pnpm config the pruned snapshots already bake in, so the manifest and the lockfile agree on every pnpm version: `overrides`, `ignoredOptionalDependencies`, `settings`, `catalogs` and the package-extension and pnpmfile checksums from the lockfile, and `pnpm.overrides`, `pnpm.ignoredOptionalDependencies` and `pnpm.packageExtensions` from the manifest. A `pnpm.patchedDependencies` the manifest inherited from the project goes too, since it names the workspace's patch paths rather than the output's and would fail the install on a patch file the output never ships.
- `@nx/js:prune-lockfile` emits every workspace module the app depends on, directly or transitively and through any of `dependencies`, `devDependencies`, `optionalDependencies` or `peerDependencies`, as a pnpm `file:` directory package (`<name>@file:workspace_modules/<name>`) carrying the module's own resolved production closure, and points the app's dependency at `file:./workspace_modules/<name>`. A module reached through `peerDependencies` moves into `dependencies`, since pnpm rejects a `file:` spec there. `copy-workspace-modules` copies the modules into `workspace_modules/`, so a standalone `pnpm install` installs each module's production dependencies. Only dependencies that are workspace packages in the project graph are rewritten, so a vendored tarball keeps its own spec.
- On pnpm 11 and above, `allowBuilds`, `supportedArchitectures` and `patchedDependencies` are read only from `pnpm-workspace.yaml`. The output carries a settings-only `pnpm-workspace.yaml` holding the ones the workspace declares, scoped to the packages the pruned output installs, plus an empty `packages: []` list, since pnpm 9 rejects the file without one. On pnpm 10 and below the same settings stay in the emitted `package.json` and the file ships carrying only that empty list. It ships on every pnpm output rather than only when there is something to carry: a cache replay restores only the files the replayed entry holds and a bundler asset pipeline can only add assets, so an artifact some builds skip would leave an earlier build's copy in place for pnpm to read. A `packages: []`-only file installs identically to no file at all on pnpm 9, 10 and 11, and leaves the pnpm 10 `package.json` build approvals in force.
- The settings those artifacts carry are hashed so a cached task re-runs when they change: the root `pnpm-workspace.yaml`, the root `package.json` (the bundler build targets narrow it through a `json` input to `packageManager` and the `pnpm` approvals, architectures and patch fields, so ordinary dependency bumps in the root manifest do not invalidate every build), and a `runtime` input probing the pnpm major, which decides whether the settings land in the emitted `pnpm-workspace.yaml` or the emitted `package.json` when the root manifest has no `packageManager` field. The probe prints only the major, so pnpm patch and minor releases do not move the hash, and a sentinel when no pnpm binary is available. The `@nx/node` application generator writes the inputs and the widened `outputs` on new prune targets, and the `@nx/webpack` and `@nx/rspack` plugins declare them on the build targets they infer when the workspace uses pnpm. The contents of vendored local-path dependencies are not hashed, since their set is derived from the lockfile at build time and no statically written input list stays correct as dependencies change; hashing them, including `link:` sources without a project-graph node, is follow-up work at the graph level.
- `pnpm patch` files ship under the output's `patches/` directory, keeping the path recorded in the pnpm config as a subpath, so a recorded `patches/foo.patch` lands at `patches/patches/foo.patch`. The pruned lockfile keeps only the `patchedDependencies` entries whose package survives the prune, matching a name-only, exact, or semver-range patch key against the resolved package versions. Two declarations that would collide on one output path abort the build, before anything is written.
- Local-path dependencies are vendored under `local_path_modules/` and their specs rewritten to resolve there. `link:` references are emitted relative to the deploy root and `file:` specs relative to the owning package, since pnpm resolves each protocol from a different base. A `file:`/`link:` peer a workspace module declares is backfilled into the pruned lockfile as the edge and package entry pnpm writes when it auto-installs it, and `validatePrunedLocalPathClosure` fails the build when a shipped local-path package requires a dependency the app does not install. Two targets cannot ship: one resolving outside the workspace root, which is left in place, and one sitting where a copied workspace module of the same name lands, which the output cannot tell apart from that module. Both warn. Closure validation and local-path shipping are skipped when the prune falls back to the root lockfile, whose importer references the whole workspace.
- `createLockFile` keeps its signature and reconciles the manifest it is given for pnpm, so a caller writes the `package.json` after calling it rather than before. It is deprecated in favor of `generatePrunedDeployOutput` and slated for removal in Nx 25.

## Public API

`@nx/js` gains one export, `generatePrunedDeployOutput(packageJson, graph, projectRoot, options)`, where `options` carries the `packageManager` plus either an `outputDirectory` to write into or an `emit` sink for a bundler's asset pipeline. `createLockFile` is marked `@deprecated` in its favor. Nothing is renamed or removed, and `@nx/devkit`'s public surface is unchanged.

> [!NOTE]
> `pnpm install --ignore-workspace` skips a `pnpm-workspace.yaml` sitting beside the output, so a workspace using `pnpm patch` gets an output that installs cleanly without the flag and fails with it. The emitted file declares an empty `packages` list, which already stops pnpm treating a parent directory as a workspace, so the flag is not needed to install the output in place. The deploy guide covers this.

## Migration coverage

- New `@nx/js:prune-lockfile` artifacts (`pnpm-workspace.yaml`, `patches/`, `local_path_modules/`) and the settings that determine them, on existing targets: migration, `update-23-2-0-add-pnpm-prune-lockfile-cache-config`. A target listing only `package.json` and the lockfile under `outputs` would have a cache replay drop the new artifacts and deploy without the install settings, patch files, or vendored local-path dependencies. Those targets also declare no `inputs`, and the build approvals and `supportedArchitectures` the artifacts carry reach no lockfile, so revoking an approval would leave the task hash unchanged and replay an output that still grants it. The migration appends the three paths next to the target's existing pnpm lockfile entry, reusing that entry's own path spelling, and adds the root `pnpm-workspace.yaml`, the root `package.json` and the pnpm-major `runtime` probe to the `inputs` array the target actually resolves: its own when it declares one, otherwise the `targetDefaults` entry it inherits from, with nx's default spelled out when it inherits none. `targetDefaults` are resolved the way the runtime does (executor key, exact-name key, then glob keys longest first, then each entry's own executor compatibility), so a defaults-only workspace is covered too; when no compatible filter-less entry under the resolving key declares `inputs`, a spelled-out default entry is prepended only where it provably cannot change which key a target resolves to or hand the executor to a same-name target of another executor, and any other key shape keeps its input gap, documented in the migration.
- Build targets that emit the deploy output, on existing executor configurations: migration, `update-23-2-0-add-pnpm-deploy-output-cache-inputs`. In workspaces with a `pnpm-lock.yaml` it adds the same settings sources, with the `package.json` fields narrowed through a `json` input, to webpack, rspack, vite and esbuild builds with `generatePackageJson` enabled and to next, remix, tsc and swc builds with `generateLockfile` enabled, using the same layer resolution as the prune migration. Targets the `@nx/webpack` and `@nx/rspack` plugins infer need no migration: the plugins declare the inputs and inference recomputes them.
- Prune targets a generator creates: nothing to migrate, `getPruneTargets` declares the three paths, the two root inputs and the pnpm-major probe for pnpm workspaces.
- No option, executor, or configuration key is renamed or removed.

## Related Issue(s)

Fixes #36055
Fixes #36066
Fixes #36140
Fixes #35425
Fixes #18402

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-36055-9a7143dd)


rendmath:
See https://github.com/nrwl/nx/issues/34655#issuecomment-4839715919 for a different approach that is currently working for us in pnpm 10.

leosvelperez:
@rendmath, thanks for your investigation and suggestion. It makes sense to me, so I pivoted this PR to use that approach.

rendmath:
Removing the development dependencies is the right call for both the `prune-lockfile` and the `copy-workspace-modules` executors. Not only are the development dependencies useless for restoring a production build, but more importantly they are a security liability. They expose internal details of the CI/CD toolchain to potential attackers.

We should move towards a commonly agreed whitelist of properties to preserve in package.json files and discard the rest.

rendmath:
I am also wondering if the new `allowBuilds` feature of pnpm 11 is going to cause issues during package restoration. I will make tests and keep you informed of my findings.

rendmath:
Yes the `allowBuilds` feature is causing issues. I did not find an easy process with the CLI on a build to add the allowed packages without getting a non-zero return code first (`ERR_PNPM_IGNORED_BUILDS`).

So the best developer experience is probably to carry over the `pnpm-workspace.yaml` file with only these fields, filtered by the packages present in the release.

I will try this solution with our internal version of the executor and let you know.

rendmath:
Another point of attention that we have discovered and that might be worthy of inclusion in your test suite. Some packages use aliased dependencies, and it has an impact on the resolution of the snapshot in the original lock file.

Example with `@isaacs/cliui`, which is an unfortunate transitive dependency of the `glob` package, so still fairly common.

```
'@isaacs/cliui@8.0.2':
  dependencies:
    string-width: 5.1.2
    string-width-cjs: string-width@4.2.3
    strip-ansi: 7.2.0
    strip-ansi-cjs: strip-ansi@6.0.1
    wrap-ansi: 8.1.0
    wrap-ansi-cjs: wrap-ansi@7.0.0
```

In that specific case we happened to be lucky because another sibling package has a non-aliased dependency on the exact same version of the same package, so the snapshot was present and the lockfile was valid. But that will not always be the case.

So in those cases, you need to use the left part of the reference to locate the snapshot to include, and not the dependency key.

rendmath:
Also you cannot simply test the presence of a `@` in the reference otherwise you will pick some peer dependency resolution contexts references like those:

```
express@5.2.1(supports-color@10.2.2):
  dependencies:
    accepts: 2.0.0
    body-parser: 2.2.2(supports-color@10.2.2)
```

So I assume that the `@` must be before any `(`. I am not sure what an aliased reference with peer dependency resolution context would look like. I assumed something like this `body-parser-cjs: body-parser@2.2.2(supports-color@10.2.2)`.

So this is what we are currently using:

```
function determineSnaphotKey(
  dependencyName: string,
  dependencyVersion: string
): string {
  const atIndex = dependencyVersion.indexOf("@");
  const parenIndex = dependencyVersion.indexOf("(");
  const isAlias = atIndex !== -1 && (parenIndex === -1 || atIndex < parenIndex);
  return isAlias ? dependencyVersion : `${dependencyName}@${dependencyVersion}`;
}
```

Armster15:
ahh tysm for working on this! we ended up having to deploy a handrolled script to add support for pruning w/ patched dependencies / overrides / `catalog:` support so im really excited to see this land 🥹

rendmath:
Yes it looks like amazing work done by @leosvelperez on this PR. We also rolled custom scripts to circumvent the issues in our team and we might be able to switch back to standard executors now.

FrozenPandaz (approved):
LGTM 🎉 Thanks everyone for chiming in and helping to solve this issue!

PR #36158, opened 2026-06-30
fix(misc): fix dead links for vite deprecation messages

colinscz:
Affected version: Version 23.0.1

## Current Behavior
Currently when running vitest tests the deprecation links shown in the warning messages end up in a 404 not found

For example:
`The nxCopyAssetsPlugin plugin from @nx/vite/plugins/nx-copy-assets.plugin is deprecated and will be removed in Nx v24. Use Vite's native publicDir option or the vite-plugin-static-copy package instead. See https://nx.dev/docs/technologies/build-tools/vite/configure-vite for details.`

## Expected Behavior
Those links should be pointing to the correct page in the documentation. Small fix, The path element "guide" was missing.


colinscz:
Fixed in the meantime via #36070

PR #36188, opened 2026-07-01
fix(webpack): prevent TS6059 when a tsc build bundles a workspace lib

leosvelperez:
## Current Behavior

Building an app with the `@nx/webpack` `tsc` compiler fails when the app imports a workspace library whose sources are bundled from source (the default `buildLibsFromSource`). The build errors with:

```
TS6059: File 'libs/lib-1/src/index.ts' is not under 'rootDir' 'apps/app-1/src'. 'rootDir' is expected to contain all source files.
```

ts-loader 9.5.7 stopped forcing `rootDir` to `undefined` (which had masked this) to avoid a TS5011 error on TypeScript 6. Because the floating `^9.3.1` range now resolves to 9.5.7+, ts-loader forwards the app tsconfig `rootDir` into `transpileModule`, and lib sources resolved outside the app `rootDir` fail the check.

## Expected Behavior

A `tsc` webpack build that imports a workspace library from source compiles without `TS6059`.

## Related Issue(s)

Fixes #35017

## Implementation Details

`rootDir` only affects output layout, which webpack owns in a bundled build, so it is safe to widen it for the loader. `createLoaderFromCompiler` now sets the ts-loader `compilerOptions.rootDir` to the workspace root, keeping it defined (avoiding TS5011) while covering the lib sources. ts-loader is raised to `^9.5.7` so the lockfile tracks the version users already resolve. Includes a unit test for the loader option and an e2e case that builds a `tsc` app importing a workspace lib from source.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-35017-c68ef038)

### DIFF
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/compiler-loaders.spec.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/compiler-loaders.spec.ts
@@ -5,8 +5,6 @@ describe('createLoaderFromCompiler tsc', () => {
   function tscOptions(): NormalizedNxAppWebpackPluginOptions {
     return {
       root: '/test',
-      projectRoot: 'apps/test',
-      sourceRoot: 'apps/test/src',
       tsConfig: 'apps/test/tsconfig.app.json',
       compiler: 'tsc',
       transformers: [],
--- a/packages/webpack/src/plugins/nx-webpack-plugin/lib/compiler-loaders.spec.ts
+++ b/packages/webpack/src/plugins/nx-webpack-plugin/lib/compiler-loaders.spec.ts
@@ -1,0 +1,24 @@
+import { createLoaderFromCompiler } from './compiler-loaders';
+import { NormalizedNxAppWebpackPluginOptions } from '../nx-app-webpack-plugin-options';
+
+describe('createLoaderFromCompiler tsc', () => {
+  function tscOptions(): NormalizedNxAppWebpackPluginOptions {
+    return {
+      root: '/test',
+      projectRoot: 'apps/test',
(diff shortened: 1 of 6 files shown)


packages/webpack/src/plugins/nx-webpack-plugin/lib/compiler-loaders.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { createLoaderFromCompiler } from './compiler-loaders';
+import { NormalizedNxAppWebpackPluginOptions } from '../nx-app-webpack-plugin-options';
+
+describe('createLoaderFromCompiler tsc', () => {
+  function tscOptions(): NormalizedNxAppWebpackPluginOptions {
+    return {
+      root: '/test',
+      projectRoot: 'apps/test',
```
alexandremaz: question : why would we need this 'projectRoot' property in this specific test (and same for 'sourceRoot' property) ? it looks like you only want to use options used in case 'tsc' in createLoaderFromCompiler function

so only : 
options.compiler
options.transformers
options.tsConfig
options.root

am I wrong ?

alexandremaz:
thx for the fix 🙏

packages/webpack/src/plugins/nx-webpack-plugin/lib/compiler-loaders.spec.ts
leosvelperez: You're not wrong. Those are leftovers from an initial change and just make the test data look a bit more realistic, but they are not needed. I cleaned it up to make what's relevant clear.

FrozenPandaz (approved):

PR #36189, opened 2026-07-01
fix(core): accept legacy plugin array in retrieveProjectConfigurations

llwt:
## Current Behavior

Creating a workspace with `create-nx-workspace@22.x --preset=nest` (and any preset that ends up calling `@nx/webpack`'s init generator) crashes during generation with:

```
NX   Cannot read properties of undefined (reading 'filter')
```

`@nx/devkit@22.x` declares `peerDependencies.nx: ">= 21 <= 23"`, so during the preset install `nx` resolves transiently to `23.x`. devkit 22.x's `add-plugin.ts` calls the `nx/src/devkit-internals` export `retrieveProjectConfigurations([plugin], root, nxJson)` with a plain `LoadedNxPlugin[]` as the first argument (the nx 22 signature). nx 23 changed that argument to `SeparatedPlugins` and immediately does `separatedPlugins.specifiedPlugins.filter(...)`, so the array-shaped argument makes `.specifiedPlugins` `undefined` and generation crashes.

Observed stack (nx 23 leaked into the temp dir; devkit/webpack/node/nest all 22.7.6):

```
at retrieveProjectConfigurations (nx@23/retrieve-workspace-files.js)
at _addPluginInternal (@nx/devkit@22.7.6/add-plugin.js)
at webpackInitGeneratorInternal (@nx/webpack@22.7.6/init.js)
at @nx/node -> @nx/nest -> @nx/workspace:preset
```

## Expected Behavior

`retrieveProjectConfigurations` continues to accept the legacy plain `LoadedNxPlugin[]` first argument used by `@nx/devkit` <= 22 (which the peer range allows against nx 23), so `create-nx-workspace@22 --preset=nest` completes successfully.

The fix widens the first parameter to `SeparatedPlugins | LoadedNxPlugin[]` and normalizes the legacy array into `{ specifiedPlugins: arr, defaultPlugins: [] }` at the top of the function. This mirrors exactly what nx 23's own `add-plugin.ts` passes for the single-plugin case. A regression test drives the legacy array shape and fails without the fix.

## Related Issue(s)

Fixes #36144

## Scope note (post-review)

[...]

Issue #36144, opened 2026-06-29
```create-nx-workspace@22.7.6 --preset=nest``` fails with ```Cannot read properties of undefined (reading 'filter')```

### ISSUE THREAD

**Issue by robanders** [nrwl/nx#36144]
### Current Behavior

Creating a new workspace using ```create-nx-workspace@22.7.6``` with ```--preset=nest``` (and likely any preset that depends on ```@nx/webpack```) fails during workspace generation with ```TypeError: Cannot read properties of undefined (reading 'filter')```

### Expected Behavior

The workspace generation completes successfully

### Steps to Reproduce

1. run the command ```npx -y create-nx-workspace@22.7.6 --preset=nest```

### Nx Report

```shell
Node           : 22.17.1
OS             : darwin-arm64
Native Target  : aarch64-macos
npm            : 10.9.2
daemon         : Available
```

### Failure Logs

```shell
Failed to create a workspace: npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
npm warn deprecated glob@10.5.0: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
 NX   Cannot read properties of undefined (reading 'filter')
Pass --verbose to see the stacktrace.
 NX   Workspace creation failed, see above.
Error: Workspace creation failed, see above.
    at ChildProcess.<anonymous> (/private/var/folders/_b/fx55h1c56s1547rxhh2pqz_40000gp/T/tmp-85890-7fEnCO391C9K/node_modules/@nx/workspace/src/generators/new/generate-preset.js:46:24)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
```

### Package Manager Version

npm 10.9.2

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Root Cause Analysis

[...]

PR #36191, opened 2026-07-01
fix(core): pin nx to the running version in temp package installs

llwt:
## Current Behavior

`installPackageToTmp` (which backs `@nx/devkit`'s `ensurePackage`, among others) installs the requested package into a bare temp directory without pinning `nx`. `@nx/devkit` declares its `nx` peer dependency with a range spanning +/- 1 major (`>= 22 <= 24` on the 23.x line), so the package manager is free to satisfy that peer with a newer `nx` major than the one the workspace runs. The plugin code loaded from the temp directory then resolves that mismatched `nx`.

This is the root cause behind #36144: during `create-nx-workspace@22.7.6 --preset=nest`, `ensurePackage('@nx/webpack', '22.7.6')` installed `@nx/webpack@22.7.6` into a temp dir where `nx` floated to `23.x`, and `addPlugin` -> `retrieveProjectConfigurations` crashed on the changed argument shape.

The same class of breakage is live at the 23 -> 24 boundary today. Verified against a local Verdaccio registry with a fake `nx@24.0.0` published: `create-nx-workspace@23.0.1 --preset=nest`'s ensurePackage temp install pulled `nx@24.0.0` next to `@nx/webpack@23.0.1`. Any incompatible internal API change in a real nx 24 would reproduce the #36144 crash for every published 23.x CLI.

## Expected Behavior

Temp installs performed by `installPackageToTmp` pin `nx` to the version of the running process, so the temp directory always contains a matching `nx` and code loaded from it runs against the same nx the workspace uses. Installs of `nx` itself (the daemon pulling `nx@latest`, `nx init`, AI-agents setup) are unchanged.

Verified end-to-end against a local Verdaccio registry with a fake `nx@24.0.0` published (tagged `latest`):

| create-nx-workspace | ensurePackage temp dir | Result |
| --- | --- | --- |
| 23.0.1 (shipped, no fix) | `nx@24.0.0` next to `@nx/webpack@23.0.1` | mismatch (future crash) |
| 23.9.10 (this fix) | `nx@23.9.10` next to `@nx/webpack@23.9.10` | pinned, versions match |

A unit test locks the install command shape (fails with the exact unpinned command if the fix is reverted); the existing exact-command tests for installing `nx` itself keep passing, proving those flows are untouched.

## Related Issue(s)

Addresses the root cause behind #36144 going forward. The crash for already-published 22.x CLIs is covered by the backward-compat shim in #36189, which remains necessary (and should be backported to 23.0.x) because the shipped 22.x CLIs can only be reached through the `nx` package they transitively float to.

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/kind-wren-3766add2)

### DIFF
--- a/packages/nx/src/utils/package-json.spec.ts
+++ b/packages/nx/src/utils/package-json.spec.ts
@@ -19,6 +19,7 @@ import {
 } from './package-json';
 import * as pacakgeManager from './package-manager';
 import { getPackageManagerCommand } from './package-manager';
+import { nxVersion } from './versions';
 import { workspaceRoot } from './workspace-root';
 
 describe('buildTargetFromScript', () => {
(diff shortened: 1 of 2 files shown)


FrozenPandaz:
This was handled in #36295

PR #36200, opened 2026-07-02
fix(linter): reconcile angular-eslint v22 breaking changes in flat config

leosvelperez:
## Current Behavior

`angular-eslint` v22 dropped the legacy eslintrc config format and removed the `no-conflicting-lifecycle` rule. When nx converts an eslintrc Angular workspace to flat config (via the `@nx/eslint:convert-to-flat-config` generator or the 23.1 migration), it carries the removed `plugin:@angular-eslint/*` shared configs over as `FlatCompat` shims and may keep the removed rule. After the v22 bump those configs no longer resolve, so ESLint fails to load the flat config (`Could not find "no-conflicting-lifecycle" in plugin "@angular-eslint"`, and unresolvable `extends`). Project configs that an older nx already converted to flat config hit the same failure.

## Expected Behavior

The converted (or already-flat) config loads cleanly on `angular-eslint` v22. Both shim shapes the converter emits are reconciled to their flat-native equivalents:

- Top-level `...compat.extends(...)` and per-override `...compat.config({ extends }).map(...)` shared configs become `angular.configs.*`, scoped to the same files.
- `process-inline-templates` becomes a `processor` block, dropped when `flat/angular` already applies it.
- The removed `no-conflicting-lifecycle` rule is dropped.
- The `angular-eslint` import and devDependency are injected, and the now-unused `FlatCompat` scaffolding is stripped.

The reconciliation runs in both the `convert-to-flat-config` generator and the `update-23-1-0` migration, so fresh conversions and already-flat project configs are both covered. It is gated on `angular-eslint` v22+ (the shims still resolve on v18-v21).

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/nx-23.1.0-beta-migration-b1195096)

### DIFF
--- a/packages/eslint/src/migrations/update-23-1-0/convert-to-flat-config.ts
+++ b/packages/eslint/src/migrations/update-23-1-0/convert-to-flat-config.ts
@@ -75,18 +75,17 @@ export default async function update(tree: Tree): Promise<{
     // conversion, so the converted flat configs load without agent repair.
     await convertToFlatConfigGenerator(tree, {
       keepExistingVersions: true,
-      skipFormat: false,
+      skipFormat: true,
     });
   }
 
-  // Reconcile angular-eslint v22's removed eslintrc configs in any flat config
-  // present. Runs for every root state: project-level flat configs can exist
-  // even when the root is JavaScript-based (so the generator never ran), already
-  // flat, or absent entirely. No-op for the converted root above (idempotent).
-  // Gated on v22.
-  if (await migrateAngularEslintV22FlatConfig(tree)) {
-    await formatFiles(tree);
-  }
+  // Reconcile angular-eslint v22's removed eslintrc configs in any flat config.
+  // Runs for every root state: project-level flat configs can exist even when the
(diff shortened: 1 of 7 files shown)


packages/eslint/src/migrations/update-23-1-0/convert-to-flat-config.ts
```diff
@@ -62,18 +69,30 @@ export default async function update(tree: Tree): Promise<{
...
-    // No ESLint configuration to migrate.
-    return;
-  }
 
   if (rootState === 'convertible') {
+    // The generator reconciles angular-eslint v22's removed configs as part of
+    // conversion, so the converted flat configs load without agent repair.
     await convertToFlatConfigGenerator(tree, {
       keepExistingVersions: true,
       skipFormat: false,
```
FrozenPandaz: skip format here and then format separately below regardless.

FrozenPandaz (approved):

PR #36201, opened 2026-07-02
fix(linter): update terminal cli regex for local-dist build

### LINKED ISSUE #36199: [@nx/eslint-plugin] enforce-module-boundaries projectGraph cache validation broken since nx 22.7.0
### Current Behavior

The projectGraph cache validation is broken in the `@nx/eslint-plugin` package. Because of this for each file that is linted with @nx/enforce-module-boundaries the projectGraph is read and parsed from the file system again. I was able to debug this and found out that the method `ensureGlobalProjectGraph` in project-graph-utils was causing this increase in time. In 22.6.5 this method was called once and afterwards it was returning the cached ProjectGraph. In 22.7.0 and later this method is executed for every single file that is linted.

The issue is a regex that used to determine if it is a terminal run or not. `isTerminalRun` in runtime-lint-utils. The regex that is used for the NX CLI is not correct anymore, since the nx package is build/puvlished with local-dist. 
So `const NRWL_CLI_REGEX = /nx[\/\\]bin[\/\\]run-executor\.js$/;` should now be  `/nx[\/\\]dist[\/\\]bin[\/\\]run-executor\.js$/;`

Node modules structure of NX package:
Image

In case of a large monorepo and a large project with many files this can cause very slow linting. In our case the lint of one large project went from 1 minutes to 11 minutes because of this issue.

NX 22.7.0:
Image

NX 22.6.5:
Image

### Expected Behavior

The expected behavior is that the projectGraph is read once and reused from cache afterwards.

### GitHub Repo

_No response_

### Steps to Reproduce

1.to reproduce you can run linting for any project with the env var TIMING=1 to see the duration per rule

### Nx Report

```shell
Node           : 24.15.0
OS             : darwin-arm64
pnpm           : 11.6.0
daemon         : Available

nx                     : 22.7.6
@nx/js                 : 22.7.6
@nx/eslint             : 22.7.6
@nx/workspace          : 22.7.6
@nx/angular            : 22.7.6
@nx/jest               : 22.7.6
@nx/cypress            : 22.7.6
@nx/devkit             : 22.7.6
@nx/esbuild            : 22.7.6
@nx/eslint-plugin      : 22.7.6
@nx/module-federation  : 22.7.6
@nx/playwright         : 22.7.6
@nx/rspack             : 22.7.6
@nx/storybook          : 22.7.6
@nx/vite               : 22.7.6
@nx/vitest             : 22.7.6
@nx/web                : 22.7.6
@nx/webpack            : 22.7.6
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

Roozenboom:
## Current Behavior

Since NX 22.7.0 the NX package is published with a dist folder. This change causes the eslint-plugin to read the ProjectGraph for every file as the regex if it runs in the terminal never matched. This change updates the regex to include the dist folder

## Expected Behavior

Only the first file reads the ProjectGraph all the other files can read it from cache (globalThis). The regex for isTerminalRun should not impact this behavior

## Related Issue(s)

Fixes #36199

### DIFF
--- a/packages/eslint-plugin/src/utils/runtime-lint-utils.spec.ts
+++ b/packages/eslint-plugin/src/utils/runtime-lint-utils.spec.ts
@@ -402,7 +402,7 @@ describe('is terminal run', () => {
     // Mac: $run npx nx lint my-nx-project
     mockProcessArgv([
       '/Users/user/.nvm/versions/node/v16.13.0/bin/node',
-      '/Users/user/my-repo/node_modules/nx/bin/run-executor.js',
+      '/Users/user/my-repo/node_modules/nx/dist/bin/run-executor.js',
       '{"targetDescription":{"project":"my-nx-project","target":"lint"}',
     ]);
     expect(isTerminalRun()).toBe(true);
@@ -412,7 +412,7 @@ describe('is terminal run', () => {
     // Windows: $run npx nx lint my-nx-project
     mockProcessArgv([
       'C:\\Program Files\\nodejs\\node.exe',
-      'C:\\dev\\my-repo\\node_modules\\nx\\bin\\run-executor.js',
+      'C:\\dev\\my-repo\\node_modules\\nx\\dist\\bin\\run-executor.js',
       '{"targetDescription":{"project":"my-nx-project","target":"lint"}',
     ]);
     expect(isTerminalRun()).toBe(true);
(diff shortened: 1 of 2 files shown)


FrozenPandaz (commented):

FrozenPandaz (approved):

PR #36219, opened 2026-07-03
feat(docker): Add Docker and Podman support with configuration generator

Optischa:
## Current Behavior

Nx has no first-class way to build container images. Teams containerizing an app have to hand-write Dockerfiles and wire up `docker build`/`buildx`/`podman` invocations themselves as raw `run-commands` targets — with no caching integration, no generated Dockerfile starter, and no consistent way to produce build metadata (tags, labels, git context) across projects in the monorepo.

## Expected Behavior

A new `@nx/docker` plugin adds:

- **`nx g @nx/docker:configuration`** — scaffolds a `container` target on a project plus a starter `Dockerfile` (templates for an empty file, NestJS, Next.js, or an nginx static site), and lets you pick the engine (`docker` or `podman`).
- **`@nx/docker:build` executor** — builds the image via Docker Buildx or Podman, supporting multi-platform builds, build args/secrets/contexts, cache import/export, and auto-generated tags/labels/CSV metadata from git context (branch, sha, etc.).

Example — add container support to an existing app:

```shell
nx g @nx/docker:configuration --project=my-api --engine=docker --template=nest
```

This generates a `Dockerfile` (if one doesn't exist yet) and adds to `apps/my-api/project.json`:

```json
{
  "targets": {
    "container": {
      "executor": "@nx/docker:build",
      "options": {
        "engine": "docker",
        "context": "."
      }
    }
  }
}
```

Building the image then goes through the normal Nx pipeline, with caching and affected-detection like any other target:

```shell
nx container my-api
# or with overrides
nx container my-api --build-args=NODE_ENV=production --cache-from=type=registry,ref=myorg/my-api:cache
```

## Why this belongs in Nx

- **Monorepo-native**: containerization is configured per-project via `project.json`, so it composes with Nx's task graph, caching, and `nx affected` — only projects whose sources actually changed get rebuilt/rebuilt-cached.
- **Consistent generator-driven setup**: instead of every team copy-pasting Dockerfiles and build scripts, `nx g @nx/docker:configuration` gives a repeatable, framework-aware starting point (NestJS/Next.js/nginx templates included).
- **Fits the existing plugin model**: mirrors how other Nx plugins (e.g. `@nx/webpack`, `@nx/esbuild`) ship a generator + executor pair, so it's discoverable and usable the same way users already work with Nx.
- **Removes boilerplate `run-commands` hacks**: normalizes options (build args, secrets, cache, multi-platform) behind a typed schema instead of ad-hoc shell strings, and supports both Docker and Podman so teams aren't locked into one engine.

Marked **experimental** for now (see `packages/docker/readme-template.md`) since the executor surface may still evolve before it's covered by semver guarantees.


Optischa:
@FrozenPandaz, would that be something for the 23.1.0 release, as a somewhat larger update to address the Docker change?

PR #36223, opened 2026-07-03
fix(core): speed up lockfile parsing and catalog resolution

leosvelperez:
## Current Behavior

Lockfile stringification (used when pruning a lockfile down to a subset of the graph) matched every external node against every package key, so the cost grew with the number of nodes times the total packages in the lockfile. On a real 3100-package pnpm lockfile this quadratic match dominated the whole operation. The pnpm, npm (v3), and yarn classic parsers all shared this full-scan shape. Separately, catalog resolution created a manager and re-read the workspace catalog file (pnpm-workspace.yaml / .yarnrc.yml) on every dependency reference.

## Expected Behavior

Package keys are bucketed by name once, so each node scans only its own name's versions. On the same 3100-package pnpm lockfile, stringifying drops from ~1240ms to ~49ms (about 25x): the quadratic key match collapses from ~1.2s to a few ms, while parse and dump are unchanged. Catalog managers are created once per pass and cache their parsed definitions per root. Workspace-only pnpm lockfiles, which omit the packages block when there are no external dependencies, are handled without error.

The pruned lockfile and project graph output are unchanged.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/catalogs-perf-a52afa38)

### DIFF
--- a/packages/nx/src/utils/catalog/yarn-manager.ts
+++ b/packages/nx/src/utils/catalog/yarn-manager.ts
@@ -21,6 +21,11 @@ const YARNRC_FILENAME = '.yarnrc.yml';
 export class YarnCatalogManager implements CatalogManager {
   readonly name = 'yarn';
   readonly catalogProtocol = 'catalog:';
+  // Parsed definitions cached per root. A manager is created per operation
+  // (getCatalogManager news one up), so defs are read once per pass instead of
+  // once per catalog reference. Only the fs (string-root) branch is cached; the
+  // Tree branch stays live since the tree is mutable within a generator.
+  private definitionsByRoot = new Map<string, CatalogDefinitions | null>();
 
   isCatalogReference(version: string): boolean {
     return version.startsWith(this.catalogProtocol);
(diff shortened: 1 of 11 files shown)


packages/nx/src/utils/catalog/yarn-manager.ts
```diff
@@ -21,6 +21,11 @@ const YARNRC_FILENAME = '.yarnrc.yml';
 export class YarnCatalogManager implements CatalogManager {
   readonly name = 'yarn';
   readonly catalogProtocol = 'catalog:';
+  // Parsed definitions cached per root. A manager is created per operation
+  // (getCatalogManager news one up), so defs are read once per pass instead of
+  // once per catalog reference. Only the fs (string-root) branch is cached; the
+  // Tree branch stays live since the tree is mutable within a generator.
+  private definitionsByRoot = new Map<string, CatalogDefinitions | null>();
```
jaysoo, line 22: Minor: Rule-of-three. Since exact pattern of map to catalog defs is repeated four times, maybe a shared util would be better.

jaysoo (approved):

packages/nx/src/utils/catalog/yarn-manager.ts
leosvelperez: It's really only repeated twice (devkit's are a copy of nx's), but it does still make sense to reuse a helper. That's fixed now.

PR #36234, opened 2026-07-06
docs(misc): note that package managers can override .env variables

### LINKED ISSUE #30298: `.npmrc` environment variables override those in `.env` files
### Current Behavior

`.npmrc` environment variables override those in `.env` files described in the [docs](https://nx.dev/recipes/tips-n-tricks/define-environment-variables#set-environment-variables) when nx is invoked `npm run nx ...` or `npx nx ...`, but not when invoked directly.

This behavior was observed using the `NODE_OPTIONS=--max-old-space-size=8192` variable.

The [docs](https://nx.dev/recipes/tips-n-tricks/define-environment-variables#set-environment-variables) suggest environment variables are overridden by those with the most specific scope. However an `.npmrc` file in **any** location: project, workspace root, external parent directories, or even the user's global `.npmrc`, all of these will override `.env` variables.

### Expected Behavior

`.npmrc` environment variables do **not** override those in `.env` files when when nx is invoked `npm run nx ...` or `npx nx ...`.

If this cannot be fixed or is as-designed, I suggest adding a note in the [docs](https://nx.dev/recipes/tips-n-tricks/define-environment-variables#set-environment-variables) to warn about this interaction. The cause is not obvious and is mistakable for a similar limitation of the `dotenv` package where the `NODE_OPTIONS` variable cannot be set while node is running.

### GitHub Repo

https://github.com/tlaurie-cl/npmrc-overrides-nx-env

### Steps to Reproduce

1. Install nx globally:
	* npm install -g nx
2. Run nx directly via:
	* nx run example:build
	* Notice the heap_size_limit is 8640266240 B (~8 GiB)
3. Run nx with npm via:
	* npm run nx run example:build
	* Notice the heap_size_limit is 4345298944 B (~4 GiB)
3. Run nx with npx via:
	* npx nx run example:build
	* Notice the heap_size_limit is 4345298944 B (~4 GiB)

### Nx Report

```shell
Node           : 18.19.1
OS             : win32-x64
npm            : 10.2.4

nx (global)  : 20.5.0
nx           : 20.5.0
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

leosvelperez:
## Current Behavior

The environment variables guide explains that Nx ignores a variable that is already loaded into the process, but it doesn't mention that a package manager can be what loads it. Running Nx through `npm run` or `npx` can set variables before Nx reads `.env` files (for example, npm's `node-options` config becomes a real `NODE_OPTIONS` variable). The `.env` value is then silently ignored, while running `nx` directly uses it. Users hit this and mistake it for an Nx bug.

## Expected Behavior

The guide documents the interaction. A new caution aside in the environment variables guide explains that a package manager can set variables before Nx runs, uses npm's `node-options` -> `NODE_OPTIONS` as the concrete example, and clarifies that this is the package manager's behavior, not Nx overriding your files.

## Why documentation only

When Nx runs via `npm run` / `npx`, npm translates its `node-options` config (from any `.npmrc`: project, user, or global) into a real `NODE_OPTIONS` environment variable before Nx starts. Nx loads `.env` files with dotenv's default `override: false`, so a variable already present in the environment wins. Direct `nx` runs have no such variable, so `.env` applies.

This process-env-wins behavior is intentional and already documented (it protects system variables like `NODE_ENV`). Letting `.env` override it would regress that protection, and Nx can't reliably distinguish an npm-injected variable from a genuine shell or CI one: npm exposes only the merged value, not its origin. Documenting the interaction is the correct fix.

## Related Issue(s)

Fixes #30298

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-30298-1d731484)

### DIFF
--- a/astro-docs/src/content/docs/guides/Tips-n-Tricks/define-environment-variables.mdoc
+++ b/astro-docs/src/content/docs/guides/Tips-n-Tricks/define-environment-variables.mdoc
@@ -79,9 +79,9 @@ for workspace-specific settings (like the [Nx Cloud token](/docs/guides/nx-cloud
 {% aside type="caution" title="Package managers can set variables before Nx runs" %}
 The rule above also applies to variables you didn't set yourself. A package manager can put variables into the process before Nx reads any `.env` file, and Nx keeps those values.
 
-For example, npm's `node-options` setting (from a project, user, or global `.npmrc`) becomes a `NODE_OPTIONS` variable when you run `npm run nx ...` or `npx nx ...`. If you also set `NODE_OPTIONS` in `.env`, the `.npmrc` value wins, while running `nx` directly uses the `.env` value.
+For example, the npm `node-options` setting (from a project, user, or global `.npmrc`) becomes a `NODE_OPTIONS` variable when you run `npm run nx ...` or `npx nx ...`. If you also set `NODE_OPTIONS` in `.env`, the `.npmrc` value wins, while running `nx` directly uses the `.env` value.
 
-This is the package manager's behavior, not Nx overriding your files. If you rely on a variable from `.env`, make sure nothing else sets it, such as npm's `node-options`.
+The package manager sets the value, not Nx. If you rely on a variable from `.env`, make sure nothing else sets it first.
 {% /aside %}
 
 {% aside type="caution" title="Env files are not loaded in batch mode" %}
--- a/astro-docs/src/content/docs/guides/Tips-n-Tricks/define-environment-variables.mdoc
+++ b/astro-docs/src/content/docs/guides/Tips-n-Tricks/define-environment-variables.mdoc
@@ -76,6 +76,14 @@ We recommend nesting your **app** specific `env` files in `apps/your-app`, and c
 for workspace-specific settings (like the [Nx Cloud token](/docs/guides/nx-cloud/access-tokens)).
 {% /aside %}
 
+{% aside type="caution" title="Package managers can set variables before Nx runs" %}
+The rule above also applies to variables you didn't set yourself. A package manager can put variables into the process before Nx reads any `.env` file, and Nx keeps those values.
+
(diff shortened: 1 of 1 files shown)


astro-docs/src/content/docs/guides/Tips-n-Tricks/define-environment-variables.mdoc
```diff
@@ -76,6 +76,14 @@ We recommend nesting your **app** specific `env` files in `apps/your-app`, and c
 for workspace-specific settings (like the [Nx Cloud token](/docs/guides/nx-cloud/access-tokens)).
 {% /aside %}
 
+{% aside type="caution" title="Package managers can set variables before Nx runs" %}
```
jaysoo, line 79: Style guide violations, mostly around use of possessives.

```
Findings

  1. Possessive on a product name (violation) — Possessives rule

  ▎ "Don't use possessives on product names."

  npm's node-options appears twice. Fix to the npm node-options setting / the node-options npm setting. (package manager's is fine — not a product name.)

  2. Mild restatement closer — No restatement closers / structural
  "This is the package manager's behavior, not Nx overriding your files" restates the opening ("Nx keeps those values"). The actionable half ("make sure nothing else sets it") is the keeper. Could tighten to one sentence:

  ▎ The package manager sets the value, not Nx. If you rely on a .env variable, make sure nothing else sets it first.

  Clean

  - Contractions used (didn't) ✓
  - No em/en dashes, no semicolons ✓
  - Second person, direct, action-oriented ✓
  - No banned AI phrases, no marketing language ✓
  - caution aside type is valid ✓
```

Checked against `astro-docs/STYLE_GUIDE.md`.

jaysoo (approved):

PR #36266, opened 2026-07-08
fix(core): avoid retaining matched file lists in fileset hash caches

leosvelperez:
## Current Behavior

The task hasher's persistent fileset caches store the full list of matched file paths alongside each fileset hash. The path lists stay in memory for the lifetime of the hasher even though they are only consumed when task inputs are collected (e.g. for graph tooling), so cache retention grows linearly with the number of matched files in the workspace. The lists are also built on every cache miss when nothing will read them, and the hash plan inspector computes fileset hashes as a side effect of listing matched files.

## Expected Behavior

The persistent caches store only the fileset hashes. Matched file lists are expanded only when inputs are actually collected, in per-invocation caches that are freed once hashing returns, so the default hashing path never allocates them. The inspector lists matched files without hashing. Task hashes, hash details, and collected inputs are unchanged, verified by comparing full per-task hash dumps between builds.

> [!NOTE]
> A follow-up PR (#36267) builds on this to deduplicate the hash detail strings themselves, reducing peak hashing memory on large workspaces.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-36152-655cd716)

### DIFF
--- a/packages/nx/src/native/tasks/hash_plan_inspector.rs
+++ b/packages/nx/src/native/tasks/hash_plan_inspector.rs
@@ -1,6 +1,6 @@
 use crate::native::tasks::hashers::{
-    ProjectFileSetCache, collect_json_input_files, hash_project_files_with_inputs_cached,
-    hash_workspace_files_with_inputs, resolve_task_output_files,
+    ProjectFilePathsCache, collect_json_input_files, collect_project_file_paths_cached,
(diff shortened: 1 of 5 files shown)


leosvelperez:
Folded into #36267; the retention improvement alone is too small to warrant a standalone PR.

PR #36282, opened 2026-07-09
fix(misc): export @nx/esbuild/executors entry point

stefreak:
## Current Behavior

`@nx/esbuild` exposes `executors.json`, but it does not expose a public TypeScript entry point for importing its executor implementation or executor options.

Other Nx plugin packages already expose small public facade entry points for this kind of package-level integration:

- `@nx/vitest/executors`
- `@nx/js/typescript`
- `@nx/eslint/plugin`

## This PR

This adds `@nx/esbuild/executors` with the same package export shape:

- `@nx/nx-source` points to `./executors.ts`
- `types` points to `./dist/executors.d.ts`
- `default` points to `./dist/executors.js`

The new facade exports `esbuildExecutor` and `EsBuildExecutorOptions`.


AgentEnder (approved):

stefreak:
@AgentEnder thank you for the approval! Could you approve workflows again? The last attempt failed due to flaky tests 🙏

meeroslav (approved):

PR #36294, opened 2026-07-10
fix(js): bump @swc/cli to 0.8.1 to patch critical decompress advisory

leosvelperez:
## Current Behavior

The daily NPM Audit workflow (`.github/workflows/npm-audit.yml`, running `pnpm dlx audit-ci --critical`) fails on a critical advisory. `@swc/cli@0.8.0` resolves `@xhmikosr/decompress@10.0.1` through its `@xhmikosr/bin-wrapper` -> `@xhmikosr/downloader` chain, and that version is vulnerable to [GHSA-mp2f-45pm-3cg9](https://github.com/advisories/GHSA-mp2f-45pm-3cg9) (critical): archive extraction can create files and links outside the target directory. Vulnerable range `< 10.2.1`; patched in `10.2.1` and `11.1.3`.

## Expected Behavior

`@swc/cli` is bumped to `0.8.1`, which moves to `@xhmikosr/bin-wrapper@14` -> `@xhmikosr/downloader@16` -> `@xhmikosr/decompress@11.1.3` (patched), so the audit passes with `critical: 0`. The bump stays within `@nx/js`'s existing `@swc/cli` peer range (`>=0.6.0 <0.9.0`) and needs no other `@swc` change (`@swc/cli`'s `@swc/core` peer is unchanged at `^1.2.66`).

The scaffolded `@swc/cli` floor in `@nx/js` moves to `~0.8.1`, and a `packageJsonUpdates` migration bumps existing projects still on `0.8.0` to `~0.8.1` so they re-resolve off the vulnerable version.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/check-critical-audit-failure-a3634f55)

### DIFF
--- a/packages/js/migrations.json
+++ b/packages/js/migrations.json
@@ -139,6 +139,18 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "23.1.0-swc-cli": {
+      "version": "23.1.0-rc.3",
+      "requires": {
+        "@swc/cli": ">=0.8.0 <0.8.1"
+      },
+      "packages": {
+        "@swc/cli": {
+          "version": "~0.8.1",
+          "alwaysAddToPackageJson": false
+        }
+      }
     }
   }
 }
--- a/packages/js/src/utils/versions.ts
+++ b/packages/js/src/utils/versions.ts
@@ -4,7 +4,7 @@ export const nxVersion = require(join('@nx/js', 'package.json')).version;
 
 export const esbuildVersion = '^0.27.0';
 export const prettierVersion = '~3.6.2';
-export const swcCliVersion = '~0.8.0';
+export const swcCliVersion = '~0.8.1';
 export const swcCoreVersion = '~1.15.5';
 export const swcHelpersVersion = '~0.5.18';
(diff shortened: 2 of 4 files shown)


packages/js/migrations.json
```diff
@@ -139,6 +139,18 @@
           "alwaysAddToPackageJson": false
         }
       }
+    },
+    "23.1.0-swc-cli": {
+      "version": "23.1.0-rc.3",
+      "requires": {
+        "@swc/cli": ">=0.8.0 <0.8.1"
```
meeroslav, line 146: This is literally saying `=0.8.0`. Why not migrate all `<0.8.1` to that safe version?

meeroslav (approved):
Approved but left a question.

packages/js/migrations.json
leosvelperez: There are prior package updates in this file that would have bumped you to at least 0.8.0. We build each package update on top of prior updates. If, for whatever reason, a workspace has opted out of the prior ~0.8.0 bump and stayed behind, we don't force the update.

PR #36298, opened 2026-07-10
feat(angular-rspack): scaffold and run SSR apps on the @angular/ssr application engine

leosvelperez:
## Current Behavior

`@nx/angular-rspack` cannot run a server written against the `@angular/ssr` application engine, which is the shape the Angular CLI scaffolds. The engine reads manifests the build never produces, so it throws at startup and prerendering finds no bootstrap. A build cannot declare a server output mode either, nor the hosts a deployed server answers. The docs list the engine gap as a known limitation.

Generating an application with `--bundler=rspack --ssr` scaffolds a server built on `CommonEngine`, the API Angular deprecated in v22. It is created with no options, and from `@angular/ssr` 20.3.17 the engine checks the request host against a list that stays empty without them. Every request is then rejected. Version 22 answers 500, and the earlier versions log an error and fall back to client-side rendering. The servers scaffolded for webpack SSR applications and for module federation hosts and remotes carry the same empty list.

## Expected Behavior

An SSR build wires the server bundle for the application engine, so a server written against those APIs renders. `outputMode: 'server'` is accepted with the application builder's meaning, and `security.allowedHosts` sets the hosts a deployed server answers. While serving, the local hosts and the dev server host are allowed without any configuration.

A new `--bundler=rspack --ssr` application is scaffolded on the application engine, with server routes and a server config, and it renders on the first run. The scaffolds that still use `CommonEngine` allow the local hosts. The application builder and rspack setups write an empty list for the deployment to fill in.

## Implementation Notes

- The generated rspack config keeps `prerender` instead of setting `outputMode: 'server'`. Under rspack the output mode turns build-time prerendering off, which would drop the `/` page the scaffold produces today.
- An allowed hosts list that is already configured is never overwritten. Angular's schematic assigns an empty one unconditionally, which would turn a working deployment into one that answers nothing.
- Host configuration is scaffolded only when every `@angular/ssr` version the workspace can resolve accepts it. The declared range is raised to reach one, so `~20.3.0` becomes `~20.3.17`. Prereleases are not matched, because the option landed part-way through the 21.2 prerelease train.
- A non-empty `security.allowedHosts` fails the build on an `@angular/ssr` that ignores it, rather than passing silently. An empty list is accepted on every version.
- Unsupported surfaces are rejected instead of misbehaving: the `static` output mode, `security.autoCsp`, and locale inlining combined with `outputMode`.
- Converting an existing SSR project folds the server builder's tsconfig into every tsconfig the build compiles with, so its server sources are type checked again. The converted project keeps its `CommonEngine` server, which is user-owned source.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-34936-4f8ace2a)


AgentEnder (commented):
lgtm. The approach lines up with what the esbuild application builder does, and the `@angular/ssr` version gates all check out against the published packages.

Comments inline on the hunks. The one worth acting on before merge is the source map in `platform-server-exports.loader.ts`: the prologue shifts every line of the user's entry while the incoming map is still forwarded unchanged, and it lands on every SSR build rather than only engine-wired ones.

The rest are one-liners, mostly docs.

Separately, generated rspack SSR apps look broken on `@angular/ssr` 22: `server.ts__tmpl__` builds `CommonEngine` with no `allowedHosts`, and 22 rethrows where 21.2.4 fell back to CSR. Different package and not yours, I'll file it.

packages/angular-rspack/src/lib/plugins/loaders/platform-server-exports.loader.ts
```diff
@@ -2,29 +2,68 @@ import {
...
+        ɵextractRoutesAndCreateRouteTree,
+        ɵdestroyAngularServerApp,
+      } from '@angular/ssr';
+      export { default as __ngRspackMainServerBootstrap } from ${JSON.stringify(
+        engineWiring.mainServerEntry
+      )};
+    `;
+  }
+
+  let source = `${prologue}
```
AgentEnder: The prologue shifts the user's entry but `this.callback(null, source, map)` still forwards the incoming map, so it no longer describes what is emitted.

| path | shift |
| --- | --- |
| no engine wiring (`prologue` empty) | 1 line |
| virtual module | 1 line |
| inline fallback (rspack < 1.5) | 44 lines |

The first row is the one I would flag: `${prologue}` is followed by a literal newline, so the shift lands even when there is no wiring at all. Base kept content on line 1 and shifted only for `import 'zone.js/node'`. `devtool` is `'source-map'` whenever `sourceMap.scripts` is on.

`${prologue}${content}` takes the virtual-module path to zero. For the inline path, either keep `generateEngineManifestSource` on one line (`//` comments would need to become `/* */`) or stop forwarding a map you know is wrong.

astro-docs/src/content/docs/kb/create-config.mdoc
```diff
@@ -313,6 +313,12 @@ Controls build optimization settings for scripts, styles, and fonts.
 `OutputHashing` `default: 'none'`
 Defines the hashing strategy for output files. Can be 'none', 'all', 'media', or 'bundles'.
 
+### outputMode
```
AgentEnder, line 321: These two are missing from the `AngularRspackPluginOptions` block above, which is introduced as "contains the following properties" and lists 25 while the prose now lists 27.

`outputMode?: 'server' | 'static';` after `outputHashing`, and `security?: { allowedHosts?: string[]; autoCsp?: boolean | { unsafeEval?: boolean } };` after `scripts`. Both land in the existing alphabetical order.

(The block has been missing ~25 other options for a while, that part isn't yours.)

astro-docs/src/content/docs/kb/create-config.mdoc
```diff
@@ -313,6 +313,12 @@ Controls build optimization settings for scripts, styles, and fonts.
...
+
+`'server' | 'static'`
+Defines the build output target for server-side rendered applications.
+Only 'server' is currently supported; 'static' (build-time prerendering of the full application) is rejected.
```
AgentEnder: Semicolon, and the same on line 346. `STYLE_GUIDE.md:408` asks for two sentences instead. vale has no rule for it so CI stays green.

The ones inside the `{ allowedHosts?: string[]; autoCsp?: ... }` signature are code and fine.

astro-docs/src/content/docs/technologies/angular/angular-rspack/introduction.mdoc
```diff
@@ -52,11 +52,28 @@ Please note that Angular Rspack support is still experimental and is not yet con
...
+- The server route configuration is extracted at runtime on the first request of each server process, rather than at build time.
+  Route configuration errors surface at request time instead of build time.
+- Automatic content security policy generation (`security.autoCsp`) is not supported.
+- Locale inlining (`localize`) disables the `@angular/ssr` application engine wiring, so it cannot be used by server entries written against the application engine APIs and cannot be combined with the `outputMode` option.
 
 If you have any other missing features or limitations, please [let us know](https://github.com/nrwl/angular-rspack/issues/new).
 
+{% aside type="note" title="Application engine host validation" %}
+
+Since `@angular/ssr` version 21.2.0 (also backported to 20.3.17 and 21.1.5), the application engine validates the request host against its allowlist; earlier versions do not validate hosts.
```
AgentEnder: Semicolon here too, same `STYLE_GUIDE.md:408` rule.

astro-docs/src/content/docs/technologies/angular/angular-rspack/introduction.mdoc
```diff
@@ -52,11 +52,28 @@ Please note that Angular Rspack support is still experimental and is not yet con
...
 
-- Server Routing is not supported - still experimental in Angular currently.
-- App Engine APIs are not supported - still experimental in Angular currently.
+- The `static` output mode (build-time prerendering of the full application) is not supported.
+- Server routes marked for build-time prerendering (`RenderMode.Prerender`) are rendered at request time instead.
+  Each request repeats the full render and critical CSS inlining, without the response streaming and critical CSS caching applied to server-rendered routes.
+- The server route configuration is extracted at runtime on the first request of each server process, rather than at build time.
+  Route configuration errors surface at request time instead of build time.
+- Automatic content security policy generation (`security.autoCsp`) is not supported.
+- Locale inlining (`localize`) disables the `@angular/ssr` application engine wiring, so it cannot be used by server entries written against the application engine APIs and cannot be combined with the `outputMode` option.
```
AgentEnder: Worth a line here for the `@rspack/core` < 1.5 case: the manifest registration is inlined into the entry and runs after the entry's own imports, rather than before them. The scaffolded `server.ts` builds the engine in the entry body so it works either way, but it bites when the engine is constructed at module scope of an imported module, and nothing warns.

packages/angular-rspack/src/lib/config/config-utils/server-config.ts
```diff
@@ -103,3 +184,75 @@ export async function getServerConfig(
...
+    new Set([
+      ...allowedHosts,
+      // The engine matcher has no implicit local set; seed the hosts the dev
+      // server accepts implicitly. Other IP literals cannot be enumerated;
+      // NG_ALLOWED_HOSTS covers those.
+      'localhost',
+      '*.localhost',
+      '127.0.0.1',
+      '[::1]',
+      devServer.host,
```
AgentEnder: `devServer.host` goes in verbatim, so `--host 0.0.0.0` seeds `0.0.0.0` into the engine allowlist. A bind-all address is never a legitimate request `Host`, and it is a known loopback-bypass value in browsers. Worth skipping `0.0.0.0` / `::` / `[::]` here.

packages/angular-rspack/src/lib/plugins/loaders/angular-partial-transform.loader.ts
```diff
@@ -49,6 +51,14 @@ export default function loader(
       return;
     }
 
+    // The engine manifest virtual module mentions '@angular' but needs no
+    // transform, and it has no on-disk file for the transformer worker to
+    // read.
+    if (basename(request) === ENGINE_MANIFEST_VIRTUAL_NAME) {
```
AgentEnder: Matching on `basename` means any file called `__ng-rspack-ssr-entry-manifest__.js` anywhere in the graph skips the partial-link transform. Comparing against the full `engineWiring.manifestModuleRequest` would be exact.

packages/angular-rspack/src/lib/plugins/tools/routes-extractor-worker.ts
```diff
@@ -41,15 +51,17 @@ async function extract(): Promise<string[]> {
...
     default: bootstrapAppFn,
+    __ngRspackMainServerBootstrap,
   } = require(serverBundlePath) as ServerBundleExports;
 
   const browserIndexInputPath = path.join(outputPath, indexFile);
   const document = await fs.promises.readFile(browserIndexInputPath, 'utf8');
 
-  const bootstrapAppFnOrModule = bootstrapAppFn || AppServerModule;
+  const bootstrapAppFnOrModule =
+    bootstrapAppFn || __ngRspackMainServerBootstrap || AppServerModule;
```
AgentEnder, line 61: Nothing covers this fallback: deleting `__ngRspackMainServerBootstrap` from the chain leaves all 208 tests green. A worker-level test asserting it resolves when the bundle exports no `default` would close it.

packages/angular-rspack/src/lib/models/angular-rspack-plugin-options.ts
```diff
@@ -292,6 +292,17 @@ export interface AngularRspackPluginOptions {
...
+   * Defines the build output target. Only 'server' is currently supported:
+   * it requires the `server` and `ssr.entry` options and the `@angular/ssr`
+   * package, and disables build-time prerendering (`prerender`/`appShell`).
+   * The `@angular/ssr` application engine wiring itself is active for any
+   * SSR build with `@angular/ssr` installed, with or without this option,
+   * except when locale inlining is enabled.
+   * 'static' (build-time prerendering of the full application) is not
+   * supported yet and is rejected.
+   */
+  outputMode?: 'server' | 'static';
```
AgentEnder: Out of the file's alphabetical order, sitting between `prerender` and `preserveSymlinks`. Belongs next to `outputHashing` / `outputPath`. (`security?` below is placed right.)

leosvelperez:
All nine addressed. One I decided differently, so calling it out.

I implemented the `0.0.0.0` / `::` / `[::]` skip and then removed it, on parity grounds. `@angular/build` does `allowedHosts.push(serverOptions.host)` unfiltered, under a comment reading "Always allow the dev server host". Every vite from 5.4 to 8.1 accepts any IP literal before it consults `allowedHosts`, and `@rspack/dev-server` does the same, so our own browser dev path already takes `Host: 0.0.0.0`. The skip would make `nx serve` return 400 where `ng serve` with the same flags works, and make our SSR path stricter than our browser path.

The premise does not hold either. Chromium 148 navigating to `http://0.0.0.0:4200`, against a server bound only to `127.0.0.1`, sends `Host: 0.0.0.0:4200`. With the skip, `@angular/ssr` 22.1.2 answers that with `Header "host" with value "0.0.0.0:4200" is not allowed.` and a 400. Firefox 150 refuses the connection and WebKit 26.4 sends nothing.

The engine has no implicit local set of its own, so we seed `localhost`, `*.localhost`, `127.0.0.1` and `[::1]` unconditionally. Angular does not, and it costs them: with default options its manifest allowlist is just `['localhost']`, so `ng serve` 400s on `http://127.0.0.1:4200` even though vite waves the same request through. Against that backdrop `0.0.0.0` is one more exact alias in a set we already seed deliberately, on a value that only appears when someone passes `--host 0.0.0.0`. It does nothing for DNS rebinding, where the request carries the attacker's hostname. One real asymmetry: Angular's SSR sits behind vite's host middleware, ours is a forked process with nothing in front of it.

The comment did surface a real bug next to it. Bare `::` was seeded verbatim and could never match, since IPv6 hostnames arrive bracketed as `[::]`. The seeded host is bracketed now.

The rest:

- sourcemap: prefix built as whole lines, map shifted down by however many it adds. Covers zone.js, the virtual module, the inlined manifest, and the stray blank first line.
- manifest `basename` match: exact `resourcePath` compare against the absolute path, passed as a loader option.
- bootstrap fallback: has a spec.
- the three semicolons: split into sentences.
- `outputMode` and `security`: added to the options block.
- `@rspack/core` < 1.5: new bullet in the limitations.
- `outputMode` ordering: moved next to `outputHashing`.

Two version comments were also wrong and are fixed. Pre-22 `@angular/ssr` does not always validate hosts, it starts at 20.3.17 / 21.1.5 / 21.2.0.

AgentEnder (approved):

PR #34820, opened 2026-03-13
fix(core): error with helpful error instead of looping nx invocations

AgentEnder:
This pull request introduces a mechanism to detect and prevent recursive task invocation loops in Nx by tracking task invocation chains across nested Nx processes. The main changes focus on adding an environment variable (`NX_TASK_INVOCATION_CHAIN`) that accumulates task keys, and implementing logic to halt execution if a task is re-invoked within the same chain. Additionally, new tests are added to verify this behavior.

**Recursive Task Invocation Detection:**

* Added the `NX_TASK_INVOCATION_CHAIN` environment variable to track the chain of task invocations across nested Nx processes in `getEnvVariablesForTask`. This helps identify when a task is re-invoked, preventing infinite loops. [[1]](diffhunk://#diff-6235da82f41dcbb027ac22e4b01d8c50390dcca9faf6ef531a8b1f442b583e81R115-R121) [[2]](diffhunk://#diff-6235da82f41dcbb027ac22e4b01d8c50390dcca9faf6ef531a8b1f442b583e81R133)
* Implemented the `detectTaskInvocationLoop` method in `TaskOrchestrator`, which checks the invocation chain and terminates the process with an error if a recursive loop is detected. This method is called before starting or running any task. [[1]](diffhunk://#diff-e9bae83332b3d6e57c023959ab2e5f191c97e0a154a8c1d36dd81f8f869e1bdfR568-R569) [[2]](diffhunk://#diff-e9bae83332b3d6e57c023959ab2e5f191c97e0a154a8c1d36dd81f8f869e1bdfR855-R856) [[3]](diffhunk://#diff-e9bae83332b3d6e57c023959ab2e5f191c97e0a154a8c1d36dd81f8f869e1bdfR1136-R1171)

**Testing:**

* Added unit tests in `task-env.spec.ts` to verify that the invocation chain is correctly set and accumulated, and that loops are detected.

**Codebase Clean-up:**

[...]

PR #36331, opened 2026-07-14
fix(core): avoid false positive in recursive task invocation detection

bojanbass:
## Current Behavior

`detectTaskInvocationLoop` (introduced in #34820) keys the loop-detection registry on `task.id` alone (`project:target:configuration`). This does not account for CLI overrides, so a target that is legitimately invoked multiple times with different overrides gets misreported as an infinite loop.

Concretely, a `nx:run-commands` target like this:

```json
"import-minio": {
  "options": {
    "commands": [
      "nx run infra:import-minio --configuration=review --bucket=media-files --sourceFilePath=dumps/media-files/bucket.tar.gz",
      "nx run infra:import-minio --configuration=review --bucket=media --sourceFilePath=dumps/media/bucket.tar.gz",
      "nx run infra:import-minio --configuration=review --bucket=invoices --sourceFilePath=dumps/invoices/bucket.tar.gz"
    ]
  }
}
```

fails on the second nested `nx run` with:

```
NX   Recursive task invocation detected

Nx detected a recursive loop of task invocations:

   -> infra:import-minio:review

Task "infra:import-minio:review" was already invoked by a parent Nx process in this chain.
```

even though these are three distinct operations (different bucket, different source file) — nothing is actually recursing. Each nested `nx run` inherits `NX_INVOCATION_ROOT_PID` from the top-level process, and the tracker's DB registration is keyed only on `(root_pid, task.id)`, so the second and third calls collide with the first purely because `task.id` is identical, ignoring the different `--bucket`/`--sourceFilePath` overrides.

## Expected Behavior

Targets invoked multiple times with different CLI overrides should not be flagged as recursive, while a genuine self-invoking loop (same id *and* same overrides) should still be caught.

## Related Issue(s)

I couldn't find an existing issue for this specific false positive (searched for "Recursive task invocation detected"), so opening this fix directly with a repro in the description above. Happy to file a companion issue if preferred.

## Fix

Fold `task.overrides` into the identity used by `detectTaskInvocationLoop` (and the matching `unregisterTask`/`registeredInvocations` cleanup), via a new `getTaskInvocationKey(task)` helper. Falls back to plain `task.id` when there are no overrides, so behavior for existing non-parameterized recursion cases is unchanged.

Added tests in `task-orchestrator.spec.ts` covering:
- same target invoked repeatedly with different overrides → not flagged
- same target + identical overrides invoked from two orchestrator instances (simulating nested Nx processes) → still flagged
- same target with no overrides invoked twice → still flagged (unchanged behavior)


AgentEnder (approved):

AgentEnder:

<details>
<summary>

</summary>

>
> No linked issue, but already approved, so the change is agreed. The body identifies the cause precisely: `detectTaskInvocationLoop` from #34820 keys the registry on `task.id` alone, which ignores CLI overrides.
>
> Branch merges cleanly. The only thing that has ever blocked this is CI.

</details>

Issue #30670, opened 2025-04-10
nx affected -t nx-release-publish does not run version/changelog presteps.


h3adache:
### Current Behavior

Running

```
pnpm nx affected -t nx-release-publish
```
does not run version and changelog presteps.

This leaves any `workspace:*` dependencies untouched and fails to publish.

Running
```
pnpm nx release
```
will properly run the presteps.

I've even tried setting `preserveLocalDependencyProtocols: false` to no avail.

### Expected Behavior

```pnpm nx affected -t nx-release-publish``` has parity with ```pnpm nx release```

### Steps to Reproduce

1. init a new nx monorepo with `pnpm` package manager `npx create-nx-workspace --pm pnpm`
2. generate publishable lib `pnpm nx g @nx/react:lib packages/my-lib --buildable --publishable`
3. generate 2nd publishable lib `pnpm nx g @nx/react:lib packages/my-lib2 --buildable --publishable`
4. add a dependency on `my-lib2` in `my-lib1` (I just use `@nx/dependency-checks` for this)
5. run `pnpm nx affected -t nx-release-publish`
6. observe difference with `pnpm nx release` (but that cannot target affected)

### Nx Report

```shell
Node           : 22.11.0
OS             : darwin-arm64
Native Target  : aarch64-macos
pnpm           : 9.12.1

nx                     : 20.5.0
@nx/js                 : 20.5.0
@nx/jest               : 20.5.0
@nx/eslint             : 20.5.0
@nx/workspace          : 20.5.0
@nx/cypress            : 20.5.0
@nx/devkit             : 20.5.0
@nx/esbuild            : 20.5.0
@nx/eslint-plugin      : 20.5.0
@nx/module-federation  : 20.5.0
@nx/next               : 20.5.0
@nx/node               : 20.5.0
@nx/playwright         : 20.5.0
@nx/react              : 20.5.0
@nx/rspack             : 20.5.0
@nx/storybook          : 20.5.0
@nx/vite               : 20.5.0
@nx/web                : 20.5.0
@nx/webpack            : 20.5.0
typescript             : 5.8.2
```

### Failure Logs

```shell

```

### Operating System

- [x] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

thdk:
@h3adache 

This is not meant to be used like that:

<img width="1173" alt="Image" src="https://github.com/user-attachments/assets/a01b5b0d-e935-423a-9c7f-3fd257fa19e4" />

[...]

Issue #29602, opened 2025-01-13
Nx release - workspace protocol versions are not resolved when publishing independently from custom dist folder


xalechez:
### Current Behavior

When using:
* Independent releases
* pnpm workspaces
* `workspace:` protocol (or `file:` protocol)
* Custom dist directory with git-tag as source of truth for versioning

and releasing a subset of packages e.g. when only a few were affected by the changes

* Versions of workspace dependencies (meaning resolved via pnpm `workspace:` protocol when one workspace package depends on another) unaffected by the current release (e.g. if there were no changes to these packages since previous run) are unexpectedly preserved as `workspace:^` resulting in empty `pnpm error` during `nx release publish` command
* Unaffected packages that are not expected to be published are unexpectedly published with their fallback version set in their package.json (0.0.0 in our case)

We didn't have these versions with npm workspaces. They began to appear since we migrated to pnpm. We use Scenario 2 from [this page](https://nx.dev/recipes/nx-release/updating-version-references#scenario-2-i-want-to-publish-from-a-custom-dist-directory-and-not-update-references-in-my-source-packagejson-files) and don't set `preserveLocalDependencyProtocols`

### Expected Behavior

* Versions of workspace dependencies need to be resolved from git tags and package.json files in custom dist directory should be updated with the latest versions before publishing when `preserveLocalDependencyProtocols` is either not set or set to false
* Packages that are not affected by this release should be skipped during `nx release publish`

### Steps to Reproduce

1. Create a pnpm workspace with packages depending on one another and following Nx release config:

```
  "release": {
    "projects": ["packages/*"],
    "projectsRelationship": "independent",
    "changelog": {
      "projectChangelogs": {
        "file": false,
        "createRelease": {
          "provider": "github-enterprise-server",
          "hostname": "**"
        }
      }
    },
    "version": {
      "preVersionCommand": "pnpm nx run-many -t build",
      "generatorOptions": {
        "updateDependents": "never",
        "skipLockFileUpdate": true,
        "specifierSource": "conventional-commits",
        "currentVersionResolver": "git-tag",
        "packageRoot": "dist/packages/{projectName}"
      }
    }
  },

```
2. Run initial release of all packages
3. Commit a change affecting a "leaf" package (meaning package depending on other packages, but not being dependent on by other packages)
4. Run `nx release version`

### Nx Report

```shell
Node           : 18.20.4
OS             : darwin-arm64
Native Target  : aarch64-macos
pnpm           : 9.15.0

[...]

Issue #28695, opened 2024-10-30
nx release version incorrectly resolves dependency version with pnpm workspaces


xalechez:
### Current Behavior

Info about the workspace:
* 17 npm packages. Independent versioning via git tags (using generic names as they're private)
* Recently migrated from Lerna
* Using pnpm workspaces and workspace protocol
* nx.json (fragment):
```
  "targetDefaults": {
    "nx-release-publish": {
      "options": {
        "packageRoot": "dist/packages/{projectName}"
      }
    }
  },
  "release": {
    "projects": ["packages/*"],
    "projectsRelationship": "independent",
    "changelog": {
      "projectChangelogs": {
        "file": false,
        "createRelease": {
          "provider": "github-enterprise-server",
          "hostname": "our.ghe.server"
        }
      }
    },
    "version": {
      "preVersionCommand": "pnpm nx run-many -t experimental-build",
      "generatorOptions": {
        "updateDependents": "auto",
        "specifierSource": "conventional-commits",
        "currentVersionResolver": "git-tag",
        "packageRoot": "dist/packages/{projectName}"
      }
    }
  },

```

When running `nx release version --dry-run --specifier patch`

* **Package A** is versioned correctly from git tag and provided specifier

```
 NX   Running release version for project: package-a

package-a 🔍 Reading data for package "package-a" from dist/packages/package-a/package.json
package-a 📄 Resolved the current version as 3.0.1 from git tag "package-a@3.0.1".
package-a 📄 Using the provided version specifier "patch".
package-a ✍️  New version 3.0.2 written to dist/packages/package-a/package.json
```

* All other packages are versioned correctly as well
* `package.json` for Package A is updated correctly

```
UPDATE dist/packages/package-a/package.json [dry-run]

    "name": "package-a",
-   "version": "0.0.0",
+   "version": "3.0.2",
    "description": "package-a",

      "json-cycle": "^1.3.0",
-     "package-b": "workspace:*",
-     "package-c": "workspace:*",
+     "package-b": "7.0.13",
+     "package-c": "4.0.18",
      "uuid": "^9.0.1"
```

* `package.json` for Package D is updated incorrectly

```
UPDATE dist/packages/package-d/package.json [dry-run]

    "name": "package-d",
-   "version": "0.0.0",
+   "version": "2.2.3",
    "description": "package-d ",

      "md5": "^2.2.1",
-     "package-b": "workspace:*",
-     "package-a": "workspace:*",
-     "package-c": "workspace:*",
[...]

PR #36334, opened 2026-07-14
fix(release): resolve out-of-set local dependency versions

meeroslav:
## Current Behavior

When Nx Release versions a subset of projects, a released package can depend on a local project outside that release set. With `preserveLocalDependencyProtocols: false`, a `workspace:` or `file:` reference in a publish manifest can remain unresolved because release core does not supply a version for that dependency. Package-manager publishing can then use a placeholder version such as `0.0.0`.

Dependency manifests can also be updated before every release group has made its final version decision. This can write a stale current version for a dependency that is bumped later in the same release.

## Expected Behavior

Release core provides version actions with a lazy dependency version resolver. It returns:

- the new version for a dependency that is being versioned;
- the current version for a dependency outside the release set; or
- the current version for an in-set dependency only after its release group has finalized without a bump.

If a safe version is not yet available, the resolver fails instead of writing a stale value. Dependency manifest updates wait until all release groups have finalized their version decisions.

The JavaScript version actions use this resolver to replace local protocols when `preserveLocalDependencyProtocols` is `false`. JavaScript-specific protocol handling remains outside release core. With `versionPrefix: "auto"`, workspace range intent is retained:

- `workspace:^` becomes `^x.y.z`;
- `workspace:~` becomes `~x.y.z`; and
- `workspace:*` becomes `x.y.z`.

### Interactive prompt behavior

Interactive runs now collect all release-group version prompts before Nx prints the buffered results for those groups. Previously, Nx printed one group's results before prompting for the next group. This change keeps output grouped while dependency manifests wait for final versions. A final logger flush also keeps late version-resolution messages visible.

Unit and end-to-end tests cover excluded dependencies, in-set final versions, no-bump groups, unsafe early resolution, workspace ranges, deferred manifest updates, and late resolution logs. The affected E2E snapshots now count actual manifest edits and omit dependency-update messages when matching ranges are preserved. The full required CI run is green.

## Related Issue(s)

Related: #28695

Related: #29602

Related: #30670


packages/nx/src/command-line/release/version/release-group-processor.ts
JamesHenry, line 1: This hardcodes JS ecosystem specific into the language agnostic core of nx release.

`VersionActions` already exposes an `isLocalDependencyProtocol` method for this reason, please rework

JamesHenry (dismissed):

packages/nx/src/command-line/release/version/release-group-processor.ts
meeroslav: This should be now resolved. Can you re-review?

leosvelperez (approved):

PR #36337, opened 2026-07-14
fix(core): use --config.frozen-lockfile=false for pnpm add during migrate

Squixx:
## Current Behavior

When running `nx migrate` in a pnpm workspace with `frozenLockfile: true` in `pnpm-workspace.yaml`, migration metadata fetching can fail during the install fallback with:

```
Failed to fetch migrations for @angular/cli@22.0.6
Command failed: pnpm add -w @angular/cli@22.0.6
[ERR_PNPM_LOCKFILE_CONFIG_MISMATCH] Cannot proceed with the frozen installation.
```

Nx already passes `--no-frozen-lockfile` for pnpm `install` commands (including post-migration workspace installs), but **not** for `pnpm add` commands used when fetching migrations in temporary directories. Because `createTempNpmDirectory()` copies `pnpm-workspace.yaml` into the temp dir, `frozenLockfile: true` is inherited and blocks `pnpm add`.

## Expected Behavior

Nx-initiated `pnpm add` / `pnpm add -D` commands should include `--no-frozen-lockfile`, matching the existing pnpm `install` behavior. This allows `nx migrate` to fetch package migrations in workspaces that enforce frozen lockfiles for developer consistency.

## Related Issue(s)

No existing upstream issue was found for this specific migrate + `frozenLockfile` failure.

## Changes

- Add `--no-frozen-lockfile` to pnpm `add` and `addDev` command templates in `getPackageManagerCommand()`
- Add regression tests for workspace and non-workspace pnpm add commands
- Update temp install test expectations in `package-json.spec.ts`

## Test plan

- [x] `jest packages/nx/src/utils/package-manager.spec.ts packages/nx/src/utils/package-json.spec.ts --config packages/nx/jest.config.cts --testNamePattern="getPackageManagerCommand|installPackageToTmp"` (10 tests passed)

### DIFF
--- a/packages/nx/src/utils/package-json.spec.ts
+++ b/packages/nx/src/utils/package-json.spec.ts
@@ -88,7 +88,7 @@ describe('installPackageToTmp', () => {
       .spyOn(pacakgeManager, 'getPackageManagerVersion')
       .mockReturnValue('9.0.0');
     jest.spyOn(pacakgeManager, 'getPackageManagerCommand').mockReturnValue({
-      addDev: 'pnpm add -Dw --no-frozen-lockfile',
+      addDev: 'pnpm add -Dw --config.frozen-lockfile=false',
       ignoreScriptsFlag: '--ignore-scripts',
     } as any);
     const execSyncSpy = jest
@@ -99,7 +99,7 @@ describe('installPackageToTmp', () => {
 
     expect(execSyncSpy).toHaveBeenCalledTimes(1);
     expect(execSyncSpy.mock.calls[0][0]).toBe(
-      'pnpm add -Dw --no-frozen-lockfile nx@latest --config.auto-install-peers=false --ignore-scripts'
+      'pnpm add -Dw --config.frozen-lockfile=false nx@latest --config.auto-install-peers=false --ignore-scripts'
     );
 
     cleanup();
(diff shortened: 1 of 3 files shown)


meeroslav (approved):

meeroslav:
The ` --no-frozen-lockfile` option isn't valid for `pnpm add` (it belongs to `pnpm install`), so pnpm rejects the whole command:
```
[ERROR] Unknown option: 'frozen-lockfile'
Did you mean 'trust-lockfile'? Use "--config.unknown=value" to force an unknown option.
```

The correct format should be adding `--config.frozen-lockfile=false`.

Squixx:
## Local validation

Validated this locally without relying on a new pnpm integration test in the Nx repo.

### Reproduced the original failure

Set up a pnpm workspace with `frozenLockfile: true` in `pnpm-workspace.yaml` and a lockfile/config mismatch (lockfile still containing `overrides` after removing them from the yaml). Running the migrate-style command without the fix fails with the same error reported in the issue:

```
pnpm add -w @angular/cli@22.0.6
→ ERR_PNPM_LOCKFILE_CONFIG_MISMATCH
```

### Confirmed the fix

With the updated Nx command template, the install succeeds:

```
pnpm add -w --config.frozen-lockfile=false @angular/cli@22.0.6
```

Also exercised the real migrate install-fallback path from Nx source (`createTempNpmDirectory` + `createFetcher` with `NX_MIGRATE_USE_REGISTRY_RESOLUTION=false`) and confirmed migrations for `@angular/cli@22.0.6` can be fetched successfully when the workspace enforces `frozenLockfile: true`.

### On test scope

I don't think adding a full pnpm CLI integration test in Nx is worth it here. Most of the cost would be testing pnpm's behavior (network, pnpm version differences, brittle stderr matching) rather than Nx logic. Nx's responsibility is generating the correct command string, which is already covered by the existing `getPackageManagerCommand` unit tests plus the small `modifyPnpmWorkspaceYamlToFitNewDirectory` case documenting that `frozenLockfile: true` is preserved when config is copied into migrate temp dirs.

meeroslav (approved):

PR #36263, opened 2026-07-08
feat(core): add a full-width TUI status bar and vim-style pane search

FrozenPandaz:
## Current Behavior

The TUI task list renders its own bottom rows (keyboard hints, Nx Cloud message, filter display) inside its own column, so they are cramped in split layouts and disappear entirely when the task list is hidden (fullscreen pane). The run title and NX badge live in the task-list table header, terminal panes draw their own keybinding hints on their bottom borders, and there is no way to search a pane's output. Much of the UI state (cloud message/link, filter text, perf-report flag) is duplicated between `TuiState` and `TasksList`, kept in sync via broadcast actions.

## Expected Behavior

**Full-width status bar** on the bottom row of the TUI:

- Left: minimal progress counts with a live overall run duration — `63/174 (1m 23s)` — which double as the clickable Nx Cloud link when a structured link exists.
- Middle: free-text cloud messages (they can carry errors), transient pane feedback ("copied to clipboard"), or the compact confirmed-search display.
- Right: context-aware keyboard hints (task-list vs focused-pane) with progressive fitting — as many whole hint items as fit the space — and the `NON-INTERACTIVE i to toggle` / `INTERACTIVE <ctrl>+z to toggle` indicator pinned right-most, never dropped.
- The task-list filter (`/`) swaps the bar row vim-style while typing; the bar is mouse-selectable (drag to highlight + copy) and always visible, including fullscreen-pane mode.
- The ` NX ` badge (run-state colored) and the run title stay at the top-left of the task list in a minimal form; both columns keep bottom-aligned scrollbars.

**Vim-style pane search**: `/` in a non-interactive pane searches the full scrollback (case-insensitive, wrap-aware) with incremental jumping while typing; Enter confirms into `n`/`N` navigation with wrap-around; Esc cancels/clears. Matches highlight reverse-video with the current match on a warning-colored background, and the bar shows `/query 2/5 (n/N)` while a confirmed search is active.

**State consolidation (started)**: `TuiState` is now the single owner of the cloud message/link, filter text, and perf-report flag — the `TasksList` mirrors and the `UpdateCloudMessage`/`UpdateCloudLink` actions are deleted, and filter persistence across TUI mode switches is automatic. Remaining mirrors (task statuses/timings, focus, pinned tasks) are named follow-ups.

## Related Issue(s)

[NXC-4610](https://linear.app/nxdev/issue/NXC-4610/tui-full-width-status-bar-and-vim-style-terminal-pane-search)

---
[...]

PR #36341, opened 2026-07-14
fix(core): report tasks running in another Nx process in the inline TUI

AgentEnder:
## Current Behavior

When a task is being run by a *different* Nx process, this process never gets a pty for it — there is no output to stream and nothing to interact with.

The full-screen terminal pane already handles this: it renders `Running in another Nx process...` for a task whose status is `Shared`/`Stopped` with no pty.

The inline TUI does not. It falls back to `Waiting for tasks to start...` for *any* missing pty, without asking why the pty is missing, so:

- A task running in another Nx process shows `Waiting for tasks to start...` indefinitely — output that will never arrive.
- The user can still enter inline mode for such a task (F11, Enter on a focused pane, double-click a pane, Enter/F11 from the run report), landing in a view that can never render anything.
- A task already displayed inline that transitions from pending (dependency view) to running-elsewhere keeps showing the same misleading message.

## Expected Behavior

The two views agree, and inline mode is never a dead end:

- **Entering inline mode is blocked** for a task another Nx process is running. A hint (`This task is running in another Nx process`) is shown instead of switching, so the user stays in full-screen where the pane explains what is happening. All four entry points route through a single `App::request_inline_mode`.
- **The inline no-pty fallback is status-aware.** A task that is (or becomes) running-elsewhere renders `Running in another Nx process...`, covering the case where the user was *already* in the inline TUI when the task transitioned. An in-progress task with no pty renders `Waiting for task results...`, matching the full-screen pane's wording.
- A new `TuiState::is_running_in_another_process` (status is `Shared`/`Stopped` **and** no local pty) gives the inline app a single named definition of "running elsewhere" that matches what the full-screen pane checks. The pane still reads its own `TerminalPaneState` copies rather than calling the helper (it works off flattened props, not `TuiState`), so the two agree today but are not yet structurally coupled — unifying the pane on the helper is a reasonable follow-up.

Note: the guard applies to a task selected in the task list as well as one pinned to a focused pane — inline always renders exactly one item, and that item would have nothing to show.

Known limitation (inherited, follow-up): because `Shared` and `Stopped` are lumped together, a *shared* continuous task that has finished (goes `Stopped`, never had a local pty) keeps rendering `Running in another Nx process...`. The full-screen pane already behaves this way, so this change inherits rather than introduces it.

### Tests

- `test_inline_mode_blocked_for_task_running_in_another_process` — a shared task shows a hint and does not switch.
- `test_inline_mode_allowed_for_local_task` — a locally running task still drops into inline.
- `test_inline_reports_task_running_in_another_process` — renders the inline view across the `NotStarted → Shared → Stopped` transition.

All 320 TUI tests pass; `cargo fmt --check` and `cargo clippy` are clean. End-to-end validation against a real second Nx process holding a shared task has not been done — behavior is covered by unit tests.

## Related Issue(s)

Relates to [NXC-4597](https://linear.app/nxdev/issue/NXC-4597/error-insert-before-failed-when-swapping-to-inline-mode)

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Fix-inline-TUI-Waiting-for-tasks-state-for-multi-process-scenarios-7a1fd4ea)

### DIFF
--- a/packages/nx/src/native/tui/app.rs
+++ b/packages/nx/src/native/tui/app.rs
@@ -991,7 +991,7 @@ impl App {
                 if matches!(key.code, KeyCode::F(11))
                     && !matches!(self.focus(), Focus::CountdownPopup)
                 {
-                    self.dispatch_action(Action::SwitchMode(TuiMode::Inline));
+                    self.request_inline_mode();
                     return Ok(false);
                 }
 
@@ -1123,7 +1123,7 @@ impl App {
                                 countdown_popup.cancel_countdown();
                                 self.core.state().lock().cancel_quit();
                                 self.close_popup(Focus::CountdownPopup);
-                                self.dispatch_action(Action::SwitchMode(TuiMode::Inline));
+                                self.request_inline_mode();
                                 return Ok(false);
                             }
                             _ => {
@@ -1303,8 +1303,7 @@ impl App {
                                     // If focused on a specific terminal pane, and not interactive, enter should
                                     // swap to inline tui mode focusing that task
                                     KeyCode::Enter => {
-                                        // dispatch action to switch to inline mode with focused task
-                                        self.dispatch_action(Action::SwitchMode(TuiMode::Inline));
+                                        self.request_inline_mode();
                                     }
                                     _ => {
                                         // Forward other keys for interactivity, scrolling (j/k) etc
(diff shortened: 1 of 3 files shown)


leosvelperez (approved):

AgentEnder:
Rebased onto latest master. The `build_status_bar_props` conflict from #36263 was purely textual (both PRs appended a method right after `dispatch_action`); kept both. The full status_bar suite plus the three new tests pass on the rebased tree.

Addressed the two non-blocking notes:

- **Rustdoc on `is_running_in_another_process`** — reworded to your suggestion. It now describes the predicate (`Shared`/`Stopped` **with no local pty**) and calls out that the pty check is the load-bearing conjunct separating a task stopped elsewhere from one stopped locally, which does have a pty. You're right that `Stopped` is terminal and set locally on the fulfilled path in `task-orchestrator.ts` — the old present-tense 'is being run by' wording didn't fit it.
- **'cannot drift apart'** — softened in the description. The pane reads its own `TerminalPaneState` (flattened props, not `TuiState`), so it can't call the helper without a larger refactor; the description now says the two agree today but aren't structurally coupled, and flags unifying the pane as a follow-up.

Also captured the finished-shared-task rendering as a known-limitation note in the description for the follow-up.

One clippy nit surfaced by the rebase (`collapsible_if` in `request_inline_mode`) is fixed with a let-chain, matching the edition-2024 idiom already used elsewhere in `app.rs`.

FrozenPandaz (approved):

PR #36353, opened 2026-07-15
fix(misc): resolve CSS url() assets on Windows in postcss-cli-resources

### LINKED ISSUE #36336: postcss-cli-resources passes pathToFileURL().pathname to resolver (fails on Windows, esp. paths with spaces)
## Current Behavior

In `@nx/rspack` (also webpack / angular-rspack copies), `postcss-cli-resources` does:

```ts
const resolvedPath = path.resolve(context, filePath.replace(/\\/g, '/'));
const { pathname } = pathToFileURL(resolvedPath);
const result = await resolve(pathname, context, resolver);
```

On Windows, `pathToFileURL(...).pathname` is `/C:/Users/.../file.svg` (and spaces become `%20` before `decodeURI`). Passing that string to `loader.resolve` / Rspack's resolver fails with:

```
RspackResolver(NotFound("/C:/Users/.../Island MC Prototype/.../dragged-dnd-handle.svg"))
```

even when the file exists on disk. Reproduced with `@nx/rspack@23.0.0` and confirmed the same pattern still exists in `@nx/rspack@23.1.0` and on `master`.

This commonly breaks CSS `url(...)` for local SVGs and packages like `monaco-editor` `codicon.ttf` when the workspace path contains spaces.

## Expected Behavior

CSS `url()` assets should resolve on Windows regardless of drive letter / spaces. Prefer resolving a filesystem path or a path relative to `context`, e.g.:

```ts
const resolveRequest = path.relative(context, resolvedPath).replace(/\\/g, '/');
const result = await resolve(resolveRequest, context, resolver);
```

or `fileURLToPath(pathToFileURL(resolvedPath))` instead of using `pathname`.

## Steps to Reproduce

1. Clone an Nx rspack React app into a Windows path that contains spaces (e.g. `C:\\Users\\me\\My Project\\app`).
2. Add a CSS file with `background: url('./icon.svg')`.
3. `nx serve` / build with `@nx/rspack`.
4. Observe `postcss-loader` / `RspackResolver(NotFound("/C:/..."))` for the existing SVG.

## Related

- Related (different failure mode, already fixed): #33052 / #34676 — that PR correctly started using `pathToFileURL` for the `new URL(windowsPath, 'file:///')` drive-letter bug, but still feeds `pathname` (`/C:/...`) into the webpack/rspack resolver.
- Related: #32582, #34092

## Environment

- OS: Windows 11
- `@nx/rspack`: 23.0.0 (also verified source still present in 23.1.0 / master)
- Package manager: pnpm 11.9.0
- Node: 24

## Workaround

pnpm-patch `postcss-cli-resources` to pass `path.relative(context, resolvedPath)` to `resolve()` instead of `pathname`.

leosvelperez:
## Current Behavior

On Windows, CSS `url(...)` references to local assets (SVGs, fonts such as `codicon.ttf`, etc.) fail to resolve during an `@nx/rspack`, `@nx/webpack`, or `@nx/angular-rspack` build, even when the file exists on disk. The build reports `RspackResolver(NotFound("/C:/.../file.svg"))`. It is most visible when the workspace path contains spaces.

## Expected Behavior

CSS `url()` assets resolve on Windows regardless of the drive letter or spaces in the path.

## Related Issue(s)

Fixes #36336

## Implementation Details

The `postcss-cli-resources` plugin resolved assets by passing `pathToFileURL(resolvedPath).pathname` to the bundler resolver. That value is a URL path (`/C:/Users/...`, with spaces percent-encoded), not a filesystem path, so the resolver could not find the file. On POSIX the URL pathname coincides with the filesystem path, which is why only Windows was affected. #34676 fixed an earlier `new URL(winPath, 'file:///')` drive-letter misparse but kept feeding the pathname into the resolver.

The fix resolves a relative filesystem path derived from `resolvedPath` instead, and drops the paired `decodeURI` that only existed to undo the pathname encoding. It is applied to the `@nx/rspack`, `@nx/webpack`, and `@nx/angular-rspack` copies of the plugin.

The plugin has no existing unit coverage and the resolver needs a real webpack/rspack loader context, so no automated test was added; the change was validated with build and lint and cross-checked against upstream `@angular-devkit/build-angular`.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-36336-b7b11fff)

### DIFF
--- a/packages/webpack/src/utils/webpack/plugins/postcss-cli-resources.ts
+++ b/packages/webpack/src/utils/webpack/plugins/postcss-cli-resources.ts
@@ -1,6 +1,5 @@
 import { interpolateName } from 'loader-utils';
 import * as path from 'path';
-import { pathToFileURL } from 'url';
 import type { Declaration } from 'postcss';
 import type { LoaderContext } from 'webpack';
 
@@ -107,7 +106,12 @@ export function PostcssCliResources(options: PostcssCliResourcesOptions) {
     // Separate URL query/hash from the file path before resolving
     const [, filePath, urlSuffix] = inputUrl.match(/^([^?#]*)(.*)$/)!;
     const resolvedPath = path.resolve(context, filePath.replace(/\\/g, '/'));
-    const { pathname } = pathToFileURL(resolvedPath);
+    // Resolve a relative filesystem path, not a file:// URL pathname: on Windows the
+    // pathname is `/C:/...` (spaces as %20), which the resolver cannot find. Relative
+    // (not absolute) lets `resolve()`'s `./`-prefixed lookup succeed on the first try.
+    const resolveRequest = path
+      .relative(context, resolvedPath)
+      .replace(/\\/g, '/');
     let hash = '';
     let search = '';
     if (urlSuffix) {
@@ -115,7 +119,7 @@ export function PostcssCliResources(options: PostcssCliResourcesOptions) {
     }
     const resolver = (file: string, base: string) =>
       new Promise<boolean | string>((resolve, reject) => {
-        loader.resolve(base, decodeURI(file), (err, result) => {
+        loader.resolve(base, file, (err, result) => {
           if (err) {
             reject(err);
             return;
@@ -123,7 +127,7 @@ export function PostcssCliResources(options: PostcssCliResourcesOptions) {
           resolve(result);
         });
       });
-    const result = await resolve(pathname as string, context, resolver);
+    const result = await resolve(resolveRequest, context, resolver);
     return new Promise<boolean | string>((resolve, reject) => {
       loader.fs.readFile(result as string, (err: Error, content: Buffer) => {
         if (err) {
--- a/packages/angular-rspack/src/lib/utils/postcss-cli-resources.ts
+++ b/packages/angular-rspack/src/lib/utils/postcss-cli-resources.ts
@@ -8,7 +8,6 @@
 
 import { interpolateName } from 'loader-utils';
 import * as path from 'node:path';
-import { pathToFileURL } from 'node:url';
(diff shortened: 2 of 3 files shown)


packages/webpack/src/utils/webpack/plugins/postcss-cli-resources.ts
```diff
@@ -107,23 +106,28 @@ export function PostcssCliResources(options: PostcssCliResourcesOptions) {
     // Separate URL query/hash from the file path before resolving
     const [, filePath, urlSuffix] = inputUrl.match(/^([^?#]*)(.*)$/)!;
     const resolvedPath = path.resolve(context, filePath.replace(/\\/g, '/'));
-    const { pathname } = pathToFileURL(resolvedPath);
+    // Resolve a relative filesystem path, not a file:// URL pathname: on Windows the
+    // pathname is `/C:/...` (spaces as %20), which the resolver cannot find. Relative
+    // (not absolute) lets `resolve()`'s `./`-prefixed lookup succeed on the first try.
+    const resolveRequest = path
```
meeroslav, line 112: Should we extract this to a function in devkit perhaps?

meeroslav (approved):

packages/webpack/src/utils/webpack/plugins/postcss-cli-resources.ts
leosvelperez: We already have some `normalizePath` with similar but slightly different logic. Introducing another helper for what seems like the same type of normalization, but with slightly different behavior, can lead to confusion or subtle errors if the wrong one is chosen. In any case, what's really reusable here is the slash-swap normalization, since the target string can differ in other uses outside these three places.

PR #36258, opened 2026-07-07
fix(core): create nx sockets under a stable root for sandbox allowlisting

AgentEnder:
## Current Behavior

Nx creates its unix sockets (daemon, plugin workers, forked task processes) in per-process hashed directories directly under `os.tmpdir()`. The paths change on every run and differ per machine/user (macOS `$TMPDIR` is per-user, and AI-agent sandboxes rewrite `$TMPDIR` per session), so sandboxed environments cannot allowlist them — the only workable Claude Code setting is the blanket `allowAllUnixSockets: true`. When a sandbox blocks socket access, the daemon client surfaces generic `connect EPERM`/timeout errors with no guidance.

## Expected Behavior

- All Nx sockets live under one stable root: `/tmp/.nx/sockets` on macOS/Linux (`.nx/sockets` under the OS temp dir on Windows, where named pipes make the path a name token only). `NX_SOCKET_DIR`/`NX_DAEMON_SOCKET_DIR` overrides are unchanged, and the shared root gets a best-effort sticky `0o1777` chmod for multi-user machines.
- `connect EPERM`/`EACCES` and daemon start-timeout errors explain the sandbox cause and name the socket root. Messages are tuned via `isAiAgent()`: agents get actionable self-serve steps (run `nx configure-ai-agents`, allowlist `/tmp/.nx/sockets`, or set `NX_SOCKET_DIR`); humans get a brief neutral pointer, and the start-timeout hint is agent-only since that failure has many other causes for humans.
- The `setup-ai-agents` generator writes machine-independent sandbox allowances into the committed `.claude/settings.json`: `sandbox.network.allowUnixSockets: ["/tmp/.nx/sockets"]` and `sandbox.filesystem.allowRead/allowWrite: ["/tmp/.nx"]`, preserving and deduping existing user entries.
- Docs: the Claude Code sandbox troubleshooting page covers the scoped allowlist (team-shareable committed config), corrects the `allowAllUnixSockets` example to live under `sandbox.network`, and notes that `allowUnixSockets` covers `connect()` only; `NX_SOCKET_DIR` reference rows state the new default.

## Related Issue(s)

No linked issue — follow-up to internal work on running Nx under AI-agent sandboxes.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Unix-Socket-Paths--Daemon-Connection-Error-Messaging-9e6b083e)

PR #36370, opened 2026-07-15
fix(core): restrict daemon and plugin worker socket access to the owning user

AgentEnder:
## Current Behavior

The Nx daemon and isolated plugin workers communicate over unix domain sockets (named pipes on Windows).

On POSIX the socket directory and files are created under the ambient umask with no explicit mode, and plugin worker sockets sit directly in the shared system temp root. Socket paths are also unpredictable: each process derives its own hashed directory under `os.tmpdir()`, so there is no single path a sandbox or container policy can allowlist.

The native binding is copied into a file cache before being loaded. That cache lives in a hash-named directory under the system temp root, and the loader accepts an existing entry on byte size alone, so a rebuild that happens to produce an identical size is not picked up.

Separately, a daemon error is tagged as internal only when its log file can be read, so a daemon that fails *before* ever writing a log aborts the command instead of falling back to a daemonless graph build.

### Layout on macOS and Linux

Every runtime artifact moves under one fixed root, with a per-user directory beneath it. `<tmp>` is `os.tmpdir()` — `/var/folders/…` on macOS, `/tmp` on Linux:

| Today | After |
| --- | --- |
| `<tmp>/<hash20>/d.sock` | `/tmp/.nx/<uid>/sockets/<hash20>/d.sock` |
| `<tmp>/<hash20>/fp<pid>-<n>.sock` | `/tmp/.nx/<uid>/sockets/<hash20>/fp<pid>-<n>.sock` |
| `<tmp>/plugin<pid>-<n>-<perfNow>.sock` | `/tmp/.nx/<uid>/sockets/<hash8>/p<pid>-<n36>-<rand8>.sock` |
| `<tmp>/nx-native-file-cache-<hash7>/<file>` | `/tmp/.nx/<uid>/native-cache/<nxVersion>/<file>` |

The root is a literal `/tmp/.nx` rather than `os.tmpdir()`, which honors `$TMPDIR` — per-user on macOS, rewritten by sandboxes, and stripped from the daemon's environment. A fixed path is identical on every machine, so a sandbox allowlist entry for it can be committed and shared with a team.

**The uid sits directly beneath the container**, so `/tmp/.nx` is the only level that can ever be shared between users, and `sockets` and `native-cache` are inside the per-user directory rather than beside it. That is what makes a single `install -d -m 1777 -o root -g root /tmp/.nx` sufficient provisioning: there is no `/tmp/.nx/sockets` for the first user to create and own.

When the container cannot be used, sockets move to `~/.nx/sockets` before falling back into the workspace. Home needs no administrator — there is no shared level to own — so a peer holding `/tmp/.nx` no longer costs the short, fixed-length socket path.

**The native cache deliberately gets no second tier.** Sockets need *somewhere* to live or the daemon and plugin workers cannot run at all, which is what justifies chaining locations for them. The cache is an optimisation with a working fallback already in hand — load the binding in place from `node_modules` — so a second location would add a directory to guard and a `.node` to trust for no capability that is otherwise lost. Skipping it is also the fail-closed answer: the only thing worse than no cache is a cache another user can write to.

The home tier is skipped when `HOME` makes `~/.nx` the shared container itself (`HOME=/tmp`); otherwise a tier-1 failure would point the owner-only guard at `/tmp/.nx` and take a root-owned `1777` directory to `0700`.

### Layout on Windows

Named pipes are not filesystem objects, so there is nothing to allowlist and no shared directory to separate users in, and `%TMP%` is already per-user. Neither reason for the shared root applies, so sockets stay directly under `%TMP%` and **no socket path is longer than it is today**:

| Today | After |
| --- | --- |
| `%TMP%\<hash20>\d.sock` | `%TMP%\<hash20>\d.sock` (unchanged) |
| `%TMP%\<hash20>\fp<pid>-<n>.sock` | `%TMP%\<hash20>\fp<pid>-<n>.sock` (unchanged) |
| `%TMP%\plugin<pid>-<n>-<perfNow>.sock` | `%TMP%\<hash8>\p<pid>-<n36>-<rand8>.sock` (3 chars shorter) |
| `%TMP%\nx-native-file-cache-<hash7>\<file>` | `%TMP%\.nx\native-cache\<nxVersion>\<file>` |

There is no per-uid segment on Windows — `%TMP%` is already per-user, and the segment would only spend path budget. This matters because `assertValidSocketPath` rejects paths over 95 characters and has no platform guard, so it applies to named pipes too, and `%TMP%` already contains the username. Longest username that still fits, with the default `%TMP%`:

| socket | today | after |
| --- | --- | --- |
| daemon | ≤39 | ≤39 |
| plugin | ≤30 | ≤33 |
| forked / pty | ≤29 | ≤29 |

The native cache is not subject to that limit.

### Shared roots

`/tmp/.nx` is `1777` sticky, like `/tmp` itself, so every user on a machine can create their own directory beneath it. It is the **only** shared level: `/tmp/.nx/<uid>` and everything under it is `0700`, created and re-checked on every use.

Before creating anything beneath a shared root, Nx checks that it is a real directory, that it carries the sticky bit if anyone beyond its owner can write to it, and that it belongs to the current user or to `root`. The ownership half is not redundant: sticky restricts renaming an entry to that entry's owner *and the directory's owner*, so a root belonging to another regular user is unsafe at any mode. `/tmp` itself is safe for the same reason — `root` owns it.

Several users therefore share the container only when something trusted created it first — a container image, or an administrator running `sudo install -d -m 1777 -o root -g root /tmp/.nx`. That one command is the whole remedy, because no descendant is shared. Otherwise everyone after the first moves to `~/.nx`, which needs no provisioning at all; the error names the `chown` that would restore sharing.

On a single-user machine, in a container, and in a sandbox, none of this applies: the first Nx run creates the container and owns it, so `isSafeSharedRoot` accepts it with no setup.

`NX_SOCKET_DIR` still overrides the socket location and is used as given — it names the socket directory itself. Pointing it at one of Nx's own roots throws `InvalidSocketDirConfigured` rather than silently substituting a default, with two distinct messages: the system temp dir and `/tmp/.nx` are reachable by other users, while the per-user roots are refused because Nx manages and cleans what lives beneath them. A directory *nested* under any of them is still accepted.

### Directory handling

- `ensureOwnedPrivateDir` accepts a directory only if it is a real directory (`lstat`, so a symlink does not qualify), owned by the current user, and carries no group or other permissions — read and search are enough to reach a socket inside it, so `0755` is tightened to `0700` rather than accepted. A directory that fails a check it cannot repair is not used.
- Shared roots are created one level at a time, so a symlink at any level is caught rather than resolved through. The verdict comes from the resulting mode rather than from whether the `chmod` succeeded, since only a root's owner can `chmod` it.
- Directory modes are changed through a descriptor opened `O_NOFOLLOW | O_NONBLOCK` and confirmed with `fstat` to be a directory, rather than by classifying errnos — the errno for a given condition is not stable across flag combinations or kernels, and `O_NONBLOCK` avoids a FIFO blocking the open.
- The daemon socket file is set to `0600` after `listen`. On Linux, connecting requires write permission on the socket file; on macOS/BSD the (already `0700`) directory is what applies.
- Anything that cannot be established is skipped rather than approximated: sockets try `~/.nx/sockets` and then a directory inside the workspace, and the native cache is bypassed so the binding loads in place from `node_modules`.

### Message scoping

Every message carries the sender's `workspaceRoot`, stamped centrally by the transport rather than by each message constructor. A receiver scoped to a different workspace refuses it: the daemon returns the mismatch and stays alive for its own workspace, and a plugin worker drops the message. This is what catches two workspaces accidentally sharing an `NX_SOCKET_DIR`.

### Daemon failure classification

A daemon failure is tagged as internal whether or not a log file exists yet, so a daemon that cannot start degrades to a daemonless graph build instead of aborting. Previously the tag was set only inside the block that read the log, so a first run — exactly when startup is most likely to fail — aborted the command.

A `connect EPERM`/`EACCES` is the one exception, and is tagged separately rather than as internal. It degrades the same way, but is deliberately *not* disabled until `nx reset`: a socket owned by someone else stops being there once it is removed or the machine reboots, so a sticky disable would outlive the cause and hide the fix from anyone who followed the advice. The message names the socket, both ways out, and keeps the errno on its first line — `EACCES` and `EPERM` need opposite remedies and it is the only token that tells them apart.

## Scope

Socket paths, permissions, and the native binary cache only.

The AI-agent and sandbox work previously folded in here is split into **#36586**, stacked on this branch (it replaces #36463, which was closed): Codex/Superset agent detection, `nx configure-ai-agents` writing the sandbox allowances, the sandbox remediation hint, and the changes that re-enable the daemon and plugin isolation inside sandboxes. On this branch both still auto-disable under a detected sandbox, exactly as on `master`.

**The Rust side is deliberately not covered, and NXC-4025 is therefore only partly met.** `packages/nx/src/native/utils/socket_path.rs` derives the Nx Console socket independently — `std::env::temp_dir()` plus a workspace hash, created with `create_dir_all` at the ambient umask — and is reached from `native/ide/nx_console/messaging.rs` on every TUI run that talks to Nx Console. It also reads `NX_SOCKET_DIR`/`NX_DAEMON_SOCKET_DIR` and applies none of the refusal list added here, so a value this branch rejects is still used as-is by the Rust path.

That means the ticket's "all Nx sockets under one allowlistable root" is not yet true, and the sandbox allowlist this branch emits (`/tmp/.nx` or `~/.nx`) does not cover the Nx Console socket. It is pre-existing code on a different runtime with its own path handling, so folding it in would widen this change materially; it is tracked as a follow-up instead.

## Behavior changes worth calling out

- **Socket paths moved on macOS and Linux.** Anything that located Nx sockets by scanning the system temp root must look under `/tmp/.nx/<uid>/sockets`, `~/.nx/sockets`, or `NX_SOCKET_DIR` instead. On macOS this is a different filesystem, not just a different name: `$TMPDIR` is `/var/folders/…` while the new root is the literal `/tmp`. Windows socket directories are unchanged.
- **The native cache directory moved on both platforms**, from `<tmp>/nx-native-file-cache-<hash7>` to `/tmp/.nx/<uid>/native-cache/<nxVersion>` (`%TMP%\.nx\native-cache\<nxVersion>` on Windows). Tooling that excluded or cleaned the old path by name needs updating.
- **Previously published `NX_DAEMON_SOCKET_DIR` guidance is reversed.** The daemon docs told users to point it at a shared directory so several long-running containers could share one daemon socket. That is exactly what this change refuses, so the page now says the opposite: never share a socket directory, and give each container its own daemon. Anyone following the old documented setup will see `InvalidSocketDirConfigured` or a refused directory rather than a silent behaviour change — that is intended, but it is a documented workflow being withdrawn, not just a docs edit.
- **The `0700` mode is applied to directories you already own**, not only ones Nx creates — in practice the workspace-local fallback (`<workspaceRoot>/.nx/workspace-data/d`, routinely `0755`) and an explicitly configured `NX_SOCKET_DIR`. This ends the docker-compose pattern of sharing one socket directory across containers running as different users, and today it happens without a message.
- **A container owned by another regular user is no longer used.** On a multi-user machine where someone else ran Nx first, sockets move to `~/.nx/sockets` and the native binding loads in place. Pre-creating `/tmp/.nx` as `root` restores the shared layout for everyone.
- **`NX_NATIVE_FILE_CACHE_DIRECTORY` and `NX_SOCKET_DIR` now behave differently on a bad value.** A socket directory Nx refuses throws `InvalidSocketDirConfigured`, because there is no safe substitute and silently relocating sockets resurfaces later as a path-length error about a directory the user never set. A native cache directory Nx refuses warns and loads the binding in place, because that fallback is complete. Both are loud; only one is fatal.
- **The native cache is keyed per uid and Nx version**, with each file additionally keyed by a hash of the resolved binding path so multiple source checkouts (all reporting `0.0.1`) do not collide. A cache entry older than its source is refreshed.

## Related Issue(s)

Fixes NXC-4658
Fixes NXC-4025

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Tighten-Nx-daemon-RPC-socket-security-0700-0600-perms-1fee7ebf)


astro-docs/src/content/docs/concepts/nx-daemon.mdoc
```diff
@@ -49,7 +49,11 @@ To see information about the running Nx Daemon (such as its background process I
 ## Customizing the socket location
 
 The Nx Daemon uses a unix socket to communicate between the daemon and the Nx processes. By default this socket gets placed in a temp directory. If you are using Nx in a docker-compose environment, however, you may want to run the daemon manually
-and control its location to enable sharing the daemon among your docker containers. To do so, set the `NX_SOCKET_DIR` environment variable to a shared directory.
+and control its location to enable sharing the daemon among your docker containers. To do so, set the `NX_SOCKET_DIR` environment variable to a directory shared by those containers.
+
+{% aside type="caution" title="Keep the socket directory private" %}
+Do not point `NX_SOCKET_DIR` (or `NX_DAEMON_SOCKET_DIR`) at a directory that other users on the machine can access, such as a world-writable temp directory. The daemon drives work inside a long-lived process, so another user who can reach its socket could execute code in your daemon. Share the socket directory only between processes that belong to the same user and the same workspace; the daemon rejects messages whose workspace root does not match its own.
+{% /aside %}
```
FrozenPandaz: Change the primary usecase of this variable to be a workaround for file permissions. Do not recommend the whole docker container sharing daemon thing. Still have a caution that points out that it should not be shared because it is a remote for code execution.

astro-docs/src/content/docs/reference/environment-variables.mdoc
```diff
@@ -234,7 +234,7 @@ The following environment variables are ones that you can set to change the beha
...
+| `NX_SOCKET_DIR`                            | string  | Directory for all Nx sockets (daemon, forked process, and plugin). Set this to a shorter path when the default temp directory produces a socket path that exceeds the OS limit. Do not set it to a location shared with other users or other workspaces: the daemon drives work inside a long-lived process, so a shared socket directory would let another user connect to your daemon. Takes precedence over `NX_DAEMON_SOCKET_DIR`.                                                                                                                                                                                       |
```
FrozenPandaz:
```suggestion
| `NX_SOCKET_DIR`                            | string  | Directory for all Nx sockets (daemon, forked process, and plugin). Be sure to read [Nx Daemon documentation](blah) before using this.                                                                                                                                                                                |
```

packages/nx/src/daemon/message-types/daemon-message.ts
```diff
@@ -1,9 +1,27 @@
...
+ * workspace (e.g. two workspaces sharing an `NX_SOCKET_DIR`) and must not be
+ * processed. The comparison is case-insensitive to match how the socket
+ * directory is keyed off the workspace root.
+ */
+export function isForeignWorkspaceMessage(
+  msg: DaemonMessage,
+  daemonWorkspaceRoot: string
+): boolean {
+  return (
+    msg.workspaceRoot !== undefined &&
```
FrozenPandaz: return false when workspaceRoot is not defined.

packages/nx/src/daemon/message-types/daemon-message.ts
```diff
@@ -1,9 +1,27 @@
...
+    msg.workspaceRoot.toLowerCase() !== daemonWorkspaceRoot.toLowerCase()
```
FrozenPandaz: Don't sanitize this. They should be identical.

packages/nx/src/daemon/server/server.ts
```diff
@@ -256,6 +260,24 @@ async function handleMessage(socket: Socket, data: string) {
   }
   serverLogger.log(`Received ${mode} message of type ${payload.type}`);
 
+  // Refuse messages from a different workspace without processing them any
+  // further. The daemon is scoped to the workspace that launched it; a mismatch
+  // means the client reached the wrong daemon (e.g. a shared NX_SOCKET_DIR). We
+  // respond with an error but keep the daemon alive for its own workspace.
+  if (
+    isDaemonMessage(payload) &&
+    isForeignWorkspaceMessage(payload, workspaceRoot)
```
FrozenPandaz: change this to a try catch { respondWithError } so that the error can utilize the information in the check.

Call it assertValidDaemonMessage

packages/nx/src/daemon/server/server.ts
```diff
@@ -731,6 +753,19 @@ export async function startServer(): Promise<Server> {
...
           serverLogger.log(`Started listening on: ${socketPath}`);
 
+          // Belt-and-suspenders: restrict the socket file itself to its owner so
+          // that only the daemon owner can connect. On Linux, connecting to a
+          // Unix socket requires write permission on the socket file; on
+          // macOS/BSD the (already-0700) socket directory is what gates access.
+          // Skipped on Windows, where `socketPath` is a named pipe.
+          if (!isWindows) {
+            try {
+              chmodSync(socketPath, 0o600);
```
FrozenPandaz, line 764: Consider if we can pass these permissions in where we create the directory / create the socket

packages/nx/src/daemon/tmp-dir.ts
```diff
@@ -54,24 +54,71 @@ function socketDirName() {
...
+}
+
+function createOwnerOnlySocketDir(dir: string): string {
+  try {
+    mkdirSync(dir, { recursive: true, mode: 0o700 });
+    // Restrict the socket directory to its owner so that other local users
+    // cannot connect to the sockets inside it (which would let them execute
+    // code in the daemon or a plugin worker). We never touch the shared system
+    // temp dir, which other processes rely on.
+    if (dir !== tmpdir) {
```
FrozenPandaz: throw when the directory is the tmpdir which we assume can be utilized by any user on the machine and thus unsafe. Then change this logic appropriately.

FrozenPandaz (changes requested):

packages/nx/src/daemon/server/server.ts
AgentEnder: confirmed socket creation can't create files w/ given perms.

packages/nx/src/daemon/tmp-dir.spec.ts
```diff
@@ -54,16 +58,19 @@ describe('socket directories', () => {
...
-  it('never locks down the shared system temp directory', () => {
+  it('refuses the shared system temp directory and falls back to the workspace data dir', () => {
     setPlatform('linux');
-    // A user pointing NX_SOCKET_DIR at the raw system temp dir must not cause us
-    // to chmod it, since other processes rely on it.
+    // Pointing NX_SOCKET_DIR at the raw system temp dir is unsafe: every user on
+    // the machine can access it, so it can never be locked down to us. We fail
+    // closed by falling back to the owner-controlled workspace data dir instead
+    // of trusting the shared root.
     process.env.NX_SOCKET_DIR = tmpdir;
```
AgentEnder: Hmmm, no we should just throw if we detect this instead of fall back to the default value. Its invalid configuration.

packages/nx/src/daemon/tmp-dir.ts
```diff
@@ -93,16 +93,37 @@ export function getPluginSocketDir() {
...
+
+    // `mode` only applies to directories mkdirSync actually creates; a
+    // pre-existing directory keeps its current permissions, so we still chmod
+    // it below to guarantee the sockets inside are reachable only by their
+    // owner.
+    mkdirSync(dir, { recursive: true, mode: 0o700 });
+    restrictToOwner(dir);
     return dir;
   } catch (e) {
+    // Fail closed: fall back to the owner-controlled workspace data dir rather
```
AgentEnder: We should catch the resolve(dir) == resolve(tmpdir), or if we do we should rethrow... We can extract a InvalidSocketDirConfigured error type or similar for control flow.

packages/nx/src/daemon/tmp-dir.ts
```diff
@@ -3,13 +3,29 @@
...
+ * Thrown when the resolved Nx socket directory points at the shared system temp
+ * directory. That location is accessible to every user on the machine, so the
+ * daemon and plugin worker sockets placed inside it could be reached by another
+ * local user and used to execute code. This is invalid configuration and must
+ * fail loudly rather than silently substituting a default.
+ */
+export class InvalidSocketDirConfigured extends Error {
+  constructor(public readonly dir: string) {
+    super(
+      `The configured Nx socket directory ('${dir}') resolves to the shared system temp directory ('${tmpdir}'), which every user on the machine can access. Pointing the Nx socket directory there is unsafe: another local user could connect to the daemon or plugin worker sockets and execute code in them. Set NX_SOCKET_DIR (or NX_DAEMON_SOCKET_DIR) to a directory that only your user can access.`
```
FrozenPandaz:
```suggestion
      `The configured Nx socket directory ${dir} cannot be the shared system temp directory. Pointing the Nx socket directory there is unsafe as another local user could connect to the daemon or plugin worker sockets and execute code in them. Set NX_SOCKET_DIR to a directory that only your user can access.`
```

packages/nx/src/daemon/message-types/daemon-message.ts
```diff
@@ -1,9 +1,43 @@
...
 }
+
+/**
+ * A daemon is scoped to the workspace that launched it. A message whose
+ * `workspaceRoot` differs from the daemon's own root came from a different
+ * workspace (e.g. two workspaces sharing an `NX_SOCKET_DIR`) and must not be
+ * processed. Both values are produced from the same workspace-root resolution,
+ * so they are compared directly.
+ */
+export function isForeignWorkspaceMessage(
```
FrozenPandaz, line 30: Should we do this for plugin workers as well?

packages/nx/src/daemon/message-types/daemon-message.ts
```diff
@@ -1,9 +1,43 @@
 export type DaemonMessage = {
   type: string;
   env?: Record<string, string>;
+  workspaceRoot?: string;
```
FrozenPandaz, line 10: Should we make this required?

FrozenPandaz (changes requested):

packages/nx/src/project-graph/plugins/isolation/plugin-worker.ts
```diff
@@ -79,6 +86,18 @@ const server = createServer((socket) => {
       if (!isPluginWorkerMessage(message)) {
         return;
       }
+      // Reject messages from a different workspace, the same way the daemon
+      // refuses foreign-workspace messages on its socket. A message whose
+      // workspaceRoot differs from this worker's own root came from a process
+      // in another workspace (e.g. one that reached this worker's socket via a
+      // shared NX_SOCKET_DIR); do not process it. An undefined workspaceRoot is
+      // treated as "not foreign", consistent with the daemon.
+      if (isForeignWorkspaceMessage(message, hostWorkspaceRoot)) {
```
FrozenPandaz: how come we don't reuse assertValidDaemonMessage?

FrozenPandaz (changes requested):

packages/nx/src/daemon/tmp-dir.ts
```diff
@@ -46,32 +65,164 @@ export function isDaemonDisabled() {
...
 }
 
+/**
+ * All Nx sockets (daemon, forked processes, plugin workers) live under this
+ * single stable root so that sandboxed environments (e.g. AI agent sandboxes)
+ * can allow unix socket access with one predictable rule. See NX_TMP_DIR for
+ * why a fixed /tmp path is used on POSIX. Windows named pipes are not subject
+ * to filesystem sandboxing, so the OS temp dir is fine there.
+ */
+export const NX_SOCKET_ROOT_POSIX = join(NX_TMP_DIR_POSIX, 'sockets');
```
AgentEnder: Lets get rid of this one.... the posix path would be accurately baked into NX_SOCKET_ROOT so we don't need both vars? NX_TMP_DIR on posix is equivalent to NX_TMP_DIR_POSIX

AgentEnder:
Supersedes #36258 — the stable socket root work (sockets + native file cache under `/tmp/.nx`) was folded into this PR via a6b83b9b10, alongside the daemon/plugin socket access restrictions. Closing #36258 in favor of this combined PR.

packages/nx/src/daemon/tmp-dir.ts
```diff
@@ -46,32 +70,160 @@ export function isDaemonDisabled() {
...
+    // A genuine fs failure (e.g. we lack permission to create the configured
+    // dir), or a leaf we refused above, is recoverable: fall back to the
+    // owner-controlled workspace data dir rather than a shared location. That
+    // path is not world-writable, but verify it the same way so the fallback is
+    // never weaker than what it replaces.
+    try {
+      mkdirSync(dirname(DAEMON_DIR_FOR_CURRENT_WORKSPACE), { recursive: true });
+      ensureOwnedPrivateDir(DAEMON_DIR_FOR_CURRENT_WORKSPACE);
+    } catch {}
     return DAEMON_DIR_FOR_CURRENT_WORKSPACE;
```
AgentEnder: Move the return into the try… we don't want to return an unsafe dir in the error case, it'd be btter to hard error?

packages/nx/src/daemon/server/server.ts
AgentEnder: Confirmed not possible for either half, so this stayed as a post-`listen` `chmod`.

`net.Server.listen()` takes no mode option, and the socket is created by the bind inside it — there is no point between creation and the file existing where a mode could be supplied. `umask` is the only other lever and it is process-global, so narrowing it around the bind would race every other write in the process.

The directory *is* created with an explicit mode (`mkdirSync(dir, { mode: 0o700 })`, non-recursive so the mode is not subject to `recursive:true`'s parent handling), which is the half that can be done at creation time. The `0600` on the socket file is defence-in-depth on top of that: Linux gates `connect()` on write permission to the socket file, macOS/BSD gate on the directory, which is already `0700`. The comment at the call site now says exactly this.

packages/nx/src/daemon/message-types/daemon-message.ts
AgentEnder: Yes — this is shared with the plugin worker now, which is what `assertValidDaemonMessage` was extracted for.

`plugin-worker.ts:103` calls it with its own `receiverDescription`, so the two receivers differ only in what they do with the throw: the daemon catches it and responds with the mismatch, the worker catches it and drops the message. `WorkspaceScopedMessage` is structural precisely so neither side has to import the other's message type.

Worth being explicit about the limit, since it reads stronger than it is: `isForeignWorkspaceMessage` treats an absent `workspaceRoot` as *not* foreign, so anyone who simply omits the field bypasses this. It is a detector for two workspaces accidentally sharing an `NX_SOCKET_DIR`, not a control against an attacker — the `0700` directory and `0600` socket are the control.

packages/nx/src/daemon/message-types/daemon-message.ts
AgentEnder: Left optional deliberately, for two reasons now recorded in the docblock above it.

The field is stamped centrally by `DaemonSocketMessenger.sendMessage` (and `sendMessageOverSocket` on the plugin-worker side) rather than by the ~30 sites that construct messages, so making it required would mean touching all of them to set a value the transport immediately overwrites.

It also has to tolerate absence on the wire: a client from a previous version sends no `workspaceRoot`, and `isForeignWorkspaceMessage` returning `false` for `undefined` is what keeps that working rather than having the daemon reject every message during a version skew.

The type-level fix that would give both — an outgoing type without the field and a stamped type where it is required — is worth doing, but it splits the message types across the daemon and plugin-worker protocols, so I would rather not land it inside this one.

FrozenPandaz (approved):

PR #36376, opened 2026-07-16
fix(angular-rspack): map bundled sourcemaps back to TypeScript sources

gultyayev:
## Current Behavior

When building an Angular application with Rspack (`@nx/angular-rspack`), the generated sourcemaps point at the intermediate **Ivy JavaScript** instead of the original TypeScript. Setting a breakpoint in (or viewing the mapped source of) a `.ts` file shows generated Ivy code rather than the authored TypeScript.

The root cause is two broken links in the sourcemap chain (`ngc` → `JavaScriptTransformer`/linker → Rspack loader → bundle), both of which Angular's own esbuild pipeline handles differently:

1. The Angular compilation emitted an **external** `.js.map` (`sourceMap: true`). The downstream `JavaScriptTransformer` (babel + the Angular linker) only reads *inline* input sourcemaps, so the external map was dropped and the chained map stopped at the Ivy JS.
2. The Angular loaders returned transformed content but never handed the sourcemap to Rspack's loader callback. Unlike esbuild, Rspack does not auto-consume inline `//# sourceMappingURL=data:...` comments embedded in loader output, so `SourceMapDevToolPlugin` mapped the bundle output straight to the Ivy JS text.

## Expected Behavior

Bundled sourcemaps map back to the original TypeScript sources, matching how `@angular/build`'s esbuild pipeline preserves the mapping.

- Emit an intermediate **inline** sourcemap from the Angular compilation instead of an external `.js.map`, so the `JavaScriptTransformer`/linker can read it and chain it back to the TypeScript.
- Extract that inline sourcemap in the Angular loaders and pass it to Rspack's loader callback so Rspack chains it into the final bundle sourcemaps.

### Changes

- `angular-rspack-compiler`: emit inline sourcemaps from the Angular compilation, threaded through a new `sourceMap` compilation option.
  - **`inlineSourceMap` stays enabled even when sourcemaps are disabled** (only `inlineSources` is gated on the user's setting): with all sourcemap compiler options off, `AotCompilation.emitAffectedFiles` skips TypeScript transpilation for `isolatedModules` projects and emits raw TypeScript (expecting esbuild to strip types downstream), which this babel-based pipeline cannot parse — production builds of generated apps failed with a syntax error on type annotations. Keeping the inline map on matches the cost profile of the previous unconditional `sourceMap: true`. (Caught by the e2e test below.)
- `angular-rspack`: new `inline-source-map.ts` helper that extracts and strips a trailing base64 inline sourcemap comment; both `angular-transform.loader` and `angular-partial-transform.loader` forward the extracted map to Rspack.
  - Extraction is **memoized per cache key** (`===` guard on the cached value: same-instance reads short-circuit on reference equality; strings compare by value, `Uint8Array` by reference) so watch-mode rebuilds don't re-run the regex + base64 decode + JSON validation on every read of unchanged entries. The map is handed to Rspack as the decoded **JSON string** (accepted natively by the loader callback), avoiding typed-object casts at the loader boundary — Rspack parses the string itself.
  - The embedded payload is **validated as a raw v3 sourcemap, including nested field types**, before being forwarded: Rspack's deserializer fails the module build not just on non-map JSON values (`{}`, `null`, arrays) but also on wrongly typed fields (e.g. `sources: [42]`, invalid `names`/`sourcesContent`/`file`/`sourceRoot`/`debugId`/`ignoreList` → `bad json: ExpectedString`), which the partial-transform loader could otherwise pick up from arbitrary Angular-related JavaScript. Every field rspack reads is type-checked (array elements included; optional fields may be `null`, matching the deserializer); anything else passes the code through untouched with no map.

### Testing

- Unit tests for the extractor (incl. CRLF handling, trailing newlines, malformed and non-sourcemap payloads such as `{}`/`null`/arrays, wrongly typed nested fields per rejected field, null-optional acceptance, memoization), both loaders' sourcemap forwarding, and the compiler-option plumbing in `setup-compilation`.
- Extended the `should successfully work with rspack for build` e2e test: after the production build, it rebuilds with the development configuration (`sourceMap: true`), parses the emitted `main.js.map`, and asserts a `.ts` source is present whose `sourcesContent` contains the original `@Component` TypeScript and no Ivy markers (`ɵcmp`). Verified locally via the local-registry e2e flow (production + development builds green).

## Related Issue(s)

N/A — no linked issue; addresses the Rspack/Angular sourcemap regression described above.

### DIFF
--- a/e2e/angular/src/projects-build-and-test.test.ts
+++ b/e2e/angular/src/projects-build-and-test.test.ts
@@ -129,6 +129,30 @@ describe('Angular Projects - Build and Test', () => {
       env: { NODE_ENV: 'production' },
     });
 
+    // build with sourcemaps enabled (the development configuration sets
+    // `sourceMap: true`) and verify the emitted sourcemap resolves back to
+    // the original TypeScript sources instead of the intermediate Ivy JS
+    runCLI(`build ${app} --skip-nx-cache`, {
+      env: { NGRS_CONFIG: 'development' },
+    });
+    const bundleMap = JSON.parse(
+      readFile(`dist/my-dir/${app}/browser/main.js.map`)
+    );
+    const tsSources = bundleMap.sources
+      .map((source: string, index: number) => ({
+        source,
+        content: bundleMap.sourcesContent?.[index],
+      }))
+      .filter(({ source }) => source.endsWith('.ts'));
+    expect(tsSources.length).toBeGreaterThan(0);
+    // the original TypeScript contains the `@Component` decorator, while the
+    // intermediate Ivy JS has it compiled away into `ɵcmp`
+    const componentSource = tsSources.find(({ content }) =>
+      content?.includes('@Component')
+    );
+    expect(componentSource).toBeDefined();
+    expect(componentSource.content).not.toContain('ɵcmp');
+
     if (runE2ETests()) {
       expect(() => runCLI(`e2e ${app}-e2e`)).not.toThrow();
       expect(await killPort(port)).toBeTruthy();
--- a/packages/angular-rspack-compiler/src/compilation/setup-compilation.spec.ts
+++ b/packages/angular-rspack-compiler/src/compilation/setup-compilation.spec.ts
@@ -1,0 +1,82 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { CompilerOptions } from 'typescript';
+import { setupCompilation } from './setup-compilation';
+
+const readConfiguration = vi.fn();
+
+vi.mock('../utils', () => ({
+  loadCompilerCli: vi.fn(async () => ({ readConfiguration })),
+}));
+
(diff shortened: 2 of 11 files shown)


leosvelperez:
Thanks for digging into this and for the PR @gultyayev! The sourcemap handling was already being reworked as part of the larger rebuild in #36268: the transform loaders now chain the maps that source-map-loader collects (inline and external .map references) and validate them against rspack's deserializer before forwarding, so dependency code resolves back to its original sources. Your e2e test for asserting the emitted sourcemap and the idea of validating maps before handing them to rspack both made it into that PR.

Closing this in favor of #36268.

gultyayev:
@leosvelperez thanks for landing the fix! I think, there's some missing piece, however. Today I was testing it and I noticed that files that don't belong to Angular components are empty at all. E.g., router config files, some other .ts files that are not component files - are completely empty in the Chrome DevTools.

PR #36381, opened 2026-07-16
docs(nx-cloud): document resource usage for runs without Nx Agents

lourw:
## Current Behavior

The resource usage page only documents two setups: Nx Agents on Nx Cloud compute, and bring your own compute. Both are distributed. 

Runs that don't run on Nx Agents (i.e CI main job) also get metrics collected and uploaded automatically, but the page doesn't mention them, and doesn't say where to find them in Nx Cloud.

## Expected Behavior

The page covers the non-distributed case:

- A third bullet in the enabling list for runs without Nx Agents, noting metrics upload automatically with nothing to configure.
- The viewing section is split into "Runs with Nx Agents" and "Runs without Nx Agents", the latter pointing at the **Resource usage** tab on the run details, with a screenshot.
- The intro and frontmatter description no longer scope the feature to distributed task execution.

## Related Issue(s)

N/A


astro-docs/src/content/docs/features/CI Features/resource-usage.mdoc
```diff
@@ -45,6 +44,8 @@ Once the add-on is active:
 - **When you [bring your own compute](/docs/guides/nx-cloud/bring-your-own-compute)** (running the agents on your own CI),
   add a single CLI step per agent job to upload metrics. See
   [the section below](#resource-metrics-when-you-bring-your-own-compute).
+- **Without Nx Agents**, metrics are collected and uploaded automatically for the machine running
```
jaysoo: I'd move this part about the BYOC since the latter would not be common.

Edit: Actually, since both with Nx Agents and without it just works, we don't need the list, say`Once the add-on is active, metrics are collected and uploaded...`. There isnt' a different from user POV.

BYOC can be an `{% aside %}` if they need to do something.

barbados-clemens: is this for manual DTE only or also for plain run/run-many commands without distribution?

lourw: plain run/run-many without distribution

lourw: **Without Nx Agents or without distributed execution**, metrics are collected and uploaded automatically for the machine running

?

barbados-clemens: nah it's fine mostly just making sure I understood since there is that tiny distinction of our internal knowledge but externally I think that's already assumed that's all I wanted to clarify for myself :)

jaysoo (approved):

PR #36386, opened 2026-07-17
cleanup(repo): route all first-party nx imports through @nx/devkit and enforce the boundary

FrozenPandaz:
## Current Behavior

First-party plugin packages import `nx` internals directly — ~474 import statements across ~330 files reaching into `nx/src/**`. Every one of those bypasses `@nx/devkit`, the layer that tracks which `nx` entry points are safe across the supported version range (see the policy in `packages/devkit/CLAUDE.md`, added in #36385). Nothing enforces the boundary, so new violations accumulate.

## Expected Behavior

Three commits, reviewable independently:

1. **Expose the needed surface** (`cleanup(core)`): 93 symbols first-party packages were deep-importing are now exported from `nx/src/devkit-internals.ts` and re-exported via `@nx/devkit/internal` (first-party-only, lockstep-versioned, no cross-version burden). The `run-commands` default executor and the batch-executor `TaskResult` are included under disambiguating names/notes. The `mock-project-graph` jest helper gets a dedicated `@nx/devkit/internal-testing-utils/mock-project-graph` subpath because importing it applies the mock — it cannot live in a barrel.

2. **The sweep** (`cleanup(repo)`): every direct `nx` import in `packages/*` (except `nx`/`devkit` themselves) rewritten to `@nx/devkit` or `@nx/devkit/internal`. Non-mechanical cases handled deliberately: disk-vs-Tree `readNxJson` call sites use `readNxJsonFromDisk`; gradle/maven batch executors use the batch `TaskResult` from `internal` (distinct type from `@nx/devkit`'s lifecycle `TaskResult`); `jest.mock('nx/src/...')` strings intentionally keep the underlying module path (mocks intercept through the barrel); one `jest.spyOn` namespace import keeps its direct path with an inline suppression (the spy must target the module the code under test binds).

3. **Enforcement** (`chore(repo)`): the shared eslint `baseConfig` restricts `nx`/`nx/**` imports with a message pointing at devkit and the CLAUDE.md policy. Non-published projects that legitimately reach into internals (e2e suites, the graph client, `packages/nx` itself) opt out via the exported `allowDirectNxImports` override; `packages/devkit` keeps its own stricter rule.

### Validation

- `nx run-many -t lint` across all 240 projects with `--skip-nx-cache`: green (cache skipped deliberately — remote-cached lint results predate the rule since the root eslint config is not part of the lint task hash; that inputs gap is a pre-existing issue worth a follow-up)
- `nx run-many -t build-base` across all packages: green (this caught and drove the fix for the two same-name/different-type collisions)
- jest resolution of the new devkit subpath smoke-tested through a generator spec

## Related Issue(s)

Follow-up to #36385 (the `packages/devkit/CLAUDE.md` policy this enforces).

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/TUI-Status-Bar-Development-11a216a4)


AgentEnder:
Supserceded by https://github.com/nrwl/nx/pull/36430

PR #36399, opened 2026-07-19
docs(misc): audit inferred plugin options and behavior across technology pages

barbados-clemens:
## Current Behavior

the technology pages drifted from what the inferred plugins actually do. matching rules, option lists, and defaults were hand-written at different times and never re-checked against `createNodes`/`createNodesV2`, so some pages list options that don't exist and omit ones that do. the jest `targetDefaults` example in the nx-json reference filters on `@nx/jest`, which matches nothing, since the filter wants the configured entry point `@nx/jest/plugin`.

## Expected Behavior

each page says which identifier goes in `nx.json`, what files the plugin matches, its options and defaults, and what targets it configures, all read off plugin source instead of the previous docs.

- treated the implementation and its tests as authoritative wherever they disagreed with the docs
- rollup and rsbuild had no inferred-task section at all, now they do
- worth a look: playwright used to say set `ciTargetName` to `false` to disable atomizer. the option is typed `string` and nothing tests `false`, though `if (options.ciTargetName)` suggests it does work in practice. i dropped that line and pointed at running the `targetName` task instead, which is not the same thing (the atomized targets still get created). put it back if someone is relying on it.
- generating this from a schema is still blocked on NXC-3871, so this is the manual accuracy pass in the meantime

draft because i haven't run the docs style check over the final state of all 25 pages yet.

## Related Issue(s)

DOC-554

### DIFF
--- a/astro-docs/src/content/docs/technologies/react/next/introduction.mdoc
+++ b/astro-docs/src/content/docs/technologies/react/next/introduction.mdoc
@@ -146,6 +146,8 @@ The `@nx/next` plugin creates tasks for any project that has a Next.js configura
 - `next.config.mjs`
 - `next.config.ts`
 
+The configuration directory must also contain a `package.json` or `project.json`.
+
 ### Plugin options
 
 Configure `@nx/next/plugin` in the `plugins` array in `nx.json`:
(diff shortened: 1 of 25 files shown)


astro-docs/src/content/docs/technologies/typescript/introduction.mdoc
```diff
@@ -360,30 +360,40 @@ Configure `@nx/js/typescript` in the `plugins` array in `nx.json`:
...
+| `build`                | Enables build inference; set to `false` or omit it to disable build tasks.    | disabled            |
+| `build.targetName`     | Name of the cached build task.                                                | `build`             |
+| `build.configName`     | The tsconfig file used for builds.                                            | `tsconfig.lib.json` |
+| `build.buildDepsName`  | Name of the task for building all dependencies.                               | `build-deps`        |
+| `build.watchDepsName`  | Name of the task for watching and rebuilding dependencies.                    | `watch-deps`        |
+| `build.skipBuildCheck` | Allows build inference without package exports that point to compiled output. | `false`             |
+| `verboseOutput`        | Adds TypeScript's `--verbose` flag to inferred commands.                      | `false`             |
+
+The inferred `typecheck` and `build` tasks are cached, declare their tsconfig-derived inputs and outputs, and depend on the corresponding tasks in dependency projects.
```
jaysoo, line 396: The `inferred` is redundant since we're in the "inferred" section.  Also, make sure we consistently say "task"; on other pages like vitest they say the "test target is cached".

astro-docs/src/content/docs/technologies/react/next/introduction.mdoc
```diff
@@ -160,19 +162,24 @@ Configure `@nx/next/plugin` in the `plugins` array in `nx.json`:
...
+| Option                  | Description                                | Default        |
+| ----------------------- | ------------------------------------------ | -------------- |
+| `buildTargetName`       | Name of the cached `next build` task.      | `build`        |
+| `devTargetName`         | Name of the continuous `next dev` task.    | `dev`          |
+| `startTargetName`       | Name of the continuous `next start` task.  | `start`        |
+| `serveStaticTargetName` | Deprecated alias for `startTargetName`.    | `serve-static` |
+| `buildDepsTargetName`   | Name of an optional dependency build task. | none           |
+| `watchDepsTargetName`   | Name of an optional dependency watch task. | none           |
+
+The build task reads `distDir` from the Next.js configuration to declare cache outputs and depends on builds of project dependencies. The start task depends on the build task.
```
jaysoo, line 182: I feel like we don't need to be too specific. Like the TS description says it's derived from tsconfig, but we don't need to mention details, that'd just be describing the code. I think a general list of what's configured and what users care about:

1. Caching? Yes/no
2. Inputs/ouputs? which files affect it?
3. Dependencies are set up? yes/no

jaysoo (approved):

PR #36417, opened 2026-07-20
docs(nx-cloud): document start-nx-agents and the .nx/ci-config.yaml file

lourw:
## Current Behavior

Nx Cloud CI is documented almost entirely around the `nx-cloud start-ci-run` command and its flags. The `.nx/ci-config.yaml` file and the `nx-cloud start-nx-agents` command are not documented anywhere, so there is no reference for the config schema and no guidance to prefer the config-file workflow.

## Expected Behavior

The docs lead with `start-nx-agents` and the `.nx/ci-config.yaml` file, while keeping `start-ci-run` documented as the flag-based alternative. The two are mutually exclusive (`start-ci-run` exits when a config file is present), so both carry a caution.

Changes:

- **New reference page** `reference/Nx Cloud/ci-config.mdoc` documenting every `.nx/ci-config.yaml` key (`lifecycle`, `ai`, `dte`, `nx-agents`, `overrides`) with types and defaults, plus a sidebar entry in the Continuous integration group.
- **New migration guide** `guides/Nx Cloud/migrate-to-ci-config.mdoc`: why to migrate (one place to control every pipeline; configuration that does not depend on command order), a before/after example, the flag-to-config mapping, and the mutually-exclusive cutover.
- **CLI reference** (`reference/nx-cloud-cli.mdoc`) gains a `start-nx-agents` section above `start-ci-run` and a flag-to-config mapping table.
- **Getting started and the Nx Agents feature page** lead with the config file and `start-nx-agents` for distribution.

The docs describe the accurate model, verified against the Nx Cloud client source: every Nx Cloud command reads `.nx/ci-config.yaml`, so the configuration applies to the run whichever command starts it. `start-nx-agents` provisions the Nx Agents; it is not the command that has to run first to configure the run. `start-ci-run` is the one command that does not read the file.

## Related Issue(s)

Closes CLOUD-4872

---
[View session information ↗](https://snapshot.app.trypolygraph.com/orgs/69cdc268b6aa527e4129c2b4/sessions/wise-moose-19f423de)


astro-docs/src/content/docs/guides/Nx Cloud/migrate-to-ci-config.mdoc
```diff
@@ -0,0 +1,107 @@
...
+lifecycle:
+  fail-after: 1
+
+overrides:
+  main:
+    lifecycle:
+      fail-after: 0
+```
+
+Here every pipeline stops on the first failure, except when `NX_CLOUD_ENV` is `main`, where the run finishes so you see every failure at once.
```
barbados-clemens: am I out of the loop. what is NX_CLOUD_ENV? is that execution env or something different?

astro-docs/src/content/docs/reference/Nx Cloud/ci-config.mdoc
```diff
@@ -0,0 +1,232 @@
...
+Environment variables forwarded from the main job to every Nx Agent.
+Variables prefixed with `NX_` are forwarded automatically, so list only the additional ones you need.
+
+Type: array of strings.
+
+## `overrides`
+
+Per-environment overrides keyed by `NX_CLOUD_ENV` value.
+Each override is a partial config that takes precedence over the base sections when its key matches the active environment.
+This has no `start-ci-run` flag equivalent. It is specific to the config file.
```
barbados-clemens: so we have to have both NX_CLOUD_ENV and NX_CI_EXECUTION_ENV or does NX_CLOUD_ENV read from execution env?

barbados-clemens (commented):
question on new env var. can we also add it to the nx environment variables lists for the cloud section?

astro-docs/src/content/docs/guides/Nx Cloud/migrate-to-ci-config.mdoc
lourw: Sorry yes, that should be NX_CI_EXECUTION_ENV

astro-docs/src/content/docs/reference/nx-cloud-cli.mdoc
```diff
@@ -346,8 +346,56 @@ npx nx-cloud onboard workspace create --org=org_123 --template=tmpl_react --name
...
+    - test
+    - e2e
+```
+
+{% aside type="caution" title="start-nx-agents and start-ci-run cannot be combined" %}
+`start-ci-run` does not read `.nx/ci-config.yaml`. If a config file is present, `start-ci-run` exits with an error and points you to `start-nx-agents`.
+Pick one path for your workspace: the config file with `start-nx-agents`, or CLI flags with `start-ci-run`.
+{% /aside %}
+
+#### Migrating from start-ci-run flags
```
jaysoo: I would not have a "migrate" section here and point to the guide instead (`astro-docs/src/content/docs/guides/Nx Cloud/migrate-to-ci-config.mdoc`).

astro-docs/src/content/docs/guides/Nx Cloud/migrate-to-ci-config.mdoc
```diff
@@ -0,0 +1,107 @@
...
+
+### Configuration that does not depend on command order
+
+Every Nx Cloud command reads `.nx/ci-config.yaml` when it runs, so your configuration applies to the CI Pipeline Execution no matter which command starts it.
+`start-nx-agents` provisions the agents your job needs. It is not a special command that has to run first to lock in the configuration.
+With `start-ci-run`, configuration instead came from that command's flags, so it had to run before any other `nx` command for its settings to apply.
+
+{% aside type="caution" title="The two commands cannot be combined" %}
+`start-ci-run` does not read `.nx/ci-config.yaml`. Once the file exists, `start-ci-run` exits with an error and points you to `start-nx-agents`.
+This is a switch, not an addition: move your flags into the file and change the command in the same step. Do not leave both in place.
```
jaysoo: Can be shortened, no need to say it's a switch.


```suggestion
Move your CLI flags into the file and change the command in the same step. Do not leave both in place.
```

astro-docs/src/content/docs/features/CI Features/distribute-task-execution.mdoc
```diff
@@ -299,11 +310,13 @@ jobs:
       - run: pnpm exec nx affected -t lint test build
 ```
 
-This command tells Nx Cloud to:
+This tells Nx Cloud to:
+
+- Start a CI run (`npx nx-cloud start-nx-agents`)
```
jaysoo: Since it's not starting a CI run we can update the text.

```suggestion
- Provisions Nx Agents (`npx nx-cloud start-nx-agents`)
```

lourw:
Good catch on `NX_CLOUD_ENV` — it was the wrong variable (that name is the launch-template env *file*). Fixed to `NX_CI_EXECUTION_ENV` across the guide and the ci-config reference, and I extended the `NX_CI_EXECUTION_ENV` entry in the environment variables reference to note it also selects the matching `overrides` entry in `.nx/ci-config.yaml`.

Note: the Ocean schema file (`ci-config.schema.json`) still has the same wrong name in its `overrides` description. That's a separate Ocean fix, deferred for now.

lourw:
Thanks — all three applied:
- Removed the flag mapping table and `Migrating from start-ci-run flags` section from the CLI reference; it now points to the migration guide, which carries the mapping.
- Shortened the caution per your suggestion.
- `Start a CI run` → `Provisions Nx Agents` on the distribute-task-execution page.

barbados-clemens (approved):

jaysoo (approved):

PR #36424, opened 2026-07-21
fix(vite): correct tsconfig paths fallback resolution

### LINKED ISSUE #36400: loadFileFromPaths in @nx/vite breaks wildcard tsconfig.paths resolution after upgrade from 22.6.4 → 22.7.6
### Current Behavior

---

After upgrading from 22.6.4 to 22.7.6, `nx run app:test` fails with `Failed to resolve import` for every cross-package alias. The root cause is two separate issues in `nxViteTsPaths`:

1. `resolvePathsBaseUrl()` resolves `baseUrl` relative to the **leaf tsconfig** instead of the tsconfig where `baseUrl` is actually declared — causing all path aliases to be constructed under the wrong directory.
2. `loadFileFromPaths()` (the fallback resolver) has three independent bugs with wildcard aliases: it only tries `paths[0]`, never substitutes the `*` wildcard in path values, and has no early return — so a correct match can be silently overwritten.

---

### Expected Behavior

Wildcard `tsconfig.paths` aliases defined in a root `tsconfig.base.json` should resolve correctly regardless of how deep the leaf tsconfig is in the directory tree.

---

### GitHub Repo

_No response_

### Steps to Reproduce

1. Have a monorepo where `paths` and `baseUrl: "."` are declared in a root `tsconfig.base.json`
2. Have an app whose leaf tsconfig (`tsconfig.app.json`) is nested 2+ levels deep (e.g. `apps/app-name/`)
3. Use wildcard aliases, e.g. `"@repo/foo/*": ["packages/foo/*.ts", "packages/foo/*/index.ts"]`
4. Run `nx run <app>:test`
5. Observe `Failed to resolve import` for all cross-package aliases

---

### Nx Report

```shell
Node           : 22.22.2
OS             : darwin-arm64
npm            : 10.9.7
daemon         : Available

nx                     : 22.7.6
@nx/js                 : 22.7.6
@nx/eslint             : 22.7.6
@nx/workspace          : 22.7.6
@nx/angular            : 22.7.6
@nx/jest               : 22.7.6
@nx/devkit             : 22.6.4
@nx/esbuild            : 22.7.6
@nx/eslint-plugin      : 22.7.6
@nx/module-federation  : 22.7.6
@nx/node               : 22.7.6
@nx/rspack             : 22.7.6
@nx/vite               : 22.7.6
@nx/vitest             : 22.7.6
@nx/web                : 22.7.6
@nx/webpack            : 22.7.6
@nx/docker             : 22.7.6
create-nx-workspace    : 22.5.4
Registered Plugins:
@nx/eslint/plugin
@nx/vite/plugin
Community plugins:
@analogjs/vite-plugin-angular : 2.4.0
@analogjs/vitest-angular      : 2.4.0
@ngrx/effects                 : 21.1.0
@ngrx/router-store            : 21.1.0
@ngrx/store                   : 21.1.0
angular-eslint                : 21.3.1
Local workspace plugins:
@app-name/nx
Cache Usage: 25.18 MB / 46.04 GB
  - @nx/devkit@22.6.4
  - create-nx-workspace@22.5.4

To fix this, run `nx migrate nx@22.7.6`
⚠️ Multiple Nx versions detected

Your workspace uses nx@22.7.6, but other packages depend on a different version:
  - @nx/rspack → @nx/devkit → nx@22.4.5

These packages should not have nx as a dependency. Please report this issue to the package maintainers.
Run `npm explain nx@22.4.5` for more details.
```

### Failure Logs

```shell
## Failure Logs


Failed to resolve import "@app-name/..." from "src/..."


---
```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Bug 1 — `resolvePathsBaseUrl` uses the leaf tsconfig directory instead of the declaring tsconfig directory

`paths` and `baseUrl: "."` are declared in `tsconfig.base.json` at the workspace root.
In 22.7.x, `resolvePathsBaseUrl()` returns the directory of the **leaf tsconfig** (`apps/storefrontapp/`) instead of the directory of the **file that actually declares `baseUrl`**.

As a result, every alias is constructed under `apps/storefrontapp/app-name/...` which doesn't exist, and all imports fail.

Other apps in the same repo are unaffected because their leaf tsconfigs sit closer to the root, so the wrong base URL accidentally still produces a valid path.

**Workaround:** add `baseUrl: "../.."` to the leaf tsconfig so `resolvePathsBaseUrl` finds it before walking up to `tsconfig.base.json` and resolves it correctly to the workspace root.

---

### Bug 2 — `loadFileFromPaths` has 3 issues with wildcard aliases

This function is the fallback path resolver — it is reached when `matchTsPathEsm` fails (which it does for all imports once Bug 1 is present). It was apparently written for non-wildcard aliases only.

**Bug 2a — only `paths[0]` is ever tried**

`tsconfig.paths` values are arrays to support fallback patterns, but the function hardcodes `paths[0]`. If the first pattern doesn't resolve to an existing file, resolution fails silently even if a second pattern would have matched.

```json
"@repo/app/clients/*": [
  "packages/clients/app/*.ts",       // tried — angular.ts doesn't exist → undefined
  "packages/clients/app/*/index.ts"  // never reached
]
```

**Bug 2b — the `*` wildcard in path values is never substituted**

After stripping `/*` from the alias key, the `*` character in the path value is left as a literal. The suffix captured from the import specifier is never substituted in.

For `importPath = "@repo/app/clients/angular"`:

```
paths[0]     = "packages/clients/app/*.ts"
after strip  = "packages/clients/app/*"           // * still literal
result       = "/workspace/packages/clients/app/*/angular"  // not a real path
```

**Bug 2c — no early return after a successful match**

After finding a valid `resolvedFile`, the loop keeps going. If a broader prefix alias later matches the same import via `startsWith`, it overwrites the correct result with `undefined`.

```
1. "@app/feature/mcs/root"  → resolvedFile = "...public_api.ts"  ✓
2. "@app/feature"            → resolvedFile = undefined           ✗ overwrites!
3. "@app/feature/mcs"        → resolvedFile = undefined           ✗ overwrites!
```

In our project this silently breaks **124 out of 153** aliases.

---

### Proposed fix for `loadFileFromPaths`

```js
function loadFileFromPaths(tsconfig, importPath) {
    let resolvedFile;

    for (const alias in tsconfig.paths) {
        const paths = tsconfig.paths[alias];
        const isWildcard = alias.endsWith('/*');
        const normalizedImport = alias.replace(/\/\*$/, '');

        if (importPath === normalizedImport ||
            importPath.startsWith(normalizedImport + '/')) {

            const suffix = importPath.slice(normalizedImport.length + 1);

            for (const pathPattern of paths) {               // fix 2a: iterate all patterns
                const resolvedPattern = isWildcard
                    ? pathPattern.replace('*', suffix)        // fix 2b: substitute wildcard
                    : pathPattern.replace(/\/\*$/, '');
                const joinedPath = isWildcard
                    ? resolve(tsconfig.absoluteBaseUrl, resolvedPattern)
                    : joinPathFragments(tsconfig.absoluteBaseUrl, resolvedPattern);
                const candidate = isWildcard
                    ? joinedPath
                    : importPath.replace(normalizedImport, joinedPath);

                resolvedFile = findFile(candidate, options.extensions);

                if (resolvedFile) return resolvedFile;        // fix 2c: early return on match
            }
        }
    }

    return resolvedFile;
}
```

leosvelperez:
## Current Behavior

The plugin looks up an alias in two passes. The first asks `tsconfig-paths`, which looks only for `.js`, `.json` and `.node` files. Most aliases miss it and fall to a second pass nx runs itself. That pass has five bugs.

- An alias fails when the `*` is not last in the mapped path. `"@app/*": ["packages/*/src"]` never resolves.
- A mapped path holding `$&`, `` $` `` or `$'` sends nx to the wrong folder.
- A project can inherit aliases from a config it extends. nx reads them against the project folder, not the folder that set them, so each one points at a missing file.
- When two aliases match one import, nx takes the one listed first, not the narrower one. A wildcard that names a file extension wins even when the narrower alias comes first.
- On Windows, nx drops the drive letter and searches the wrong volume.

A workspace may have no root config. nx then searches upward from the current folder, and finds one outside the workspace or throws.

nx reads the chain of extended configs as strict JSON. One comment drops a file from that chain. Users hit this through the `@nx/vitest` executor.

## Expected Behavior

An alias resolves wherever the `*` sits in the mapped path. `"@app/*": ["packages/*/src"]` finds `packages/one/src/index.ts`. nx resolves a mapped path holding a `$` sequence as written.

A project that inherits aliases reads them against the folder that set them. When two aliases match, nx picks the narrower one. An exact alias beats a wildcard, and the longer prefix wins between two wildcards. The order they are listed in stops mattering.

On Windows the drive letter survives. A workspace with no root config hands the import back to Vite. A project config still works on its own.

nx walks the chain of extended configs the way TypeScript does. Comments and stray commas leave aliases pointing where the config says.

## Related Issue(s)

Fixes #36400

## Implementation Notes

- `tsconfig-paths` ranks an alias by the text before its `*`. An alias without one counts as zero, so its matcher never prefers an exact alias. Two passes run ahead of it.
- The first added pass keeps package entry resolution, which the file probe does not do. The second knows the extensions the plugin is configured with, so it finds an index file the first pass misses.
- Picking the narrower alias changes which file a workspace gets. It affects a workspace that lists a broad alias first and leans on the old order.
- Three cases stay looser than TypeScript. An exact alias can match a prefix. A lookup can fall through to another alias. A plain import can match a `"/*"` key. Workspaces build on all three today, so tightening them waits for a major.
- Resolution is checked against `ts.resolveModuleName`. The check spans mapped path shapes, import extensions and three module resolution modes.
- The Windows fix carries no test. On POSIX the old and new path helpers agree for the shapes this resolver builds.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/gh-36400-796cc426)

### DIFF
--- a/packages/vite/plugins/nx-tsconfig-paths.plugin.spec.ts
+++ b/packages/vite/plugins/nx-tsconfig-paths.plugin.spec.ts
@@ -1,0 +1,105 @@
+import { TempFs } from 'nx/src/internal-testing-utils/temp-fs';
+import { join } from 'node:path';
+
+// `var` rather than `let`: transitive imports read `workspaceRoot` while the
+// module graph is still loading, before a `let` would leave its temporal dead
+// zone.
+var workspaceRootMock: string | undefined;
+jest.mock('@nx/devkit', () => {
+  const actual = jest.requireActual('@nx/devkit');
+  return {
+    ...actual,
+    get workspaceRoot() {
+      return workspaceRootMock ?? actual.workspaceRoot;
+    },
+  };
+});
+
+var failRootTsConfigLoad = false;
+jest.mock('tsconfig-paths', () => {
+  const actual = jest.requireActual('tsconfig-paths');
+  return {
+    ...actual,
+    loadConfig: (path?: string) =>
+      failRootTsConfigLoad && path?.endsWith('tsconfig.base.json')
+        ? { resultType: 'failed', message: "Couldn't find tsconfig.json" }
+        : actual.loadConfig(path),
+  };
+});
+
+import { nxViteTsPaths } from './nx-tsconfig-paths.plugin';
+
+describe('nxViteTsPaths', () => {
+  let tempFs: TempFs;
+  let originalTsConfigPath: string | undefined;
+
+  beforeEach(() => {
+    tempFs = new TempFs('nx-vite-ts-paths');
+    workspaceRootMock = tempFs.tempDir;
+    originalTsConfigPath = process.env.NX_TSCONFIG_PATH;
+    failRootTsConfigLoad = false;
+  });
+
+  afterEach(() => {
+    if (originalTsConfigPath === undefined) {
+      delete process.env.NX_TSCONFIG_PATH;
+    } else {
+      process.env.NX_TSCONFIG_PATH = originalTsConfigPath;
+    }
+    tempFs.cleanup();
+  });
+
+  const resolveWith = async (importPath: string) => {
+    const plugin = nxViteTsPaths();
+    await (plugin as any).configResolved({
+      root: join(tempFs.tempDir, 'app'),
+      command: 'build',
+      plugins: [],
+    });
+    return (plugin as any).resolveId(importPath);
+  };
+
+  const withProjectTsConfigOutsideWorkspace = async () => {
+    await tempFs.createFiles({
+      'external/tsconfig.json': JSON.stringify({
+        compilerOptions: { baseUrl: '.', paths: { '@ext/*': ['libs/*'] } },
+      }),
+      'app/src/main.ts': '',
+    });
+    process.env.NX_TSCONFIG_PATH = join(
+      tempFs.tempDir,
+      'external/tsconfig.json'
+    );
+  };
+
+  it('should defer to other resolvers when the workspace has no root-level tsconfig', async () => {
+    await withProjectTsConfigOutsideWorkspace();
+
+    await expect(resolveWith('@nope/missing')).resolves.toBeNull();
+  });
+
+  it('should defer to other resolvers when the root-level tsconfig cannot be loaded', async () => {
+    await withProjectTsConfigOutsideWorkspace();
+    await tempFs.createFiles({ 'tsconfig.base.json': JSON.stringify({}) });
+    failRootTsConfigLoad = true;
+
+    await expect(resolveWith('@nope/missing')).resolves.toBeNull();
+  });
+
+  it('should resolve a workspace alias through the root-level tsconfig', async () => {
+    await tempFs.createFiles({
+      'tsconfig.base.json': JSON.stringify({
+        compilerOptions: {
+          baseUrl: '.',
+          paths: { '@repo/util': ['libs/util/index.ts'] },
+        },
+      }),
+      'libs/util/index.ts': '',
+      'app/src/main.ts': '',
+    });
+
+    await expect(resolveWith('@repo/util')).resolves.toEqual(
+      join(tempFs.tempDir, 'libs/util/index.ts')
+    );
+  });
+});
--- a/packages/vite/plugins/nx-tsconfig-paths.plugin.ts
+++ b/packages/vite/plugins/nx-tsconfig-paths.plugin.ts
@@ -1,7 +1,6 @@
 import {
   createProjectGraphAsync,
   getPackageManagerCommand,
-  joinPathFragments,
   workspaceRoot,
 } from '@nx/devkit';
 import {
@@ -20,7 +19,7 @@ import {
 } from 'tsconfig-paths';
 import { Plugin } from 'vite';
 import { warnNxViteTsPathsDeprecation } from '../src/utils/deprecation';
-import { findFile } from '../src/utils/nx-tsconfig-paths-find-file';
+import { loadFileFromPaths } from '../src/utils/nx-tsconfig-paths-load-file';
 import { getProjectTsConfigPath } from '../src/utils/options-utils';
 import { nxViteBuildCoordinationPlugin } from './nx-vite-build-coordination.plugin';
 
@@ -40,7 +39,7 @@ export interface nxViteTsPathsOptions {
   mainFields?: (string | string[])[];
   /**
    * extensions to check when resolving files when package.json resolution fails
-   * @default ['.ts', '.tsx', '.js', '.jsx', '.json', '.mjs', '.cjs']
+   * @default ['.ts', '.tsx', '.js', '.jsx', '.json', '.mts', '.mjs', '.cts', '.cjs', '.css', '.scss', '.less']
    **/
   extensions?: string[];
   /**
@@ -74,7 +73,7 @@ export function nxViteTsPaths(options: nxViteTsPathsOptions = {}) {
   let matchTsPathEsm: MatchPath;
   let matchTsPathFallback: MatchPath | undefined;
   let tsConfigPathsEsm: ConfigLoaderSuccessResult;
-  let tsConfigPathsFallback: ConfigLoaderSuccessResult;
+  let tsConfigPathsFallback: ConfigLoaderSuccessResult | undefined;
 
   options.extensions ??= [
     '.ts',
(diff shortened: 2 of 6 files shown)


packages/vite/plugins/nx-tsconfig-paths.plugin.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+  const actual = jest.requireActual('@nx/devkit');
+  return {
+    ...actual,
+    get workspaceRoot() {
+      return workspaceRootMock ?? actual.workspaceRoot;
+    },
+  };
+});
+
+var failRootTsConfigLoad = false;
```
jaysoo: Replace `var` with `let`.

packages/vite/plugins/nx-tsconfig-paths.plugin.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+  afterEach(() => {
+    if (originalTsConfigPath === undefined) {
+      delete process.env.NX_TSCONFIG_PATH;
+    } else {
+      process.env.NX_TSCONFIG_PATH = originalTsConfigPath;
+    }
+    tempFs.cleanup();
+  });
+
+  const resolveWith = async (importPath: string) => {
```
jaysoo, line 29: Nit: Weird to use arrow functions when we use `function` elsewhere.

packages/vite/plugins/nx-tsconfig-paths.plugin.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+      'app/src/main.ts': '',
+    });
+    process.env.NX_TSCONFIG_PATH = join(
+      tempFs.tempDir,
+      'external/tsconfig.json'
+    );
+  };
+
+  it('should defer to other resolvers when the workspace has no root-level tsconfig', async () => {
+    await withProjectTsConfigOutsideWorkspace();
```
jaysoo: This feels like a `describe` block like `describe('when tsconfig is outside workspace')`.

packages/vite/plugins/nx-tsconfig-paths.plugin.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+
+    await expect(resolveWith('@ext/foo')).resolves.toEqual(
+      join(tempFs.tempDir, 'external/libs/foo.ts')
+    );
+  });
+
+  it('should defer to other resolvers when the root-level tsconfig cannot be loaded', async () => {
+    await withProjectTsConfigOutsideWorkspace();
+    await tempFs.createFiles({ 'tsconfig.base.json': JSON.stringify({}) });
+    failRootTsConfigLoad = true;
```
jaysoo: This test says root tsconfig cannot be loaded, but the line above creates one? Is that needed? I guess so before of `tsconfig-paths` mock, but reads a bit weird.

packages/vite/plugins/nx-tsconfig-paths.plugin.ts
```diff
@@ -277,54 +309,64 @@ export function nxViteTsPaths(options: nxViteTsPathsOptions = {}) {
...
-          ) {
-            const foundExtension = options.extensions.find((ext) =>
-              importPath.endsWith(ext)
-            );
-            const pathWithoutExtension = importPath
-              .replace(normalizedImport, joinedPath)
-              .slice(0, -foundExtension.length);
-            resolvedFile = findFile(pathWithoutExtension, options.extensions);
-          }
+    logIt(
```
jaysoo, line 346: TBH, the logging is super noisy when we set `NX_VERBOSE_LOGGING`. While we're in here maybe we change it to `NX_VITE_TSCONFIG_PATHS_LOGGING` or something.

packages/vite/src/utils/nx-tsconfig-paths-load-file.ts
```diff
@@ -0,0 +1,103 @@
...
+function sortAliasesBySpecificity(
+  aliases: string[],
+  importPath: string
+): string[] {
+  const isWildcard = (alias: string) => alias.endsWith('/*');
+  const rank = (alias: string) =>
+    isWildcard(alias) ? 1 : importPath === alias ? 2 : 0;
+  const prefixLength = (alias: string) => alias.replace(/\/\*$/, '').length;
+
+  return [...aliases].sort((a, b) => {
```
jaysoo, line 95: Since aliases is from `Object.keys(tsconfig.paths)`, the mutation here doesn't matter and we can avoid `[...aliases]`.

jaysoo (approved):

Issue #36425, opened 2026-07-21
Ambiguous scope "*" is still checked with useCommitScope: false


snapydziuba:
### Current Behavior

When setting `"useCommitScope": false` for `conventionalCommits` releasing with a commit like `feat: test` will work but `feat(*): test` would break as https://github.com/nrwl/nx/blob/a6dd4f0bde4a57be1008dcf8bbabe1c494086fc2/packages/nx/src/command-line/release/utils/shared.ts#L497 does not check for `config.useCommitScope` thus running always the ambiguity check.

### Expected Behavior

When releasing with `feat(*): test` and `"useCommitScope": false` the ambiguity should not be applied as we use the affected files logic that determines each project’s bump.

### Steps to Reproduce

1. `ns.json` settings
```
    "conventionalCommits": {
      "useCommitScope": false
    },
    "projectsRelationship": "independent",
    "releaseTag": {
      "requireSemver": true,
      "strictPreid": true
    },
    "version": {
      "conventionalCommits": true
    }
```

2. `git commit --allow-empty -m 'feat(*): test'`
3. `nx release version --dry-run`

This results in the error `Ambiguous scope "*" in commit "feat(*): test".` from https://github.com/nrwl/nx/blob/a6dd4f0bde4a57be1008dcf8bbabe1c494086fc2/packages/nx/src/command-line/release/utils/shared.ts#L516

### Nx Report

```shell
Node           : 22.23.1
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.9.8
daemon         : Available

nx             : 23.0.1
@nx/js         : 23.0.1
@nx/workspace  : 23.0.1
@nx/devkit     : 23.0.1
typescript     : 6.0.3
---------------------------------------
Registered Plugins:
@nx/js/typescript
---------------------------------------
Cache Usage: 110.32 MB / 90.33 GB
```

### Failure Logs

```shell

```

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

### Additional Information

https://github.com/nrwl/nx/issues/35744 fixed a similar issue but only so that ambiguity checks scopes stay in group

PR #36427, opened 2026-07-21
fix(release): honor commit scope configuration (#36425)

snapydziuba:
## Current Behavior

nx release currently always checks for `scopePatterns` even when `useCommitScope` is set to `false`
 https://github.com/nrwl/nx/blob/a6dd4f0bde4a57be1008dcf8bbabe1c494086fc2/packages/nx/src/command-line/release/utils/shared.ts#L497
 thus running into the ambiguity check where `feat(*)` would trigger a throw.
 
## Expected Behavior

I would expect that when setting `useCommitScope` to falls my scopes are not checked and we would directly fall back to the file-affectedness path matching. Even when we commit `feat(lib-b)` with a lib-a file i would expect it to be ignored and match against lib-a.

## Related Issue(s)
#36425

### DIFF
--- a/packages/nx/src/command-line/release/utils/shared.spec.ts
+++ b/packages/nx/src/command-line/release/utils/shared.spec.ts
@@ -1326,6 +1326,32 @@ describe('shared', () => {
       expect(scopedCommits?.[0].isProjectScopedCommit).toBe(true);
     });
 
+    it('should ignore commit scope when useCommitScope is disabled', async () => {
+      mockReleaseConfig!.conventionalCommits = {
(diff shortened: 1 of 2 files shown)


AgentEnder:

<details>
<summary>

</summary>

>
> The title and body both point at #36425, which is open, but there is no closing keyword so GitHub has not linked them. Adding `Fixes #36425` to the description would close the issue automatically on merge.
>
> Branch merges cleanly.

</details>

snapydziuba:
@AgentEnder rebased :+1:

PR #36434, opened 2026-07-22
feat(core): add bun dependency-catalog support

cogwirrel:
## Current Behavior

Bun supports dependency catalogs: the root `package.json` carries a `catalog` field (the default catalog) and/or a `catalogs` field (named catalogs), and dependencies reference them with the `catalog:` / `catalog:<name>` protocol. From bun 1.2.14 the fields are read from the object form of `workspaces`; from 1.2.19 they may also live at the top level of `package.json`.

Nx resolves `catalog:` references through per-package-manager catalog managers (pnpm and yarn exist today), but `getCatalogManager` returns `null` for bun. As a result, `getDependencyVersionFromPackageJson` returns the raw unresolved `"catalog:"` string in bun workspaces, and generators that feed that into `semver.coerce(...).version` without a null guard crash:

```
TypeError: Cannot read properties of null (reading 'version')
```

e.g. `@nx/vitest` `getInstalledViteVersion` and `@nx/react` `getInstalledReactVersion` crash when `vite` / `react` are catalogued in a bun workspace, so generators like `@nx/vitest:configuration` fail outright.

Additionally, the `dependency-checks` lint rule and the yarn catalog manager hardcode pnpm's rule that `"default"` aliases the `catalog` field. Verified against the real binaries, that rule holds only for pnpm — yarn berry (>= 4.10) and bun both treat `catalog` and `catalogs.default` as separate catalogs.

## Expected Behavior

- `getCatalogManager` returns a `BunCatalogManager` for bun workspaces, which reads/writes catalogs from the root `package.json`. Catalog references now resolve to their declared ranges in generators, migrations, release versioning, lockfile pruning, and lint dependency checks — matching the existing pnpm/yarn behavior. This alone fixes the `@nx/vitest` / `@nx/react` crashes, since the resolved range reaches semver instead of the raw `"catalog:"` string; no changes to those packages are needed.
- The bun JSON-based helpers live in a dedicated `bun-manager-utils.ts`, keeping `manager-utils.ts` scoped to the YAML-based pnpm/yarn logic.
- The manager follows bun's actual resolution semantics, verified empirically against bun 1.3.14 (they differ from pnpm in two ways):
  - **Locations are all-or-nothing:** when either catalog field exists nested inside the object form of `workspaces`, bun ignores the top-level fields entirely (no merging across locations). The manager reads, and routes updates to, whichever location is active.
  - **"default" is not special:** `catalog:` resolves only against the singular `catalog` field, and `catalog:default` addresses a named catalog literally called `default` (bun fails to resolve `catalog:` from `catalogs.default` and vice versa). Whitespace-only names (e.g. `catalog:  `) are treated as the default catalog.
- Catalog updates preserve the user's `package.json` formatting (surgical edits via `jsonc-parser`), tolerate null placeholders, and are no-ops when the version already matches.
- Default-catalog candidate selection in the `dependency-checks` lint rule moves behind the `CatalogManager` interface via a new `getCatalogReferencesForPackage` method, so `nx lint --fix` emits specifiers the workspace's package manager actually accepts (e.g. `catalog:default` rather than `catalog:` for a bun `catalogs.default` entry).
- A follow-up commit aligns the yarn manager with yarn berry's real semantics (same separate-catalogs rule as bun), removing its pnpm-style `"default"` aliasing; `updateCatalogVersionsInFile` gains an `aliasDefaultCatalog` option so pnpm keeps its behavior unchanged.
- `nx release` no longer logs the hardcoded `pnpm-workspace.yaml` filename for catalog updates; it derives the file from the active catalog manager.

Covered by new unit tests for the bun manager (reference parsing, default/named/`workspaces`-nested resolution, location precedence, validation, updates including null-placeholder edge cases), `getCatalogReferencesForPackage` tests for all three managers, `dependency-checks` bun fixtures, updated yarn manager tests asserting yarn's separate-catalogs semantics, catalog-dependency detection tests, and devkit `getDependencyVersionFromPackageJson` bun tests. Existing pnpm catalog suites pass unchanged. Also verified end-to-end in a real bun workspace: with published nx, `nx g @nx/vitest:configuration` reproduces the crash; with this branch's catalog module in place, the generator succeeds and the catalogued `vite` version resolves correctly for both definition locations.

### DIFF
--- a/packages/devkit/src/utils/package-json.ts
+++ b/packages/devkit/src/utils/package-json.ts
@@ -37,9 +37,10 @@ const NON_SEMVER_TAGS = {
 /**
  * Get the resolved version of a dependency from package.json.
  *
- * Retrieves a package version and automatically resolves PNPM catalog references
- * (e.g., "catalog:default") to their actual version strings. By default, searches
- * `dependencies` first, then falls back to `devDependencies`.
+ * Retrieves a package version and automatically resolves package manager
+ * catalog references (e.g., "catalog:default") to their actual version
+ * strings. By default, searches `dependencies` first, then falls back to
+ * `devDependencies`.
  *
  * **Tree-based usage** (generators and migrations):
  * Use when you have a `Tree` object, which is typical in Nx generators and migrations.
--- a/packages/nx/src/utils/catalog/manager-factory.ts
+++ b/packages/nx/src/utils/catalog/manager-factory.ts
@@ -1,4 +1,5 @@
 import { detectPackageManager } from '../package-manager';
+import { BunCatalogManager } from './bun-manager';
 import type { CatalogManager } from './manager';
 import { PnpmCatalogManager } from './pnpm-manager';
 import { YarnCatalogManager } from './yarn-manager';
@@ -16,6 +17,8 @@ export function getCatalogManager(
       return new PnpmCatalogManager();
     case 'yarn':
       return new YarnCatalogManager();
+    case 'bun':
+      return new BunCatalogManager();
     default:
       return null;
   }
--- a/packages/nx/src/utils/catalog/manager-utils.ts
+++ b/packages/nx/src/utils/catalog/manager-utils.ts
@@ -168,8 +168,17 @@ export function updateCatalogVersionsInFile(
     packageName: string;
     version: string;
     catalogName?: string;
-  }>
+  }>,
+  options?: {
+    /**
+     * Treat "default" as an alias for the `catalog` field and route default
+     * updates through a populated `catalogs.default` (pnpm semantics). When
+     * false, "default" is an ordinary named catalog (yarn semantics).
+     */
+    aliasDefaultCatalog?: boolean;
+  }
 ): void {
+  const aliasDefaultCatalog = options?.aliasDefaultCatalog ?? true;
   let checkExists: () => boolean;
   let readYaml: () => string;
   let writeYaml: (content: string) => void;
@@ -234,14 +243,18 @@ export function updateCatalogVersionsInFile(
     for (const update of updates) {
       const { packageName, version, catalogName } = update;
       const normalizedCatalogName =
-        catalogName === 'default' ? undefined : catalogName;
+        aliasDefaultCatalog && catalogName === 'default'
+          ? undefined
+          : catalogName;
 
       let targetPath: string[];
       if (!normalizedCatalogName) {
         // An empty `catalog:` placeholder must not claim the default route
         // when `catalogs.default` is populated; that would create a
-        // duplicate-default config rejected by pnpm.
-        if (isMapAt(doc, ['catalog'])) {
+        // duplicate-default config rejected by pnpm. Without the alias,
+        // "default" is an ordinary named catalog and the default route is
+        // always the `catalog` field.
+        if (!aliasDefaultCatalog || isMapAt(doc, ['catalog'])) {
           targetPath = ['catalog', packageName];
         } else if (existsAt(doc, ['catalogs', 'default'])) {
           targetPath = ['catalogs', 'default', packageName];
--- a/packages/devkit/src/utils/package-json.spec.ts
+++ b/packages/devkit/src/utils/package-json.spec.ts
@@ -2,7 +2,11 @@ import * as packageManagerUtils from 'nx/src/utils/package-manager';
 import { createTree } from 'nx/src/generators/testing-utils/create-tree';
 import type { Tree } from 'nx/src/generators/tree';
 import { readJson, writeJson } from 'nx/src/generators/utils/json';
-import { addDependenciesToPackageJson, ensurePackage } from './package-json';
+import {
+  addDependenciesToPackageJson,
+  ensurePackage,
+  getDependencyVersionFromPackageJson,
+} from './package-json';
 
 describe('addDependenciesToPackageJson', () => {
   let tree: Tree;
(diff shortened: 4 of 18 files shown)


packages/vitest/src/utils/version-utils.ts
```diff
@@ -9,15 +9,24 @@ export function getInstalledViteVersion(tree: Tree): string {
...
   ) {
-    return clean(vitestVersion) ?? coerce(vitestVersion).version;
+    return fallbackVersion;
   }
 
-  return clean(installedViteVersion) ?? coerce(installedViteVersion).version;
+  // coerce() returns null for an unparseable specifier (e.g. an unresolved
+  // "catalog:" reference), so guard the `.version` access and fall back.
+  return (
+    clean(installedViteVersion) ??
```
jaysoo: These now return vite version not vitest.

Also, don't think we need to account for `catalog:` here since `getDependencyVersionFromPackageJson` already handles it once bun manager is registered.

packages/react/src/utils/version-utils.ts
```diff
@@ -72,15 +72,23 @@ export function getInstalledReactVersion(tree: Tree): string {
     'react'
   );
 
+  const fallbackVersion = clean(reactVersion) ?? coerce(reactVersion)?.version;
```
jaysoo: We shouldn't need to touch these utils now that Bun manager is registered as `catalog:` will resolve to the right version with `getDependencyVersionFromPackageJson`.

I'd revert all changes to `packages/react` and `packages/vitest`.

packages/nx/src/utils/catalog/manager-utils.ts
```diff
@@ -269,3 +271,243 @@ export function updateCatalogVersionsInFile(
     throw error;
   }
 }
+
+// Bun catalogs live in package.json, either at the top level or nested under
+// the object form of `workspaces`.
+interface BunPackageJson {
```
jaysoo: It looks like most of the Bun/JSON logic doesn't overlap with the YAML-based pnpm/yarn utils. I'd move these out to `bun-manager-utils.ts` so it's each module is scoped to a specific use-case.

packages/vitest/src/utils/version-utils.ts
cogwirrel: Good catch, and yes agreed - have removed changes to the `vitest` and `react` packages :)

packages/react/src/utils/version-utils.ts
cogwirrel: Done, thank you!

packages/nx/src/utils/catalog/manager-utils.ts
cogwirrel: Moved to a new `bun-manager-utils.ts` file :)

packages/devkit/src/utils/package-json.ts
```diff
@@ -37,8 +37,9 @@ const NON_SEMVER_TAGS = {
 /**
  * Get the resolved version of a dependency from package.json.
  *
- * Retrieves a package version and automatically resolves PNPM catalog references
- * (e.g., "catalog:default") to their actual version strings. By default, searches
+ * Retrieves a package version and automatically resolves package manager catalog
+ * references (e.g., "catalog:default") to their actual version strings. By default,
+ * searches
```
leosvelperez: nit: formatting is not great here, `searches` is alone in this line instead of being on the next one. This doesn't fail `nx format`, but it's not well-formatted.

packages/nx/src/utils/catalog/manager-factory.ts
```diff
@@ -16,6 +17,8 @@ export function getCatalogManager(
       return new PnpmCatalogManager();
     case 'yarn':
       return new YarnCatalogManager();
+    case 'bun':
+      return new BunCatalogManager();
```
leosvelperez, line 21: Flagging here because this arm brings bun into the lint rule's catalog path. The fix belongs in the rule.

`dependency-checks.ts` hardcodes pnpm's rule that `catalog:` and `catalogs.default` are the same catalog:

https://github.com/nrwl/nx/blob/3824674d4b46f3b698bf912327b898c1b8e00c40/packages/eslint-plugin/src/rules/dependency-checks.ts#L216-L236

Two consequences. A package found only in `catalogs.default` gets pushed as `catalog:` (line 222). And `catalog:default` can never be produced at all, because line 228 skips `name === 'default'` before reaching the `catalog:${name}` push on lines 231-234.

That rule holds for pnpm only. Verified against the real binaries, across versions:

| | `catalog:` from `catalogs.default` | `catalog:default` from top-level `catalog` |
|---|---|---|
| pnpm 9.15.9 / 10.34.5 / 11.16.0 | resolves | resolves |
| yarn 4.10.1 / 4.13.0 / 4.17.1 | `default catalog not found or empty` | `catalog "default" not found or empty` |
| bun 1.2.19 / 1.3.0 / 1.3.14 | failed to resolve | failed to resolve |

pnpm additionally rejects both being defined at once (`The 'default' catalog was defined multiple times`), so the comment on line 217 is accurate for pnpm. Yarn and bun both accept the coexistence and resolve the two references to different versions, which is the direct evidence they are separate catalogs. Yarn gained catalogs in 4.10.1, so 4.9.4 and earlier have none. The bun row starts at 1.2.19 because that is the first version supporting the top-level location; 1.2.14 through 1.2.18 support only `workspaces`-nested catalogs, and 1.2.14 gives the same result when tested there.

`BunCatalogManager` models this correctly (`isDefault = !catalogName`, and `catalogToUse = catalogDefs.catalog` with no `?? catalogs.default`). The lint rule does not, so once bun resolves through the factory the rule applies pnpm's semantics to bun workspaces.

For a bun workspace declaring only `catalogs.default`, running the real rule through ESLint's `Linter.verifyAndFix` autofixes a project manifest to:

```json
{ "dependencies": { "random-external": "catalog:" } }
```

So `nx lint --fix` writes a specifier that this PR's own manager rejects (`resolveCatalogReference` returns `null`, `validateCatalogReference` throws `No default catalog defined in package.json`) and that real bun refuses to install. `checkMissingDependencies` defaults to `true`, so this is the default configuration.

**Scope:** this only misfires when a catalog is literally named `default`. `catalog` -> `catalog:` and `catalogs.foo` -> `catalog:foo` are both correct.

**The same assumption is wrong for yarn**, as the table shows, which is the argument for fixing it at the interface rather than patching in a bun case. Moving default-catalog candidate selection behind `CatalogManager`, which already owns this knowledge per manager, fixes both at once. The stopgap to keep this PR small is to verify each candidate with `manager.resolveCatalogReference(...)` before adding it to `matches`.

No test covers this. `dependency-checks.spec.ts` passes with 57 tests and has no bun fixtures, and its existing `should use catalog: for missing dep when package is in catalogs.default` asserts the pnpm-correct behavior, which is what breaks for bun and yarn.

leosvelperez (commented):
Small description nit: the stated bun support window is slightly off. Checking 1.2.13 through 1.2.23, catalogs arrive in 1.2.14 but only under the object form of `workspaces`; the top-level `catalog`/`catalogs` fields start resolving in 1.2.19. So ">= 1.2.14, either at the top level or nested" holds from 1.2.19 onward. Might be worth splitting the two windows in the description.

packages/nx/src/utils/catalog/manager-factory.ts
leosvelperez: I understand this is already pre-existing with Yarn, so I'm fine with leaving it for a follow-up; we can handle it if you don't have the time. So, please let me know if that's the case.

cogwirrel: Ah good catch, thank you @leosvelperez! I'll have a go at moving it behind `CatalogManager` :)

cogwirrel: I've pushed the yarn fix as a separate commit too :)

leosvelperez (approved):
Thank you! 🙏🏻

Issue #35272, opened 2026-04-12
Non-relative paths are not allowed when 'baseUrl' is not set


JonathanWilbur:
### Current Behavior

When I try to compile a Typescript project with TypeScript 6.0, I now get this error: `error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?`. I removed `baseUrl` as a part of the upgrade, because this option is deprecated.

### Expected Behavior

This technically might be expected behavior, since TypeScript 6.0.0 is not officially [supported](https://nx.dev/docs/technologies/typescript/introduction#requirements) by Nx yet.

I am still writing it up as a bug because eventually you will want to support TypeScript 6.0.0 and this is a problem that is blocking that upgrade.

### GitHub Repo

https://github.com/Wildboar-Software/asn1-typescript-libraries/tree/823f9a76d6dd45664aa9423155d87847bcac4f94

### Steps to Reproduce

1. Install the dependencies as usual
2. Run `npx nx run x500:build`.

### Nx Report

```shell
Node           : 23.8.0
OS             : linux-x64
Native Target  : x86_64-linux
npm            : 10.9.2
daemon         : Available

nx             : 22.6.5
@nx/js         : 22.6.5
@nx/workspace  : 22.6.5
@nx/devkit     : 22.6.5
@nx/vite       : 22.6.5
@nx/vitest     : 22.6.5
typescript     : 6.0.2
---------------------------------------
Registered Plugins:
@nx/js/typescript
@nx/vitest
---------------------------------------
Cache Usage: 0.00 B / 91.47 GB
```

### Failure Logs

```shell

```

### Operating System

- [ ] macOS
- [x] Linux
- [ ] Windows
- [ ] Other (Please specify)

JonathanWilbur:
I suspect this is caused by similar code to the issue behind this bug: https://github.com/nrwl/nx/issues/35198

JonathanWilbur:
Actually, this looks like it might be fixed here: https://github.com/nrwl/nx/commit/dc479c50a58231d421ac324e5783b20214710c8d. You can close it now if you want, but I will be testing it again as soon as the next release drops.

leosvelperez:
Yes, this is fixed by https://github.com/nrwl/nx/pull/34965, which should be released soon.

PR #36445, opened 2026-07-23
fix(js): resolve tsconfig to absolute path so ${configDir} paths type…

edusperoni:
## Current Behavior

In a workspace whose tsconfigs have **no `baseUrl`** (the TypeScript 6 migration path — `baseUrl` is deprecated in TS 6 and removed in TS 7) and whose base tsconfig uses `${configDir}` in `paths`, e.g.:

```jsonc
// tsconfig.base.json (no baseUrl)
"paths": {
  "@/*": ["${configDir}/src/*"],
  "~/*": ["${configDir}/src/*"]
}
```

every `@nx/esbuild:esbuild` build fails its type check with a **location-less** error:

```
error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?
Found 1 error.
```

The error has no file/line because it comes from `program.getOptionsDiagnostics()` — it is a diagnostic about the compiler `options` (the `paths` map), not about any source file. A plain `tsc -p <tsconfig>` on the exact same config **passes**, which makes the failure look like an Nx-only problem.

### Root cause

Two things compose to produce the bug:

1. **The esbuild executor relativizes the config path.** `packages/esbuild/src/executors/esbuild/esbuild.impl.ts` → `getTypeCheckOptions()` sets `tsConfigPath: relative(process.cwd(), join(context.root, tsConfig))`, i.e. a value like `"apps/my-app/tsconfig.app.json"`. This is then handed to `runTypeCheck()`.

2. **`readTsConfig()` parses with a relative `basePath` and no `configFileName`.** `packages/js/src/utils/typescript/ts-config.ts` did:

   ```ts
   const readResult = tsModule.readConfigFile(tsConfigPath, sys.readFile);
   return tsModule.parseJsonConfigFileContent(
     readResult.config,
     sys,
     dirname(tsConfigPath) // relative basePath, e.g. "apps/my-app"
   );
   ```

   TypeScript substitutes `${configDir}` in `paths` using that `basePath`, so `${configDir}/src/*` becomes `"apps/my-app/src/*"` — a value that is **neither absolute nor `./`-relative**. `parseJsonConfigFileContent` returns **zero errors** for this. The TS5090 diagnostic only surfaces later, when `runTypeCheck()` (`packages/js/src/utils/typescript/run-type-check.ts`) builds the program and something reads `program.getOptionsDiagnostics()`/`getPreEmitDiagnostics()`.

`tsc -p <path>` does not hit this because it resolves the config path to absolute and passes the config source file, so `${configDir}` expands to an absolute directory and the resulting `paths` values are absolute (which TS5090 allows without a `baseUrl`).

Minimal isolated repro of the diagnostic:

```js
const ts = require('typescript');
const parsed = ts.parseJsonConfigFileContent(
  ts.readConfigFile(relativeCfgPath, ts.sys.readFile).config,
  ts.sys,
  path.dirname(relativeCfgPath) // relative basePath
);
// parsed.errors === []  BUT  parsed.options.paths['@/*'] === ['apps/my-app/src/*'] (non-relative!)
const program = ts.createProgram(parsed.fileNames.slice(0, 1), { ...parsed.options, noEmit: true });
program.getOptionsDiagnostics(); // → TS5090
```

### Affected versions

Reproduced against installed **nx 23.1.0** and published **@nx/js + @nx/esbuild 23.2.0-beta.2** — the relevant code (`getTypeCheckOptions`, `readTsConfig`) is identical, and the same bug is present on `master`.

### Precedent inside Nx

`packages/js/src/utils/buildable-libs-utils.ts` → `readTsConfigWithRemappedPaths()` already absolutizes all remapped `paths` values (with an explicit comment about not needing `baseUrl` because it is deprecated in TS 6), so the generated tmp tsconfig it produces is correct. The `readTsConfig` → `runTypeCheck` route used by the esbuild executor's type check was simply missed.

## Expected Behavior

Type-checking a `baseUrl`-less workspace that uses `${configDir}` in `paths` succeeds — the `@nx/esbuild:esbuild` type check behaves like `tsc -p`, with no spurious location-less TS5090.

## Fix

Make `readTsConfig()` in `packages/js/src/utils/typescript/ts-config.ts` resolve the tsconfig path to **absolute** before parsing, so the `basePath` passed to `parseJsonConfigFileContent` is absolute and `${configDir}` expands to an absolute directory (matching `tsc`). This is a central fix: it repairs **every** `runTypeCheck` consumer (the `@nx/esbuild` executor at minimum), not just one call site.

```ts
const usingDefaultSys = !sys || sys === tsModule.sys;
sys ??= tsModule.sys;

const configPath = usingDefaultSys ? resolve(tsConfigPath) : tsConfigPath;

const readResult = tsModule.readConfigFile(configPath, sys.readFile);
return tsModule.parseJsonConfigFileContent(
  readResult.config,
  sys,
  dirname(configPath)
);
```

### Two deliberate constraints on the fix

1. **Absolutization is gated to the default (real) filesystem `sys`.** `readTsConfigFromTree()` calls `readTsConfig(tsConfigPath, treeBackedSys)` with a `sys` whose `readFile` is `(path) => tree.read(path, 'utf-8')`, and `FsTree.read` normalizes via `relative(root, join(root, path))` — passing it an **absolute** path re-roots it (`join('/root', '/root/x')` → `/root/root/x`) and breaks the read. So we only absolutize when using the default `ts.sys`; any caller that supplies a custom `sys` (tree-backed or otherwise) keeps the previous relative behavior. All the affected `runTypeCheck` consumers use the default `sys`, so the bug is fixed for them without touching the Tree path.

2. **`configFileName` (the 5th arg) is intentionally *not* passed.** Only the `basePath` needs to be absolute for `${configDir}` to expand correctly. Passing `tsConfigPath` as `configFileName` would additionally set `options.configFilePath`, which changes how TypeScript infers the common source directory and **alters declaration emit output paths** — this was caught by the existing `runTypeCheck` "should support emitting declarations" spec, which started emitting to `dist/proj/…` instead of `dist/…`. Leaving `configFileName` unset keeps emit behavior unchanged while still fixing the diagnostic. (Trade-off: TS5090-class option diagnostics remain location-less, exactly as before — this PR does not attach them to the config file.)

The esbuild executor's `relative(process.cwd(), …)` wrapper in `getTypeCheckOptions()` is left as-is; with `readTsConfig` now absolutizing, the relative path it produces is handled correctly, and centralizing the fix avoids touching each consumer individually.

## Tests

Added to `packages/js/src/utils/typescript/`:

- **`ts-config.spec.ts`** — a new `describe('readTsConfig')` block: parsing a `baseUrl`-less config whose `paths` use `${configDir}`, read via a **relative** path (with `process.cwd()` set to the workspace), yields **absolute** `options.paths` values and no parse errors; and building a program from that parsed config produces **no TS5090** in `program.getOptionsDiagnostics()`.
- **`run-type-check.spec.ts`** — a new end-to-end case exercising the actual esbuild path (`runTypeCheck` → `readTsConfig`): a `baseUrl`-less base tsconfig with `${configDir}` `paths`, type-checked via a **relative** `tsConfigPath`, returns **zero errors** (before the fix it returned the location-less TS5090).

### Test runs

All runs go through the workspace's own Nx (`pnpm nx test <project>`), with the jest 30 path filter passed through via `--testPathPatterns`:

- `pnpm nx test js --testPathPatterns="typescript/(ts-config|run-type-check)"` — the new/edited `ts-config.spec.ts` and `run-type-check.spec.ts`: **2 suites, 22 tests passed**.
- `pnpm nx test esbuild` — the full `@nx/esbuild` package suite (the primary consumer of the fixed code path, exercised as a regression sweep): **7 suites, 36 tests passed**.

Reverting only the `ts-config.ts` change makes the three new assertions fail with the location-less TS5090, while the pre-existing `runTypeCheck` "should support emitting declarations" spec continues to pass — confirming the `configFileName` trade-off noted above. Edited files were formatted with the repo's Prettier.

## Related Issue(s)

Related: #35272 (same TS5090 symptom, closed as fixed by #34965 — that PR absolutized the remapped paths written by `readTsConfigWithRemappedPaths` and stopped generating `baseUrl`, but did not cover `readTsConfig` itself, which is the path `runTypeCheck` consumers like `@nx/esbuild` go through; this PR fixes that remaining case)

### DIFF
--- a/packages/js/src/utils/typescript/run-type-check.spec.ts
+++ b/packages/js/src/utils/typescript/run-type-check.spec.ts
@@ -89,6 +89,61 @@ describe('runTypeCheck', () => {
     expect(existsSync(join(cacheDir, '.tsbuildinfo'))).toBe(true);
   });
 
+  it('should type-check a baseUrl-less config using ${configDir} paths passed via a relative path', async () => {
+    // The relative config path esbuild passes here must not trip TS5090.
+    writeFileSync(
+      join(workspaceRoot, 'tsconfig.base.json'),
+      JSON.stringify(
+        {
+          compilerOptions: {
+            skipLibCheck: true,
+            strict: false,
+            module: 'esnext',
+            moduleResolution: 'bundler',
+            paths: {
+              '@/*': ['${configDir}/src/*'],
+              '~/*': ['${configDir}/src/*'],
+            },
+          },
+        },
+        null,
+        2
+      )
+    );
+    const relativeTsConfigPath = 'proj/tsconfig.json';
+    writeFileSync(
+      join(workspaceRoot, relativeTsConfigPath),
+      JSON.stringify(
+        {
+          extends: '../tsconfig.base.json',
+          include: ['src/**/*.ts'],
+        },
+        null,
+        2
+      )
+    );
+    mkdirSync(join(projectRoot, 'src'), { recursive: true });
+    writeFileSync(
+      join(projectRoot, 'src', 'valid-source.ts'),
+      `export const msg = 'Hello';`
+    );
+
+    // chdir so the relative tsConfigPath resolves from the workspace root.
+    const originalCwd = process.cwd();
+    process.chdir(workspaceRoot);
+    try {
+      const result = await runTypeCheck({
+        workspaceRoot,
+        tsConfigPath: relativeTsConfigPath,
+        mode: 'noEmit',
+      });
+
+      expect(result.errors).toHaveLength(0);
+    } finally {
+      process.chdir(originalCwd);
+    }
+  });
+
   it('should support emitting declarations', async () => {
     const outDir = join(workspaceRoot, 'dist');
     writeFileSync(
--- a/packages/js/src/utils/typescript/ts-config.spec.ts
+++ b/packages/js/src/utils/typescript/ts-config.spec.ts
@@ -1,6 +1,10 @@
 import { TempFs } from 'nx/src/internal-testing-utils/temp-fs';
-import { readTsConfigPaths, resolvePathsBaseUrl } from './ts-config';
-import { join } from 'path';
+import {
+  readTsConfig,
+  readTsConfigPaths,
+  resolvePathsBaseUrl,
+} from './ts-config';
+import { isAbsolute, join } from 'path';
 
 describe('resolvePathsBaseUrl', () => {
   let tempFs: TempFs;
@@ -314,5 +318,86 @@ describe('readTsConfigPaths', () => {
     );
 
     expect(readTsConfigPaths(join(tempFs.tempDir, 'tsconfig.json'))).toBeNull();
+  });
+});
+
+describe('readTsConfig', () => {
+  let tempFs: TempFs;
+  let originalCwd: string;
+
+  beforeEach(() => {
+    tempFs = new TempFs('read-ts-config', false);
+    originalCwd = process.cwd();
+  });
+
+  afterEach(() => {
+    process.chdir(originalCwd);
+    tempFs.cleanup();
+  });
+
+  it('should expand ${configDir} paths to absolute values when a baseUrl-less config is read via a relative path', () => {
+    // A relative basePath would leave `paths` non-relative and trip TS5090.
+    tempFs.createFileSync(
+      'tsconfig.base.json',
+      JSON.stringify({
+        compilerOptions: {
+          paths: {
+            '@/*': ['${configDir}/src/*'],
+            '~/*': ['${configDir}/src/*'],
+          },
+        },
+      })
+    );
+    tempFs.createFileSync(
+      'proj/tsconfig.json',
+      JSON.stringify({
+        extends: '../tsconfig.base.json',
+        include: ['src/**/*.ts'],
+      })
+    );
+    tempFs.createFileSync('proj/src/index.ts', `export const a = 1;`);
+
+    process.chdir(tempFs.tempDir);
+    const parsed = readTsConfig('proj/tsconfig.json');
+
+    expect(parsed.errors).toHaveLength(0);
+    for (const values of Object.values(parsed.options.paths ?? {})) {
+      for (const value of values) {
+        expect(isAbsolute(value)).toBe(true);
+      }
+    }
+  });
+
+  it('should not report option diagnostics for a baseUrl-less ${configDir} config read via a relative path', () => {
+    const ts = require('typescript') as typeof import('typescript');
+    tempFs.createFileSync(
+      'tsconfig.base.json',
+      JSON.stringify({
+        compilerOptions: {
+          skipLibCheck: true,
+          module: 'esnext',
+          moduleResolution: 'bundler',
+          paths: { '@/*': ['${configDir}/src/*'] },
+        },
+      })
+    );
+    tempFs.createFileSync(
+      'proj/tsconfig.json',
+      JSON.stringify({
+        extends: '../tsconfig.base.json',
+        include: ['src/**/*.ts'],
+      })
+    );
+    tempFs.createFileSync('proj/src/index.ts', `export const a = 1;`);
+
+    process.chdir(tempFs.tempDir);
+    const parsed = readTsConfig('proj/tsconfig.json');
+    const program = ts.createProgram(parsed.fileNames, {
+      ...parsed.options,
+      noEmit: true,
+    });
+
+    const optionDiagnostics = program.getOptionsDiagnostics();
+    expect(optionDiagnostics.map((d) => d.code)).not.toContain(5090);
   });
 });
--- a/packages/js/src/utils/typescript/ts-config.spec.ts
+++ b/packages/js/src/utils/typescript/ts-config.spec.ts
@@ -9,11 +9,12 @@ import {
 import { tmpdir } from 'os';
 import {
   createTreeParseConfigHost,
+  readTsConfig,
   readTsConfigPaths,
   resolvePathsBaseUrl,
   type TreeParseConfigHost,
 } from './ts-config';
-import { join } from 'path';
+import { isAbsolute, join } from 'path';
 import * as ts from 'typescript';
 import { TempFs } from '@nx/devkit/internal-testing-utils';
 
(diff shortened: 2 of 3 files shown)


packages/js/src/utils/typescript/ts-config.ts
```diff
@@ -20,13 +20,36 @@ export function readTsConfig(
...
+  // but later surfaces as a location-less
+  // `TS5090: Non-relative paths are not allowed when 'baseUrl' is not set` from
+  // `program.getOptionsDiagnostics()` — the failure mode for TS6 workspaces
+  // that drop the deprecated `baseUrl` and use `${configDir}` in `paths` (e.g.
+  // when type-checking via `@nx/esbuild`, which passes a relative config path).
+  //
+  // Note: only the `basePath` is absolutized. Passing `tsConfigPath` as the
+  // `configFileName` (5th arg) would additionally set `options.configFilePath`,
+  // which changes how TS infers the common source directory and would alter
+  // declaration emit output paths — so it is intentionally left unset.
```
leosvelperez:
```suggestion
  // Resolve to absolute so `${configDir}` in `paths` expands to absolute
  // values. A relative basePath produces the TS5090 "no baseUrl" error.
  //
  // Don't also pass `configPath` as `configFileName` (5th arg). It doesn't
  // affect `${configDir}`, only shifts emit output paths (dist/src vs dist).
```

packages/js/src/utils/typescript/ts-config.ts
```diff
@@ -20,13 +20,36 @@ export function readTsConfig(
     tsModule = require('typescript');
   }
 
+  // A caller may pass a custom `sys` whose `readFile` expects workspace-relative
+  // paths (e.g. the tree-backed system in `readTsConfigFromTree`, where
+  // `FsTree.read` re-roots the path and breaks on absolute input). Only the
+  // default real filesystem is safe to absolutize against.
```
leosvelperez:
```suggestion
  // A tree-backed `sys` (readTsConfigFromTree) reads via `FsTree.read`, which
  // re-roots an absolute path and breaks, so only absolutize the default sys.
```

packages/js/src/utils/typescript/run-type-check.spec.ts
```diff
@@ -89,6 +89,70 @@ describe('runTypeCheck', () => {
     expect(existsSync(join(cacheDir, '.tsbuildinfo'))).toBe(true);
   });
 
+  it('should type-check a baseUrl-less config using ${configDir} paths passed via a relative path', async () => {
+    // Reproduces the TS6-migration failure: a workspace with NO `baseUrl`
+    // (deprecated in TS6) whose base tsconfig uses `${configDir}` in `paths`.
+    // The esbuild executor relativizes the config path before type-checking, so
+    // `readTsConfig` receives a relative path. If `${configDir}` is substituted
+    // against a relative basePath, `paths` values become non-relative, which
+    // later triggers a location-less `TS5090` from `getOptionsDiagnostics()`.
```
leosvelperez:
```suggestion
    // The relative config path esbuild passes here must not trip TS5090.
```

packages/js/src/utils/typescript/run-type-check.spec.ts
```diff
@@ -89,6 +89,70 @@ describe('runTypeCheck', () => {
...
+      )
+    );
+    mkdirSync(join(projectRoot, 'src'), { recursive: true });
+    writeFileSync(
+      join(projectRoot, 'src', 'valid-source.ts'),
+      `export const msg = 'Hello';`
+    );
+
+    // The executor passes a path relative to the current working directory, so
+    // exercise that here.
```
leosvelperez:
```suggestion
    // chdir so the relative tsConfigPath resolves from the workspace root.
```

packages/js/src/utils/typescript/ts-config.spec.ts
```diff
@@ -316,3 +320,91 @@ describe('readTsConfigPaths', () => {
...
+  afterEach(() => {
+    process.chdir(originalCwd);
+    tempFs.cleanup();
+  });
+
+  it('should expand ${configDir} paths to absolute values when a baseUrl-less config is read via a relative path', () => {
+    // TS6-migration shape: no `baseUrl` (deprecated in TS6), `${configDir}` in
+    // `paths`. When the config is read via a relative path (as the esbuild
+    // executor does), `${configDir}` must still resolve against an absolute
+    // directory — otherwise `paths` end up non-relative and trigger TS5090.
```
leosvelperez:
```suggestion
    // A relative basePath would leave `paths` non-relative and trip TS5090.
```

packages/js/src/utils/typescript/run-type-check.spec.ts
```diff
@@ -89,6 +89,70 @@ describe('runTypeCheck', () => {
...
+    writeFileSync(
+      join(workspaceRoot, 'tsconfig.base.json'),
+      JSON.stringify(
+        {
+          compilerOptions: {
+            skipLibCheck: true,
+            strict: false,
+            module: 'esnext',
+            moduleResolution: 'bundler',
+            // No `baseUrl` — paths resolve from `${configDir}`.
```
leosvelperez: Let's drop it. The test name and preamble already say baseUrl-less, and the `${configDir}` paths are right below.
```suggestion
```

packages/js/src/utils/typescript/run-type-check.spec.ts
```diff
@@ -89,6 +89,70 @@ describe('runTypeCheck', () => {
...
+    process.chdir(workspaceRoot);
+    try {
+      const result = await runTypeCheck({
+        workspaceRoot,
+        tsConfigPath: relativeTsConfigPath,
+        mode: 'noEmit',
+      });
+
+      // Before the fix this contained the location-less
+      // `TS5090: Non-relative paths are not allowed when 'baseUrl' is not set`.
```
leosvelperez: Let's drop it. "Before the fix" has no referent once merged, and the TS5090 is already named in the preamble.
```suggestion
```

packages/js/src/utils/typescript/ts-config.spec.ts
```diff
@@ -316,3 +320,91 @@ describe('readTsConfigPaths', () => {
...
+        include: ['src/**/*.ts'],
+      })
+    );
+    tempFs.createFileSync('proj/src/index.ts', `export const a = 1;`);
+
+    process.chdir(tempFs.tempDir);
+    const parsed = readTsConfig('proj/tsconfig.json');
+
+    // No parse errors, and every remapped path value is absolute (not the
+    // non-relative "proj/src/*" that TS5090 rejects).
```
leosvelperez: Let's drop it. Same point as the preamble above (non-relative paths -> TS5090). The `isAbsolute` loop reads fine on its own.
```suggestion
```

packages/js/src/utils/typescript/ts-config.spec.ts
```diff
@@ -316,3 +320,91 @@ describe('readTsConfigPaths', () => {
...
+    process.chdir(tempFs.tempDir);
+    const parsed = readTsConfig('proj/tsconfig.json');
+    const program = ts.createProgram(parsed.fileNames, {
+      ...parsed.options,
+      noEmit: true,
+    });
+
+    const optionDiagnostics = program.getOptionsDiagnostics();
+    // Before the fix this contained TS5090 ("Non-relative paths are not allowed
+    // when 'baseUrl' is not set").
```
leosvelperez: Let's drop it. "Before the fix" has no referent once merged, and the assertion already checks 5090 directly.
```suggestion
```

leosvelperez (commented):
Thank you!

The change looks good, but let's simplify the code comments. Most restate the code or the test names, and the `readTsConfig` block narrates the bug the PR description already covers. The inline suggestions keep just the non-obvious why.

edusperoni:
@leosvelperez  Thank you for the review! I've adjusted the comments.

leosvelperez (approved):
Thanks!

PR #36449, opened 2026-07-23
chore(repo): add sigint sigsegv hunt diagnostic workflow

AgentEnder:
> **Diagnostic PR — not intended to merge.** It exists to run the `SIGINT SIGSEGV Hunt` workflow on CI infrastructure and force a reproduction of the intermittent native segfault with core dumps enabled.

## Current Behavior

The `SIGINT process cleanup` e2e tests (`e2e/nx/src/run.test.ts`) spawn nx inside a real pseudo-terminal with the native TUI forced on (`NX_TUI=true` + `NX_TUI_SKIP_CAPABILITY_CHECK`), send SIGINT, and assert a clean exit. In CI they intermittently die with SIGSEGV instead (`[sigint-test] onExit: Exited with code 139`) — 12 failed jobs between 2026-07-13 and 2026-07-23, 3 on master. This is currently the most frequently-reproducing shape of the intermittent CI segfault, but like the `check-imports` fingerprint (#36448) it has never left a native stack.

## Expected Behavior

A `workflow_dispatch`/branch-push workflow (`.github/workflows/sigint-segv-hunt.yml`) loops exactly those tests, 3 shards × N iterations (default 20), with `kernel.core_pattern` repointed away from apport and `ulimit -c unlimited`, so a reproduction yields a core. Each shard reports `RESULT shard=N: X/Y iterations had signal deaths` in its step summary and uploads cores + gdb backtraces + per-iteration logs as `sigint-segv-hunt-shard-*` artifacts. Shards are report-only (always green). Detection greps are case-sensitive — case-insensitive matching hits lowercase branch/commit metadata echoed by Nx Cloud (learned in #36303 round 1).

If the backtrace here matches whatever the #36448 catcher eventually captures for the `check-imports` fingerprint, we have one root cause and a cheap local repro; if not, we have two well-separated bugs.

## Related Issue(s)

Companion to #36448 (core catcher). Prior hunt: #36303.

### DIFF
--- a/.github/workflows/sigint-segv-hunt.yml
+++ b/.github/workflows/sigint-segv-hunt.yml
@@ -1,0 +1,147 @@
+name: SIGINT SIGSEGV Hunt
+
+# Diagnostic workflow — loops the `SIGINT process cleanup` e2e tests
+# (e2e/nx/src/run.test.ts) which intermittently die with SIGSEGV (exit 139)
+# in CI (12 failed jobs between 2026-07-13 and 2026-07-23, 3 on master),
+# trying to force a reproduction with core dumps enabled so we finally get
+# a native backtrace for the intermittent CI segfaults.
+#
+# Report-only: shards never go red on a crash — read the step summaries
+# ("RESULT shard=N: X/Y iterations had signal deaths") and download the
+# sigint-segv-hunt-shard-* artifacts for cores + gdb backtraces.
+
+on:
+  workflow_dispatch:
+    inputs:
+      iterations:
+        description: 'Iterations per shard'
+        required: false
+        default: '20'
+  push:
+    branches:
+      - ci-sigint-segv-hunt
+
+env:
+  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+  PNPM_HOME: ~/.pnpm
+  # Pin corepack to the pnpm version from packageManager. Without this, corepack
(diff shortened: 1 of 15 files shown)


AgentEnder:
Closing this. The segfault stopped reproducing once the repo's own nx picked up #36673.

The crash was in the nx running the e2e tasks, not the nx under test. #36673 landed 2026-08-19, shipped in 23.2.0-beta.8, and reached this repo when #36751 bumped us from beta.7 to beta.9 on 08-21. That's why the 08-19 hunt still crashed with the fix already in its tree.

| hunt run    | installed nx | signal deaths |
| ----------- | ------------ | ------------- |
| 08-13 19:41 | beta.7       | 0/60          |
| 08-13 20:30 | beta.7       | 0/60          |
| 08-14 18:19 | beta.7       | 1/60          |
| 08-14 18:29 | beta.7       | 1/60          |
| 08-19 17:17 | beta.7       | 2/60          |
| 08-24 21:47 | beta.9       | 0/60          |
| 08-27 02:21 | beta.9       | 0/60          |
| 08-27 17:39 | beta.9       | 0/60          |
| 08-27 20:34 | beta.9       | 0/60          |

4 signal deaths in 300 iterations before, 0 in 240 after. At the old rate a clean 240 lands about 4% of the time.

The 2026-07-23 run reading 60/60 is not in the table. That was the log grep matching this workflow's own commit message, fixed in da2bf811a by detecting exit code 139 and new core files instead.

Not merging the workflow. The branch keeps it if this comes back.

PR #36465, opened 2026-07-25
fix(core): prune git linked worktrees from workspace walks

AgentEnder:
## Current Behavior

A linked worktree created inside the workspace is a full second checkout of the repository, and Nx walks it like any other directory. `git worktree add ./wt/feature`, or the `.claude/worktrees` directory agent tooling uses, means the project graph hashes the entire workspace twice and the daemon registers watches for every directory in the worktree.

Nothing filters these today. A worktree is not ignored by git — `git check-ignore` reports it as untracked, not ignored — so it does not fall out of the existing `.gitignore` handling unless someone adds it by hand.

## Expected Behavior

Linked worktrees nested inside the workspace are pruned from the walk, and their events are kept out of the daemon.

Git already records every linked worktree under `<git-dir>/worktrees/<name>`, so the roots are read from there rather than probed for on disk — no subprocess and no new dependency. The list is revalidated against that directory's mtime, which git touches whenever a worktree is added or removed: a worktree created while the daemon runs is still picked up on the next graph construction, while a walk that finds nothing costs one `stat` (1.79µs, against 7.06µs resolving from scratch).

**Walking** — `walker.rs`

`create_walker` and `nx_walker_sync` both prune worktree roots. The second matters more than it looks: the daemon expands a directory-creation event by walking the new directory, and backfills a newly watched directory the same way, so without it a `git worktree add` reports every file of the new checkout as created. This applies whatever `use_ignores` says — that flag does not mean "not the workspace", and the daemon's output watcher passes it false while walking the workspace root. Cache output expansion passes it false too and pays the revalidation for nothing, which is now one `stat`.

**Watching** — `watcher.rs`, `watch_filterer.rs`

The filterer blocks everything under a known worktree root. That is the only defence on macOS, where the root is watched recursively and the walker's prune never gets a say. Roots reach it three ways, because no one of them is sufficient:

- seeded at boot from the registry;
- from the event stream, when a checkout's gitfile is created under an existing watch;
- checked once more at the moment a watch would be registered, because `git worktree add` makes the directory first and registers it a moment later, so a walk can land in between and see something that is not yet a worktree.

`track_new_worktrees` runs ahead of the filterer — it has to, or the gitfile would be discarded as a `.git` path before it could be read — so it does no filesystem I/O on a `.git` it has not already established is a file created just now. Answering a `.git` event by opening a `.git` publishes another one, and that loop produced ~1.9M events per run.

**Cache siting** — `worktree.rs`

`get_main_worktree_root` uses the shared gitfile helpers instead of shelling out to `git rev-parse --git-common-dir`. Two bugs fall out of the consolidation: a submodule was read as a worktree and resolved its cache and workspace-data database into `<superproject>/.git/modules/…`, and the memoized root was a `OnceLock` that returned the first caller's answer to every later call regardless of its argument.

`common_git_dir` no longer fails open. It returned the git dir it was handed when `commondir` was missing or unparseable, which named `<repo>/.git/worktrees` as a repository root — a path `git worktree prune`, and so `git gc`, deletes.

### Deliberately handled

- **Submodules keep being scanned.** They use the very same gitfile mechanism (`.git` is a regular file containing `gitdir: …`, identical in shape to a worktree's), but they are registered under `<git-dir>/modules/`. Reading the worktree registry rather than sniffing `.git` files means submodule contents are never mistaken for a worktree — their files are real workspace sources.
- **Stale registrations do not prune real directories.** A registration outlives a worktree deleted by hand, and `gc.worktreePruneExpire` defaults to three months. Pruning a path that no longer exists is harmless, but an ordinary directory later created there must stay in the graph, so the worktree's gitfile has to still be present for the path to be pruned.

### Testing

Fixtures build the layouts git builds, so the suite does not need a `git` binary — with one exception. `a_real_git_worktree_added_after_boot_is_blocked` drives real `git worktree add`, because no fixture reproduces git's ordering (directory, then registration, then gitfile, then checkout), and that ordering is what the last of the three defences above exists for. It skips itself when `git` is unavailable.

`watch_filterer.rs` has unit tests on the filterer itself rather than through a watcher, so the macOS-only half has a witness that fails on Linux, which is the only platform where CI runs cargo.

## Related Issue(s)

N/A — found while looking at directories Nx should never scan or watch.

Fixes NXC-4730

### DIFF
--- a/packages/nx/src/native/utils/git.rs
+++ b/packages/nx/src/native/utils/git.rs
@@ -116,6 +116,25 @@ fn common_git_dir(git_dir: &Path) -> PathBuf {
     }
 }
 
+/// Whether `path` is the root of a git linked worktree.
+///
+/// A worktree's gitfile points at `<git-dir>/worktrees/<name>`, while a
+/// submodule's points at `<git-dir>/modules/<name>` - the parent segment is
+/// what separates them. Costs one read, so it suits deciding about a single
+/// directory (a watch event) rather than scanning a whole workspace; use
+/// [`nested_linked_worktrees`] for that.
+pub fn is_linked_worktree_root<P: AsRef<Path>>(path: P) -> bool {
+    let path = path.as_ref();
+    let Some(git_dir) = read_gitfile(&path.join(".git"), path) else {
+        return false;
+    };
+
+    git_dir
+        .parent()
+        .and_then(Path::file_name)
+        .is_some_and(|segment| segment == "worktrees")
+}
+
 /// Roots of git's linked worktrees (`git worktree add`) that live inside
 /// `workspace_root`, relative to it.
 ///
@@ -192,6 +211,29 @@ mod test {
             format!("gitdir: {}\n", metadata_dir.display()),
         )
         .unwrap();
+    }
+
+    #[test]
+    fn identifies_linked_worktree_roots() {
+        let temp = TempDir::new().unwrap();
+        create_dir_all(temp.path().join(".git")).unwrap();
+
+        let worktree = temp.path().join("wt");
+        register_worktree(temp.path(), "wt", &worktree);
+        assert!(is_linked_worktree_root(&worktree));
+
+        // A submodule's gitfile is identical in shape - only the segment it
+        // points into differs.
+        let submodule = temp.path().join("libs/sub");
+        create_dir_all(&submodule).unwrap();
+        write(
+            submodule.join(".git"),
+            "gitdir: ../../.git/modules/libs/sub\n",
+        )
+        .unwrap();
+        assert!(!is_linked_worktree_root(&submodule));
+
+        assert!(!is_linked_worktree_root(temp.path().join("libs")));
     }
 
     #[test]
--- a/packages/nx/src/native/utils/git.rs
+++ b/packages/nx/src/native/utils/git.rs
@@ -1,4 +1,9 @@
+use std::fs::read_to_string;
 use std::path::{Path, PathBuf};
+
+/// Prefix of a "gitfile" - the plain file git writes in place of a `.git`
+/// directory for linked worktrees and submodules.
+const GITDIR_PREFIX: &str = "gitdir:";
 
 /// Find the nearest git repository root by walking up the directory tree
 fn find_git_root<P: AsRef<Path>>(start_path: P) -> Option<PathBuf> {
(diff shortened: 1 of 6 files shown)


packages/nx/src/native/utils/git.rs
FrozenPandaz, line 1: We should have some existing worktree utils somewhere. See if those can be reused or if this can be used there.

FrozenPandaz (changes requested):

packages/nx/src/native/utils/git.rs
AgentEnder: There was one — `native/worktree.rs`'s `get_main_worktree_root`, which shelled out to `git rev-parse --git-common-dir`. It's on the shared gitfile helpers as of 8501d0c, so there's one implementation and resolving a cache directory no longer needs a git binary on PATH.

Two bugs fell out of consolidating it: a submodule also replaces `.git` with a gitfile, so it was read as a worktree and resolved to a "main root" of `<superproject>/.git/modules/<..>` — putting a submodule workspace's cache and workspace-data inside the superproject's git directory. And the memoized root was a `OnceLock` seeded by the first caller's `workspace_root` and returned for every later call regardless of argument; it's keyed by workspace root now.

AgentEnder:
The self-healing fix on this PR should be rejected — it reorders the `register_worktree` test helper on the premise that `git worktree add` writes metadata before the gitfile. It's the other way round: `gitdir` → the checkout's `.git` → `commondir`.

Measured by mtime against git 2.44:

```
1786467381.403310611  .git/worktrees/wt/gitdir      <- 1st
1786467381.403395862  wt/.git                       <- 2nd  (what the watcher acts on)
1786467381.403461529  .git/worktrees/wt/commondir   <- 3rd
```

The fixture was right and the test was catching a real race. `is_linked_worktree_root` keyed on `commondir`, which lands ~66us *after* the gitfile the watcher acts on, so Linux (inotify) lost the race and macOS (FSEvents) won it — hence the platform split rather than an outright failure. Applying the fix would have made the fixture diverge from git and shipped the bug, in exactly the workflow this PR exists for: `git worktree add` against a running daemon.

Fixed in 7b4f670 by keying on `gitdir`, which git writes first. It separates a worktree from a submodule just as well — a submodule's `.git/modules/<path>` holds neither `gitdir` nor `commondir`, recording its checkout as `core.worktree` in `config` — and answers the dangling-worktree case identically.

That fix also made a second path reachable, so it is closed here too: `common_git_dir` failed open, returning the git dir it was handed when no `commondir` was there. For a half-registered worktree that named `<main>/.git/worktrees` as the repo root, which is where `resolve_main_worktree_root` would site the cache and the workspace-data database — a path `git worktree prune`, and so `git gc`, deletes. It returns `None` now.

AgentEnder:
## Status: `nx watch` e2e regression, bisected to two commits on this branch

Flagging a regression on this branch that is **not** flake and **not** environmental, so it doesn't get lost behind the self-healing reruns.

### `e2e-nx:e2e-ci--src/watch.test.ts` fails here and passes elsewhere

`nx watch` emits nothing — every assertion receives `[]` or `undefined`:

```
● Nx Watch › should watch for project changes
    - Array [ "proj11951784" ]
    + Array []
```

Two unrelated PRs from the same window ran the suite on a **Cache Miss** and passed, so it is not the environment:

| Branch | watch.test.ts | Duration |
| --- | --- | --- |
| `migrate-nx-23.2.0-beta.6` | ✔ Completed | 3m29s |
| `fix/nxc-4659-shell-injection-master` | ✔ Completed | 2m52s |
| this branch | ✖ Failed | 2m18s – 10m2s |

Note that most `master` runs **Remote Cache Hit** this task at 0s, so a green master says nothing about it. The comparisons above are all verified cache misses.

### Bisected to two commits, from this branch's own CI history

| Commit | watch.test.ts |
| --- | --- |
| `8501d0c` resolve the main worktree root without shelling out to git | ✔ **Cache Miss, 5m28s — passes** |
| `ab76cda` tell a worktree from a submodule by commondir, not by a path segment | not run alone |
| `aa42ab0` prune worktrees for the output watcher and for worktrees added later | ✖ fails |

So the regression is in `ab76cda` or `aa42ab0`. (An earlier "success" on the branch was a cache hit and is not evidence.)

### Ruled out

- **The native watcher.** Probed with `use_ignore: true`, a real `.git`, and nested project dirs — events arrive correctly. Also probed with the workspace root *itself* being a linked worktree, and with a worktree nested inside the workspace: events arrive in both.
- **The native walker.** Returns the expected file set when the workspace root is a linked worktree.
- **Worktree code misfiring in e2e.** e2e workspaces contain no worktrees, and no worktree-related log lines appear anywhere in the e2e daemon output (native debug logging is live in that suite via `NX_NATIVE_LOGGING`).
- **Staleness.** Still fails after rebasing onto current master.

The daemon logs `Updated workspace context based on watched changes, recomputing project graph...`, so changes *are* detected. The break is downstream of detection, in the mapping from changed files to projects — which is why the output is empty rather than wrong.

### Unrelated to the above: the Rust unit failure is fixed

`a_worktree_created_after_boot_is_blocked` was failing on Linux and passing on macOS. That was a real race, not flake — `is_linked_worktree_root` keyed on `commondir`, which `git worktree add` writes *after* the checkout's gitfile that the watcher acts on (verified by mtime against git 2.44: `gitdir` → `.git` → `commondir`). Fixed by keying on `gitdir`, which git writes first. A second leak went through `nx_walker_sync`, which had no worktree prune, so a directory-creation event expanded into every file of a new checkout.

Rust suite is green in CI: **621 passed, 0 failed**.

### Not done

The mechanism behind the `nx watch` failure. Next step is bisecting `ab76cda` vs `aa42ab0`, or running `watch.test.ts` locally with an instrumented daemon.

AgentEnder:
## Resolved: the `nx watch` regression was a self-sustaining event storm

Following up on my earlier comment, which reported the bisect but not the mechanism. Both are settled now, and the fix is pushed in c68dc11.

### Root cause

`track_new_worktrees` answers a `.git` event by *opening* a `.git`. On Linux that open is an access on a watched path, so it publishes another `.git` event, which it answers the same way.

Nothing broke the cycle. The function runs ahead of the filterer deliberately — it has to, or the gitfile would be discarded as a `.git` path before it could be read — so the hardcoded `.git` ignore never saw these events.

Measured in an e2e workspace: **~1.93M invocations in a single run**, every one for the workspace's own `.git`, and **not one** ever identified a worktree. The main repository's `.git` is a directory, so `is_linked_worktree_root` answered false each time while still paying the read that published the next event. Real changes starved behind the storm, so `nx watch` emitted nothing at all — hence `Received: []` on every assertion, and why the failing subset varied with how far behind the consumer was.

### Fix

Two guards ahead of the read, each closing a different loop:

- **`meta_is_dir`** — only a file can be a gitfile. The event already carries the stat, and `fs::metadata` publishes nothing (unlike a read), so the main repository's `.git` directory is skipped without touching the disk. This closes the ~2M loop.
- **`EventKind::Create(_)`** — a worktree only *appears* on a create. Without this, the modify and access events on a real worktree's own gitfile reopen it and spin the same loop one directory over — which the first guard cannot catch, because that gitfile really is a file.

### Evidence

Bisected against `e2e-nx:e2e-ci--src/watch.test.ts` in a Linux container, swapping only the native sources and holding everything else constant:

| Native sources | Result |
| --- | --- |
| master (PR reverted) | 8/8 pass |
| `ab76cda` — commondir discriminator | 8/8 pass |
| `aa42ab0` — walker half only | 8/8 pass |
| `aa42ab0` — with the single `track_new_worktrees` call removed | 8/8 pass |
| full branch | **5 fail / 3 pass** |
| branch + these guards | **8/8 pass** |

Rust unit suite: 591 passed, 0 failed.

Note for anyone comparing against `master`: most master runs **Remote Cache Hit** this task at 0s, so a green master does not mean it ran. The comparisons above are all verified cache misses.

### Unrelated, but worth knowing

`assertLocalRegistryIsRunning` in `scripts/local-registry/populate-storage.js` does a bare `await fetch(url)` with no timeout inside a `while (true)` retry loop. If the first request is issued while verdaccio is still coming up, that fetch can hang forever — the loop never retries and never logs. It cost 83 minutes of silent hang while reproducing this. An `AbortSignal.timeout` on that fetch would make the retry loop actually work.

AgentEnder:
Superseded by #36734.

This grew to 20 commits and ~1470 lines of Rust pruning worktrees from every workspace walk *and* from the daemon's file watcher. Two review rounds kept surfacing edge cases — event-ordering races, registry-cache invalidation, a self-feeding `.git` event loop — and after discussing it we decided that complexity isn't warranted for a situation users can avoid by gitignoring the directory.

#36734 takes the other route: detect the case where it actually surfaces and say what to do about it. Duplicate project names whose roots sit inside a linked worktree now name the path to gitignore instead of advising a rename, which is the wrong fix for a checkout that shouldn't be in the graph at all. The check runs only on the way to throwing, so workspaces without duplicates pay nothing.

Thanks @FrozenPandaz for the two rounds of review here — the findings were real, and the tracing in the second round is what made it clear how much surface this approach needed to cover.

Leaving the branch in place rather than deleting it, in case the pruning is wanted later.

PR #36469, opened 2026-07-26
feat(dotnet): per-package affected and cache granularity for Central Package Management

mikehaas763:
## Current Behavior

With Central Package Management, the dotnet plugin adds the whole `Directory.Packages.props` file as an input on every project's cacheable targets. Bumping a single package version invalidates the cache for every project, and `nx affected` selects everything under the manifest, including projects that don't reference the package at all.

## Expected Behavior

Bumping a package version only affects the projects that actually reference that package.

To make that possible this adds a `createTouchedDependencies` hook to core. It's a `[filePattern, fn]` tuple like `createNodes`. When a matching file changes, `nx affected` hands the hook the file contents at the base and head revisions, and the hook returns which external dependencies changed (or `'*'` if it can't tell). A new locator maps those onto external nodes and the usual reverse traversal picks up the consuming projects. There's nothing dotnet-specific in core, and plugins that don't implement the hook behave exactly as before.

The dotnet plugin then uses the hook for CPM. The MSBuild analyzer already evaluates project files, so it joins PackageReference items with their PackageVersion entries (VersionOverride wins) and emits a `nuget:<Id>@<Version>` external node for each resolved package, with edges from the projects that reference it. Cacheable targets drop the whole-file input in favor of a per-package `externalDependencies` input, and the hook diffs the PackageVersion items between the two revisions.

Anything that can't be attributed cleanly falls back to the old behavior: malformed or deleted manifests, versions supplied through MSBuild properties, unreadable base revisions, or a version the analyzer can't resolve. In those cases rebuilding too much beats a stale cache hit.

Tested with unit tests on the core locator, the plugin, and the analyzer, plus an e2e suite: in a four-project CPM workspace, bumping a package referenced by one project rebuilds only that project, comment-only manifest edits select nothing, and a project with no package references is never invalidated by version bumps.

## Related Issue(s)

N/A

### DIFF
--- a/astro-docs/src/content/docs/kb/project-graph-plugins.mdoc
+++ b/astro-docs/src/content/docs/kb/project-graph-plugins.mdoc
@@ -315,6 +315,40 @@ Breaking down this example, we can see that it follows this flow:
 7. Pushes it into the located dependency array
 8. Returns the located dependencies
 
+## Attributing manifest changes to specific dependencies
+
+A centralized dependency manifest, such as a lock file or a .NET `Directory.Packages.props`, lists versions for many packages in one file.
+By default, a change to a file like this marks every project that declares it as an input as affected, even when only one package version moved.
+
+Plugins that add external nodes can narrow this with `createTouchedDependencies`.
+Like `createNodes`, it's a tuple of a file pattern and a function, exported from the plugin's entry point:
+
+```typescript
+export const createTouchedDependencies: CreateTouchedDependencies = [
+  '**/Directory.Packages.props',
+  (touchedFiles, options, context) => {
+    const changed = new Set<string>();
+    for (const { baseContent, headContent } of touchedFiles) {
+      if (baseContent === null || headContent === null) {
+        // The manifest was added, deleted, or unreadable.
+        return '*';
+      }
+      diffVersions(baseContent, headContent).forEach((id) => changed.add(id));
+    }
+    return Array.from(changed);
+  },
+];
+```
+
+When `nx affected` runs and a changed file matches the pattern, Nx calls the function with the file's contents at the base and head revisions.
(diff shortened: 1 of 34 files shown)


mikehaas763:
@MaxKless Do you have time to take an initial look at this? The most fundamental change that I could see needing some discussion is the addition of `getTouchedDependencies` to core / project-graph.

I'm happy to answer any questions you may have!

AgentEnder:
Thanks for this, and apologies it's taken us so long to come back to it.

We're going to close this one, and it isn't about the implementation. The blocker is `createTouchedDependencies`. It's a new public plugin extension point for how `nx affected` computes touched projects, and that's a design decision the core team hasn't made yet.

You landed on the right shape, for what it's worth. `TouchedProjectLocator` already exists in core, and `getTouchedProjectsFromLockFile` / `getTouchedNpmPackages` do exactly this for npm. They're just wired into a hardcoded array in `affected-project-graph.ts` whose comment reads "Additional affected logic should be in this array", so today the JS plugin has a hook no other plugin can reach. Opening that up is something we want to do deliberately and once, with agreement on the contract, rather than as part of a .NET feature. That conversation is still open on our side and I can't give you a timeline for it. None of that was visible from outside, and you shouldn't have had to guess at it.

One thing that might be worth pursuing separately: the caching half of this may not need the hook at all. If the analyzer emits `nuget:<Id>@<Version>` external nodes and cacheable targets take per-package `externalDependencies` inputs instead of the whole `Directory.Packages.props`, a version bump should stop invalidating unrelated projects' caches on its own. `nx affected` would still over-select until the hook question is settled, but that's the smaller half of the pain. If you want to try that as its own PR, I'd review it.

In the meantime, #36527 is the one I'd most like to get over the line. I've just approved CI on it, and your changes there addressed everything from the last round.

PR #36472, opened 2026-07-27
fix(core): quote args unless they parse as one shell word

webpro:
## Current Behavior

`isAlreadyQuoted` accepts any value with matching first and last quotes. Both argument forwarding in `run-commands` and task override serialization use it to skip quoting, and the result is concatenated into a shell command string.

A single forwarded argument can become two arguments:

```text
forwarded argument   --opt="a" "b"
Nx emits             --opt="a" "b"
child receives       ["--opt=a", "b"]
```

It can also execute additional commands:

```text
forwarded argument   --opt="x" && echo INJECTED ; echo "y"
Nx emits             --opt="x" && echo INJECTED ; echo "y"
child receives       ["--opt=x"]
extra shell output   INJECTED
                     y
```

## Expected Behavior

Parse quoted-looking values with [unbash](https://github.com/webpro-nl/unbash) and pass them through only when they occupy one command word, with no extra statements, arguments, assignments, or redirects.

Both examples now reach the child as a single argument, and neither extra `echo` command runs:

```text
Nx emits         --opt="\"a\" \"b\""
child receives   ["--opt=\"a\" \"b\""]

Nx emits         --opt="\"x\" && echo INJECTED ; echo \"y\""
child receives   ["--opt=\"x\" && echo INJECTED ; echo \"y\""]
```

Existing expansion behavior inside double quotes is preserved.

This adds `unbash@^4.0.11`, a synchronous Bash parser with no runtime dependencies. I maintain unbash.

Verification: all 165 tests across `shell-quoting`, `serialize-overrides-into-command-line`, and `run-commands` pass. Package type checking, formatting, lockfile lint, and `pnpm install --frozen-lockfile --ignore-scripts` pass. Oxlint passes with module-boundary checks skipped because the workspace graph is unavailable. The two examples above were also checked against master using Bash 5.3.20. `nx prepush` cannot construct the workspace graph locally because Java and .NET runtimes are unavailable.

## Related Issue(s)

Extends the argument quoting work in https://github.com/nrwl/nx/pull/34491.

### DIFF
--- a/packages/nx/jest.config.cts
+++ b/packages/nx/jest.config.cts
@@ -5,6 +5,8 @@ module.exports = {
   displayName: 'nx',
   preset: '../../jest.preset.js',
   resolver: './jest-resolver.js',
+  // unbash is ESM-only, so let swc transform it rather than ignoring it
+  transformIgnorePatterns: ['node_modules/(?!(\\.pnpm/)?unbash)'],
   // Ensure cargo insta snapshots do not get picked up by jest
   testPathIgnorePatterns: ['<rootDir>/src/native/tui'],
 };
--- a/packages/nx/package.json
+++ b/packages/nx/package.json
@@ -80,7 +80,7 @@
     "tmp": "catalog:",
     "tsconfig-paths": "catalog:typescript",
     "tslib": "catalog:typescript",
-    "unbash": "^4.0.3",
+    "unbash": "^4.0.4",
     "which": "catalog:",
     "yaml": "^2.8.3",
     "yargs": "catalog:",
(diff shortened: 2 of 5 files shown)


AgentEnder:
Thanks for the PR, and sorry it sat this long. It needs a push from you to move. The branch conflicts with master, and its CI runs have gone stale in a way that cannot be restarted from our side, so the rebase is also what gets it building again.

<details>
<summary>

</summary>

> Added `blocked: needs rebase` because `mergeStateStatus` is DIRTY against current master.
>
>
> No linked issue. Kept open because the body frames it as finishing #34491, which added `shell-quoting.ts` to handle `|`, and this covers the values that are not a single word.
>
> Worth a look before review: it adds a `transformIgnorePatterns` entry to `packages/nx/jest.config.cts` for an ESM-only dependency, and this repo's `packages/nx` specs run under vitest rather than jest, so that hunk may be aimed at the wrong runner.

</details>

webpro:
No worries! Rebased and updated to simpler PR description.

PR #36481, opened 2026-07-27
fix(core): only report recursive task invocations from ancestor processes

AgentEnder:
## Current Behavior

Nx reports `Recursive task invocation detected` for task graphs that contain no recursion.

The detector keys `task_invocations` on `(root_pid, task_id)`, and every nested process inherits `root_pid`. That makes the table a set of every task seen anywhere in the process tree rather than a chain, so two sibling Nx processes running the same task collide on the primary key, and we report the collision as recursion and exit 1. Neither process is an ancestor of the other.

```
app-e2e:e2e-ci--a.cy.ts -> app-e2e:serve-static
app-e2e:e2e-ci--b.cy.ts -> app-e2e:serve-static
```

Any other exception out of `registerTask` is read as a loop too, so a locked or full database prints the same error and exits.

## Expected Behavior

Only a task re-invoked by an ancestor is a loop. Siblings running the same task proceed.

The table is rekeyed on `(root_pid, pid, task_id)` and the check is scoped to `NX_INVOCATION_ANCESTOR_PIDS`, a chain each process appends its own pid to on the way down. Registration is keyed on the overrides as well as the task id, so `nx run app:build --flag` from inside `app:build` is not a loop either. `registerTask` returns the chain instead of throwing, so a db failure fails open.

`addRunningTask` also moves to just after the `getRunningTasks` check rather than after the fork. That part is load bearing: once siblings are allowed past the detector, the claim is what keeps them sharing one server instead of racing for the same port.

## Related Issue(s)

Fixes #

Regression from #34820, which added the detector. Internal: NXC-4732.


AgentEnder (commented):
Deep review at \`5d9025b\` — four independent lanes (implementation, verification, approach, security), each required to prove it opened the diff. Reviewed as a draft.

The diagnosis is right and the ancestry model is the correct one — \`(root_pid, task_id)\` genuinely could not distinguish a sibling from an ancestor, and the new key plus an env-propagated ancestor list can. The Rust tests are real and discriminating. \*\*Four Important findings\*\*, one of which the lanes disagree about; I've kept the disagreement visible rather than picking a side.

\## Important

\*\*1. \`environment-variables.mdoc:211\` — the docs state the wrong separator.\*\* The row says \`NX_INVOCATION_ANCESTOR_PIDS\` is "a \*\*colon\*\*-separated list of ancestor process IDs". The code writes and reads it \*\*comma\*\*-separated: \`appendAncestorPid\` returns \`\` \`${inherited},${process.pid}\` \`\` (\`task-env.ts:135\`) and \`getInvocationAncestorPids\` does \`.split(',')\` (\`:144\`). Confirmed both ends independently. Anyone following this page to reproduce or forward the chain builds \`123:456\`, which \`Number()\` turns into \`NaN\`, which the \`Number.isInteger\` filter drops — silently disabling ancestor detection. Impact is bounded by the row's own "Do not set this manually", but this is a factual error on a public reference page introduced by this diff. Fix: \`colon-separated\` → \`comma-separated\`.

\*\*2. \`task-env.ts:27-42\` — the ancestor pid is appended at only one of the two fork points.\*\* \`getNxEnvVariablesForTask\` appends \`process.pid\`; \`getEnvVariablesForBatchProcess\` — which produces \`this.batchEnv\`, the env for \`forkProcessForBatch\` — spreads \`...process.env\` and never appends. So a batch worker inherits its parent's \*ancestors\* without the parent itself.

Batch tasks do reach the detector (\`task-orchestrator.ts:771\` calls \`detectTaskInvocationLoop\` for every task in the batch run graph), so the parent registers rows under a pid no descendant will ever carry in \`ancestor_pids\`, and a loop passing through a batch executor becomes undetectable.

\*\*This is net-new for the nested case.\*\* On base the key was \`(root_pid, task_id)\` and \`NX_INVOCATION_ROOT_PID\` \*was\* inherited through \`...process.env\`, so base did flag it. Two lanes reached this independently. Fix is a one-liner: add \`NX_INVOCATION_ANCESTOR_PIDS: appendAncestorPid(process.env.NX_INVOCATION_ANCESTOR_PIDS)\` (and, for symmetry, \`NX_INVOCATION_ROOT_PID\`) to \`getEnvVariablesForBatchProcess\`.

\*\*3. \`task_invocation_tracker.rs:131-143\` — the "chain ordered by ancestry, not \`created_at\`" claim has no effective coverage.\*\* Proven by mutation, not argued: replacing the ancestry loop in \`ancestor_invocation_chain\` with the base ordering (filter \`invocations_for_root\`'s \`ORDER BY created_at ASC\` result by \`ancestor_pids\`) leaves \*\*all 5 new tests passing\*\*.

\`reports_a_loop_through_an_intermediate_process\` looks like it pins the order, but its insertion order coincides with its ancestry order, so it cannot fail for the regression it appears to guard.

A discriminating test was written and verified in both directions — it fails under the mutation (\`left: \["mid:task", "root:task"\]\`) and passes on HEAD — by making the root's row the \*newest\* so \`created_at\` ordering would invert the chain:

\`\`\`rust

db.lock().unwrap().execute(

    "UPDATE task_invocations SET created_at = datetime('now', '+1 hour') WHERE pid = ?1",

    params!\[ROOT_PID\],

).unwrap();

let chain = leaf.register_task(300, "mid:task".into()).unwrap().expect("loop");

assert_eq!(rendered, vec!\["root:task", "mid:task"\]);

\`\`\`

Worth knowing: a first attempt that reversed only \*insertion\* order passed under the mutation by luck — SQLite's tie order really is arbitrary, which is itself evidence for the PR's premise.

\*\*4. \`task-orchestrator.ts:1381\` — moving \`addRunningTask\` before the fork opens a window where the claim is held but nothing owns it. The two lanes that looked at this disagree, and it's worth your judgment rather than mine.\*\*

Both agree the claim is released only via \`completeContinuousTask\`, reachable through the \`onExit\` handler wired \*after\* \`runTask\` returns — i.e. after \`await this.processedTasks.get(task.id)\`, \`preRunSteps\`, \`pipeOutputCapture\` and \`getEnvVariablesForTask\`. A throw anywhere in there leaves a row in \`running_tasks\` whose pid is a \*live\* nx process, and \`is_task_running\` validates liveness + cmdline + cwd, so the row does not self-heal.

\- \*\*implementation lane calls it a defect\*\*: \`runContinuousTasks\` (\`init-tasks-runner.ts:162\`) hands back one promise per task, so a single rejection does \*\*not\*\* end the process — meaning the stale claim survives and every sibling asking for that task gets \`SharedRunningTask\` and waits indefinitely for a task nobody is running. Proposed fix: wrap from the claim to \`runningContinuousTasks.set(...)\` in \`try { … } catch (e) { this.runningTasksService?.removeRunningTask(task.id); throw e; }\`.

\- \*\*verification lane calls it a callout\*\*: no caller catches, so the process is already on its way down, and the row left by the loop path's \`process.exit(1)\` is inert.

They can't both be right about whether the process survives. \`init-tasks-runner.ts:162\` is the specific claim; if it holds, the defect reading is correct. Worth settling before merge — the fix is cheap either way.

\## Strengths

\- \*\*The ancestry model is right and is genuinely exercised.\*\* \`(root_pid, pid, task_id)\` plus the env-propagated list makes sibling and ancestor distinguishable, which the old key could not. \`invoked_by_ancestor\` short-circuits to \`false\` on an empty list, so a root invocation can never self-report.

\- \*\*The Rust tests are discriminating, not padding.\*\* \`cargo test\` (with the 1.95.0 toolchain pin): 5 passed. \`allows_sibling_processes_to_run_the_same_task\` inserts two rows that collide under the base primary key; \`unregistering_does_not_drop_a_sibling_record\` fails if the \`DELETE\` is not pid-scoped.

\- \*\*Failing open on a DB error fixes a second, separate bug.\*\* On base \*any\* \`execute\` failure — \`SQLITE_BUSY\` after the busy handler gives up, a full disk, a read-only db — surfaced as "Recursive task invocation detected" + \`process.exit(1)\`, because a unique-constraint violation and an I/O failure were the same exception. The deeper fix is structural: loops are now a \*return value\*, so a DB failure can no longer be mistaken for one by construction.

\- \*\*\`DB_VERSION\` 3→4 degrades loudly, not silently.\*\* The version is a \*filename\* component (\`db/mod.rs:28\` → \`{machine}-v4.db\`), so a v4 process opens a fresh file rather than reading a v3 table under a new primary key — which \`CREATE TABLE IF NOT EXISTS\` would otherwise have left in place, making every \`registerTask\` error. The v3 file is untouched and swept only after 7 days of disuse.

\- \*\*The binding artifacts are complete.\*\* A \`#\[napi\]\` change can need four regenerated committed files; here only \`index.d.ts\` moves, and that is correct — \`native-bindings.js:594\` exports \`TaskInvocationTracker\` by \*name\* only, carrying no signature, and neither wasi shim references it. This diff changes signatures, not the exported symbol set.

\- \*\*API migration is complete\*\*: no \`parentPid\` / \`getInvocationChain\` references remain under \`packages/\`.

\- \*\*Security is clean and was checked properly.\*\* \`ancestor_pids\` never reaches SQLite — all four statements are \`'static\` strings with \`params!\[\]\` binding, and ancestor filtering happens in Rust via \`Vec::contains\`, so the \`IN\`-clause risk doesn't exist. The env parser was run against an adversarial matrix (\`-100\`, \`1e20\`, \`abc\`, an ANSI payload, and \`100,200'); DROP TABLE task_invocations;--\`); every non-positive-integer element is dropped and both injection payloads parse to \`\[100\]\`. Forging the var can suppress or force detection, but that's not net-new (base keyed on the equally forgeable \`NX_INVOCATION_ROOT_PID\`) and the sink is a diagnostic message plus an exit code.

\- \*\*Docs mechanics are right\*\* apart from finding 1: \`vale\` runs clean on the new line (and is genuinely running, not silently passing — its only two hits are pre-existing), prettier is clean, the \`Internal.\` prefix and "Do not set this manually." closer match the adjacent rows, and no other page describes recursive-invocation detection.

\## Maintainer calls

\- \*\*Fail-open on DB error — endorsed.\*\* Killing a correct build with a misleading message to protect a diagnostic is the wrong trade. The worst case of failing open (an actual loop runs until the user notices) is what shipped before the detector existed.

\- \*\*Bundling the \`addRunningTask\` race fix with the detector fix — endorsed.\*\* The false positive was \*masking\* the race: with siblings allowed past the detector, shipping the detector fix alone would turn "false recursion error" into "several processes each start the server and race for the port." They are coupled by behavior. It would read better as its own commit in the series, but it doesn't belong in a separate PR.

\- \*\*An env chain rather than reading real ancestry via \`sysinfo\` — endorsed, for a specific reason.\*\* \`sysinfo\` is already a dependency and the walk is implementable, but it loses on failure \*direction\*: the OS tree contains shells/pty/node wrappers, is subject to reparenting when an intermediate exits, and to pid recycling — so a bad walk \*invents\* an ancestor and produces a false positive plus \`exit 1\`, which is the exact bug being fixed. The env chain's failure mode is always a \*shorter\* list, i.e. a missed loop, which is the safe direction for a diagnostic.

\## Suggestions

\- \`task-orchestrator.ts:108-110\` — the comment on \`registeredInvocations\` still says it prevents tripping "the DB uniqueness constraint"; that's no longer an error path (\`INSERT OR REPLACE\`, sibling rows legal). Stale as written.

\- \`task_invocation_tracker.rs\` — \`register_task\` queries \`invocations_for_root()\` twice on the loop path; reusing the first result halves the work and removes a race where an ancestor unregisters in between and the chain renders empty.

\- \`TaskInvocationTracker\` has no \`Drop\` impl, unlike \`RunningTasksService\`, so a killed process leaves rows until \`cleanup_stale\`'s 24h sweep. A \`Drop\` deleting \`WHERE root_pid = ? AND pid = ?\` restores base's self-sweeping without giving up pid scoping.

\- \`ancestor_invocation_chain\` returns \*every\* task each ancestor is running, so a parallel root can render unrelated concurrent tasks as if they were a call chain. Better than base (which printed the whole tree), so not a regression — but filtering to the offending \`task_id\` would make it a real call path.

\- Add an \`NX_INVOCATION_ANCESTOR_PIDS\` case to the existing \`NX_INVOCATION_ROOT_PID\` describe block in \`task-env.spec.ts\` (unset → \`String(process.pid)\`; \`'123'\` → \`'123,<pid>'\`; plus a \`getInvocationAncestorPids\` round-trip). That would also have caught finding 1.

\- \`NX_INVOCATION_ROOT_PID\` is now derivable from the new var (\`ancestorPids\[0\] ?? process.pid\`); worth collapsing the two in a follow-up.

\## Note for whoever merges

\`DB_VERSION\` 3→4 also splits the \`running_tasks\` table. A \*\*mixed-version workspace loses cross-version shared-continuous-task visibility\*\* until both sides are on v4 — two Nx versions running concurrently won't see each other's continuous tasks and may each start one. Not a defect in this PR, but release-note-worthy.

\---

\*Verdict: \*\*needs-changes\*\* — four Important findings, no Critical. Finding 3 is not downgraded as a coverage gap: the calibration that makes missing coverage advisory keeps a test which cannot fail for the regression it appears to guard as a defect, and the mutation proved this one cannot. Finding 4 is recorded with the inter-lane disagreement intact rather than resolved.\*

AgentEnder:
Assigned. It conflicts with master as it stands, so `blocked: needs rebase` is on it and the rebase comes before any review. There is also a decision to make first: #36331 is a community fix for the same false positive, in `packages/nx/src/tasks-runner` rather than in native, and only one of the two should land. Which layer the fix belongs in settles which one that is.

<details>
<summary>

</summary>

> Neither PR references the other. The overlap was found by comparing subjects and touched paths across the open PR queue:
>
> ```
> #36481  only report recursive task invocations from ancestor processes
>         packages/nx/src/native, native/db, native/tasks   (author: AgentEnder)
> #36331  avoid false positive in recursive task invocation detection
>         packages/nx/src/tasks-runner                      (author: bojanbass, approved)
> ```
>
> Whether the two actually conflict has not been checked. That is a review question rather than a triage one.
>
> Tracked in Linear as NXC-4732, so there is no GitHub closing reference to look for.

</details>

PR #36488, opened 2026-07-28
fix(maven): ignore comments and string literals when detecting test class names

### LINKED ISSUE #36487: @nx/maven: a comment containing "class " is read as the test class declaration, merging targets into a cyclic task graph
## Current Behavior

`TestClassDiscovery` extracts the test class name by scanning the raw text of each Java file for the token `class` and taking the next token (`packages/maven/maven-plugin/src/main/kotlin/dev/nx/maven/targets/TestClassDiscovery.kt`, `extractClassName`). Comments are not excluded, so prose in a Javadoc block is read as the declaration. A comment containing the words "class name" above the real declaration yields the class name `name`.

Two files in the same Maven module whose comments both match therefore collapse onto one generated target name. Because each atomized target is created with a `dependsOn` copy that already contains the previously generated atomized targets (`NxTargetFactory.kt`, the `deepCopy()` inside `testClasses.forEach`), the merged target ends up with edges to targets that already depend on it, and the task graph becomes cyclic.

In the repro below, 11 test classes produce 9 atomized targets — three files collapse into a single `com.example.name.name` target — and `test-ci` cannot run.

## Expected Behavior

Comments and string literals are ignored when detecting the package and class name, so the class name always comes from the declaration and every test class gets its own target.

## GitHub Repo

No repo — the steps below build the reproduction from scratch in about two minutes. Happy to publish one if that helps.

## Steps to Reproduce

1. Create a Maven workspace with a root `pom.xml` (packaging `pom`) and one module `app`, then add @nx/maven:

       npm install -D nx@23.0.0 @nx/maven@23.0.0
       npx nx g @nx/maven:init --no-interactive

   `nx.json`:

       { "plugins": ["@nx/maven"] }

2. In `app/src/test/java/com/example`, add three test classes whose Javadoc contains the word `class` followed by another word, for example:

       package com.example;

       import org.junit.jupiter.api.Test;

       /// A `$` in a top-level class name is treated as a nested class by Surefire, so this
       /// test is named without one.
       class FirstTest {

         @Test
         void works() {}
       }

   The other two use `/// The class name intentionally drops the `$` of the type under test.` and `/// Surefire reads the class name from the file, so keep the two in sync.` — all three extract to `name`.

3. Add eight ordinary test classes in the same package, each with a `@Test` method and no `class <word>` sequence in its comments:

       /// Ordinary test with nothing unusual in its documentation.
       class ThirdTest {

         @Test
         void works() {}
       }

4. `npx nx show project com.example:app --json` shows 9 atomized targets for 11 files, one of them named after the comment rather than a class:

       surefire:test@default-test--com.example.name.name

   Its `dependsOn` contains itself plus targets that in turn depend on it:

       test-compile-ci, install-ci,
       ...--com.example.TenthTest.TenthTest,
       ...--com.example.NinthTest.NinthTest,
       ...--com.example.name.name,
       ...--com.example.EighthTest.EighthTest,
       ...--com.example.SixthTest.SixthTest,
       ...

5. `npx nx run-many -t test-ci -p com.example:app` fails immediately.

Notes on the repro: two colliding files are enough to merge the targets, but the cycle additionally requires at least one ordinary test class to be generated between them, so the repro uses three colliding files among eight ordinary ones to make it deterministic (reproduced on 3/3 runs with `nx reset` in between). Comments that read "class name", "class is", "class allows", "class runs" are common in real test files, which makes the extracted name `name`, `is`, `allows` or `runs`.

## Nx Report

    Node           : 26.5.0
    OS             : darwin-arm64
    npm            : 11.17.0
    daemon         : Available

    nx          : 23.0.0
    @nx/devkit  : 23.0.0
    @nx/maven   : 23.0.0
    Registered Plugins:
    @nx/maven

## Failure Logs

    NX   Could not execute command because the task graph has a circular dependency

    com.example:app:test-ci --> com.example:app:surefire:test@default-test--com.example.name.name --> com.example:app:surefire:test@default-test--com.example.EighthTest.EighthTest --> com.example:app:surefire:test@default-test--com.example.name.name

## Package Manager Version

    npm 11.17.0

## Operating System

- [x] macOS
- [ ] Linux
- [ ] Windows
- [ ] Other

## Additional Information

Root cause is in the Maven-side Kotlin plugin (`nx-maven-plugin`, published to Maven Central at 0.0.17); `extractPackageName` has the same raw-text flaw. Still present on master. I have a fix that blanks comments and string literals before extraction, with unit tests, and will open a PR referencing this issue.

Azmah-Bad:
## Current Behavior

`TestClassDiscovery.extractClassName` scans the raw text of each Java test file for the token `class` and takes the next token, so a comment or string literal containing "class " is read as the class declaration. A Javadoc comment mentioning "a top-level class name" above the real declaration produces the class name `name`.

Two files in one module whose comments both match then collapse onto the same generated target name. Since each atomized target is created with a `dependsOn` copy that already lists the previously generated atomized targets, the merged target ends up with edges to targets that already depend on it and `nx run-many -t test-ci` fails with "the task graph has a circular dependency". `extractPackageName` has the same flaw.

## Expected Behavior

Comments and string literals are blanked out before the package and class name are extracted, so the name always comes from the declaration. Line breaks are preserved, so the existing line-based scan is unchanged.

Against the reproduction in #36487, where 11 test classes collapsed into 9 targets and `test-ci` could not run: all 11 classes now get their own target and the task graph is valid. 7 unit tests are added, 5 of which fail without this change.

Three notes for reviewers:

- The fix lives in the Central-published `nx-maven-plugin` jar, so it only reaches users with a new plugin version plus a `mavenPluginVersion` bump and pom migration. Do you want that in this PR, or do you handle it at release?
- The Kotlin tests here do not run on PRs (`nx-maven-plugin` has no `test` target, and @nx/maven is not applied to this repo yet — #33412). They pass locally with `mvn -pl maven-plugin -am test`. Happy to wire up a `test` target if you want it.
- Building the plugin on JDK 24+ needs `-Dkotlin.version=2.1.20`; the pinned Kotlin 1.9.22 fails in IntelliJ's `JavaVersion.parse`. Unrelated to this change.

## Related Issue(s)

Fixes #36487

### DIFF
--- a/packages/maven/maven-plugin/src/main/kotlin/dev/nx/maven/targets/TestClassDiscovery.kt
+++ b/packages/maven/maven-plugin/src/main/kotlin/dev/nx/maven/targets/TestClassDiscovery.kt
@@ -90,11 +90,15 @@ class TestClassDiscovery() {
             return null
         }
 
+        // Comments and string literals contain prose like "a top-level class name", which string
+        // matching cannot tell apart from a declaration, so only scan actual code
+        val code = withoutCommentsAndLiterals(content)
+
         // Extract package name (simple approach)
-        val packageName = extractPackageName(content)
+        val packageName = extractPackageName(code)
 
         // Extract class name (simple approach)
-        val className = extractClassName(content) ?: return null
+        val className = extractClassName(code) ?: return null
 
         // Double check: does this class actually have test methods?
         if (!hasTestMethodsInClass(content, className)) {
@@ -112,6 +116,22 @@ class TestClassDiscovery() {
             packageName = packageName
         )
     }
+
+    // Line comment, block comment, text block, string literal. Leftmost-first matching is what
+    // keeps a literal's own comment markers from opening a comment, and """ must precede "
+    private val commentOrLiteral = Regex(
+        """//[^\n]*|/\*[\s\S]*?\*/|"{3}[\s\S]*?"{3}|"(?:\\.|[^"\\\n])*""""
+    )
+    private val nonLineBreak = Regex("[^\n]")
+
+    /**
+     * Blank out comments and string literals, keeping line breaks so the result can still be
+     * scanned line by line. Without this, a Javadoc block or a string literal that happens to
+     * contain "class " is read as the class declaration: two files whose comments both mention it
+     * end up with the same generated target name, which produces a cyclic task graph.
+     */
+    private fun withoutCommentsAndLiterals(content: String): String =
+        commentOrLiteral.replace(content) { it.value.replace(nonLineBreak, " ") }
 
     /**
      * Extract package name from Java file content using simple string matching
--- a/packages/maven/maven-plugin/src/test/kotlin/dev/nx/maven/targets/TestClassDiscoveryTest.kt
+++ b/packages/maven/maven-plugin/src/test/kotlin/dev/nx/maven/targets/TestClassDiscoveryTest.kt
@@ -1,0 +1,153 @@
+package dev.nx.maven.targets
+
+import org.apache.maven.project.MavenProject
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import java.io.File
+import kotlin.test.assertEquals
+
+class TestClassDiscoveryTest {
+
+  @Nested
+  inner class ClassNameExtraction {
+
+    @Test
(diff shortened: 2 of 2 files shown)


AgentEnder:

<details>
<summary>

</summary>

>
> Closes #36487. Branch merges cleanly.

</details>

PR #36491, opened 2026-07-28
feat(linter): add the @nx/oxlint plugin

FrozenPandaz:
## Current Behavior

`@nx/oxlint` resides in the nx-labs repository, shipping on its own cadence and unavailable through generators. No Nx generator offers an option to request an Oxlint project.

Additionally, the `linter` option has never reflected workspace choices. Every generator carried static JSON-schema defaults, so `nx g @nx/react:lib` in an Oxlint-adopting workspace still produced ESLint projects. Workspaces created with `--linter=none` had ESLint inferred on first generator run without explicit flags. Plugin workspace creation ignored the choice entirely—`@nx/plugin:preset` hardcoded ESLint.

## Expected Behavior

**`@nx/oxlint` graduates into `packages/oxlint`, shipping experimental.** Additions include an inference plugin detecting Oxlint from config files, a configuration generator, and a bridge enabling Nx's `enforce-module-boundaries` rule under Oxlint's JS-plugin API.

**Project generators now follow workspace linter preferences.** The new `detectLinters` function reads the workspace and returns every linter it has, most-preferred first, so `[0]` is the one a generator should follow:

| Workspace State | Result |
|---|---|
| Has `oxlint` | Oxlint |
| Has `eslint` | ESLint |
| Has neither | None |

Previously, detection only distinguished between Oxlint and ESLint, collapsing "uses ESLint" and "uses nothing" into one category.

Verification involved 17 generators tested against all three workspace configurations without `--linter` flags—51/51 matched workspace preferences. An additional 25-generator sweep confirmed detection never overrides explicit requests.

### Breaking-ish Changes

- **`detectLinters` replaces four hand-rolled `isEslintInstalled` helpers.** Those probed `require('eslint')`, which resolves from the generator's own scope; `eslint` is a peer dependency of several first-party plugins, so they returned true in workspaces that do not use it. `detectLinters` reads the tree and is exported through `@nx/js/internal`.

- **12 generators declare `linter` as optional** on input `schema.d.ts` where previously required, though required on `NormalizedSchema`. This is strictly more permissive; missing resolution now becomes a compile error rather than silent failure.

- **Non-interactive `nx g <framework>:app|lib` in linter-free workspaces now generates no linter,** where previously ESLint was generated. User-visible on paths unrelated to Oxlint; warrants a release note.

- **Generators no longer ask which linter to use when the workspace already has one.** Following the workspace is not a question; the prompt is reserved for a workspace with no linter, where there is a real choice. Pass `--linter` to override. Cost: opting a single project out interactively now needs `--linter=none`.

- **The linter prompt moved from schemas into every generator.** Previously, JSON `x-prompt` always highlighted the first enum value regardless of workspace setup. Generators now resolve through `normalizeLinterOption`.

- **`@nx/angular:host` resolves the linter once for host and remotes.** Angular's `normalizeOptions` returns a new object, so resolved values never reached callers—each delegation resolved independently, allowing divergent answers.

- **`create-nx-plugin` now prompts for Nx Cloud in an interactive terminal.** Declaring `--interactive` (below) also un-skips `determineNxCloud`, which read the same flag. The prompt was never a decision to omit—it was dead because the flag did not exist—and this brings `create-nx-plugin` in line with `create-nx-workspace`.

- **`@nx/web` no longer scaffolds a Jest `src/test-setup.ts`.** Its `web-components` setup file only ever held the `document-register-element` polyfill, which was removed from the template long ago; the value has since produced an empty file wired into `setupFilesAfterEnv`. The file, its `setupFilesAfterEnv` entry and its `tsconfig.spec.json` include are all gated on the same value and go together.

- **`@nx/oxlint` declares `engines: { node: "^20.19.0 || >=22.12.0" }`.** It's the only ESM package under `packages/`, requiring synchronous `require` compatibility.

### Workspace Creation

Plugin workspace creation now respects `--linter`. The option threads through `@nx/plugin:preset`, with resolution occurring once—child generators prompt independently.

`create-nx-plugin` gained `--interactive` (default true), which it never declared. Without it, prompts reading this flag saw `undefined` and skipped themselves.

A detected `none` no longer pins into `nx.json`. Recording `eslint` or `oxlint` preserves real choices, but freezing `none` prevents workspaces from adopting linting later.

**The linter question now precedes the test-runner questions.** It previously trailed `unitTestRunner` and `e2eTestRunner`, sitting beside the formatter—a workspace-level question deliberately asked last. Every stack now asks it after the appearance questions (bundler, style, SSR) and before the test runners, so linter, unit-test runner and e2e runner read as one block. The react stack's preset branch was split to make this possible: the bundler resolves first, since `determineUnitTestRunner` still keys off `preferVitest: bundler === 'vite'`. The Next.js and React Native/Expo arms were identical and merged rather than duplicated.

Each stack resolves its own linter inside its `determine*Options` function, alongside every other per-stack option (`appName`, `framework`, `unitTestRunner`, `e2eTestRunner`, `formatter`). Resolution is deliberately not hoisted above the preset switch: `apps`, `ts` and `npm` reach no generator that takes a linter, so a hoisted call would ask a question and discard the answer on the most travelled path in the CLI.

`--preset=web-components` received its own `web` stack. It was the only `Preset` member mapped to the `unknown` catch-all, which skips option resolution—harmless until this PR replaced `@nx/web:application`'s internal `|| 'eslint'` with detection, at which point it scaffolded unlinted. All 24 `Preset` members are now explicitly cased.

`--preset=ts-standalone` previously asked twice—parent and child each resolved independently. The preset now forwards `--linter` so the child short-circuits.

`--no-workspaces` now asks about linters. Previously, only workspaced stacks determined linters; this layout independence is now respected.

At creation time, ESLint remains the default when prompts cannot run (CI, `--no-interactive`)—`determineLinterOptions` keeps `initial: 0` deliberately while Oxlint remains experimental.

### Generated Code Passes Lint

Generating a project and immediately linting it surfaced three failures, all measured against oxlint 1.75 rather than reasoned about.

- **The React routing scaffold used `<div role="navigation">`**, which `jsx-a11y/prefer-tag-over-role` reports; it is now a `<nav>`, which carries that role implicitly. `create-nx-workspace` defaults React workspaces to routing, so every generated app hit this. Nothing selected the wrapper by role — no stylesheet, no attribute selector, no test query — and `addRoute` locates `Route` and `Link` by tag name.
- **The `@nx/next` welcome page lacked the `prefer-tag-over-role` suppression** that the React and Remix welcome templates already carry. `@nx/next` enables the `jsx-a11y` plugin and the page embeds five inline SVGs with `role="img"`, which has no tag to swap for. Both routers inject the same content, so both wrappers carry it. Sweeping every package, `jsx-a11y` is enabled only for react, next and remix — expo and react-native get `react`/`react-perf`, vue and nuxt get `vue`, and the `role=` attributes in angular and web live in template literals rather than JSX.
- **The empty Jest setup file** described above, reported by `unicorn/no-empty-file`.

## Documentation

New Oxlint pages cover setup, inferred tasks, config format, task naming, type-aware linting, and module-boundaries bridging. Four knowledge-base pages show `create-nx-workspace` terminal transcripts with the linter prompt in the position it actually occupies—before the test-runner questions. The prompt is absent from `--preset=apps` transcripts, where it cannot fire.

## Related Issue(s)

Relates to NXC-4312

Fixes NXC-4774 — Oxlint a11y rules give contradictory guidance on `role="navigation"` vs `<nav>`. Note the reported *second, circular* violation did not reproduce: `<nav>` and even `<nav role="navigation">` are both clean, and oxlint has no `no-redundant-roles` rule.

Fixes NXC-4776 — Oxlint `unicorn/no-empty-file` flags the Jest test-setup file in the generated web app.

Known follow-up, deliberately not addressed here: NXC-4784 — the inference plugin decides on lint targets by extension count alone, so a project whose files are all ignored still gets a target, and that target fails.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Graduate-nx-oxlint-from-labs-into-packages-oxlint-NXC-4312-33c21af4)


astro-docs/src/content/docs/features/enforce-module-boundaries.mdoc
```diff
@@ -196,6 +196,59 @@ tagged with "scoped:shared" or "scope:admin".
...
+
+[`@nx/oxlint`](/docs/technologies/oxlint/introduction) exposes the rule to Oxlint as a JavaScript plugin.
+It reads the project graph and takes the options documented for ESLint, so your constraints carry over unchanged.
+
+{% aside type="caution" title="Experimental" %}
+The bridge depends on the Oxlint JavaScript plugin API. Oxlint excludes that API from its semantic versioning policy, so it can change in any release.
+{% /aside %}
+
+{% aside type="caution" title="It skips silently without a project graph" %}
+When no cached project graph is available, the rule reports nothing and the task still passes. The notice it prints is not a lint diagnostic, so `--deny-warnings` and `--max-warnings` don't catch it either. Run it behind `nx lint` rather than invoking `oxlint` directly, and don't rely on it as your only boundary check until that's fixed.
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/features/enforce-module-boundaries.mdoc
```diff
@@ -196,6 +196,59 @@ tagged with "scoped:shared" or "scope:admin".
...
+
+The ESLint rule loads the project graph through the same code path, so switching linters doesn't avoid this.
+{% /aside %}
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/features/enforce-module-boundaries.mdoc
```diff
@@ -196,6 +196,59 @@ tagged with "scoped:shared" or "scope:admin".
...
+
+`@nx/eslint-plugin` is an optional peer of `@nx/oxlint`, so no package manager installs it for you. `nx add` takes one package at a time, so add both:
+
+```shell
+nx add @nx/oxlint
+nx add @nx/eslint-plugin
+```
```
FrozenPandaz:
```suggestion
```

This is weird.. perhaps it should be a dependency

astro-docs/src/content/docs/kb/enforce-module-boundaries.mdoc
```diff
@@ -12,7 +12,7 @@ The `@nx/enforce-module-boundaries` ESLint rule enables you to define strict rul
 different projects in the repository. Enforcing strict boundaries helps to prevent unplanned cross-dependencies.
 
 {% aside type="note" %}
-This rule requires ESLint and only works for JavaScript/TypeScript projects. For language-agnostic boundary enforcement across all project dependencies, see the [Conformance plugin's Enforce Project Boundaries rule](/docs/enterprise/conformance).
+This rule only works for JavaScript/TypeScript projects. It runs on ESLint, and experimentally on [Oxlint](/docs/technologies/oxlint/introduction). On both, it reports nothing and still passes if no cached project graph is available, so don't rely on it as your only boundary check. For language-agnostic boundary enforcement across all project dependencies, see the [Conformance plugin's Enforce Project Boundaries rule](/docs/enterprise/conformance).
```
FrozenPandaz:
```suggestion
This rule only works for JavaScript/TypeScript projects. It runs on ESLint, and experimentally on [Oxlint](/docs/technologies/oxlint/introduction). For language-agnostic boundary enforcement across all project dependencies, see the [Conformance plugin's Enforce Project Boundaries rule](/docs/enterprise/conformance).
```

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+title: Nx with Oxlint
+description: Set up and use the @nx/oxlint plugin with inferred tasks, hybrid ESLint migration, and experimental module-boundary enforcement.
+sidebar:
+  label: Introduction
+weight: 2
+filter: 'type:References'
+---
+
+[Oxlint](https://oxc.rs/docs/guide/usage/linter/) is a linter written in Rust.
+The `@nx/oxlint` plugin runs it as a cached Nx task, so a change re-lints only the projects it affects.
```
FrozenPandaz:
```suggestion
The `@nx/oxlint` plugin runs it as a cacheable Nx task, so a change re-lints only the projects it affects.
```

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+`@nx/oxlint` is experimental. The module boundaries bridge depends on the Oxlint JavaScript plugin API, which Oxlint excludes from its semantic versioning policy and can change in any release.
+{% /aside %}
+
+## Set up
+
+```shell
+nx add @nx/oxlint
+```
+
+This installs `oxlint`, registers `@nx/oxlint`, and creates a root `.oxlintrc.json` if the workspace has no Oxlint config.
```
FrozenPandaz:
```suggestion
This installs `oxlint` if it is not already installed, registers `@nx/oxlint`, and creates a root `.oxlintrc.json` if the workspace has no Oxlint config.
```

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+## Inferred tasks
+
+`@nx/oxlint` infers a lint task for projects governed by an Oxlint config. It recognizes the four config files Oxlint discovers on its own:
+
+- `.oxlintrc.json`
+- `.oxlintrc.jsonc`
+- `oxlint.config.ts`
+- `oxlint.config.mts`
+
+A project gets a task when it owns an Oxlint config, or when it contains files Oxlint can lint. Documentation-only and non-JavaScript projects stay untouched unless they opt in with a config of their own.
```
FrozenPandaz:
```suggestion
A project gets a task when it owns an Oxlint config, or when it contains files Oxlint can lint. Documentation-only and non-JavaScript projects do not get a target inferred unless they opt in with a config of their own.
```

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+
+Install the official [Oxc extension](https://marketplace.visualstudio.com/items?itemName=oxc.oxc-vscode) (`oxc.oxc-vscode`) for VS Code and Cursor. `nx add @nx/oxlint` appends it to `.vscode/extensions.json` when that file already exists.
+
+Oxc also ships extensions for Zed, JetBrains IDEs, and Neovim. Any editor with LSP support can use `oxlint --lsp` as its language server.
+
+## Config file format
+
+Oxlint walks up from each file to the nearest config. Two configs in one directory is an error, not a precedence, so pick one format per directory. Prefer `.oxlintrc.json`:
+
+- `oxlint.config.ts` needs a Node runtime that executes TypeScript (Node 22.18 or later), and doesn't work with the standalone binary. Node 20 can't, whatever Oxlint's own error message says.
```
FrozenPandaz: This is a confusing sentence. Can you clearify this please?

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+- Only JSON configs get their `extends` chain tracked as task inputs. Nx can't read a TypeScript config statically.
+- Framework generators can't register Oxlint plugins for you under a TypeScript config. `nx g @nx/react:app --linter=oxlint` warns and leaves the `plugins` array to you.
```
FrozenPandaz: Hm, we can't? We should be able to by finding the area to mutate via the AST right?

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+
+## Task naming
+
+Oxlint is designed to run alongside ESLint during a migration, so the plugin claims the first available name from:
+
+1. `lint`
+2. `oxlint`
+3. `oxlint-lint`
```
FrozenPandaz: Don't other plugins also try for something like `oxlint:lint`?

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+      "error",
+      {
+        "depConstraints": []
+      }
+    ]
+  }
+}
+```
+
+The rule runs as JavaScript rather than native Rust, so it's slower than the built-in Oxlint rules. It also reloads the project graph for every file it checks, so the gap widens with workspace size.
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,133 @@
...
+
+{% aside type="caution" title="It skips silently without a project graph" %}
+When no cached project graph is available, the rule reports nothing and the task still passes. The notice it prints is not a lint diagnostic, so `--deny-warnings` and `--max-warnings` don't catch it either. Run it behind `nx lint` rather than invoking `oxlint` directly, and don't rely on it as your only boundary check until that's fixed.
+{% /aside %}
```
FrozenPandaz:
```suggestion
```

e2e/oxlint/src/oxlint.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const nxJson = readJson('nx.json');
+    expect(
+      nxJson.plugins.some(
+        (p: string | { plugin: string }) =>
+          (typeof p === 'string' ? p : p.plugin) === '@nx/oxlint'
+      )
+    ).toBe(true);
+  });
+
+  it('should infer a cached Oxlint task distinct from any ESLint task', () => {
```
FrozenPandaz: This test feels like it can be a unit test of the create nodes plugin

e2e/oxlint/src/oxlint.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    expect(target.cache).toBe(true);
+    expect(target.inputs).toContainEqual({ externalDependencies: ['oxlint'] });
+    expect(target.inputs).toContain('{workspaceRoot}/.oxlintrc.json');
+    // Oxlint reports to stdout and has no output-file flag, so the task
+    // declares no outputs — the cache replays terminal output only. Declaring
+    // one would make Nx expect a file that never appears.
+    expect(target.outputs).toBeUndefined();
+  });
+
+  it('should not infer a task for a project with no lintable files', () => {
```
FrozenPandaz: also can be a unit test

e2e/oxlint/src/oxlint.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      nxJson.plugins.some(
+        (p: string | { plugin: string }) =>
+          (typeof p === 'string' ? p : p.plugin) === '@nx/oxlint'
+      )
+    ).toBe(true);
+  });
+
+  it('should infer a cached Oxlint task distinct from any ESLint task', () => {
+    const lib = uniq('oxlintlib');
+    runCLI(
```
FrozenPandaz: we pass linter none but then expect a oxlint target? that seems weird

e2e/oxlint/src/oxlint.test.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/** `nx show project` resolves inferred command targets onto `nx:run-commands`. */
+function commandOf(target: Record<string, any>): string {
+  return target.command ?? target.options?.command ?? '';
+}
+
+describe('Oxlint', () => {
+  beforeAll(() => {
+    newProject({ packages: ['@nx/oxlint', '@nx/js'] });
```
FrozenPandaz, line 55: does the js lib default to a particular unit test runner? if so that should also be in this array

e2e/oxlint/src/oxlint.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      JSON.stringify({ rules: { 'no-debugger': 'error' } })
+    );
+    updateFile(
+      `packages/${lib}/src/index.ts`,
+      `export function clean() {\n  return 1;\n}\n`
+    );
+
+    const { name: targetName } = requireOxlintTarget(lib);
+    const run = () =>
+      runCLI(`run ${lib}:${targetName}`, { silenceError: true });
```
FrozenPandaz: why do we silence the error? Can't we try catch it?

packages/angular/src/generators/storybook-configuration/lib/generate-storybook-configuration.ts
```diff
@@ -12,7 +12,12 @@ export async function generateStorybookConfiguration(
   return await configurationGenerator(tree, {
     project: options.project,
     uiFramework: '@storybook/angular',
-    linter: options.linter,
+    // `@nx/storybook`'s schema types `linter` from the *published* `@nx/eslint`,
+    // whose `LinterType` predates oxlint, so `'oxlint'` does not typecheck here
+    // until that version ships. Coercing is safe: the storybook generator never
+    // reads this option — it decides what to do from `findEslintFile` instead —
+    // so all three values produce identical output.
+    linter: options.linter === 'eslint' ? 'eslint' : 'none',
```
FrozenPandaz: This comment is.. too long.. and is there a reason why we can't use LinterType? All of these will be the same version.

packages/create-nx-workspace/bin/create-nx-workspace.ts
```diff
@@ -1137,27 +1153,28 @@ async function determineFormatterOptions(
...
-        },
-        {
-          name: 'No',
-        },
+        { name: 'eslint' },
+        { name: 'oxlint', message: 'oxlint (experimental)' },
+        { name: 'none' },
       ],
-      initial: opts?.preferEslint ? 0 : 1,
+      initial: opts?.preferEslint ? 0 : 2,
```
FrozenPandaz: In what cases do we prefer eslint? can we change this more to.. prefer oxlint instead maybe?

packages/cypress/src/utils/add-linter.ts
```diff
@@ -51,8 +52,16 @@ export async function addLinterToCyProject(
...
+  // ignore entries — which have no equivalent in other linters. They only
+  // need the linter registering, which the helper handles (including `none`).
+  if (options.linter && options.linter !== 'eslint') {
+    return addLintingToProject(tree, {
+      linter: options.linter,
+      project: options.project,
+      addPlugin: options.addPlugin,
+      skipPackageJson: options.skipPackageJson,
+    });
   }
```
FrozenPandaz: I think I'd prefer if this was written...
```
addLintingToProject()

if (linter === 'eslint') {
  // doExtraStuff
}
```

packages/eslint-plugin/internal.ts
```diff
@@ -0,0 +1,20 @@
+// Semi-private surface for first-party Nx packages.
```
FrozenPandaz: Comment too long. Please trim.

packages/expo/src/generators/library/library.ts
```diff
@@ -93,6 +93,7 @@ export async function expoLibraryGeneratorInternal(
   }
 
   const lintTask = await addLinting(host, {
+    unitTestRunner: options.unitTestRunner,
     ...options,
```
FrozenPandaz: Wouldn't this already be in ...options?

packages/expo/src/generators/application/application.ts
```diff
@@ -77,6 +77,7 @@ export async function expoApplicationGeneratorInternal(
   }
 
   const lintTask = await addLinting(host, {
+    unitTestRunner: options.unitTestRunner,
     ...options,
```
FrozenPandaz: Wouldn't this already be in ...options?

packages/express/src/generators/application/schema.d.ts
```diff
@@ -1,4 +1,9 @@
-import type { Linter, LinterType } from '@nx/eslint';
+// `LinterType` comes from `@nx/js`, not `@nx/eslint`: this package is the one
+// `@nx/eslint` importer that does not declare it as a dependency, so the
+// specifier resolves to the published tarball, whose `LinterType` predates
+// `oxlint` and would contradict the enum in schema.json.
+import type { Linter } from '@nx/eslint';
```
FrozenPandaz: Mm.. we should just not import from '@nx/eslint'?

packages/js/src/utils/add-linting-to-project.spec.ts
```diff
@@ -0,0 +1,89 @@
+import 'nx/src/internal-testing-utils/mock-project-graph';
```
FrozenPandaz: can this be imported from @nx/devkit/internals instead?

packages/js/src/utils/linter.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export function detectLinter(tree: Tree): Exclude<LinterType, 'none'> {
+  if (
+    hasPlugin(tree, '@nx/oxlint') ||
+    hasAnyDependency(tree, ['@nx/oxlint', 'oxlint'])
+  ) {
+    return 'oxlint';
+  }
+
+  return 'eslint';
```
FrozenPandaz: Shouldn't this also check if eslint is being used or not? if not, it should return none?

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,134 @@
...
+1. `lint`
+2. `oxlint`
+3. `oxlint:lint`
+4. `oxlint-lint`
+
+A workspace with no other linter gets `lint`. A workspace where ESLint already owns `lint` gets `oxlint`.
+
+## Customizing the task
+
+`@nx/oxlint` has no executor. Every Oxlint task comes from inference, so the plugin is registered even when `useInferencePlugins` is disabled.
```
FrozenPandaz:
```suggestion
```

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,134 @@
...
+
+```json
+{
+  "targets": {
+    "lint": {
+      "command": "oxlint --type-aware ."
+    }
+  }
+}
+```
```
FrozenPandaz: Use args as an example instead.. otherwise too much of the target will get overwritten.a

astro-docs/src/content/docs/technologies/oxlint/introduction.mdoc
```diff
@@ -0,0 +1,134 @@
...
+    }
+  ]
+}
+```
+
+## Add Oxlint to an ESLint workspace
+
+Oxlint runs alongside ESLint, so the setup above is all you need and nothing about your ESLint config changes. ESLint keeps the `lint` target, so Oxlint registers as `oxlint` and every project with lintable files gets one.
+
+Nx doesn't translate ESLint rules into Oxlint rules. To generate an `.oxlintrc.json` from your existing ESLint config, run [`@oxlint/migrate`](https://github.com/oxc-project/oxlint-migrate). To stop ESLint from re-reporting what Oxlint already covers, add [`eslint-plugin-oxlint`](https://www.npmjs.com/package/eslint-plugin-oxlint) to the ESLint config.
```
FrozenPandaz: this is still a weird section... I think maybe let's remove it for now until we know what the questions need to be answered.

packages/cypress/src/utils/add-linter.ts
```diff
@@ -51,21 +52,20 @@ export async function addLinterToCyProject(
   tree: Tree,
   options: CyLinterOptions
 ) {
-  if (options.linter === 'none') {
```
FrozenPandaz, line 54: maybe this should stay?

packages/cypress/src/utils/add-linter.ts
```diff
@@ -6,7 +6,7 @@ import {
   runTasksInSerial,
   Tree,
 } from '@nx/devkit';
-import { Linter, LinterType, lintProjectGenerator } from '@nx/eslint';
+import { Linter, LinterType } from '@nx/eslint';
```
FrozenPandaz: why does this come from eslint? It should probably only LinterType from js

packages/oxlint/src/generators/configuration/configuration.ts
```diff
@@ -0,0 +1,58 @@
...
+  /** Oxlint plugins to enable for this project, e.g. `['react', 'jsx-a11y']`. */
+  plugins?: string[];
+}
+
+export function configurationGenerator(
+  tree: Tree,
+  options: ConfigurationGeneratorSchema
+) {
+  return configurationGeneratorInternal(tree, options);
+}
```
FrozenPandaz: Do we need an internal one vs a public one? they're the same.. so probably not

packages/oxlint/src/generators/init/init.ts
```diff
@@ -0,0 +1,172 @@
...
+  if (!options.skipFormat) {
+    await formatFiles(tree);
+  }
+
+  return runTasksInSerial(...tasks);
+}
+
+export function initGenerator(tree: Tree, options: InitGeneratorSchema) {
+  return initGeneratorInternal(tree, options);
+}
```
FrozenPandaz: again also don't need a separate internal method if they're the same.

leosvelperez (approved):

PR #36495, opened 2026-07-28
fix(core): make tui cloud icon visible on light terminal themes

FrozenPandaz:
## Current Behavior

The Nx Cloud icon in the TUI status bar (left of the task counts) is effectively invisible on a light-theme terminal.

The icon was written as `U+2601` followed by **VS16** (`U+FE0F`), the *emoji* variation selector:

```rust
let mut icon_style = Style::default().fg(THEME.secondary_fg);
...
spans.push(Span::styled("☁\u{fe0f} ", icon_style));
```

VS16 instructs the terminal to draw the glyph from its **color emoji font**, which ignores the SGR foreground color. So the `fg(THEME.secondary_fg)` on that span never took effect, and the terminal painted the Apple/Noto cloud instead — a near-white glyph with a pale gray outline. Fine on a dark background, invisible on white.

The theme system was working correctly the whole time; the color just never reached the glyph.

## Expected Behavior

The icon renders in `THEME.secondary_fg` as originally intended — `Color::DarkGray` on the light theme, `Color::Gray` on the dark theme — so it stays legible in both.

The fix swaps VS16 for **VS15** (`U+FE0E`, text presentation), so the terminal draws a monochrome glyph from the text font, which honors the foreground color:

```rust
spans.push(Span::styled("☁\u{fe0e} ", icon_style));
```

The color itself is deliberately unchanged. Bumping the icon to `THEME.primary_fg` for extra weight was tried and rejected in favor of the icon matching the gray of the counts it prefixes, consistent with the existing "deliberately quiet" design note on `status_line()`.

### Layout impact

The icon narrows from two cells to one. No layout work was needed — the status bar derives its widths from `status_line.width()` and had no hardcoded icon widths, so this flowed through automatically. Only two width assertions and the affected snapshots changed. Total row width is unchanged; the freed cell becomes padding.

The regenerated snapshots no longer carry the `Hidden by multi-width symbols: [(2, " ")]` trailer — that was ratatui reporting the second cell of the double-width emoji, which no longer exists.

### Validation

- `cargo test -p nx --lib` — 570 passed, 0 failed
- `cargo fmt -p nx -- --check` — clean
- `cargo clippy -p nx --lib` — no new warnings in the changed file

One gap worth flagging for review: `Theme::is_dark_mode()` hard-returns `true` under `#[cfg(test)]`, so **no automated test exercises the light palette**. The tests here confirm the glyph and the layout; the light-theme improvement itself was verified by eye. Making that palette testable is out of scope for this fix but may be worth a follow-up.

This was the only VS16 emoji-presentation sequence in `packages/nx/src/native/`, so no other TUI glyph has the same defect.

## Related Issue(s)

No filed issue — reported internally while using the TUI on a light-theme terminal.

---
[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Fix-invisible-Nx-Cloud-icon-in-TUI-status-bar-on-light-themes-46534fda)

### DIFF
--- a/packages/nx/src/native/tui/components/status_bar.rs
+++ b/packages/nx/src/native/tui/components/status_bar.rs
@@ -426,13 +426,19 @@ impl<'a> StatusBar<'a> {
         // A cloud icon marks a cloud-enabled run. Kept outside the underlined
         // span so the click affordance stays on the numbers themselves.
         if props.cloud_enabled || props.cloud_link.is_some() || props.cloud_message.is_some() {
+            // Same muted foreground as the counts it prefixes — the icon is a
+            // marker, not a call to action.
             let mut icon_style = Style::default().fg(THEME.secondary_fg);
             if props.is_dimmed {
                 icon_style = icon_style.add_modifier(Modifier::DIM);
             }
-            // U+2601 + VS16 forces the filled emoji glyph (two cells wide),
-            // followed by a single space before the counts.
-            spans.push(Span::styled("☁\u{fe0f} ", icon_style));
+            // U+2601 + VS15 forces the *text* presentation (one cell wide).
+            // The emoji presentation (VS16) is drawn by the terminal's color
+            // emoji font, which ignores our foreground color and paints a near
+            // white cloud — invisible against a light background. The text
+            // glyph is monochrome and takes `icon_style`, so it stays legible
+            // in both themes. One space separates it from the counts.
+            spans.push(Span::styled("☁\u{fe0e} ", icon_style));
         }
         spans.push(Span::styled(
             format!("{}/{}", props.completed_count, props.total_count),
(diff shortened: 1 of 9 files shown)


packages/nx/src/native/tui/components/status_bar.rs
```diff
@@ -426,13 +426,25 @@ impl<'a> StatusBar<'a> {
...
+            // emoji font, which ignores our foreground color and paints a near
+            // white cloud — invisible against a light background. The text
+            // glyph is monochrome and takes `icon_style`, so it stays legible
+            // in both themes.
+            //
+            // Two trailing spaces, not one: the emoji spanned two cells and
+            // used the second as slack, so a single space around the narrower
+            // text glyph reads cramped. This keeps the icon segment three
+            // cells wide, leaving the counts in the column they have always
+            // occupied.
```
AgentEnder: nit: trim comment?

AgentEnder (approved):

PR #36499, opened 2026-07-29
fix(core): apply legacy array-form targetDefaults instead of silently dropping them

### LINKED ISSUE #36489: targetDefaults dependsOn silently dropped on an inferred (@nx/docker) target when not redeclared in project.json
## Current Behavior

A `targetDefaults` entry in `nx.json` that adds `dependsOn` to an **inferred** target is silently dropped when the target is inferred by a plugin that already sets its own `dependsOn` (e.g. `@nx/docker`, which hardcodes `dependsOn: ['build', '^build']` on the `docker:build` target it infers).

Given:

```jsonc
// nx.json
{
  "plugins": [
    {
      "plugin": "@nx/docker",
      "options": {
        "buildTarget": { "name": "docker:build", "args": ["--tag", "{projectName}:latest"] },
        "runTarget":   { "name": "docker:run",   "args": ["--rm"] }
      }
    }
  ],
  "targetDefaults": [
    { "target": "docker:build", "dependsOn": ["build", "prune"] }
  ]
}
```

and a project that has a `Dockerfile` (so `docker:build` is inferred) plus its own `build` and `prune` targets, but does **not** redeclare `docker:build`, the resolved target is:

```
docker:build.dependsOn = ["build", "^build"]   // the plugin's value — the "prune" from targetDefaults is gone
```

The `targetDefaults` contribution (`prune`) is discarded entirely rather than merged.

## Expected Behavior

The `dependsOn` from `targetDefaults` should be applied to the inferred target, as it was prior to Nx 23. Expected:

```
docker:build.dependsOn = ["build", "prune"]   // or a merge that also retains the plugin's "^build"
```

This is the same class of regression as #36067 / #36142, but for the case where the inferred target is **not** redeclared in `project.json` — the fix in #36142 only covers "a `project.json` target overrides an inferred `nx:run-commands` target", so this variant remains unfixed.

## Steps to Reproduce

Minimal, self-contained reproduction (no plugins other than `@nx/docker`):

1. Create a workspace with this `nx.json` (see above): `@nx/docker` plugin inferring `docker:build`, plus an array-shape `targetDefaults` entry `{ "target": "docker:build", "dependsOn": ["build", "prune"] }`.
2. Add a project `apps/api/` containing:
   - `apps/api/Dockerfile` (any valid Dockerfile — this triggers `@nx/docker` inference)
   - `apps/api/project.json`:
     ```jsonc
     {
       "name": "api",
       "targets": {
         "build": { "executor": "nx:run-commands", "options": { "command": "echo build" } },
         "prune": { "executor": "nx:run-commands", "options": { "command": "echo prune" } }
       }
     }
     ```
     (Note: `docker:build` is intentionally **not** declared here.)
3. Run `nx show project api --json` and inspect `targets['docker:build'].dependsOn`.

**Observed:** `["build", "^build"]` — `prune` is missing.
**Expected:** `prune` present (e.g. `["build", "prune"]`).

## Nx Report

Reproduced on the latest stable and the newest canary. Minimal repro dependencies:

```
Node : 24.11.0
OS   : darwin-arm64

nx         : 23.1.0
@nx/docker : 23.1.0
```

Also reproduced (identical result) on `nx@23.2.0-canary.20260724-9c735d4` + `@nx/docker@23.2.0-canary.20260724-9c735d4`, which includes #36257 — so the most recent fix in this area does not address this variant.

## Failure Logs

Root cause appears to be the interaction between two pieces of the target-defaults merge:

- `targetDefaults` are synthesized into a middle "synthetic plugin" layer in [`createTargetDefaultsResults`](https://github.com/nrwl/nx/blob/master/packages/nx/src/project-graph/utils/project-configuration/target-defaults.ts#L42), which is then merged against the inferring plugin's target.
- During that merge, an array property with no `'...'` spread token **wholesale-replaces** the base rather than merging — [`mergeArrayValue`, "No spread: newValue replaces baseValue entirely"](https://github.com/nrwl/nx/blob/master/packages/nx/src/project-graph/utils/project-configuration/utils.ts#L162-L169).

Because the inferred `@nx/docker` target already provides `dependsOn: ['build', '^build']` (a plain array, no spread), it replaces the `targetDefaults` `dependsOn` instead of merging with it, so `prune` never reaches the resolved target.

Notably, placing a `'...'` spread token **in the `targetDefaults` entry itself** does not help — at that layer the spread expands against an empty base:

```jsonc
// nx.json targetDefaults — does NOT work
{ "target": "docker:build", "dependsOn": ["build", "prune", "..."] }
// resolved: ["build", "^build"]   (prune still dropped)
```

The spread only takes effect when moved to project-level config (see workaround below), which is the top merge layer.

## Additional Information

**Workaround (verified on both 23.1.0 and the canary):** declare `docker:build` at the project level using the spread token so the inferred deps are retained:

```jsonc
// apps/api/project.json
"docker:build": { "dependsOn": ["prune", "..."] }
// resolved: ["prune", "build", "^build"]  ✅
```

Summary of what was tested against `nx show project`:

| Where `dependsOn` is defined | Resolved `docker:build.dependsOn` | Works? |
| --- | --- | --- |
| `targetDefaults`: `["build", "prune"]` | `["build", "^build"]` | ❌ |
| `targetDefaults`: `["build", "prune", "..."]` | `["build", "^build"]` | ❌ |
| `project.json`: `["build", "prune"]` | `["build", "prune"]` | ✅ (loses `^build`) |
| `project.json`: `["prune", "..."]` | `["prune", "build", "^build"]` | ✅ |

**Real-world workspace with this config** (public, if a non-minimal reproduction helps): [`thdk/nx-monorepo-demo@f009451`](https://github.com/thdk/nx-monorepo-demo/tree/f0094515927c8036095322630b8c5f5272d1a7b3) — see the [`docker:build` targetDefault](https://github.com/thdk/nx-monorepo-demo/blob/f0094515927c8036095322630b8c5f5272d1a7b3/nx.json#L180-L183) and the [`node-fastify-tsc` project.json](https://github.com/thdk/nx-monorepo-demo/blob/f0094515927c8036095322630b8c5f5272d1a7b3/apps/nx-demo/node-fastify-tsc/project.json) (which defines `prune`/`prune-lockfile`/`copy-workspace-modules` but not `docker:build`), then run `nx show project node-fastify-tsc --json`.

Related: #36067, #36142 (fixed the sibling "project.json overrides an inferred run-commands target" case).

Mohith26:
Fixes #36489, reported by @thdk.

The Nx 23 beta shipped a migration (`23-0-0-convert-target-defaults-to-array`) that rewrote `nx.json` `targetDefaults` into an array of `{target/executor, ...config}` entries. Stable Nx 23 settled on the map shape and deleted the migration, but `normalizeTargetDefaults` reads a leftover array as a record, producing integer keys ("0", "1", ...) that never match any target, so every default, like the reporter's `docker:build` `dependsOn`, is silently dropped.

For what it's worth, the map form merges onto inferred @nx/docker targets correctly (verified with a harness), so the trigger is purely the leftover array shape; the reporter's demo repo has the beta migration in its git history.

The fix teaches `normalizeTargetDefaults` (the single choke point used by both graph-merge synthesis and `readTargetDefaultsForTarget`) to convert the legacy array back to map form: `target` (or `executor` when no target) becomes the key, and `projects`/`plugin` matchers map onto the filter namespace. Eight regression tests (six normalizer, one reader, one end-to-end repro of the issue's docker scenario) fail without the fix; 198 tests pass across the four related spec files with zero new failures. tsc and prettier clean.

### DIFF
--- a/packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
+++ b/packages/nx/src/project-graph/utils/project-configuration-utils.spec.ts
@@ -1538,6 +1538,95 @@ describe('project-configuration-utils', () => {
       expect(errors).toEqual([]);
     });
 
+    it('should apply legacy array-form targetDefaults to an inferred target that project.json does not redeclare (#36489)', () => {
+      // Repro for #36489: nx.json still carries the top-level-array
+      // `targetDefaults` shape written by the Nx 23 beta migration
+      // (`23-0-0-convert-target-defaults-to-array`). @nx/docker infers
+      // `docker:build` with its own `dependsOn`, and project.json declares
+      // `build` and `prune` but never redeclares `docker:build`. Before the
+      // fix the array was read as a record, so its integer keys matched no
+      // target and the `docker:build` default was silently dropped — the
+      // resolved target kept only the plugin's `['build', '^build']`.
+      const specifiedResults = [
+        [
+          [
+            '@nx/docker',
+            'apps/api/Dockerfile',
+            {
+              projects: {
+                'apps/api': {
+                  root: 'apps/api',
+                  targets: {
+                    'docker:build': {
+                      dependsOn: ['build', '^build'],
+                      command: 'docker build .',
+                      options: {
+                        cwd: 'apps/api',
+                        args: ['--tag', 'api:latest'],
+                      },
+                      inputs: ['default', '^default'],
+                    },
+                  },
+                },
+              },
+            },
+          ],
+        ],
+      ] as const;
+
+      const defaultResults = [
+        [
+          [
+            'nx/core/project-json',
+            'apps/api/project.json',
+            {
+              projects: {
+                'apps/api': {
+                  name: 'api',
+                  root: 'apps/api',
+                  targets: {
+                    build: {
+                      executor: 'nx:run-commands',
+                      options: { command: 'echo build' },
+                    },
+                    prune: {
+                      executor: 'nx:run-commands',
+                      options: { command: 'echo prune' },
+                    },
+                  },
+                },
+              },
+            },
+          ],
+        ],
+      ] as const;
+
+      const errors = [];
+      const result = mergeCreateNodesResults(
+        specifiedResults as any,
+        defaultResults as any,
+        {
+          targetDefaults: [
+            { target: 'lint', dependsOn: ['build'] },
+            { target: 'docker:build', dependsOn: ['build', 'prune'] },
+          ] as any,
+        },
+        '/tmp/test',
+        errors
+      );
+
+      const dockerBuild =
+        result.projectRootMap['apps/api'].targets!['docker:build'];
+      // The targetDefaults contribution replaces the inferred dependsOn.
+      expect(dockerBuild.dependsOn).toEqual(['build', 'prune']);
+      // The inferred target's identity is untouched.
+      expect(dockerBuild.executor).toEqual('nx:run-commands');
+      expect(dockerBuild.options).toEqual(
+        expect.objectContaining({ command: 'docker build .' })
+      );
+      expect(errors).toEqual([]);
+    });
+
     it('should merge multiple specified plugins contributing to the same project', () => {
       const specifiedResults = [
         [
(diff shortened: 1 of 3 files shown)


AgentEnder (commented):
Hey, thanks for the work that went into this. Im going to close it out though, the array form from the v23 betas is unsupported and as it was only available in the betas it doesn't make sense to add backwards compatibility

PR #36514, opened 2026-07-29
fix(core): share worktree cache under ~/.nx so agent sandboxes can reach it

AgentEnder:
## Current Behavior

Worktrees share one cache and workspace-data directory, and that directory is the main checkout's `.nx`. Its path is absolute and unique to the machine.

Agent sandboxes grant paths, so reaching it needs a per-machine allowance, and a per-machine allowance cannot be committed. Inside an agentic migration, which runs in a worktree, a sandboxed agent can read the shared cache but every write to it fails with `EPERM`.

## Expected Behavior

Cache and workspace data live in `~/.nx/<hash>/{cache,workspace-data}`, keyed on the repository. That path is the same string on every machine, so the existing `~/.nx` allowance already covers it and the sandbox configuration stays in the committed `.claude/settings.json`.

Every checkout resolves there, not only linked worktrees. A main checkout that kept its own `.nx` would stop sharing with its worktrees the moment the first one was added.

A CI step that archives `.nx/cache` needs to point at `~/.nx` instead, or set `NX_CACHE_DIRECTORY`.

An explicitly configured `cacheDirectory` or `NX_WORKSPACE_DATA_DIRECTORY` is left where the user put it.

A process that cannot write under `~/.nx` falls back to its own checkout. Nx establishes this by writing a marker rather than by recognizing the caller, so a sandbox Nx has never heard of is covered.

## Related Issue(s)

Fixes NXC-4625

https://linear.app/nxdev/issue/NXC-4625/sandbox-safe-shared-worktree-cache-in-agentic-migration

Stacked on #36586.

---

<p><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.ops.cloud.nx.app/polygraph/session-logo-v4-dark.svg">Polygraph</picture> <a href="https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/NXC-4625-df9780df">View session ↗</a></p>


packages/nx/src/utils/worktree-data-root.ts
```diff
@@ -0,0 +1,238 @@
...
+  ensureOwnedPrivateDir,
+} from './owned-private-dir';
+
+/**
+ * Lives here rather than in `cache-directory` because the sharing decision
+ * turns on it and that module imports this one.
+ */
+export function readCacheDirectoryProperty(root: string): string | undefined {
+  try {
+    const nxJson = readJsonFile<NxJsonConfiguration>(join(root, 'nx.json'));
```
FrozenPandaz: use readNxJson

packages/nx/src/utils/worktree-data-root.ts
```diff
@@ -0,0 +1,238 @@
...
+    return undefined;
+  }
+}
+
+/** The data a set of worktrees shares, each in its own directory. */
+export type WorktreeDataKind = 'cache' | 'workspace-data';
+
+const WORKTREE_DATA_KINDS: readonly WorktreeDataKind[] = [
+  'cache',
+  'workspace-data',
```
FrozenPandaz: This is basically only the db.. and the shared db at that

packages/nx/src/utils/worktree-data-root.ts
FrozenPandaz, line 1: shared-data

packages/nx/src/utils/worktree-data-root.ts
```diff
@@ -0,0 +1,238 @@
...
+): WorktreeDataLocation {
+  let mainRoot: string | null;
+  try {
+    mainRoot = getMainWorktreeRoot(root);
+  } catch {
+    // Worktree detection is best-effort.
+    return { share: 'none' };
+  }
+
+  if (!mainRoot) {
```
FrozenPandaz: Even if there are no multiple worktrees yet.. we should prefer the user location

packages/nx/src/utils/worktree-data-root.ts
```diff
@@ -0,0 +1,238 @@
...
+    return { share: 'none' };
+  }
+
+  // Process-global, and pointing at one location for this run whichever
+  // checkout asks. There is nothing left to share.
+  if (
```
FrozenPandaz: this needs to move up

packages/nx/src/utils/cache-directory.ts
```diff
@@ -1,22 +1,12 @@
 import { existsSync } from 'fs';
 import { isAbsolute, join } from 'path';
-import { NxJsonConfiguration } from '../config/nx-json';
-import { getMainWorktreeRoot } from '../native';
-import { readJsonFile } from './fileutils';
+import {
+  readCacheDirectoryProperty,
+  resolveWorktreeDataLocation,
+  sharedUserDataDir,
+} from './worktree-data-root';
```
FrozenPandaz: These files seem highly related to each other.. is there a reason why it's not 1 file?

packages/nx/src/utils/db-connection.ts
```diff
@@ -1,32 +1,31 @@
...
-    // Fall back to local workspace data if worktree detection fails
-  }
-  _sharedDir = workspaceDataDirectoryForWorkspace(root);
+  const location = resolveWorktreeDataLocation(root);
+  _sharedDir =
+    location.share === 'user'
+      ? sharedUserDataDir(location.mainRoot, 'workspace-data')
+      : workspaceDataDirectoryForWorkspace(
+          location.share === 'main' ? location.mainRoot : root
+        );
```
FrozenPandaz: This sort of code appears multiple times.. Can we reuse a function?

FrozenPandaz (changes requested):

packages/nx/src/utils/worktree-data-root.ts
AgentEnder: Done. It also picks up a `cacheDirectory` reached through `extends`, which the raw read missed. Kept the try/catch, since `readNxJson` throws on malformed JSON and this runs at module scope.

packages/nx/src/utils/worktree-data-root.ts
AgentEnder: Right, and nothing said so. `getDbConnection` is the only consumer of the shared `workspace-data` directory. Daemon logs and the `disabled` marker stay in the checkout's own, because they describe that checkout. Written down at the resolver now.

packages/nx/src/utils/worktree-data-root.ts
AgentEnder: Went with the merge instead, per your comment on `cache-directory.ts`, so `worktree-data-root.ts` is gone rather than renamed.

Say the word if you want the merged file called `shared-data.ts`. It is a rename plus 27 importers, which is why I did not fold it in here.

packages/nx/src/utils/worktree-data-root.ts
AgentEnder: Done, and this one is worth a second look because it is bigger than it sounds.

Taken literally it moves every workspace's cache to `~/.nx/<hash>/cache`, not just worktrees. A CI step that archives `.nx/cache` is then archiving a directory Nx no longer writes to. I updated `nx reset` to clear both locations, and the five docs pages that named `.nx/cache` as the default, with a callout for the CI case.

That goes past what NXC-4625 asked for. If you meant only "a main checkout that already has worktrees", say so and I will scope it back.

packages/nx/src/utils/worktree-data-root.ts
AgentEnder: Moved above the `getMainWorktreeRoot` call, so a pinned location settles before anything touches git or the filesystem.

packages/nx/src/utils/cache-directory.ts
AgentEnder: No good reason, so they are one file now. `worktree-data-root.ts` folded into `cache-directory.ts`, and its spec came with it.

packages/nx/src/utils/db-connection.ts
AgentEnder: One `sharedDataDirectory(root, kind)` in `cache-directory.ts`, used by the cache, the DB and `nx reset`.

First pass took the per-kind fallback as a callback. Once the two files merged that was indirection with nothing behind it, so it switches on `kind` instead.

packages/nx/src/utils/owned-private-dir.ts
```diff
@@ -198,6 +200,40 @@ function chmodRealDirectory(path: string, mode: number): boolean {
...
+ *
+ * Only `ENOENT` walks up — any other errno means the path exists and cannot be
+ * read through, which `ensureOwnedPrivateDir` cannot establish either.
+ */
+export function canonicalDir(dir: string): string {
+  const resolved = resolve(dir);
+  const missing: string[] = [];
+  let candidate = resolved;
+
+  for (;;) {
```
FrozenPandaz: Make this a while loop instead…

packages/nx/src/utils/db-connection.ts
```diff
@@ -1,32 +1,24 @@
...
-    const mainRoot = getMainWorktreeRoot(root);
-    if (mainRoot) {
-      _sharedDir = workspaceDataDirectoryForWorkspace(mainRoot);
-      return _sharedDir;
-    }
-  } catch {
-    // Fall back to local workspace data if worktree detection fails
-  }
-  _sharedDir = workspaceDataDirectoryForWorkspace(root);
+  _sharedDir ??= sharedDataDirectory(root, 'workspace-data');
```
FrozenPandaz, line 21: It's only dbs now so let's call the directory dbs for the shared folder

FrozenPandaz (approved):

Issue #20421, opened 2023-11-26
Target project does not exist: npm:rimraf@2.4.5


markabrahams:
### Current Behavior

When I run:
`npx nx run myproject:serve`

I get the error message:
```
>  NX   An error occured while creating pruned lockfile

   Please open an issue at `https://github.com/nrwl/nx/issues/new?template=1-bug.yml` and provide a reproduction.
   To prevent the build from breaking we are returning the root lock file.
   If you run `npm install --package-lock-only` in your output folder it will regenerate the correct pruned lockfile.
   
Original error: Target project does not exist: npm:rimraf@2.4.5


   Error: Target project does not exist: npm:rimraf@2.4.5
    at validateCommonDependencyRules (/home/mark/myproject/node_modules/nx/src/project-graph/project-graph-builder.js:323:15)
    at validateDependency (/home/mark/myproject/node_modules/nx/src/project-graph/project-graph-builder.js:313:5)
    at ProjectGraphBuilder.addDependency (/home/mark/myproject/node_modules/nx/src/project-graph/project-graph-builder.js:190:9)
    at ProjectGraphBuilder.addStaticDependency (/home/mark/myproject/node_modules/nx/src/project-graph/project-graph-builder.js:89:14)
    at /home/mark/myproject/node_modules/nx/src/plugins/js/lock-file/project-graph-pruning.js:81:17
    at Array.forEach (<anonymous>)
    at traverseNode (/home/mark/myproject/node_modules/nx/src/plugins/js/lock-file/project-graph-pruning.js:78:36)
    at /home/mark/myproject/node_modules/nx/src/plugins/js/lock-file/project-graph-pruning.js:80:9
    at Array.forEach (<anonymous>)
    at traverseNode (/home/mark/myproject/node_modules/nx/src/plugins/js/lock-file/project-graph-pruning.js:78:36)

```

So I'm opening this as per the above request in the error message.

Note that this doesn't occur on the initial compile, but does occur on every recompile after that.

Also note that this doesn't *seem* to have any impact on the actual serving of the app, which seems fine.

### Expected Behavior

No reported error message.

### Steps to Reproduce

1. Run command:

`npx nx run myproject:serve`

2. Edit a file and save.  On recompile the error is produced.

### Nx Report

```shell
mark@hope:~/myproject$ npx nx report
>  NX  Falling back to ts-node for local typescript execution. This may be a little slower.
  - To fix this, ensure @swc-node/register and @swc/core have been installed

[...]

PR #36517, opened 2026-07-30
fix(core): stop pruneProjectGraph from mutating the source project graph

### LINKED ISSUE #36470: pruneProjectGraph mutates node.name on the source graph, breaking every prune after the first when a package has multiple versions
### Current Behavior

`pruneProjectGraph` mutates the `ProjectGraph` it is given. `switchNodeToHoisted` reassigns `node.name` on the node object that is **shared by reference** with the source graph, so the source graph is left with a node stored under key `npm:<pkg>@<version>` whose `name` is now `npm:<pkg>`.

Every subsequent prune against that same graph object then resolves the node by key, registers it in the builder under the mutated `name`, and the dependency edge no longer resolves:

```
NX   An error occurred while creating pruned lockfile
Original error: Target project does not exist: npm:cookie@1.1.1

Error: Target project does not exist: npm:cookie@1.1.1
    at validateCommonDependencyRules (nx/dist/src/project-graph/project-graph-builder.js:325:15)
    at validateDependency (nx/dist/src/project-graph/project-graph-builder.js:316:5)
    at ProjectGraphBuilder.addDependency (nx/dist/src/project-graph/project-graph-builder.js:191:9)
    at ProjectGraphBuilder.addStaticDependency (nx/dist/src/project-graph/project-graph-builder.js:90:14)
    at nx/dist/src/plugins/js/lock-file/project-graph-pruning.js:115:17
    at traverseNode (nx/dist/src/plugins/js/lock-file/project-graph-pruning.js:112:36)
```

Nx swallows the error and writes the **full workspace lockfile** into `dist/apps/<app>/`. That is not benign for Docker builds: the generated `package.json` has the pruned dependency set while the emitted lockfile has the root one, so the image stage fails with

```
ERR_PNPM_OUTDATED_LOCKFILE: Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/package.json
  * 72 dependencies were removed: ...
```

Preconditions: a package resolves to **more than one version** in the tree, and at least one project prunes down to a single version of it (triggering the rehoist) while another project still needs the multi-version shape.

This looks like the unexplained cause behind #20421, which was closed as `not planned` for lack of a reproduction (note the reporter there also saw it only on the **second** compile, not the first — consistent with a first prune poisoning the shared graph). It is distinct from #34322, which was the `closest === undefined` crash in `rehoistNodes` and is already fixed.

### Expected Behavior

`pruneProjectGraph` should be a pure function of `(graph, packageJson)`. Pruning for project A must not change the result of pruning for project B, and the caller-supplied `ProjectGraph` must not be mutated.

### GitHub Repo

_No response_ — the reproduction below is standalone and needs no workspace.

### Steps to Reproduce

Save as `repro.js` in any directory where `nx` is installed, then `node repro.js`:

```js
const { pruneProjectGraph } = require("nx/src/plugins/js/lock-file/project-graph-pruning");

const ext = (name, packageName, version) => ({
  type: "npm",
  name,
  data: { version, packageName, hash: `sha512-${packageName}${version}` },
});

const graph = {
  nodes: {},
  externalNodes: {
    "npm:cookie": ext("npm:cookie", "cookie", "2.0.1"),
    "npm:cookie@1.1.1": ext("npm:cookie@1.1.1", "cookie", "1.1.1"),
    "npm:light-my-request": ext("npm:light-my-request", "light-my-request", "6.6.0"),
  },
  dependencies: {
    "npm:light-my-request": [
      { source: "npm:light-my-request", target: "npm:cookie@1.1.1", type: "static" },
    ],
  },
};

// app-b does NOT depend on cookie directly -> 1.1.1 is the only version left -> rehoisted
const appB = { name: "app-b", dependencies: { "light-my-request": "6.6.0" } };
// app-a depends on cookie 2.0.1 directly -> both versions must stay
const appA = { name: "app-a", dependencies: { cookie: "2.0.1", "light-my-request": "6.6.0" } };

const nameOf = () => graph.externalNodes["npm:cookie@1.1.1"].name;
console.log("before         :", nameOf());

pruneProjectGraph(graph, appB);
console.log("after prune(B) :", nameOf(), " <-- source graph mutated");

try {
  pruneProjectGraph(graph, appA);
  console.log("after prune(A) : OK");
} catch (e) {
  console.log("after prune(A) : THROWS ->", e.message);
}
```

Actual output on nx 23.1.0:

```
before         : npm:cookie@1.1.1
after prune(B) : npm:cookie  <-- source graph mutated
after prune(A) : THROWS -> Target project does not exist: npm:cookie@1.1.1
```

Expected: `prune(A)` succeeds, and `nameOf()` still reports `npm:cookie@1.1.1` after `prune(B)`.

In a real workspace this surfaces as `nx run-many --target=build` (or `nx build <app>` with `generatePackageJson`) succeeding for the first project and failing pruning for later ones. Ordering decides which project breaks: pruning `[api, worker, scheduler]` fails on `scheduler`, and pruning `[scheduler, worker, api]` on the same graph fails on `worker` and `api`.

### Nx Report

```shell
Node           : 24.18.0
OS             : win32-x64
pnpm           : 11.14.0
daemon         : Available

nx                : 23.1.0
@nx/js            : 23.1.0
@nx/eslint        : 23.1.0
@nx/workspace     : 23.1.0
@nx/jest          : 23.1.0
@nx/devkit        : 23.1.0
@nx/eslint-plugin : 23.1.0
@nx/nest          : 23.1.0
@nx/node          : 23.1.0
@nx/vite          : 23.1.0
@nx/vitest        : 23.1.0
@nx/web           : 23.1.0
@nx/webpack       : 23.1.0
@nx/docker        : 23.1.0
typescript        : 6.0.2
```

Also reproduces on `linux-x64` (GitHub Actions runner, Docker build).

### Failure Logs

_See above._

### Package Manager Version

pnpm 11.14.0

### Operating System

- [ ] macOS
- [X] Linux
- [X] Windows
- [ ] Other (Please specify)

### Additional Information

The mutation is in `switchNodeToHoisted` (`nx/src/plugins/js/lock-file/project-graph-pruning.ts`):

```js
function switchNodeToHoisted(node, builder, invBuilder) {
    const previousName = node.name;
    ...
    builder.removeNode(node.name);
    invBuilder.removeNode(node.name);
    // modify the node and re-add it
    node.name = `npm:${node.data.packageName}`;   // <-- mutates the node owned by the source graph
    builder.addExternalNode(node);
    invBuilder.addExternalNode(node);
    ...
}
```

`traverseNode` passes the node straight from `graph.externalNodes` into `builder.addExternalNode(node)`, so builder and source graph share the same object.

**Suggested fix** — rehoist onto a copy instead of mutating in place:

```js
const hoisted = { ...node, name: `npm:${node.data.packageName}` };
builder.addExternalNode(hoisted);
invBuilder.addExternalNode(hoisted);
```

and use `hoisted.name` for the subsequent `addStaticDependency` calls. Deep-copying the graph on entry to `pruneProjectGraph` would also work but is more expensive on large graphs.

I verified `23.2.0-beta.2` and `23.2.0-canary.20260724-9c735d4` ship an identical `project-graph-pruning.js`, and patching either build in still reproduces, so this is not fixed on the current pre-release line.

ATKasem:
## Current Behavior

`pruneProjectGraph` mutates the `ProjectGraph` it is handed, so pruning one project changes the result of pruning the next.

`switchNodeToHoisted` renames the node in place:

```ts
node.name = `npm:${node.data.packageName}`;
```

but `traverseNode` adds nodes to the builder **by reference** from the caller's graph (`builder.addExternalNode(node)`, no copy). The rename therefore lands on the source graph, which is left holding a node keyed `npm:<pkg>@<version>` whose `name` is now `npm:<pkg>`. Every later prune against that same graph object resolves the node by key, registers it under the mutated `name`, and the dependency edge no longer resolves:

```
NX   An error occurred while creating pruned lockfile
Original error: Target project does not exist: npm:cookie@1.1.1
```

The preconditions are a package resolving to more than one version, where at least one project prunes down to a single version (triggering the rehoist) while another still needs the multi-version shape.

This is worse than a thrown error, because Nx swallows it and writes the **full workspace lockfile** into the build output. The generated `package.json` then has the pruned dependency set while the lockfile has the root one, so Docker builds fail later and further from the cause:

```
ERR_PNPM_OUTDATED_LOCKFILE: Cannot install with "frozen-lockfile" because
```

It also explains why this is typically seen on the *second* build rather than the first.

## Expected Behavior

`pruneProjectGraph` is a pure function of `(graph, packageJson)`. Pruning for project A does not change the result of pruning for project B, and the caller-supplied `ProjectGraph` is never mutated.

## What changed

`switchNodeToHoisted` now re-adds the node under the hoisted name as a **new object** rather than renaming the shared one:

```ts
const hoistedNode: ProjectGraphExternalNode = {
  ...node,
  name: `npm:${node.data.packageName}`,
};
```

`node.name` on that line was the only mutation of caller-owned state in the file (I grepped for others), so this is sufficient to make the function pure with respect to its `graph` argument. Behaviour inside a single prune is unchanged — the same node content is registered under the same hoisted name, and the dependency rewiring just uses `hoistedNode.name`.

## Verification

Three regression tests added from the reproduction in the issue, all of which fail on `master`:

```
● when a rehoist happens › does not mutate the source graph
● when a rehoist happens › stays deterministic when the same graph is pruned repeatedly
● when a rehoist happens › keeps both versions resolvable after an earlier prune rehoisted one

    expect(received).not.toThrow()
    Error message: "Target project does not exist: npm:cookie@1.1.1"

Tests:       3 failed, 26 skipped, 4 passed, 33 total
```

That last message is the reported error verbatim. With the fix:

```
$ npx jest --config packages/nx/jest.config.cts --testPathPatterns "project-graph-pruning"
Tests:       33 passed, 33 total
```

And the whole lock-file suite, since this function is shared by every parser:

```
$ npx jest --config packages/nx/jest.config.cts --testPathPatterns "lock-file"
Test Suites: 6 passed, 6 total
Tests:       216 passed, 216 total
Snapshots:   61 passed, 61 total
```

`prettier` reports both changed files unchanged.

I ran the focused package suites rather than `nx affected`, because the local `@nx/dotnet` and `@nx/gradle` plugins fail to create nodes without `dotnet`/`gradle` installed, which drowns the run in unrelated errors. Happy to add anything else CI flags.

## Related Issue(s)

Fixes #36470

This also looks like the mechanism behind #20421, which was closed as `not planned` for lack of a reproduction — note that reporter also saw it only on the second compile, consistent with a first prune poisoning the shared graph. It is distinct from #34322 (the `closest === undefined` crash in `rehoistNodes`), which is already fixed.

### DIFF
--- a/packages/nx/src/plugins/js/lock-file/project-graph-pruning.spec.ts
+++ b/packages/nx/src/plugins/js/lock-file/project-graph-pruning.spec.ts
@@ -912,5 +912,94 @@ describe('project-graph-pruning', () => {
       expect(prunedGraph.externalNodes?.['npm:lodash@4.17.21']).toBeDefined();
       expect(prunedGraph.externalNodes?.['npm:lodash@4.17.20']).toBeDefined();
     });
+
+    describe('when a rehoist happens', () => {
+      const externalNode = (
+        name: string,
+        packageName: string,
+        version: string
+      ) => ({
+        type: 'npm' as const,
+        name,
+        data: { version, packageName, hash: `sha512-${packageName}${version}` },
+      });
+
+      // A package resolved to two versions, with the nested one reachable only
+      // through another dependency. Pruning to just that dependency leaves 1.1.1
+      // as the sole version, which triggers the rehoist.
+      const createGraph = (): ProjectGraph => ({
+        nodes: {},
+        externalNodes: {
+          'npm:cookie': externalNode('npm:cookie', 'cookie', '2.0.1'),
+          'npm:cookie@1.1.1': externalNode(
+            'npm:cookie@1.1.1',
+            'cookie',
+            '1.1.1'
+          ),
+          'npm:light-my-request': externalNode(
+            'npm:light-my-request',
+            'light-my-request',
+            '6.6.0'
+          ),
+        },
+        dependencies: {
+          'npm:light-my-request': [
+            {
+              source: 'npm:light-my-request',
+              target: 'npm:cookie@1.1.1',
+              type: 'static' as any,
+            },
+          ],
+        },
+      });
+
+      // Prunes down to a single cookie version, so the rehoist runs.
+      const rehoistingPackageJson: PackageJson = {
+        name: 'app-b',
+        version: '1.0.0',
+        dependencies: { 'light-my-request': '6.6.0' },
+      };
+
+      // Needs both cookie versions, so it depends on the source graph being intact.
+      const multiVersionPackageJson: PackageJson = {
+        name: 'app-a',
+        version: '1.0.0',
+        dependencies: { cookie: '2.0.1', 'light-my-request': '6.6.0' },
+      };
+
+      it('does not mutate the source graph', () => {
+        const graph = createGraph();
+
+        pruneProjectGraph(graph, rehoistingPackageJson);
+
+        expect(graph.externalNodes['npm:cookie@1.1.1'].name).toEqual(
+          'npm:cookie@1.1.1'
+        );
+        expect(graph.externalNodes['npm:cookie'].name).toEqual('npm:cookie');
+      });
+
+      it('stays deterministic when the same graph is pruned repeatedly', () => {
+        const graph = createGraph();
+
+        const first = pruneProjectGraph(graph, multiVersionPackageJson);
+        pruneProjectGraph(graph, rehoistingPackageJson);
+        const third = pruneProjectGraph(graph, multiVersionPackageJson);
+
+        expect(Object.keys(third.externalNodes ?? {}).sort()).toEqual(
+          Object.keys(first.externalNodes ?? {}).sort()
+        );
+        expect(third.dependencies).toEqual(first.dependencies);
+      });
+
+      it('keeps both versions resolvable after an earlier prune rehoisted one', () => {
+        const graph = createGraph();
+
+        pruneProjectGraph(graph, rehoistingPackageJson);
+
+        expect(() =>
+          pruneProjectGraph(graph, multiVersionPackageJson)
+        ).not.toThrow();
+      });
+    });
   });
 });
--- a/packages/nx/src/plugins/js/lock-file/project-graph-pruning.ts
+++ b/packages/nx/src/plugins/js/lock-file/project-graph-pruning.ts
@@ -280,18 +280,24 @@ function switchNodeToHoisted(
   builder.removeNode(node.name);
   invBuilder.removeNode(node.name);
 
-  // modify the node and re-add it
-  node.name = `npm:${node.data.packageName}`;
-  builder.addExternalNode(node);
-  invBuilder.addExternalNode(node);
+  // Re-add under the hoisted name as a new object. The node is shared by
+  // reference with the caller's graph, so renaming it in place would leave that
+  // graph with a node keyed `npm:<pkg>@<version>` but named `npm:<pkg>`, and
+  // every later prune of the same graph would fail to resolve the edge.
+  const hoistedNode: ProjectGraphExternalNode = {
+    ...node,
+    name: `npm:${node.data.packageName}`,
+  };
+  builder.addExternalNode(hoistedNode);
+  invBuilder.addExternalNode(hoistedNode);
 
   targets.forEach((target) => {
-    builder.addStaticDependency(node.name, target);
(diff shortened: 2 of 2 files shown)


leosvelperez (approved):
Thanks!

PR #36527, opened 2026-07-30
feat(dotnet): split .NET test targets into one task per test class or method

mikehaas763:
## Current Behavior

A .NET test project gets a single `test` target, so a long suite caches as one unit and runs on one agent. `@nx/playwright`, `@nx/jest`, `@nx/vitest`, `@nx/cypress` and `@nx/gradle` all split their test target into one task per test unit. `@nx/dotnet` has no equivalent.

## Expected Behavior

Naming a CI target produces one target per test class (or method), plus an `nx:noop` parent carrying `nonAtomizedTarget`.

```jsonc
// nx.json
{
  "plugins": [
    {
      "plugin": "@nx/dotnet",
      "options": { "test": { "ciTargetName": "test-ci" } }
    }
  ]
}
```

`ciTargetName` enables splitting and is unset by default. `ciGroupName` overrides the group label. `ciSplitBy` chooses `class` (default) or `method`.

Discovery is a syntax-only Roslyn parse of the project's MSBuild `Compile` items, so it needs no build. Roslyn is excluded at runtime like the MSBuild packages and loaded from the SDK that MSBuildLocator selects, so nothing new ships in the npm package. If that SDK has no Roslyn, splitting is unavailable and projects keep their ordinary `test` target.

Covers MSTest on Microsoft.Testing Platform. MTP hasn't standardized filtering across frameworks yet (microsoft/testfx#7160), so tasks use the syntax MSTest accepts. Projects that don't qualify keep their `test` target.

The analyzer cache now also keys on the C# sources these targets derive from.

## Related Issue(s)

N/A


packages/dotnet/analyzer/Utilities/TestClassScanner.cs
```diff
@@ -0,0 +1,383 @@
...
+        }
+
+        // Generic classes are name-mangled in both filter syntaxes.
+        if (declaration.TypeParameterList is not null)
+        {
+            exclusions.Generic++;
+            return false;
+        }
+
+        return true;
```
AgentEnder, line 422: These are the only exclusions, so any `[TestClass]` that won't actually run a test still gets its own
target. Checked against MSTest 3.6.4 on MTP with the args `BuildAtomizedArgs` emits:

| class shape | exit |
| --- | --- |
| ordinary test class | 0 |
| `[TestClass]` with only `[AssemblyInitialize]` | 1 (`Total: 0`) |
| `internal [TestClass]` | 1 (`Total: 0`) |
| `[TestClass, Ignore]` | 1 (`Skipped: 1`) |

The fixture case isn't rare. MSTest only runs `[AssemblyInitialize]` if its declaring class has
`[TestClass]`, so any suite with an assembly fixture hits it. `[Ignore]` exits 1 even without
`--minimum-expected-tests`. Today all three are one passing `test` target, so this turns a green
suite red.

Three checks would cover it:

- class mode: only yield when the declaration has a `[TestMethod]`/`[DataTestMethod]` or a
  `BaseList`, unioned across partial halves the way `DoNotParallelize` already is
- skip `[Ignore]` on classes and methods
- require `public` unless a source declares `[assembly: DiscoverInternals]`, using the same pre-pass
  as `DeclaresAssemblyDoNotParallelize`

packages/dotnet/analyzer/Utilities/TestClassScanner.cs
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    /// <summary>
+    /// Builds the dotted namespace for a declaration, joining nested namespace
+    /// blocks outermost-first. Handles both block and file-scoped forms.
+    /// </summary>
+    private static string GetNamespace(SyntaxNode node) =>
+        string.Join('.', node.Ancestors()
+            .OfType<BaseNamespaceDeclarationSyntax>()
+            .Select(ns => ns.Name.ToString())
```
AgentEnder: `ToString()` returns the raw source span including interior trivia, so whatever sits between the name
tokens ends up in the namespace. `ClassName`/`MethodName` are fine, they use `Identifier.ValueText`.
This value feeds the filter, the results directory, the target name and the declared outputs.

| source | resulting filter |
| --- | --- |
| namespace wrapped over lines | contains newlines |
| `namespace Acme // note` then `.Tests;` | comment lands in the filter |
| `namespace Acme . Tests;` | `ClassName=Acme . Tests.T`, matches nothing, fails every run |
| `namespace Company.@base;` | `@` isn't in the runtime name, never matches |

It also gets joined into the shell command run-commands executes, so normalizing here closes that off
too. `SyntaxKind` is already imported.

```suggestion
            .Select(ns => string.Join('.', ns.Name.DescendantTokens()
                .Where(t => t.IsKind(SyntaxKind.IdentifierToken))
                .Select(t => t.ValueText)))
```

Worth a guard before a unit becomes a target as well: skip and tally any `Id` that isn't a dotted
identifier sequence.

packages/dotnet/analyzer/Models/TestUnit.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// The <c>=</c> operator is an exact match rather than a prefix, which keeps
+    /// <c>LoginTest</c> and <c>LoginTestWithMfa</c> in separate tasks. MSTest
+    /// folds data rows under the method's own FullyQualifiedName, so one task
+    /// runs all of a <c>[DataRow]</c> method's cases.
+    ///
+    /// <c>ClassName</c> requires the namespace; MSTest matches nothing for a
+    /// bare class name.
+    ///
+    /// Values are quoted for the shell Nx runs commands through. A C# identifier
+    /// cannot contain a double-quote, so no escaping is needed inside them.
```
AgentEnder: The premise is true of identifiers, but the value being quoted isn't one. `ClassFqn` includes the
namespace, which `GetNamespace` builds with `ns.Name.ToString()` (raw syntax text, see the other
comment). Worth updating alongside that fix.

packages/dotnet/analyzer/Utilities/TargetBuilder.TestAtomizer.cs
```diff
@@ -0,0 +1,187 @@
...
+        string fileName)
+    {
+        var ciTargetName = options.TestCiTargetName!;
+        var units = discovery.Units;
+
+        // A no-op parent with no dependencies would pass while running nothing.
+        if (units.Count == 0)
+        {
+            return null;
+        }
```
AgentEnder: This returns before `ReportExclusions`, so zero units means no target and no message.

That path is reachable without an empty project. With an abstract `[TestClass]` base holding the
`[TestMethod]`s and a concrete `[TestClass]` subclass, class mode gives 1 unit but method mode gives
0, with `SkippedNested` and `SkippedGeneric` both 0, because `ScanSource` only walks
`declaration.Members` and nothing was actually excluded. So a `ciSplitBy: method` user with a shared
test base gets no `test-ci` target and no explanation.

Worth reporting when discovery finds classes but produces no units.

packages/dotnet/analyzer/Utilities/TargetBuilder.TestAtomizer.cs
```diff
@@ -0,0 +1,187 @@
...
+        var (nxBaseDirectory, cwdRelativeBase) = resultsBase.Value;
+
+        ReportExclusions(discovery, projectName, options.TestTargetName);
+
+        var technologies = ProjectUtilities.GetTechnologies(fileName);
+        var groupName = options.TestCiGroupName ?? $"{ciTargetName.ToUpperInvariant()} (CI)";
+        var dependsOn = new List<object>();
+        var groupMembers = new List<string>();
+
+        foreach (var unit in units)
```
AgentEnder, line 74: This mutates the caller's `targets` as it goes, so if anything threw partway through the loop the
`catch` in `BuildTargets` would leave orphan `test-ci--*` leaves with no parent and no group.
Building into a local dictionary and merging on success would make that catch clean.

packages/dotnet/analyzer/Utilities/TargetBuilder.TestAtomizer.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        targets[ciTargetName] = new Target
+        {
+            Executor = "nx:noop",
+            Cache = baseTestTarget.Cache,
+            Inputs = baseTestTarget.Inputs,
+            Outputs = baseTestTarget.Outputs,
+            DependsOn = [.. dependsOn],
+            // Otherwise Nx could schedule the parent alongside its own tasks.
+            Parallelism = units.All(unit => unit.DoNotParallelize) ? false : null,
```
AgentEnder: `dependsOn` already contains every leaf, so the graph orders the parent after all of them regardless
of `parallelism`. The line itself is right and matches `@nx/playwright`, it's the stated reason
that's off.

packages/dotnet/analyzer/Utilities/TestClassScanner.cs
```diff
@@ -0,0 +1,383 @@
...
+                // Excluded rather than guessed at.
+                if (!HasAttribute(method.AttributeLists, "TestMethod", "DataTestMethod"))
+                {
+                    continue;
+                }
+
+                if (method.TypeParameterList is not null)
+                {
+                    exclusions.Generic++;
+                    continue;
```
AgentEnder: This comment explains why generic *methods* are excluded, but it sits above the attribute check. It
belongs above the `TypeParameterList` check below it.

packages/dotnet/analyzer/Utilities/TargetBuilder.PathHelpers.cs
```diff
@@ -296,4 +296,64 @@ private static string StripConfiguration(string path)
...
+                .GetRelativePath(workspaceRoot, projectDirectory)
+                .Replace('\\', '/');
+            var upToWorkspaceRoot = projectRelativeToWorkspace is "." or ""
+                ? string.Empty
+                : string.Concat(Enumerable.Repeat("../", projectRelativeToWorkspace.Split('/').Length));
+
+            return (baseDirectory, $"{upToWorkspaceRoot}{baseDirectory[workspaceRootToken.Length..]}");
+        }
+
+        return null;
```
AgentEnder, line 355: Also returns null when `baseDirectory` is exactly `{projectRoot}` or `{workspaceRoot}` with no
trailing segment, since both `StartsWith` checks include the trailing slash. The caller then prints
"resolves outside the workspace", which is the wrong reason in that case.

packages/dotnet/analyzer/Analyzer.cs
```diff
@@ -292,7 +314,15 @@ private static Dictionary<string, string> CollectProperties(ProjectInstance proj
...
-            "TestResultsDirectory"
+            "TestResultsDirectory",
+
+            // Microsoft.Testing.Platform. Splitting a test target relies on the
+            // platform's filtering options, so these gate whether a project can
+            // be atomized at all. MSTest.Sdk is not detectable as an SDK
+            // attribute, but it sets EnableMSTestRunner, so it is covered here.
+            "EnableMSTestRunner",
+            "TestingPlatformDotnetTestSupport",
+            "UseMicrosoftTestingPlatformRunner"
```
AgentEnder: `UseMicrosoftTestingPlatformRunner` is collected but never read. `SupportsTestSplitting` gates on the
other two only, so the comment describes three gates where there are two. Either wire it in or drop
it, and adjust the comment either way.

packages/dotnet/src/plugins/create-nodes.ts
```diff
@@ -103,10 +149,54 @@ export interface DotNetPluginOptions {
...
+ * `test-ci--*` leaf behind, all pointing at a target that no longer exists.
+ * The group is located by its `nonAtomizedTarget` back-reference rather than by
+ * name, since the disabling option carries no `ciTargetName` to look up.
+ */
+function removeAtomizedTargetsFor(
+  targets: Record<string, TargetConfiguration>,
+  metadata: ProjectConfiguration['metadata'],
+  nonAtomizedTargetName: string
+): ProjectConfiguration['metadata'] {
+  const parentName = Object.keys(targets).find(
```
AgentEnder: I don't think this is reachable. `createNodes` sets
`testOptions = normalizedOptions.test === false ? undefined : ...`, so `ciTargetName` is undefined
exactly when `test: false`, and the analyzer is never given `testCiTargetName`. No parent target and
no `metadata.nonAtomizedTarget` get emitted for this to find. Options are per-registration and the
node always comes from `analyzeProjects` with those same options.

Separately, the lookup key is `defaultTargetName` (`'test'`) while the analyzer writes
`NonAtomizedTarget = options.TestTargetName`, so it wouldn't match under a custom `targetName`
either.

The four specs covering it assert on nodes the plugin can't produce. Since `TestTargetConfiguration`
already makes `test: false` and `ciTargetName` mutually exclusive, I'd drop the helper, the call site
and those specs.

packages/dotnet/src/analyzer/analyzer-client.ts
```diff
@@ -75,6 +120,65 @@ async function calculateProjectFilesHash(files: string[]): Promise<string> {
...
+  if (!result.atomizedRoots?.length) {
+    return null;
+  }
+
+  return await hashWithWorkspaceContext(workspaceRoot, [
+    ...result.atomizedRoots.map((root) => rootGlob(root, '**/*.cs')),
+    // Discovery reads MSBuild's Compile items, which can point outside the
+    // project. A linked test file would otherwise change the discovered units
+    // without changing this hash.
+    ...(result.atomizedExternalSources ?? []),
```
AgentEnder: These reach `hashWithWorkspaceContext` as glob patterns, not literal paths, so a source path
containing `*`, `!` or `{}` would silently change which files get hashed. Worth escaping the
metacharacters before adding them.

packages/dotnet/analyzer.Tests/SupportsTestSplittingTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    [Fact]
+    public void BothProperties_AreRequiredAndSufficient()
+    {
+        Assert.True(Analyzer.SupportsTestSplitting(
+            Props((Runner, "true"), (DotnetTestSupport, "true"))));
+    }
+
+    [Fact]
+    public void MsTestSdk_IsCoveredByThoseProperties()
+    {
```
AgentEnder: This is identical in inputs and assertion to `BothProperties_AreRequiredAndSufficient` above it.
There's no MSTest.Sdk-specific input, so it can't fail unless that one does either. Probably better
as a comment on the first test.

astro-docs/src/content/docs/kb/split-dotnet-tests.mdoc
```diff
@@ -0,0 +1,197 @@
...
+target.
+
+That project now has a `test-ci--<class>` target per test class, plus a `test-ci` target that
+depends on all of them:
+
+```shell
+nx show project IntegrationTests
+```
+
+Run `nx test-ci IntegrationTests` and Nx expands it into the individual tasks.
```
AgentEnder: Following the page in order, this is the first thing the reader runs, and it exits 1. Line 62 further
down explains that Nx refuses the parent target without Nx Cloud
(`run-command.ts` -> `validateNoAtomizedTasks`).

Either point this at a single task (the form shown at line 69) or move the Nx Cloud note above the
walkthrough.

astro-docs/src/content/docs/kb/split-dotnet-tests.mdoc
```diff
@@ -0,0 +1,197 @@
...
+Nested and generic test classes are not split out, because their names have no stable command-line
+form to filter on. Generic test methods are excluded for the same reason under `ciSplitBy: method`.
+Those tests still run as part of the ordinary `test` target, and the plugin reports how many it left
+out while building the project graph, so the gap is visible up front.
+
+An abstract test class isn't split either, but that costs you nothing — the concrete classes that
+derive from it are what carry tests, and each of those gets its own task.
+
+Tests declared on a base class and inherited by a subclass that carries no `[TestClass]` of its own
+are also missed. If you rely on that pattern, keep the project on its `test` target.
```
AgentEnder: Both sentences are wrong under `ciSplitBy: method`, which this page also documents. With an abstract
`[TestClass]` base holding the `[TestMethod]`s and a concrete `[TestClass]` subclass, method mode
produces 0 units, so the subclass gets no task at all rather than one task per inherited method. The
miss also isn't limited to a subclass without its own `[TestClass]`.

It's not reported either, since the skip counts stay 0, so the "visible up front" promise a few lines
above doesn't hold here.

e2e/dotnet/src/dotnet-atomizer.test.ts
```diff
@@ -0,0 +1,270 @@
...
+    });
+  });
+
+  describe('Nx Cloud requirement', () => {
+    it('should refuse to run the parent without Nx Cloud', () => {
+      // Splitting a suite across a single machine is slower than not splitting
+      // it, so Nx core hard-exits rather than letting it run.
+      const output = runCLI(`run ${project}:test-ci`, { silenceError: true });
+
+      expect(output).toContain('Nx Cloud');
```
AgentEnder: Plenty of unrelated Nx output contains "Nx Cloud", so this passes for the wrong reasons. Matching the
actual message would be tighter, something like
`The ${project}:test-ci task should only be run with Nx Cloud.`

AgentEnder (changes requested):
Thanks for this. Solid work, and the analyzer side is well thought through.

Two things I'd want fixed before merge, both in `TestClassScanner`: discovery emits targets for
classes that can never pass, and the namespace is built from raw syntax text so ordinary formatting
breaks the filter. Details are inline, along with some smaller notes.

I checked these against a real MSTest 3.6.4 on Microsoft.Testing.Platform project (SDK 9.0.316),
running the argument vector `BuildAtomizedArgs` emits, and against the shipped
`TestClassScanner.ScanSources` on the same sources.

Two questions:

1. Is a task whose filter matches nothing meant to be a diagnostic, per the e2e case? That decides
   whether the discovery issue is a fix or a documented limitation.
2. `GroupNameFallsBackWhenNoneIsProvided` pins `TEST-CI (CI)`, which the plugin can't produce since
   `create-nodes.ts` always sends a derived name. Keep the fallback reachable for direct analyzer
   use, or make `TestCiGroupName` required?

One rebase note: the `deriveGroupNameFromTarget` import will need to move to `@nx/devkit/internal`.
It's correct as written on this branch, master just moved since you branched.

mikehaas763:
Thanks for the review @AgentEnder!

> 1. Is a task whose filter matches nothing meant to be a diagnostic, per the e2e case? That decides
>    whether the discovery issue is a fix or a documented limitation.

Yeah, it was meant to be a diagnostic. I pushed a commit to fix discovery of the test units instead of documenting it as a limitation, then updated the tests accordingly.

> 2. `GroupNameFallsBackWhenNoneIsProvided` pins `TEST-CI (CI)`, which the plugin can't produce since
>    `create-nodes.ts` always sends a derived name. Keep the fallback reachable for direct analyzer
>    use, or make `TestCiGroupName` required?

~Idk, I suppose I could drop it, I don't think there's any intention for direct analyzer use ever right?~

I decided to just drop the fallback, seemed like the cleaner fix for now

packages/dotnet/analyzer/Utilities/TestClassScanner.cs
mikehaas763: fixed, excludes empty/ignored/non-public now

packages/dotnet/analyzer/Utilities/TargetBuilder.TestAtomizer.cs
mikehaas763: builds local, merges on success now

packages/dotnet/src/analyzer/analyzer-client.ts
mikehaas763: hashed by exact path now, no more glob

packages/dotnet/analyzer.Tests/SupportsTestSplittingTests.cs
mikehaas763: migrated into a comment on the first test

astro-docs/src/content/docs/kb/split-dotnet-tests.mdoc
mikehaas763: points at a single task now

astro-docs/src/content/docs/kb/split-dotnet-tests.mdoc
mikehaas763: broadened, and it's reported now too

e2e/dotnet/src/dotnet-atomizer.test.ts
mikehaas763: pinned to the real message

packages/dotnet/analyzer/Utilities/TargetBuilder.TestAtomizer.cs
mikehaas763: moved it earlier, plus a new count for the inherited-only case

packages/dotnet/analyzer/Utilities/TestClassScanner.cs
mikehaas763: identifier tokens now, and an Id that isn't a dotted sequence gets skipped and tallied

packages/dotnet/analyzer/Utilities/TargetBuilder.TestAtomizer.cs
mikehaas763: dropped the comment

packages/dotnet/analyzer/Utilities/TestClassScanner.cs
mikehaas763: moved down to the TypeParameterList check

packages/dotnet/src/plugins/create-nodes.ts
mikehaas763: dropped the helper and call site and the four specs

packages/dotnet/analyzer/Analyzer.cs
mikehaas763: dropped it, nothing was reading it

packages/dotnet/analyzer/Models/TestUnit.cs
mikehaas763: updated

packages/dotnet/analyzer/Utilities/TargetBuilder.PathHelpers.cs
mikehaas763: fixed

mikehaas763:
@AgentEnder would you wanna release this with only the mstest support to start, as it's currently written? Or would you wanna see xunit and/or others added before this goes in? It should be relatively easy and straightforward to add additional testing frameworks after with separate PRs but maybe xunit is one you wouldn't wanna release without support right away?

mikehaas763:
@AgentEnder will the pipeline release a preview nx build from this branch? Just wondering because I wouldn't mind testing it out that way as well

AgentEnder:
Yes, I'm kicking one off now, so there'll be a preview build from this branch shortly and you can try it against a real workspace.

Heads up on two other things while you're here. CI had never actually run on this PR until today: it's a fork PR and the workflow approval gate was never cleared, so every round so far was reviewed with no build, test or e2e signal. That's on us. It's green now.

And I've pushed a commit to your branch (`2252709754`) rather than sending you another round for it. Full notes to follow in a moment.

mikehaas763:
Thanks for publishing a build! I'll give this a test this weekend and get back here

mikehaas763:
@AgentEnder I tested this out in one of my workspaces, it didn't work right away out of the box 😬. No test splitting was happening. Once I applied a small workaround to my workspace it split! The issue is the check to detect whether the MSTest project is on MTP was a little too narrow as of v4. I pushed a commit to broaden that check just a bit. I also resolved some conflicts that were merged in upstream recently. Should be good to go again!

AgentEnder:
Thanks for the PR. Two things. The branch conflicts with master now, so it needs a rebase before anyone can read it against current code, and the head commit had no approved workflow run, so I have approved it.

<details>
<summary>

</summary>

> Added `blocked: needs rebase` because `mergeStateStatus` is DIRTY against current master.
>
> The head commit here has no approved workflow run, so no GitHub Actions check runs exist against it. On a fork pull request the run waits for a maintainer to approve it, and until that happens the green tick comes from third-party checks only, so it says nothing about the build. Approved as part of triage.
>
> No linked issue, but there is an open review here already, so the change is not unprompted and the unlinked-PR rule does not apply.
>
> The conflict is not surprising. #36958, #36981, #36804 and #36788 have all landed in `packages/dotnet` since this branch was cut, and #36991 is in flight across fourteen of the same files.

</details>

mikehaas763:
@AgentEnder I resolved those conflicts

PR #36554, opened 2026-08-04
fix(release): use the release group name when resolving the previous tag for fixed groups

ShwethaSundar:
## Current Behavior

For a fixed release group, the changelog resolves the previous git tag by interpolating `releaseTag.pattern` without the release group name. `{releaseGroupName}` is left unresolved and becomes a match-anything wildcard, so the latest tag of any release group can be selected. Workspaces with multiple configured release groups hit this by default, since `{releaseGroupName}-v{version}` is the default pattern for a fixed group.

## Expected Behavior

The release group name is interpolated into the pattern, so only that group's own tags match. This is what tag creation and the versioning path already do.


ShwethaSundar:
@JamesHenry requesting review.

leosvelperez (changes requested):
Thanks for this, the fix is correct. It matches what tag creation (`utils/shared.ts`), the versioning read (`utils/release-graph.ts:647`) and the independent-group branch (`changelog.ts:513`) already do, and I confirmed the bug against master: resolving `group-a` with tags `group-b-v2.0.0` and `group-a-v1.0.0` returns `group-b`'s tag.

I've updated the title and description to describe the mechanism.

One thing before we merge: could you add a unit test? `changelog.spec.ts` already has the shape you need in `should not resolve a changelog from ref for an unversioned dependent project`, which asserts on `resolveChangelogFromSHA`'s `tagPatternValues`. A fixed-group equivalent (set `projectsRelationship: 'fixed'`, assert `tagPatternValues` is `{ releaseGroupName: '...' }`) fails on master and passes with your change.

mpsanchis:
thanks @ShwethaSundar !

JamesHenry:
@leosvelperez @ShwethaSundar FYI I'm not sure why flake is affecting this one so badly - I have an agent loop monitoring and rerunning the main CI job when a failure comes in so no action needed from you until further notice

leosvelperez (approved):
Thanks!

PR #36565, opened 2026-08-04
fix(core): separate daemon runtime env from graph identity

### LINKED ISSUE #36564: Yarn Berry temporary PATH invalidates an unchanged daemon graph
## Current Behavior

“Berry” is Yarn's name for Yarn 2 and later; this reproduction uses Yarn 4.15.0. When Yarn runs a workspace binary, it creates a temporary `xfs-*` directory containing wrappers for the selected `node`, `yarn`, `node-gyp`, and package binaries.

Yarn exposes that runtime directory and puts it first on `PATH`:

```text
BERRY_BIN_FOLDER = /tmp/.../xfs-0a01a6bc
PATH              = <BERRY_BIN_FOLDER><delimiter><existing toolchain PATH>
```

The first `PATH` entry is exactly the current `BERRY_BIN_FOLDER` value. The delimiter is `:` on macOS/Linux and `;` on Windows; every remaining `PATH` entry is the pre-existing toolchain path.

The folder is deleted when the command exits. The next Yarn command receives a different random path even when the workspace, lockfile and available binaries are unchanged.

Two consecutive captures produced different folder names, but:

- the complete remaining `PATH` was byte-for-byte identical;
- both folders contained the same 73 wrapper names; and
- every wrapper SHA-256 fingerprint matched.

The path string changed; the executable environment it represented did not.

The Nx client sends its runtime environment to the persistent daemon. The daemon compares the raw values and calls `invalidateGraphCache()` whenever a non-excluded value differs. Every `yarn nx` invocation therefore changes both `BERRY_BIN_FOLDER` and the first `PATH` segment, causing an unchanged graph to be rebuilt.

Observed daemon log on all three requests:

```text
Graph recompute necessary due to env variable refresh.
Changed keys: PATH, BERRY_BIN_FOLDER
```

The public workflow reproduces this under stock Nx 23.1.1 on Windows, macOS and Ubuntu. In the large workspace that exposed the cost, six unchanged requests took 10.030-16.543 seconds. When the same graph was actually reused, requests completed in 266-270ms.

## Expected Behavior

Nx needs two environment views:

| View | Required behavior |
|---|---|
| Runtime environment | Forward the latest full environment, including the current real Berry folder, so plugin workers can execute Yarn wrappers |
| Graph identity | Compare stable inputs that can change projects, targets, dependencies, SDK selection or compiler behavior |

For graph comparison only, Nx should replace the random Berry value with a stable Yarn marker and remove only the matching temporary `PATH` segment. Every other `PATH` segment and SDK/compiler variable must remain significant.

Yarn and direct invocation must also remain distinct identities. Normalization must not collapse them together.

## GitHub Repo

https://github.com/sdjayna/nx-daemon-reproductions

Successful cross-platform evidence run:
https://github.com/sdjayna/nx-daemon-reproductions/actions/runs/30946807413

## Steps to Reproduce

1. Install the reproduction with Yarn 4.15.0.
2. Run `corepack yarn repro:berry`.
3. The script resets Nx and makes three unchanged `nx show projects` requests.
4. Inspect the emitted `invalidations` array.
5. Observe Berry/`PATH` invalidation before every unchanged request.

## Nx Report

```shell
Node           : 22.22.2
OS             : darwin-arm64
yarn           : 4.15.0
daemon         : Available

nx  : 23.1.1
```

The public workflow additionally reproduces the same invalidation on Node 24.18.0 Windows, macOS and Ubuntu runners.

## Failure Logs

```text
Graph recompute necessary due to env variable refresh.
Changed keys: PATH, BERRY_BIN_FOLDER
```

Windows reports the same two keys in the reverse iteration order: `BERRY_BIN_FOLDER, PATH`.

## Package Manager Version

Yarn 4.15.0

## Operating System

- [x] macOS
- [x] Windows
- [x] Linux

## Additional Information

Draft PR #36565 implements runtime-versus-graph-identity separation:
https://github.com/nrwl/nx/pull/36565

There is no public Nx daemon environment-ignore configuration and no Yarn setting for a stable script bin directory. [yarnpkg/berry#6130](https://github.com/yarnpkg/berry/issues/6130) documents the same temporary Node/`PATH` behavior in another persistent-tool integration.

sdjayna:
## Current Behavior

"Berry" is Yarn's name for Yarn 2 and later; this reproduction uses Yarn 4.15.0. When Yarn runs a workspace binary it creates a fresh temporary `xfs-*` directory for that invocation, writes wrappers into it for the selected `node`, `yarn`, `node-gyp` and package binaries, exports it as `BERRY_BIN_FOLDER`, and prepends it to `PATH`.

The daemon compares the raw client environment against the previous one and invalidates the project graph when any value differs. The Berry folder differs on every invocation by design, so every `yarn nx` command on a completely unchanged workspace pays full graph construction:

```
Graph recompute necessary due to env variable refresh.
Changed keys: PATH, BERRY_BIN_FOLDER
```

Measured on a large workspace: 10.0-16.5s per invocation, against 0.27s from the retained graph when the daemon is reached without Yarn's wrapper environment. Reproduced on stock Nx across Windows, macOS and Ubuntu (workflow linked from #36564).

## Expected Behavior

The daemon separates two concerns that currently share one comparison:

- **Runtime environment**: the daemon and plugin workers always receive the full, latest client environment, including the currently valid Berry wrapper folder. Unchanged.
- **Graph identity**: invalidation is decided on a normalized copy. `normalizeDaemonEnvironmentForGraph` removes the exact `BERRY_BIN_FOLDER` segment from `PATH` (case- and separator-insensitively on win32) and pins `BERRY_BIN_FOLDER` to a sentinel, so two Yarn invocations compare equal while a Yarn invocation and a direct Node invocation remain distinct identities.

The decision happens in `handle-client-env.ts` and unions two normalized diffs: `process.env` before versus after the existing `applyDaemonEnvFromClient` (whose signature, behaviour and generation semantics are unchanged), and the previously applied client environment versus the incoming one. A real toolchain change (a different SDK on `PATH`) still invalidates; lockfile and file changes continue to invalidate through file watching.

Tests: 54/54 in the touched suites, including every pre-existing upstream test unchanged, plus a Berry-rotation test (two environments differing only by `xfs-*` folder produce no invalidation and the new folder is still forwarded to workers) and unit coverage for the normalization and changed-key helpers.

## Related Issue(s)

Fixes #36564


leosvelperez:
@sdjayna thanks for the contribution! Could you please rebase? I merged a fix for a failing e2e that should unblock this one.

meeroslav (approved):

leosvelperez (approved):
Thanks!

Issue #36563, opened 2026-08-04
Windows output tracking duplicates per-directory HANDLE cost across workspace


sdjayna:
## Current Behavior

The Nx daemon starts a workspace-wide output-tracking watcher in addition to its source watcher. On Windows, both use non-recursive per-directory registrations.

This duplicates resource cost for normal source directories and extends watcher cost into Git-ignored directories:

| Directory type | Source watcher | Output watcher | Current HANDLE cost |
|---|:---:|:---:|---:|
| Normal source directory | Yes | Yes | 4 per directory |
| Git-ignored directory | No | Yes | 2 per directory |
| Hard-excluded (`node_modules`, `.git`, `.nx/cache`, `.nx/workspace-data`) | No | No | 0 |

The public fixture is Git-ignored specifically to isolate the output watcher's two-HANDLE slope. The defect itself is workspace-wide duplication, not Git-ignore handling.

Nx uses `notify` 8.2.0. For every Windows directory registration, `notify` retains:

1. a directory HANDLE created by `CreateFileW`, and
2. a completion-semaphore HANDLE created by `CreateSemaphoreW`.

A HANDLE is a Windows kernel resource owned by the process. The count therefore grows with workspace directories and remains allocated for the daemon lifetime.

The public synthetic reproduction proves the slope on native Windows:

| Generated directories | Daemon HANDLEs | Private memory (MB) | Working set (MB) | Graph command |
|---:|---:|---:|---:|---:|
| 0 | 288 | 96.6 | 91.1 | 2,664ms |
| 8,192 | 16,674 | 236.3 | 230.4 | 3,912ms |
| 32,768 | 65,826 | 644.4 | 639.7 | 8,110ms |
| 131,072 | 262,434 | 2,311.2 | 2,303.7 | 26,146ms |

Across all four points:

- HANDLE slope: **2.000008 per directory**, R² effectively 1.0.
- Private-memory slope: **0.0169 MB per directory** (~16.9 KB), R² 0.99999.
- Graph-command slope: **0.180ms per directory**, R² 0.99969.
- The measured daemon PID exited after reset at all four points.

A separate large-workspace field observation reached 161,876 handles and approximately 2.48 GB. It motivated the investigation; the public synthetic result is the independently reproducible proof.

## Expected Behavior

Optional output tracking should not add a second workspace-wide per-directory watcher on Windows. Nx must preserve cache correctness through its existing conservative fallback while leaving source watching unchanged.

[...]

PR #36566, opened 2026-08-04
fix(core): recover from kernel watch-event overflow via rescan

sdjayna:
## Current Behavior

When the kernel cannot deliver every file event, Nx loses the difference silently and permanently.

- On Linux this is live today: an inotify queue overflow arrives as notify's rescan flag and Nx discards it. Forced with `fs.inotify.max_queued_events=16` against Nx's real watcher, a 5,000-file burst delivered 2,026-2,745 per-path events across runs and a 20,000-file burst delivered 7,731 - 45-61% of changes lost with no signal of any kind.
- On Windows the loss mode measured during #36811's review (16.7% under bulk writes with a recursive root) is structurally invisible: notify 8 never raises rescan on Windows and Nx handles rescan on no platform.

In both cases the daemon's view of the workspace diverges from disk and stays wrong until someone diagnoses the mystery and runs `nx reset`.

## Expected Behavior

Overflow becomes visible and bounded, per the triage guidance on #36563 ("#36811's strategy change plus something that makes the event loss either visible or bounded"):

1. The native watcher recognises notify's rescan flag and emits a single rescan event that supersedes the accumulated batch - it is emitted in place of, never alongside, per-path events. The new `rescan` member of the napi `EventType` string enum is additive; the regenerated `index.d.ts` is included.
2. The daemon intercepts a rescan ahead of normal change routing, re-walks the workspace, diffs it against the known file map, and synthesizes exactly the missed create/update/delete events through the same recomputation path ordinary changes take. A rescan that finds no differences preserves the cached graph promise untouched.
3. Recorded output hashes are cleared once, so cache restoration re-verifies instead of trusting possibly stale output state - the cost is a single re-verify per affected task, not the permanent copy-mode fallback of `disableOutputsTracking()`. Because the outputs watcher is also the only reporter of gitignored dotenv edits, its rescan additionally invalidates the graph cache as a fail-safe.

This PR deliberately does not change watch registration: the recovery hangs off each watcher's event pipeline, so it remains correct under any registration strategy - including #36811's recursive root and the single-shared-registration direction discussed in #36810. The per-directory handle slope is removed by #36811, and the two compose - validated by applying #36811's changes onto this branch and running the full suites and measurements on the combined tree (Windows 11 ARM64, real watcher). Footprint on that tree is flat at 3 handles / ~0.2 MB up to 192,206 directories (one-time warm-up of 4 handles / 2.8 MB at the 1,000-directory rung), with watcher startup flat at 1.8s against 34.1s on main at the top rung. With the notify 9 pin also applied (Windows rescan requires notify 9; see below), a genuine `ReadDirectoryChangesW` buffer overflow - a four-thread rename storm against the 16 KB buffer - produced exactly one rescan event. The same recovery on Linux under the current notify 8 pin: exactly one rescan per forced overflow at both burst sizes.

Recovery cost, measured at 129,348 files: 2.2s cold / 0.3s warm (5.2s / 0.8s at double). Before this PR the comparable number does not exist, because there is no recovery.

notify versions: the Linux rescan flag arrives under notify 8, so that path is live with this PR alone. The Windows rescan flag requires notify 9 (notify-rs/notify#964, first in 9.0.0-rc.5); a branch bumping the pin is validated end to end and can follow when notify 9.0.0 stable is released.

Tests: a Rust unit test for rescan superseding the accumulated batch; daemon integration tests proving missed changes are recovered exactly and a no-change rescan preserves the cached graph; an outputs-watcher rescan test. Full Rust suite 576/576, all daemon suites 268/268, watch integration tests 16/16 on Windows 11 ARM64; prepush, lint and typecheck green.

## Related Issue(s)

Addresses the recovery half of #36563; composes with #36811.

### DIFF
--- a/packages/nx/src/daemon/server/handle-outputs-changes.spec.ts
+++ b/packages/nx/src/daemon/server/handle-outputs-changes.spec.ts
@@ -5,6 +5,7 @@ vi.mock('../logger', () => ({
   serverLogger: { watcherLog: vi.fn() },
 }));
 vi.mock('./outputs-tracking', () => ({
+  clearRecordedOutputsHashes: vi.fn(),
   disableOutputsTracking: vi.fn(),
   processFileChangesInOutputs: vi.fn(),
 }));
@@ -26,6 +27,7 @@ describe('handleOutputsChanges', () => {
   let handleOutputsChanges: typeof import('./handle-outputs-changes').handleOutputsChanges;
   let getOutputsWatcherTerminalError: typeof import('./handle-outputs-changes').getOutputsWatcherTerminalError;
   let outputsTracking: {
+    clearRecordedOutputsHashes: Mock;
     disableOutputsTracking: Mock;
     processFileChangesInOutputs: Mock;
   };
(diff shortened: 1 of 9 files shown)


AgentEnder:
Thanks for the rescan recovery work — the recovery half of the Windows watcher fix. We've consolidated it into #36912 with co-author credit, alongside the recursive-root change from #36811 and the notify 9.0.0-rc.5 bump. That combined branch is what's headed for merge.

Closing this as consolidated into #36912.

PR #36578, opened 2026-08-05
fix(core): keep agentic migrations within the Windows command line limit

### LINKED ISSUE #36551: Agentic migrations fail on Windows with “The command line is too long”
### Current Behavior

When running Nx 23 migrations on Windows, agentic migration prompts fail with:

`The command line is too long.`

For example:

```text
→ Running prompt with OpenAI Codex…
The command line is too long.

NX   The agent run ended without a usable handoff
Agent exited with code 1.
```

The deterministic part of the migration completes successfully, but Nx fails when launching Codex for the prompt-based step. This occurs even when running with `--no-validate`, because the migration itself invokes an agentic prompt.

It appears Nx is passing the generated prompt to Codex as a command-line argument, exceeding the Windows command-line length limit. Passing the prompt through stdin or a temporary file would likely avoid the issue.

### Expected Behavior

Agentic migrations are tested and work on Windows 11.

### GitHub Repo

_No response_

### Steps to Reproduce

Run agentic migrations under Windows 11 using:

`npx nx migrate --run-migrations`

or 

`npx nx migrate --run-migrations --no-validate`

### Nx Report

```shell
Node           : 24.19.0
OS             : win32-x64
npm            : 11.17.0
daemon         : Disabled

nx (global)            : 23.1.1
nx                     : 23.1.1
@nx/js                 : 23.1.1
@nx/eslint             : 23.1.1
@nx/workspace          : 23.1.1
@nx/angular            : 23.1.1
@nx/jest               : 23.1.1
@nx/cypress            : 23.1.1
@nx/devkit             : 23.1.1
@nx/eslint-plugin      : 23.1.1
@nx/module-federation  : 23.1.1
@nx/rspack             : 23.1.1
@nx/web                : 23.1.1
@nx/webpack            : 23.1.1
typescript             : 6.0.3
Community plugins:
@ngrx/component-store : 22.0.0-beta.0
@ngrx/effects         : 22.0.0-beta.0
@ngrx/entity          : 22.0.0-beta.0
@ngrx/operators       : 22.0.0-beta.0
@ngrx/router-store    : 22.0.0-beta.0
@ngrx/schematics      : 22.0.0-beta.0
@ngrx/store           : 22.0.0-beta.0
@ngrx/store-devtools  : 22.0.0-beta.0
angular-eslint        : 22.1.0
Cache Usage: 0.00 B / 95.30 GB
```

### Failure Logs

```shell

```

### Package Manager Version

_No response_

### Operating System

- [ ] macOS
- [ ] Linux
- [x] Windows
- [ ] Other (Please specify)

### Additional Information

_No response_

leosvelperez:
## Current Behavior

On Windows, `nx migrate --run-migrations` cannot start the coding agent for a migration step:

```
-> Running prompt with OpenAI Codex...
The command line is too long.

NX   The agent run ended without a usable handoff
Agent exited with code 1.
```

npm installs these agents as `.cmd` shims, which Node cannot run directly, so nx starts them through `cmd.exe`. That caps the command line at 8,191 characters. nx passed both the system prompt and the step instructions as arguments. The system prompt alone runs about 5,000 characters, which leaves under 3,300 for the instructions before any escaping. A migration that logged 20 lines already produces 3,400 characters of instructions. One that logged 500 produces 23,000, or 28,000 once its file changes and advisory notes are counted.

Two escaping bugs sit on the same path. nx caret-escaped `%`, which does nothing, because cmd.exe expands variables before it reads carets, so a `%PATH%` in a prompt reached the agent as its value. A `\r` or `\n` cannot be escaped for a `.cmd` invocation at all, so cmd.exe truncated the command line at the break and the agent silently received partial instructions.

## Expected Behavior

The agent starts for every migration step, whatever its generator produced, and its instructions arrive intact.

nx writes both prompts to files in the run directory and puts only a pointer to them on the command line, so its length no longer tracks the size of the migration. A prompt carrying `%` reaches the agent unchanged, and an argument carrying a line break stops the step with an error instead of being truncated in silence.

## Related Issue(s)

Fixes #36551

NXC-4760

## Implementation Notes

- Claude Code reads the system prompt with `--system-prompt-file`, opencode through the `{file:...}` substitution its config supports. cmd.exe drops any inherited variable over 8,191 characters too, so the environment was no way out.
- opencode expands `{env:...}` and then `{file:...}` over its config text, and parses the JSON afterwards. nx points it at the file only when the path comes back out of that unchanged, and inlines the system prompt otherwise, writing every `{` as `\u007b` so that no pattern can open. The instructions are never inlined, which bounds that value.
- codex cannot load instructions from a file, so it carries a reduced system context inline: the handoff contract plus a pointer at the rest. nx TOML-encodes it and round-trips it through a parser, because codex falls back to raw text on a parse failure instead of reporting one.
- The runner measures the line before dispatching. When it does not fit, agents carrying an inline context swap in a shorter one. When that still does not fit, the step stops with an explanation.
- `%` goes through the `%%cd:~,%` substitution Rust's standard library uses, with `/e:on` and `/v:off` passed so neither expansion mode is inherited from a machine-wide registry setting.
- Prompts travel as files on every platform, so there is one path to keep working.

> [!NOTE]
> nx has no Windows CI for this flow, so no step has run through a real `cmd.exe`. A suite drives the real prompt builders, each agent's real arguments and the real adapter at a 260-character `MAX_PATH` workspace root, asserting the line stays in budget, carries no line break, and does not grow with the generator's output.

---
<p><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.ops.cloud.nx.app/polygraph/session-logo-v4-dark.svg">Polygraph</picture> <a href="https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/nxc-4760-6de92ebb">View session ↗</a></p>

### DIFF
--- a/packages/nx/src/command-line/migrate/agentic/definitions.spec.ts
+++ b/packages/nx/src/command-line/migrate/agentic/definitions.spec.ts
@@ -121,9 +121,9 @@ describe('codexDefinition', () => {
     });
   });
 
-  // codex parses `-c key=value` as TOML and, on a parse failure, silently
-  // falls back to the raw text with quotes trimmed, so an encoding that does
-  // not survive a real TOML parse would ship a mangled system context.
+  // codex parses `-c key=value` as TOML and, on a parse failure, falls back to
+  // the raw text as a literal, so an encoding that does not survive a real TOML
+  // parse would ship a mangled system context.
   it.each([
     ['embedded newlines', 'line1\nline2\n\nline4'],
     ['carriage returns', 'line1\r\nline2'],
@@ -207,13 +207,10 @@ describe('opencodeDefinition', () => {
     expect(opencodeDefinition.wellKnownPaths()).toEqual([]);
   });
 
-  // opencode expands `{env:<name>}` over the raw config text, then
-  // `{file:<path>}`, and parses the JSON only afterwards (1.18.27,
-  // `packages/opencode/src/config/variable.ts`). The env pass splices its value
-  // in unescaped, so what nx builds has to survive it, which a `JSON.parse`
-  // round trip does not show. Only that pass is reproduced here, alongside the
-  // file token opencode would look up: these tests are about which file the
-  // config names, not what reading it would return.
+  // opencode expands `{env:<name>}` then `{file:<path>}` over the raw config
+  // text and parses the JSON afterwards (1.18.27, `config/variable.ts`). The
+  // env pass splices unescaped, which a `JSON.parse` oracle would not show, so
+  // only that pass is reproduced here; the file token is reported, not read.
   function readEnvExpandedConfig(
     configContent: string,
     env: Record<string, string> = {}
@@ -269,10 +266,9 @@ describe('opencodeDefinition', () => {
     }
   );
 
-  // The inlined prompt quotes the workspace root and the handoff path, so a
-  // workspace directory named after a substitution pattern lands inside it.
-  // Expansion must not rewrite the prompt nx wrote, and must not break the
-  // document when the value it would splice in carries a quote of its own.
+  // The inlined prompt quotes the workspace root, so a directory named like a
+  // substitution pattern lands inside it. Expansion must neither rewrite the
+  // prompt nor break the JSON with a quote of its own.
   it.each([
     ['a plain value', 'replaced'],
     ['a value with a quote and a backslash', 'a"b\\c'],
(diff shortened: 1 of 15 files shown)


leosvelperez:
We'll address this in a separate PR as part of a rework to agentic orchestration.

FrozenPandaz (approved):

PR #36586, opened 2026-08-05
feat(core): let the daemon and plugin workers run inside agent sandboxes

AgentEnder:
## Current Behavior

Nx opens unix sockets for the daemon, plugin workers and forked tasks. Agent sandboxes block that, so those features fail inside Claude Code, Codex and the Copilot CLI. The failure does not say what was refused or what to change, and `configure-ai-agents` writes no sandbox settings at all.

## Expected Behavior

`configure-ai-agents` writes the allowances Claude Code needs into `.claude/settings.json`: the Nx socket roots on `allowUnixSockets`, read and write on the same roots, and the analytics domain. Every entry is the same string on any machine, so the file stays shareable.

The daemon and plugin isolation now stay on inside a sandbox rather than being switched off.

When a socket is refused, the hint says permission was denied and names both the setting the detected agent gates sockets on and the file it goes in. A sandbox is the usual cause but not the only one, so it is one entry in the remedy list rather than the first sentence. Codex gates on `network_access`, which also grants general network access and has no path-scoped equivalent, so Nx offers it rather than writing it.

The Copilot CLI is detected as its own agent instead of being reported as the VS Code extension, in `nx` and in `create-nx-workspace`.

`allowAllUnixSockets` is deliberately not written. Testing against Claude Code 2.1.241 showed the scoped entry already permits binding, while the blanket grant only widens connect access to every socket on the machine.

## Related Issue(s)

Fixes NXC-4026

https://linear.app/nxdev/issue/NXC-4026/sandbox-setup-allowunixsockets-settings-for-llms-sandbox-during

Also carries the Copilot CLI detection tracked by NXC-4622.

---

[View session information ↗](https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/Apply-nrwl-nx-PR-review-fixups-d67768f8)


AgentEnder:
## Sandbox testing in a fixture workspace — `allowAllUnixSockets: true` looks unnecessary

I ran the generator from this branch against a throwaway workspace and then drove **Claude Code 2.1.241 headless (`claude -p`)** against the generated settings, varying one key at a time. The probe is a plain Node script that does what the daemon does: `mkdirSync` the socket root, `net.createServer().listen(sock)`, then `net.connect(sock)`.

| Variant | `filesystem.allowRead/Write` | `network.allowUnixSockets` | `network.allowAllUnixSockets` | mkdir | **bind** | connect |
|---|---|---|---|---|---|---|
| A — control | – | – | – | OK | **EPERM** | – |
| D | `/tmp/.nx`, `~/.nx` | – | – | OK | **EPERM** | – |
| C | `/tmp/.nx`, `~/.nx` | `/tmp/.nx`, `~/.nx` | – | OK | **OK** | OK |
| B — as generated | `/tmp/.nx`, `~/.nx` | `/tmp/.nx`, `~/.nx` | `true` | OK | **OK** | OK |

Two things fall out:

**1. The allowances work.** A vs B is the headline: without them the daemon's socket bind dies with `EPERM`, with them it succeeds. That is exactly the failure this PR exists to fix, reproduced and then fixed under a real sandboxed agent. Variant D also confirms the filesystem grant alone is not enough — socket creation is a distinct seatbelt operation — so `allowUnixSockets` is genuinely load-bearing and not redundant with `allowWrite`.

**2. But the scoped entry is create-capable, so the blanket grant is not needed.** Variant C is the important one: with `allowAllUnixSockets` removed and only the scoped `['/tmp/.nx', '~/.nx']` list present, **bind and connect both succeed**. Reproduced 3/3.

That contradicts the premise of 84fd7e4191:

> Claude Code's scoped allowlist permits *connecting* to a socket that already exists; it does not permit *creating* one.

On 2.1.241 the scoped allowlist does permit creating. I can't tell from here whether that was always true or whether Claude Code gained it between early August and now — the commit may well have been correct when written. Worth re-checking against whatever version you consider the floor before acting.

### Why it's worth acting on

`allowAllUnixSockets: true` lets every sandboxed command connect to **every** Unix socket on the machine — `/var/run/docker.sock` (which is root-equivalent and a container-escape primitive), the SSH agent socket, database sockets, and so on. `configure-ai-agents` writes it into `.claude/settings.json`, the **committed** file, so it lands in the repo for everyone who checks it out, next to a scoped list that makes the grant read narrower than it is.

The scoped pair is what the ticket actually asks for ("point to the common socket directory for nx"), and per the test it is sufficient.

### Suggested change

Drop `allowAllUnixSockets` from the generator, keep the scoped `allowUnixSockets` + the filesystem grants, and update the spec plus the KB page (`astro-docs/.../nx-sandbox-unix-sockets.mdoc` currently tells users all three entries are written). If the blanket grant has to stay for a Claude Code version you still support, it would be worth a comment naming that version and the behaviour it works around, since the current comment asserts the scoped entry cannot create sockets and that is not what this build does.

### Repro

```bash
# generate into a fixture, then for each variant edit .claude/settings.json and:
claude -p "Run exactly this command and show me its complete raw output, nothing else: node ./socket-probe.js" \
  --allowedTools Bash --output-format text < /dev/null
```
`sandbox.enabled: true` + `autoAllowBashIfSandboxed: true` were added by hand for the test — worth noting the generator never writes `sandbox.enabled`, so these allowances are inert until the user turns the sandbox on themselves. That's probably intended, but it means a user following the KB page without running `/sandbox` sees no change.

AgentEnder:
Ran this repo's review agents against the PR in an isolated sandbox (four lanes: implementation,
verification, approach, security), plus a set of measurements taken against the real agents. Four
findings came back at Critical. Two are fixed, two are accepted. Recording the accepted ones here so
they are not mistaken for something nobody looked at.

## Accepted, not fixed

**The generated `allowUnixSockets` grant reaches the daemon socket, and the daemon runs code by
design.** A sandboxed agent that can open `/tmp/.nx/<uid>/sockets/…` can send `PROCESS_IN_BACKGROUND`,
which does `require(payload.requirePath)`. The reviewer verified it executing. The daemon a developer
started in an ordinary terminal is not inside the sandbox, so this is code execution outside it.

This is not something the allowlist can fix: the daemon socket is exactly what has to be reachable
for the daemon to work at all, which is the point of the PR. Constraining `requirePath` is a separate
change to the daemon's own RPC surface, and `PROCESS_IN_BACKGROUND` exists to run code. Accepted as a
known property of the daemon rather than something this PR introduces.

**The `allowWrite` grant covers the native binary cache, whose freshness gate is size plus mtime.**
Both are forgeable, verified: a same-size, newer-mtime file with different bytes is accepted and then
loaded by the next `nx` run. The 0700 per-uid directory does not help, because a sandboxed agent runs
as the same uid.

Accepted for the same reason. Same-uid substitution needs a targeted attack, and defending the
process against code running as the same user is not what an agent sandbox is for.

Both predate this PR: `handle-process-in-background.ts` and `native/index.js` are unchanged here. What
the diff changes is reach, and only for a workspace whose owner ran `nx configure-ai-agents` and
turned sandboxing on.

## Fixed

**A sandbox socket failure disabled the daemon permanently, including outside the sandbox.**
Enabling the daemon under `isSandbox()` routed a refusal into the branch that calls
`markDaemonAsDisabled`, which writes `.nx/workspace-data/d/disabled`. That is read on every later run
of the workspace, so one blocked command turned the daemon off for the checkout everywhere, silently,
until `nx reset` — including after the user had done exactly what Nx told them to do. The
neighbouring `daemonPermissionException` already argues this case in a comment.

Now `disableDaemonForThisProcess` turns it off in memory and leaves no marker. A refusal describes
the environment, not the workspace.

The plugin fallback got the same treatment, which also fixed a second problem the review found: the
decision was per-plugin, so every plugin spawned a worker, waited for it to die and reprinted the
advice. It latches once per process now.

**`reference/nx-daemon.mdoc` told readers the opposite of what Nx does.** Two sentences still said the
daemon is disabled in sandboxed environments, which was true before this PR. The reader that
misinforms is the one the change is for.

Also corrected two `certain: true` assertions the review caught: `plugin-worker.ts` asserted a sandbox
on the errno alone, and `get-plugins.ts` asserted one with no errno at all, against the contract
written for that option in this same PR.

## Checked and cleared

`isSandbox()` does return true under Claude Code — two reviewers flagged that the Claude-facing half
would be inert otherwise. Measured inside a sandboxed session: `SANDBOX_RUNTIME=1`.

packages/create-nx-workspace/src/utils/ai/ai-output.ts
```diff
@@ -17,6 +17,7 @@ export function isAiAgent(): boolean {
     _isAiAgent =
       isClaudeCode() ||
       isOpenCode() ||
+      isCodex() ||
```
FrozenPandaz, line 20: Also detect copilot CLI

packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
```diff
@@ -432,17 +477,21 @@ export async function setupAiAgentsGeneratorImpl(
...
+
+function appendAllMissing(
+  existing: string[] | undefined,
+  values: readonly string[]
+): string[] {
+  return values.reduce<string[]>(
+    (acc, value) => appendIfMissing(acc, value),
+    existing ?? []
+  );
 }
```
FrozenPandaz: Use a Set so that we don't have to dedupe

packages/nx/src/utils/nx-tmp-dir.ts
```diff
@@ -25,6 +31,23 @@ export const NX_USER_TMP_DIR =
...
+ * generator. These two spellings are identical on every machine and are
+ * expanded by the sandbox per user, which is the property that makes the entry
+ * worth committing at all.
+ *
+ * Both are listed because socket resolution walks a chain: `/tmp/.nx` first,
+ * `~/.nx` when a peer already owns the shared container. Allowing only the
+ * first leaves every user on a machine where someone else got there first
+ * silently uncovered.
+ */
+export const NX_ALLOWLIST_ROOTS = ['/tmp/.nx', '~/.nx'] as const;
```
FrozenPandaz: Use the `NX_TMP_DIR_POSIX` defined above. Move this into the ai agents config

packages/nx/src/daemon/sandbox-socket-hint.ts
```diff
@@ -0,0 +1,115 @@
...
+}: SandboxSocketHintOptions = {}): string[] {
+  const roots = socketRoots();
+  return [
+    certain
+      ? `Your sandbox is blocking unix socket access. Nx creates its sockets under ${roots}.`
+      : `A sandbox blocking unix socket access is a common cause. Nx creates its sockets under ${roots}.`,
+    'To fix this, do one of the following:',
+    ...(isAiAgent()
+      ? [
+          '  - Run `nx configure-ai-agents` to write the sandbox allowances Nx needs (Claude Code today). Agent sandboxes usually block writes to their own settings file, so this may have to be run from a regular terminal rather than through the agent.',
```
FrozenPandaz: Remove (Claude Code today)

packages/nx/src/daemon/sandbox-socket-hint.ts
```diff
@@ -0,0 +1,115 @@
...
+      return [
+        `  - Add ${roots} to \`sandbox.network.allowUnixSockets\` and to \`sandbox.filesystem.allowRead\`/\`allowWrite\` in your Claude Code settings. The scoped entry covers creating sockets, not only connecting to them.`,
+      ];
+    case 'copilot-cli':
+      return [
+        `  - Add ${roots} to \`sandbox.userPolicy.filesystem.readwritePaths\` in \`~/.copilot/settings.json\`. Copilot CLI reads sandbox policy only from that user-level file, so a copy committed to the workspace has no effect.`,
+      ];
+    case 'codex':
+      return [
+        `  - Codex only permits unix sockets when \`sandbox_workspace_write.network_access\` is true, and has no setting that scopes them to a path. Enabling it also grants network access, so prefer one of the options below unless you want that. If you do enable it, add ${roots} to \`sandbox_workspace_write.writable_roots\` as well.`,
```
FrozenPandaz: Tell the user the path

packages/nx/src/daemon/sandbox-socket-hint.ts
```diff
@@ -0,0 +1,115 @@
...
+      ? [
+          '  - Run `nx configure-ai-agents` to write the sandbox allowances Nx needs (Claude Code today). Agent sandboxes usually block writes to their own settings file, so this may have to be run from a regular terminal rather than through the agent.',
+        ]
+      : []),
+    ...sandboxSpecificRemedy(roots),
+    '  - Set NX_SOCKET_DIR to a directory your sandbox allows.',
+    // The one remedy that needs no sandbox change at all. Worth naming
+    // explicitly: a denied bind is fatal to plugin isolation, so without this
+    // an agent that cannot edit its own sandbox config has no way forward.
+    '  - Keep going without sockets: NX_ISOLATE_PLUGINS=false runs plugins in the main process, and NX_DAEMON=false skips the daemon. Both cost performance but need no sandbox changes.',
```
FrozenPandaz: We don't need this line

packages/nx/src/daemon/sandbox-socket-hint.ts
```diff
@@ -0,0 +1,115 @@
...
+      ? `Your sandbox is blocking unix socket access. Nx creates its sockets under ${roots}.`
+      : `A sandbox blocking unix socket access is a common cause. Nx creates its sockets under ${roots}.`,
+    'To fix this, do one of the following:',
+    ...(isAiAgent()
+      ? [
+          '  - Run `nx configure-ai-agents` to write the sandbox allowances Nx needs (Claude Code today). Agent sandboxes usually block writes to their own settings file, so this may have to be run from a regular terminal rather than through the agent.',
+        ]
+      : []),
+    ...sandboxSpecificRemedy(roots),
+    '  - Set NX_SOCKET_DIR to a directory your sandbox allows.',
```
FrozenPandaz: Set NX_SOCKET_DIR to a directory that you have permissions for.. or get permissions for \[socketdir\]

packages/nx/src/daemon/sandbox-socket-hint.ts
```diff
@@ -0,0 +1,115 @@
...
+ * client's module-level `DaemonClient` singleton.
+ */
+export function sandboxSocketHint({
+  certain = false,
+}: SandboxSocketHintOptions = {}): string[] {
+  const roots = socketRoots();
+  return [
+    certain
+      ? `Your sandbox is blocking unix socket access. Nx creates its sockets under ${roots}.`
+      : `A sandbox blocking unix socket access is a common cause. Nx creates its sockets under ${roots}.`,
```
FrozenPandaz: This should point out permission error rather than assuming it is a sandbox

packages/nx/src/daemon/client/client.ts
```diff
@@ -1134,7 +1135,15 @@ export class DaemonClient {
...
+          error = daemonProcessException(
+            [
+              'The Daemon Server is not running',
+              // A denied bind leaves no socket file behind, so the client sees
+              // a missing socket rather than a refused connection. Outside a
+              // sandbox this is just the ordinary "not started yet" case.
+              ...(isSandbox() ? sandboxSocketHint() : []),
+            ].join('\n')
+          );
         } else if (isPermissionErrno(err as NodeJS.ErrnoException)) {
```
FrozenPandaz: Validate that the EPERM does not fall into this case.. and it is indeed an ENOENT

packages/nx/src/daemon/client/client.ts
```diff
@@ -1553,7 +1568,7 @@ export function daemonPermissionException(socketPath: string, cause: string) {
       `Socket: ${socketPath}`,
       '',
       'Most often the socket belongs to a different user: a daemon left behind by running Nx under `sudo`, a different uid inside a container, or a working copy shared between accounts. If the socket is your own, a sandbox is refusing the connection instead.',
-      'If it belongs to another user, delete the socket above or set NX_SOCKET_DIR to a directory only your user can reach. If you are in a sandbox, allow unix sockets under the Nx socket root — in Claude Code a scoped `allowUnixSockets` only permits connecting, so starting a daemon there needs `allowAllUnixSockets: true`. See https://nx.dev/docs/kb/nx-sandbox-unix-sockets',
+      'If it belongs to another user, delete the socket above or set NX_SOCKET_DIR to a directory only your user can reach. If you are in a sandbox, allow unix sockets under the Nx socket root: `nx configure-ai-agents` writes that for Claude Code, and the agent-specific setting is listed at https://nx.dev/docs/kb/nx-sandbox-unix-sockets',
```
FrozenPandaz: Remove.. "for Claude Code"

packages/nx/src/ai/utils.ts
```diff
@@ -298,3 +298,26 @@ export async function configureAgents(
...
+
+/**
+ * Detects a filesystem permission error thrown while writing agent
+ * configuration files. Agent harnesses (e.g. Claude Code) deny writes to
+ * their own settings files (such as .claude/settings.json) from sandboxed
+ * shell commands, so when `nx configure-ai-agents` or `nx init` runs inside
+ * an agent's bash tool the flush fails with EPERM even though the same
+ * command succeeds in a regular terminal.
+ */
+export function isPermissionWriteError(error: unknown): boolean {
```
FrozenPandaz: Do we have this somewhere else.. or can other places use this function? It's probably more general than an ai util

packages/nx/src/command-line/init/init-v2.ts
```diff
@@ -431,27 +444,43 @@ async function runInit(
...
   }
 
+  // AI agent mode configures the detected agent automatically (aiAgents is
+  // pre-filled above); interactive humans are always prompted — even in the
+  // minimum setup — since agent configuration also writes the sandbox
+  // allowances the Nx daemon needs when driven by an agent later.
   const selectedAgents = await determineAiAgents(
     options.aiAgents,
-    options.interactive && guided
+    options.interactive
```
FrozenPandaz, line 450: Why did this condition change?

FrozenPandaz (changes requested):

packages/nx/src/ai/set-up-ai-agents/set-up-ai-agents.ts
AgentEnder: Done. `appendIfMissing` and `appendAllMissing` are one `withEntries` now, and it collapses duplicates the user's file already had.

packages/nx/src/utils/nx-tmp-dir.ts
AgentEnder: Moved to `ai/constants.ts` and built from `NX_TMP_DIR_POSIX`. The generator was its only consumer, so it had no business on the native binding loader's path.

packages/nx/src/daemon/sandbox-socket-hint.ts
AgentEnder: Removed.

packages/nx/src/daemon/sandbox-socket-hint.ts
AgentEnder: Each remedy names its file now: `.claude/settings.json`, `~/.copilot/settings.json`, `~/.codex/config.toml`.

The KB page said `.codex/config.toml`. Codex loads from `~/.codex/config.toml` or `$CODEX_HOME`, checked against codex-cli 0.149.1, so anyone following that page wrote the setting where nothing reads it. Fixed there too.

packages/nx/src/daemon/sandbox-socket-hint.ts
AgentEnder: Dropped. The Codex remedy pointed at it as "the options below", so that wording went with it.

packages/nx/src/daemon/sandbox-socket-hint.ts
AgentEnder: Now reads: `Get permission to create sockets under <roots>, or set NX_SOCKET_DIR to a directory you do have permission for.`

packages/nx/src/daemon/sandbox-socket-hint.ts
AgentEnder: Reworded. It leads with `Nx was denied permission to use its unix socket under <roots>`, which is the part Nx can observe, and the sandbox is one entry in the remedy list.

That changed what `certain` means, so `plugin-worker.ts` no longer conjoins `isSandbox()` to earn it. The errno alone proves denied permission.

packages/nx/src/daemon/client/client.ts
AgentEnder: It does not, but only by branch order. The ENOENT arm matched the message prefix `connect ENOENT`, so an EPERM missed it by luck of ordering rather than by test. Both are errno tests now and the permission check runs first.

packages/nx/src/daemon/client/client.ts
AgentEnder: Removed. It says "the agents it supports" now.

packages/nx/src/ai/utils.ts
AgentEnder: Agreed, and there were three copies: `isPermissionErrno` in the daemon client, `isPermissionWriteError` here, and an inline `refusedByOs` in the plugin worker.

All three are now `isPermissionDenied` in `utils/permission-errors.ts`, import-free so the plugin worker entry point can take it.

packages/nx/src/command-line/init/init-v2.ts
AgentEnder: Deliberate, and it is the one hunk here that is not about sockets. Agent configuration is also what writes the sandbox allowances, so a minimum-setup user who later drives Nx from an agent never gets them and hits the exact failure this PR fixes.

Split it into its own PR if you would rather keep this one to sockets.

packages/create-nx-workspace/src/utils/ai/ai-output.ts
AgentEnder: Added `isCopilotCli()` on `COPILOT_CLI`, the same variable `detect_ai_agent` reads on the Rust side, and wired it into `isAiAgent()` and `detectAiAgentName()`. It also goes on the daemon env exclusion list, where it took the slot of a duplicate `CODEX_THREAD_ID`.

packages/nx/src/daemon/sandbox-socket-hint.ts
```diff
@@ -0,0 +1,111 @@
...
+      return [
+        `  - Add ${roots} to \`sandbox.userPolicy.filesystem.readwritePaths\` in \`~/.copilot/settings.json\`. Copilot CLI reads sandbox policy only from that user-level file, so a copy committed to the workspace has no effect.`,
+      ];
+    case 'codex':
+      return [
+        `  - In \`~/.codex/config.toml\`, set \`sandbox_workspace_write.network_access\` to true and add ${roots} to \`sandbox_workspace_write.writable_roots\`. Codex has no setting that scopes sockets to a path, and \`network_access\` also grants general network access, so this is a broader grant than the other agents need.`,
+      ];
+    default:
+      return [
+        `  - Allow your sandbox to create unix sockets under ${roots}, not only connect to them, and grant read and write on those directories.`,
```
FrozenPandaz: Don't specify sandbox here

FrozenPandaz (approved):

PR #36614, opened 2026-08-09
fix(release): allow pnpm v11+ to publish without npm on PATH

### LINKED ISSUE #36613: nx release publish (@nx/js:release-publish) hard-requires the npm binary, breaking pnpm v11+ environments that ship no npm
## Current Behavior

`@nx/js:release-publish` (`packages/js/src/executors/release-publish/release-publish.impl.ts`) hard-requires the `npm` binary for every package manager except `bun`:

```ts
try { isNpmInstalled = execSync('npm --version', …) !== ''; }
catch {
  if (pm !== 'bun') {
    throw new Error(`npm was not found in the current environment. This is only supported when using \`bun\` …`);
  }
}
```

`npm` is then used for `npm view` (~line 200) and `npm dist-tag add` (~line 240) even when the detected package manager is `pnpm` and the actual publish runs `pnpm publish`.

## Why this breaks now

Since **pnpm v11.0.0**, installing a Node.js runtime via `pnpm runtime set node` — and the new `pnpm/setup` GitHub Action, which uses it — intentionally does **not** extract the bundled `npm`, `npx`, or `corepack` (https://pnpm.io/cli/runtime#supported-runtimes). Any workflow that provisions Node via `pnpm/setup@v2` (`runtime: node@…`) therefore has no `npm` on PATH, and `nx release publish` fails immediately with _"npm was not found in the current environment"_, even though the workspace uses pnpm and never invokes `npm publish`.

## Expected Behavior

`nx release publish` should succeed in a pnpm-managed workspace that ships no `npm`, since `pnpm publish`, `pnpm view`, and `pnpm dist-tag add` are all available. The `npm` binary should only be required for package managers whose publish is npm-backed (`npm` itself, and `yarn` whose publish command is `npm publish`).

## Reproduction

1. pnpm workspace (pnpm v11+)
2. Node provisioned via `pnpm/setup@v2` with `runtime: node@22`
3. Run `nx release publish` → `@nx/js:release-publish` returns `{ success: false }` with _"npm was not found in the current environment"_.

## Affected versions

Confirmed present in **22.7.8**, latest stable **23.1.1**, and unchanged on **`master`**. (pnpm v10 and below are unaffected in practice because they always ship npm alongside; the npm-less scenario only arises with pnpm v11+.)

## Suggested fix

Only hard-require `npm` for `npm`/`yarn`. For `pnpm` v11+ and `bun`, fall back to the package manager's own registry-metadata commands (`pnpm view` / `pnpm dist-tag add`) and its own `config get` for `.npmrc` resolution when `npm` is absent.

aqeelat:
## Current Behavior

`@nx/js:release-publish` hard-required the `npm` binary for every package manager except `bun`. Since pnpm v11.0.0, provisioning Node via `pnpm runtime set node` / the `pnpm/setup` action intentionally does not extract the bundled `npm`, so `nx release publish` failed immediately in such environments even though the workspace uses pnpm and never needs `npm publish`.

## Expected Behavior

For pnpm v11+ the entire publish flow runs through pnpm — `pnpm view`, `pnpm dist-tag add`, `pnpm config get`, `pnpm publish` — and npm is never invoked, even when it is installed. `npm` is only required for package managers whose publish is npm-backed (`npm`, `yarn`); pnpm v10 and below still require npm; bun keeps npm solely as an auth-error fallback.

## Summary of changes

- `release-publish.impl.ts`: `usePnpmNatively` gate (pnpm v11+ via shared `isPnpmV11Plus()`); the `npm --version` probe is skipped entirely; view/dist-tag commands and error labels route through pnpm.
- `npm-config.ts`: exported `isPnpmV11Plus()` helper; `getNpmConfigValue` uses `pnpm config get` for pnpm v11+, `npm config get` otherwise.
- Tests (38 passing): pnpm v11+ with and without npm on PATH (asserting zero npm invocations), pnpm v10 without npm fails, yarn without npm fails, `isPnpmV11Plus` unit tests, config resolution via the correct binary.

## Related Issue(s)

Fixes #36613

### DIFF
--- a/e2e/release/src/first-release.test.ts
+++ b/e2e/release/src/first-release.test.ts
@@ -125,7 +125,10 @@ describe('nx release first run', () => {
       ).toEqual(1);
       expect(
         releaseOutput1.match(
-          new RegExp('Skipped npm view because --first-release was set', 'g')
+          new RegExp(
+            'Skipped (?:npm|pnpm) view because --first-release was set',
+            'g'
+          )
         ).length
       ).toEqual(3);
 
@@ -160,7 +163,10 @@ describe('nx release first run', () => {
       ).toEqual(1);
       expect(
         releaseOutput2.match(
-          new RegExp('Skipped npm view because --first-release was set', 'g')
+          new RegExp(
+            'Skipped (?:npm|pnpm) view because --first-release was set',
+            'g'
+          )
         ).length
       ).toEqual(3);
     });
@@ -196,7 +202,10 @@ describe('nx release first run', () => {
       ).toEqual(1);
       expect(
         releaseOutput1.match(
-          new RegExp('Skipped npm view because --first-release was set', 'g')
+          new RegExp(
+            'Skipped (?:npm|pnpm) view because --first-release was set',
+            'g'
+          )
         ).length
       ).toEqual(3);
 
@@ -231,7 +240,10 @@ describe('nx release first run', () => {
       ).toEqual(1);
       expect(
         releaseOutput2.match(
-          new RegExp('Skipped npm view because --first-release was set', 'g')
+          new RegExp(
+            'Skipped (?:npm|pnpm) view because --first-release was set',
+            'g'
+          )
         ).length
       ).toEqual(3);
     });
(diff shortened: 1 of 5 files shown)


aqeelat:
Fixed reported issues.

@AgentEnder @FrozenPandaz can you please approve the workflow?

PR #36632, opened 2026-08-11
fix(angular): import optional `ng-packagr` lazily

BenjaminBrienen-CanonPP:
`@nx/angular/executors` eagerly loads the package executor path, which loads `ng-packagr` at module evaluation time via `packages/angular/src/executors/package/ng-packagr-adjustments/ng-packagr.ts`.

## Current Behavior
Today this can break Angular application builds in downstream wrappers that import `@nx/angular/executors` only to call the webpack browser builder, because the barrel import evaluates the package executor path and throws `Cannot find module 'ng-packagr'` before any library-packaging code is selected.

## Expected Behavior
- app-build imports from `@nx/angular/executors` should work without `ng-packagr`
- `ng-packagr` should only be required when `@nx/angular:package` or related library packaging executors actually run

## Related Issue(s)
none


leosvelperez (approved):
Thanks!

PR #36638, opened 2026-08-11
chore(repo): sort tailwind classes to match prettier

AgentEnder:
## Current Behavior

50 `.tsx` files across `nx-dev`, `graph` and `astro-docs` have drifted from `prettier-plugin-tailwindcss` output. They pass CI today only because `nx format:check` gates *changed* files — so the drift is invisible until an unrelated change touches one of them, at which point that PR is forced to carry an unrelated reformat.

That is exactly what happened in #36624: a two-line import fix in `nx-dev/ui-blog/src/lib/blog-details.tsx` had to ship with three unrelated `className` reorderings to keep `format:check` green.

## Expected Behavior

The drift is cleared, so touching any of these files no longer conscripts the author into reformatting it.

The cause is **Tailwind v4 sort order** — v4 reordered the inset and layout groups, so the plugin now sorts differently than when these files were last written:

```diff
-<header className="mx-auto mb-16 mt-8 px-4">
+<header className="mx-auto mt-8 mb-16 px-4">

-'pointer-events-none absolute left-0 top-0 transform-gpu stroke-2'
+'pointer-events-none absolute top-0 left-0 transform-gpu stroke-2'

-'whitespace-nowrap border-b-2 py-2 text-sm font-medium'
+'border-b-2 py-2 text-sm font-medium whitespace-nowrap'
```

## Why this is safe to merge on sight

**Formatting only — 97 insertions, 97 deletions.** Every changed line is a rewrite of an existing line; none is added or removed.

**Every changed line was verified to be a pure permutation.** For all 97 changed line pairs, the multiset of whitespace-separated class tokens is identical before and after — no class added, dropped, or altered. Reordering utility classes cannot change rendered output: Tailwind emits each utility once, and precedence comes from the generated stylesheet's order, not from the order in `className`.

## Scope notes

- **`nx-dev/ui-blog/src/lib/blog-details.tsx` is deliberately excluded** — #36624 already reformats it. The two PRs are disjoint by file, so merge order does not matter.
- **`packages/gradle/batch-runner/build/nx/gradle-batch-runner.json` is excluded** — `nx format:check --all` flags it, but it is an untracked, gitignored build artifact. Worth knowing that `--all` does not respect `.gitignore`.

## Verification

| Check | Scope | Result |
| --- | --- | --- |
| Pure-permutation check | 97 changed line pairs | all identical token multisets |
| Insertions / deletions | 50 files | 97 / 97 |
| `nx format:check --all` | whole repo | clean* |
| `lint` | 16 owning projects, 149 tasks | 0 errors |

\* apart from `blog-details.tsx` (owned by #36624) and the untracked gradle artifact.

## Related Issue(s)

No filed issue. Follow-up to #36624, which surfaced the drift.

---
<p><a href="https://app.trypolygraph.com/orgs/6a061dcb561c062131116eca/sessions/fix-double-exports-c05225fb">View Polygraph session ↗</a></p>

### DIFF
--- a/astro-docs/src/components/PluginDirectory.tsx
+++ b/astro-docs/src/components/PluginDirectory.tsx
@@ -202,7 +202,7 @@ function PluginCard({ plugin }: { plugin: Plugin }) {
       <div className="flex h-full flex-col">
         <div className="mb-3 flex items-center gap-3">
           <GitHubIcon className="h-6 w-6 flex-shrink-0 text-gray-600 dark:text-slate-400" />
-          <h3 className="text-base font-medium leading-tight text-gray-900 dark:text-slate-100">
+          <h3 className="text-base leading-tight font-medium text-gray-900 dark:text-slate-100">
             {plugin.name}
           </h3>
         </div>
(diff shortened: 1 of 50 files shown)


AgentEnder:
Green and mergeable since Aug 11, no review yet. Small scope — applies prettier's tailwind class sort to nx-dev so the formatting check stops drifting. Happy for anyone to pick it up.

jaysoo (approved):

